PR #1538, opened 2022-03-30
CASSANDRA-16456 - Add Plugin Support for CQLSH

bhouse99:
Added plugin support for CQLSH.  Support is available as part of Command line and via additions to the CQLSHRC. file.
For examples see CEP Documentation.

	* JIRA: https://issues.apache.org/jira/browse/CASSANDRA-16456
	* CEP: https://cwiki.apache.org/confluence/display/CASSANDRA/CEP-16%3A+Auth+Plugin+Support+for+CQLSH


bin/cqlsh.py
```diff
@@ -2384,7 +2441,8 @@ def main(options, hostname, port):
                       single_statement=options.execute,
                       request_timeout=options.request_timeout,
                       connect_timeout=options.connect_timeout,
-                      encoding=options.encoding)
+                      encoding=options.encoding,
+                      auth_provider=load_auth_provider(options))
```
maulin-vasavada: Can you please provide any test cases and samples for these changes?

bin/cqlsh.py
```diff
@@ -98,6 +99,33 @@
...
+    if len(splits) > 2:
+        raise optparse.OptionValueError("Invalid --AuthProviderExtended argument only single period should be found: %s" % opt_str)
+    if len(splits) < 2:
+        raise optparse.OptionValueError("Invalid --AuthProviderExtended must provide a param using period: %s" % opt_str)
+    key = splits[1]
+
+    if not hasattr(passed_parser.values, 'auth_provider_extended') or parser.values.auth_provider_extended is None:
+        passed_parser.values.auth_provider_extended = {}
+
+    passed_parser.values.auth_provider_extended[key] = value
```
maulin-vasavada: The name `auth_provider_extended` is little confusing in my opinion. Since it is set of properties that the pluggable AuthProvider can use, would `auth_provider_properties` or `auth_provider_config` make more sense?

bin/cqlsh.py
```diff
@@ -98,6 +99,33 @@
...
+
+
+def add_auth_provider_extended_options(passed_parser, argv):
+    for arg in argv:
+        if arg.startswith('--AuthProviderExtended'):
```
maulin-vasavada: Similar to my previous comment, would it make more sense to name this `AuthProviderProperty`/`AuthProviderConfig` (I believe this is a single property that can be overridden by the command line)?

maulin-vasavada (changes requested):

bin/cqlsh.py
bhouse99: Will add tests in the vein of sslhandling.

bin/cqlsh.py
bhouse99: Recently changed this to auth_provider_
config.

bin/cqlsh.py
bhouse99: Renamed to auth_provider_config.

bin/cqlsh.py
bhouse99: tests added

maulin-vasavada (approved):

bin/cqlsh.py
maulin-vasavada: thanks!

bin/cqlsh.py
maulin-vasavada: thanks!

CHANGES.txt
```diff
@@ -24,6 +24,7 @@
  * Add diagnostic events for guardrails (CASSANDRA-17197)
  * Pre hashed passwords in CQL (CASSANDRA-17334)
  * Increase cqlsh version (CASSANDRA-17432)
+ * Added cqlsh custom auth provider support (CASSANDRA-16456)
```
smiklosovic: @bhouse99 remove this line, I ll take care of that on the merge itself. No worries. We would have to rebase / fix it all over again as other people are adding stuff there in the meanwhile.

bin/cqlsh.py
```diff
@@ -405,7 +405,6 @@ def deserialize_date_fallback_int(byts, protocol_version):
     # Return cassandra.cqltypes.EMPTY instead of None for empty values
     cassandra.cqltypes.CassandraType.support_empty_values = True
 
-
```
smiklosovic: should not be this kept?

pylib/cqlshlib/authproviderhandling.py
```diff
@@ -0,0 +1,79 @@
...
+            return None
+    
+    def get_auth_provider_config_settings():
+        if 'auth_provider_config' in configs.sections():
+            return dict(configs.items("auth_provider_config"))
+        else:
+            return None
+    
+    module_name = get_setting('auth_provider', 'module')
+    class_name = get_setting('auth_provider', 'classname')
```
smiklosovic: what if there is no `class_name` but `module` is there? So `class_name` will be `None` but you are not checking it as you do upon `module_name`. Could you also add a test for that, please?

pylib/cqlshlib/authproviderhandling.py
```diff
@@ -0,0 +1,79 @@
...
+
+import configparser
+from importlib import import_module
+
+def load_custom_auth_provider(config_file = None):
+    """
+    Function which loads a custom auth provider from available config.
+
+    Params:
+    * config_file ..: path to cqlsh config file (usually ~/.cqlshrc).
```
smiklosovic: the default path is actually `~/.cassandra/cqlshrc`

CHANGES.txt
bhouse99: Kk, will do will avoid modifying in the future.

bin/cqlsh.py
```diff
@@ -2091,7 +2095,6 @@ def is_file_secure(filename):
     # This is to allow "sudo cqlsh" to work with user owned credentials file.
     return (uid == 0 or st.st_uid == uid) and stat.S_IMODE(st.st_mode) & (stat.S_IRGRP | stat.S_IROTH) == 0
 
-
```
smiklosovic, line 2094: @bhouse99 this line should not be removed. I think there is 2 lines policy in Python between methods.

bin/cqlsh.py
```diff
@@ -2169,14 +2180,15 @@ def read_options(cmdlineargs, environment):
...
         # use the username from credentials file but fallback to cqlshrc if username is absent from the command line parameters
-        options.username = option_with_default(credentials.get, 'plain_text_auth', 'username', username_from_cqlshrc)
+        options.username = username_from_cqlshrc
 
     if not options.password:
         rawcredentials = configparser.RawConfigParser()
         rawcredentials.read(options.credentials)
 
         # handling password in the same way as username, priority cli > credentials > cqlshrc
         options.password = option_with_default(rawcredentials.get, 'plain_text_auth', 'password', password_from_cqlshrc)
```
smiklosovic: @bhouse99 any reason why you have not removed this line, similarly as it is done for username case above?

bin/cqlsh.py
```diff
@@ -2169,14 +2180,15 @@ def read_options(cmdlineargs, environment):
         credentials.read(options.credentials)
 
         # use the username from credentials file but fallback to cqlshrc if username is absent from the command line parameters
-        options.username = option_with_default(credentials.get, 'plain_text_auth', 'username', username_from_cqlshrc)
+        options.username = username_from_cqlshrc
```
smiklosovic, line 2167: @bhouse99 I do not get this change. So we are populating `options.username` with `username_from_cqlshrc`. So for what reason do we read `credentials.read(options.credentials)` 3 lines above? That seems to me like dead code?

smiklosovic:
merged here https://github.com/apache/cassandra/commit/381c2a4fa82db80543fd67da6669de022d7576c4

PR #1543, opened 2022-03-31
CASSANDRA-17509

jmckenzie-dev:


src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -332,6 +332,20 @@ public void setUserTimestampsEnabled(boolean enabled)
                                   x -> config.user_timestamps_enabled = x);
     }
 
+    @Override
+    public boolean getGroupByEnabled()
```
dcapwell, line 336: should this be `is` rather than `get`, since `boolean` rather than `Boolean`?

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -456,6 +461,18 @@ public void setUserTimestampsEnabled(boolean enabled)
         DEFAULT_CONFIG.setUserTimestampsEnabled(enabled);
     }
 
+    @Override
+    public boolean getGroupByEnabled()
```
dcapwell, line 525: same as other comment, should this be `is` rather than `get`?

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -129,6 +129,11 @@
                     state -> !CONFIG_PROVIDER.getOrCreate(state).getUserTimestampsEnabled(),
                     "User provided timestamps (USING TIMESTAMP)");
 
+    public static final DisableFlag groupByEnabled =
+    new DisableFlag("group_by",
+                    state -> !CONFIG_PROVIDER.getOrCreate(state).getGroupByEnabled(),
+                    "GROUP BY functionality");
```
dcapwell, line 143: for me: `GROUP BY functionality is not allowed` will be the error message

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -129,6 +129,11 @@
                     state -> !CONFIG_PROVIDER.getOrCreate(state).getUserTimestampsEnabled(),
                     "User provided timestamps (USING TIMESTAMP)");
 
+    public static final DisableFlag groupByEnabled =
+    new DisableFlag("group_by",
+                    state -> !CONFIG_PROVIDER.getOrCreate(state).getGroupByEnabled(),
```
dcapwell, line 142: outside the scope of this issue:

Rather than `DisableFlag` maybe we make it `EnabledFlag` and undo the `!` prefix every usage uses...

test/unit/org/apache/cassandra/db/guardrails/GuardrailGroupByTest.java
```diff
@@ -0,0 +1,57 @@
...
+        {
+            assertValid(query);
+        }
+        catch (AssertionError ae)
+        {
+            // We expect the query to _warn_ (i.e. fail the assertValid check), but with a non-guardrails warning and
+            // not a failure. Rather than extending the GuardrailsTester framework to handle this case of "warns but not
+            // with a guardrail warning", we just confirm here we saw the acceptable warning we expect but the query was
+            // allowed to run.
+            Assert.assertTrue(ae.getMessage().contains("used without partition key"));
```
adelapena: Maybe it would be easier to just use a query that doesn't throw a warning like, for example:
```java
private static final String query = String.format("SELECT * FROM %s.%s WHERE keyspace_name='%s' GROUP BY table_name",
                                                  SchemaConstants.SCHEMA_KEYSPACE_NAME,
                                                  SchemaKeyspaceTables.TABLES,
                                                  KEYSPACE);
```
Alternatively, modifying `GuardrailTester#assertValid` to accept non-guardrail warnings should be easy and has potential for reutilization on other tests, for example [this way](https://github.com/adelapena/cassandra/commit/de425dc1ad994aa01eb0f15a4670bbd376fa7b22).

test/unit/org/apache/cassandra/db/guardrails/GuardrailGroupByTest.java
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+}
```
adelapena: Nit: missing newline at the end of the file

test/unit/org/apache/cassandra/db/guardrails/GuardrailGroupByTest.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+package org.apache.cassandra.db.guardrails;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class GuardrailGroupByTest extends GuardrailTester
+{
+    private final String query = "SELECT * FROM system_schema.keyspaces GROUP BY keyspace_name";
```
adelapena: Nit: could be static

test/unit/org/apache/cassandra/db/guardrails/GuardrailGroupByTest.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Test
+    public void checkExplicitlyDisabled() throws Throwable
+    {
+        setGuardrail(false);
+        assertFails(query, "GROUP BY functionality");
+    }
+
+    @Test
+    public void checkEnabled() throws Throwable
```
adelapena, line 53: We could add a test to verify that internal queries and superusers are not affected by the guardrail. If `GuardrailGroupByTest.query` doesn't throw additional warnings it could be:
```java
@Test
public void testExcludedUsers() throws Throwable
{
    setGuardrail(false);
    testExcludedUsers(() -> query);
}
```
Otherwise it would be:
```java
@Test
public void testExcludedUsers() throws Throwable
{
    setGuardrail(false);
    testExcludedUsers(() -> String.format("SELECT * FROM %s.%s WHERE keyspace_name='%s' GROUP BY table_name",
                                          SchemaConstants.SCHEMA_KEYSPACE_NAME,
                                          SchemaKeyspaceTables.TABLES,
                                          KEYSPACE));
}
```

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
adelapena: Makes sense to me, probably every future enabled/disabled guardrail is going to have that negation.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
adelapena: The `what` argument in the constructor of `DisableFlag` doesn't contain the full message but just the name of what is disallowed, the ` is not allowed` part will be added by the guardrail, [here](https://github.com/apache/cassandra/blob/18bae4628c50ae927da7dfa4e7e1fb8648dd765b/src/java/org/apache/cassandra/db/guardrails/DisableFlag.java#L84). So I think this is correct. Maybe we could add the full message to the calls to `assertFails` in the new test to be completely sure:
```java
assertFails(query, "GROUP BY functionality is not allowed");
```

jmckenzie-dev:
```
@Test
public void checkExplicitlyDisabled() throws Throwable
{
    setGuardrail(false);
    assertFails(query, "GROUP BY functionality is not allowed");
}
```

Modified the unit test to check the full string. Passes.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
jmckenzie-dev: I was thinking the same thing. I'll make a note to create a follow up JIRA for that if you're not already on it @adelapena

test/unit/org/apache/cassandra/db/guardrails/GuardrailGroupByTest.java
jmckenzie-dev: Good call. I went ahead and switched to using the group by table_name form of the query instead of jumping through these hoops. If we keep seeing this need to differentiate guardrail vs. non-guardrail warnings we can revisit assertValid's functionality.

test/unit/org/apache/cassandra/db/guardrails/GuardrailGroupByTest.java
jmckenzie-dev: Went ahead and changed the query as mentioned in another comment and included this test. Thanks Andres!

CHANGES.txt
```diff
@@ -1,4 +1,5 @@
 4.1
+ * Add Guardrail for GROUP BY queries (CASSANDRA-17509)
```
adelapena: Super nit: other similar entries in `CHANGES.txt` in use "Add guardrail...", without capitalizing.

adelapena (approved):

PR #1546, opened 2022-04-01
CASSANDRA-17150 trunk: Guardrails for disk usage

adelapena:


src/java/org/apache/cassandra/db/guardrails/Guardrail.java
```diff
@@ -75,6 +85,9 @@ protected void warn(String message)
 
     protected void warn(String message, String redactedMessage)
     {
+        if (skipNotifying(true))
```
adelapena, line 100: I'm a bit on the fence about omitting the client warnings when `minNotifyIntervalInMs` is set. We want that property to not overflow server logs nor emit millions of diagnostic events when there are replicas with stuffed disks, given that a node can remain is that state for a long time and keep accepting writes. But it might be legit to still send the warning to the client, although on the other hand we could also overflow client's logs if we emit client warnings for every write sent to a stuffed node.

adelapena, line 88: I'm a bit on the fence about omitting the client warning when `minNotifyIntervalInMs` is set. We want that property to not overflow server logs nor emit millions of diagnostic events when there are replicas with stuffed disks. But it might be legit to still send the warning to the client, although on the other hand we could also overflow client's logs if we emit client warnings for every write done when a node is stuffed.

src/java/org/apache/cassandra/service/disk/usage/DiskUsageBroadcaster.java
```diff
@@ -0,0 +1,169 @@
...
+            usageState = DiskUsageState.valueOf(value.value);
+        }
+        catch (IllegalArgumentException e)
+        {
+            noSpamLogger.warn(String.format("Found unknown DiskUsageState: %s. Using default state %s instead.",
+                                            value.value, usageState));
+        }
+        usageInfo.put(endpoint, usageState);
+
+        hasStuffedOrFullNode = usageState.isStuffedOrFull() || usageInfo.values().stream().anyMatch(DiskUsageState::isStuffedOrFull);
```
smiklosovic: @adelapena do you think using stream() here is performance friendly / without any penalty when hundreds of nodes are used? Would not be simple for loop better?

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * Start monitoring local disk usage and call notifier when local disk usage state changed.
+     */
+    public void start(Consumer<DiskUsageState> notifier)
+    {
+        // start the scheduler regardless guardrail is enabled, so we can enable it later without a restart
+        ScheduledExecutors.scheduledTasks.scheduleAtFixedRate(() -> {
+
+            if (!Guardrails.localDiskUsage.enabled(null))
```
smiklosovic: @adelapena can't there be a helper method like `Guardrails.localDiskUsage.enabled()`? (without null)

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
```diff
@@ -0,0 +1,174 @@
...
+import org.apache.cassandra.io.util.FileUtils;
+
+/**
+ * Schedule periodic task to monitor local disk usage and notify {@link DiskUsageBroadcaster} if local state changed.
+ */
+public class DiskUsageMonitor
+{
+    private static final Logger logger = LoggerFactory.getLogger(DiskUsageMonitor.class);
+
+    private static final int MONITOR_INTERVAL_MS = Integer.parseInt(System.getProperty(Config.PROPERTY_PREFIX + "disk_usage.monitor_interval_ms",
```
smiklosovic: @adelapena should not this be in `CassandraRelevantProperties`? Then it can be done like "getInt()" once you have that property.

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
adelapena: I've added that utility method.

src/java/org/apache/cassandra/service/disk/usage/DiskUsageBroadcaster.java
adelapena: Changed to a loop

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
adelapena: Good idea, done

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
```diff
@@ -0,0 +1,171 @@
...
+        BigDecimal used = BigDecimal.ZERO;
+        BigDecimal total = BigDecimal.ZERO;
+
+        for (Directories.DataDirectory dir : dataDirectoriesSupplier.get())
+        {
+            used = used.add(BigDecimal.valueOf(dir.getSpaceUsed()));
+            total = total.add(BigDecimal.valueOf(dir.getTotalSpace()));
+        }
+
+        if (config.getDataDiskUsageMaxDiskSize().toGibibytes() > 0)
```
smiklosovic: I can not use anything less than 1GiB for `data_disk_usage_max_disk_size`. Is this what we want? I was trying to set it like to 200MiB and it will be rounded down to zero, so this will not be executed:

        if (config.getDataDiskUsageMaxDiskSize().toGibibytes() > 0)
        {
            long diskUsageMaxSizeInBytes = config.getDataDiskUsageMaxDiskSize().toBytes();
            total = total.min(BigDecimal.valueOf(diskUsageMaxSizeInBytes));
        }

Which means my total will be still total of the disk. For testing purposes, when I am trying to generate something quick in docker or whatever, megabytes are completely enough, no need to do gigabytes. I this we could lower it down to mebibytes maybe, on the other hand, it would be nice if we do not allow it to be super low, like 1MiB, at least 10MiB, because an empty Cassandra node, after bootstrapping, will occupy some space already.

An ideal solution would be to create a startup check for this - it would check if the value you have set makes sense - if it is acutally bigger than the stuff on the disk.

adelapena: Good catch, we should use `config.getDataDiskUsageMaxDiskSize().toBytes()` to avoid rounding problems. I'm also adding JMX methods to allow specifying different units than gibibytes.

src/java/org/apache/cassandra/cql3/statements/BatchStatement.java
```diff
@@ -411,16 +411,22 @@ public ResultMessage execute(QueryState queryState, BatchQueryOptions options, l
...
 
+        ClientState clientState = queryState.getClientState();
         Guardrails.writeConsistencyLevels.guard(EnumSet.of(options.getConsistency(), options.getSerialConsistency()),
-                                                queryState.getClientState());
+                                                clientState);
+
+        for (int i = 0; i < statements.size(); i++ )
+        {
+            statements.get(i).validateDiskUsage(options.forStatement(i), clientState);
+        }
```
ekaterinadimitrova2: nit: I would probably remove the braces of the for loop

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -446,20 +452,23 @@ public void setWriteConsistencyLevelsDisallowed(Set<ConsistencyLevel> consistenc
...
 
     @Override
+    @Nullable
     public DataStorageSpec getCollectionSizeFailThreshold()
     {
         return config.collection_size_fail_threshold;
     }
 
-    public void setCollectionSizeThreshold(DataStorageSpec warn, DataStorageSpec fail)
+    public void setCollectionSizeThreshold(@Nullable DataStorageSpec warn, @Nullable DataStorageSpec fail)
```
ekaterinadimitrova2, line 484: Good call adding @Nullable!

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -387,20 +389,26 @@
...
+    @Nullable
+    String getCollectionSizeFailThreshold();
 
     /**
-     * @param warnInKiB The threshold to warn when encountering larger size of collection data than threshold, in KiB.
-     * @param failInKiB The threshold to prevent collections with larger data size than threshold, in KiB.
+     * @param warnSize The threshold to warn when encountering larger size of collection data than threshold, as a
+     *                 string formatted as in, for example, {@code 10GiB}, {@code 20MiB}, {@code 30KiB} or {@code 40B}.
+     * @param failSize The threshold to prevent collections with larger data size than threshold, as a string formatted
+     *                 as in, for example, {@code 10GiB}, {@code 20MiB}, {@code 30KiB} or {@code 40B}.
```
ekaterinadimitrova2: I know you documented the null being disabled in cassandra.yaml but I would say it is probably worth it to add a point here too.

conf/cassandra.yaml
```diff
@@ -1671,6 +1671,18 @@ drop_compact_storage_enabled: false
 # Default -1 to disable.
 # fields_per_udt_warn_threshold: -1
 # fields_per_udt_fail_threshold: -1
+# Guardrail to warn or fail when local data disk usage percentage exceeds threshold. Valid values are in (1, 100].
```
ekaterinadimitrova2: This is a bit different from dynamic_snitch_badness_threshold where 1.0 means 100%, but I find it better than double numbers so I would encourage it actually. Also why not [1,100]? Do we exclude 1? I don't think so. 
Also, do we plan recommendations on how to choose the values?

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -265,6 +265,9 @@
     TEST_IGNORE_SIGAR("cassandra.test.ignore_sigar", "false"),
     PAXOS_EXECUTE_ON_SELF("cassandra.paxos.use_self_execution", "true"),
 
+    // properties for guardrails
```
ekaterinadimitrova2: I would prefer if we add a JavaDoc, the way it was when I introduced the class. I can see that after that people were adding properties just with comment but I think it is better to have a JavaDoc here

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -558,19 +609,44 @@ private static void validateIntThreshold(int warn, int fail, String name)
...
 
-    private static void validateSizeThreshold(DataStorageSpec warn, DataStorageSpec fail, String name)
+    private static void validateSize(DataStorageSpec size, boolean allowZero, String name)
+    {
+        if (size == null)
+            return;
+
+        if (!allowZero && size.toBytes() == 0)
+            throw new IllegalArgumentException(format("Invalid value for %s: 0 is not allowed; " +
+                                                      "if attempting to disable use null/empty",
```
ekaterinadimitrova2: nit: empty value

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -558,19 +609,44 @@ private static void validateIntThreshold(int warn, int fail, String name)
         validateWarnLowerThanFail(warn, fail, name);
     }
 
+    private static void validatePercentageThreshold(int warn, int fail, String name)
```
ekaterinadimitrova2, line 627: nit: I think we should reorder the methods to keep the code style. 
For instance: `validatePercentage` should be before `validatePositiveNumeric`. Caller before called one, no?

src/java/org/apache/cassandra/db/guardrails/Guardrail.java
ekaterinadimitrova2: I guess if this is configurable with the trade-offs explained we can leave it to the operators to decide what and with what frequency they would like to log depending on their environment? Does it make sense?

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -42,11 +46,11 @@ public final class Guardrails implements GuardrailsMBean
 {
     public static final String MBEAN_NAME = "org.apache.cassandra.db:type=Guardrails";
 
-    private static final GuardrailsConfigProvider CONFIG_PROVIDER = GuardrailsConfigProvider.instance;
+    public static final GuardrailsConfigProvider CONFIG_PROVIDER = GuardrailsConfigProvider.instance;
     private static final GuardrailsOptions DEFAULT_CONFIG = DatabaseDescriptor.getGuardrailsConfig();
 
     @VisibleForTesting
-    static final Guardrails instance = new Guardrails();
+    public static final Guardrails instance = new Guardrails();
```
ekaterinadimitrova2: Did this one become public for test purposes?

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -228,8 +232,8 @@ public final class Guardrails implements GuardrailsMBean
      */
     public static final Threshold collectionSize =
     new Threshold("collection_size",
-                  state -> CONFIG_PROVIDER.getOrCreate(state).getCollectionSizeWarnThreshold().toBytes(),
-                  state -> CONFIG_PROVIDER.getOrCreate(state).getCollectionSizeFailThreshold().toBytes(),
+                  state -> sizeToBytes(CONFIG_PROVIDER.getOrCreate(state).getCollectionSizeWarnThreshold()),
```
ekaterinadimitrova2, line 240: What is the reason to convert to -1 when we use null for disabled?

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -708,6 +749,36 @@ public void setFieldsPerUDTThreshold(int warn, int fail)
         DEFAULT_CONFIG.setFieldsPerUDTThreshold(warn, fail);
     }
 
+    public int getDataDiskUsagePercentageWarnThreshold()
```
ekaterinadimitrova2, line 771: Missing @Override

src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
```diff
@@ -222,4 +226,22 @@
      * @return The threshold to fail when creating a UDT with more fields than threshold.
      */
     int getFieldsPerUDTFailThreshold();
+
+    /**
+     * @return The threshold to warn when local disk usage percentage exceeds that threshold.
+     * Allowed values are in the range {@code (1, 100]}, and -1 means disabled.
```
ekaterinadimitrova2: I believe this needs to be `[1,100]`, not `(1,100]`?

src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
```diff
@@ -222,4 +226,22 @@
...
+     */
+    int getDataDiskUsagePercentageWarnThreshold();
+
+    /**
+     * @return The threshold to fail when local disk usage percentage exceeds that threshold.
+     * Allowed values are in the range {@code (1, 100]}, and -1 means disabled.
```
ekaterinadimitrova2: `[1,100]`

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -433,4 +441,37 @@
...
+     *             Allowed values are in the range {@code (1, 100]}, and -1 means disabled.
+     */
+    public void setDataDiskUsagePercentageThreshold(int warn, int fail);
+
+    /**
+     * @return The max disk size of the data directories when calculating disk usage thresholds, as a string formatted
+     * as in, for example, {@code 10GiB}, {@code 20MiB}, {@code 30KiB} or {@code 40B}.
+     */
+    @Nullable
+    String getDataDiskUsageMaxDiskSize();
```
ekaterinadimitrova2, line 489: I think we should probably mention `null` in the java doc for this and the following method

src/java/org/apache/cassandra/db/guardrails/PercentageThreshold.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.function.ToLongFunction;
+
+import org.apache.cassandra.service.ClientState;
+
+/**
+ * A {@link Threshold} guardrail whose values represent a percentage
+ * <p>
+ * This works exactly as a {@link Threshold}, but provides slightly more convenient error messages for percentage
+ */
+public class PercentageThreshold extends Threshold
```
ekaterinadimitrova2, line 30: Wondering whether this should be final class?

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
```diff
@@ -0,0 +1,229 @@
...
+     * @return local node disk usage state
+     */
+    @VisibleForTesting
+    public DiskUsageState state()
+    {
+        return localState;
+    }
+
+    /**
+     * @return disk usage (including all memtable sizes) ratio
```
ekaterinadimitrova2: disk space occupied by data directories?

ekaterinadimitrova2: So looking into `getUsableSpace`, I feel it will be nice maybe to explain also on our end a bit the approximations for someone newbie? WDYT?

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
```diff
@@ -0,0 +1,229 @@
...
+        }
+
+        // The total disk size for data directories is the space that is actually used by those directories plus the
+        // free space on disk that might be used for storing those directories in the future.
+        BigDecimal total = used.add(usable);
+
+        // That total space can be limited by the config property data_disk_usage_max_disk_size.
+        DataStorageSpec diskUsageMaxSize = guardrailsConfigSupplier.get().getDataDiskUsageMaxDiskSize();
+        if (diskUsageMaxSize != null)
+            total = total.min(BigDecimal.valueOf(diskUsageMaxSize.toBytes()));
```
ekaterinadimitrova2: I am wondering whether we shouldn't workaround this calculation every time, poking the conversion toBytes.... Maybe we should have it somewhere in bytes and use that one and make the JMX to be able to provide only bytes... Thinking about performance which is hard without having performance testing unfortunately. Thoughts?

src/java/org/apache/cassandra/service/disk/usage/DiskUsageState.java
```diff
@@ -0,0 +1,55 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.service.disk.usage;
+
+public enum DiskUsageState
+{
+    NOT_AVAILABLE("Not Available"), // either disk usage guardrail is not enabled or gossip state is not ready
```
ekaterinadimitrova2: Wondering whether we want some java doc here

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailDiskUsageTest.java
```diff
@@ -0,0 +1,225 @@
...
+        return cluster;
+    }
+
+    @Test
+    public void testDiskUsage() throws Throwable
+    {
+        schemaChange("CREATE TABLE %s (k int PRIMARY KEY, v int)");
+        String insert = format("INSERT INTO %s(k, v) VALUES (?, 0)");
+
+        // With both nodes is SPACIOUS state, we can write without warnings nor failures
```
ekaterinadimitrova2:
```suggestion
       // With both nodes in SPACIOUS state, we can write without warnings nor failures
```

conf/cassandra.yaml
```diff
@@ -1671,6 +1671,18 @@ drop_compact_storage_enabled: false
...
+# This is only used for the disks storing data directories, so it won't count any separate disks used for storing
+# the commitlog, hints nor saved caches.
+# The two thresholds default to -1 to disable.
+# data_disk_usage_percentage_warn_threshold: -1
+# data_disk_usage_percentage_fail_threshold: -1
+# Allows defining the max disk size of the data directories when calculating thresholds for
+# disk_usage_percentage_warn_threshold and disk_usage_percentage_fail_threshold, so if this is greater than zero they
+# become percentages of a fixed size on disk instead of percentages of the physically available disk size.
+# Valid values are in (1, max available disk size of all data directories].
```
ekaterinadimitrova2: [1, max available disk size of all data directories]

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
```diff
@@ -0,0 +1,229 @@
...
+            for (Directories.DataDirectory dir : e.getValue())
+                used = used.add(BigDecimal.valueOf(dir.getRawSize()));
+        }
+
+        // The total disk size for data directories is the space that is actually used by those directories plus the
+        // free space on disk that might be used for storing those directories in the future.
+        BigDecimal total = used.add(usable);
+
+        // That total space can be limited by the config property data_disk_usage_max_disk_size.
+        DataStorageSpec diskUsageMaxSize = guardrailsConfigSupplier.get().getDataDiskUsageMaxDiskSize();
```
ekaterinadimitrova2, line 143: This feels to me as one of the advanced config parameters that we would hide from the yaml maybe, not sure. So with the new version of the calculations warnings can be actually faster reached. And things require more advanced knowledge and recommendations and testing in order to decide what to set. Btw we need to add also NEWS.txt

test/unit/org/apache/cassandra/db/guardrails/GuardrailDiskUsageTest.java
```diff
@@ -0,0 +1,615 @@
...
+        guardrails().setDataDiskUsageMaxDiskSize("1GiB");
+        assertThat(monitor.getDiskUsage()).isEqualTo(5.0);
+    }
+
+    @Test
+    public void testDiskUsageCalculationWithMaxDiskSizeAndSmallUnits() throws IOException
+    {
+        FileStore store = mock(FileStore.class);
+        long freeDiskSizeInBytes = DataStorageSpec.inGibibytes(100).toBytes() - DataStorageSpec.inMebibytes(5).toBytes();
+        when(store.getUsableSpace()).thenReturn(DataStorageSpec.inBytes(freeDiskSizeInBytes).toBytes()); // 100GiB disk
```
ekaterinadimitrova2, line 344: Maybe good to add to the comment 5GiB used out of 100GiB disk

test/unit/org/apache/cassandra/db/guardrails/GuardrailDiskUsageTest.java
```diff
@@ -0,0 +1,615 @@
...
+        guardrails().setDataDiskUsageMaxDiskSize("5GiB");
+        assertThat(monitor.getDiskUsage()).isEqualTo(1.0);
+
+        // max disk size < space used
+        guardrails().setDataDiskUsageMaxDiskSize("1GiB");
+        assertThat(monitor.getDiskUsage()).isEqualTo(5.0);
+    }
+
+    @Test
+    public void testDiskUsageCalculationWithMaxDiskSizeAndSmallUnits() throws IOException
```
ekaterinadimitrova2, line 338: I like this test :-)

test/unit/org/apache/cassandra/db/guardrails/GuardrailDiskUsageTest.java
```diff
@@ -0,0 +1,615 @@
...
+        assertThat(monitor.getDiskUsage()).isEqualTo(1.0);
+
+        // max disk size < space used
+        guardrails().setDataDiskUsageMaxDiskSize("1MiB");
+        assertThat(monitor.getDiskUsage()).isEqualTo(5.0);
+    }
+
+    @Test
+    public void testDiskUsageCalculationWithMaxDiskSizeAndMultipleVolumes() throws IOException
+    {
```
ekaterinadimitrova2, line 374: Nice one!

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
ekaterinadimitrova2: One more thing. Maybe a good thing will be to have a getter for the value in the default internal unit; returns the parameter value in the unit we should use internally which is the one used in Config class. In this case we use Bytes. Then everywhere internally we can use that getter ensuring we don't get tempted to use wrong conversion, going up for example and loosing precision. The reason why we have the smallest unit introduced for users to ensure they don't run into that problem with the old parameters I transferred without changing what the parameters use internally. Every parameter is different of course so we always need to think what is the best for the particular one. Nothing has changed from before in regards to thinking what should be the boundaries for a parameter that we want to accept and what units we want to encourage or not, except time/storage and rate are not negative things.. At the end of the day the new types are only wrappers that we need for a  parser but upper boundary is something we still need to consider on a per case basis. 
I am actually thinking maybe to move to abstract Boundary classes and split into Int and long ones to make people think more in particular about that... That is a topic for another ticket of course, just sharing thoughts.

ekaterinadimitrova2 (commented):
Left minor comments and questions. It is solid patch. I am still not sure why before it wasn't done considering only the data occupation + free space, but everything that occupies + free space. Probably worth to check again with Zhao?  I see that now we will be hitting the warnings faster.
One thing I wanted to mention - we try to pick for a parameter in Config the smallest allowed unit as small as it makes sense, if we can set in Bytes that's great, but sometimes it might not make sense. Then internally we use everywhere the parameter converted to that unit of choice and nothing else. So in this case I made the calculation - picking up bytes, we give the chance to people to use at most 8388608 TiB

src/java/org/apache/cassandra/config/GuardrailsOptions.java
adelapena: Done

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
adelapena: Changed to JavaDoc. I think that the other line comments around are being used to comment groups of related properties, instead of individual properties.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
adelapena: I wasn't aware of that code style rule. My intention here was to leave generic utility methods on top (sizes, numbers, etc.), and place the methods validating specific guardrails at the bottom (table properties, disk usage, etc). That way new validation rules should tend to accumulate at the end of the file, leaving the top part undisturbed.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
adelapena: Nope, I guess I changed it while testing something and forgot to leave it as it was. Reverted.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
adelapena: Here we are translating from config's `DataStorageSpec` to the internal long used by `Threshold`. The `Threshold` class is intended to be agnostic to specific data types on config, and it always uses a long even if the property is an int or a data size, and the disable value is always -1, so we can keep it simple.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
adelapena: Added

src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
adelapena: Right, it's [1, 100] and it has been copy-pasted everywhere. Fixed.

src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
adelapena: Fixed

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
adelapena: Good idea, mentioned here and in `GuardrailsConfig`.

src/java/org/apache/cassandra/db/guardrails/PercentageThreshold.java
adelapena: Could be, but guardrail classes seem prone to be extended, as we have just extended `Threshold`. I can imagine that making these final could make things harder for 3rd parties trying to extend guardrails, but I'll be happy changing it to `final`.

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
adelapena: I've tried to add some explanations here.

src/java/org/apache/cassandra/service/disk/usage/DiskUsageState.java
adelapena: Added some JavaDoc.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailDiskUsageTest.java
adelapena: Fixed

conf/cassandra.yaml
adelapena: Same as (1, 100], fixed.

test/unit/org/apache/cassandra/db/guardrails/GuardrailDiskUsageTest.java
adelapena: Just added that comment.

test/unit/org/apache/cassandra/db/guardrails/GuardrailDiskUsageTest.java
adelapena: I'm glad you like it :)

src/java/org/apache/cassandra/config/GuardrailsOptions.java
ekaterinadimitrova2: Actually I think it's me and not official rule. I don't see it in our code style guide, I think I mixed it with the Clear Code/Clear Architecture books plus that I see it all over the place in our codebase so I guess I took it as rule. I am ok the way it is

src/java/org/apache/cassandra/db/guardrails/PercentageThreshold.java
ekaterinadimitrova2: Ok, let's leave it this way

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
adelapena: In this particular case the conversion, which is a series of relatively simple arithmetic operations, happens once when the periodic monitoring task is run. The default rate is 30 seconds, so having a `toBytes` every 30 seconds doesn't seem very problematic compared to checks on the hot path. Probably the `toBytes` call done in the guardrail for collection size could be more problematic, since it's done for every collection cell during writes.

I guess that memoizing the number of bytes into `DataStorageSpec` (or a subclass of it) could improve possible performance problems. If I'm right `DataStorageSpec` is only used in config so we are not going to have many instances on memory, so memoizing is not going to create memory issues. Also, we would only need this optimization on size properties on the hot path.

Performance-wise, an alternative approach would be splitting `Config` in two classes, one for parsing and other for holding the status. The parsing class, say `YamlConfig`, would be short-lived and only used to parse the yaml at startup. It would contain the `DataStorageSpec` instances and other parse-only classes. The class holding the status, say `ConfigStatus`, would be a singleton containing the `volatile` properties in the format that they are going to be internally consumed, such as a longs for bytes. JMX methods would directly write into that `ConfigStatus` class. That's just idea, and I'm not sure it would worth the additional complexity.

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
adelapena: The property `data_disk_usage_max_disk_size` is used to allow to set the disk usage limits as an absolute amount of space instead of relative to the free space on disk. Combined with the the percentages it's an alternative to have a pair of `data_disk_usage_size_warn_threshold`/`data_disk_usage_size_fail_threshold` properties, so we don't have two different limits. The use case would be, for example, a box with 60 TB local SSD volume running multiple Cassandra instances, or reserving disk space for other processes. So I'd say that this property is as important as the percentages themselves, since we are providing two alternative ways of defining the disk usage.

adelapena: I have tried to extend the comments about the new properties on yaml. As for `NEWS.txt`, the plan was to add a single entry about guardrails in general once we were closer to the release date. Now it could be that moment, so I have added that entry.

adelapena:
I have rebased the PR, CI is running:
* [j8](https://app.circleci.com/pipelines/github/adelapena/cassandra/1487/workflows/9fbcfe8a-97b3-4af1-b161-c1b65e5d2a16)
* [j11](https://app.circleci.com/pipelines/github/adelapena/cassandra/1487/workflows/bad9f922-c222-4ef2-b0de-d3b7d285999d)

conf/cassandra.yaml
adelapena: I have added some explanations about how to use the guardrail, and fixed the range.

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
adelapena: Yep, I have added a mention about the null value.

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
ekaterinadimitrova2: I had some conversation with @blambov yesterday and he doesn't think that single multiplication (we don't do division to prevent any precision problems, that's why we set minimum unit) especially when we use 1024 would have any impact. 
Regarding `DataStorageSpec` - yes, it is used only by config. @blambov any concerns in regards to the exact guardrail mentioned - collection, since it's done for every collection cell during writes?

NEWS.txt
```diff
@@ -56,6 +56,34 @@ using the provided 'sstableupgrade' tool.
...
+      user queries, and superusers and internal queries are excluded. Reaching the soft limit raises a client warning,
+      whereas reaching the hard limit aborts the query. In both cases a log message and a diagnostic event are emitted.
+      Additionally, some guardrails are not linked to specific user queries due to techincal limitations, such as
+      detecting the size of large collections during compaction or periodically monitoring the disk usage. These
+      guardrails would only emit the proper logs and diagnostic events when triggered, without aborting any processes.
+      Guardrails config is defined through cassandra.yaml properties, and they can be dynamically updated through the
+      JMX MBean `org.apache.cassandra.db:type=Guardrails`. There are guardrails for:
+        - Number of user keyspaces.
+        - Number of user tables.
+        - Number of columns per table.
```
ekaterinadimitrova2, line 70: there is also the one to enable or disable the ability to create uncompressed tables

conf/cassandra.yaml
ekaterinadimitrova2: LGTM thanks

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
ekaterinadimitrova2: I still see it public? Maybe you didn't commit that change

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
ekaterinadimitrova2: LGTM

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
blambov: In this particular piece of code, I would consider the `Config` conversions trivial compared to the cost of operating on `BigDecimal`. Any of the additions in the code above is at least one, possibly two orders of magnitude more complex than that conversion.

I would not worry about the configuration changes. Outside the scope of this ticket, it may be a good idea to consider why a `BigDecimal` is required instead of `BigInteger` or `long`, possibly pre-multiplied by some fixed power of two.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
adelapena: I reverted it to default visibility in the same commit where I addressed most of the review feedback, [this one](https://github.com/apache/cassandra/pull/1546/commits/15099b433a8fb1d543526c5b9f6b0be1cf74e456#diff-ee9b238deb86b4f297d2bd8a400cdfd799e33d08238bde9652b73a2f64a89c08). Maybe you were looking this second time at `Guardrails.CONFIG_PROVIDER`, a couple of lines above, that has also become public because we need it in the disk usage monitor?

NEWS.txt
adelapena: It's a few lines below, at the end of the list, although I wrongly wrote "compressed" instead of "uncompressed". Just fixed it.

src/java/org/apache/cassandra/service/disk/usage/DiskUsageMonitor.java
adelapena: Thanks for the feedback. Yep, the conversion config properties doesn't look like a big problem here, since it happens in the middle of `BigDecimal` operations and the calculation of the directories size. Besides, this is a periodic task which triggers every 30s, different to other config accesses in the hot path.

As for the use of `BigDecimal`, I guess we can use `BigInteger` until the final division. We use those types instead of double/long to avoid problems while managing very large file systems.

ekaterinadimitrova2: Thanks @blambov , actually the question was more about the collection guardrail that was mentioned. l guess we can discuss this separately from this ticket but as it was slightly touched here in this patch (prohibit 0) and there is no ticket (as there is uncertainty whether there is a real issue to handle or not) I was just interested to get an opinion on that one.

src/java/org/apache/cassandra/service/disk/usage/DiskUsageState.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.cassandra.db.guardrails.GuardrailsConfig;
+
+public enum DiskUsageState
+{
+    /** Either disk usage guardrail is not enabled or gossip state is not ready. */
+    NOT_AVAILABLE("Not Available"),
+
+    /**
+     * Disk usage is below both {@link GuardrailsConfig#getCollectionSizeWarnThreshold()} and
+     * {@link GuardrailsConfig#getCollectionSizeFailThreshold()}.
```
ekaterinadimitrova2: I started working on the 0 patch and noticed this java doc, GuardrailsConfig#getCollectionSizeWarnThreshold(), I think this was copy-paste mistake, also the other JavaDocs in this class. Apologize for not catching it earlier :(

adelapena: Good catch, I should have been thinking on those 0 values for collection sizes when using autocompletion :). Just fixed it.

PR #1548, opened 2022-04-05
Cassandra-17365 remove deprecated version specific TLS

bschoening:
Replaces version specific TLS with auto-negotiation of the highest protocol version that both the client and server support.


pylib/cqlshlib/sslhandling.py
```diff
@@ -48,32 +48,11 @@ def get_option(section, option):
...
     if ssl_validate is None:
         ssl_validate = get_option('ssl', 'validate')
     ssl_validate = ssl_validate is None or ssl_validate.lower() != 'false'
 
-    ssl_version_str = env.get('SSL_VERSION')
-    if ssl_version_str is None:
-        ssl_version_str = get_option('ssl', 'version')
-
-    ssl_version = get_best_tls_protocol(ssl_version_str)
-    if ssl_version is None:
```
smiklosovic, line 73: @bschoening remove this if as get_best_tls_protocol will always return ssl.PROTOCOL_TLS

pylib/cqlshlib/sslhandling.py
```diff
@@ -96,5 +75,5 @@ def get_best_tls_protocol(ssl_ver_str):
 
     return dict(ca_certs=ssl_certfile,
                 cert_reqs=ssl.CERT_REQUIRED if ssl_validate else ssl.CERT_NONE,
-                ssl_version=ssl_version,
+                ssl_version=ssl.PROTOCOL_TLS,
```
smiklosovic, line 78: leave `ssl_version` as it will be always resolved as `ssl.PROTOCOL_TLS`

pylib/cqlshlib/sslhandling.py
```diff
@@ -48,32 +48,11 @@ def get_option(section, option):
         except configparser.Error:
             return None
 
-    def get_best_tls_protocol(ssl_ver_str):
-        # newer python versions suggest to use PROTOCOL_TLS to negotiate the highest TLS version.
-        # older protocol versions have been deprecated:
-        # https://docs.python.org/2/library/ssl.html#ssl.PROTOCOL_TLS
-        # https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLS
-        if ssl_ver_str:
-            return getattr(ssl, "PROTOCOL_%s" % ssl_ver_str, None)
```
smiklosovic, line 57: add here warning instead of this return, remove the next for loop completely and just return `ssl.PROTOCOL_TLS`.

PR #1556, opened 2022-04-07
C175330 cleanup migration coord trunk

jonmeredith:
See [CASSANDRA-17533](https://issues.apache.org/jira/browse/CASSANDRA-17533)

### DIFF
--- a/src/java/org/apache/cassandra/schema/MigrationCoordinator.java
+++ b/src/java/org/apache/cassandra/schema/MigrationCoordinator.java
@@ -343,7 +343,7 @@ public class MigrationCoordinator
     private boolean shouldPullImmediately(InetAddressAndPort endpoint, UUID version)
     {
         UUID localSchemaVersion = schemaVersion.get();
-        if (SchemaConstants.emptyVersion.equals(localSchemaVersion) || ManagementFactory.getRuntimeMXBean().getUptime() < MIGRATION_DELAY_IN_MS)
+        if (SchemaConstants.emptyVersion.equals(localSchemaVersion) || getUptimeFn.getAsLong() < MIGRATION_DELAY_IN_MS)
         {
             // If we think we may be bootstrapping or have recently started, submit MigrationTask immediately
             logger.debug("Immediately submitting migration task for {}, " +
--- a/src/java/org/apache/cassandra/schema/MigrationCoordinator.java
+++ b/src/java/org/apache/cassandra/schema/MigrationCoordinator.java
@@ -38,6 +38,7 @@ import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
+import java.util.function.LongSupplier;
 import java.util.function.Supplier;
 
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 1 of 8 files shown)


src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -90,15 +93,20 @@
     private static final Logger logger = LoggerFactory.getLogger(MigrationCoordinator.class);
     private static final Future<Void> FINISHED_FUTURE = ImmediateFuture.success(null);
 
+    private static LongSupplier getUptimeFn = () -> ManagementFactory.getRuntimeMXBean().getUptime();
```
dcapwell, line 97: this is dead code in your patch

test/distributed/org/apache/cassandra/distributed/action/GossipHelper.java
```diff
@@ -228,7 +229,7 @@ public void accept(IInvokableInstance pullTo)
                 InetAddressAndPort endpoint = toCassandraInetAddressAndPort(pullFrom);
                 EndpointState state = Gossiper.instance.getEndpointStateForEndpoint(endpoint);
                 Gossiper.instance.doOnChangeNotifications(endpoint, ApplicationState.SCHEMA, state.getApplicationState(ApplicationState.SCHEMA));
-                Schema.instance.waitUntilReady(Duration.ofSeconds(10));
+                assertTrue("schema is ready", Schema.instance.waitUntilReady(Duration.ofSeconds(10)));
```
dcapwell, line 232: you are a good person here, but not in 4.0?  why the hate of 4.0!?!?!!

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
```diff
@@ -631,6 +633,7 @@ public void startup(ICluster cluster)
                 StorageService.instance.registerDaemon(CassandraDaemon.getInstanceForTesting());
                 if (config.has(GOSSIP))
                 {
+                    MigrationCoordinator.setUptimeFn(() -> TimeUnit.NANOSECONDS.toMillis(nanoTime() - startedAt));
```
dcapwell: spoke in slack, can you also fix the startedAt bug?  This is object init, which != uptime.

When we run for 1m and restart, the restart should have uptime=0 but will be uptime=1m due to the fact we don't cleanup the object.

dcapwell (changes requested):

test/distributed/org/apache/cassandra/distributed/action/GossipHelper.java
jonmeredith: Puzzler. It was there, I'll put it back.

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -467,16 +476,29 @@ void announce(UUID schemaVersion)
...
+            {
+                skipped = true;
+                return ImmediateFuture.success(null);
+            }
+            return executor.submit(task);
+        }
+        catch (RejectedExecutionException ex)
+        {
+            skipped = true;
             return ImmediateFuture.success(null);
```
dcapwell, line 494: shouldn't this return an failed future rather than success future?

dcapwell: 4.0 returned `null`.  The issue with this patch is that `executor.submit(task)` returns `Void`, which can only ever be `null`; so the caller has 0 way to know that the future was rejected; in 4.0 we could null check the future reference

dcapwell: removing my comment due to the fact this looks to be the desired behavior of this function, so you are not changing anything, you are just being consistent with existing logic...

dcapwell (approved):

jonmeredith:
Reviewed, starting merge. Thanks for the review.

PR #1557, opened 2022-04-08
CASSANDRA-17379: don't start if old and new config params have different values

krummas:


src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -227,6 +227,9 @@
 
     PAXOS_REPAIR_RETRY_TIMEOUT_IN_MS("cassandra.paxos_repair_retry_timeout_millis", "60000"),
 
+    ALLOW_DUPLICATE_CONFIG_KEYS("cassandra.allow_duplicate_config_keys", "true"),
+    ALLOW_CONFLICTING_CONFIG_VALUES("cassandra.allow_conflicting_config_values", "false"),
```
ekaterinadimitrova2: Maybe we can add some JavaDoc?

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1090,7 +1089,7 @@ public static void setOverrideLoadConfig(Supplier<Config> loadConfig)
         exception
     }
 
-    private static final List<String> SENSITIVE_KEYS = new ArrayList<String>() {{
+    public static final Set<String> SENSITIVE_KEYS = new HashSet<String>() {{
```
ekaterinadimitrova2: I am not sure we want those to be public In Config, they might appear at other places? @dcapwell maybe has some opinion as I know in his patch he was filtering things a lot?

src/java/org/apache/cassandra/config/YamlConfigurationLoader.java
```diff
@@ -71,7 +73,7 @@ public class YamlConfigurationLoader implements ConfigurationLoader
      * Inspect the classpath to find storage configuration file
      */
     @VisibleForTesting
-    private static URL getStorageConfigURL() throws ConfigurationException
+    public static URL getStorageConfigURL() throws ConfigurationException
```
ekaterinadimitrova2: This can be protected

test/unit/org/apache/cassandra/config/FailStartupDuplicateParamsTest.java
```diff
@@ -0,0 +1,171 @@
...
+        ALLOW_CONFLICTING_CONFIG_VALUES.setBoolean(false);
+    }
+
+    @Test
+    public void testDuplicateParamThrows() throws IOException
+    {
+        ALLOW_DUPLICATE_CONFIG_KEYS.setBoolean(false);
+        testYaml("found duplicate key endpoint_snitch", true,
+                 "endpoint_snitch: org.apache.cassandra.locator.RackInferringSnitch");
+    }
```
ekaterinadimitrova2, line 75: I guess we can add also 
        ```ALLOW_DUPLICATE_CONFIG_KEYS.setBoolean(false);

             testYaml("nothing", false,
                 "user_defined_functions_enabled: true",
                 "user_defined_functions_enabled: true");```

test/unit/org/apache/cassandra/config/FailStartupDuplicateParamsTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+    @Test
+    public void testReplacementDupesMultiReplace() throws IOException
+    {
+        /*
+    @Replaces(oldName = "internode_socket_send_buffer_size_in_bytes", converter = Converters.BYTES_DATASTORAGE, deprecated = true)
+    @Replaces(oldName = "internode_send_buff_size_in_bytes", converter = Converters.BYTES_DATASTORAGE, deprecated = true)
+    public DataStorageSpec internode_socket_send_buffer_size = new DataStorageSpec("0B");
+         */
```
ekaterinadimitrova2: Indentation is off a bit

ekaterinadimitrova2 (commented):
Overall LGTM, left small comments/questions; missing config  doc update, NEWS.txt, CHANGES.txt
I added something small to the doc in this commit https://github.com/ekaterinadimitrova2/cassandra/commit/9f82520039166610b0f6565defc2a45e96dc1822

src/java/org/apache/cassandra/config/YamlConfigurationLoader.java
```diff
@@ -148,6 +151,38 @@ public Config loadConfig(URL url) throws ConfigurationException
...
+                {
+                    String msg = String.format("old: [%s], new: [%s]", r.oldName, r.newName);
+                    duplicates.add(msg);
+                    logger.warn("Config contains both new and old parameters: {}, please migrate config to use the new parameters", msg);
+                }
+            }
+        }
+
+        if (!duplicates.isEmpty() && !ALLOW_NEW_OLD_CONFIG_KEYS.getBoolean())
+            throw new ConfigurationException("Config contains duplicate keys for old and new configuration parameters:\n" + String.join("\n", duplicates));
```
ekaterinadimitrova2: Probably we can say duplicate or both old and new key of the same parameter. And we can point to the flag even maybe? I noticed you keep the warning but when I was personally testing, I noticed the exception and ignored the warning and I think other people can do the same. WDYT?

ekaterinadimitrova2 (commented):
Overall LGTM, I left just one suggestion in regards to an exception message. Otherwise, I believe we need only NEWS.txt, update of the doc similar to what I suggested in the previous review and full CI.

.build/build-rat.xml
```diff
@@ -58,6 +58,7 @@
                  <exclude NAME="**/doc/antora.yml"/>
                  <exclude name="**/test/conf/cassandra.yaml"/>
                  <exclude name="**/test/conf/cassandra-old.yaml"/>
+                 <exclude name="**/test/conf/cassandra-dupes-template.yaml"/>
```
dcapwell: I don't see this file in this patch

src/java/org/apache/cassandra/config/YamlConfigurationLoader.java
```diff
@@ -199,6 +239,7 @@ public static <T> T fromMap(Map<String,Object> map, boolean shouldCheck, Class<T
     {
         Constructor constructor = new YamlConfigurationLoader.CustomConstructor(klass, klass.getClassLoader());
         Map<Class<?>, Map<String, Replacement>> replacements = getNameReplacements(Config.class);
+        verifyReplacements(replacements, map);
```
dcapwell, line 264: when you rebase you will find a new method

```
public static <T> T updateFromMap(Map<String, ? extends Object> map, boolean shouldCheck, T obj)
```

Can you add this same logic there as well?  replacement is there + same style map, so just need this one line; cool doing this on merge

src/java/org/apache/cassandra/config/YamlConfigurationLoader.java
```diff
@@ -519,6 +559,11 @@ private static void addReplacement(Class<?> klass,
             // by default deprecated is false
             this.deprecated = deprecated;
         }
+
+        public boolean isValueFormatReplacement()
```
dcapwell: this got moved to its own top-level class, so will be a merge conflict on rebase

src/java/org/apache/cassandra/config/YamlConfigurationLoader.java
```diff
@@ -71,7 +73,7 @@ public class YamlConfigurationLoader implements ConfigurationLoader
      * Inspect the classpath to find storage configuration file
      */
     @VisibleForTesting
-    private static URL getStorageConfigURL() throws ConfigurationException
+    protected static URL getStorageConfigURL() throws ConfigurationException
```
dcapwell: looks like the tests do not use, can we go back to private?

dcapwell (approved):

PR #1559, opened 2022-04-08
CASSANDRA-17524 shutdown gossip on drain trunk

jonmeredith:
See [CASSANDRA-17524](https://issues.apache.org/jira/browse/CASSANDRA-17524)

### DIFF
--- a/src/java/org/apache/cassandra/concurrent/Stage.java
+++ b/src/java/org/apache/cassandra/concurrent/Stage.java
@@ -42,27 +42,34 @@ import static org.apache.cassandra.concurrent.ExecutorFactory.Global.executorFac
 
 public enum Stage
 {
-    READ              ("ReadStage",             "request",  DatabaseDescriptor::getConcurrentReaders,        DatabaseDescriptor::setConcurrentReaders,        Stage::multiThreadedLowSignalStage),
-    MUTATION          ("MutationStage",         "request",  DatabaseDescriptor::getConcurrentWriters,        DatabaseDescriptor::setConcurrentWriters,        Stage::multiThreadedLowSignalStage),
-    COUNTER_MUTATION  ("CounterMutationStage",  "request",  DatabaseDescriptor::getConcurrentCounterWriters, DatabaseDescriptor::setConcurrentCounterWriters, Stage::multiThreadedLowSignalStage),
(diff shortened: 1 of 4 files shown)


src/java/org/apache/cassandra/concurrent/Stage.java
```diff
@@ -42,27 +42,34 @@
...
 
     public final String jmxName;
     private final Supplier<ExecutorPlus> executorSupplier;
     private volatile ExecutorPlus executor;
+    /* Set true if this executor should be gracefully shutdown before stopping
+     * the commitlog allocator. Tasks on executors that issue mutations may
+     * block indefinitely waiting for a new commitlog segment, preventing a
+     * clean drain/shutdown.
+     */
+    public final boolean shutdownBeforeCommitlog;
```
maedhroz, line 68: nit: Might as well make it a JavaDoc comment?

src/java/org/apache/cassandra/concurrent/Stage.java
```diff
@@ -152,6 +167,18 @@ public static void shutdownNow()
         ExecutorUtils.shutdownNow(executors());
     }
 
+    public static void shutdownAndAwaitMutatingExecutors(boolean interrupt, long timeout, TimeUnit units) throws InterruptedException, TimeoutException
+    {
+        List<ExecutorPlus> executors = mutatingExecutors();
+        ExecutorUtils.shutdown(interrupt, executors);
+        ExecutorUtils.awaitTermination(timeout, units, executors);
```
maedhroz, line 174: Does `ExecutorUtils.awaitTermination()` fail all the termination awaits if the very first one times out? I guess that might not be a terrible thing, and we'll at lest have something visibly exploding, but wanted to make sure that was cool.

maedhroz (approved):
Looks good to me, with a couple minor questions.

src/java/org/apache/cassandra/concurrent/Stage.java
jonmeredith: Yes, `ExecutorUtils.awaitTermination()` will stop waiting for the other executors as soon as it hits a timeout awaiting. Not an issue in this case, but it is interesting that if there was an exception before waiting for a {{Thread}}, it would never be joined.

jonmeredith:
Committed.

PR #1563, opened 2022-04-11
CASSANDRA-17493 the second take

smiklosovic:


test/unit/org/apache/cassandra/service/StorageServiceTest.java
```diff
@@ -156,4 +169,30 @@ public void testGetChangedReplicasForLeaving() throws Exception
...
+        // we would see 4th counter run which would fail the test
+        Thread.sleep(20_000);
+
+        assertEquals(3, numberOfRuns.get().intValue());
+
+        assertTrue(ScheduledExecutors.scheduledTasks.isTerminated());
+        assertTrue(ScheduledExecutors.nonPeriodicTasks.isTerminated());
+        assertTrue(ScheduledExecutors.scheduledFastTasks.isTerminated());
+        assertTrue(ScheduledExecutors.optionalTasks.isTerminated());
+    }
```
jonmeredith:
```suggestion
    @Test
    public void testScheduledExecutorsShutdownOnDrain() throws Throwable
    {
        final AtomicInteger numberOfRuns = new AtomicInteger(0);

        ScheduledFuture<?> f = ScheduledExecutors.scheduledTasks.scheduleAtFixedRate(() -> numberOfRuns.incrementAndGet(),
                                                                                 0, 1, MILLISECONDS);

        // Prove the task was scheduled more than once before checking cancelled.
        Awaitility.await("first run").atMost(1, TimeUnit.MINUTES).until(() -> numberOfRuns.get() > 1);

        assertFalse(f.isCancelled());
        StorageService.instance.drain();
        assertTrue(f.isCancelled());

        assertTrue(ScheduledExecutors.scheduledTasks.isTerminated());
        assertTrue(ScheduledExecutors.nonPeriodicTasks.isTerminated());
        assertTrue(ScheduledExecutors.scheduledFastTasks.isTerminated());
        assertTrue(ScheduledExecutors.optionalTasks.isTerminated());
    }
```

jonmeredith: I think this would be a more sympathetic test for CI - rather than taking 65second.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5287,12 +5287,16 @@ protected synchronized void drain(boolean isFinalShutdown) throws IOException, I
             CommitLog.instance.shutdownBlocking();
 
             // wait for miscellaneous tasks like sstable and commitlog segment deletion
-            ScheduledExecutors.nonPeriodicTasks.shutdown();
-            if (!ScheduledExecutors.nonPeriodicTasks.awaitTermination(1, MINUTES))
-                logger.warn("Unable to terminate non-periodic tasks within 1 minute.");
```
jonmeredith, line 5292: So this log message would now appear as a TimeoutException in the main catch block on L5301?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5287,12 +5287,16 @@ protected synchronized void drain(boolean isFinalShutdown) throws IOException, I
...
-
             ColumnFamilyStore.shutdownPostFlushExecutor();
-            setMode(Mode.DRAINED, !isFinalShutdown);
+
+            try
+            {
+                ScheduledExecutors.shutdownNowAndWait(1, MINUTES);
```
jonmeredith: Did you think through the consequences of also shutting down `scheduledFastTasks` here too? `approxTime()` will stop updating and tests like `isAfter()` will get stuck returning `false` even as real time passes.

As `drain()` can be called by `nodetool` independently of the shutdown hook, is there any functionality that will be impacted by `approxTime` sticking until the node is finally shutdown?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5287,12 +5287,16 @@ protected synchronized void drain(boolean isFinalShutdown) throws IOException, I
             CommitLog.instance.shutdownBlocking();
 
             // wait for miscellaneous tasks like sstable and commitlog segment deletion
-            ScheduledExecutors.nonPeriodicTasks.shutdown();
-            if (!ScheduledExecutors.nonPeriodicTasks.awaitTermination(1, MINUTES))
-                logger.warn("Unable to terminate non-periodic tasks within 1 minute.");
-
             ColumnFamilyStore.shutdownPostFlushExecutor();
```
jonmeredith, line 5301: Why did you move the shutdownPostFlushExecutor ahead of the refactored non-periodic executor termination?

I couldn't find any tasks that obviously scheduled on the post flush executor that should still be running - it looks mostly related to the commit log and column family truncation which should not be possible on the node at this point in the drain.

So I think this is safe, but was wondering if there was a reason for it or it was an incidental change.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5287,12 +5287,16 @@ protected synchronized void drain(boolean isFinalShutdown) throws IOException, I
...
-            setMode(Mode.DRAINED, !isFinalShutdown);
+
+            try
+            {
+                ScheduledExecutors.shutdownNowAndWait(1, MINUTES);
+            }
+            finally
+            {
+                setMode(Mode.DRAINED, !isFinalShutdown);
```
jonmeredith, line 5314: Interesting when a node should decide it is `DRAINED` if there's an exception. You can't attempt draining again if an exception is thrown after the collection of mutation stages. I'm guessing you structured it like this so you would retain behavior on timeout from the scheduled executor shutdown as before.

jonmeredith (commented):
I've left some comments on the PR.  To summarize my comments/concerns
1) The unit test takes a long time in current form - I've proposed a change below
2) approxTime() will stop updating once the scheduled executors shutdown on a `nodetool drain`, is that safe?
3) Is it safe to reorder the post flush executor shutdown before the scheduled executors.

test/unit/org/apache/cassandra/service/StorageServiceTest.java
smiklosovic: ah yes this is better way how to look at it, I couldnt remember that AtomicInteger as I was in the rush :D No problem with Awaitility either.

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: Yes. With one caveat and that is that instead of waiting 1 minute _for each executor_ (here we had just one), we are waiting 1 minute _for all of them_ to be shutdown / terminated. I consider this to be the flaw of the previous solution. They were not covering all other executors and I think that if whoever did that was taking other executors into account, he would end up with same solution.

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: I am sorry I am not getting your first paragraph at all. What approxTime? The second paragraph builds on top of me not understanding the first.

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: Well, the movement was done purely by looking at it - I similarly think nothing is scheduling on that executor. There is stuff related to flushing commit logs etc. There seems to be nothing which would schedule stuff for ColumnFamilyStore to these executors.

This is again something I consider to be the flaw of the previous solution. I think that all logic should be "stopped" first before shutting down all executors. I would say that the subsystem should clean up after itself first and then we are safer to shut down everything else (even a task that subsystem forgot to shut down itself) rather than thinking that we submitted a task to these executors on which ColumnFamilyStore depends. Also, if we shut down executors first and then we shut down a subsystem, is not it the case that we might shut down something prematurely?

Any subsystem should keed track of own tasks it submitted into these executors and they should be stopped internally first. By this solution, we are making it better to shut it all down. It may happen that if the subsystem is not shutting its own tasks down on drain that subsystem would "leak" unstopped tasks.

The tests / pipeline proves me right that, I do not see any regression.

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: Exactly my reasoning behind that. We do not have idempotent drain right now. This is quite sad to be honest. We can not "re-drain". So this is our only chance to be drained even it throws.

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: @jonmeredith  I think I got it, I rewrote it a little bit. You will see it in the next commit. I think, to be super safe, it is better if we move the shutdown of that very specific fast tasks executor to Runtime's shutdown hooks so we shut down as the very last thing so we still progress time at least while c* is between drain and the actual shut down area.

jonmeredith (approved):

PR #1567, opened 2022-04-12
CASSANDRA-17537: nodetool compact should support using a key string to find the range to avoid operators having to manually do this

dcapwell:


src/java/org/apache/cassandra/db/compaction/CompactionController.java
```diff
@@ -255,10 +255,7 @@ public LongPredicate getPurgeEvaluator(DecoratedKey key)
 
         for (SSTableReader sstable: filteredSSTables)
         {
-            // if we don't have bloom filter(bf_fp_chance=1.0 or filter file is missing),
-            // we check index file instead.
-            if (sstable.getBloomFilter() instanceof AlwaysPresentFilter && sstable.getPosition(key, SSTableReader.Operator.EQ, false) != null
-                || sstable.getBloomFilter().isPresent(key))
+            if (sstable.isPresent(key))
```
dcapwell: moved this logic into the sstable so it can be reused

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -928,10 +929,10 @@ protected void runMayThrow()
         return futures;
     }
 
-    public void forceCompactionForTokenRange(ColumnFamilyStore cfStore, Collection<Range<Token>> ranges)
+    public void forceCompaction(ColumnFamilyStore cfStore, Supplier<Collection<SSTableReader>> sstablesFn, com.google.common.base.Predicate<SSTableReader> sstablesPredicate)
```
krummas, line 932: It is slightly weird to have both the `Supplier` and the `Predicate` here - but i guess its needed due to that `sstablesInBounds` method. Maybe we should add an assertion that `sstablesPredicate.test(..)` returns true for sstables from `sstablesFn`

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -990,6 +996,23 @@ private static Collection<SSTableReader> sstablesInBounds(ColumnFamilyStore cfs,
...
 
+    public void forceCompactionForKey(ColumnFamilyStore cfStore, DecoratedKey key)
+    {
+        forceCompaction(cfStore, () -> sstablesWithKey(cfStore, key), sstable -> sstable.isPresent(key));
+    }
+
+    private static Collection<SSTableReader> sstablesWithKey(ColumnFamilyStore cfs, DecoratedKey key)
+    {
+        final Set<SSTableReader> sstables = new HashSet<>();
+        Iterable<SSTableReader> liveTables = cfs.getTracker().getView().select(SSTableSet.LIVE);
```
krummas: this could be
```
Iterable<SSTableReader> liveTables = cfs.getTracker().getView().liveSSTablesInBounds(key.getToken().minKeyBound(), 
                                                                                     key.getToken().maxKeyBound());
```
to avoid false positives and to reduce the number of `.isPresent` checks below

src/java/org/apache/cassandra/dht/Splitter.java
```diff
@@ -92,6 +92,21 @@ else if (token.compareTo(unwrapped.left) < 0)
         return elapsedTokens;
     }
 
+    public Token next(Token token)
```
krummas: these are unused

src/java/org/apache/cassandra/io/sstable/format/SSTableReader.java
```diff
@@ -2003,6 +2003,14 @@ public void addTo(Ref.IdentityCollection identities)
 
     }
 
+    public boolean isPresent(DecoratedKey key)
```
krummas: maybe rename this to indicate that there are no guarantees here - or invert it to say `isNotPresent` ?

src/java/org/apache/cassandra/io/sstable/format/SSTableReader.java
```diff
@@ -2003,6 +2003,14 @@ public void addTo(Ref.IdentityCollection identities)
...
+    {
+        // if we don't have bloom filter(bf_fp_chance=1.0 or filter file is missing),
+        // we check index file instead.
+        IFilter bf = getBloomFilter();
```
krummas:
```
     [echo] Running Eclipse Code Analysis.  Output logged to /.../build/ecj/eclipse_compiler_checks.txt
     [java] ----------
     [java] 1. ERROR in /.../src/java/org/apache/cassandra/io/sstable/format/SSTableReader.java (at line 2011)
     [java]     return bf instanceof AlwaysPresentFilter && getPosition(key, Operator.EQ, false) != null || bf.isPresent(key);
     [java]     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     [java] Potential resource leak: 'bf' may not be closed at this location
```

can probably just avoid `getBloomFilter()` here

krummas (changes requested):

src/java/org/apache/cassandra/io/sstable/format/SSTableReader.java
dcapwell: adding `eclipse-warnings` to my scripts so I run this more often!

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
dcapwell: not sure about this one, `sstablesFn` is called when compaction is paused, so if I call earlier than that could that not cause problems?  Unless I call early not sure how to get access to the SSTables

krummas: I meant that after calling `sstablesFn.get()` in `taskCreator` we just `assert sstables.stream().allMatch(sstablesPredicate);`

but its fine leaving out as well

krummas (approved):

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
dcapwell: added

```
List<SSTableReader> sstablesNotInPredicate = sstables.stream().filter(a -> ! sstablesPredicate.test(a)).collect(Collectors.toList());
            assert sstablesNotInPredicate.isEmpty() : String.format("Attempted to force compact %s, but predicate does not include", sstablesNotInPredicate);
```

I prefer it show which sstables not included than existence

dcapwell: this check broke some tests due to `(32, 31]` finding files that had bounds `(32, 32]`

PR #1572, opened 2022-04-18
CASSANDRA-17560: Migrate track_warnings to more standard naming conventions and use latest configuration types rather than long

dcapwell:


test/unit/org/apache/cassandra/config/YamlConfigurationLoaderTest.java
```diff
@@ -40,41 +40,43 @@ public void trackWarningsFromConfig()
...
+        "row_index_size_fail_threshold", "1024KiB"
+        );
+
+        Config c = YamlConfigurationLoader.fromMap(map, Config.class);
+        assertThat(c.read_thresholds_enabled).isTrue();
 
-        assertThat(tw.coordinator_read_size.warn_threshold_kb).isEqualTo(1024);
-        assertThat(tw.coordinator_read_size.abort_threshold_kb).isEqualTo(0);
+        assertThat(c.coordinator_read_size_warn_threshold).isEqualTo(DataStorageSpec.inKibibytes(1024));
+        assertThat(c.coordinator_read_size_fail_threshold).isNull();;
```
maedhroz: nit: extra semicolon

src/java/org/apache/cassandra/cql3/selection/ResultSetBuilder.java
```diff
@@ -22,6 +22,8 @@
 import java.util.Arrays;
 import java.util.List;
 
+import org.apache.cassandra.config.Config;
+import org.apache.cassandra.config.DataStorageSpec;
```
maedhroz: nit: unused import

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -913,7 +913,17 @@ static void applyRepairCommandPoolSize(Config config)
     @VisibleForTesting
     static void applyTrackWarningsValidations(Config config)
     {
-        config.track_warnings.validate("track_warnings");
+        validateTrackWarnings("coordinator_read_size", config.coordinator_read_size_warn_threshold, config.coordinator_read_size_fail_threshold);
+        validateTrackWarnings("local_read_size", config.local_read_size_warn_threshold, config.local_read_size_fail_threshold);
+        validateTrackWarnings("row_index_size", config.row_index_size_warn_threshold, config.row_index_size_fail_threshold);
+    }
+
+    private static void validateTrackWarnings(String name, DataStorageSpec warn, DataStorageSpec fail)
```
maedhroz:
```suggestion
    private static void validateReadThresholds(String name, DataStorageSpec warn, DataStorageSpec fail)
```

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -913,7 +913,17 @@ static void applyRepairCommandPoolSize(Config config)
...
+    {
+        if (fail != null && warn != null && fail.toBytes() < warn.toBytes())
+            throw new ConfigurationException(String.format("%s (%s) must be greater than or equal to %s (%s)",
+                                                           name + "_fail_threshold", fail.toString(),
+                                                           name + "_warn_threshold", warn.toString()));
```
maedhroz: nit: The `toString()` isn't strictly necessary.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -3932,74 +3942,97 @@ public static SubnetGroups getInternodeErrorReportingExclusions()
...
+        conf.row_index_size_fail_threshold = value;
     }
 
-    public static void setRowIndexSizeAbortThresholdKiB(int value)
+    private static DataStorageSpec parseDataStorageSpec(String threshold)
     {
-        conf.track_warnings.row_index_size.setAbortThresholdKb(value);
+        return threshold == null
+               ? null
+               : new DataStorageSpec(threshold);
```
maedhroz: What happens if someone passes something like `""` or `"      "`?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
maedhroz: I guess this would also extend to things like `QueryOptions.TrackWarnings`? (i.e. We'd rename to `ReadThresholds` or something similar?)

src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
```diff
@@ -885,9 +885,9 @@ private void maybeWarn(ResultSetBuilder result, QueryOptions options)
         ColumnFamilyStore store = cfs();
         if (store != null)
             store.metric.coordinatorReadSize.update(result.getSize());
-        if (result.shouldWarn(options.getCoordinatorReadSizeWarnThresholdKB()))
+        if (result.shouldWarn(options.getCoordinatorReadSizeWarnThresholdBytes()))
         {
-            String msg = String.format("Read on table %s has exceeded the size warning threshold of %,d kb", table, options.getCoordinatorReadSizeWarnThresholdKB());
+            String msg = String.format("Read on table %s has exceeded the size warning threshold of %,d", table, options.getCoordinatorReadSizeWarnThresholdBytes());
```
maedhroz: Would it make sense to replace "kb" w/ "bytes" instead of nothing?

maedhroz: Another idea...what if we just used the actual `DataStorageSpec` instances as fields in `DefaultTrackWarnings` and nulls in `DisabledTrackWarnings`? Then we could use the `toString()` here.

src/java/org/apache/cassandra/db/ReadCommand.java
```diff
@@ -707,16 +709,16 @@ protected DeletionTime applyToDeletion(DeletionTime deletionTime)
             private void addSize(long size)
             {
                 this.sizeInBytes += size;
-                if (abortThresholdBytes != 0 && this.sizeInBytes >= abortThresholdBytes)
+                if (failBytes != Config.DISABLED_GUARDRAIL && this.sizeInBytes >= failBytes)
                 {
-                    String msg = String.format("Query %s attempted to read %d bytes but max allowed is %d; query aborted  (see track_warnings.local_read_size.abort_threshold_kb)",
-                                               ReadCommand.this.toCQLString(), this.sizeInBytes, abortThresholdBytes);
+                    String msg = String.format("Query %s attempted to read %d bytes but max allowed is %s; query aborted  (see local_read_size_fail_threshold)",
+                                               ReadCommand.this.toCQLString(), this.sizeInBytes, failThreshold.toString());
```
maedhroz: nit: unnecessary `toString()`

src/java/org/apache/cassandra/db/RowIndexEntry.java
```diff
@@ -352,31 +353,31 @@ public RowIndexEntry<IndexInfo> deserialize(DataInputPlus in, long indexFilePosi
...
-            if (abortThreshold != 0 && estimatedMemory > abortThreshold)
+            if (abortThreshold != null && estimatedMemory > abortThreshold.toBytes())
             {
                 String msg = String.format("Query %s attempted to access a large RowIndexEntry estimated to be %d bytes " +
-                                           "in-memory (total entries: %d, total bytes: %d) but the max allowed is %d;" +
-                                           " query aborted  (see row_index_size_abort_threshold_kb)",
-                                           command.toCQLString(), estimatedMemory, entries, bytes, abortThreshold);
+                                           "in-memory (total entries: %d, total bytes: %d) but the max allowed is %s;" +
+                                           " query aborted  (see row_index_size_failed_threshold)",
+                                           command.toCQLString(), estimatedMemory, entries, bytes, abortThreshold.toString());
```
maedhroz: nit: unnecessary `toString()`

src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -920,23 +920,20 @@ public void enableAuditLog(String loggerName, String includedKeyspaces, String e
     public void setCompactionTombstoneWarningThreshold(int count);
     public int getCompactionTombstoneWarningThreshold();
 
-    public boolean getTrackWarningsEnabled();
-    public void setTrackWarningsEnabled(boolean value);
```
maedhroz, line 924: Will these no longer be configurable at runtime?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
dcapwell: both will fail due to `new DataStorageSpec` not being able to parse it; so we throw an exception.

src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
dcapwell: main reason I didn't do that is due to the fact `maybeFail` is called frequently (each row), so didn't like calling `toBytes` constantly... One option is to have the interface hold both, but feels like overkill; `bytes` I think is a good blend as humans can always convert to a higher form if needed, but know explicitly what the limit is

src/java/org/apache/cassandra/service/StorageServiceMBean.java
dcapwell: thanks, looks like I forgot to add back; it was only added to DD

maedhroz (approved):

src/java/org/apache/cassandra/db/RowIndexEntry.java
```diff
@@ -352,31 +353,31 @@ public RowIndexEntry<IndexInfo> deserialize(DataInputPlus in, long indexFilePosi
         private void checkSize(int entries, int bytes)
         {
             ReadCommand command = ReadCommand.getCommand();
-            if (command == null || SchemaConstants.isSystemKeyspace(command.metadata().keyspace) || !DatabaseDescriptor.getTrackWarningsEnabled())
+            if (command == null || SchemaConstants.isSystemKeyspace(command.metadata().keyspace) || !DatabaseDescriptor.getReadThresholdsEnabled())
```
adelapena, line 356: `getReadThresholdsEnabled` should be faster than `isSystemKeyspace`, we could change the order to check `getReadThresholdsEnabled` first.

test/unit/org/apache/cassandra/config/YamlConfigurationLoaderTest.java
```diff
@@ -40,41 +40,43 @@ public void trackWarningsFromConfig()
     {
         // this test just makes sure snakeyaml loads the test config properly and populates the fields (track warnings uses final in some places)
         // if the config is changed, its ok to update this test to reflect that change
-        TrackWarnings tw = load("test/conf/cassandra.yaml").track_warnings;
-        assertThat(tw.enabled).isTrue();
+        Config c = load("test/conf/cassandra.yaml");
```
adelapena, line 41: The methods `trackWarningsFromConfig` and `trackWarningsFromMap` could be named `readThresholdsFromConfig` and `readThresholdsFromMap`.

src/java/org/apache/cassandra/db/RowIndexEntry.java
dcapwell: I rather leave as is for this patch.

> getReadThresholdsEnabled should be faster than isSystemKeyspace

🤷 may or may not, 3 set contains clauses on non-thread safe data structures vs a memory fence... without measuring I could see it go both ways

conf/cassandra.yaml
```diff
@@ -1575,29 +1575,24 @@ drop_compact_storage_enabled: false
...
+# coordinator_read_size_warn_threshold:
+# coordinator_read_size_fail_threshold:
+# When read_thresholds_enabled: true, this tracks the size of the local read (as defined by
+# heap size), and will warn/fail based off these thresholds; undefined disables these checks.
+# local_read_size_warn_threshold:
+# local_read_size_fail_threshold:
+# When read_thresholds_enabled: true, this tracks the expected memory size of the RowIndexEntry
+# and will warn/fail based off these thresholds; undefined disables these checks
+# row_index_size_warn_threshold:
+# row_index_size_fail_threshold:
```
adelapena: Since we are renaming to read thresholds, shouldn't these be named `row_index_read_size_warn_threshold`/`row_index_read_size_fail_threshold`? I think we only check the index size when reading, but there is no limit when writing.

src/java/org/apache/cassandra/service/reads/trackwarnings/WarningContext.java
```diff
@@ -28,9 +28,9 @@
 
 public class WarningContext
 {
-    private static EnumSet<ParamType> SUPPORTED = EnumSet.of(ParamType.TOMBSTONE_WARNING, ParamType.TOMBSTONE_ABORT,
-                                                             ParamType.LOCAL_READ_SIZE_WARN, ParamType.LOCAL_READ_SIZE_ABORT,
-                                                             ParamType.ROW_INDEX_SIZE_WARN, ParamType.ROW_INDEX_SIZE_ABORT);
+    private static EnumSet<ParamType> SUPPORTED = EnumSet.of(ParamType.TOMBSTONE_WARNING, ParamType.TOMBSTONE_FAIL,
```
adelapena, line 31: Should we rename the `org.apache.cassandra.service.reads.trackwarnings` package to `org.apache.cassandra.service.reads.readthresholds`, or perhaps `org.apache.cassandra.service.reads.thresholds`?

dcapwell: how about `package org.apache.cassandra.service.thresholds;`?  CASSANDRA-17258 adds a write threshold (blocked on 2 JIRAs), so rather than a `read.thresholds` and `write.thresholds` we just have `thresholds`?

conf/cassandra.yaml
dcapwell:
```
case ROW_INDEX_SIZE_FAIL:
                    reason = RequestFailureReason.READ_SIZE;
```

our error code says its READ_SIZE, so makes sense, will do

dcapwell: pushed change... there is prob a regression somewhere =D

src/java/org/apache/cassandra/service/reads/trackwarnings/WarningContext.java
adelapena: `org.apache.cassandra.service.thresholds` makes sense to me

dcapwell:
with latest changes track warnings no longer exist

```
$ egrep -r 'track_warning|track-warning|trackwarning' src/ test/
$
```

src/java/org/apache/cassandra/service/reads/thresholds/CoordinatorWarnings.java
```diff
@@ -32,12 +32,11 @@
 import org.apache.cassandra.metrics.TableMetrics;
 import org.apache.cassandra.schema.Schema;
 import org.apache.cassandra.service.ClientWarn;
-import org.apache.cassandra.service.reads.ReadCallback;
 
 public class CoordinatorWarnings
 {
     private static final Logger logger = LoggerFactory.getLogger(CoordinatorWarnings.class);
-    private static final boolean ENABLE_DEFENSIVE_CHECKS = Boolean.getBoolean("cassandra.track_warnings.coordinator.defensive_checks_enabled");
+    private static final boolean ENABLE_DEFENSIVE_CHECKS = Boolean.getBoolean("cassandra.reads.thresholds.coordinator.defensive_checks_enabled");
```
adelapena, line 39: This still appears with its previous name (`cassandra.track_warnings.coordinator.defensive_checks_enabled`) on `build.xml` and `ide/idea/workspace.xml`

conf/cassandra.yaml
```diff
@@ -1575,29 +1575,24 @@ drop_compact_storage_enabled: false
 #    - 127.0.0.1
 #    - 127.0.0.0/31
 
-# Enables tracking warnings/aborts across all replicas for reporting back to client.
+# Enables read thresholds (warn/fail) across all replicas for reporting back to the client.
```
adelapena, line 1578: We should update [the entry in `NEWS.txt`](https://github.com/apache/cassandra/blob/ab192b26a28e84f025d6be475786067f22f97cd5/NEWS.txt#L92-L101), it is still mentioning the previous properties.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -911,9 +911,19 @@ static void applyRepairCommandPoolSize(Config config)
...
-    static void applyTrackWarningsValidations(Config config)
+    static void applyReadThresholdsValidations(Config config)
     {
-        config.track_warnings.validate("track_warnings");
+        validateReadThresholds("coordinator_read_size", config.coordinator_read_size_warn_threshold, config.coordinator_read_size_fail_threshold);
+        validateReadThresholds("local_read_size", config.local_read_size_warn_threshold, config.local_read_size_fail_threshold);
+        validateReadThresholds("row_index_read_size", config.row_index_read_size_warn_threshold, config.row_index_read_size_fail_threshold);
+    }
+
+    private static void validateReadThresholds(String name, DataStorageSpec warn, DataStorageSpec fail)
```
adelapena, line 921: Are fine with accepting 0 as a valid value, even if it makes the db unusable? Probably any other very low value is going to be as bad as 0, so I guess that rejecting it wouldn't make a big difference.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -6371,106 +6371,87 @@ public void addSnapshot(TableSnapshot snapshot) {
...
 
     @Override
-    public void setCoordinatorLargeReadWarnThresholdKB(long threshold)
+    public void setCoordinatorLargeReadWarnThreshold(String threshold)
     {
-        if (threshold < 0)
-            throw new IllegalArgumentException("threshold " + threshold + " is less than 0; must be positive or zero");
-        DatabaseDescriptor.setCoordinatorReadSizeWarnThresholdKB(threshold);
-        logger.info("updated track_warnings.coordinator_large_read.warn_threshold_kb to {}", threshold);
+        DatabaseDescriptor.setCoordinatorReadSizeWarnThreshold(threshold);
```
adelapena: It seems that we are doing the conversion from `String` to `DataStorageSpec` on `DatabaseDescriptor`, and the opposite conversion from `DataStorageSpec` to `String` on `StorageService`. Wouldn't it make more sense to do both conversions on the same class, probably on `StorageService` which is the one dealing with JMX?

src/java/org/apache/cassandra/cql3/QueryOptions.java
```diff
@@ -283,27 +280,27 @@ public boolean isEnabled()
         }
 
         @Override
-        public long getCoordinatorReadSizeWarnThresholdKB()
+        public long getCoordinatorReadSizeWarnThresholdBytes()
         {
-            return 0;
+            return Config.DISABLED_GUARDRAIL;
```
adelapena, line 285: With the new config standardisation `-1` and `null` are going to be the disabled value for all numeric/size properties, so maybe we should rename those constants to something more generic, or just directly use `-1`/`null` as it's done by other non-guardrail properties. Whatever we do, I think there is no need to do it on this ticket.

src/java/org/apache/cassandra/service/reads/thresholds/CoordinatorWarnings.java
dcapwell: before

```
$ grep -r 'cassandra.track_warnings.coordinator.defensive_checks_enabled' *
build.xml:        <jvmarg value="-Dcassandra.track_warnings.coordinator.defensive_checks_enabled=true" /> <!-- enable defensive checks -->
build.xml:        <jvmarg value="-Dcassandra.track_warnings.coordinator.defensive_checks_enabled=true" /> <!-- enable defensive checks -->
ide/idea/workspace.xml:      <option name="VM_PARAMETERS" value="-Dcassandra.config=file://$PROJECT_DIR$/conf/cassandra.yaml -Dcassandra.storagedir=$PROJECT_DIR$/data -Dlogback.configurationFile=file://$PROJECT_DIR$/conf/logback.xml -Dcassandra.logdir=$PROJECT_DIR$/data/logs -Djava.library.path=$PROJECT_DIR$/lib/sigar-bin -DQT_SHRINKS=0 -ea -Dcassandra.track_warnings.coordinator.defensive_checks_enabled=true" />
ide/idea/workspace.xml:      <option name="VM_PARAMETERS" value="-Dcassandra.config=file://$PROJECT_DIR$/test/conf/cassandra.yaml -Dlogback.configurationFile=file://$PROJECT_DIR$/test/conf/logback-test.xml -Dcassandra.logdir=$PROJECT_DIR$/build/test/logs -Djava.library.path=$PROJECT_DIR$/lib/sigar-bin -Dlegacy-sstable-root=$PROJECT_DIR$/test/data/legacy-sstables -Dinvalid-legacy-sstable-root=$PROJECT_DIR$/test/data/invalid-legacy-sstables -Dcassandra.ring_delay_ms=1000 -Dcassandra.skip_sync=true -ea -XX:MaxMetaspaceSize=384M -XX:SoftRefLRUPolicyMSPerMB=0 -Dcassandra.strict.runtime.checks=true -Dlegacy-sstable-root=$PROJECT_DIR$/test/data/legacy-sstables -Dinvalid-legacy-sstable-root=$PROJECT_DIR$/test/data/invalid-legacy-sstables -Dmigration-sstable-root=$PROJECT_DIR$/test/data/migration-sstables -Dcassandra.ring_delay_ms=1000 -Dcassandra.tolerate_sstable_size=true -Dcassandra.skip_sync=true -Dcassandra.track_warnings.coordinator.defensive_checks_enabled=true" />
ide/idea/workspace.xml:      <option name="VM_PARAMETERS" value="-Dcassandra-foreground=yes -Dcassandra.config=file://$PROJECT_DIR$/conf/cassandra.yaml -Dcassandra.storagedir=$PROJECT_DIR$/data -Dlogback.configurationFile=file://$PROJECT_DIR$/conf/logback.xml -Dcassandra.logdir=$PROJECT_DIR$/data/logs -Djava.library.path=$PROJECT_DIR$/lib/sigar-bin -Dcom.sun.management.jmxremote -Dcom.sun.management.jmxremote.port=7199 -Dcom.sun.management.jmxremote.local.only=false -Dcom.sun.management.jmxremote.authenticate=false -Dcom.sun.management.jmxremote.ssl=false -ea -Xmx1G -Dcassandra.track_warnings.coordinator.defensive_checks_enabled=true" />
```

After

```
$ grep -r 'cassandra.track_warnings.coordinator.defensive_checks_enabled' *
$
```

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
dcapwell: yeah, 0b and 1b are basically the same, so don't think blocking 0 matters much.

2 checks are memory based (coordinator and index), so those also depend on the JVM running, so not sure if a lower bound would make any sense

src/java/org/apache/cassandra/service/StorageService.java
dcapwell: yes it would

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -3932,74 +3942,84 @@ public static SubnetGroups getInternodeErrorReportingExclusions()
...
     {
-        return conf.track_warnings.coordinator_read_size.getWarnThresholdKb();
+        return conf.coordinator_read_size_warn_threshold;
     }
 
-    public static void setCoordinatorReadSizeWarnThresholdKB(long threshold)
+    public static void setCoordinatorReadSizeWarnThreshold(DataStorageSpec value)
     {
-        conf.track_warnings.coordinator_read_size.setWarnThresholdKb(threshold);
+        logger.info("updated coordinator_read_size_warn_threshold to {}", value);
```
adelapena, line 3966: Nit: if this is before actually setting the property the message should be `updating coordinator_read_size_warn_threshold to {}`. Or it could be after setting the property if we want to use `updated`. The same applies to the other two properties.

adelapena (approved):
The changes look good to me, +1. I have just added a nit about how we log the updating of properties. That can be addressed on commit.

PR #1579, opened 2022-04-20
Cassandra 17563

dcapwell:


.circleci/generate.sh
```diff
@@ -116,49 +116,53 @@ if $has_env_vars && $check_env_vars; then
...
+  cat $BASEDIR/license.yml $BASEDIR/config.yml.MIDRES.tmp > $BASEDIR/config.yml.MIDRES
+  rm $BASEDIR/config-2_1.yml.MIDRES $BASEDIR/config.yml.MIDRES.tmp
+}
+
+highers() {
+  $BASEDIR/circle-ci-config.py HIGHER --stdout > $BASEDIR/config-2_1.yml.HIGHRES
+  circleci config process $BASEDIR/config-2_1.yml.HIGHRES > $BASEDIR/config.yml.HIGHRES.tmp
+  cat $BASEDIR/license.yml $BASEDIR/config.yml.HIGHRES.tmp > $BASEDIR/config.yml
+  rm $BASEDIR/config-2_1.yml.HIGHRES $BASEDIR/config.yml.HIGHRES.tmp
+}
```
dcapwell: there were slight differences between the single case and the `-a` case, so by leveraging functions we can do this in a consistent way

.circleci/circle-ci-config.py
```diff
@@ -0,0 +1,167 @@
...
+            '_default_job_':                            {'parallelism': 100,    'exec_resource_class': 'xlarge'},
+            'jobs': {
+                'j8_jvm_upgrade_dtests':                {'parallelism': 2},
+                'j11_jvm_dtests':                       {'parallelism': 5},
+                'j11_jvm_dtests_vnode':                 {'parallelism': 5},
+                'j8_jvm_dtests':                        {'parallelism': 5},
+                'j8_jvm_dtests_vnode':                  {'parallelism': 5},
+            },
+        },
+}
```
dcapwell: The core idea is that we maintain the list of jobs which need different resources, rather than patches which assume offers do not change...

This also doesn't rely on custom executors as executors are created per job already, so rather than having many executors, we can really get away with 2-4 executors (j8/j11, java/python) and define resources per job

.circleci/circle-ci-config.py
```diff
@@ -0,0 +1,196 @@
...
+
+try:
+    import yaml
+except ImportError:
+    import subprocess
+    # dependencies not installed; install them outside of the normal system
+
+    subprocess.check_call([sys.executable, '-m', 'pip', 'install', "--target=" + dep_dir, 'pyyaml'])
+
+    import yaml
```
dcapwell, line 39: this is to make sure the script "just works", so if you don't have the package installed the script can move forward by installing local to this directory

.circleci/circle-ci-config.py
```diff
@@ -0,0 +1,196 @@
...
+
+try:
+    from yaml import CLoader as Loader, CDumper as Dumper
+except ImportError:
+    from yaml import Loader, Dumper
```
dcapwell, line 44: this just handles the different implementations; just for portability

.circleci/circle-ci-config.py
```diff
@@ -0,0 +1,196 @@
...
+        },
+        'HIGHER': {
+            # apply defaults to all jobs, so the below just overrides jobs which do not match the default
+            '_default_job_':                            {'parallelism': 100,    'exec_resource_class': 'xlarge'},
+            'jobs': {
+                'j8_jvm_upgrade_dtests':                {'parallelism': 2},
+                'j11_jvm_dtests':                       {'parallelism': 5},
+                'j11_jvm_dtests_vnode':                 {'parallelism': 5},
+                'j8_jvm_dtests':                        {'parallelism': 5},
+                'j8_jvm_dtests_vnode':                  {'parallelism': 5},
```
dcapwell, line 97: don't know why MID/HIGh use different values here... ideally we should make this based off tests...

in my personal scripts I have the following to figure out better parallel

```
ef java_parallelism(src_dir, kind, num_file_in_worker, include = lambda a, b: True):
    d = os.path.join(src_dir, 'test', kind)
    num_files = 0
    for root, dirs, files in os.walk(d):
        for f in files:
            if f.endswith('Test.java') and include(os.path.join(root, f), f):
                num_files += 1
    return math.floor(num_files / num_file_in_worker)
...
    unit_parallelism                = java_parallelism(args.src, 'unit', 20)
    jvm_dtest_parallelism           = java_parallelism(args.src, 'distributed', 4, lambda full, name: 'upgrade' not in full)
    jvm_dtest_upgrade_parallelism   = java_parallelism(args.src, 'distributed', 2, lambda full, name: 'upgrade' in full)
```

if you use that script our jobs run faster and workflows complete faster...

.circleci/circle-ci-config.py
```diff
@@ -0,0 +1,196 @@
...
+
+                'utests_fqltool':                       {'parallelism': 1},
```
dcapwell, line 99: I assume its an oversight that these use xlarge when they complete quickly with medium...

.circleci/circle-ci-job-resources.py
```diff
@@ -0,0 +1,98 @@
...
+        target_jobs = target['jobs']
+        jobs = contents['jobs']
+        default_spec = target.get('_default_job_')
+        if default_spec:
+            for name, job in jobs.items():
+                apply_job_spec(job, default_spec)
+
+        for name, spec in target_jobs.items():
+            apply_job_spec(jobs[name], spec)
+
```
dcapwell: copy/paste code, removed locally but can't commit without stopping CI =/

dcapwell:
Talking to @ekaterinadimitrova2 to lower the risk of breaking `generate.sh` I reverted that logic and used the new scripts to generate the diff

.circleci/circle-ci-config.py
ekaterinadimitrova2: I would say if you have to add a comment on the PR for reviewer, probably a good idea to add a comment in the code for this :0)

.circleci/circle-ci-config.py
ekaterinadimitrova2: Same comment as before ;-)

.circleci/circle-ci-config.py
ekaterinadimitrova2: Highres uses XLarge and Midres uses Large :-) Let's not change that for now and focus on the fix :-)

.circleci/circle-ci-config.py
ekaterinadimitrova2: Well, as far as I can see Highres is just bump to the highest resources, it didn't consider per suite what is enough or not, just bumped on max :-) But I neither use it/nor created it so I would refrain from changing

.circleci/readme.md
```diff
@@ -102,7 +102,7 @@ Read below for details how to generate the files manually without the `generate.
    config-2_1.yml)
    Note that if the patch no longer applies to `config-2_1.yml` a new patch file
    is needed, do this by manually making `config-2_1.yml` high resource and create
-   the patch file based on the diff (don't commit it though).
```
ekaterinadimitrova2, line 105: I am lost... so you say the same as before - update the config-2_1.yml file and the script will just make the diffs for you?
Unfortunately running the script failed for me:
edimitri-rmbp15:.circleci ekaterina.dimitri$ ./create-patches.sh
```
Traceback (most recent call last):
  File "/Users/ekaterina.dimitri/IdeaProjects/april/cassandra/.circleci/circle-ci-config.py", line 32, in <module>
    import yaml
ModuleNotFoundError: No module named 'yaml'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/cassandra/.circleci/circle-ci-config.py", line 39, in <module>
    import yaml
ModuleNotFoundError: No module named 'yaml'
```

ekaterinadimitrova2: Can we say what the other scripts do and when do we update them? 
Also, based on your feedback, I guess we need to add that "apply" means make the changes as per your expectations not just create file without error.

.circleci/circle-ci-config.py
dcapwell: agree, just pointing out its a waste of resources; we could be nicer if we don't need xlarge

.circleci/readme.md
dcapwell: @ekaterinadimitrova2 did it actually fail or did it just log that with no other output?  when that error happens we install yaml and move on

ekaterinadimitrova2: In my experience people will complain if they have to install modules and it doesn't work our of the box. Probably one of the reasons why shell scripts were chosen

jmckenzie-dev:
Could we get more detail on the help output for these scripts? It's not clear what they're supposed to do, how they work, how to test, etc w/out opening them up. Nothing in readme.md and the help output is:
```./circle-ci-config.py
usage: circle-ci-config.py [-h] [--stdout] target
circle-ci-config.py: error: the following arguments are required: target
```

Also, having "HIGHER" as a target when we have LOWRES, MIDRES, HIGHRES seems pretty confusing for me:
```
51 targets = {
52         'MIDRES': {
...
89         'HIGHER': {
```

PR #1580, opened 2022-04-20
[CASSANDRA-17568] add datapaths subcommand to nodetool

rtib:


src/java/org/apache/cassandra/tools/nodetool/DataPaths.java
```diff
@@ -0,0 +1,54 @@
...
+import io.airlift.airline.Command;
+import io.airlift.airline.Option;
+import org.apache.cassandra.tools.NodeProbe;
+import org.apache.cassandra.tools.NodeTool.NodeToolCmd;
+import org.apache.cassandra.tools.nodetool.stats.DataPathsHolder;
+import org.apache.cassandra.tools.nodetool.stats.DataPathsPrinter;
+import org.apache.cassandra.tools.nodetool.stats.StatsHolder;
+import org.apache.cassandra.tools.nodetool.stats.StatsPrinter;
+
+@Command(name = "datapaths", description = "Print paths where the table data is stored in")
```
smiklosovic: _data_ is plural. _datum_ is singular. You can not use "data is stored". I would do it like "Print all directories where data of tables are stored".

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
```diff
@@ -0,0 +1,81 @@
...
+import org.apache.cassandra.db.ColumnFamilyStoreMBean;
+import org.apache.cassandra.tools.NodeProbe;
+
+public class DataPathsHolder implements StatsHolder
+{
+    public final Map<String, Object> pathsHash;
+
+    public DataPathsHolder(NodeProbe probe, List<String> tableNames)
+    {
+        this.pathsHash = new HashMap();
```
smiklosovic: I would do it like `new HashMap<>();` to keep IDE silent about it.

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
```diff
@@ -0,0 +1,81 @@
...
+
+        Iterator<Map.Entry<String, ColumnFamilyStoreMBean>> tableMBeans = probe.getColumnFamilyStoreMBeanProxies();
```
smiklosovic: I would rename this to `mbeansIterator`. The thing the method on probe returns is not "tableMBeans". It is an iterator you need to iterate over to get them.

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
```diff
@@ -0,0 +1,81 @@
...
+        while (tableMBeans.hasNext())
+        {
+            Map.Entry<String, ColumnFamilyStoreMBean> entry = tableMBeans.next();
+            String keyspaceName = entry.getKey();
+            String tableName = entry.getValue().getTableName();
+
+            if (!(tableNames.isEmpty() || 
+                  tableNames.contains(keyspaceName + "." + tableName) ||
```
smiklosovic: `"."` can be simplified with `'.'`

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
```diff
@@ -0,0 +1,81 @@
...
+            Map<String, List<String>> ksPaths;
+            List<String> dataPaths;
+
+            try
+            {
+                dataPaths = entry.getValue().getDataPaths();
+            }
+            catch (IOException ex)
+            {
+                continue;
```
smiklosovic, line 63: This should be logged with logger on warn level.

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
```diff
@@ -0,0 +1,81 @@
...
+            }
+
+            if (pathsHash.containsKey(keyspaceName))
+            {
+                ksPaths = (Map<String, List<String>>) pathsHash.get(keyspaceName);
+            }
+            else
+            {
+                ksPaths = new HashMap();
```
smiklosovic: `ksPaths = new HashMap<>();`

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
```diff
@@ -0,0 +1,81 @@
...
+                dataPaths = entry.getValue().getDataPaths();
+            }
+            catch (IOException ex)
+            {
+                continue;
+            }
+
+            if (pathsHash.containsKey(keyspaceName))
+            {
+                ksPaths = (Map<String, List<String>>) pathsHash.get(keyspaceName);
```
smiklosovic, line 68: it is sad we need to do this and that `StatsHolder` is not generic in this regard.

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsPrinter.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.io.PrintStream;
+import java.util.List;
+import java.util.Map;
+
+public class DataPathsPrinter<T extends StatsHolder>
+{
+    public static <T extends StatsHolder> StatsPrinter<T> from (String format)
+    {
+        switch(format)
```
smiklosovic: I would simplify this to plain ifs. There is no need to have case for this really. It reads shorter and we are hardly ever going to have more output formats anyway.

    public static <T extends StatsHolder> StatsPrinter<T> from (String format)
    {
        if (format.equals("json"))
            return new StatsPrinter.JsonPrinter<T>();
        if (format.equals("yaml"))
            return new StatsPrinter.YamlPrinter<T>();
        
        return (StatsPrinter<T>) new DefaultPrinter();
    }

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsPrinter.java
```diff
@@ -0,0 +1,63 @@
...
+            case "yaml":
+                return new StatsPrinter.YamlPrinter<T>();
+            default:
+                return (StatsPrinter<T>) new DefaultPrinter();
+        }
+    }
+    
+    public static class DefaultPrinter implements StatsPrinter<DataPathsHolder>
+    {
+
```
smiklosovic: This empty line seems to be redundant.

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsPrinter.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+package org.apache.cassandra.tools.nodetool.stats;
+
+import java.io.PrintStream;
+import java.util.List;
+import java.util.Map;
+
+public class DataPathsPrinter<T extends StatsHolder>
+{
+    public static <T extends StatsHolder> StatsPrinter<T> from (String format)
```
smiklosovic: @rtib by the way this whole method might be changed to this:

    public static StatsPrinter<DataPathsHolder> from(String format)
    {
        if (format.equals("json"))
            return new StatsPrinter.JsonPrinter<>();
        if (format.equals("yaml"))
            return new StatsPrinter.YamlPrinter<>();

        return new DefaultPrinter();
    }

You do not need to do it with generics.

Then in nodetool command, you need to change it like this:

        DataPathsHolder holder = new DataPathsHolder(probe, tableNames);
        StatsPrinter<DataPathsHolder> printer = DataPathsPrinter.from(outputFormat);
        printer.print(holder, probe.output().out);

It will be nicely done without any casts or whatever.

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsPrinter.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+    }
+    
+    public static class DefaultPrinter implements StatsPrinter<DataPathsHolder>
+    {
+
+        @Override
+        public void print(DataPathsHolder data, PrintStream out)
+        {
+            for (String keyspace : data.pathsHash.keySet())
```
smiklosovic: my IDE informs me that this might be reworked with

            for (Map.Entry<String, Object> entry : data.pathsHash.entrySet())
            {
                out.println("Keyspace : " + entry.getKey());
                Map<String, List<String>> ksPaths = (Map<String, List<String>>) entry.getValue();

You would work with entry instead of getting keyspaces as keys and then getting its value via "get" from the map again. That is not necessary. It is better to work with isolated entry instead only.

smiklosovic: or maybe not huh, this is complicated / confusing, i need to go over this again.

smiklosovic: Yes, the ultimate version of this is like this:

            for (Map.Entry<String, Object> entry : data.pathsHash.entrySet())
            {
                out.println("Keyspace : " + entry.getKey());
                Map<String, List<String>> ksPaths = (Map<String, List<String>>) entry.getValue();
                for (Map.Entry<String, List<String>> table : ksPaths.entrySet())
                {
                    out.println("\tTable : " + table.getKey());
                    out.println("\tPaths :");
                    for (String path : table.getValue())

test/unit/org/apache/cassandra/tools/nodetool/DataPathsTest.java
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    @Test
+    public void testHelp()
+    {
+        ToolRunner.ToolResult tool = ToolRunner.invokeNodetool("help", SUBCOMMAND);
+        tool.assertOnCleanExit();
+        
+        String help = "NAME\n" +
+                      "        nodetool datapaths - Print paths where the table data is stored in\n" +
```
smiklosovic: you need to change the description here as well

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsPrinter.java
smiklosovic: @rtib I do not like the fact that there is "------------" at the end of the output when there is not other table to print. You would need to rewrite this to solution using Iterator and you would ask if you are going to print "-----" only if iterator has next entry, like this:

        public void print(DataPathsHolder data, PrintStream out)
        {
            Iterator<Map.Entry<String, Object>> iterator = data.pathsHash.entrySet().iterator();

            while (iterator.hasNext())
            {
                Map.Entry<String, Object> entry = iterator.next();

                out.println("Keyspace : " + entry.getKey());
                Map<String, List<String>> ksPaths = (Map<String, List<String>>) entry.getValue();
                for (Map.Entry<String, List<String>> table : ksPaths.entrySet())
                {
                    out.println("\tTable : " + table.getKey());
                    out.println("\tPaths :");

                    for (String path : table.getValue())
                        out.println("\t\t" + path);

                    out.println("");
                }

                if (iterator.hasNext())
                    out.println("----------------");
            }
        }

doc/modules/cassandra/pages/troubleshooting/use_nodetool.adoc
```diff
@@ -240,3 +240,46 @@ concurrent compactions such that compactions complete quickly but don't
...
+	Paths :
+		var/lib/cassandra/data/system_auth/resource_role_permissons_index-5f2fbdad91f13946bd25d5da3a5c35ec
+
+	Table : roles
+	Paths :
+		var/lib/cassandra/data/system_auth/roles-5bc52802de2535edaeab188eecebb090
+
+	Table : role_members
+	Paths :
+		var/lib/cassandra/data/system_auth/role_members-0ecdaa87f8fb3e6088d174fb36fe5c0d
```
smiklosovic: why are there no starting slashes here? should be `/var/...` not `var/`.

doc/modules/cassandra/pages/troubleshooting/use_nodetool.adoc
```diff
@@ -240,3 +240,46 @@ concurrent compactions such that compactions complete quickly but don't
...
+
+----------------
```
smiklosovic: remove this line if you follow my advice to do it via iterators.

doc/modules/cassandra/pages/troubleshooting/use_nodetool.adoc
```diff
@@ -240,3 +240,46 @@ concurrent compactions such that compactions complete quickly but don't
 take too many resources away from query threads is very important for
 performance. If you notice compaction unable to keep up, try tuning
 Cassandra's `concurrent_compactors` or `compaction_throughput` options.
+
+[[nodetool-datapaths]]
+== Paths used for data files
+
+Cassandra is persisting data on disk within the configured directories. Data
```
smiklosovic, line 247: mentioning that it supports json as well as yaml output formats would be nice.

src/java/org/apache/cassandra/tools/nodetool/DataPaths.java
```diff
@@ -0,0 +1,54 @@
...
+    @Override
+    protected void execute(NodeProbe probe)
+    {
+        if (!outputFormat.isEmpty() && !"json".equals(outputFormat) && !"yaml".equals(outputFormat))
+        {
+            throw new IllegalArgumentException("arguments for -F are yaml and json only.");
+        }
+
+        StatsHolder holder = new DataPathsHolder(probe, tableNames);
+        StatsPrinter printer = DataPathsPrinter.from(outputFormat);
```
smiklosovic: I would rename this method from `from` to `for`. We want to get a printer _for_ an output format, not _from_ it.
EDIT: aha ... we can not use `for` as it is reserved keyword :) Thats a bummer ... Ok, `from` is fine too then.
EDIT 2: `forFormat` would be nice as well.

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.cassandra.db.ColumnFamilyStoreMBean;
+import org.apache.cassandra.tools.NodeProbe;
+
+public class DataPathsHolder implements StatsHolder
+{
+    public final Map<String, Object> pathsHash;
```
smiklosovic: is there any reason to not have it initialised directly here with `= new HashMap<>()`? I think you do not need to initialise it via `this.pathsHash` in the constructor.

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
```diff
@@ -0,0 +1,81 @@
...
+                pathsHash.put(keyspaceName, ksPaths);
+            }
+            ksPaths.put(tableName, dataPaths);
+        }
+    }
+
+    @Override
+    public Map<String, Object> convert2Map()
+    {
+        return this.pathsHash;
```
smiklosovic: `this` is redundant.

NEWS.txt
```diff
@@ -56,6 +56,7 @@ using the provided 'sstableupgrade' tool.
 
 New features
 ------------
+    - Tool to list data paths of existing tables
```
smiklosovic: I do not think this is necessary. It is not _tool_ as in a standalone tool, it is still the part of _nodetool_. Mentioning this in CHANGES.txt is enough.

test/unit/org/apache/cassandra/tools/nodetool/DataPathsTest.java
```diff
@@ -0,0 +1,169 @@
...
+        assertThat(tool.getStdout()).contains("Table : roles");
+        assertThat(tool.getStdout()).contains("Table : role_members");
+        assertThat(StringUtils.countMatches(tool.getStdout(), "\tTable :")).isEqualTo(2);
+        assertThat(StringUtils.countMatches(tool.getStdout(), "\tPaths :")).isEqualTo(2);
+    }
+
+    @Test
+    public void testFormatArgJson()
+    {
+        ToolRunner.ToolResult tool = ToolRunner.invokeNodetool(SUBCOMMAND, "--format", "json");
```
smiklosovic, line 153: for both json and yaml output formats, I would get the output as string and then I would put it back to some json lib to parse it so we would eat what we produced - doing full circle here. Hence we would be sure that what we produced makes sense and it is valid output.

I leave this for the author to consider it but it is not the absolute must to do.

rtib: This is done in [StatsPrinterTest.java](https://github.com/apache/cassandra/blob/trunk/test/unit/org/apache/cassandra/tools/nodetool/stats/StatsPrinterTest.java), isn't it?

smiklosovic: Probably, we can leave it out, yes. I think that is better place to do these kinds of tests.

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
rtib: Hm... NetBeans didn't complain on that. Oops, did I just revealed my age?

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
rtib: Does nodetool has a logger? Or just print a warning to stderr?

rtib: I was looking into it, and I don't understand why [ColumnFamilyStoreMBean.getDataPaths()](https://github.com/apache/cassandra/blob/d80bb323c9dd8f814fe02e16dec510557c7d5101/src/java/org/apache/cassandra/db/ColumnFamilyStoreMBean.java#L268) does announce IOException to be thrown. The only productive code implementing that interface is [ColumnFamilyStore.getDataPaths()](https://github.com/apache/cassandra/blob/d80bb323c9dd8f814fe02e16dec510557c7d5101/src/java/org/apache/cassandra/db/ColumnFamilyStore.java#L600) which doesn't seem to throw any exception. Haven't check the bunch of MockCFS implementations, but it seems they are for testing purposes only. Did I miss something?

src/java/org/apache/cassandra/tools/nodetool/DataPaths.java
rtib: I could refactor that, but the other implementations I used as blueprint are also using this naming:

- https://github.com/apache/cassandra/blob/d80bb323c9dd8f814fe02e16dec510557c7d5101/src/java/org/apache/cassandra/tools/nodetool/stats/TableStatsPrinter.java#L28
- https://github.com/apache/cassandra/blob/d80bb323c9dd8f814fe02e16dec510557c7d5101/src/java/org/apache/cassandra/tools/nodetool/stats/TpStatsPrinter.java#L36
- https://github.com/apache/cassandra/blob/d80bb323c9dd8f814fe02e16dec510557c7d5101/src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryPrinter.java#L32

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
smiklosovic: If you checkout the original commit which was pushed, back in time it was possible that the exception might be thrown with `dataPath.getCanonicalPath()`

But then there was patch of CASSANDRA-16926 CEP-10 Phase 1, which changed that code so it stopped to throw it but the method signature stayed.

If you dig deep enough, there is `PathUtils.toRealPath` called on that line which wrapps IOException and it throws unchecked exception and if you eventually dig even further it will ultimately wrap that IOError into FSReadError which is runtime exception.

Hence, I would change this to catching `Throwable t` there and I would log that one.

https://issues.apache.org/jira/browse/CASSANDRA-16335

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsPrinter.java
rtib: Oh, I was also stumbling over that one, however, I found the readability of having a loop variable of type `String` some better than `Map.Entry<String, List<String>>`. What if solving the separator problem otherwise?

smiklosovic: I prefer the iterator approach and working with one entry only at the time. Having keys in for loop and then performing get on that map again is not too much performance friendly. It also solves not printing the separator at the very end of the output by looking if hasNext returns true.

src/java/org/apache/cassandra/tools/nodetool/DataPaths.java
smiklosovic: ok keep it.

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
smiklosovic: You can reuse

        output.err.println("error: " + e.getMessage());
        output.err.println("-- StackTrace --");
        output.err.println(getStackTraceAsString(e));

This code is in `NodeTool.err(Throwable)` method but it is protected so it can not be called directly in `DataPathsHolder` but since you have access to `NodeProbe` in that method, you can call it via `probe.output().err.println()`.

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
rtib: I'm afraid there's nothing we can do with that right now.

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsHolder.java
smiklosovic: @rtib please `catch (Throwable t)` instead of IOException

rtib: Changed that, but some IDE (e.g. NetBeans) hint on that to change back.

smiklosovic: Yes because it still thinks that we are throwing that from the method based on its signature which is, as explained not true. To be safe, lets just catch it all and thats it.

src/java/org/apache/cassandra/tools/nodetool/DataPaths.java
```diff
@@ -38,16 +45,46 @@ public class DataPaths extends NodeToolCmd
     @Option(title = "format", name = {"-F", "--format"}, description = "Output format (json, yaml)")
     private String outputFormat = "";
 
+    @Option(title = "",
```
smiklosovic: What is the empty title good for? What about "orphaned"?

smiklosovic: I would also add short form like "-o".

src/java/org/apache/cassandra/tools/nodetool/DataPaths.java
```diff
@@ -38,16 +45,46 @@ public class DataPaths extends NodeToolCmd
...
-        {
             throw new IllegalArgumentException("arguments for -F are yaml and json only.");
-        }
 
         DataPathsHolder holder = new DataPathsHolder(probe, tableNames);
-        StatsPrinter<DataPathsHolder> printer = DataPathsPrinter.from(outputFormat);
-        printer.print(holder, probe.output().out);
+        if (orphaned)
+        {
+            if (!outputFormat.isEmpty() || !tableNames.isEmpty())
```
smiklosovic: I would move this up, outside from this `if`. Because here we might fail but you were calling Cassandra by `DataPathsHolder` already and unnecessarilly. You queried Cassandra to return you stuff but you failed. I would fail as soon as possible, doing JMX query only if we are sure we are going to need it.

src/java/org/apache/cassandra/tools/nodetool/DataPaths.java
```diff
@@ -38,16 +45,46 @@ public class DataPaths extends NodeToolCmd
...
+                throw new IllegalArgumentException("Format option, keyspace and table name arguments aren't supported when listing orphans.");
+
+            Set<String> dataDirs = new HashSet<>();
+            dataDirs.addAll(Arrays.asList(probe.getStorageService().getNonLocalSystemKeyspacesDataFileLocations()));
+            dataDirs.addAll(Arrays.asList(probe.getStorageService().getLocalSystemKeyspacesDataFileLocations()));
+            
+            Set<File> allFiles = new HashSet<>();
```
smiklosovic: If you want to use `File`, we have a special `File` class in `org.apache.cassandra.io.util.File` instead of `java.io.File`. I think it would fail to compile by ant if you used the original Java one, there is some check for it.

src/java/org/apache/cassandra/tools/nodetool/DataPaths.java
```diff
@@ -38,16 +45,46 @@ public class DataPaths extends NodeToolCmd
...
+            for (String dataDir : dataDirs)
```
smiklosovic: I need to digest what this is doing. I ll get back to you today / tomorrow. Thanks for your work so far. Stellar cooperation.

src/java/org/apache/cassandra/tools/nodetool/DataPaths.java
rtib: I've reverted that commit, as Brandon confirmed my concerns.

src/java/org/apache/cassandra/tools/nodetool/DataPaths.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.cassandra.tools.nodetool.stats.DataPathsPrinter;
+import org.apache.cassandra.tools.nodetool.stats.StatsPrinter;
+
+@Command(name = "datapaths", description = "Print all directories where data of tables are stored")
+public class DataPaths extends NodeToolCmd
+{
+    @Arguments(usage = "[<keyspace.table>...]", description = "List of table (or keyspace) names")
+    private List<String> tableNames = new ArrayList<>();
+
+    @Option(title = "format", name = {"-F", "--format"}, description = "Output format (json, yaml)")
```
smiklosovic, line 38: @rtib why is this upper cased? "F" ? What is wrong with just "f". Is it used somewhere already?

src/java/org/apache/cassandra/tools/nodetool/stats/DataPathsPrinter.java
```diff
@@ -0,0 +1,66 @@
...
+        @Override
+        public void print(DataPathsHolder data, PrintStream out)
+        {
+            Iterator<Map.Entry<String, Object>> iterator = data.pathsHash.entrySet().iterator();
+
+            while (iterator.hasNext())
+            {
+                Map.Entry<String, Object> entry = iterator.next();
+
+                out.println("Keyspace : " + entry.getKey());
```
smiklosovic: @rtib could we have just "Keyspace: " (without space between "Keyspace" and ":" ? Same for Table and Paths. I do not think that there is space in normal English. Wherever I looked, colon follows the previous word immediately. Please change the documentation too.

src/java/org/apache/cassandra/tools/nodetool/DataPaths.java
rtib: There is nothing wrong with the lower case "f" option, I guess. But, TableStats, TpStats and CompactionStats are using it identically. That is the whole bunch of subcommands built on the StatsHolder/StatsPrinter infrastructure of org.apache.cassandra.tools.nodetool.stats. Even if datapaths subcommand is not a statistics feature, I found the use of StatsHolder/StatsPrinter useful, convenient and a suitable pattern.

smiklosovic: fair enough.

PR #1582, opened 2022-04-21
CASSANDRA-17500 Create Maximum Keyspace Replication Factor Guardrail

thingtwin1:


test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,228 @@
...
+import org.apache.cassandra.service.StorageService;
+
+import static java.lang.String.format;
+
+public class GuardrailMaximumReplicationFactorTest extends ThresholdTester
+{
+    private static final int MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD = 2;
+    private static final int MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD = 4;
+    private static final int DISABLED_GUARDRAIL = -1;
+    private static final String WHAT = "maximum_replication_factor";
```
adelapena: Unused, probably because, differently from `GuardrailMinimumReplicationFactorTest`, there isn't a `testConfigValidation` test method.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -333,6 +333,19 @@ public final class Guardrails implements GuardrailsMBean
...
+     */
+    public static final MaxThreshold maximumReplicationFactor =
+    new MaxThreshold("maximum_replication_factor",
+                  state -> CONFIG_PROVIDER.getOrCreate(state).getMaximumReplicationFactorWarnThreshold(),
+                  state -> CONFIG_PROVIDER.getOrCreate(state).getMaximumReplicationFactorFailThreshold(),
+                  (isWarning, what, value, threshold) ->
+                  isWarning ? format("The keyspace %s has a replication factor of %s, above the warning threshold of %s.",
+                                     what, value, threshold)
+                            :format("The keyspace %s has a replication factor of %s, above the failure threshold of %s.",
+                                    what, value, threshold));
```
adelapena: Nit: misaligned
```suggestion
                            : format("The keyspace %s has a replication factor of %s, above the failure threshold of %s.",
                                     what, value, threshold));
```

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,228 @@
...
+                    format("The keyspace %s has a replication factor of 6, above the failure threshold of %s", KS, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD));
+    }
+
+    @Test
+    public void testMaxKeyspaceRFWarnBelow() throws Throwable
+    {
+        guardrails().setMaximumReplicationFactorThreshold(MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD);
+        assertMaxThresholdValid("CREATE KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 1}");
+        assertMaxThresholdValid("ALTER KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 2}");
+
```
adelapena: Nit: unneeded blank line

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    @Test
+    public void testMaxKeyspaceRFOnlyFailBelow() throws Throwable
+    {
+        guardrails().setMaximumReplicationFactorThreshold(DISABLED_GUARDRAIL, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD);
+        assertMaxThresholdValid("CREATE KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 2}");
+        assertMaxThresholdValid("ALTER KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 3}");
+
```
adelapena: Nit: unneeded blank line.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,228 @@
...
+    @Test
+    public void testMaxKeyspaceRFOnlyWarnAbove() throws Throwable
+    {
+        guardrails().setMaximumReplicationFactorThreshold(MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD, DISABLED_GUARDRAIL);
+        assertWarns("CREATE KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 3}",
+                    format("The keyspace %s has a replication factor of 3, above the warning threshold of %s.", KS, MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD));
+        assertWarns("ALTER KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 4}",
+                    format("The keyspace %s has a replication factor of 4, above the warning threshold of %s.", KS, MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD));
+
```
adelapena: Nit: unneeded blank line.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,228 @@
...
+        execute("DROP KEYSPACE IF EXISTS ks1");
+    }
+
+    @Test
+    public void testMaxKeyspaceRFOnlyWarnBelow() throws Throwable
+    {
+        guardrails().setMaximumReplicationFactorThreshold(MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD, DISABLED_GUARDRAIL);
+        assertMaxThresholdValid("CREATE KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 2}");
+        assertMaxThresholdValid("ALTER KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 1}");
+
```
adelapena: Nit: unneeded blank line.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,228 @@
...
+                    format("he keyspace %s has a replication factor of 5, above the failure threshold of %s", KS, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD));
+    }
+
+    @Test
+    public void testMaxKeyspaceRFFailAboveAlter() throws Throwable
+    {
+        guardrails().setMaximumReplicationFactorThreshold(MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD);
+        execute("CREATE KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 4}");
+       assertFails("ALTER KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 5}",
+                   format("he keyspace %s has a replication factor of 5, above the failure threshold of %s", KS, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD));
```
adelapena:
```suggestion
        assertFails("ALTER KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 5}",
                    format("The keyspace %s has a replication factor of 5, above the failure threshold of %s", KS, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD));
```

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -82,6 +82,7 @@ public GuardrailsOptions(Config config)
         validatePercentageThreshold(config.data_disk_usage_percentage_warn_threshold, config.data_disk_usage_percentage_fail_threshold, "data_disk_usage_percentage");
         validateDataDiskUsageMaxDiskSize(config.data_disk_usage_max_disk_size);
         validateMinRFThreshold(config.minimum_replication_factor_warn_threshold, config.minimum_replication_factor_fail_threshold, "minimum_replication_factor");
+        validateMaxIntThreshold(config.maximum_replication_factor_warn_threshold, config.maximum_replication_factor_fail_threshold, "maximum_replication_factor");
```
adelapena: Probably we should validate this against `default_keyspace_rf`, as we did with the guardrail for min RF.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -333,6 +333,19 @@ public final class Guardrails implements GuardrailsMBean
...
 
+    /**
+     * Guardrail on the maximum replication factor.
+     */
+    public static final MaxThreshold maximumReplicationFactor =
+    new MaxThreshold("maximum_replication_factor",
+                  state -> CONFIG_PROVIDER.getOrCreate(state).getMaximumReplicationFactorWarnThreshold(),
+                  state -> CONFIG_PROVIDER.getOrCreate(state).getMaximumReplicationFactorFailThreshold(),
+                  (isWarning, what, value, threshold) ->
+                  isWarning ? format("The keyspace %s has a replication factor of %s, above the warning threshold of %s.",
```
jmckenzie-dev: May be able to tidy this up with something like:
```
format("The keyspace %s has a replication factor of %s, above the %s threshold of %s),
             what, value, isWarning ? "warning" : "failure", threshold);
```
Gets you down to one line with the ternary on the string argument only.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,228 @@
...
+
+import static java.lang.String.format;
+
+public class GuardrailMaximumReplicationFactorTest extends ThresholdTester
+{
+    private static final int MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD = 2;
+    private static final int MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD = 4;
+    private static final int DISABLED_GUARDRAIL = -1;
+    private static final String WHAT = "maximum_replication_factor";
+    private static final String KS = "ks";
```
jmckenzie-dev: You have inconsistent usage of this String throughout the tests. I'd either settle on using the KS string everywhere or just hardcode to ks everywhere.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,228 @@
...
+
+    @Override
+    protected List<String> getWarnings()
+    {
+        List<String> warnings = ClientWarn.instance.getWarnings();
+
+        return warnings == null
+               ? Collections.emptyList()
+               : warnings.stream()
+                         .filter(w -> !w.contains("keyspace ks is higher than the number of nodes 1 for datacenter1") &&
```
jmckenzie-dev, line 81: It isn't immediately obvious to me why you're filtering out this set of warnings from your list. Maybe leave a quick comment to explain that these are expected warnings if that is indeed the case.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,228 @@
...
+    @Test
+    public void testMaxKeyspaceRFDisabled() throws Throwable
+    {
+        guardrails().setMaximumReplicationFactorThreshold(DISABLED_GUARDRAIL, DISABLED_GUARDRAIL);
+        assertMaxThresholdValid("CREATE KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 6}");
+        assertMaxThresholdValid("ALTER KEYSPACE ks WITH replication = { 'class' : 'NetworkTopologyStrategy', 'datacenter1': 10}");
+    }
+
+    @Test
+    public void testSimpleStrategy() throws Throwable
```
jmckenzie-dev: nit: you're testing CREATE + warn, and ALTER + fail. Might be worth testing all four case combinations just for the sake of completeness. Really would only matter in some timeline where someone decouples the two in the future from each other.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,228 @@
...
+            @Override
+            public int compareEndpoints(InetAddressAndPort target, Replica a1, Replica a2) { return 0; }
+        });
+
+        List<String> twoWarnings = Arrays.asList(format("The keyspace %s has a replication factor of 3, above the warning threshold of %s.", KS, MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD),
+                                                 format("The keyspace %s has a replication factor of 3, above the warning threshold of %s.", KS, MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD));
+
+        StorageService.instance.getTokenMetadata().updateHostId(UUID.randomUUID(), InetAddressAndPort.getByName("127.0.0.255"));
+        guardrails().setMaximumReplicationFactorThreshold(MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD);
+        assertValid("CREATE KEYSPACE ks WITH replication = { 'class' : 'NetworkTopologyStrategy', 'datacenter1': 2, 'datacenter2' : 2 };");
```
jmckenzie-dev: Same feedback here. Might want to have this test walk through the cases like:
```
assertValid("CREATE... (with small #)
assertWarns("CREATE... (with medium #)
assertFails("CREATE... (with very big #)

assertValid("ALTER... (small)
assertWarns("ALTER... (medium)
assertFails("ALTER... (large)
```

That way you have all your bases covered.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,228 @@
...
+    public void testMaxKeyspaceRFWarnBelow() throws Throwable
+    {
+        guardrails().setMaximumReplicationFactorThreshold(MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD);
+        assertMaxThresholdValid("CREATE KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 1}");
+        assertMaxThresholdValid("ALTER KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 2}");
+
+    }
+
+    @Test
+    public void testMaxKeyspaceRFWarnFailBetween() throws Throwable
```
jmckenzie-dev, line 206: Why is this named "WarnFail"? Looks like it's a warn only test?

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,228 @@
...
+        assertFails("CREATE KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 5}",
+                    format("he keyspace %s has a replication factor of 5, above the failure threshold of %s", KS, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD));
+    }
+
+    @Test
+    public void testMaxKeyspaceRFFailAboveAlter() throws Throwable
+    {
+        guardrails().setMaximumReplicationFactorThreshold(MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD);
+        execute("CREATE KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 4}");
+       assertFails("ALTER KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 5}",
```
jmckenzie-dev: nit: alignment on this line

jmckenzie-dev (changes requested):

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
thingtwin1: hardcoded the keyspace everywhere

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
thingtwin1: separated out the tests with CREATE(warn + fail) and ALTER(warn + fail) cases

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
thingtwin1: it tests values that are in between warn and fail

conf/cassandra.yaml
```diff
@@ -1791,6 +1791,10 @@ drop_compact_storage_enabled: false
 # Suggested value for use in production: 2 or higher
 # minimum_replication_factor_warn_threshold: -1
 # minimum_replication_factor_fail_threshold: -1
+# Guardrail to warn or fail when the maximum replication factor is greater than threshold.
```
adelapena, line 1816: Recent changes have introduced a blank line separating the configs of every guardrail, we should also introduce that empty line here to keep the format:
```suggestion
#
# Guardrail to warn or fail when the maximum replication factor is greater than threshold.
```

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -648,6 +649,31 @@ public void setMinimumReplicationFactorThreshold(int warn, int fail)
...
+
+    @Override
+    public int getMaximumReplicationFactorFailThreshold()
+    {
+        return config.maximum_replication_factor_fail_threshold;
+    }
+
+    public void setMaximumReplicationFactorThreshold(int warn, int fail)
+    {
+        validateMaxIntThreshold(warn, fail, "maximum_replication_factor");
```
adelapena: This should be `validateMaxRFThreshold`. This would have been detected by a test equivalent to [`GuardrailMinimumReplicationFactorTest#testMinRFGreaterThanDefaultRF()`](https://github.com/apache/cassandra/blob/12751587818de4ae586a02c7d8568d815a4c0956/test/unit/org/apache/cassandra/db/guardrails/GuardrailMinimumReplicationFactorTest.java#L259-L276), but we don't have an equivalent test in `GuardrailMaximumReplicationFactorTest`.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -738,6 +770,15 @@ private static void validateMinRFVersusDefaultRF(int fail, String name) throws I
         }
     }
 
+    private static void validateMaxRFVersusDefaultRF(int fail, String name) throws IllegalArgumentException
```
adelapena: I think we don't need the explicit `throws IllegalArgumentException`.

adelapena: The only caller of `validateMaxRFVersusDefaultRF` is `validateMaxRFThreshold`, and I don't think we are going to have any other callers. We might want to inline the contents of this method into `validateMaxRFThreshold`. Same for `validateMinRFVersusDefaultRF`.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -738,6 +770,15 @@ private static void validateMinRFVersusDefaultRF(int fail, String name) throws I
...
+    {
+        if (fail != -1 && fail < DatabaseDescriptor.getDefaultKeyspaceRF())
+        {
+            throw new IllegalArgumentException(String.format("%s_fail_threshold to be set (%d) cannot be less than default_keyspace_rf (%d)",
```
adelapena:
```suggestion
            throw new IllegalArgumentException(String.format("%s_fail_threshold to be set (%d) cannot be lesser than default_keyspace_rf (%d)",
```

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -341,6 +341,17 @@ public final class Guardrails implements GuardrailsMBean
...
+    /**
+     * Guardrail on the maximum replication factor.
+     */
+    public static final MaxThreshold maximumReplicationFactor =
+    new MaxThreshold("maximum_replication_factor",
+                  state -> CONFIG_PROVIDER.getOrCreate(state).getMaximumReplicationFactorWarnThreshold(),
+                  state -> CONFIG_PROVIDER.getOrCreate(state).getMaximumReplicationFactorFailThreshold(),
+                  (isWarning, what, value, threshold) ->
+                  format("The keyspace %s has a replication factor of %s, above the %s threshold of %s",
+                         what, value, isWarning ? "warning" : "failure", threshold));
```
adelapena: We could use this same cleaner string formatting in the analogous `minimumReplicationFactor`.

adelapena: This should probably end with a period:
```suggestion
                  format("The keyspace %s has a replication factor of %s, above the %s threshold of %s.",
                         what, value, isWarning ? "warning" : "failure", threshold));
```
The new unit tests seem to expect that dot, and I think that's why some of them fail. Did you run them locally?

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -553,4 +553,21 @@
      */
     void setMinimumReplicationFactorThreshold (int warn, int fail);
 
+    /**
+     * @return The threshold to fail when replication factor is greater threshold.
```
adelapena:
```suggestion
     * @return The threshold to fail when replication factor is greater than threshold.
```
This also applies to the equivalent `getMinimumReplicationFactorWarnThreshold`

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -553,4 +553,21 @@
...
+     */
+    int getMaximumReplicationFactorWarnThreshold();
+
+    /**
+     * @return The threshold to fail when replication factor is greater threshold.
```
adelapena:
```suggestion
     * @return The threshold to fail when replication factor is greater than threshold.
```
This also applies to the equivalent `getMinimumReplicationFactorFailThreshold`.

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -553,4 +553,21 @@
...
+     */
+    int getMaximumReplicationFactorFailThreshold();
+
+    /**
+     * @param warn the threshold to warn when the maximum replication factor is greater than
+     *             threshold -1 means disabled.
+     * @param fail the threshold to fail when the maximum replication factor is greater than
+     *             threshold -1 means disabled.
```
adelapena:
```suggestion
     * @param warn The threshold to warn when the maximum replication factor is greater than
     *             threshold. -1 means disabled.
     * @param fail The threshold to fail when the maximum replication factor is greater than
     *             threshold. -1 means disabled.
```
This also applies to the equivalent `setMinimumReplicationFactorThreshold`.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,246 @@
...
+import org.apache.cassandra.locator.AbstractEndpointSnitch;
+import org.apache.cassandra.locator.IEndpointSnitch;
+import org.apache.cassandra.locator.InetAddressAndPort;
+import org.apache.cassandra.locator.Replica;
+import org.apache.cassandra.service.ClientWarn;
+import org.apache.cassandra.service.StorageService;
+
+import static java.lang.String.format;
+
+public class GuardrailMaximumReplicationFactorTest extends ThresholdTester
```
adelapena, line 43: As mentioned above, multiple tests in this class fail due to a missed period in the error message provider for the testes guardrail.

Also, we are missing tests equivalent to `GuardrailMinimumReplicationFactorTest#testConfigValidation` and `GuardrailMinimumReplicationFactorTest#testMinRFGreaterThanDefaultRF`. I think those tests are useful because they help us to detect things like the missed validation in  `GuardrailOptions#setMaximumReplicationFactorThreshold` that we mentioned above.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,246 @@
...
+        assertValid("ALTER KEYSPACE ks WITH replication = { 'class' : 'NetworkTopologyStrategy', 'datacenter1': 2, 'datacenter2' : 2 };");
+        assertWarns("ALTER KEYSPACE ks WITH replication = { 'class' : 'NetworkTopologyStrategy', 'datacenter1': 2, 'datacenter2' : 3 };",
+                    format("The keyspace ks has a replication factor of 3, above the warning threshold of %s.", MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD));
+        assertWarns("ALTER KEYSPACE ks WITH replication = { 'class' : 'NetworkTopologyStrategy', 'datacenter1': 3, 'datacenter2' : 3 };", twoWarnings);
+        assertFails("ALTER KEYSPACE ks WITH replication = { 'class' : 'NetworkTopologyStrategy', 'datacenter1': 2, 'datacenter2' : 5 };",
+                    format("The keyspace ks has a replication factor of 5, above the failure threshold of %s", MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD));
+        assertFails("ALTER KEYSPACE ks WITH replication = { 'class' : 'NetworkTopologyStrategy', 'datacenter1': 5, 'datacenter2' : 5 };",
+                    format("The keyspace ks has a replication factor of 5, above the failure threshold of %s", MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD));
```
adelapena: The phrases of the form `The keyspace ks has a replication factor...` are repeated multiple times across the test. We could try to reduce duplication and make the tests a bit more compact with a couple of utility methods like, for example:
```java
private void assertWarns(String query, int rf) throws Throwable
{
    assertWarns(query, format("The keyspace ks has a replication factor of %d, above the warning threshold of %s.",
                              rf, MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD));
}

private void assertFails(String query, int rf) throws Throwable
{
    assertFails(query, format("The keyspace ks has a replication factor of %d, above the failure threshold of %s.",
                              rf, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD));
}
```

adelapena (changes requested):

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
thingtwin1: sorry, missed that when I ran the tests, added the . back in.

thingtwin1: I will make all the same cleanup changes for minimumReplicationFactor in another pr ?

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
thingtwin1: `ThresholdTester#testValidationOfThresholdProperties` covers all the combinations tested in `GuardrailMinimumReplicationFactorTest#testConfigValidation` . The method was added separately in `GuardrailsMinimumReplicationFactorTest` because checking if minimumReplicationFactor = Integer.MAX_VALUE would fail since we checked the value against the default_rf. maximumReplicationFactor doesn't have that constraint so we can use `ThresholdTester#testValidationOfThresholdProperties` to check the same thing ?

jmckenzie-dev:
👍 w/Andres' changes.

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
adelapena: As mentioned above, I would also apply this fix to `getMinimumReplicationFactorFailThreshold`

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
adelapena: As mentioned above, I would also apply this fix to `getMinimumReplicationFactorWarnThreshold`

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
adelapena: As mentioned above, I would also apply this fix to `setMinimumReplicationFactorThreshold`.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
adelapena: As mentioned above, I would use the same cleaner string formatting in `minimumReplicationFactor`.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -709,6 +735,16 @@ private static void validateMinRFThreshold(int warn, int fail, String name)
...
+    private static void validateMaxRFThreshold(int warn, int fail, String name)
+    {
+        validateMaxIntThreshold(warn, fail, name);
+        if (fail != -1 && fail < DatabaseDescriptor.getDefaultKeyspaceRF())
+        {
+            throw new IllegalArgumentException(String.format("%s_fail_threshold to be set (%d) cannot be lesser than default_keyspace_rf (%d)",
+                                                             name, fail, DatabaseDescriptor.getDefaultKeyspaceRF()));
+        }
+    }
+
```
adelapena: I would also inline `validateMinRFVersusDefaultRF` into `validateMinRFThreshold`, to keep consistency between these mostly symmetrical guardrails. Also, we can use `format` instead of `String.format` since we already have a static import for it. And we can also inline the `name` arguments, since they always have the same value, and try not to exceed the line width:
```java
private static void validateMinRFThreshold(int warn, int fail)
{
    validateMinIntThreshold(warn, fail, "maximum_replication_factor");

    if (fail > DatabaseDescriptor.getDefaultKeyspaceRF())
        throw new IllegalArgumentException(format("minimum_replication_factor_fail_threshold to be set (%d) " +
                                                  "cannot be greater than default_keyspace_rf (%d)",
                                                  fail, DatabaseDescriptor.getDefaultKeyspaceRF()));
}

private static void validateMaxRFThreshold(int warn, int fail)
{
    validateMaxIntThreshold(warn, fail, "maximum_replication_factor");

    if (fail != -1 && fail < DatabaseDescriptor.getDefaultKeyspaceRF())
        throw new IllegalArgumentException(format("maximum_replication_factor_fail_threshold to be set (%d) " +
                                                  "cannot be lesser than default_keyspace_rf (%d)",
                                                  fail, DatabaseDescriptor.getDefaultKeyspaceRF()));
}
```

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,257 @@
...
+        {
+            DatabaseDescriptor.setDefaultKeyspaceRF(1);
+            guardrails().setMaximumReplicationFactorThreshold(MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD);
+        }
+        catch (IllegalArgumentException e)
+        {
+            String expectedMessage = "";
+
+            if(guardrails().getMaximumReplicationFactorFailThreshold() < DatabaseDescriptor.getDefaultKeyspaceRF())
+                expectedMessage = format("%maximum_replication_factor_fail_threshold to be set (%d) cannot be lesser than default_keyspace_rf (%d)",
```
adelapena:
```suggestion
            if (guardrails().getMaximumReplicationFactorFailThreshold() < DatabaseDescriptor.getDefaultKeyspaceRF())
                expectedMessage = format("maximum_replication_factor_fail_threshold to be set (%d) cannot be lesser than default_keyspace_rf (%d)",
```

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,257 @@
...
+    @Test
+    public void testMaxKeyspaceRFFailAboveAlter() throws Throwable
+    {
+        guardrails().setMaximumReplicationFactorThreshold(MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD);
+        execute("CREATE KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 4}");
+        assertFails("ALTER KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 5}", 5);
+    }
+
+    @Test
+    public void testMinRFGreaterThanDefaultRF()
```
adelapena: This should be named `testMaxRFLesserThanDefaultRF`.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,257 @@
...
+    {
+        try
+        {
+            DatabaseDescriptor.setDefaultKeyspaceRF(1);
+            guardrails().setMaximumReplicationFactorThreshold(MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD);
```
adelapena: A value of `1` wouldn't trigger the failure. However, this test doesn't fail because it only checks the exception contents if it's thrown, but it doesn't verify that the exception is thrown. It should be something like:
```suggestion
                    {
            DatabaseDescriptor.setDefaultKeyspaceRF(MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD + 1);
            guardrails().setMaximumReplicationFactorThreshold(MAXIMUM_REPLICATION_FACTOR_WARN_THRESHOLD, MAXIMUM_REPLICATION_FACTOR_FAIL_THRESHOLD);
            Assert.fail("Should have failed due to being lesser than default_keyspace_rf");
```
We would also need that call to `Assert.fail` in `GuardrailMinimumReplicationFactorTest#testMinRFGreaterThanDefaultRF`.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
adelapena: Seems that the tests are passing after the last commit. We are still missing that test analogous to `GuardrailMinimumReplicationFactorTest#testConfigValidation`, though.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,257 @@
...
+        }
+        catch (IllegalArgumentException e)
+        {
+            String expectedMessage = "";
+
+            if(guardrails().getMaximumReplicationFactorFailThreshold() < DatabaseDescriptor.getDefaultKeyspaceRF())
```
adelapena: No sure why we need to check this here

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
adelapena: We could also simplify and just use `assertThatThrownBy`:
```java
@Test
public void testMaxRFLesserThanDefaultRF()
{
    DatabaseDescriptor.setDefaultKeyspaceRF(3);
    Assertions.assertThatThrownBy(() -> guardrails().setMaximumReplicationFactorThreshold(1, 2))
              .isInstanceOf(IllegalArgumentException.class)
              .hasMessageContaining("maximum_replication_factor_fail_threshold to be set (2) cannot be lesser than default_keyspace_rf (3)");
}
```

src/java/org/apache/cassandra/config/GuardrailsOptions.java
thingtwin1: will fix the MinimumRF methods in a different commit

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
thingtwin1: removed

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
thingtwin1: removed

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -723,6 +749,18 @@ private static void validateMinRFThreshold(int warn, int fail, String name)
         validateMinRFVersusDefaultRF(fail, name);
     }
 
+    private static void validateMaxRFThreshold(int warn, int fail, String name)
```
adelapena: The `name` argument always has the same value, so we can inline it as we do in `validateDataDiskUsageMaxDiskSize`, and as I initially suggested for this method on a previous comment:
```java
private static void validateMaxRFThreshold(int warn, int fail)
{
    validateMaxIntThreshold(warn, fail, "maximum_replication_factor");

    if (fail != -1 && fail < DatabaseDescriptor.getDefaultKeyspaceRF())
        throw new IllegalArgumentException(format("maximum_replication_factor_fail_threshold to be set (%d) " +
                                                  "cannot be lesser than default_keyspace_rf (%d)",
                                                  fail, DatabaseDescriptor.getDefaultKeyspaceRF()));
}
```
The same would happen for max RF:
```java
private static void validateMinRFThreshold(int warn, int fail)
{
    validateMinIntThreshold(warn, fail, "minimum_replication_factor");

    if (fail > DatabaseDescriptor.getDefaultKeyspaceRF())
        throw new IllegalArgumentException(format("minimum_replication_factor_fail_threshold to be set (%d) " +
                                                  "cannot be greater than default_keyspace_rf (%d)",
                                                  fail, DatabaseDescriptor.getDefaultKeyspaceRF()));
}
```

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -723,6 +749,18 @@ private static void validateMinRFThreshold(int warn, int fail, String name)
...
+    {
+        validateMaxIntThreshold(warn, fail, name);
+
+        if (fail != -1 && fail < DatabaseDescriptor.getDefaultKeyspaceRF())
+        {
+            throw new IllegalArgumentException(format("maximum_replication_factor_fail_threshold to be set (%d) " +
+                                                      "cannot be lesser than default_keyspace_rf (%d)",
+                                                      fail, DatabaseDescriptor.getDefaultKeyspaceRF()));
+        }
```
adelapena: This is checked at startup and when we dynamically set a new `maximum_replication_factor_fail_threshold`. However, we are missing an equivalent check for dynamic updates of `default_keyspace_rf`, as it's done for max RF in [`DatabaseDescriptor.setDefaultKeyspaceRF(int)`](https://github.com/apache/cassandra/blob/1b3aec8eef6d1348177dc5a1f0ef30d4ab8b79ee/src/java/org/apache/cassandra/config/DatabaseDescriptor.java#L4087-L4090).

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumReplicationFactorTest.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    @Test
+    public void testMaxRFLesserThanDefaultRF()
+    {
+        DatabaseDescriptor.setDefaultKeyspaceRF(3);
+        Assertions.assertThatThrownBy(() -> guardrails().setMaximumReplicationFactorThreshold(1, 2))
+                  .isInstanceOf(IllegalArgumentException.class)
+                  .hasMessageContaining("maximum_replication_factor_fail_threshold to be set (2) cannot be lesser than default_keyspace_rf (3)");
+    }
```
adelapena: We are missing the reverse case, when an invalid `default_keyspace_rf` is set:
```java
@Test
public void testMaxRFLesserThanDefaultRF()
{
    DatabaseDescriptor.setDefaultKeyspaceRF(3);
    Assertions.assertThatThrownBy(() -> guardrails().setMaximumReplicationFactorThreshold(1, 2))
              .isInstanceOf(IllegalArgumentException.class)
              .hasMessageContaining("maximum_replication_factor_fail_threshold to be set (2) cannot be lesser than default_keyspace_rf (3)");

    DatabaseDescriptor.setDefaultKeyspaceRF(1);
    guardrails().setMaximumReplicationFactorThreshold(1, 2);
    Assertions.assertThatThrownBy(() -> DatabaseDescriptor.setDefaultKeyspaceRF(3))
              .isInstanceOf(IllegalArgumentException.class)
              .hasMessageContaining("default_keyspace_rf to be set (3) cannot be greater than maximum_replication_factor_fail_threshold (2)");
}
```
This validation should be done in `DatabaseDescriptor.setDefaultKeyspaceRF`, but we are missing it.

adelapena (changes requested):

adelapena (approved):
Looks good to me, +1. I have left a few formatting nits on [this commit](https://github.com/adelapena/cassandra/commit/f5ff92ed108c72dc39b373ae87a2ce1104a14842), they can be incorporated during commit if you agree.

thingtwin1:
> Looks good to me, +1. I have left a few formatting nits on [this commit](https://github.com/adelapena/cassandra/commit/f5ff92ed108c72dc39b373ae87a2ce1104a14842), they can be incorporated during commit if you agree.

yes the formatting comments look good to me!

PR #1321, opened 2021-11-15
CASSANDRA-17147 trunk: Guardrails prototype

adelapena:

PR #1583, opened 2022-04-21
CASSANDRA-17212 : Migrate threshold for minimum keyspace replication factor to guardrails

thingtwin1:


src/java/org/apache/cassandra/config/Config.java
```diff
@@ -630,10 +630,9 @@ public static Set<String> splitCommaDelimited(String src)
 
     public volatile boolean diagnostic_events_enabled = false;
 
-    // Default and minimum keyspace replication factors allow validation of newly created keyspaces
+    // Default keyspace replication factors allow validation of newly created keyspaces
     // and good defaults if no replication factor is provided by the user
     public volatile int default_keyspace_rf = 1;
-    public volatile int minimum_keyspace_rf = 0;
```
adelapena, line 642: This property should also be removed from `cassandra.yaml`

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -814,6 +813,8 @@ public static void setClientMode(boolean clientMode)
     public volatile int items_per_collection_fail_threshold = DISABLED_GUARDRAIL;
     public volatile int fields_per_udt_warn_threshold = DISABLED_GUARDRAIL;
     public volatile int fields_per_udt_fail_threshold = DISABLED_GUARDRAIL;
+    public volatile int minimum_keyspace_rf_warn_threshold = DISABLED_GUARDRAIL;
+    public volatile int minimum_keyspace_rf_fail_threshold = DISABLED_GUARDRAIL;
```
adelapena: These new properties should also be added to `cassandra.yaml`

src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
```diff
@@ -229,4 +229,15 @@
      * @return The threshold to fail when creating a UDT with more fields than threshold.
      */
     int getFieldsPerUDTFailThreshold();
+
+    /**
+     * @return The threshold to warn when minimum keyspace replication factor more than threshold.
```
adelapena:
```suggestion
     * @return The threshold to warn when keyspace replication factor is lesser than threshold.
```

src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
```diff
@@ -229,4 +229,15 @@
...
+     */
+    int getMinimumKeyspaceRFWarnThreshold();
+
+    /**
+     * @return The threshold to warn when minimum keyspace replication factor more than threshold.
```
adelapena:
```suggestion
     * @return The threshold to fail when keyspace replication factor is lesser than threshold.
```

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -51,11 +51,11 @@ public final class Guardrails implements GuardrailsMBean
     /**
      * Guardrail on the total number of user keyspaces.
      */
-    public static final Threshold keyspaces =
-    new Threshold("keyspaces",
+    public static final MaxThreshold keyspaces =
+    new MaxThreshold("keyspaces",
```
adelapena, line 59: Nit: The parameters of all the thresholds declarations in this class have been misaligned due to the rename, they should be in the form:
```java
    public static final MaxThreshold keyspaces =
    new MaxThreshold("keyspaces",
                     state -> CONFIG_PROVIDER.getOrCreate(state).getKeyspacesWarnThreshold(),
                     state -> CONFIG_PROVIDER.getOrCreate(state).getKeyspacesFailThreshold(),
                     isWarning ? format("Creating keyspace %s, current number of keyspaces %s exceeds warning threshold of %s.",
                                        what, value, threshold)
                               : format("Cannot have more than %s keyspaces, aborting the creation of keyspace %s",
                                        threshold, what));
```

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -257,16 +257,29 @@ public final class Guardrails implements GuardrailsMBean
...
+     */
+    public static final MinThreshold minimumKeyspaceRF =
+    new MinThreshold("minimum_keyspace_rf",
+                  state -> CONFIG_PROVIDER.getOrCreate(state).getMinimumKeyspaceRFWarnThreshold(),
+                  state -> CONFIG_PROVIDER.getOrCreate(state).getMinimumKeyspaceRFFailThreshold(),
+                  (isWarning, what, value, threshold) ->
+                  isWarning ? format("Keyspaces with %s " + "exceeds warn threshold of %s.",
+                                     what, value, threshold)
+                            : format("Keyspaces with %s " + "exceeds fail threshold of %s.",
+                                     what, value, threshold));
```
adelapena: These calls to `String.format` seem wrong, there are three arguments for only two placeholders, and the `+` seems unneeded.

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -447,4 +447,23 @@
...
      */
     void setFieldsPerUDTThreshold(int warn, int fail);
+
+    /**
+     * @return The threshold to warn when minimum keyspace replication factor more than threshold.
+     */
+    int getMinimumKeyspaceRFWarnThreshold();
+
+    /**
+     * @return The threshold to warn when minimum keyspace replication factor more than threshold.
```
adelapena:
```suggestion
     * @return The threshold to fail when keyspace replication factor is lesser than threshold.
```

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -447,4 +447,23 @@
      * @param fail The threshold to prevent creating a UDT with more fields than threshold. -1 means disabled.
      */
     void setFieldsPerUDTThreshold(int warn, int fail);
+
+    /**
+     * @return The threshold to warn when minimum keyspace replication factor more than threshold.
```
adelapena:
```suggestion
     * @return The threshold to warn when keyspace replication factor is lesser than threshold.
```

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -447,4 +447,23 @@
...
+     */
+    int getMinimumKeyspaceRFFailThreshold();
+
+    /**
+     * @param warn the threshold to warn when the minimum replcation factor for a keyspace is greater than
+     *             threshold -1 means disabled.
+     * @param fail the threshold to fail when the minimum replication factor for a keyspace is greater than
+     *             threshold -1 means disabled.
```
adelapena:
```suggestion
     * @param warn the threshold to warn when the minimum replication factor for a keyspace is lesser than
     *             threshold -1 means disabled.
     * @param fail the threshold to fail when the minimum replication factor for a keyspace is lesser than
     *             threshold -1 means disabled.
```

src/java/org/apache/cassandra/db/guardrails/Threshold.java
```diff
@@ -23,146 +23,79 @@
...
 import org.apache.cassandra.service.ClientState;
 
-/**
- * A guardrail based on numeric threshold(s).
- *
- * <p>A {@link Threshold} guardrail defines (up to) 2 thresholds, one at which a warning is issued, and a higher one
- * at which the operation is aborted with an exception. Only one of those thresholds can be activated if desired.
- *
- * <p>This guardrail only handles guarding positive values.
- */
```
adelapena: Most of this JavaDoc can be preserved with some minor changes, so we don't leave the class undocumented:
```java
/**
 * A guardrail based on numeric threshold(s).
 *
 * <p>A {@link MinThreshold} guardrail defines (up to) 2 thresholds, one at which a warning is issued, and anoone at
 * which the operation is aborted with an exception. Only one of those thresholds can be activated if desired.
 *
 * <p>This guardrail only handles guarding positive values.
 */
```

src/java/org/apache/cassandra/db/guardrails/Threshold.java
```diff
@@ -23,146 +23,79 @@
...
-public class Threshold extends Guardrail
+public abstract class Threshold extends Guardrail
```
adelapena, line 34: All attributes and methods in this class are misaligned, they have have double indentation.

src/java/org/apache/cassandra/db/guardrails/Threshold.java
```diff
@@ -23,146 +23,79 @@
...
-        return errMsg(isWarning, REDACTED, value, thresholdValue);
-    }
-
-    private long failValue(ClientState state)
-    {
-        long failValue = failThreshold.applyAsLong(state);
-        return failValue <= 0 ? Long.MAX_VALUE : failValue;
-    }
+        private final ToLongFunction<ClientState> warnThreshold;
+        private final ToLongFunction<ClientState> failThreshold;
```
adelapena: These two can be protected, so we don't have to define them again in `MinThreshold` and `MaxThreshold`.

src/java/org/apache/cassandra/db/guardrails/Threshold.java
```diff
@@ -23,146 +23,79 @@
...
          */
-        String createMessage(boolean isWarning, String what, long value, long threshold);
-    }
+        interface ErrorMessageProvider
+        {
+            /**
+             * Called when the guardrail is triggered to build the corresponding error message.
+             *
+             * @param isWarning Whether the trigger is a warning one; otherwise it is a failure one.
+             * @param what      A string, provided by the call to the {@link #guard} method, describing what the guardrail
```
adelapena: The reference to `#guard` is broken. I think we should declare the method as abstract in this class, and probably also `triggersOn`, so `MinThreshold` and `MaxThreshold` have to implement these abstract methods. That way we could use references to `Threshold` around the code, without needing to know if it's a min or max threshold.

src/java/org/apache/cassandra/db/guardrails/Threshold.java
```diff
@@ -23,146 +23,79 @@
...
-     * @param value            The value to check.
-     * @param what             A string describing what {@code value} is a value of. This is used in the error message
-     *                         if the guardrail is triggered. For instance, say the guardrail guards the size of column
-     *                         values, then this argument must describe which column of which row is triggering the
-     *                         guardrail for convenience.
-     * @param containsUserData whether the {@code what} contains user data that should be redacted on external systems.
-     * @param state            The client state, used to skip the check if the query is internal or is done by a superuser.
-     *                         A {@code null} value means that the check should be done regardless of the query.
-     */
-    public void guard(long value, String what, boolean containsUserData, @Nullable ClientState state)
```
adelapena: We can leave this as an abstract method:
```java
    /**
     * Apply the guardrail to the provided value, warning or failing if appropriate.
     *
     * @param value            The value to check.
     * @param what             A string describing what {@code value} is a value of. This is used in the error message
     *                         if the guardrail is triggered. For instance, say the guardrail guards the size of column
     *                         values, then this argument must describe which column of which row is triggering the
     *                         guardrail for convenience.
     * @param containsUserData whether the {@code what} contains user data that should be redacted on external systems.
     * @param state            The client state, used to skip the check if the query is internal or is done by a superuser.
     *                         A {@code null} value means that the check should be done regardless of the query.
     */
    public abstract void guard(long value, String what, boolean containsUserData, @Nullable ClientState state);
```

src/java/org/apache/cassandra/db/guardrails/Threshold.java
```diff
@@ -23,146 +23,79 @@
...
-     * argument to {@link #guard} is expensive to build to save doing so in the common case (of the guardrail
-     * not being triggered).
-     *
-     * @param value the value to test.
-     * @param state The client state, used to skip the check if the query is internal or is done by a superuser.
-     *              A {@code null} value means that the check should be done regardless of the query.
-     * @return {@code true} if {@code value} is above the warning or failure thresholds of this guardrail,
-     * {@code false otherwise}.
-     */
-    public boolean triggersOn(long value, @Nullable ClientState state)
```
adelapena: We can leave this as an abstract method:
```java
    /**
     * Checks whether the provided value would trigger a warning or failure if passed to {@link #guard}.
     *
     * <p>This method is optional (does not have to be called) but can be used in the case where the "what"
     * argument to {@link #guard} is expensive to build to save doing so in the common case (of the guardrail
     * not being triggered).
     *
     * @param value the value to test.
     * @param state The client state, used to skip the check if the query is internal or is done by a superuser.
     *              A {@code null} value means that the check should be done regardless of the query.
     * @return {@code true} if {@code value} is above the warning or failure thresholds of this guardrail,
     * {@code false otherwise}.
     */
    public abstract boolean triggersOn(long value, @Nullable ClientState state);
```

src/java/org/apache/cassandra/tools/NodeProbe.java
```diff
@@ -2034,14 +2036,17 @@ public int getDefaultKeyspaceReplicationFactor()
         return ssProxy.getDefaultKeyspaceReplicationFactor();
     }
 
-    public void setMinimumKeyspaceReplicationFactor(int value)
+    public void setMinimumKeyspaceReplicationFactor(int warn, int fail)
```
adelapena: Currently all guardrails can be dynamically changed through direct JMX methods, without offering nodetool setters nor setters. Instead, there are plans to make them accessible through virtual tables, although sadly that will be after 4.1. Even if we wanted to add nodetool setters/getters, we could consider adding some `nodetool setguardrail/getguardrail` common command. Thus, I wonder if we want to keep the `nodetool getdefaultrf/setdefaultrf` commands added by CASSANDRA-14557 or just remove them. @dcapwell wdyt?

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinKeyspaceReplicationFactorTest.java
```diff
@@ -0,0 +1,245 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.db.guardrails;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Set;
+import java.util.function.BiConsumer;
```
adelapena: Nit: unused imports

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinKeyspaceReplicationFactorTest.java
```diff
@@ -0,0 +1,245 @@
...
+import java.util.stream.Collectors;
+
+import org.junit.After;
+import org.junit.Assert;
```
adelapena: Nit: unused import

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinKeyspaceReplicationFactorTest.java
```diff
@@ -0,0 +1,245 @@
...
+               ? Collections.emptyList()
+               : warnings.stream()
+                         .filter(w -> !w.contains("keyspace ks is higher than the number of nodes 1 for datacenter") &&
+                                      !w.contains("When increasing replication factor you need to run a full (-full) repair to distribute the data") &&
+                                      !w.contains("keyspace ks is higher than the number of nodes 1"))
+                         .collect(Collectors.toList());
+    }
+
+    @Test
+    public void testConfigValidation()
```
adelapena: Nit: This method has multiple missed whitespaces after commas and around `+` operators.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinKeyspaceReplicationFactorTest.java
```diff
@@ -0,0 +1,245 @@
...
+    public void testMinKeyspaceRFOnlyWarnBelow() throws Throwable
+    {
+        guardrails().setMinimumKeyspaceRFThreshold(MINIMUM_KEYSPACE_RF_WARN_THRESHOLD, DISABLED_GUARDRAIL);
+        assertWarns("CREATE KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 3}",
+                    format("Keyspaces with %s exceeds warn threshold of %d",
+                           WHAT, MINIMUM_KEYSPACE_RF_WARN_THRESHOLD - 1));
+        assertWarns("ALTER KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 2}",
+                    format("Keyspaces with %s exceeds warn threshold of %d",
+                           WHAT, MINIMUM_KEYSPACE_RF_WARN_THRESHOLD - 2));
+
```
adelapena: Nit: unneeded blank line

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinKeyspaceReplicationFactorTest.java
```diff
@@ -0,0 +1,245 @@
...
+        {
+            String expectedMessage = "";
+
+            if(guardrails().getMinimumKeyspaceRFFailThreshold() > DatabaseDescriptor.getDefaultKeyspaceRF())
+                expectedMessage = format("%s_fail_threshold to be set (%d) cannot be greater than default_keyspace_rf (%d)",
+                                         WHAT, guardrails().getMinimumKeyspaceRFFailThreshold(), DatabaseDescriptor.getDefaultKeyspaceRF());
+
+            Assertions.assertThat(e.getMessage()).contains(expectedMessage);
+        }
+
```
adelapena: Nit: unneeded blank line

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinKeyspaceReplicationFactorTest.java
```diff
@@ -0,0 +1,245 @@
...
+                    format("Keyspaces with %s exceeds fail threshold of %d.",
+                           WHAT, MINIMUM_KEYSPACE_RF_FAIL_THRESHOLD - 1));
+    }
+    @Test
+    public void testMinKeyspaceRFOnlyWarnAbove() throws Throwable
+    {
+        guardrails().setMinimumKeyspaceRFThreshold(MINIMUM_KEYSPACE_RF_WARN_THRESHOLD, DISABLED_GUARDRAIL);
+        assertMinThresholdValid("CREATE KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 6}");
+        assertMinThresholdValid("ALTER KEYSPACE ks WITH replication = { 'class': 'NetworkTopologyStrategy', 'datacenter1': 5}");
+
```
adelapena: Nit: unneeded blank line

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinKeyspaceReplicationFactorTest.java
```diff
@@ -0,0 +1,245 @@
...
+                                      !w.contains("When increasing replication factor you need to run a full (-full) repair to distribute the data") &&
+                                      !w.contains("keyspace ks is higher than the number of nodes 1"))
+                         .collect(Collectors.toList());
+    }
+
+    @Test
+    public void testConfigValidation()
+    {
+        Long two = 2L;
+        Long one = 1L;
```
adelapena: Why can't we use these numbers directly on the last assertion, which is their only use?

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinKeyspaceReplicationFactorTest.java
```diff
@@ -0,0 +1,245 @@
...
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.db.Keyspace;
+import org.apache.cassandra.exceptions.ConfigurationException;
+import org.apache.cassandra.service.ClientWarn;
+import org.assertj.core.api.Assertions;
+
+import static java.lang.String.format;
+import static org.junit.Assert.assertNotNull;
+
+public class GuardrailMinKeyspaceReplicationFactorTest extends ThresholdTester
```
adelapena: I think we are missing some tests for `SimpleStrategy`, and probably also for  `NetworkTopologyStrategy` with multiple DCs. No need to repeat all threshold combinations for those two IMO, probably a simple test for each would be enough.

test/unit/org/apache/cassandra/db/guardrails/ThresholdTester.java
```diff
@@ -118,15 +118,41 @@ protected void testValidationOfThresholdProperties(String warnName, String failN
                   .hasMessageContaining(guardrail.name + "_warn_threshold should be lower than the fail threshold");
     }
 
-    protected void assertThresholdValid(String query) throws Throwable
+    protected void assertMaxThresholdValid(String query) throws Throwable
```
adelapena, line 139: We might make this class and this method abstract, and have new `MinThreshodTester` and `MaxThresholdTester` subclasses implementing the method. Not sure it's worth the fuss, though.

test/unit/org/apache/cassandra/tools/nodetool/GetMinimumKeyspaceRFTest.java
```diff
@@ -1,87 +0,0 @@
...
-import org.junit.Test;
-
-import org.apache.cassandra.config.DatabaseDescriptor;
-import org.apache.cassandra.cql3.CQLTester;
-import org.apache.cassandra.tools.ToolRunner;
-import org.assertj.core.api.Assertions;
-
-import static org.assertj.core.api.Assertions.assertThat;
-
-public class GetMinimumKeyspaceRFTest extends CQLTester
```
adelapena, line 31: I think we should get rid of the nodetool commands for min RF, but this PR still has them. If we were going to keep them we should also keep their tests.

src/java/org/apache/cassandra/config/Config.java
adelapena: I think replication factors are always for keyspaces, so I wonder if this would be better named just `minimum_replication_factor_fail_threshold`, or perhaps `min_replication_factor_fail_threshold`.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -51,11 +51,11 @@ public final class Guardrails implements GuardrailsMBean
     /**
      * Guardrail on the total number of user keyspaces.
      */
-    public static final Threshold keyspaces =
-    new Threshold("keyspaces",
+    public static final MaxThreshold keyspaces =
```
adelapena, line 58: If we add `Threshold#guard` and `Threshold#triggersOn` abstract methods we could keep the type of these instances as `Threshold`, with a `MaxThreshold` or `MinThreshold` constructor, since callers don't strictly need to know if the type of threshold. No particular preference on this, though.

src/java/org/apache/cassandra/cql3/statements/schema/AlterKeyspaceStatement.java
```diff
@@ -99,6 +102,15 @@ public void authorize(ClientState client)
         client.ensureKeyspacePermission(keyspaceName, Permission.ALTER);
     }
 
+    @Override
+    public void validate(ClientState state)
+    {
+        super.validate(state);
```
dcapwell: if you rebase I believe this was done by Josh, so you don't need to store the state

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
dcapwell: I like types as its clear at usage, but agree that 90% of the code should be in a shared base type of Threshold

src/java/org/apache/cassandra/db/guardrails/MinThreshold.java
```diff
@@ -0,0 +1,118 @@
...
+        if (value < failValue)
+        {
+            triggerFail(value, failValue, what, containsUserData, state);
+            return;
+        }
+
+        long warnValue = warnValue(state);
+        if (value < warnValue)
+            triggerWarn(value, warnValue, what, containsUserData);
+    }
```
dcapwell: This is all copy/paste, can we refactor so the operator is pluggable?  Either have an abstract compare method

```
protected boolean compare(long a, long b)
```

or make the comparison an enum like

```
enum Operator {
  GREATER_THAN: { public boolean compare(long a, long b) { return a > b; }}
...
```

src/java/org/apache/cassandra/locator/AbstractReplicationStrategy.java
```diff
@@ -422,10 +434,7 @@ protected void validateReplicationFactor(String s) throws ConfigurationException
         {
             ReplicationFactor rf = ReplicationFactor.fromString(s);
 
-            if (rf.fullReplicas < DatabaseDescriptor.getMinimumKeyspaceRF())
-            {
-                throw new ConfigurationException(String.format("Replication factor cannot be less than minimum_keyspace_rf (%d), found %d", DatabaseDescriptor.getMinimumKeyspaceRF(), rf.fullReplicas));
-            }
+            // mininum replication factor checked in #maybeWarnOnOptions
```
dcapwell: please remove

src/java/org/apache/cassandra/locator/LocalStrategy.java
```diff
@@ -70,7 +71,7 @@ public void validateOptions() throws ConfigurationException
     }
 
     @Override
-    public void maybeWarnOnOptions()
+    public void maybeWarnOnOptions(ClientState state)
```
dcapwell: please revert, you do not have a body so you should take in the default implementations

src/java/org/apache/cassandra/tools/nodetool/GetMinimumKeyspaceRF.java
```diff
@@ -26,8 +26,11 @@
 @Command(name = "getminimumrf", description = "Gets minimum keyspace replication factor.")
 public class GetMinimumKeyspaceRF extends NodeTool.NodeToolCmd
 {
+    @Arguments(title = "<getWarn>", usage = "<getWarn>", description = "warn flag", required = true)
+    private Boolean getWarn = false;
```
dcapwell: use `boolean`

src/java/org/apache/cassandra/tools/nodetool/GetMinimumKeyspaceRF.java
```diff
@@ -26,8 +26,11 @@
 @Command(name = "getminimumrf", description = "Gets minimum keyspace replication factor.")
 public class GetMinimumKeyspaceRF extends NodeTool.NodeToolCmd
 {
+    @Arguments(title = "<getWarn>", usage = "<getWarn>", description = "warn flag", required = true)
```
dcapwell: why is it required when this is optional?  if you added tests you should see this didn't work.

Also, `<getWarn>` doesn't make sense, if you added a test you should see how this impacts the CLI

test/unit/org/apache/cassandra/tools/nodetool/GetMinimumKeyspaceRFTest.java
dcapwell: @adelapena I don't know about getting rid of them, would need to see original JIRA more closely to see the full intent.  I would be concerned getting rid of a new feature when we can port it to the new configs without much prob

src/java/org/apache/cassandra/tools/nodetool/SetMinimumKeyspaceRF.java
```diff
@@ -18,19 +18,31 @@
...
 
+import static java.lang.Boolean.parseBoolean;
+import static java.lang.Integer.parseInt;
+
 @Command(name = "setminimumrf", description = "Sets minimum keyspace replication factor.")
 public class SetMinimumKeyspaceRF extends NodeTool.NodeToolCmd
 {
-    @Arguments(title = "minimum_rf", usage = "<value>", description = "Minimum replication factor", required = true)
-    private Integer minimumRF = null;
+    @Arguments(title = "<setWarn> <value>", usage = "<setWarn> <value>", description = "warn flag and minimum keyspace rf value", required = true)
```
adelapena: In the initial guardrails patch we decided to use a single JMX setter for both warn and fail values (see [here](https://github.com/apache/cassandra/pull/1321#discussion_r760697524)), probably we should do the same here and require both values instead of a boolean.

adelapena:
> @adelapena I don't know about getting rid of them, would need to see original JIRA more closely to see the full intent. I would be concerned getting rid of a new feature when we can port it to the new configs without much prob

I don't see a particular reason in the original ticket for having the nodetool commands beyond the need to be able to dynamically udpdate the properties. Raw JMX already does that, although the nodetool command make it more handy. I don't have a problem keeping those trivial commands, beyond the lack of homogeneity with the other guardrails, which doesn't seem a big problem, even it they become redundant when we add the ability to manipulate guardrails through virtual tables.

src/java/org/apache/cassandra/config/Config.java
dcapwell: I am cool with all 3 names, slight preference for `minimum_replication_factor_fail_threshold`

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -257,16 +257,29 @@ public final class Guardrails implements GuardrailsMBean
...
 
+    /**
+     * Guardrail on the minimum keyspace replication factor.
+     */
+    public static final MinThreshold minimumKeyspaceRF =
+    new MinThreshold("minimum_keyspace_rf",
+                  state -> CONFIG_PROVIDER.getOrCreate(state).getMinimumKeyspaceRFWarnThreshold(),
+                  state -> CONFIG_PROVIDER.getOrCreate(state).getMinimumKeyspaceRFFailThreshold(),
+                  (isWarning, what, value, threshold) ->
+                  isWarning ? format("Keyspaces with %s " + "exceeds warn threshold of %s.",
```
dcapwell: why do you do string concat here?

src/java/org/apache/cassandra/db/guardrails/MaxThreshold.java
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * <p>A {@link MaxThreshold} guardrail defines (up to) 2 thresholds, one at which a warning is issued, and a higher one
+ * at which the operation is aborted with an exception. Only one of those thresholds can be activated if desired.
+ *
+ * <p>This guardrail only handles guarding positive values.
+ */
+public class MaxThreshold extends Threshold
+{
+    private final ToLongFunction<ClientState> warnThreshold;
+    private final ToLongFunction<ClientState> failThreshold;
```
dcapwell: why do you have ref if your parent has it as well?  can you remove?

src/java/org/apache/cassandra/db/guardrails/MinThreshold.java
```diff
@@ -0,0 +1,118 @@
...
+ * at which the operation is aborted with an exception. Only one of those thresholds can be activated if desired.
+ * A {@Link MinThreshold} is defined for minimum guardrails, the value is checked to see if
+ * it is less than the warn and fail thresholds.
+ *
+ * <p>This guardrail only handles guarding positive values.
+ */
+public class MinThreshold extends Threshold
+{
+    private final ToLongFunction<ClientState> warnThreshold;
+    private final ToLongFunction<ClientState> failThreshold;
```
dcapwell: same comment, you reference this twice in the same type

src/java/org/apache/cassandra/tools/NodeProbe.java
dcapwell: > Also, added nodetool commands to override default rf and minimum rf. This is still cumbersome to do it for each node, but yes it serves the purpose of avoiding node restart. We may ideally want to put an endpoint in the [sidecar](https://github.com/apache/cassandra-sidecar), that could potentially change these configurations across the cluster.

from 14557, sounds like its there for simplicity and not sure sidecar ever got the functionality.

I am +0 to removing with a slight preference to keep.

@sumanth-pasupuleti  as the author what is your stance here?  The TL;DR is that your min_rf feature is now min_rf_warn_threshold and min_rf_fail_threshold, which is now awkward in nodetool (doable, just awkward)

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -1239,7 +1239,7 @@ protected com.datastax.driver.core.ResultSet executeViewNet(String query, Object
         return sessionNet().execute(formatViewQuery(query), values);
     }
 
-    protected com.datastax.driver.core.ResultSet executeNet(ProtocolVersion protocolVersion, Statement statement)
+    protected com.datastax.driver.core.ResultSet executeNet(ProtocolVersion protocolVersion, Statement statement) throws Throwable
```
dcapwell: revert plz

test/unit/org/apache/cassandra/cql3/statements/DescribeStatementTest.java
```diff
@@ -51,7 +51,7 @@
 public class DescribeStatementTest extends CQLTester
 {
     @Test
-    public void testSchemaChangeDuringPaging()
+    public void testSchemaChangeDuringPaging() throws Throwable
```
dcapwell: revert plz

conf/cassandra.yaml
```diff
@@ -1688,6 +1683,11 @@ drop_compact_storage_enabled: false
 # Valid values are in [1, max available disk size of all data directories].
 # Defaults to null to disable and use the physically available disk size of data directories during calculations.
 # data_disk_usage_max_disk_size:
+# Guardrail to warn or fail when the minimum keyspace replication factor is greater than threshold.
```
adelapena:
```suggestion
# Guardrail to warn or fail when the keyspace replication factor is lesser than threshold.
```

src/java/org/apache/cassandra/locator/SimpleStrategy.java
```diff
@@ -100,12 +102,13 @@ public void validateOptions() throws ConfigurationException
...
 
     @Override
-    public void maybeWarnOnOptions()
+    public void maybeWarnOnOptions(ClientState state)
     {
         if (!SchemaConstants.isSystemKeyspace(keyspaceName))
         {
             int nodeCount = StorageService.instance.getHostIdToEndpoint().size();
             // nodeCount==0 on many tests
+            Guardrails.minimumKeyspaceRF.guard(rf.fullReplicas, "minimum_keyspace_rf", false, state);
```
adelapena: The `what` argument refers to specific object that is evaluated by the guardrail, not the name of the guardrail. Here we are evaluating the replication factor of a specific keyspace, and `minimum_keyspace_rf` is the name of the guardrail. I would pass the name of the keyspace instead, and use a message provider such as, for example:
```java
isWarning ? format("The keyspace %s has a replication factor of %s, below the warning threshold of %s.",
                   what, value, threshold)
          : format("The keyspace %s has a replication factor of %s, below the failure threshold of %s.",
                   what, value, threshold)
```

src/java/org/apache/cassandra/locator/NetworkTopologyStrategy.java
```diff
@@ -347,6 +349,7 @@ public void maybeWarnOnOptions()
 
                 String dc = e.getKey();
                 ReplicationFactor rf = getReplicationFactor(dc);
+                Guardrails.minimumKeyspaceRF.guard(rf.fullReplicas, "minimum_keyspace_rf", false, state);
```
adelapena: I would pass the name of the keyspace instead of the name of the guardrail.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
adelapena: The parameters remain misaligned after the last changes, they were aligned on column, as in the example above.

src/java/org/apache/cassandra/db/guardrails/Threshold.java
```diff
@@ -21,20 +21,13 @@
 import java.util.function.ToLongFunction;
 import javax.annotation.Nullable;
 
+import io.airlift.airline.Cli;
```
adelapena: Unused import

src/java/org/apache/cassandra/db/guardrails/MinThreshold.java
```diff
@@ -0,0 +1,89 @@
...
+     * @param messageProvider a function to generate the warning or error message if the guardrail is triggered
+     */
+    public MinThreshold(String name,
+                        ToLongFunction<ClientState> warnThreshold,
+                        ToLongFunction<ClientState> failThreshold,
+                        Threshold.ErrorMessageProvider messageProvider)
+    {
+        super(name, warnThreshold, failThreshold, messageProvider);
+        this.warnThreshold = warnThreshold;
+        this.failThreshold = failThreshold;
```
adelapena: Unneeded, this is already done by the super constructor.

src/java/org/apache/cassandra/db/guardrails/MaxThreshold.java
```diff
@@ -0,0 +1,87 @@
...
+     * @param messageProvider a function to generate the warning or error message if the guardrail is triggered
+     */
+    public MaxThreshold(String name,
+                        ToLongFunction<ClientState> warnThreshold,
+                        ToLongFunction<ClientState> failThreshold,
+                        Threshold.ErrorMessageProvider messageProvider)
+    {
+        super(name, warnThreshold, failThreshold, messageProvider);
+        this.warnThreshold = warnThreshold;
+        this.failThreshold = failThreshold;
```
adelapena: Unneeded, this is already done by the super constructor.

src/java/org/apache/cassandra/db/guardrails/MaxThreshold.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.cassandra.service.ClientState;
+
+/**
+ * A guardrail based on numeric threshold(s).
+ *
+ * <p>A {@link MaxThreshold} guardrail defines (up to) 2 thresholds, one at which a warning is issued, and a higher one
+ * at which the operation is aborted with an exception. Only one of those thresholds can be activated if desired.
+ *
+ * <p>This guardrail only handles guarding positive values.
+ */
```
adelapena: As mentioned before, I think we should keep the class JavaDoc on `Threshold`. The class JavaDoc in the min/max subclasses could be an extension of that JavaDoc, something like:
```java
/**
 * {@link Threshold} for minimum guardrails, the value is checked to see if it is lesser than the warn and fail thresholds.
 */
```

test/unit/org/apache/cassandra/db/guardrails/GuardrailsTest.java
```diff
@@ -63,10 +63,10 @@ private void testDisabledThreshold(Threshold guard) throws Throwable
     @Test
     public void testThreshold() throws Throwable
     {
-        Threshold guard = new Threshold("x",
+        MaxThreshold guard = new MaxThreshold("x",
```
adelapena, line 67: We should probably add some test for `MinThreshold`. We could rename `testThreshold` to `testMaxThreshold`, and add a new analogous `testMinThreshold`.

src/java/org/apache/cassandra/config/Config.java
adelapena: I'd prefer either `minimum_replication_factor_fail_threshold` or `min_replication_factor_fail_threshold `. I guess that the prefix choice that we make now would determine how future min/max (or minimum/maximum) properties would look like, since new similar properties should use the same prefix.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -653,6 +691,25 @@ private static void validateWarnLowerThanFail(long warn, long fail, String name)
...
+    {
+        if (warn == -1|| fail == -1)
+            return;
+
+        if (fail > warn)
+            throw new IllegalArgumentException(format("The warn threshold %d for %s_warn_threshold should be greater " +
+                                                      "than the fail threshold %d", warn, name, fail));
+    }
+
+    private static void validateMinRFversesDefaultRF(int fail, String name)
```
dcapwell: capital `V` in `verses`

src/java/org/apache/cassandra/cql3/statements/schema/AlterKeyspaceStatement.java
```diff
@@ -99,6 +102,15 @@ public void authorize(ClientState client)
         client.ensureKeyspacePermission(keyspaceName, Permission.ALTER);
     }
 
+    @Override
+    public void validate(ClientState state)
```
dcapwell: can you remove this method?  its defined to do this in `org.apache.cassandra.cql3.statements.schema.AlterSchemaStatement#validate`

src/java/org/apache/cassandra/cql3/statements/schema/AlterKeyspaceStatement.java
```diff
@@ -53,6 +53,9 @@ public final class AlterKeyspaceStatement extends AlterSchemaStatement
     private final HashSet<String> clientWarnings = new HashSet<>();
 
     private final KeyspaceAttributes attrs;
+
+    private ClientState state;
```
dcapwell: remove, its already accessible

src/java/org/apache/cassandra/cql3/statements/schema/CreateKeyspaceStatement.java
```diff
@@ -53,6 +53,8 @@ public final class CreateKeyspaceStatement extends AlterSchemaStatement
     private final boolean ifNotExists;
     private final HashSet<String> clientWarnings = new HashSet<>();
 
+    private ClientState state;
```
dcapwell: remove, already accessible

src/java/org/apache/cassandra/cql3/statements/schema/CreateKeyspaceStatement.java
```diff
@@ -118,6 +120,9 @@ public void validate(ClientState state)
 
         // Guardrail on number of keyspaces
         Guardrails.keyspaces.guard(Schema.instance.getUserKeyspaces().size() + 1, keyspaceName, false, state);
+
+        // save the query state to use it for guardrails validation in #apply
+        this.state = state;
```
dcapwell: remove, its already done for you

src/java/org/apache/cassandra/db/guardrails/Threshold.java
```diff
@@ -64,24 +65,15 @@ protected String errMsg(boolean isWarning, String what, long value, long thresho
                                              Long.toString(thresholdValue));
     }
 
-    private String redactedErrMsg(boolean isWarning, long value, long thresholdValue)
+    protected String redactedErrMsg(boolean isWarning, long value, long thresholdValue)
```
dcapwell: don't need to change the visibility

src/java/org/apache/cassandra/db/guardrails/Threshold.java
```diff
@@ -136,24 +128,24 @@ public void guard(long value, String what, boolean containsUserData, @Nullable C
...
         }
 
         long warnValue = warnValue(state);
-        if (value > warnValue)
+        if (compare(value, warnValue))
             triggerWarn(value, warnValue, what, containsUserData);
     }
 
-    private void triggerFail(long value, long failValue, String what, boolean containsUserData, ClientState state)
+    protected void triggerFail(long value, long failValue, String what, boolean containsUserData, ClientState state)
```
dcapwell: don't need to change visibility

src/java/org/apache/cassandra/db/guardrails/Threshold.java
```diff
@@ -136,24 +128,24 @@ public void guard(long value, String what, boolean containsUserData, @Nullable C
...
     {
         String fullMessage = errMsg(false, what, value, failValue);
         fail(fullMessage, containsUserData ? redactedErrMsg(false, value, failValue) : fullMessage, state);
     }
 
-    private void triggerWarn(long value, long warnValue, String what, boolean containsUserData)
+    protected void triggerWarn(long value, long warnValue, String what, boolean containsUserData)
```
dcapwell: don't need to change the visibility

src/java/org/apache/cassandra/locator/SimpleStrategy.java
```diff
@@ -100,12 +102,13 @@ public void validateOptions() throws ConfigurationException
...
 
     @Override
-    public void maybeWarnOnOptions()
+    public void maybeWarnOnOptions(ClientState state)
     {
         if (!SchemaConstants.isSystemKeyspace(keyspaceName))
         {
             int nodeCount = StorageService.instance.getHostIdToEndpoint().size();
             // nodeCount==0 on many tests
+            Guardrails.minimumRF.guard(rf.fullReplicas, keyspaceName, false, state);
```
dcapwell: your error messages says datacenter but you are providing a keyspace, can you fix?

src/java/org/apache/cassandra/tools/NodeTool.java
```diff
@@ -102,7 +102,6 @@ public int execute(String... args)
                 Compact.class,
                 CompactionHistory.class,
                 CompactionStats.class,
-                DataPaths.class,
```
dcapwell: please add back, its unrelated to Min RF

src/java/org/apache/cassandra/tools/NodeTool.java
```diff
@@ -144,7 +143,6 @@ public int execute(String... args)
                 GetInterDCStreamThroughput.class,
                 GetLoggingLevels.class,
                 GetMaxHintWindow.class,
-                GetMinimumKeyspaceRF.class,
```
dcapwell, line 147: Please update `NEWS.txt` removing reference to this since you are removing

test/unit/org/apache/cassandra/db/guardrails/GuardrailsTest.java
```diff
@@ -40,13 +40,13 @@ public class GuardrailsTest extends GuardrailTester
...
-    public void testDisabledThreshold() throws Throwable
+    public void testDisabledMaxThreshold() throws Throwable
     {
         Threshold.ErrorMessageProvider errorMessageProvider = (isWarn, what, v, t) -> "Should never trigger";
-        testDisabledThreshold(new Threshold("x", state -> DISABLED, state -> DISABLED, errorMessageProvider));
+        testDisabledMaxThreshold(new MaxThreshold("x", state -> DISABLED, state -> DISABLED, errorMessageProvider));
     }
 
-    private void testDisabledThreshold(Threshold guard) throws Throwable
+    private void testDisabledMaxThreshold(MaxThreshold guard) throws Throwable
```
adelapena: We can leave `testDisabledThreshold(Threshold)` as it was and just call it from `testDisabledMaxThreshold()` and `testDisabledMinThreshold` with a `MaxThreshold` and a `MinThreshold` instance.

conf/cassandra.yaml
```diff
@@ -1688,6 +1683,11 @@ drop_compact_storage_enabled: false
 # Valid values are in [1, max available disk size of all data directories].
 # Defaults to null to disable and use the physically available disk size of data directories during calculations.
 # data_disk_usage_max_disk_size:
+# Guardrail to warn or fail when the minimum replication factor is lesser than threshold.
+# This would also apply to system keyspaces.
+# Suggested value for use in production: 2 or higher
+# minimum_rf_warn_threshold: -1
+# minimum_rf_fail_threshold: -1
```
adelapena: Didn't we say that we preferred `minimum_replication_factor_fail_threshold`? That would also apply to the getters/setters around, that should be:
* `getMinimumReplicationFactorWarnThreshold`
* `getMinimumReplicationFactorFailThreshold`
* `getMinimumReplicationFactor`

test/unit/org/apache/cassandra/config/DatabaseDescriptorRefTest.java
```diff
@@ -105,7 +105,7 @@
     "org.apache.cassandra.config.GuardrailsOptions$ConsistencyLevels",
     "org.apache.cassandra.config.GuardrailsOptions$IntThreshold",
     "org.apache.cassandra.config.GuardrailsOptions$TableProperties",
-    "org.apache.cassandra.config.GuardrailsOptions$Threshold",
+    "org.apache.cassandra.config.GuardrailsOptions$MaxThreshold",
```
adelapena: Actually, we don't need any of the threshold classes here, not for guardrails nor for track warnings. We forgot to remove them during CASSANDRA-17353 and CASSANDRA-17560.

src/java/org/apache/cassandra/tools/NodeTool.java
adelapena: I think no entry was added to `NEWS.txt`. We should however mention the new guardrail in that file, adding it to [this list](https://github.com/apache/cassandra/blob/0965710345b1e9887d6bbd53b3df45278f8743f4/NEWS.txt#L68-L87).

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -55,72 +55,72 @@ public final class Guardrails implements GuardrailsMBean
...
+                                        threshold, what));
 
     /**
      * Guardrail disabling user's ability to create secondary indexes
      */
     public static final DisableFlag createSecondaryIndexesEnabled =
     new DisableFlag("secondary_indexes",
                     state -> !CONFIG_PROVIDER.getOrCreate(state).getSecondaryIndexesEnabled(),
-                    "User creation of secondary indexes");
+              "User creation of secondary indexes");
```
adelapena: Misaligned parameter.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -315,6 +315,19 @@ public final class Guardrails implements GuardrailsMBean
         replicaDiskUsage.minNotifyIntervalInMs(minNotifyInterval);
     }
 
+    /**
+     * Guardrail on the minimum replication factor.
+     */
+    public static final MinThreshold minimumRF =
```
adelapena: Maybe this could be named `minimumReplicationFactor`, in line with the property names.

src/java/org/apache/cassandra/cql3/statements/schema/AlterKeyspaceStatement.java
```diff
@@ -53,6 +53,7 @@ public final class AlterKeyspaceStatement extends AlterSchemaStatement
     private final HashSet<String> clientWarnings = new HashSet<>();
 
     private final KeyspaceAttributes attrs;
+
```
adelapena: Unneeded blank line, I think.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -653,6 +691,25 @@ private static void validateWarnLowerThanFail(long warn, long fail, String name)
...
+    {
+        if (warn == -1|| fail == -1)
+            return;
+
+        if (fail > warn)
+            throw new IllegalArgumentException(format("The warn threshold %d for %s_warn_threshold should be greater " +
+                                                      "than the fail threshold %d", warn, name, fail));
+    }
+
+    private static void validateMinRFVersesDefaultRF(int fail, String name)
```
adelapena: Nit: I think its "Versus", instead of "Verses"

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -653,6 +691,25 @@ private static void validateWarnLowerThanFail(long warn, long fail, String name)
...
+    {
+        if (fail > DatabaseDescriptor.getDefaultKeyspaceRF())
+        {
+            throw new ConfigurationException(String.format("%s_fail_threshold to be set (%d) cannot be greater than default_keyspace_rf (%d)",
```
adelapena: We should throw `IllegalArgumentException` or some other generic Java exception, so `ConfigurationException` doesn't leak to JMX. At startup, [`DatabaseDescriptor.applyGuardrails`](https://github.com/apache/cassandra/blob/0965710345b1e9887d6bbd53b3df45278f8743f4/src/java/org/apache/cassandra/config/DatabaseDescriptor.java#L931-L938) would catch that `IllegalArgumentException` and throw a `ConfigurationException` instead.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -595,6 +596,31 @@ public void setDataDiskUsageMaxDiskSize(@Nullable DataStorageSpec diskSize)
...
+        validateMinRFThreshold(warn, fail, "minimum_rf");
+        updatePropertyWithLogging("minimum_rf_warn_threshold",
+                                  warn,
+                                  () -> config.minimum_rf_warn_threshold,
+                                  x -> config.minimum_rf_warn_threshold = x);
+        updatePropertyWithLogging("minimum_rf_fail_threshold",
+                                  fail,
+                                  () -> config.minimum_rf_fail_threshold,
+                                  x -> config.minimum_rf_fail_threshold = x);
+    }
```
adelapena, line 621: Nit: missed blank line after method declaration

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -629,20 +655,32 @@ private static void validatePercentage(long value, String name)
...
+        validatePositiveNumeric(warn, Integer.MAX_VALUE, name + "_warn_threshold");
+        validatePositiveNumeric(fail, Integer.MAX_VALUE, name + "_fail_threshold");
+        validateWarnGreaterThanFail(warn, fail, name);
     }
 
+    private static void validateMinRFThreshold(int warn, int fail, String name)
+    {
+        validateMinIntThreshold(warn, fail, name);
+        validateMinRFVersesDefaultRF(fail, name);
+    }
```
adelapena, line 682: Nit: missed blank line after method declaration

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -653,6 +691,25 @@ private static void validateWarnLowerThanFail(long warn, long fail, String name)
                                                       "than the fail threshold %d", warn, name, fail));
     }
 
+    private static void validateWarnGreaterThanFail(long warn, long fail, String name)
+    {
+        if (warn == -1|| fail == -1)
```
adelapena: Nit: missed white space:
```suggestion
        if (warn == -1 || fail == -1)
```

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinimumReplicationFactorTest.java
```diff
@@ -0,0 +1,272 @@
...
+
+    @Test
+    public void testConfigValidation()
+    {
+        assertNotNull(guardrail);
+        setter.accept(guardrails(), DISABLED_GUARDRAIL, DISABLED_GUARDRAIL);
+
+        assertInvalidPositiveProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), Integer.MIN_VALUE, Integer.MAX_VALUE, WHAT + "_fail_threshold");
+        assertInvalidPositiveProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), -2, Integer.MAX_VALUE, WHAT + "_fail_threshold");
+        assertValidProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), DISABLED_GUARDRAIL);
```
adelapena:
```suggestion
        assertValidProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a), DISABLED_GUARDRAIL);
```

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinimumReplicationFactorTest.java
```diff
@@ -0,0 +1,272 @@
...
+        assertInvalidPositiveProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), DISABLED_GUARDRAIL == 0 ? -1 : 0, Integer.MAX_VALUE, WHAT + "_fail_threshold");
+        assertValidProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), 1L);
+        assertValidProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), 2L);
+
+        assertInvalidPositiveProperty((g, w) -> setter.accept(g, w.intValue(), DISABLED_GUARDRAIL), Integer.MIN_VALUE, Integer.MAX_VALUE, WHAT + "_warn_threshold");
+        assertInvalidPositiveProperty((g, w) -> setter.accept(g, w.intValue(), DISABLED_GUARDRAIL), -2, Integer.MAX_VALUE, WHAT + "_warn_threshold");
+        assertValidProperty((g,w) ->setter.accept(g,w.intValue(),DISABLED_GUARDRAIL), DISABLED_GUARDRAIL);
```
adelapena:
```suggestion
        assertValidProperty((g, w) ->setter.accept(g, w, DISABLED_GUARDRAIL), DISABLED_GUARDRAIL);
```

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinimumReplicationFactorTest.java
```diff
@@ -0,0 +1,272 @@
...
+        assertInvalidPositiveProperty((g, w) -> setter.accept(g, w.intValue(), DISABLED_GUARDRAIL), DISABLED_GUARDRAIL == 0 ? -1 : 0, Integer.MAX_VALUE, WHAT + "_warn_threshold");
```
adelapena: `DISABLED_GUARDRAIL == 0` is always false

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinimumReplicationFactorTest.java
```diff
@@ -0,0 +1,272 @@
...
+    @Test
+    public void testConfigValidation()
+    {
+        assertNotNull(guardrail);
+        setter.accept(guardrails(), DISABLED_GUARDRAIL, DISABLED_GUARDRAIL);
+
+        assertInvalidPositiveProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), Integer.MIN_VALUE, Integer.MAX_VALUE, WHAT + "_fail_threshold");
+        assertInvalidPositiveProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), -2, Integer.MAX_VALUE, WHAT + "_fail_threshold");
+        assertValidProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), DISABLED_GUARDRAIL);
+        assertInvalidPositiveProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), DISABLED_GUARDRAIL == 0 ? -1 : 0, Integer.MAX_VALUE, WHAT + "_fail_threshold");
```
adelapena: `DISABLED_GUARDRAIL == 0` is always false

src/java/org/apache/cassandra/db/guardrails/Threshold.java
```diff
@@ -56,6 +55,8 @@ public Threshold(String name,
         this.messageProvider = messageProvider;
     }
 
+    protected abstract boolean compare(long value, long threshold);
```
adelapena, line 59: Can we add `@Override` annotation for `compare`, `failValue` and `warnValue`?

src/java/org/apache/cassandra/db/guardrails/MinThreshold.java
```diff
@@ -0,0 +1,62 @@
...
+     */
+    public MinThreshold(String name,
+                        ToLongFunction<ClientState> warnThreshold,
+                        ToLongFunction<ClientState> failThreshold,
+                        Threshold.ErrorMessageProvider messageProvider)
+    {
+        super(name, warnThreshold, failThreshold, messageProvider);
+    }
+
+    protected boolean compare(long value, long threshold)
```
adelapena, line 46: Can we add `@Override` annotations for `compare`, `failValue` and `warnValue`?

src/java/org/apache/cassandra/config/GuardrailsOptions.java
dcapwell: For context on that comment: https://issues.apache.org/jira/browse/CASSANDRA-17580

src/java/org/apache/cassandra/db/guardrails/Threshold.java
dcapwell: to be clear, the `@Override` isn't in this class, its in `MinThreshold` and `MaxThreshold`

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -868,10 +868,10 @@ else if (conf.max_value_size.toMebibytes() >= 2048)
 
         validateMaxConcurrentAutoUpgradeTasksConf(conf.max_concurrent_automatic_sstable_upgrades);
 
-        if (conf.default_keyspace_rf < conf.minimum_keyspace_rf)
+        if (conf.default_keyspace_rf < conf.minimum_replication_factor_fail_threshold)
         {
-            throw new ConfigurationException(String.format("default_keyspace_rf (%d) cannot be less than minimum_keyspace_rf (%d)",
-                                                           conf.default_keyspace_rf, conf.minimum_keyspace_rf));
+            throw new ConfigurationException(String.format("default_keyspace_rf (%d) cannot be less than minimum_rf_fail_threshold (%d)",
```
dcapwell: fix comment, no longer `rf`

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -4081,30 +4081,14 @@ public static void setDefaultKeyspaceRF(int value) throws ConfigurationException
             throw new ConfigurationException("default_keyspace_rf cannot be less than 1");
         }
 
-        if (value < getMinimumKeyspaceRF())
+        if (value < guardrails.getMinimumReplicationFactorFailThreshold())
         {
-            throw new ConfigurationException(String.format("default_keyspace_rf to be set (%d) cannot be less than minimum_keyspace_rf (%d)", value, getMinimumKeyspaceRF()));
+            throw new ConfigurationException(String.format("default_keyspace_rf to be set (%d) cannot be less than minimum_rf_fail_threshold (%d)", value, guardrails.getMinimumReplicationFactorFailThreshold()));
```
dcapwell: fix comment, no longer `rf`

src/java/org/apache/cassandra/cql3/statements/schema/AlterKeyspaceStatement.java
```diff
@@ -51,7 +51,6 @@ public final class AlterKeyspaceStatement extends AlterSchemaStatement
     private static final boolean allow_alter_rf_during_range_movement = Boolean.getBoolean(Config.PROPERTY_PREFIX + "allow_alter_rf_during_range_movement");
     private static final boolean allow_unsafe_transient_changes = Boolean.getBoolean(Config.PROPERTY_PREFIX + "allow_unsafe_transient_changes");
     private final HashSet<String> clientWarnings = new HashSet<>();
-
```
dcapwell: can you add the space back; patch is large, so best to avoid style changes

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinimumReplicationFactorTest.java
```diff
@@ -0,0 +1,277 @@
...
+
+    @Test
+    public void testConfigValidation()
+    {
+        assertNotNull(guardrail);
+        setter.accept(guardrails(), DISABLED_GUARDRAIL, DISABLED_GUARDRAIL);
+
+        assertInvalidPositiveProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), Integer.MIN_VALUE, Integer.MAX_VALUE, WHAT + "_fail_threshold");
+        assertInvalidPositiveProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), -2, Integer.MAX_VALUE, WHAT + "_fail_threshold");
+        assertValidProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), DISABLED_GUARDRAIL);
```
adelapena: Nit: we don't need the conversion to int
```suggestion
        assertValidProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a), DISABLED_GUARDRAIL);
```

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinimumReplicationFactorTest.java
```diff
@@ -0,0 +1,277 @@
...
+        assertInvalidPositiveProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), 0, Integer.MAX_VALUE, WHAT + "_fail_threshold");
+        assertValidProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), 1L);
+        assertValidProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), 2L);
+
+        assertInvalidPositiveProperty((g, w) -> setter.accept(g, w.intValue(), DISABLED_GUARDRAIL), Integer.MIN_VALUE, Integer.MAX_VALUE, WHAT + "_warn_threshold");
+        assertInvalidPositiveProperty((g, w) -> setter.accept(g, w.intValue(), DISABLED_GUARDRAIL), -2, Integer.MAX_VALUE, WHAT + "_warn_threshold");
+        assertValidProperty((g,w) ->setter.accept(g,w.intValue(),DISABLED_GUARDRAIL), DISABLED_GUARDRAIL);
```
adelapena: Nit: we don't need the conversion to int, missed spaces
```suggestion
        assertValidProperty((g, w) ->setter.accept(g, w, DISABLED_GUARDRAIL), DISABLED_GUARDRAIL);
```

src/java/org/apache/cassandra/cql3/statements/schema/AlterKeyspaceStatement.java
adelapena: Nit: unneeded change

adelapena (approved):
Looks good to me, +1. I'm leaving a couple of comments about nits that were missed in the last commit. I've started a new CI run over the rebased patch:
* https://app.circleci.com/pipelines/github/adelapena/cassandra/1513/workflows/61df50c2-96e2-45a6-8310-89f3d4393695
* https://app.circleci.com/pipelines/github/adelapena/cassandra/1513/workflows/217a824a-9cfc-414a-bbcc-53821e63d757

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -34,6 +34,7 @@
 import org.apache.cassandra.db.guardrails.GuardrailsConfig;
 import org.apache.cassandra.io.util.FileUtils;
 import org.apache.cassandra.service.disk.usage.DiskUsageMonitor;
+import org.apache.cassandra.exceptions.ConfigurationException;
```
dcapwell: unused import

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinimumReplicationFactorTest.java
```diff
@@ -0,0 +1,277 @@
...
+        {
+            String expectedMessage = "";
+
+            if(guardrails().getMinimumReplicationFactorFailThreshold() > DatabaseDescriptor.getDefaultKeyspaceRF())
+                expectedMessage = format("%s_fail_threshold to be set (%d) cannot be greater than default_keyspace_rf (%d)",
+                                         WHAT, guardrails().getMinimumReplicationFactorFailThreshold(), DatabaseDescriptor.getDefaultKeyspaceRF());
+            Assertions.assertThat(e.getMessage()).contains(expectedMessage);
+        }
+    }
+}
```
dcapwell: missing newline, plz add one

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinimumReplicationFactorTest.java
```diff
@@ -0,0 +1,277 @@
...
+    {
+        assertNotNull(guardrail);
+        setter.accept(guardrails(), DISABLED_GUARDRAIL, DISABLED_GUARDRAIL);
+
+        assertInvalidPositiveProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), Integer.MIN_VALUE, Integer.MAX_VALUE, WHAT + "_fail_threshold");
+        assertInvalidPositiveProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), -2, Integer.MAX_VALUE, WHAT + "_fail_threshold");
+        assertValidProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), DISABLED_GUARDRAIL);
+        assertInvalidPositiveProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), 0, Integer.MAX_VALUE, WHAT + "_fail_threshold");
+        assertValidProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), 1L);
+        assertValidProperty((g, a) -> setter.accept(g, DISABLED_GUARDRAIL, a.intValue()), 2L);
```
dcapwell: why are you setting `long` when the type is `int`?  Can you switch this to `1` and `2`?

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinimumReplicationFactorTest.java
```diff
@@ -0,0 +1,277 @@
...
+
+        assertInvalidPositiveProperty((g, w) -> setter.accept(g, w.intValue(), DISABLED_GUARDRAIL), Integer.MIN_VALUE, Integer.MAX_VALUE, WHAT + "_warn_threshold");
+        assertInvalidPositiveProperty((g, w) -> setter.accept(g, w.intValue(), DISABLED_GUARDRAIL), -2, Integer.MAX_VALUE, WHAT + "_warn_threshold");
+        assertValidProperty((g, w) -> setter.accept(g, w.intValue(), DISABLED_GUARDRAIL), DISABLED_GUARDRAIL);
+        assertInvalidPositiveProperty((g, w) -> setter.accept(g, w.intValue(), DISABLED_GUARDRAIL), 0, Integer.MAX_VALUE, WHAT + "_warn_threshold");
+        assertValidProperty((g, w) -> setter.accept(g, w.intValue(), DISABLED_GUARDRAIL), 1L);
```
dcapwell: why are you setting long when the type is int? Can you switch this to 1 and 2?

dcapwell (approved):

PR #1584, opened 2022-04-23
CASSANDRA-17425: Add new CQL function maxWritetime

yifan-c:


src/java/org/apache/cassandra/cql3/selection/Selectable.java
```diff
@@ -245,18 +266,20 @@ public Selector.Factory newSelectorFactory(TableMetadata table,
             if (column.isPrimaryKeyColumn())
                 throw new InvalidRequestException(
                         String.format("Cannot use selection function %s on PRIMARY KEY part %s",
-                                      isWritetime ? "writeTime" : "ttl",
+                                      kind.name,
                                       column.name));
-            if (column.type.isCollection())
+
+            // maxwritetime is allowed for collection
+            if (kind != Kind.MAX_WRITE_TIME && column.type.isCollection())
```
frankgh: we could potentially move this to the enum? `if (column.type.isCollection() && !kind.allowedForCollection()) then throw...`. If we ever add additional `Kind`s, we wouldn't need to touch this logic, so it becomes a little more maintainable

src/java/org/apache/cassandra/cql3/selection/ResultSetBuilder.java
```diff
@@ -98,6 +102,27 @@ public void add(ByteBuffer v)
...
+
+        long timestamp = -1L;
+        if (selectors.collectMaxTimestamps())
+        {
+            Iterator<Cell<?>> cells = complexColumnData.iterator();
+            while (cells.hasNext())
+            {
+                timestamp = Math.max(timestamp, cells.next().timestamp());
+            }
+        }
```
frankgh, line 121: We could use streams here, although we might need to worry about performance 
```suggestion
        long timestamp = -1L;
        if (selectors.collectMaxTimestamps())
        {
            timestamp = StreamSupport.stream(complexColumnData.spliterator(), false)
                                     .mapToLong(Cell::timestamp)
                                     .max()
                                     .orElse(-1L);
        }
```

test/unit/org/apache/cassandra/cql3/selection/SelectorSerializationTest.java
```diff
@@ -60,8 +60,8 @@ public void testSerDes() throws IOException
         checkSerialization(table.getColumn(new ColumnIdentifier("c1", false)), table);
 
         // Test WritetimeOrTTLSelector serialization
-        checkSerialization(new Selectable.WritetimeOrTTL(table.getColumn(new ColumnIdentifier("v", false)), true), table);
-        checkSerialization(new Selectable.WritetimeOrTTL(table.getColumn(new ColumnIdentifier("v", false)), false), table);
+        checkSerialization(new Selectable.WritetimeOrTTL(table.getColumn(new ColumnIdentifier("v", false)), Selectable.WritetimeOrTTL.Kind.WRITE_TIME), table);
+        checkSerialization(new Selectable.WritetimeOrTTL(table.getColumn(new ColumnIdentifier("v", false)), Selectable.WritetimeOrTTL.Kind.TTL), table);
```
frankgh, line 64: Can we add the `MAX_WRITE_TIME` kind here as well?
```suggestion
        checkSerialization(new Selectable.WritetimeOrTTL(table.getColumn(new ColumnIdentifier("v", false)), Selectable.WritetimeOrTTL.Kind.TTL), table);
        checkSerialization(new Selectable.WritetimeOrTTL(table.getColumn(new ColumnIdentifier("v", false)), Selectable.WritetimeOrTTL.Kind.MAX_WRITE_TIME), table);
```

src/java/org/apache/cassandra/cql3/selection/WritetimeOrTTLSelector.java
```diff
@@ -169,28 +176,28 @@ public boolean equals(Object o)
...
+                + TypeSizes.sizeofUnsignedVInt(kind.ordinal());
     }
 
     @Override
     protected void serialize(DataOutputPlus out, int version) throws IOException
     {
         ByteBufferUtil.writeWithVIntLength(column.name.bytes, out);
         out.writeInt(idx);
-        out.writeBoolean(isWritetime);
+        out.writeUnsignedVInt(kind.ordinal());
```
frankgh: we could make this serialization more compact. I don't think we'll have more that 256 enum types 
```suggestion
        out.writeByte(kind.ordinal());
```

frankgh (commented):
Looks good to me. I've left a few small comments.

src/java/org/apache/cassandra/cql3/selection/ResultSetBuilder.java
yifan-c: Only 1 minus line. Not very motivative.

src/java/org/apache/cassandra/cql3/selection/WritetimeOrTTLSelector.java
yifan-c: yep. In this case, `VInt` writes only 1 byte. I think it makes sense to explicitly use writeByte so to avoid checking value size.

frankgh (approved):
+1 Looks good to me

frankgh (approved):
+1 Thanks for addressing the comments.

src/java/org/apache/cassandra/cql3/selection/Selectable.java
```diff
@@ -245,18 +271,20 @@ public Selector.Factory newSelectorFactory(TableMetadata table,
             if (column.isPrimaryKeyColumn())
                 throw new InvalidRequestException(
                         String.format("Cannot use selection function %s on PRIMARY KEY part %s",
-                                      isWritetime ? "writeTime" : "ttl",
+                                      kind.name,
                                       column.name));
-            if (column.type.isCollection())
+
+            // maxwritetime is allowed for collection
```
adelapena:
```suggestion
            // only maxwritetime is allowed for collections
```

src/java/org/apache/cassandra/cql3/selection/Selection.java
```diff
@@ -376,6 +376,12 @@ private static List<ByteBuffer> rowToJson(List<ByteBuffer> row,
          */
         public boolean collectTimestamps();
 
+        /**
+         * Checks if one of the selectors collect maxTimestamps.
```
adelapena:
```suggestion
         * Checks if one of the selectors collects maxTimestamps.
```

src/java/org/apache/cassandra/cql3/selection/SelectorFactories.java
```diff
@@ -46,6 +46,11 @@ final class SelectorFactories implements Iterable<Selector.Factory>
      */
     private boolean containsWritetimeFactory;
 
+    /**
+     * <code>true</code> if one of the factory creates maxWritetime selectors.
```
adelapena:
```suggestion
     * {@code true} if one of the factories creates maxWritetime selectors.
```

src/java/org/apache/cassandra/cql3/selection/SelectorFactories.java
```diff
@@ -165,6 +171,17 @@ public boolean containsWritetimeSelectorFactory()
         return containsWritetimeFactory;
     }
 
+    /**
+     * Checks if this <code>SelectorFactories</code> contains at least one factory for maxWritetime selectors.
```
adelapena:
```suggestion
     * Checks if this {@link SelectorFactories} contains at least one factory for maxWritetime selectors
```

src/java/org/apache/cassandra/cql3/selection/SelectorFactories.java
```diff
@@ -165,6 +171,17 @@ public boolean containsWritetimeSelectorFactory()
...
+     *
+     * @return <code>true</code> if this <code>SelectorFactories</code> contains at least one factory for maxWritetime
+     * selectors, <code>false</code> otherwise.
```
adelapena:
```suggestion
     * <code>true</code> if this <code>SelectorFactories</code> contains at least one factory for maxWritetime
     * selectors, <code>false</code> otherwise.
     * @return {@link true} if this {@link SelectorFactories} contains at least one factory for maxWritetime
     * selectors, {@link false} otherwise.
```

test/unit/org/apache/cassandra/cql3/validation/entities/TimestampTest.java
```diff
@@ -97,6 +97,42 @@ public void testTimestampTTL() throws Throwable
...
+        String myType = createType("CREATE TYPE %s (a int, b int)");
+        createTable("CREATE TABLE %s (k int PRIMARY KEY, a text, " +
+                    "l list<int>, fl frozen<list<int>>," +
+                    "s set<int>, fs frozen<set<int>>," +
+                    "m map<int, text>, fm frozen<map<int, text>>," +
+                    "t " + myType + ", ft frozen<" + myType + ">)");
+        execute("INSERT INTO %s (k, a, l, fl, s, fs, m, fm, t, ft) VALUES " +
+                "(1, 'test', [1], [2], {1}, {2}, {1 : 'a'}, {2 : 'b'}, {a : 1, b : 1 }, {a : 2, b : 2}) USING TIMESTAMP 1");
+        // update the list, so its maxWritetime should be more advanced than other columns
+        execute("UPDATE %s USING TIMESTAMP 2 SET l = l + [10] WHERE k = 1");
```
adelapena: This only exercises the selection among multiple values for the list. It could be useful to test all the multicell types, probably adding a separate update and timestamp per type.

test/unit/org/apache/cassandra/cql3/validation/entities/TimestampTest.java
```diff
@@ -97,6 +97,42 @@ public void testTimestampTTL() throws Throwable
...
+    @Test
+    public void testMaxTimestamp() throws Throwable
+    {
+        String myType = createType("CREATE TYPE %s (a int, b int)");
+        createTable("CREATE TABLE %s (k int PRIMARY KEY, a text, " +
+                    "l list<int>, fl frozen<list<int>>," +
+                    "s set<int>, fs frozen<set<int>>," +
+                    "m map<int, text>, fm frozen<map<int, text>>," +
+                    "t " + myType + ", ft frozen<" + myType + ">)");
+        execute("INSERT INTO %s (k, a, l, fl, s, fs, m, fm, t, ft) VALUES " +
```
adelapena: Before writing any data on the collections, we could insert a row with only the primary key and then verify that `maxwritetime` returns `null` for all the types.

src/java/org/apache/cassandra/cql3/selection/Selectable.java
```diff
@@ -245,18 +271,20 @@ public Selector.Factory newSelectorFactory(TableMetadata table,
...
+            if (column.type.isCollection() && !kind.allowedForMultiCell())
```
adelapena, line 279: Not sure why this error was originally thrown for frozen collections, nor why it wasn't thrown for non-frozen, multi-cell UDTs. Maybe it should be `column.type.isMultiCell()` instead of `column.type.isCollection()`?

src/java/org/apache/cassandra/cql3/selection/WritetimeOrTTLSelector.java
```diff
@@ -45,29 +43,30 @@ protected Selector deserialize(DataInputPlus in, int version, TableMetadata meta
             ByteBuffer columnName = ByteBufferUtil.readWithVIntLength(in);
             ColumnMetadata column = metadata.getColumn(columnName);
             int idx = in.readInt();
-            boolean isWritetime = in.readBoolean();
-            return new WritetimeOrTTLSelector(column, idx, isWritetime);
+            int ordinal = in.readByte();
```
adelapena, line 46: Don't know if this would require a protocol bump, since I think 4.1 is going to be released with the current `v6-beta`.

src/java/org/apache/cassandra/cql3/selection/SelectorFactories.java
yifan-c: `<code>true</code>` was used to be consistent with others. There are many such occurrences in this file.  I can change the nearby ones.

src/java/org/apache/cassandra/cql3/selection/Selectable.java
yifan-c: Yes and nice catch. It should use `isMultiCell` instead of `isCollection`. Otherwise, `writetime` function on non-frozen UDT is allowed. I was planning to fix it later, since it is relatively unrelated with the patch.

src/java/org/apache/cassandra/cql3/selection/WritetimeOrTTLSelector.java
yifan-c: Good question! I think this patch is going to be merged with trunk instead of 4.1. 

Luckily, both `readBoolean` and `readByte` consume one byte. When the cluster runs in mix mode, it should be able to ser/deser w/o throwing exceptions. 
The prior logic is to use `ttl` when reads 0 and use `writetime` when reads 1. I need to adjust the `Kind` enum ordering to reflect that, i.e. `TTL: 0, WRITE_TIME: 1, MAX_WRITE_TIME: 2`. Thanks for pointing it out.
When the cluster is upgrading, we should not expect client to start sending queries that use the new function. So the value of the byte field should be either 0 or 1, which can be interpreted correctly on both old and new versions. 

Does it sound good to you?

adelapena: Sounds good to me. Indeed it's fortunate that `readBoolean` also reads a byte. Probably we should add a note in the upgrade notes in `NEWS.txt` warning about `MAX_WRITE_TIME` not working on rolling upgrade scenarios. By the way, we should also add an entry about the new feature in the section about new features of that file.

src/java/org/apache/cassandra/cql3/selection/Selectable.java
```diff
@@ -224,9 +224,9 @@ public static class WritetimeOrTTL implements Selectable
     {
         public enum Kind
         {
+            TTL("ttl", Int32Type.instance),
             WRITE_TIME("writetime", LongType.instance),
-            MAX_WRITE_TIME("maxwritetime", LongType.instance),
-            TTL("ttl", Int32Type.instance);
+            MAX_WRITE_TIME("maxwritetime", LongType.instance);
```
adelapena: Perhaps we should add a comment mentioning that versions up to 4.1 don't have `MAX_WRITE_TIME`.

adelapena: And/or maybe a comment telling that the order of the enum matters for serialization, so we or others don't repeat the same mistake if we modify this again?

src/java/org/apache/cassandra/cql3/selection/Selectable.java
adelapena: I guess we can fix it in a separate bugfix ticket, also targeting past releases.

adelapena: Just opened [CASSANDRA-17628](https://issues.apache.org/jira/browse/CASSANDRA-17628) to deal with this.

adelapena (approved):

adelapena (approved):
Looks good to me, +1

PR #949, opened 2021-04-02
CASSANDRA-16555 - Add out-of-the-box snitch for Ec2 IMDSv2

ocluf:
- added two new Ec2 snitches which use the Instance Meta Data Service v2 which requires a session token

PR #1587, opened 2022-04-25
CASSANDRA-16555 - Add out-of-the-box snitch for Ec2 IMDSv2 (new PR)

paulrutter:
New PR since the original branch for https://github.com/apache/cassandra/pull/949 was removed.

### DIFF
--- a/src/java/org/apache/cassandra/locator/Ec2MultiRegionSnitchIMDSv2.java
+++ b/src/java/org/apache/cassandra/locator/Ec2MultiRegionSnitchIMDSv2.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.cassandra.locator;
+
+import java.io.IOException;
+import java.net.InetAddress;
+
+
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.exceptions.ConfigurationException;
+import org.apache.cassandra.gms.ApplicationState;
+import org.apache.cassandra.gms.Gossiper;
+import org.apache.cassandra.service.StorageService;
+
+/**
+ * Is exactly the same as the Ec2MultiRegionSnitch except it uses the Instance MetaData Service v2 (IMDSv2) which
+ * requires you to get a session token first.
+ */
+public class Ec2MultiRegionSnitchIMDSv2 extends Ec2SnitchIMDSv2
+{
+
+    private static final String PUBLIC_IP_QUERY_URL = "http://169.254.169.254/latest/meta-data/public-ipv4";
+    private static final String PRIVATE_IP_QUERY_URL = "http://169.254.169.254/latest/meta-data/local-ipv4";
+    private final String localPrivateAddress;
+
+    public Ec2MultiRegionSnitchIMDSv2() throws IOException, ConfigurationException
+    {
+        super();
+        InetAddress localPublicAddress = InetAddress.getByName(awsApiCall(PUBLIC_IP_QUERY_URL));
+        logger.info("EC2Snitch using publicIP as identifier: {}", localPublicAddress);
+        localPrivateAddress = awsApiCall(PRIVATE_IP_QUERY_URL);
+        // use the Public IP to broadcast Address to other nodes.
+        DatabaseDescriptor.setBroadcastAddress(localPublicAddress);
+        if (DatabaseDescriptor.getBroadcastRpcAddress() == null)
+        {
+            logger.info("broadcast_rpc_address unset, broadcasting public IP as rpc_address: {}", localPublicAddress);
+            DatabaseDescriptor.setBroadcastRpcAddress(localPublicAddress);
+        }
+    }
+
+    @Override
+    public void gossiperStarting()
+    {
+        super.gossiperStarting();
+        Gossiper.instance.addLocalApplicationState(ApplicationState.INTERNAL_IP, StorageService.instance.valueFactory.internalIP(localPrivateAddress));
+        Gossiper.instance.register(new ReconnectableSnitchHelper(this, ec2region, true));
+    }
+}
--- a/src/java/org/apache/cassandra/locator/Ec2SnitchIMDSv2.java
+++ b/src/java/org/apache/cassandra/locator/Ec2SnitchIMDSv2.java
@@ -1,0 +1,108 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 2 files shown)


src/java/org/apache/cassandra/locator/Ec2MultiRegionSnitchIMDSv2.java
```diff
@@ -0,0 +1,63 @@
...
+    private final String localPrivateAddress;
+
+    public Ec2MultiRegionSnitchIMDSv2() throws IOException, ConfigurationException
+    {
+        super();
+        InetAddress localPublicAddress = InetAddress.getByName(awsApiCall(PUBLIC_IP_QUERY_URL));
+        logger.info("EC2Snitch using publicIP as identifier: {}", localPublicAddress);
+        localPrivateAddress = awsApiCall(PRIVATE_IP_QUERY_URL);
+        // use the Public IP to broadcast Address to other nodes.
+        DatabaseDescriptor.setBroadcastAddress(localPublicAddress);
```
paulrutter, line 48: @driftx this is the same as in the original snitch: https://github.com/blueconic/cassandra/blob/cassandra-2.2/src/java/org/apache/cassandra/locator/Ec2MultiRegionSnitch.java#L53

So we left it the same.

src/java/org/apache/cassandra/locator/Ec2SnitchIMDSv2.java
```diff
@@ -0,0 +1,108 @@
...
+                "Ec2SnitchIMDSv2 was unable to execute the API call. Not an ec2 node?");
+            }
+            // Read the information. I wish I could say (String) conn.getContent() here...
+            final int cl = conn.getContentLength();
+            final byte[] b = new byte[cl];
+            d = new DataInputStream((FilterInputStream) conn.getContent());
+            d.readFully(b);
+            return new String(b, StandardCharsets.UTF_8);
+        }
+        finally
```
paulrutter, line 102: The original doesn't do this, so we left this the same: https://github.com/blueconic/cassandra/blob/trunk/src/java/org/apache/cassandra/locator/Ec2Snitch.java#L104

conradkleinespel:
I'm also interested in seeing this merged. Let me know if there is anything I can do to help.

src/java/org/apache/cassandra/locator/Ec2SnitchIMDSv2.java
```diff
@@ -0,0 +1,108 @@
...
+        DataInputStream d = null;
+        try
+        {
+            if (conn.getResponseCode() != 200)
+            {
+                throw new ConfigurationException(
+                "Ec2SnitchIMDSv2 was unable to execute the API call. Not an ec2 node?");
+            }
+            // Read the information. I wish I could say (String) conn.getContent() here...
+            final int cl = conn.getContentLength();
```
jacek-lewandowski, line 96: Could you extract this content reading code into some method maybe in `EC2Snitch` or somewhere else and then, use there and in the legacy EC2 snitch impl? (only for trunk)

Also, would be good to cover the case where content length is returned -1, and read the whole input stream then (or read the whole input stream always, for example using `ByteStreams.toByteArray`)

src/java/org/apache/cassandra/locator/Ec2SnitchIMDSv2.java
```diff
@@ -0,0 +1,108 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * An implementation of the Ec2Snitch which uses Instance Meta Data Service v2 (IMDSv2) which requires you
+ * to get a session token first before calling the metaData service
+ */
+public class Ec2SnitchIMDSv2 extends Ec2Snitch
+{
+    protected static final Logger logger = LoggerFactory.getLogger(Ec2SnitchIMDSv2.class);
+    private static final long REFRESH_TOKEN_TIME = 21600;
```
jacek-lewandowski, line 39: The documentation says that it is the maximum TTL for the token - would be good to let the user configure that time using snitch properties (please remember to include the unit suffix in property name)

src/java/org/apache/cassandra/locator/Ec2SnitchIMDSv2.java
```diff
@@ -0,0 +1,108 @@
...
+    public Ec2SnitchIMDSv2() throws IOException, ConfigurationException
+    {
+        super();
+    }
+
+    @Override
+    String awsApiCall(final String url) throws IOException, ConfigurationException
+    {
+        // Populate the region and zone by introspection, fail if 404 on metadata
+        if (myToken == null || myLastTokenTime == null
```
jacek-lewandowski, line 58: nit: perhaps you can use `Suppliers.memoizeWithExpiration` to automatically control token lifetime

src/java/org/apache/cassandra/locator/Ec2MultiRegionSnitchIMDSv2.java
```diff
@@ -0,0 +1,63 @@
...
+public class Ec2MultiRegionSnitchIMDSv2 extends Ec2SnitchIMDSv2
+{
+
+    private static final String PUBLIC_IP_QUERY_URL = "http://169.254.169.254/latest/meta-data/public-ipv4";
+    private static final String PRIVATE_IP_QUERY_URL = "http://169.254.169.254/latest/meta-data/local-ipv4";
+    private final String localPrivateAddress;
+
+    public Ec2MultiRegionSnitchIMDSv2() throws IOException, ConfigurationException
+    {
+        super();
```
jacek-lewandowski, line 43: nit: redundant

src/java/org/apache/cassandra/locator/Ec2SnitchIMDSv2.java
```diff
@@ -0,0 +1,108 @@
...
+    private static final String AWS_EC2_METADATA_HEADER_TTL = "X-aws-ec2-metadata-token-ttl-seconds";
+    private static final String TOKEN_TTL_SECONDS = String.valueOf(REFRESH_TOKEN_TIME);
+    private static final String AWS_EC2_METADATA_HEADER = "X-aws-ec2-metadata-token";
+    private static final String TOKEN_ENDPOINT = "http://169.254.169.254/latest/api/token";
+
+    private String myToken;
+    private Long myLastTokenTime;
+
+
+    public Ec2SnitchIMDSv2() throws IOException, ConfigurationException
```
jacek-lewandowski, line 49: This constructor alone is redundant. You should add a constructor which takes `SnitchProperties` so that both are justified

src/java/org/apache/cassandra/locator/Ec2SnitchIMDSv2.java
```diff
@@ -0,0 +1,108 @@
...
+ * to get a session token first before calling the metaData service
+ */
+public class Ec2SnitchIMDSv2 extends Ec2Snitch
+{
+    protected static final Logger logger = LoggerFactory.getLogger(Ec2SnitchIMDSv2.class);
+    private static final long REFRESH_TOKEN_TIME = 21600;
+    private static final String AWS_EC2_METADATA_HEADER_TTL = "X-aws-ec2-metadata-token-ttl-seconds";
+    private static final String TOKEN_TTL_SECONDS = String.valueOf(REFRESH_TOKEN_TIME);
+    private static final String AWS_EC2_METADATA_HEADER = "X-aws-ec2-metadata-token";
+    private static final String TOKEN_ENDPOINT = "http://169.254.169.254/latest/api/token";
```
jacek-lewandowski, line 43: The endpoint IP is common for all of the addressed used here and in other EC2 classes. Would be good to extract it into a single field in `Ec2Snitch` class. If it was additionally configurable via `SnitchProperties` those classes would be more testable as we could simulate the EC2 endpoint (applies only for PR against trunk)

src/java/org/apache/cassandra/locator/Ec2MultiRegionSnitchIMDSv2.java
```diff
@@ -0,0 +1,63 @@
...
+ * requires you to get a session token first.
+ */
+public class Ec2MultiRegionSnitchIMDSv2 extends Ec2SnitchIMDSv2
+{
+
+    private static final String PUBLIC_IP_QUERY_URL = "http://169.254.169.254/latest/meta-data/public-ipv4";
+    private static final String PRIVATE_IP_QUERY_URL = "http://169.254.169.254/latest/meta-data/local-ipv4";
+    private final String localPrivateAddress;
+
+    public Ec2MultiRegionSnitchIMDSv2() throws IOException, ConfigurationException
```
jacek-lewandowski, line 41: Similar to `Ec2SnitchIMDSv2` - please add a constructor accepting `SnitchProperties`

jacek-lewandowski (changes requested):
Overall looks good. For trunk I'd consider refactoring it a bit though

PR #1595, opened 2022-04-28
CASSANDRA-16843: List snapshots of dropped tables

pauloricardomg:


src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -2974,6 +2974,7 @@ public double getDroppableTombstoneRatio()
 
     public long trueSnapshotsSize()
     {
+
```
smiklosovic: nit: this line can be deleted.

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -1244,9 +1249,8 @@ private static String join(String... s)
     private class SSTableSizeSummer extends DirectorySizeCalculator
     {
         private final Set<String> toSkip;
-        SSTableSizeSummer(File path, List<File> files)
+        SSTableSizeSummer(List<File> files)
         {
-            super(path);
             toSkip = files.stream().map(f -> f.name()).collect(Collectors.toSet());
```
smiklosovic: can be simplified by `map(File::name)`

src/java/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDC.java
```diff
@@ -274,9 +274,12 @@ private static class CDCSizeTracker extends DirectorySizeCalculator
         // track the total size between two dictionary size calculations
         private final AtomicLong sizeInProgress;
 
+        private final File path;
+
         CDCSizeTracker(CommitLogSegmentManagerCDC segmentManager, File path)
         {
-            super(path);
+            super();
```
smiklosovic: why do we need this `super()` here to be called explicitly?

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
```diff
@@ -0,0 +1,141 @@
...
+    private boolean walked = false;
+
+    public SnapshotLoader()
+    {
+        this(DatabaseDescriptor.getAllDataFileLocations());
+    }
+
+    public SnapshotLoader(String[] dataDirs)
+    {
+        this.dataDirectories = Arrays.asList(dataDirs).stream().map(d -> Paths.get(d)).collect(Collectors.toList());
```
smiklosovic: can be all simplified to `this.dataDirectories = Arrays.stream(dataDirs).map(Paths::get).collect(Collectors.toList());`

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
```diff
@@ -0,0 +1,141 @@
...
+                    Files.walkFileTree(dataDir, Collections.EMPTY_SET, 5, this);
+                }
+                catch (IOException e)
+                {
+                    throw new RuntimeException(String.format("Error while loading snapshots from %s", dataDir));
+                }
+            }
+            walked = true;
+        }
+        return snapshots.values().stream().map(s -> s.build()).collect(Collectors.toSet());
```
smiklosovic: I prefer method referencies here like "TableSnapshot.Builder::build" instead of `s -> s.build()`

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
```diff
@@ -0,0 +1,141 @@
...
+            logger.trace("Processing directory " + subdir);
+            Matcher snapshotDirMatcher = SNAPSHOT_DIR_PATTERN.matcher(subdir.toString());
+            if (snapshotDirMatcher.find())
+            {
+                try
+                {
+                    loadSnapshotFromDir(snapshotDirMatcher, subdir);
+                } catch (Throwable e)
+                {
+                    logger.warn("Could not load snapshot from %s.", subdir, e);
```
smiklosovic: IDEA says: "More arguments provided (1) than placeholders specified (0) ". I think you forgot to String.format it.

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
```diff
@@ -0,0 +1,141 @@
...
+                }
+            }
+            return FileVisitResult.SKIP_SUBTREE;
+        }
+
+        return subdir.equals(Directories.BACKUPS_SUBDIR)
```
smiklosovic: you are comparing Path with String here.

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * Given an UUID string without dashes (ie. c7e513243f0711ec9bbc0242ac130002)
+     * return an UUID object (ie. c7e51324-3f07-11ec-9bbc-0242ac130002)
+     */
+    protected static UUID parseUUID(String uuidWithoutDashes) throws IllegalArgumentException
+    {
+        assert uuidWithoutDashes.length() == 32 && !uuidWithoutDashes.contains("-");
+        String dashedUUID = uuidWithoutDashes.replaceFirst("([0-9a-f]{8})([0-9a-f]{4})([0-9a-f]{4})([0-9a-f]{4})([0-9a-f]+)", "$1-$2-$3-$4-$5");
```
smiklosovic, line 134: you might replace this by compiled Pattern

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
```diff
@@ -17,50 +17,76 @@
...
 package org.apache.cassandra.service.snapshot;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.time.Instant;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.Map;
```
smiklosovic: you have few unused imports here

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
```diff
@@ -113,54 +139,171 @@ public long computeSizeOnDiskBytes()
...
     public boolean equals(Object o)
     {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
-        TableSnapshot that = (TableSnapshot) o;
-        return Objects.equals(keyspace, that.keyspace) && Objects.equals(table, that.table)
-               && Objects.equals(tag, that.tag) && Objects.equals(createdAt, that.createdAt)
-               && Objects.equals(expiresAt, that.expiresAt) && Objects.equals(snapshotDirs, that.snapshotDirs);
+        TableSnapshot snapshot = (TableSnapshot) o;
+        return Objects.equals(keyspaceName, snapshot.keyspaceName) && Objects.equals(tableName, snapshot.tableName) && Objects.equals(tableId, snapshot.tableId) && Objects.equals(tag, snapshot.tag) && Objects.equals(createdAt, snapshot.createdAt) && Objects.equals(expiresAt, snapshot.expiresAt) && Objects.equals(snapshotDirs, snapshot.snapshotDirs);
```
smiklosovic: could you please make this line shorter?

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
```diff
@@ -113,54 +139,171 @@ public long computeSizeOnDiskBytes()
...
+
+        for (File snapshotDir : snapshotDirs)
+        {
+            try
+            {
+                Files.walkFileTree(snapshotDir.toPath(), visitor);
+            }
+            catch (IOException e)
+            {
+                logger.error("Could not calculate the size of {}. {}", snapshotDir, e);
```
smiklosovic: Should not this be replaced by `error(String msg, Throwable t)`? I do not think that this version of the method is the one we want. Sorry if I am wrong.

src/java/org/apache/cassandra/utils/DirectorySizeCalculator.java
```diff
@@ -49,12 +45,14 @@ public boolean isAcceptable(Path file)
     public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) throws IOException
     {
         if (isAcceptable(file))
+        {
```
smiklosovic: I think that if there is just one statement in the body of `if`, the parenthesis might be left out completely.

test/unit/org/apache/cassandra/service/snapshot/TableSnapshotTest.java
```diff
@@ -216,25 +210,41 @@ public void testGetCreatedAt() throws IOException
...
-        (File file) -> 0L
-        );
+        folders);
         assertThat(withoutCreatedAt.getCreatedAt()).isEqualTo(Instant.ofEpochMilli(folders.stream().mapToLong(f -> f.lastModified()).min().getAsLong()));
     }
 
+    @Test
+    public void testGetLiveFileFromSnapshotFile()
+    {
+        testGetLiveFileFromSnapshotFile("~/.ccm/test/node1/data0/test_ks/tbl-e03faca0813211eca100c705ea09b5ef/snapshots/1643481737850/me-1-big-Data.db",
```
smiklosovic, line 235: are these .ccm methods really something we are interested in? I think this is the leftover from your local testing.

smiklosovic: ah no sorry, I see what you did there now. Forget it.

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
smiklosovic: could not be `this.dataDirectories` renamed to `this.dataDirs` as you are putting `dataDirs` into the constructor? I just dont like this minor discrepancy.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4129,24 +4130,26 @@ public void clearSnapshot(String tag, String... keyspaceNames) throws IOExceptio
 
     public Map<String, TabularData> getSnapshotDetails(Map<String, String> options)
     {
+        boolean skipExpiring = options != null && Boolean.parseBoolean(options.getOrDefault("no_ttl", "false"));
+
+        SnapshotLoader loader = new SnapshotLoader();
         Map<String, TabularData> snapshotMap = new HashMap<>();
-        for (Keyspace keyspace : Keyspace.all())
+
+        for (TableSnapshot snapshot : loader.loadSnapshots())
```
smiklosovic, line 4144: I see you have 'walked' variable in loadSnapshots method, I guess it is for not needing to go over whole data structure multiple times. However, is it really used in this method? This is called from JMX when I am not mistaken, plus you create new instance of SnapshotLoader every time we use nodetool, so it will parse all data dirs anyway. Maybe we should cache this? I think it would be better to do this via SnapshotManager and I believe you have this in mind, just the proximity of the release is blocking you from doing it "properly".

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
pauloricardomg: 👍

src/java/org/apache/cassandra/db/Directories.java
pauloricardomg: 👍

src/java/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDC.java
pauloricardomg: Good catch. Removed it.

src/java/org/apache/cassandra/service/StorageService.java
pauloricardomg: > I see you have 'walked' variable in loadSnapshots method, I guess it is for not needing to go over whole data structure multiple times. However, is it really used in this method?

Removed `walked` variable from `SnapshotLoader` since it's no longer neeeded. This is a leftover from a previous implementation.


> Maybe we should cache this? I think it would be better to do this via SnapshotManager and I believe you have this in mind, just the proximity of the release is blocking you from doing it "properly".

Yes, I have a longer version of this patch where loaded snapshots are cached in the `SnapshotManager` ([here](https://github.com/apache/cassandra/compare/trunk...pauloricardomg:CASSANDRA-16843-final?expand=1#diff-17b12c3f18d24601e8d6dcdf288f6583624ebf1b5cd1b135cfe1ed5abd6d349dR71)). However in order to cache this properly we need to move snapshot cleanup logic to `SnapshotManager`, which makes this patch bigger than it should be and riskier for 4.1. I plan to create follow-up patches based on the longer patch above centralizing snapshot logic on `SnapshotManager` for 4.2.

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
pauloricardomg: 👍

pauloricardomg: updated constructor to match `dataDirectories`

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
pauloricardomg: 👍

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
pauloricardomg: good catch, fixed.

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
pauloricardomg: good catch, fixed.

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
pauloricardomg: I could, but I prefer not to touch this right now as I'm not really sure how to use `Pattern` here and this is working as expected. Feel free to offer a suggestion on how to improve this.

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
pauloricardomg: Fixed.

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
pauloricardomg: Good catch, fixed it.

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
pauloricardomg: Done.

src/java/org/apache/cassandra/utils/DirectorySizeCalculator.java
pauloricardomg: fixed

test/unit/org/apache/cassandra/service/snapshot/TableSnapshotTest.java
pauloricardomg: 👍

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
```diff
@@ -0,0 +1,137 @@
...
+            logger.trace("Processing directory " + subdir);
+            Matcher snapshotDirMatcher = SNAPSHOT_DIR_PATTERN.matcher(subdir.toString());
+            if (snapshotDirMatcher.find())
+            {
+                try
+                {
+                    loadSnapshotFromDir(snapshotDirMatcher, subdir);
+                } catch (Throwable e)
+                {
+                    logger.warn("Could not load snapshot from {}.", subdir, e);
```
smiklosovic, line 105: I think this is still not fixed, is it? this will call warn(String, Object, Object). It is different from warn(String,Throwable). And there are not two {} anyway.

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
```diff
@@ -113,54 +136,174 @@ public long computeSizeOnDiskBytes()
...
+
+        for (File snapshotDir : snapshotDirs)
+        {
+            try
+            {
+                Files.walkFileTree(snapshotDir.toPath(), visitor);
+            }
+            catch (IOException e)
+            {
+                logger.error("Could not calculate the size of {}.", snapshotDir, e);
```
smiklosovic, line 149: same as above.

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
```diff
@@ -113,54 +136,174 @@ public long computeSizeOnDiskBytes()
...
+            try
+            {
+                logger.debug("Loading snapshot manifest from {}", manifestFile);
+                SnapshotManifest manifest = SnapshotManifest.deserializeFromJsonFile(manifestFile);
+                createdAt = manifest.createdAt;
+                expiresAt = manifest.expiresAt;
+            }
+            catch (IOException e)
+            {
+                logger.warn("Cannot read manifest file {} of snapshot {}.", manifestFile, tag, e);
```
smiklosovic, line 258: same ... I live with the opinion that this API expects Object's, not two objects and Throwable. There are other methods for logging throwables. This does not work I think or I am completely wrong here.
I am used to do it like logger.warn(String.format("%s %s", a, b), e); so it calls warn(String,Throwable). If there is warn(Objects...) I can indeed do warn("{} {} {}", a, b, c) but I would not put Throwable into it.

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
pauloricardomg: Yes this is fixed - check this: https://www.slf4j.org/faq.html#paramException

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
pauloricardomg: This is working as expected: https://www.slf4j.org/faq.html#paramException

The first parameter will be added to the message, and the exception trace will be print.

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
pauloricardomg: Read this: https://www.slf4j.org/faq.html#paramException

pauloricardomg:
```
The SLF4J API supports parametrization in the presence of an exception, assuming the exception is the last parameter. Thus, 
```

I've been using this for a while and it's the recommended way to log exception traces, and no longer use the previous workaround `logger.warn(String.format("%s %s", a, b), e)` which is required for `SLF4J < 1.6.0`.

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
smiklosovic: Oh wow, that is very nice. So resolved.

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
smiklosovic: great.

PR #1596, opened 2022-04-30
CASSANDRA-16790: Add auto_snapshot_ttl configuration

pauloricardomg:


conf/cassandra.yaml
```diff
@@ -876,6 +876,14 @@ snapshot_before_compaction: false
 # lose data on truncation or drop.
 auto_snapshot: true
 
+# Adds a time-to-live (TTL) to auto snapshots generated by table
+# truncation or drop (when enabled).
+# After the TTL is elapsed, the snapshot is automatically cleared.
+# By default, auto snapshots *do not* have TTL, unncomment the property below
```
smiklosovic: typo on "unncomment", double n

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2805,6 +2818,17 @@ public static boolean isAutoSnapshot()
...
 
+    public static DurationSpec getAutoSnapshotTtl()
+    {
+        return autoSnapshoTtl;
+    }
+
+    @VisibleForTesting
+    public static void setAutoSnapshotTtl(DurationSpec autoSnapshotTtl)
+    {
+        autoSnapshoTtl = autoSnapshotTtl;
```
smiklosovic: this is strage at first sight, you are basically assigning variable to static variable of the same name, can not be this somehow changed?

test/distributed/org/apache/cassandra/distributed/test/AutoSnapshotTtlTest.java
```diff
@@ -0,0 +1,150 @@
...
+        properties.close();
+    }
+
+    /**
+     * Check that when auto_snapshot_ttl=5s, snapshots created from TRUNCATE are expired after 10s
+     */
+    @Test
+    public void testAutoSnapshotTTlOnTruncate() throws IOException, InterruptedException
+    {
+        try (Cluster cluster = build().withNodes(1)
```
smiklosovic: if you wrap whole `build()....start()` in `init()` from `DistributedTestBase` (which `TestBaseImpl` extends), it will create a keyspace for you with replication factor of 1 so you do not need to create another one on line 70. Then, on line 71, you wrap your `CREATE TABLE ...` with `withKeyspace` method from `DistributedTestBase` which will format it with the keyspace it created. 

Same might be done for all tests in this test class.

test/distributed/org/apache/cassandra/distributed/test/AutoSnapshotTtlTest.java
```diff
@@ -0,0 +1,150 @@
...
+            populate(cluster);
+
+            // Drop Table
+            cluster.schemaChange("DROP TABLE default.tbl;");
+
+            // Check snapshot is listed after table is dropped
+            instance.nodetoolResult("listsnapshots").asserts().success().stdoutContains(ColumnFamilyStore.SNAPSHOT_DROP_PREFIX);
+
+            // Check snapshot is removed after 10s
+            Thread.sleep(2 * FIVE_SECONDS * 1000L);
```
smiklosovic: This should be replaced by `Awaitility`. Await.poll.until(...) with timeout more than 10s.

conf/cassandra.yaml
pauloricardomg: 👍

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
pauloricardomg: renamed variable to `newTtl`

test/distributed/org/apache/cassandra/distributed/test/AutoSnapshotTtlTest.java
pauloricardomg: 👍 good suggestion

test/distributed/org/apache/cassandra/distributed/test/AutoSnapshotTtlTest.java
pauloricardomg: Updated on both `testAutoSnapshotTTlOnTruncate` and `testAutoSnapshotTTlOnDrop`.

Did not update on `testAutoSnapshotTtlDisabled` because it needs to wait the full 10 seconds to ensure the auto snapshots are *not* removed.

PR #1606, opened 2022-05-05
CASSANDRA-17593: StorageService read threshold get methods throw NullPointerException due to not handling null configs

dcapwell:


src/java/org/apache/cassandra/locator/TokenMetadata.java
```diff
@@ -601,7 +601,7 @@ public Collection<Token> getTokens(InetAddressAndPort endpoint)
         lock.readLock().lock();
         try
         {
-            assert isMember(endpoint); // don't want to return nulls
+            assert isMember(endpoint): String.format("Unable to get tokens for %s; it is not a member", endpoint); // don't want to return nulls
```
dcapwell, line 604: was found in testing, I don't like `null` messages, so added message

test/distributed/org/apache/cassandra/distributed/test/jmx/JMXGetterCheckTest.java
```diff
@@ -0,0 +1,143 @@
...
+import org.apache.cassandra.distributed.test.TestBaseImpl;
+import org.apache.cassandra.utils.JMXServerUtils;
+
+import static org.apache.cassandra.config.CassandraRelevantProperties.IS_DISABLED_MBEAN_REGISTRATION;
+import static org.apache.cassandra.cql3.CQLTester.getAutomaticallyAllocatedPort;
+
+public class JMXGetterCheckTest extends TestBaseImpl
+{
+    private static final Set<String> IGNORE_ATTRIBUTES = ImmutableSet.of(
+    "org.apache.cassandra.net:type=MessagingService:BackPressurePerHost" // throws unsupported saying the feature was removed... this feels like a regression...
```
bereng: Should a ticket be created for this?

test/distributed/org/apache/cassandra/distributed/test/jmx/JMXGetterCheckTest.java
```diff
@@ -0,0 +1,143 @@
...
+    "org.apache.cassandra.db:type=StorageService:stopDaemon", // halts the instance, which then causes the JVM to exit
+    "org.apache.cassandra.db:type=StorageService:drain", // don't drain, it stops things which can cause other APIs to be unstable as we are in a stopped state
+    "org.apache.cassandra.db:type=StorageService:stopGossiping" // if we stop gossip this can causes other issues, so avoid
+    );
+
+    @Test
+    public void test() throws Exception
+    {
+        // start JMX server, which the instance will register with
+        InetAddress loopback = InetAddress.getLoopbackAddress();
```
bereng, line 63: Asking genuinely from my ignorance with jvm dtests: isn't/shouldn't there be helpers to do JMX in  the jvm-dtest package so you don't have to handcraft it? Maybe we could add this now or maybe it's food for another ticket. Wdyt?

dcapwell: > there be helpers to do JMX in the jvm-dtest package

JMX can not be isolated in the ClassLoader (or at least the last attempt failed), so its global; due to this reason we don't normally do JMX in dtest, but we do support Metrics, which is part of the API (rather than register with JMX, we register with a global map)

> shouldn't there be helpers 

Personally I don't like helpers until there are more than 1 usage

test/distributed/org/apache/cassandra/distributed/test/jmx/JMXGetterCheckTest.java
dcapwell: looks like it was removed in https://issues.apache.org/jira/browse/CASSANDRA-15375

dcapwell: updated comment to say what JIRA removed.

PR #1607, opened 2022-05-06
CASSANDRA-17580: Clients using JMX are unable to handle non-standard java types but we leak this into our Exceptions

dcapwell:


src/java/org/apache/cassandra/config/Config.java
```diff
@@ -853,6 +853,8 @@ public static void setClientMode(boolean clientMode)
     public volatile DurationSpec repair_state_expires = DurationSpec.inDays(3);
     public volatile int repair_state_size = 100_000;
 
+    public boolean jmx_hide_non_java_exceptions = true;
```
ekaterinadimitrova2: Let's make this one `false` by default as we spoke.
In that sense I suggest on this branch (https://github.com/ekaterinadimitrova2/cassandra/commits/17580-review) a few commits to change the config to throw `IllegalArgumentException`  as we actually make `loadConfig` to throw ConfigurationException.

I also corrected a few setters in the DB descriptor which were not released yet
All in all there  I see only three problematic now which were already released.
Wondering whether we should add new versions that throw `IllegalArgumentException` and deprecate the old ones.

`setRepairSessionSpaceInMiB, setMaxConcurrentAutoUpgradeTasks `
`checkValidForByteConversion` is used both on startup and in a few setters.

ekaterinadimitrova2: Also, we need NEWS.txt entry

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -147,6 +147,8 @@ private void maybeInitJmx()
         {
             logger.warn("JMX settings in cassandra-env.sh have been bypassed as the JMX connector server is " +
                         "already initialized. Please refer to cassandra-env.(sh|ps1) for JMX configuration info");
+
+            JMXServerUtils.hackPatchGlobalMBeanServerToFixExceptions();
```
ekaterinadimitrova2: I guess this was WIP name :D How about `patchGlobalMBeanServerToFixExceptions`?

src/java/org/apache/cassandra/config/Config.java
ekaterinadimitrova2: Maybe also a bit of JavaDoc will be nice

test/unit/org/apache/cassandra/utils/JMXServerUtilsTest.java
```diff
@@ -0,0 +1,118 @@
...
+        public String getValueNested()
+        {
+            int levels = 10;
+            RuntimeException e = new ConfigurationException("bottom level");
+            for (int i = 0; i < levels; i++)
+                e = new RuntimeException("level " + i, e);
+            throw e;
+        }
+    }
+}
```
ekaterinadimitrova2: new line at the end

src/java/org/apache/cassandra/utils/JMXServerUtils.java
```diff
@@ -143,6 +161,148 @@ public static JMXConnectorServer createJMXServer(int port, String hostname, bool
...
+                    catch (Throwable t)
+                    {
+                        logger.warn("Unable to intersept JMX", t);
+                    }
+                }
+            }
+        }
+        catch (Throwable t)
+        {
+            logger.warn("Unable to intersept JMX", t);
```
ekaterinadimitrova2: typo: intercept

src/java/org/apache/cassandra/utils/ForwardingMBeanServer.java
```diff
@@ -0,0 +1,274 @@
...
+        this.delegate = Objects.requireNonNull(mbeanServer);
+    }
+
+    protected MBeanServer delegate()
+    {
+        return delegate;
+    }
+
+    @Override
+    public ObjectInstance createMBean(String className, ObjectName name) throws ReflectionException, InstanceAlreadyExistsException, MBeanRegistrationException, MBeanException, NotCompliantMBeanException
```
ekaterinadimitrova2, line 60: In theory I guess we can get rid of  `MBeanException` here, but I see that `MBeanServer#createMBean` keeps it so I guess I am fine with it

src/java/org/apache/cassandra/utils/ForwardingMBeanServer.java
```diff
@@ -0,0 +1,274 @@
...
+    @Override
+    public Object instantiate(String className, ObjectName loaderName, Object[] params, String[] signature) throws ReflectionException, MBeanException, InstanceNotFoundException
+    {
+        return delegate().instantiate(className, loaderName, params, signature);
+    }
+
+    @Override
+    public ObjectInputStream deserialize(ObjectName name, byte[] data) throws InstanceNotFoundException, OperationsException
+    {
+        return delegate().deserialize(name, data);
```
ekaterinadimitrova2, line 242: Seems those `deserialize` methods are deprecated in favor of `getClassLoader`, did you consider that?

src/java/org/apache/cassandra/config/Config.java
ekaterinadimitrova2: Oh also in my branch I removed the validation of the config file as it is not needed anymore, the two problematic properties were set with `0ms` default value. Not being null anymore doesn't allow to have the problem that the validation was trying to prevent. Let me know if it can be part of this patch or you want me to pull it. Just int doesn't make sense to read one more time the config file when not actually needed anymore.

ekaterinadimitrova2: Oh I have bad sad news. I pushed my commit and then got an email from Circle that Java11 build failed:
```
_build_java:
     [echo] Compiling for Java 11...
    [javac] Compiling 2137 source files to /home/cassandra/cassandra/build/classes/main
    [javac] Note: Processing compiler hints annotations
    [javac] /home/cassandra/cassandra/src/java/org/apache/cassandra/utils/JMXServerUtils.java:67: error: package com.sun.jmx.mbeanserver is not visible
    [javac] import com.sun.jmx.mbeanserver.JmxMBeanServer;
    [javac]                   ^
    [javac]   (package com.sun.jmx.mbeanserver is declared in module java.management, which does not export it)
    [javac] Note: Processing compiler hints annotations
    [javac] Note: Writing compiler command file at META-INF/hotspot_compiler
    [javac] Note: Done processing compiler hints annotations
    [javac] 1 error
```

I feel the community won't be happy to open more internals for this :(

ekaterinadimitrova2 (commented):
I have bad news about the build in J11,but already provided some solutions for some of the issues we know about at least. Please check and let me know what you think.

src/java/org/apache/cassandra/utils/ForwardingMBeanServer.java
dcapwell: this class is meant to proxy all requests to the base to simplify overloading classes, so shouldn't call a different method than expected.

src/java/org/apache/cassandra/config/Config.java
dcapwell: hmmm, not sure how to *fix*, in the case where we let the JVM setup the server, we need to be able to inject our logic... ill try to look into another way

dcapwell: have a solution by plugging deeper into jmx startup

PR #1611, opened 2022-05-09
CASSANDRA-17613 Improve the way we handle repair message timeouts to avoid hanging repairs

krummas:


src/java/org/apache/cassandra/config/Config.java
```diff
@@ -134,6 +134,8 @@
     public volatile Integer repair_session_max_tree_depth = null;
     public volatile Integer repair_session_space_in_mb = null;
 
+    public volatile long repair_message_timeout_in_ms = TimeUnit.MILLISECONDS.convert(1, TimeUnit.MINUTES);
```
dcapwell: when this gets added to 4.1 should this also be migrated?  Feels weird to migrate as its new... @ekaterinadimitrova2 thoughts?  Should we just add the needed data types so we can do `DurationSpec repair_message_timeout = DurationSpec.ofMinutes(1);`?

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
```diff
@@ -26,10 +42,57 @@
  */
 public abstract class RepairMessage
 {
+    private static final CassandraVersion SUPPORTS_TIMEOUTS = new CassandraVersion("4.0.4-SNAPSHOT"); // todo: should be 4.0.5 (4.0.4 is about to be cut)
```
dcapwell: should be 4.0.6 at this point: 4.0.5 released on July 18

dcapwell (approved):

src/java/org/apache/cassandra/config/Config.java
ekaterinadimitrova2: I see your point but honestly, I am not sure. It seems like something that might require consensus on the mailing list? If we consider we might also have other similar cases in the future?

ekaterinadimitrova2: Side note: if we decide not to port them and to deprecate as others in. 4.1, don't forget we also need a ccm patch, a new entry here - https://github.com/riptano/ccm/blob/4609ac46beccd3ab86463a345ed59e998d664d04/ccmlib/common.py#L155

PR #1615, opened 2022-05-10
CASSANDRA-17619: Remove expired snapshots of dropped tables after restart

pauloricardomg:


src/java/org/apache/cassandra/service/snapshot/SnapshotManager.java
```diff
@@ -116,8 +108,15 @@ public synchronized void addSnapshot(TableSnapshot snapshot)
...
-        logger.debug("Loading snapshots");
-        snapshotLoader.get().forEach(this::addSnapshot);
+        SnapshotLoader loader = new SnapshotLoader(DatabaseDescriptor.getAllDataFileLocations());
+        addSnapshots(loader.loadSnapshots());
+    }
+
+    @VisibleForTesting
+    protected synchronized void addSnapshots(Collection<TableSnapshot> snapshots)
+    {
+        logger.debug("Adding snapshots: {}", snapshots.stream().map(s -> s.getId()).collect(Collectors.toList()));
```
smiklosovic: To be explicit, this will be printed like: "Adding snapshots: ["snapshot1", "snapshot2","snapshot3"]" ?
I think there is some joining collector which just separates it by commas and leaves brackets. I consider it to be just nicer.

src/java/org/apache/cassandra/service/snapshot/SnapshotManager.java
```diff
@@ -116,8 +108,15 @@ public synchronized void addSnapshot(TableSnapshot snapshot)
     @VisibleForTesting
     protected synchronized void loadSnapshots()
     {
-        logger.debug("Loading snapshots");
-        snapshotLoader.get().forEach(this::addSnapshot);
+        SnapshotLoader loader = new SnapshotLoader(DatabaseDescriptor.getAllDataFileLocations());
```
smiklosovic, line 112: Dont you think we might have `loader` as a property of `SnapshotManager` ? Why do we need to create an instance every time?

We are also creating it in `StorageService.getSnapshotDetails`. Can not we somehow make it by one instance?

btw, I just went through that SnapshotLoader class as such and I would change one of the constructors to this:

    public SnapshotLoader(String[] dataDirectories)
    {
        this(Arrays.stream(dataDirectories).map(Paths::get).collect(Collectors.toList()));
    }

I know it is not relevant to this patch and changes related to this but I find it just easy to change, really. Just an idea.

test/distributed/org/apache/cassandra/distributed/test/SnapshotsTest.java
```diff
@@ -222,6 +224,71 @@ public void testListSnapshotOfDroppedTable()
         instance.nodetoolResult("listsnapshots").asserts().success().stdoutContains("tag1");
     }
 
+    @Test
+    public void testTTLSnapshotOfDroppedTable()
+    {
+        IInvokableInstance instance = cluster.get(1);
+
+        cluster.schemaChange("CREATE KEYSPACE IF NOT EXISTS default WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 2};");
```
smiklosovic: why rf 2? dont we have just one node? Also, can no you use that distributed keyspace which is created already?

test/distributed/org/apache/cassandra/distributed/test/SnapshotsTest.java
```diff
@@ -222,6 +224,71 @@ public void testListSnapshotOfDroppedTable()
...
+               .until(() -> !instance.nodetoolResult("listsnapshots").getStdout().contains("tag1"));
+    }
+
+    @Test
+    public void testTTLSnapshotOfDroppedTableAfterRestart()
+    {
+        int TWENTY_SECONDS = 20; // longer TTL to allow snapshot to survive node restart
+        IInvokableInstance instance = cluster.get(1);
+
+        cluster.schemaChange("CREATE KEYSPACE IF NOT EXISTS default WITH replication = {'class': 'SimpleStrategy', 'replication_factor': 2};");
```
smiklosovic: same, rf = 2, why not 1 and why dont we use default keyspace.

test/distributed/org/apache/cassandra/distributed/test/SnapshotsTest.java
```diff
@@ -35,7 +35,10 @@
 import org.apache.cassandra.distributed.api.NodeToolResult;
 import org.apache.cassandra.distributed.shared.WithProperties;
 
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.apache.cassandra.db.ColumnFamilyStore.SNAPSHOT_DROP_PREFIX;
 import static org.apache.cassandra.distributed.shared.ClusterUtils.stopUnchecked;
+import static org.awaitility.Awaitility.await;
 
 public class SnapshotsTest extends TestBaseImpl
```
smiklosovic, line 43: do we also test (somewhere) that dropping of whole keyspace will create snapshots on individual tables?

src/java/org/apache/cassandra/service/snapshot/SnapshotManager.java
pauloricardomg: > Dont you think we might have loader as a property of SnapshotManager ? Why do we need to create an instance every time?

Since this is a bug fix I would prefer not to make loader a property of SnapshotManager because it's out of the scope of this task.

The SnapshotLoader class was originally designed to be used only once, like a builder. In order to make the class reusable I would need to change the design and update tests so I would prefer to leave for post 4.1.

After 4.1 is out I will create a new PR with a major refactoring moving all snapshot logic to SnapshotManager, based on this branch implementation:
* https://github.com/pauloricardomg/cassandra/tree/CASSANDRA-16843-final

> We are also creating it in StorageService.getSnapshotDetails. Can not we somehow make it by one instance?

same as above, prefer to not change this for now but will do later as part of the refactoring.

Is this ok for now? I can add a comment with a small TODO if you'd like.

pauloricardomg: > btw, I just went through that SnapshotLoader class as such and I would change one of the constructors to this:

updated 👍

src/java/org/apache/cassandra/service/snapshot/SnapshotManager.java
pauloricardomg: Updated. New output:
```
DEBUG [main] 2022-05-12 17:08:30,773 SnapshotManager.java:119 - Adding snapshots: ks:tbl:c7483545-103a-493e-b66e-6d96e60d3878:expired, ks:tbl:984bd8f1-35fd-44d7-b69b-52422614e25f:non-expired, ks:tbl:deff9907-495b-4497-8275-ffdfdfc3ad7b:non-expiring.
```

test/distributed/org/apache/cassandra/distributed/test/SnapshotsTest.java
pauloricardomg: good catch - added new test on 834229664d6c2b3b6a044e7fd2571eca48df6a74

test/distributed/org/apache/cassandra/distributed/test/SnapshotsTest.java
pauloricardomg: > why rf 2?

copied over from previous tests - fixed all tests to use RF=1 on 007958f8746799f4ca89a51b82bdd9d20eae1494.

> Also, can no you use that distributed keyspace which is created already?

This whole class uses this format and fixing would be out of the scope of this ticket. Is it fine if we refactor this later?

test/distributed/org/apache/cassandra/distributed/test/SnapshotsTest.java
pauloricardomg: fixed to use rf=1 on 007958f8746799f4ca89a51b82bdd9d20eae1494

src/java/org/apache/cassandra/service/snapshot/SnapshotManager.java
smiklosovic: Ehm ... should not it be the name of the snapshot rather than keyspace:table:id? I mean, we are adding _snapshots_, I do not see the _names_ of these snapshots or am I getting something wrong?

src/java/org/apache/cassandra/service/snapshot/SnapshotManager.java
```diff
@@ -116,16 +109,24 @@ public synchronized void addSnapshot(TableSnapshot snapshot)
...
+        addSnapshots(loader.loadSnapshots());
+    }
+
+    @VisibleForTesting
+    protected synchronized void addSnapshots(Collection<TableSnapshot> snapshots)
+    {
+        logger.debug("Adding snapshots: {}.", Joiner.on(", ").join(snapshots.stream().map(s -> s.getId()).collect(Collectors.toList())));
```
smiklosovic, line 119: this looks better but what I see is that we have "keyspace:table:id" rather than _name of that snapshot_ (snapshot tag).

smiklosovic: Ahaaaa I get this, so after "id" there is the snapshot tag, the whole name has four components, `ks:tb:id:tag`. I overlooked the last part. All good.

PR #1621, opened 2022-05-12
CASSANDRA-17623 - Fix issue where frozen maps may not be serialized in the correct order

JeetKunDoug:


frankgh (approved):
+1

test/unit/org/apache/cassandra/cql3/validation/operations/InsertTest.java
```diff
@@ -247,4 +252,42 @@ public void testCKInsertWithValueOver64K() throws Throwable
         assertInvalidThrow(InvalidRequestException.class,
                            "INSERT INTO %s (a, b) VALUES ('foo', ?)", new String(TOO_BIG.array()));
     }
+
+    @Test
+    public void testInsertingMapDataWithParameterizedQueriesIsKeyOrderIndependent() throws Throwable
```
adelapena: This test is mostly identical to the one on `SelectTest`, we could combine them and put them in `CollectionsTest.java`.

test/unit/org/apache/cassandra/cql3/validation/operations/InsertTest.java
```diff
@@ -247,4 +252,42 @@ public void testCKInsertWithValueOver64K() throws Throwable
...
+    {
+        UUID uuid1 = UUIDs.timeBased();
+        UUID uuid2 = UUIDs.timeBased();
+        createTable("CREATE TABLE %s (k text, c frozen<map<timeuuid, text>>, PRIMARY KEY (k, c));");
```
adelapena: The fix is for both maps and sets but the tests only exercise maps. I think we should add equivalent tests for sets. for example:
```java
@Test
public void testInsertingSetDataWithParameterizedQueriesIsKeyOrderIndependent() throws Throwable
{
    UUID uuid1 = UUIDs.timeBased();
    UUID uuid2 = UUIDs.timeBased();
    createTable("CREATE TABLE %s (k text, c frozen<set<timeuuid>>, PRIMARY KEY (k, c));");
    // We use `executeNet` in this test because `execute` passes parameters through CqlTester#transformValues(), which calls
    // AbstractType#decompose() on the value, which "fixes" the map order, but wouldn't happen normally.
    executeNet("INSERT INTO %s (k, c) VALUES ('0', ?)", set(uuid1, uuid2));
    executeNet("INSERT INTO %s (k, c) VALUES ('0', ?)", set(uuid2, uuid1));
    beforeAndAfterFlush(() -> {
        assertRowCountNet(executeNet("SELECT * FROM %s WHERE k='0' AND c={" + uuid1 + ", " + uuid2 + '}'), 1);
        assertRowCountNet(executeNet("SELECT * FROM %s WHERE k='0' AND c={" + uuid2 + ", " + uuid1 + '}'), 1);
        assertRowCountNet(executeNet("SELECT * FROM %s WHERE k='0' AND c={" + uuid1 + ", " + uuid2 + '}'), 1);
        assertRowCountNet(executeNet("SELECT * FROM %s WHERE k='0' AND c={" + uuid2 + ", " + uuid1 + '}'), 1);
    });
}

@Test
public void testInsertingSetDataWithParameterizedQueriesIsKeyOrderIndependentWithSelectSlice() throws Throwable
{
    UUID uuid1 = UUIDs.timeBased();
    UUID uuid2 = UUIDs.timeBased();
    createTable("CREATE TABLE %s (k text, c frozen<set<timeuuid>>, PRIMARY KEY (k, c));");
    // We use `executeNet` in this test because `execute` passes parameters through CqlTester#transformValues(), which calls
    // AbstractType#decompose() on the value, which "fixes" the map order, but wouldn't happen normally.
    beforeAndAfterFlush(() -> {
        executeNet("INSERT INTO %s (k, c) VALUES ('0', ?)", set(uuid2, uuid1));
        assertRowsNet(executeNet("SELECT k, c[" + uuid2 + "] from %s"), row("0", uuid2));
        assertRowsNet(executeNet("SELECT k, c[" + uuid1 + "] from %s"), row("0", uuid1));
    });
}
```

test/unit/org/apache/cassandra/cql3/validation/operations/InsertTest.java
```diff
@@ -247,4 +252,42 @@ public void testCKInsertWithValueOver64K() throws Throwable
...
+        // We use `executeNet` in this test because `execute` passes parameters through CqlTester#transformValues(), which calls
+        // AbstractType#decompose() on the value, which "fixes" the map order, but wouldn't happen normally.
+        beforeAndAfterFlush(() -> {
+            executeNet("INSERT INTO %s (k, c) VALUES ('0', ?)", buildInsertionOrderedMap(uuid2, "2", uuid1, "1"));
+            // This works before the fix because uuid2 was inserted/serialized first
+            assertRowsNet(executeNet("SELECT k, c[" + uuid2 + "] from %s"), row("0", "2"));
+            // This fails because uuid1 is < uuid2, but is is serialized after,
+            // we short-circuit the projection once we think we're past the potential map key.
+            // Therefore, the query ends up returning `null` for the map value
```
adelapena: I would either update or remove the comments of the form "this works"/"this fails". These comments are useful in the context of presenting this bugfix but, once we had fixed the bug, the comments are obsolete because the bug is already fixed and the query doesn't fail anymore.

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -2351,4 +2359,12 @@ public boolean equals(Object o)
                 && Objects.equal(password, u.password);
         }
     }
+
+    protected static <K,V> Map<K,V> buildInsertionOrderedMap(K k1, V v1, K k2, V v2)
```
adelapena: We can use the already existent `CQLTester#map(Object...)` instead of adding a new method.

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
```diff
@@ -947,6 +952,180 @@ public void testDateType() throws Exception
...
+        writer.close();
+        loadSSTables(dataDir, keyspace);
+
+        UntypedResultSet rs = QueryProcessor.executeInternal("SELECT * FROM " + qualifiedTable + ";");
+        assertEquals(2, rs.size());
+        Map<String, Map<TimeUUID, Integer>> map3 = StreamSupport.stream(rs.spliterator(), false)
+                                                                .collect(Collectors.toMap(r -> r.getString("k"),
+                                                                                          r -> r.getFrozenMap("c",
+                                                                                                              TimeUUIDType.instance,
+                                                                                                              Int32Type.instance)));
```
adelapena: What's the purpose of this third map?

test/unit/org/apache/cassandra/cql3/CQLTester.java
JeetKunDoug: What do you think about moving the implementation of `map` in this (and fixing up the implementation to take an arbitrary number of parameters), and then delegating `map` to this. The reason I say this is that this method is explicit in its intention to build a specifically-ordered map (which we need for these tests), while `map` doesn't document any contract about the kind of map it produces, even though it happens to do what we need?

test/unit/org/apache/cassandra/cql3/validation/operations/InsertTest.java
JeetKunDoug: I actually originally split the two tests out specifically so we could test inserting with parameterized queries and selecting with parameterized queries specifically (note that each test tests parameterizing a different query). Without this, you could theoretically break one side without being able to test it on the other side.
I'm happy to move it to `CollectionsTest.java` if you feel it's a more appropriate place for them to live, but I think it's important to make sure there are two different, distinct tests so we cover both.

test/unit/org/apache/cassandra/cql3/validation/operations/InsertTest.java
JeetKunDoug: Interestingly, sets already worked because they already used `SortedSet` everywhere... that said, happy to add some tests if we want to make sure we don't later _break_ sets somehow... will wait on the quesiton on where to put them (here or Collections.java) before implementing though.

test/unit/org/apache/cassandra/cql3/validation/operations/InsertTest.java
JeetKunDoug: Fair... will remove/rewrite them (I think it's worth at least documenting why we try to individually select out each field).

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
JeetKunDoug: Left-over from some previous asserts I later realized weren't necessary... will remove.

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -1429,6 +1430,13 @@ protected void assertRowsNet(ProtocolVersion protocolVersion, ResultSet result,
                                         rows.length>i ? "less" : "more", rows.length, i, protocolVersion), i == rows.length);
     }
 
+    protected void assertRowCountNet(ResultSet r1, int expectedCount)
+    {
+        Assert.assertFalse("Received a null resultset when expected count was > 0", expectedCount > 0 && r1 == null);
+        final int actualRowCount = Iterables.size(r1);
```
maedhroz: super nit: could ditch the `final`

test/unit/org/apache/cassandra/cql3/CQLTester.java
maedhroz: Why don't we just have `CQLTester#map()` return a `LinkedHashMap`, codifying the behavior it already exhibits?

test/unit/org/apache/cassandra/cql3/validation/operations/InsertTest.java
maedhroz: Fair. What I usually do is tell the story. For example...

> Before CASSANDRA-17623, this failed, because uuid1 is < uuid2 but was serialized after...

test/unit/org/apache/cassandra/cql3/CQLTester.java
JeetKunDoug: I suppose, to me, unless the method name/comments/something explicitly say _why_ it's returning a LinkedHashMap, someone else could easily come along next year and go "huh, this new Java 22 thing would totally replace this method, so I may as well just use it instead - no reason we have to use a LinkedHashMap here anyway..." I suppose we could document that we expect the return value from `map` to be iterable in insertion order via code comments as well, but then you can never change `map` (which otherwise seems really generic and has no need for in-order iteration) without then adding this new method with a different behavior, and then figuring out where it would break tests. If we have a method (maybe rename this one to `inOrderMap`?) that `map` delegates to, if we ever want to touch `map` we can just split it out without breaking the original purpose of having the map be explicitly an in order traversable map?

maedhroz (approved):
LGTM, assuming we resolve @adelapena 's nits.

test/unit/org/apache/cassandra/cql3/CQLTester.java
maedhroz: I'd be cool w/ just renaming `map()` -> `linkedHashMap()` (that returns a `LinkedHashMap`) and then adding back a `map()` method, so we can avoid renaming for dozens of existing call-sites.

test/unit/org/apache/cassandra/cql3/validation/operations/InsertTest.java
JeetKunDoug: Moving all of these tests, plus the new ones for sets, into CollectionTests.

test/unit/org/apache/cassandra/cql3/validation/operations/InsertTest.java
JeetKunDoug: Added a narrative at the top of the new set of tests in CollectionsTest and a marker at the end of those tests specifying they were all for CASSANDRA-17623, which also allowed me to remove some of the duplicated comments in the tests themselves.

test/unit/org/apache/cassandra/cql3/CQLTester.java
JeetKunDoug: Did what @maedhroz suggested, and added a comment above `linkedHashMap()` to explain that some tests require in-order enumeration of maps so it make sense to someone who sees the two and thinks "I'll just inline this" in case we later want to, for example, add a `sortedHashMap` method or something else more specific.

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -2120,13 +2128,26 @@ protected Object set(Object...values)
         return ImmutableSet.copyOf(values);
     }
 
+    protected LinkedHashSet<Object> linkedHashSet(Object...values)
```
adelapena, line 2132: This could have a comment/JavaDoc analogous to the one in `linkedHashMap`

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -2351,4 +2372,5 @@ public boolean equals(Object o)
                 && Objects.equal(password, u.password);
         }
     }
+
```
adelapena: Nit: unneeded blank line

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
```diff
@@ -45,6 +46,9 @@
 import org.apache.cassandra.cql3.functions.types.*;
 import org.apache.cassandra.db.Keyspace;
 import org.apache.cassandra.db.commitlog.CommitLog;
+import org.apache.cassandra.db.marshal.Int32Type;
+import org.apache.cassandra.db.marshal.TimeUUIDType;
```
adelapena: Nit: unneeded imports

test/unit/org/apache/cassandra/cql3/validation/entities/CollectionsTest.java
```diff
@@ -1980,4 +1981,99 @@ public void testSelectionOfEmptyCollections() throws Throwable
             assertRows(execute("SELECT m['0'..'1']['3'..'5'], s[0..1][3..5] FROM %s WHERE k = 3"), row(null, null));
         });
     }
+
+    // Tests for CASSANDRA-17623
+    // Before CASSANDRA-17623, parameterized queries with maps as values would fail because frozen maps were
+    // required to be sorted by the sort order of their key type, but weren't always sorted correctly.
+    // Also adding tests for Sets, which did work because they always used SortedSet, to make sure this behavior is maintained.
+    // We use `executeNet` in these tests because `execute` passes parameters through CqlTester#transformValues(), which calls
+    // AbstractType#decompose() on the value, which "fixes" the map order, but wouldn't happen normally.
```
adelapena: Nit: this could be a block comment:
```suggestion
    /*
     Tests for CASSANDRA-17623.
     Before CASSANDRA-17623, parameterized queries with maps as values would fail because frozen maps were required to 
     be sorted by the sort order of their key type, but weren't always sorted correctly. Also adding tests for Sets, 
     which did work because they always used SortedSet, to make sure this behavior is maintained.
     We use `executeNet` in these tests because `execute` passes parameters through CqlTester#transformValues(), which 
     calls AbstractType#decompose() on the value, which "fixes" the map order, but wouldn't happen normally.
     */

```

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
```diff
@@ -947,6 +952,172 @@ public void testDateType() throws Exception
...
+        {
+            LinkedHashMap<String, String> map = new LinkedHashMap<>();
+            map.put("b_key", "zv" + i);
+            map.put("a_key", "av" + i);
+            writer.addRow(String.valueOf(i), map);
+        }
+        writer.close();
+        loadSSTables(dataDir, keyspace);
+
+        UntypedResultSet rs = QueryProcessor.executeInternal("SELECT * FROM " + qualifiedTable + ";");
```
adelapena, line 986: Nit: we don't need the `;` at the end of the queries. It's quite common to see it in tests, so feel free to ignore it if you prefer it this way.

adelapena (approved):
Changes look good to me, +1. Just commented a couple of nits. Now we need PRs/patches for the older branches and CI runs.

PR #1622, opened 2022-05-13
CASSANDRA-17577 add new CQLSH command SHOW REPLICAS

bschoening:
Note, for naming I used the Cassandra python driver convention of 'replicas'.  Replicas is a frequently used term. Nodetool uses 'endpoints', but this is otherwise not a commonly used term in Cassandra in my experience.

### DIFF
--- a/bin/cqlsh.py
+++ b/bin/cqlsh.py
@@ -33,7 +33,6 @@ import traceback
 import warnings
 import webbrowser
 from contextlib import contextmanager
-from cassandra.metadata import Murmur3Token
 from glob import glob
 from io import StringIO
 from uuid import UUID
@@ -601,11 +600,12 @@ class Shell(cmd.Cmd):
     def show_session(self, sessionid, partial_session=False):
         print_trace_session(self, self.session, sessionid, partial_session)
 
-    def show_replicas(self, token, keyspace=None):
+    def show_replicas(self, token_value, keyspace=None):
         ks = self.current_keyspace if keyspace is None else keyspace
-        nodes = self.conn.metadata.token_map.get_replicas(self.current_keyspace, Murmur3Token(token))
+        token_map = self.conn.metadata.token_map
+        nodes = token_map.get_replicas(ks, token_map.token_class(token_value))
         addresses = [x.address for x in nodes]
-        print(f"Replicas for token {token} are {addresses}")
+        print(f"Replicas {addresses}")
 
     def get_connection_versions(self):
         result, = self.session.execute("select * from system.local where key = 'local'")
--- a/bin/cqlsh.py
+++ b/bin/cqlsh.py
@@ -605,7 +605,7 @@ class Shell(cmd.Cmd):
         token_map = self.conn.metadata.token_map
         nodes = token_map.get_replicas(ks, token_map.token_class(token_value))
         addresses = [x.address for x in nodes]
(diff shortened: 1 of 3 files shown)


bin/cqlsh.py
```diff
@@ -600,6 +601,12 @@ def show_version(self):
     def show_session(self, sessionid, partial_session=False):
         print_trace_session(self, self.session, sessionid, partial_session)
 
+    def show_replicas(self, token, keyspace=None):
+        ks = self.current_keyspace if keyspace is None else keyspace
+        nodes = self.conn.metadata.token_map.get_replicas(self.current_keyspace, Murmur3Token(token))
```
smiklosovic: should not this be `ks` instead of `self.current_keyspace` in `get_replicas` method?

bschoening: Yes, I'll fix that. Thanks!

bin/cqlsh.py
```diff
@@ -600,6 +601,12 @@ def show_version(self):
...
+        addresses = [x.address for x in nodes]
+        print(f"Replicas for token {token} are {addresses}")
```
smiklosovic: I do not think we need `Replicas for token ... are ...`. That is just redundant information. It is also harder to parse in scripts if we use `cqlsh -e 'show replicas 95'`.

doc/modules/cassandra/pages/tools/cqlsh.adoc
```diff
@@ -167,7 +167,7 @@ Example:
 [source,none]
 ----
 cqlsh> SHOW VERSION
-[cqlsh 5.0.1 | Cassandra 3.8 | CQL spec 3.4.2 | Native protocol v4]
+[cqlsh 6.1.0 | Cassandra 4.0.1 | CQL spec 3.4.5 | Native protocol v5]
```
smiklosovic, line 170: why these versions? Especially for Cassandra.
edit: I do not know what the version here should be

bin/cqlsh.py
```diff
@@ -33,6 +33,7 @@
 import warnings
 import webbrowser
 from contextlib import contextmanager
+from cassandra.metadata import Murmur3Token
```
smiklosovic: @bschoening you need to move this lower around line 135, this is too soon to import anything from cassandra as it is not loaded yet.

doc/modules/cassandra/pages/tools/cqlsh.adoc
bschoening: It's an example, and its what I had installed at the time.

bin/cqlsh.py
```diff
@@ -600,6 +600,13 @@ def show_version(self):
     def show_session(self, sessionid, partial_session=False):
         print_trace_session(self, self.session, sessionid, partial_session)
 
+    def show_replicas(self, token_value, keyspace=None):
+        ks = self.current_keyspace if keyspace is None else keyspace
+        token_map = self.conn.metadata.token_map
```
smiklosovic, line 605: @bschoening very nice!

bin/cqlsh.py
bschoening: Ok, changed.

bin/cqlsh.py
bschoening: I was able to remove the import entirely by using self.conn.metadata.token_map.

PR #1632, opened 2022-05-17
CASSANDRA-17602 4.0 remove ssl storage port from sstableloader

smiklosovic:


src/java/org/apache/cassandra/tools/BulkLoadConnectionFactory.java
```diff
@@ -28,26 +28,24 @@
...
-    private final boolean outboundBindAny;
-    private final int secureStoragePort;
+    private final int storagePort;
     private final EncryptionOptions.ServerEncryptionOptions encryptionOptions;
 
-    public BulkLoadConnectionFactory(int secureStoragePort, EncryptionOptions.ServerEncryptionOptions encryptionOptions, boolean outboundBindAny)
+    public BulkLoadConnectionFactory(EncryptionOptions.ServerEncryptionOptions encryptionOptions, int storagePort)
     {
-        this.secureStoragePort = secureStoragePort;
+        this.storagePort = storagePort;
```
smiklosovic, line 36: we need to acutally propagate storagePort, this comes from the cli of sstableloader.

src/java/org/apache/cassandra/tools/BulkLoadConnectionFactory.java
```diff
@@ -28,26 +28,24 @@
...
     {
-        // Connect to secure port for all peers if ServerEncryptionOptions is configured other than 'none'
-        // When 'all', 'dc' and 'rack', server nodes always have SSL port open, and since thin client like sstableloader
-        // does not know which node is in which dc/rack, connecting to SSL port is always the option.
+        // storage port can handle both encrypted and unencrypted traffic from 4.0
+        // so from sstableloader's point of view we can use just storage port for both cases
+
+        template = template.withConnectTo(template.to.withPort(storagePort));
 
         if (encryptionOptions != null && encryptionOptions.internode_encryption != EncryptionOptions.ServerEncryptionOptions.InternodeEncryption.none)
```
smiklosovic, line 47: we fill template with encryptionOptions if there are any

test/distributed/org/apache/cassandra/distributed/test/SSTableLoaderEncryptionOptionsTest.java
```diff
@@ -112,7 +113,7 @@ public void bulkLoaderCannotAgreeOnClientTLSProtocol()
                                                             "--truststore-password", validTrustStorePassword,
                                                             "test/data/legacy-sstables/na/legacy_tables/legacy_na_clust");
         assertNotEquals(0, tool.getExitCode());
-        assertTrue(tool.getStdout().contains("SSLHandshakeException"));
+        assertTrue(tool.getStdout().contains("com.datastax.driver.core.exceptions.TransportException"));
```
driftx: This doesn't make sense, the driver doesn't connect to the storage port.

smiklosovic: @driftx this code belongs to the test method `bulkLoaderCannotAgreeOnClientTLSProtocol` which specified `--ssl-protocol` to be `TLSv1` which is invalid as in config it is `TLSv1.2` for _client_ encryption. Bulkloader internally connects via CQL to a node, fetching metadata, getting token ranges and so on before actually loading sstables. So this exception comes from the fact that we have not been able to talk to that node via CQL because we have set invalid `--ssl-protocol` flag. Basically, it fails way sooner it gets to sstable loading and this test verifies that setting invalid ssl protocol to connect through fails and driver throws that exception.

smiklosovic: What is puzzling is that when I checkout cassandra-4.0 as is now, without this patch applied, running this test in IDEA, locally, fails as well, huh. It is actually good question why do I need to suddenly change this assert to check if the log contains some other exception. Should be the same as it was before. But this test fails even as-is in cassandra-4.0.

Would you do me a favour and run this test yourself locally on cassandra-4.0 to see if it passes there for you? Thanks a lot.

driftx: This fails for me too, both in isolation and in the suite.  It's a bit worrisome this doesn't appear in jenkins or circle

smiklosovic: Thanks, interesting, I will try to poke around whats up.

smiklosovic: I think that catching whatever exception we get from initialisation and wrapping it into ConfigurationException and throwing that one would do the job. At least we would be sure that it throws one exception type no matter what is the cause.

PR #1639, opened 2022-05-20
Cassandra 17647

jmckenzie-dev:


src/java/org/apache/cassandra/cql3/statements/schema/AlterKeyspaceStatement.java
```diff
@@ -76,6 +78,9 @@ public Keyspaces apply(Keyspaces schema)
 
         KeyspaceMetadata newKeyspace = keyspace.withSwapped(attrs.asAlteredKeyspaceParams(keyspace.params));
 
+        if (attrs.getReplicationStrategyClass() != null && attrs.getReplicationStrategyClass().equals(SimpleStrategy.class.getSimpleName()))
```
iamaleksey, line 81: So, the `CREATE` case is pretty obvious. With `ALTER`, however, I wonder - should we allow altering the replication factor on a table that has had `SimpleStrategy` enabled prior to its prohibition by guardrails?

In other words, if current strategy is `SimpleStrategy` should we still allow such an alter? Some system keyspaces are created with `SimpleStrategy` - and are expected to be RF-changed by the user to a cluster-size appropriate RF. Should we allow upping it if the guardrail is on - while keeping it on `SimpleStrategy`?

I'm leaning towards interpreting the guardrail on as "don't allow `ALTER` to *switch* to `SimpleStrategy`", myself. WDYT?

iamaleksey (approved):
+1

PR #1640, opened 2022-05-23
CASSANDRA-17616 4.1: Add in Circle CI an option to repeat jvm dtests with vnodes

adelapena:


.circleci/config-2_1.yml
```diff
@@ -1140,6 +1144,7 @@ jobs:
           target: test-jvm-dtest-some
           class: ${REPEATED_JVM_UPGRADE_DTEST_CLASS}
           methods: ${REPEATED_JVM_UPGRADE_DTEST_METHODS}
+          vnodes: "false"
```
bereng, line 1149: Is this intentionally false?

adelapena: Yes, it is intentional. The repeated jobs try to mimic the configuration of the regular jobs, so we can reproduce failures in them. The only job for running all JVM upgrade tests (`j8_jvm_upgrade_dtests`) doesn't use virtual nodes, so the task for repeating individual upgrade tests doesn't use them either. 

If at some point we added a `j8_jvm_upgrade_dtests_vnode` task, analogous to `j8_jvm_dtests_vnode`, we would need to add a new `REPEATED_JVM_UPGRADE_DTEST_VNODES` property and pass it here.

PR #1642, opened 2022-05-23
CASSANDRA-17656: jvm-dtest upgrade tests run all supported pairs of upgrades between from/to but does not actually test all patches from/to

dcapwell:


test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeFrom3ReplicationTest.java
```diff
@@ -18,21 +18,65 @@
...
-{
-    @Test
-    public void testSimpleStrategy30to3X() throws Throwable
-    {
-        testSimpleStrategy(v30, v3X);
-    }
+import static org.apache.cassandra.distributed.shared.AssertUtils.assertRows;
+import static org.apache.cassandra.distributed.shared.AssertUtils.row;
 
+public class MixedModeFrom3ReplicationTest extends UpgradeTestBase
```
dcapwell, line 31: `MixedModeReplicationTestBase` only was extended by this class, so was simpler to just inline all the logic here...

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -153,20 +189,29 @@ public TestCase upgradesFrom(Semver from)
...
-                .filter(upgradePath -> (upgradePath.left.compareTo(from) >= 0 && upgradePath.right.compareTo(to) <= 0))
-                .forEachOrdered(upgradePath ->
-                {
-                    this.upgrade.add(
-                            new TestVersions(versions.getLatest(upgradePath.left), versions.getLatest(upgradePath.right)));
-                });
+            List<TestVersions> upgrades = findUpgradePaths(from, to);
+            // order by most direct to largest path
+            Collections.sort(upgrades, Comparator.comparingInt(a -> a.upgrade.size()));
+            logger.info("Adding upgrades of\n{}", upgrades.stream().map(TestVersions::toString).collect(Collectors.joining("\n")));
```
dcapwell: this isn't needed... I did this as I tried to make tests dynamic rather than running everything as a single test... I kept it assuming that if an issue exists it is likely to be in the direct path rather than the longer path (unless dealing with old data problems like 2.x -> 4.x)

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -153,20 +189,29 @@ public TestCase upgradesFrom(Semver from)
...
+                                          .collect(Collectors.toList());
+        }
+
         /** Will test this specific upgrade path **/
         public TestCase singleUpgrade(Semver from)
         {
-            this.upgrade.add(new TestVersions(versions.getLatest(from), versions.getLatest(CURRENT)));
+            TestVersions target = new TestVersions(versions.getLatest(from), Arrays.asList(versions.getLatest(CURRENT)));
+            Set<TestVersions> supported = new HashSet<>(findUpgradePaths(from, CURRENT));
+            if (!supported.contains(target))
```
dcapwell: the thought process is that there may be cases in the future where from can't directly go to to, such as 2.1 -> 4.2.... this just makes sure we detect that when the test is running

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -226,24 +271,40 @@ public void run() throws Throwable
...
+                                triggerGC();
+                                cluster.get(n).setVersion(nextVersion);
+                                runBeforeNodeRestart.run(cluster, n);
+                                cluster.get(n).startup();
+                                runAfterNodeUpgrade.run(cluster, n);
+                            }
+
+                            runAfterClusterUpgrade.run(cluster);
+
+                            cluster.checkAndResetUncaughtExceptions();
```
dcapwell, line 344: if logs have a bad issue, trigger early so we don't deal with "there was an error somewhere"... can still happen after the fact... just trying to be nice!

dcapwell:
spoke with @jonmeredith in slack, I will revert back to `upgradeFrom(v30)`.  He pointed out that we really want to test all upgrade paths, but don't have a clear solution on how to do that given CI and JVM Metaspace limitations (testing out POC but likely to punt if not working by EOD)

dcapwell:
I gave up on trying to make the tests better... we really need a template system or code generation system where we create 1 class for each path...  given how we currently do this we need `*Test.java` files for each, so something would need to know how and generate it!  We could do this via a script, but then we need this to work on all environments... we could do this with an annotation processor, but need to make things more complex at build time since annotation processors need to be compiled before the tests compile

dcapwell:
looking into `org.apache.cassandra.distributed.upgrade.CompactStorageUpgradeTest#compactStoragePagingTest`, not sure why it OOMs when it matches the same pattern as other tests...

If you disable the queries (that do paging) then it passes... so something must leak there?

dcapwell:
just switched to `.singleUpgrade` for CompactStorageUpgradeTest; I think a leak is happening but don't have the bandwidth to look into that...

test/distributed/org/apache/cassandra/distributed/upgrade/CompactStorageUpgradeTest.java
```diff
@@ -37,7 +37,7 @@ public void compactStorageColumnDeleteTest() throws Throwable
         new TestCase()
         .nodes(2)
         .nodesToUpgrade(2)
-        .upgradesFrom(v30)
+        .singleUpgrade(v30)
```
dcapwell: there are too many tests to try multiple upgrade paths from 3.0... so avoid and just run 3.0 -> trunk

jonmeredith (approved):

test/distributed/org/apache/cassandra/distributed/upgrade/CompactStorageUpgradeTest.java
michaelsembwever: upgrading from 3.11 is (a slightly more) recommended upgrade path, not just 3.0

the upgrade tests need to test our supported and recommended upgrade paths, IMHO.

see here: https://lists.apache.org/thread/odk114pfpzpwj81sj2vdzmzt5rgn74wk

dcapwell: the function mutates a list, so we could always do

```
.singleUpgrade(v30)
.singleUpgrade(v3x)
```

so we test both

dcapwell: updated, we now have

```
INFO  [main] 2022-05-26 11:29:59,284 UpgradeTestBase.java:207 - Adding upgrade of 3.0.28 -> [4.2]
INFO  [main] 2022-05-26 11:29:59,285 UpgradeTestBase.java:207 - Adding upgrade of 3.11.14 -> [4.2]
```

michaelsembwever: wfm. (presuming `[4.2]` becomes whatever current is, so long as the from-to remains something found in the supported_paths)

test/unit/org/apache/cassandra/utils/SimpleGraph.java
```diff
@@ -0,0 +1,135 @@
...
+    private final ImmutableMap<V, ImmutableSet<V>> edges;
+
+    private SimpleGraph(ImmutableMap<V, ImmutableSet<V>> edges)
+    {
+        if (edges == null || edges.isEmpty())
+            throw new AssertionError("Edges empty");
+        this.edges = edges;
+    }
+
+    public static <T> SimpleGraph<T> of(T... values)
```
michaelsembwever: (trivial) for readability i would have passed in pairs (it confused me reading the code where this was used and i had to go here and read it to understand). i understand that it would make the calling code more verbose 🤷🏻‍♀️

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -153,20 +184,40 @@ public TestCase upgradesFrom(Semver from)
...
+                // from previous versions; so special case that
+                for (Semver end : vertices.subSet(from, false, to, false)) //from->to included above, so don't need to readd
                 {
-                    this.upgrade.add(
-                            new TestVersions(versions.getLatest(upgradePath.left), versions.getLatest(upgradePath.right)));
-                });
+                    if (SUPPORTED_UPGRADE_PATHS.hasEdge(CURRENT, end))
+                        upgrade.add(new TestVersions(versions.getLatest(CURRENT), Collections.singletonList(versions.getLatest(end))));
+                }
+            }
```
michaelsembwever: curious if this can be simplified to
```

            List<TestVersions> upgrade = new ArrayList<>();
            NavigableSet<Semver> vertices = sortedVertices(SUPPORTED_UPGRADE_PATHS);
            for (Semver vertex : vertices.subSet(from, true, to, true))
            {
                if (SUPPORTED_UPGRADE_PATHS.hasEdge(CURRENT, vertex))
                    upgrade.add(new TestVersions(versions.getLatest(CURRENT), Collections.singletonList(versions.getLatest(vertex))));
                
                if (SUPPORTED_UPGRADE_PATHS.hasEdge(vertex, CURRENT))
                    upgrade.add(new TestVersions(versions.getLatest(vertex), Collections.singletonList(versions.getLatest(CURRENT))));
            }
```

test/unit/org/apache/cassandra/utils/SimpleGraph.java
dcapwell: IMO pairs don't improve things, as it would be simpler to work with the builder directly... I can always switch back to builder

dcapwell: moved this to SimpleGraphTest to keep that code smaller, and switched upgrade base to use builder + addEdge

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
dcapwell: doesn't work given the API and would logically break `org.apache.cassandra.distributed.upgrade.MixedModeFrom3LoggedBatchTest`

if CURRENT=4.2, from=3.0, and to=4.0, then CURRENT isn't involved in the test (which the API allows and some tests are doing...)...

this logic would have `3.0 -> 4.0` do the following

```
3.0 -> 4.2
3.11 -> 4.2
4.0 -> 4.2
```

dcapwell: moved this convo back to slack as there isn't agreement on the desired semantic... blocked until we agree https://the-asf.slack.com/archives/CK23JSY2K/p1654110195106499?thread_ts=1653327219.358099&cid=CK23JSY2K

this is getting more complex, so need feedback from all.  What "should" be the test cases for the following cases
* CURRENT=3.0, from=3.0, to=4.2
* CURRENT=4.0, from=3.0, to=4.2
* CURRENT=4.2, from=3.0, to=4.2

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -145,28 +179,77 @@ public TestCase nodes(int nodeCount)
...
         public TestCase upgrades(Semver from, Semver to)
         {
-            SUPPORTED_UPGRADE_PATHS.stream()
-                .filter(upgradePath -> (upgradePath.left.compareTo(from) >= 0 && upgradePath.right.compareTo(to) <= 0))
-                .forEachOrdered(upgradePath ->
-                {
-                    this.upgrade.add(
-                            new TestVersions(versions.getLatest(upgradePath.left), versions.getLatest(upgradePath.right)));
-                });
+            Assume.assumeTrue("Unable to do upgrades(" + from + ", " + to + "); does not contain CURRENT=" + CURRENT, contains(from, to, CURRENT));
```
jonmeredith, line 231: worth assuming that `from < to`?

jonmeredith (approved):

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
dcapwell: `assume` causes tests to skip, `from > to` is a bug and should fail; currently the subSet call checks that and fails

dcapwell: `java.util.TreeMap.NavigableSubMap#NavigableSubMap` blocks this

```
if (!fromStart && !toEnd) {
                if (m.compare(lo, hi) > 0)
                    throw new IllegalArgumentException("fromKey > toKey");
```

michaelsembwever (approved):

PR #1645, opened 2022-05-25
 CASSANDRA-17661 Adding support to perform certificate based internode authentication

jyothsnakonisa:


src/java/org/apache/cassandra/auth/IInternodeAuthenticator.java
```diff
@@ -35,10 +36,51 @@
...
+     * Decides whether a peer is allowed to connect to this node.
+     * If this method returns false, the socket will be immediately closed.
+     *
+     * @param certificates client certificates
+     * @return true if the connection should be accepted, false otherwise.
+     */
+    default boolean authenticate(Certificate[] certificates)
+    {
+        return false;
+    }
```
jonmeredith: I don't think you need this new method. It would be up to the implementor to write an `authenticator` method that checks the certificates if that's what they need.

src/java/org/apache/cassandra/auth/IInternodeAuthenticator.java
```diff
@@ -35,10 +36,51 @@
...
+     * @param remorePort port of the connecting node.
+     * @param certificates client certificates
+     * @return true if the connection should be accepted, false otherwise.
+     */
+    default boolean authenticate(InetAddress remoteAddress, int remorePort, Certificate[] certificates)
+    {
+        if (certificates != null)
+        {
+            return authenticate(certificates);
+        }
```
jonmeredith: I think this can be replaced with a comment saying something like `Default implementation calls the deprecated authenticate by IP and port method`.

src/java/org/apache/cassandra/auth/IInternodeAuthenticator.java
```diff
@@ -35,10 +36,51 @@
...
+    /**
+     * Decides whether a peer is allowed to connect to this node.
+     * If this method returns false, the socket will be immediately closed.
+     *
+     * @param remoteAddress ip address of the connecting node.
+     * @param remorePort port of the connecting node.
+     * @param certificates client certificates
+     * @return true if the connection should be accepted, false otherwise.
+     */
+    default boolean authenticate(InetAddress remoteAddress, int remorePort, Certificate[] certificates)
```
jonmeredith: The other piece of information that could be useful is whether this is an inbound or outbound internode connection being authenticated. Organizations may have different standards for the two types of collection and/or create different audit messages depending on the connection. 

Maybe an `enum` argument for connections direction `INTERNODE_INBOUND`, `INTERNODE_OUTBOUND` ?

src/java/org/apache/cassandra/auth/IInternodeAuthenticator.java
```diff
@@ -35,10 +36,51 @@
...
      * @throws ConfigurationException when there is a configuration error.
      */
     void validateConfiguration() throws ConfigurationException;
+
+    /**
+     * Setup is called once upon system startup to initialize the IAuthenticator.
+     *
+     * For example, use this method to create any required keyspaces/column families.
+     */
+    default void setup()
```
jonmeredith: If somebody wanted to centralize all their client `IAuthenticator` and internode auth `IInternodeAuthenticator` by implementing both in the same class the compiler would forbid with an `inherits unrelated defaults for setup()` error. WDYT about renaming this `setupInternode` to make that possible.

src/java/org/apache/cassandra/net/OutboundConnectionInitiator.java
```diff
@@ -211,6 +204,15 @@ public void initChannel(SocketChannel channel) throws Exception
...
                 pipeline.addFirst("ssl", sslHandler);
             }
+            else
+            {
+                // If a connection is SSL based we don't need to make additional authentication when making a client
+                // connection. Trusted roots in the client's truststore would suffice for authentication.
+                //
+                // Outbound connection is a client connection we don't need to extract identity from server certificate and authenticate.
+                // Add this ServerAuthenticationHandler only when encryption is not enabled just for backward compatibility.
+                pipeline.addLast("server-authentication", new ServerAuthenticationHandler());
```
jonmeredith: I think you should call the authenticator for outbound too. This changes the behavior of always authenticating on outbound regardless of SSL.

It may not be true for all operators that the trusted root check is sufficient and they may also want to check the outbound IP addresses too, the age of the certificate or log the authentication attempt. Adding the direction to the `authenticate` method would make it easy to implement different behaviors on inbound/outbound.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -40,6 +48,103 @@
...
+
+                OutboundConnections outbound = getOnlyElement(MessagingService.instance().channelManagers.values());
+                assertTrue(outbound.small.isConnected() || outbound.large.isConnected() || outbound.urgent.isConnected());
+            };
+
+            cluster.get(1).runOnInstance(runnable);
+            cluster.get(2).runOnInstance(runnable);
+        }
+    }
+
```
jonmeredith, line 181: Could you add a test that checks the `MessagingService.instance().interruptOutbound(settings.to);` functionality
I think you could add an authenticator that works on the first attempt, but fails all subsequent attempts. You could use the `internode_outbound` table to confirm no connections were made.  Might be nice to add that check to the other tests just to make sure the connection hadn't been established then immediately dropped.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -40,6 +48,103 @@
...
+        try (Cluster cluster = builder.start())
+        {
+            executorService.submit(() -> {
+                openConnections(cluster);
+            });
+
+            /*
+             * instance (1) should connect to instance (2) without any issues;
+             * instance (2) should connect to instance (1) without any issues.
+             */
```
jonmeredith: Comment does not match code.

How do you know the failed connection was due to the authenticator failing the attempt rather than another issue? You could add an @Shared class that either holds counters or a log of authentication attempts in the test.

jonmeredith (commented):
Looking in good shape already. I've made some suggestions for improvements to the interface, and I think we still need to call it on outbound for the reasons given in comments.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
jyothsnakonisa: Updated the test to verify that.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
jyothsnakonisa: Added tests to verify this codepath.

src/java/org/apache/cassandra/auth/AllowAllInternodeAuthenticator.java
```diff
@@ -30,6 +31,11 @@ public boolean authenticate(InetAddress remoteAddress, int remotePort)
         return true;
     }
 
+    public boolean authenticate(Certificate[] certificates)
+    {
+        return true;
+    }
+
```
yifan-c: The method is not referenced. I think it can be removed.

src/java/org/apache/cassandra/auth/IInternodeAuthenticator.java
```diff
@@ -35,10 +36,53 @@
...
+    /**
+     * Decides whether a peer is allowed to connect to this node.
+     * If this method returns false, the socket will be immediately closed.
+     * <p>
+     * Default implementation calls authenticate method by IP and port method
+     * <p>
+     * 1. If it is IP based authentication ignore the certificates & connectionType parameters in the implementation
+     * of this method.
+     * 2. For certificate based authentication like mTLS, server's identity for outbound connections is verified by the
+     * trusted root certificates in the outbound_keystore. In such cases this method may be overriden to return true
```
yifan-c: typo.
this method may be ~overriden~ overridden ...

src/java/org/apache/cassandra/auth/IInternodeAuthenticator.java
```diff
@@ -35,10 +36,53 @@
...
+     * when certificateType is OUTBOUND, as the authentication of the server happens during SSL Handshake.
+     *
+     * @param remoteAddress  ip address of the connecting node.
+     * @param remorePort     port of the connecting node.
```
yifan-c: typo
```suggestion
     * @param remotePort     port of the connecting node.
```

src/java/org/apache/cassandra/auth/IInternodeAuthenticator.java
```diff
@@ -35,10 +36,53 @@
...
+     */
+    default void setupInternode()
+    {
+
+    }
+
+    /**
+     * Enum that represents connection type of an internode connection.
+     */
+    public enum InternodeConnectionType
```
yifan-c: nit: `public` is redundant since it is defined within an interface.

src/java/org/apache/cassandra/net/CertificateUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+import io.netty.channel.Channel;
+import io.netty.handler.ssl.SslHandler;
+
+/**
+ * Interface that contains certificate utility methods.
+ */
+interface CertificateUtils
+{
+    final Logger logger = LoggerFactory.getLogger(CertificateUtils.class);
```
yifan-c: Hmm... interface fields are `public static final`. You do not need to add `final` here. 

A step back. Do you really want to define an instance field? The logger's bound class would be confusing when the logger is used in the subclasses and it is public unnecessarily. 

I think you just want a Utils class, which provides `certificates(Channel channel)` as a static method. (maybe `extract/getCertificates` is a better name).

src/java/org/apache/cassandra/net/InboundConnectionInitiator.java
```diff
@@ -233,30 +286,6 @@ public void handlerAdded(ChannelHandlerContext ctx) throws Exception
                 failHandshake(ctx);
             }, HandshakeProtocol.TIMEOUT_MILLIS, MILLISECONDS);
```
yifan-c: nit: remove the empty tailing line in the body.

yifan-c: another nit: can you remove `"and authenticate the remote address."` in line 280. The authenticate part has been moved to the new handler.

src/java/org/apache/cassandra/auth/IInternodeAuthenticator.java
```diff
@@ -35,10 +36,53 @@
      */
     boolean authenticate(InetAddress remoteAddress, int remotePort);
```
yifan-c: Wdyt about adding `@Deprecated` for this method? There is not active reference of this method in the production code.

src/java/org/apache/cassandra/net/CertificateUtils.java
```diff
@@ -0,0 +1,56 @@
...
+
+    default Certificate[] certificates(Channel channel)
+    {
+        final SslHandler sslHandler = (SslHandler) channel.pipeline().get("ssl");
```
yifan-c: This method has implicit dependency on the handler name... can you add comments in the both pipeline defining code? Can you also define a constant for `ssl`?

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -40,6 +50,104 @@
...
+            /*
+             * instance (1) should not connect to instance (2) as authentication fails;
+             * instance (2) should not connect to instance (1) as authentication fails.
+             */
+            IIsolatedExecutor.SerializableCallable<Boolean> callable = () ->
+            {
+                InboundMessageHandlers inbound = getOnlyElement(MessagingService.instance().messageHandlers.values());
+                assertEquals(0, inbound.count());
+                OutboundConnections outbound = getOnlyElement(MessagingService.instance().channelManagers.values());
+                assertTrue(!outbound.small.isConnected() && !outbound.large.isConnected() && !outbound.urgent.isConnected());
```
yifan-c, line 79: Please add assert descriptions, especially for `assertTrue/False`

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -40,6 +50,104 @@
...
+
+                // Verify that the failure is due to authentication failure
+                final RejectInboundConnections authenticator = (RejectInboundConnections) DatabaseDescriptor.getInternodeAuthenticator();
+                assertTrue(authenticator.authenticationFailed);
+                return true;
+            };
+
+            // Wait for cluster to get started
+            Thread.sleep(3000);
+            assertTrue(cluster.get(1).callsOnInstance(callable).call());
```
yifan-c: The `callable` always returns `true`. Why are the assertions needed? If the assertions are not needed, what about just `runOnInstance`?

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -40,6 +50,104 @@
...
+            {
+                InboundMessageHandlers inbound = getOnlyElement(MessagingService.instance().messageHandlers.values());
+                assertEquals(0, inbound.count());
+                OutboundConnections outbound = getOnlyElement(MessagingService.instance().channelManagers.values());
+                assertTrue(!outbound.small.isConnected() && !outbound.large.isConnected() && !outbound.urgent.isConnected());
+
+                // Verify that the failure is due to authentication failure
+                final RejectOutboundAuthenticator authenticator = (RejectOutboundAuthenticator) DatabaseDescriptor.getInternodeAuthenticator();
+                assertTrue(authenticator.authenticationFailed);
+                return true;
```
yifan-c: Same as the above comment. The `callable` returns `true` only.

src/java/org/apache/cassandra/auth/IInternodeAuthenticator.java
jyothsnakonisa: We can but I think this method is valid for IP based authenticator.

jonmeredith: I think it would be good to mark it as deprecated. Implementations should move towards `authenticate(InetAddress remoteAddress, int remotePort,
                                  Certificate[] certificates, InternodeConnectionType connectionType)` and just ignoring the certificate/connectionType if not needed.

src/java/org/apache/cassandra/auth/IInternodeAuthenticator.java
```diff
@@ -35,10 +36,53 @@
...
+     */
+    default void setupInternode()
+    {
+
+    }
+
+    /**
+     * Enum that represents connection type of an internode connection.
+     */
+    enum InternodeConnectionType
```
jonmeredith: Would you mind renaming `InternodeConnectionDirection` or something like that as we already have `org.apache.cassandra.net.ConnectionType` and I don't want any confusion.

src/java/org/apache/cassandra/net/InboundConnectionInitiator.java
```diff
@@ -198,6 +204,54 @@ public static ChannelFuture bind(InboundConnectionSettings settings, ChannelGrou
...
+
+        @Override
+        protected void decode(ChannelHandlerContext channelHandlerContext, ByteBuf byteBuf, List<Object> list) throws Exception
+        {
+            // Extract certificates from SSL handler(handler with name "ssl").
+            final Certificate[] certificates = CertificateUtils.certificates(channelHandlerContext.channel());
+            if (!authenticate(channelHandlerContext.channel().remoteAddress(), certificates))
+            {
+                logger.error("Unable to authenticate peer {} for internode authentication", channelHandlerContext.channel());
+                channelHandlerContext.close();
```
jonmeredith: We should make sure any buffered data is consumed here, so handlers further down the pipeline don't have to deal with unauthenticated payloads after this handler is removed.

A bit like `failHandshake()`, but you could use `channel.pipeline().names()` to remove all pipelines after this one (and leave the exception logging intact).

src/java/org/apache/cassandra/net/OutboundConnectionInitiator.java
```diff
@@ -220,6 +216,31 @@ public void initChannel(SocketChannel channel) throws Exception
...
+        protected void decode(ChannelHandlerContext channelHandlerContext, ByteBuf byteBuf, List<Object> list) throws Exception
+        {
+            // Extract certificates from SSL handler(handler with name "ssl").
+            final Certificate[] certificates = CertificateUtils.certificates(channelHandlerContext.channel());
+            if (!settings.authenticator.authenticate(settings.to.getAddress(), settings.to.getPort(), certificates, OUTBOUND))
+            {
+                // interrupt other connections, so they must attempt to re-authenticate
+                MessagingService.instance().interruptOutbound(settings.to);
+                logger.error("authentication failed to " + settings.connectToId());
+                channelHandlerContext.close();
```
jonmeredith: Same note to consume any buffered data.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -40,6 +49,110 @@
...
+                OutboundConnections outbound = getOnlyElement(MessagingService.instance().channelManagers.values());
+                assertTrue(!outbound.small.isConnected() && !outbound.large.isConnected() && !outbound.urgent.isConnected());
+
+                // Verify that the failure is due to authentication failure
+                final RejectInboundConnections authenticator = (RejectInboundConnections) DatabaseDescriptor.getInternodeAuthenticator();
+                assertTrue(authenticator.authenticationFailed);
+            };
+
+            // Wait for cluster to get started
+            Thread.sleep(3000);
```
jonmeredith: Sleeps aren't always reliable especially on CI infra - is there a log message we could wait for instead? Maybe the Reject*Connections.class could log something so you could check it happened.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -155,4 +268,78 @@ private void openConnections(Cluster cluster)
...
+    public static class RejectConnectionsAuthenticator implements IInternodeAuthenticator
+    {
+        boolean authenticationFailed = false;
+
+        @Override
+        public boolean authenticate(InetAddress remoteAddress, int remotePort)
+        {
+            authenticationFailed = true;
+            return false;
+        }
```
jonmeredith: This doesn't need to be implemented - or you could throw an IllegalStateException as your override four-argument authentication method should have been called.

jonmeredith (commented):
Left some more comments. Continuing to look good.

On the testing front I don't think the interrupt connections is being tested still, and I don't see anything that actually retrieves certificates and authenticates/denies to prove it works.

src/java/org/apache/cassandra/auth/IInternodeAuthenticator.java
jyothsnakonisa: sounds good!

src/java/org/apache/cassandra/net/CertificateUtils.java
```diff
@@ -0,0 +1,57 @@
...
+import javax.net.ssl.SSLPeerUnverifiedException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.netty.channel.Channel;
+import io.netty.handler.ssl.SslHandler;
+
+/**
+ * Interface that contains certificate utility methods.
```
yifan-c: nit: update the java doc. It still mentions "Interface"

src/java/org/apache/cassandra/net/CertificateUtils.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+class CertificateUtils
+{
+    private static final Logger logger = LoggerFactory.getLogger(CertificateUtils.class);
+    private static String SSL_PIPELINE_NAME = "ssl";
```
yifan-c: Please replace the usage of `ssl` in inbound and outbound initiators with the defined constant.

yifan-c (commented):
Looks good. Final 2 nits :p 
Thanks for addressing them!

src/java/org/apache/cassandra/net/InternodeConnectionUtils.java
```diff
@@ -0,0 +1,76 @@
...
+            catch (SSLPeerUnverifiedException e)
+            {
+                logger.debug("Failed to get peer certificates for peer {}", channel.remoteAddress(), e);
+            }
+        }
+        return certificates;
+    }
+
+    /**
+     * Discard handler relases the received data silently. when internode authentication fails, the channel is closed,
```
jonmeredith: typo - relases -> releases

src/java/org/apache/cassandra/net/InternodeConnectionUtils.java
```diff
@@ -0,0 +1,76 @@
...
+     * but the pending buffered data may still be fired through the pipeline. To avoid that, authentication handler is
+     * replaced with this DiscardHandler to release all the buffered data, to avoid handling unauthenticated data in the
+     * following handlers.
+     */
+    public static class DiscardHandler extends ChannelInboundHandlerAdapter
+    {
+        @Override
+        public void channelRead(ChannelHandlerContext ctx, Object msg)
+        {
+            ((ByteBuf) msg).release();
```
yifan-c: A bit worried about the type casting. `msg` might not be `ByteBuf` depending how it is used. How about this?

```suggestion
            if (msg instanceof ByteBuf)
            {
                ((ByteBuf) msg).release();
            }
            else
            {
                ctx.fireChannelRead(msg);
            }
```

The handler is to discard the buffered data. Maybe calling it `ByteBufDiscardHandler`?

build.xml
```diff
@@ -549,7 +549,7 @@
           <dependency groupId="junit" artifactId="junit" version="4.12" scope="test">
             <exclusion groupId="org.hamcrest" artifactId="hamcrest-core"/>
           </dependency>
-          <dependency groupId="org.mockito" artifactId="mockito-core" version="3.2.4" scope="test"/>
+          <dependency groupId="org.mockito" artifactId="mockito-inline" version="3.4.0" scope="test"/>
```
jonmeredith: Do you still need the mockito-inline change?

src/java/org/apache/cassandra/auth/AllowAllInternodeAuthenticator.java
```diff
@@ -20,6 +20,7 @@
 package org.apache.cassandra.auth;
 
 import java.net.InetAddress;
+import java.security.cert.Certificate;
```
jonmeredith, line 23: Don't think this is used as you're relying on the deprecated IP/port authenticator.  It would probably be better to switch to using the newer `authenticate` method below so that when the deprecated method is eventually removed the test will still work.

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -476,6 +476,7 @@ public void interruptOutbound(InetAddressAndPort to)
         OutboundConnections pool = channelManagers.get(to);
         if (pool != null)
             pool.interrupt();
+        logger.info("Interrupted outbound connections to {}", to);
```
jonmeredith: this should probably be inside the `if` clause, otherwise it would look like there was an interruption even if there was no `pool`.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -40,6 +54,135 @@
...
+            {
+                openConnections(cluster);
+            }
+            catch (RuntimeException ise)
+            {
+                assertThat(ise.getMessage(), containsString("agreement not reached"));
+            }
+
+            // Verify that authentication is failed and Interrupt is called on outbound connections.
+            cluster.get(1).logs().watchFor("authentication failed to");
```
jonmeredith, line 149: How do you know that the first connection succeeded to be interrupted later? Is it possible to wait for a connection established method before the authentication failure message?

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
jonmeredith: 👍

test/unit/org/apache/cassandra/net/MessagingServiceTest.java
```diff
@@ -228,19 +269,56 @@ private static void addDCLatency(long sentAt, long nowTime)
...
+
+        final IInternodeAuthenticator rejectAllAuthenticator = new InternodeEncryptionEnforcementTest.RejectConnectionsAuthenticator();
+        final OutboundConnectionSettings settings = new OutboundConnectionSettings(InetAddressAndPort.getByName("127.0.0.1")).withAuthenticator(rejectAllAuthenticator);
+        final OutboundConnectionInitiator.ServerAuthenticationHandler handler = new OutboundConnectionInitiator.ServerAuthenticationHandler(settings);
+        final ChannelHandlerContext context = mock(ChannelHandlerContext.class, RETURNS_DEEP_STUBS);
+        when(context.channel().pipeline().get(anyString())).thenReturn(null);
+
+        // Verify that when authentication fails interruptOutbound is called
+        handler.decode(context, Unpooled.buffer(2), new ArrayList<>());
+        verify(ms, times(1)).interruptOutbound(any());
```
jonmeredith: I don't think you need this any more now you've added the distributed test to check interruption. If you agree that it doesn't add anything beyond that, it would be better to just have one test for maintenance going forward.

If you decide you don't need it, remember all the imports above.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
jyothsnakonisa: We are not logging on successful authentication. I didn't add the log thinking that it might create log spamming on successful connections. Should I use NoSpamLogger instead?

In line no 159, I am checking if there is only one successful authentication attempt or not.

src/java/org/apache/cassandra/auth/AllowAllInternodeAuthenticator.java
jyothsnakonisa: Added newly added method in this class.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
jonmeredith: That's sufficient I think. Sorry I missed it.

jonmeredith (approved):
+1 thanks for working through all the rounds of review.

PR #1646, opened 2022-05-25
CASSANDRA-15511(4.0): Utilising BTree Improvements

blerer:
Reduce garbage using new BTree features


src/java/org/apache/cassandra/db/Columns.java
```diff
@@ -96,8 +98,32 @@ public static Columns of(ColumnMetadata c)
...
+    * @param row the row from which to create the new {@code Columns}.
+    * @return the newly created {@code Columns} containing the columns from {@code row}.
+    */
+   public static Columns from(Row row)
+   {
+       try (BTree.FastBuilder<ColumnMetadata> builder = BTree.fastBuilder())
+       {
+           for (ColumnData cd : row)
+               builder.add(cd.column());
+           Object[] tree = builder.build();
```
blambov, line 113: Nit: `return from(builder);`?

src/java/org/apache/cassandra/db/rows/BTreeRow.java
```diff
@@ -227,16 +227,7 @@ public <A> long accumulate(BiLongAccumulator<A, ColumnData> accumulator, A arg,
...
-        min = BTree.<ColumnData>accumulate(btree, (cd, l) -> {
-            int m = Math.min((int) l, minDeletionTime(cd));
-            return m != Integer.MIN_VALUE ? m : Long.MAX_VALUE;
-        }, min);
-
-        if (min == Long.MAX_VALUE)
-            return Integer.MIN_VALUE;
-
-        return Ints.checkedCast(min);
+        return (int) BTree.<ColumnData>accumulate(btree, (cd, l) -> Math.min(l, minDeletionTime(cd)) , min);
```
blambov:
```suggestion
        return (int) BTree.<ColumnData>accumulate(btree, (cd, l) -> Math.min(l, minDeletionTime(cd)), min);
```

src/java/org/apache/cassandra/db/rows/ColumnData.java
```diff
@@ -35,6 +40,154 @@
...
 
+
+    /**
+     * Construct an UpdateFunction for reconciling normal ColumnData
+     * (i.e. not suitable for ComplexColumnDeletion sentinels, but suitable ComplexColumnData or Cell)
+     *
+     * @param updateF a consumer receiving all pairs of reconciled cells
+     * @param maxDeletion the row or partition deletion time to use for purging
+     */
+    public static Reconciler reconciler(ReconcileUpdateFunction updateF, DeletionTime maxDeletion)
```
blambov: "max" in `maxDeletion` is misleading: it is the deletion that should be applied to these cells. I'd rename it to `deletion` or `activeDeletion`.

src/java/org/apache/cassandra/db/rows/ColumnData.java
```diff
@@ -35,6 +40,154 @@
...
+
+        public ColumnData merge(ColumnData existing, ColumnData update)
+        {
+            if (!(existing instanceof ComplexColumnData))
+            {
+                Cell<?> existingCell = (Cell) existing, updateCell = (Cell) update;
+                boolean isExistingShadowed = maxDeletion.deletes(existingCell);
+                boolean isUpdateShadowed = maxDeletion.deletes(updateCell);
+
+                Cell<?> result = isExistingShadowed || isUpdateShadowed
```
blambov: What if both are deleted? `BTree.update` can't deal with `null` from `merge`; perhaps throw an exception?

src/java/org/apache/cassandra/db/rows/ColumnData.java
```diff
@@ -35,6 +40,154 @@
...
+        {
+            return insert;
+        }
+
+        public void onAllocatedOnHeap(long delta)
+        {
+        }
+    };
+
+    public interface ReconcileUpdateFunction
```
blambov: I would call this `PostReconciliationFunction`, which better explains why we use it: we reconcile, and then apply this.

src/java/org/apache/cassandra/db/rows/ColumnData.java
```diff
@@ -35,6 +40,154 @@
...
+            {
+                cells = BTree.transformAndFilter(existingComplex.tree(), this::retain);
+            }
+            return BTree.isEmpty(cells) ? null : new ComplexColumnData(existingComplex.column, cells, complexDeletion);
+        }
+
+        public void close()
+        {
+            pool.offer(this);
+            modifier = null;
```
blambov, line 199: The pool is thread-local, so this probably does not matter at the moment, but to avoid future surprises and making readers wonder, let's clear the fields before returning the item to the pool.

src/java/org/apache/cassandra/utils/btree/BTree.java
```diff
@@ -334,64 +334,77 @@ public static <Compare> Object[] update(Object[] toUpdate, Object[] insert, Comp
...
+                                                                                              UpdateFunction<Insert, Existing> updateF)
     {
         // perform some initial obvious optimisations
         if (isEmpty(insert))
-            return update; // do nothing if update is empty
+        {
+            if (isSimple(updateF))
+                return toUpdate; // do nothing if update is empty and updateF is trivial
+
+            toUpdate = BTree.transformAndFilter(toUpdate, updateF::retain);
```
blambov: This is very different from what will happen when `insert` contains data. In particular, the normal `update` path will only call `retain` on keys in the leaf nodes where data is inserted. `retain` will not be called on paths that weren't descended into, or on keys in non-leaf nodes. I am not sure `update` can properly handle the removal of entries in a leaf making it smaller than `MIN_KEYS` either.

IMHO it will be prohibitively expensive to apply `retain` correctly.

Do we actually need this functionality? Would it not be easier to `transformAndFilter` after applying the update? Alternatively, use a `fastBuilder` with a merging iterator?

src/java/org/apache/cassandra/utils/btree/UpdateFunction.java
```diff
@@ -20,19 +20,22 @@
...
 
-import com.google.common.base.Function;
 /**
  * An interface defining a function to be applied to both the object we are replacing in a BTree and
  * the object that is intended to replace it, returning the object to actually replace it.
  */
-public interface UpdateFunction<K, V> extends Function<K, V>
+public interface UpdateFunction<K, V> 
 {
+    V insert(K update);
```
blambov: `insert` and `retain` need some javaDoc.

src/java/org/apache/cassandra/utils/memory/NativeAllocator.java
```diff
@@ -98,6 +99,42 @@ public DecoratedKey clone(DecoratedKey key, OpOrder.Group writeOp)
...
+                    public DecoratedKey clone(DecoratedKey key)
+                    {
+                        return NativeAllocator.this.clone(key, opGroup);
+                    }
+
+                    @Override
+                    public DeletionInfo clone(DeletionInfo deletionInfo)
+                    {
+                        // TODO Auto-generated method stub
+                        return null;
```
blambov: Either implement or throw unsupported operation or similar.

test/unit/org/apache/cassandra/db/CellTest.java
```diff
@@ -349,29 +349,29 @@ private static FieldIdentifier field(String field)
...
-        Cell<?> r2m2 = BufferCell.live(m, ts2, bb(1), CellPath.create(bb(2)));
-        Cell<?> r2m3 = BufferCell.live(m, ts2, bb(2), CellPath.create(bb(3)));
-        Cell<?> r2m4 = BufferCell.live(m, ts2, bb(3), CellPath.create(bb(4)));
-        List<Cell<?>> cells2 = Lists.newArrayList(r2m2, r2m3, r2m4);
-
-        RowBuilder builder = new RowBuilder();
-        Cells.reconcileComplex(m, cells1.iterator(), cells2.iterator(), DeletionTime.LIVE, builder);
-        Assert.assertEquals(Lists.newArrayList(r1m1, r2m2, r2m3, r2m4), builder.cells);
-    }
+//    @Test
```
blambov, line 352: What is the problem with this test?

src/java/org/apache/cassandra/db/Columns.java
blerer: `from` takes a `Builder`  as argument not a `FastBuilder` and I was unsure if it makes sense to add a new method.

test/unit/org/apache/cassandra/db/CellTest.java
blerer: The reconcileComplex method that was tested has been removed

src/java/org/apache/cassandra/utils/memory/NativeAllocator.java
blerer: The method was not used. I removed it.

src/java/org/apache/cassandra/db/rows/ColumnData.java
```diff
@@ -200,8 +200,9 @@ public ColumnData retain(ColumnData existing)
 
         public void close()
         {
-            pool.offer(this);
+            activeDeletion = null;
             modifier = null;
+            pool.offer(this);
             pool = null;
```
blambov, line 202: This could still race with the pool being set by another thread. Copy locally, clear and then offer.

src/java/org/apache/cassandra/db/partitions/AtomicBTreePartition.java
```diff
@@ -433,6 +431,13 @@ public ColumnData apply(ColumnData insert)
             return insert;
         }
 
+        @Override
+        public void retain(ColumnData existing)
+        {
+            dataSize -= existing.dataSize();
```
blambov, line 437: This doesn't look right. If we retain, the size should not change. This should happen only if we delete.

blerer: You are right. The name is the issue because we only call it for deletion. I will change it.

src/java/org/apache/cassandra/utils/btree/BTree.java
```diff
@@ -2569,30 +2575,31 @@ void copyNoOverflow(Object[] source, int offset, int length)
...
         {
-            if (isSimple(updateF))
+            if (isSimple(apply))
             {
                 copy(source, offset, length);
                 return;
             }
 
-            if (count + length > MAX_KEYS)
+            for (int i = 0; i < length; ++i)
```
blambov: This change should no longer be necessary, and we should state in `update`s definition that the update function can't return nulls as other parts of `update` cannot handle them.

blambov (approved):

src/java/org/apache/cassandra/db/rows/ColumnData.java
```diff
@@ -35,6 +40,176 @@
...
+            {
+                ComplexColumnData existingComplex = (ComplexColumnData) existing;
+
+                DeletionTime existingDeletion = existingComplex.complexDeletion();
+
+                Object[] existingTree = existingComplex.tree();
+
+                if (existingDeletion.supersedes(activeDeletion))
+                {
+                    try (Reconciler reconciler = reconciler(modifier, existingDeletion))
```
belliottsmith: If the `existingDeletion` supersedes the `activeDeletion` then surely we have nothing to do here, assuming the complex data is correct?

src/java/org/apache/cassandra/utils/btree/BTree.java
```diff
@@ -487,7 +493,7 @@ else if (c < 0)
                             c = until & 0x80000000; // must find less or equal; set >= 0 (equal) to 0, otherwise leave intact
                             if (until < 0)
                                 until = -(1 + until);
-                            builder.leaf().copy(inode, ipos, until - ipos, updateF);
+                            builder.leaf().copy(inode, ipos, until - ipos);
```
belliottsmith: Did you mean to remove the `updateF` parameter? It looks like it should be retained.

src/java/org/apache/cassandra/utils/btree/UpdateFunction.java
```diff
@@ -20,19 +20,36 @@
...
+     */
+    V insert(K insert);
+
+    /**
+     * Checks if the specified value  should be deleted or not.
+     *
+     * @param existing the existing value to check
+     * @return {@code null} if the value should be removed from the BTree or the existing value if it should not.
+     */
+    V retain(K existing);
```
belliottsmith: This method is only implemented _and used_ by `ColumnData.Reconciler`, and the concrete type is always known, so we can safely remove it from this interface (and also avoid confusion in BTree.java, where we do not use it)

belliottsmith: I believe this would also permit us to restore the behaviour of implementing `Function` and save allocations for updateF::insert, but this is a small win, and I don't really mind if we retain the new nomenclature other.

PR #1311, opened 2021-11-10
CASSANDRA-17001: Optionally prune CDC segments if consumer fails to consume them fast enough

yifan-c:

PR #1650, opened 2022-05-26
CASSANDRA-17666: Option to disable write path during streaming for CDC enabled tables

yifan-c:


src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
```diff
@@ -23,9 +23,12 @@
 import java.util.List;
 import java.util.Set;
 
+import javax.xml.crypto.Data;
```
smiklosovic: what is this import for?

conf/cassandra.yaml
```diff
@@ -298,6 +298,16 @@ partitioner: org.apache.cassandra.dht.Murmur3Partitioner
 # containing a CDC-enabled table if at space limit in cdc_raw_directory).
 cdc_enabled: false
 
+# Specify whether writes to the CDC-enabled tables should be blocked when CDC data on disk has reached to the limit.
+# When setting to false, the writes will not be blocked and the oldest CDC data on disk will be deleted to
+# ensure the size constraint. The default is true.
+# cdc_block_writes: true
```
jmckenzie-dev, line 304: Wasn't this done as part of CASSANDRA-17004? https://github.com/apache/cassandra/pull/1311/files

conf/cassandra.yaml
```diff
@@ -298,6 +298,16 @@ partitioner: org.apache.cassandra.dht.Murmur3Partitioner
...
+
+# Specify whether enable write path for CDC during internodes streaming.
```
jmckenzie-dev: rephrase to `Specify whether CDC mutations are replayed through the write path on streaming`

conf/cassandra.yaml
```diff
@@ -298,6 +298,16 @@ partitioner: org.apache.cassandra.dht.Murmur3Partitioner
...
+# When enabled, CDC data streamed to the destination node will be written into commit log first. When setting to false,
+# the CDC data will be written into SSTables just the same as normal streaming. The default is true.
+# write_path_for_cdc_enabled: true
```
jmckenzie-dev: Recommend adding a bit as to "why" someone should care about this. For example: `If this is set to false, streaming will be considerably faster however it's possible that, in extreme situations (losing > RF # nodes in a replica set), you may have data in your SSTables that never makes it to the CDC log.`

jmckenzie-dev: Noodling on this a bit; I think this name is confusing. By default, things on the write path w/CDC enabled will have their data in CDC. Maybe we should rename this something like `cdc_on_streaming_enabled` to clarify it? Or `use_write_path_for_cdc_while_streaming`, for example. I like the 1st though (cdc_on_streaming_enabled) since it's pretty straightforward as to what you do and don't get.

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -443,6 +439,29 @@ public void setCDCBlockWrites(boolean val)
...
+
+    @Override
+    public void setWritePathForCdcEnabled(boolean value)
+    {
+        ensureCDCEnabled();
+        DatabaseDescriptor.setWritePathForCDCEnabled(value);
+        logger.info("Set write_path_for_cdc_enabled to {}", value);
+    }
+
+    private void ensureCDCEnabled()
```
jmckenzie-dev: We lose the context (`unable to set block_writes`) or whatever the new usage is here. Maybe take in a string and prepend that as part of the message to indicate what the user was trying to do that's incompatible w/the currently disabled CDC state?

src/java/org/apache/cassandra/db/commitlog/CommitLogMBean.java
```diff
@@ -88,4 +88,10 @@
     public boolean getCDCBlockWrites();
 
     public void setCDCBlockWrites(boolean val);
+
+    /** Returns true if internodes streaming of CDC data should go through write path */
+    boolean isWritePathForCDCEnabled();
+
+    /** Set whether enable write path for CDC data during internodes streaming, e.g. repair */
+    void setWritePathForCdcEnabled(boolean value);
```
jmckenzie-dev: nit: capitalization `CDC`

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
jmckenzie-dev: Think this was a stray IDE autocompletion on an import probably; should be removed.

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
```diff
@@ -23,9 +23,12 @@
...
+
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Iterables;
 
+import org.apache.cassandra.config.DatabaseDescriptor;
```
jmckenzie-dev: import ordering (this should be down below w/the rest of the o.a.c, and the other 2 adjacent moved down)

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
```diff
@@ -172,23 +175,31 @@ private boolean hasCDC(ColumnFamilyStore cfs)
...
      *
-     * For CDC-enabled tables, we want to ensure that the mutations are run through the CommitLog so they
-     * can be archived by the CDC process on discard.
+     * For CDC-enabled tables and write path for CDC is enabled, we want to ensure that the mutations are
+     * run through the CommitLog, so they can be archived by the CDC process on discard.
      */
     private boolean requiresWritePath(ColumnFamilyStore cfs)
     {
-        return hasCDC(cfs) || cfs.streamToMemtable() || (session.streamOperation().requiresViewBuild() && hasViews(cfs));
+        return shouldWriteCommitLog(cfs)
```
jmckenzie-dev: nit: consider renaming the method to something like "cdcRequiresCommitLog" or something to indicate that it's specific to CDC's needs rather than just "writes for this CFS need the CommitLog". Unless you plan on us extending `shouldWriteCommitLog` in the future w/other needs that aren't CDC specific... but I suppose we could do them at that time anyway.

test/distributed/org/apache/cassandra/distributed/test/cdc/ToggleWritePathInStreamingForCDCTest.java
```diff
@@ -0,0 +1,98 @@
...
+    @Test
+    public void testWritePathForCDCEnabled() throws Exception
+    {
+        testWritePathForCDC(true, cluster -> {
+            cluster.get(2).runOnInstance(() -> {
+                boolean containCDCInLog = CommitLog.instance.segmentManager
+                                              .getActiveSegments()
+                                              .stream()
+                                              .anyMatch(s -> s.getCDCState() == CommitLogSegment.CDCState.CONTAINS);
+                assertTrue("Mutaion should be added to commit log when write_path_for_cdc_enabled is true",
```
jmckenzie-dev: not: `Mutation`

test/distributed/org/apache/cassandra/distributed/test/cdc/ToggleWritePathInStreamingForCDCTest.java
```diff
@@ -0,0 +1,98 @@
...
+    @Test
+    public void testWritePathForCDCDisabled() throws Exception
+    {
+        testWritePathForCDC(false, cluster -> {
+            cluster.get(2).runOnInstance(() -> {
+                boolean containCDCInLog = CommitLog.instance.segmentManager
+                                              .getActiveSegments()
+                                              .stream()
+                                              .allMatch(s -> s.getCDCState() != CommitLogSegment.CDCState.CONTAINS);
+                assertTrue("No mutaion should be added to commit log when write_path_for_cdc_enabled is false",
```
jmckenzie-dev: nit: `Mutation`

test/distributed/org/apache/cassandra/distributed/test/cdc/ToggleWritePathInStreamingForCDCTest.java
```diff
@@ -0,0 +1,98 @@
...
+            // repair
+            cluster.get(1).flush(KEYSPACE);
+            cluster.get(2).nodetool("repair", KEYSPACE, "tbl");
+
+            // verify node2 now have data
+            result = cluster.get(2).executeInternal(withKeyspace("SELECT * FROM %s.tbl WHERE k = 1"));
+            assertRows(result, row(1, 1));
+
+            assertion.accept(cluster);
+
```
jmckenzie-dev: ultra nit: whitespace

jmckenzie-dev (changes requested):

conf/cassandra.yaml
yifan-c: Yes..

conf/cassandra.yaml
yifan-c: 👍 Yours reads better.

conf/cassandra.yaml
yifan-c: Sounds good. I changed `RF # nodes` to `quorum # nodes`

yifan-c: Good point. I see how it is confusing now, since the normal write path must produce CDC data when enabled. 
Regarding `cdc_on_streaming_enabled`, I am a bit worried that it could be confused with streaming CDC to downstream consumer... And the name of the second option is probably too long, preferably a short one.  

Essentially, it is a repair mechanism for CDC data. How about `cdc_repair_enabled`?

jmckenzie-dev: That makes it sound like it repairs cdc data. What about `cdc_on_repair_enabled`?

PR #1653, opened 2022-05-27
CASSANDRA-17073: Send all gossip notifications also to the local node

pkolaczk:
For symmetry, the Gossiper sends onJoin, onAlive and onDead notifications
for the local node.

Previously onChange notifications were sent for all nodes including
the local one, but onJoin, onAlive and onDead were sent for
remote nodes only. Hence, the subscribers saw incomplete state
of the local node, e.g. they missed the local state that was initially set
on Gossiper startup.


pmcfadin:

PR #1655, opened 2022-05-31
CASSANDRA-17569 DOC - Fixed H2 markdown to H1 to force nodetool commands to become pa…

ErickRamirezAU:
…ge titles

patch by Erick Ramirez; reviewed by ... for CASSANDRA-17569


ossarga (approved):
@ErickRamirezAU the change looks good to me. Thank you for doing this. We just need to make sure that it is committed into the different branches. 3.0, 3.11, 4.0, 4.1, and trunk.

polandll:
Does this build correctly? If the headers were switched from H2 to H1, anywhere an include is on another page, it will cause an error without an `leveloffset=+1`. Maybe the nodetool pages are never included in another page? I'd have to look.

michaelsembwever:
got fixed in https://github.com/apache/cassandra/commit/1e2c88fff832d891b296165e9adda786182e850d#diff-6640099a8316e15b9c559ecc7b6b7acec5d3cfe258257a1f687e4f2406df42db

PR #1658, opened 2022-06-01
CASSANDRA-17677: Fix BulkLoader to load entireSSTableThrottle and entireSSTableInterDcThrottle

frankgh:
`entire_sstable_stream_throughput_outbound` and `entire_sstable_inter_dc_stream_throughput_outbound` were
introduced in CASSANDRA-17065. They were added to the LoaderOptions class but they are not being used in
the BulkLoader class. In this commit, we make use of the entire_sstable streaming options in BulkLoader.
In addittion, the `throttle` and `interDcThrottle` are being passed as MB/s but they should be MiB/s.
This commit fixes the verbiage as well as setting the correct value for those properties in BulkLoader.

### DIFF
--- a/doc/modules/cassandra/pages/operating/bulk_loading.adoc
+++ b/doc/modules/cassandra/pages/operating/bulk_loading.adoc
@@ -80,13 +80,23 @@ The following options are supported, with `-d,--nodes <initial hosts>` required:
 -d,--nodes <initial hosts>                                   Required.
                                                              Try to connect to these hosts (comma separated) initially for ring information
 
+-emib,--entire-sstable-throttle-mib <throttle-mib>           Entire SSTable throttle
+                                                             speed in MiB/s (default 0 for unlimited).
+
+-eidctmib,--entire-sstable-inter-dc-throttle-mib <inter-dc-throttle-mib>
+                                                             Entire SSTable inter-datacenter throttle
+                                                             speed in MiB/s (default 0 for unlimited).
+
 -f,--conf-path <path to config file>                         cassandra.yaml file path for streaming throughput and client/server SSL.
 
 -h,--help                                                    Display this help message
 
 -i,--ignore <NODES>                                          Don't stream to this (comma separated) list of nodes
 
--idct,--inter-dc-throttle <inter-dc-throttle>                Inter-datacenter throttle speed in Mbits (default unlimited)
+-idct,--inter-dc-throttle <inter-dc-throttle>                (deprecated) Inter-datacenter throttle speed in Mbits (default 0 for unlimited)
+                                                             Use -idctmib / --inter-dc-throttle-mib instead
+
+-idctmib,--inter-dc-throttle-mib <inter-dc-throttle-mib>     Inter-datacenter throttle speed in MiB/s (default 0 for unlimited)
 
 -k,--target-keyspace <target keyspace name>                  Target
                                                              keyspace name
@@ -111,8 +121,10 @@ The following options are supported, with `-d,--nodes <initial hosts>` required:
                                                              for TLS internode communication (default 7001)
 -st,--store-type <STORE-TYPE>                                Client SSL:
                                                              type of store
--t,--throttle <throttle>                                     Throttle
-                                                             speed in Mbits (default unlimited)
+-t,--throttle <throttle>                                     (deprecated) Throttle speed in Mbits (default 0 for unlimited)
+                                                             Use -tmib / --throttle-mib instead
+-tmib,--throttle-mib <throttle-mib>                          Throttle
+                                                             speed in MiB/s (default 0 for unlimited)
 -ts,--truststore <TRUSTSTORE>                                Client SSL:
                                                              full path to truststore
 -tspw,--truststore-password <TRUSTSTORE-PASSWORD>            Client SSL:
--- a/src/java/org/apache/cassandra/tools/LoaderOptions.java
+++ b/src/java/org/apache/cassandra/tools/LoaderOptions.java
@@ -22,25 +22,37 @@ package org.apache.cassandra.tools;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
-import java.net.*;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.MalformedURLException;
+import java.net.UnknownHostException;
 import java.util.HashSet;
 import java.util.Set;
 
 import com.google.common.base.Throwables;
 import com.google.common.net.HostAndPort;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.GnuParser;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import org.apache.cassandra.config.*;
+import com.datastax.driver.core.AuthProvider;
+import com.datastax.driver.core.PlainTextAuthProvider;
+import org.apache.cassandra.config.Config;
+import org.apache.cassandra.config.DataRateSpec;
+import org.apache.cassandra.config.EncryptionOptions;
+import org.apache.cassandra.config.YamlConfigurationLoader;
 import org.apache.cassandra.exceptions.ConfigurationException;
 import org.apache.cassandra.io.util.File;
 import org.apache.cassandra.locator.InetAddressAndPort;
 import org.apache.cassandra.tools.BulkLoader.CmdLineOptions;
 
-import com.datastax.driver.core.AuthProvider;
-import com.datastax.driver.core.PlainTextAuthProvider;
-import org.apache.commons.cli.*;
-import org.apache.commons.lang3.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import static org.apache.cassandra.config.DataRateSpec.DataRateUnit.MEBIBYTES_PER_SECOND;
 
 public class LoaderOptions
 {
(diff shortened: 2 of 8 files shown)


src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -653,10 +653,10 @@ private static CmdLineOptions getCmdLineOptions()
...
-        options.addOption("idct", INTER_DC_THROTTLE_MBITS, "inter-dc-throttle", "inter-datacenter throttle speed in Mbits (default unlimited)");
-        options.addOption("e", ENTIRE_SSTABLE_THROTTLE_MBITS, "entire-sstable-throttle", "entire SSTable throttle speed in Mbits (default unlimited)");
-        options.addOption("eidct", ENTIRE_SSTABLE_INTER_DC_THROTTLE_MBITS, "entire-sstable-inter-dc-throttle", "entire SSTable inter-datacenter throttle speed in Mbits (default unlimited)");
+        options.addOption("t", THROTTLE_MEGABITS, "throttle", "throttle speed in Mbps (default unlimited)");
+        options.addOption("idct", INTER_DC_THROTTLE_MEGABITS, "inter-dc-throttle", "inter-datacenter throttle speed in Mbps (default unlimited)");
+        options.addOption("e", ENTIRE_SSTABLE_THROTTLE_MEBIBYTES, "entire-sstable-throttle", "entire SSTable throttle speed in MiB/s (default unlimited)");
+        options.addOption("eidct", ENTIRE_SSTABLE_INTER_DC_THROTTLE_MEBIBYTES, "entire-sstable-inter-dc-throttle", "entire SSTable inter-datacenter throttle speed in MiB/s (default unlimited)");
```
ekaterinadimitrova2: Note for the second reviewer: we were discussing with Francisco that it would have been nice if we had all 4 parameters with the same unit. Probably we can add a note that those in bits are deprecated(but still keep them for compatibility) and add two more for them which can be used for the mebibytes/s. On the other hand we are already in a freeze so probably the best is to leave things as is now and add for the next release in trunk directly the opportunity to set them with units, similar to the cassandra.yaml

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -653,10 +653,10 @@ private static CmdLineOptions getCmdLineOptions()
         options.addOption("p",  NATIVE_PORT_OPTION, "native transport port", "port used for native connection (default 9042)");
         options.addOption("sp",  STORAGE_PORT_OPTION, "storage port", "port used for internode communication (default 7000)");
         options.addOption("ssp",  SSL_STORAGE_PORT_OPTION, "ssl storage port", "port used for TLS internode communication (default 7001)");
-        options.addOption("t", THROTTLE_MBITS, "throttle", "throttle speed in Mbits (default unlimited)");
-        options.addOption("idct", INTER_DC_THROTTLE_MBITS, "inter-dc-throttle", "inter-datacenter throttle speed in Mbits (default unlimited)");
-        options.addOption("e", ENTIRE_SSTABLE_THROTTLE_MBITS, "entire-sstable-throttle", "entire SSTable throttle speed in Mbits (default unlimited)");
-        options.addOption("eidct", ENTIRE_SSTABLE_INTER_DC_THROTTLE_MBITS, "entire-sstable-inter-dc-throttle", "entire SSTable inter-datacenter throttle speed in Mbits (default unlimited)");
+        options.addOption("t", THROTTLE_MEGABITS, "throttle", "throttle speed in Mbps (default unlimited)");
```
yifan-c: If you are updating the default throttling, also update here. However, I'd prefer to keep it the same (unlimited), since it is an external tool.

doc/modules/cassandra/pages/operating/bulk_loading.adoc
```diff
@@ -80,13 +80,20 @@ The following options are supported, with `-d,--nodes <initial hosts>` required:
 -d,--nodes <initial hosts>                                   Required.
                                                              Try to connect to these hosts (comma separated) initially for ring information
 
+-e,--entire-sstable-throttle <throttle>                      Entire SSTable throttle
+                                                             speed in MiB/s (default 24MiB/s).
```
yifan-c: Talked offline. Although the document says 24 MiB/s, the default is actually unlimited. I feel it makes sense with unlimited for the tool

src/java/org/apache/cassandra/tools/LoaderOptions.java
yifan-c: In this approach, would it make sense for all four options has the same unit as MEGABITS, or declare new names for the ones using MEBIBYTES? I'd prefer the later one. 
The deprecation approach requires to declare _new_ options and mark the old ones as deprecated. Both new and old options are present. Currently, the option names are consistent with different unit. When the new option is added, there will be options names that are not consistent, but the unit is. Let's say, the new option name for throttle are `tmib` and `throttle-mib` (shorthand and full accordingly). `tmib` and `e` has inconsistent name, but they use the same unit.

doc/modules/cassandra/pages/operating/bulk_loading.adoc
frankgh: yes, this is correct for the bulk loading. I will adjust it

src/java/org/apache/cassandra/tools/LoaderOptions.java
frankgh: I agree that we should keep the `unlimited` as default since it's an external tool.

src/java/org/apache/cassandra/tools/LoaderOptions.java
frankgh: what do you think @ekaterinadimitrova2 about Yifan's proposal? my only concern with this approach is that we need to worry about the case where the user provides both parameters with different units. So we'll need to address that case

ekaterinadimitrova2: The deprecation approach suggested by @yifan-c  makes sense to me. 
I would expect if someone wants to use the 4 of them to use the new ones mebibytes/s for all of them.
The old two properties with megabits/s are for compatibility with scripts. If someone wants to add also the other 2 they will decide whether they want to migrate the old ones to the new units or just add the two new properties with different units and be inconsistent.

doc/modules/cassandra/pages/operating/bulk_loading.adoc
ekaterinadimitrova2: Agreed, the change was triggered by the new doc and not the opposite - those were unlimited in the code for the tool for the 4 options

src/java/org/apache/cassandra/tools/LoaderOptions.java
ekaterinadimitrova2: By the way I noticed we mention unlimited but it is not explicitly said that to make it unlimited you need to set it to 0. I checked also in cassandra.yaml and I didn't find that. Maybe we should document it?

frankgh: maybe the text should read `(default 0 for unlimited)`

src/java/org/apache/cassandra/tools/LoaderOptions.java
frankgh: So with Yifan's suggestion we'll have the following:
`t`, `throttle` -> megabits (deprecated)
`tmib`, `throttle-mib` -> mebibytes (new)
`emib`, `entire-sstable-throttle-mib` -> mebibytes (new)
`idct`, `inter-dc-throttle` -> megabits (deprecated)
`idctmib`, `inter-dc-throttle-mib` -> mebibytes (new)
`eidctmib`, `entire-sstable-inter-dc-throttle-mib` -> mebibytes (new)

What do you think @yifan-c, @ekaterinadimitrova2 ?

src/java/org/apache/cassandra/tools/LoaderOptions.java
ekaterinadimitrova2: SGTM

src/java/org/apache/cassandra/tools/LoaderOptions.java
yifan-c: Looks good to me.

ekaterinadimitrova2: `idctib` and `eidctmib` sound a bit long but they follow the already established naming convention and when people know where it comes from I guess it will be intuitive.

frankgh: I pushed the changes above. Let me know what you think

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -85,10 +101,10 @@
     public final String user;
     public final String passwd;
     public final AuthProvider authProvider;
-    public final int throttle;
-    public final int interDcThrottle;
-    public final int entireSSTableThrottle;
-    public final int entireSSTableInterDcThrottle;
+    public final double throttleMegabits;
+    public final double interDcThrottleMegabits;
```
ekaterinadimitrova2: @yifan-c , @frankgh considering our later discussions and the ticket that @yifan-c pointed to earlier today, probably we should find a way to workaround this change too? WDYT?

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -60,10 +72,14 @@
...
-    public static final String ENTIRE_SSTABLE_THROTTLE_MBITS = "entire-sstable-throttle";
-    public static final String ENTIRE_SSTABLE_INTER_DC_THROTTLE_MBITS = "entire-sstable-inter-dc-throttle";
+    @Deprecated
+    public static final String THROTTLE_MEGABITS = "throttle";
+    public static final String THROTTLE_MEBIBYTES = "throttle-mib";
+    @Deprecated
+    public static final String INTER_DC_THROTTLE_MEGABITS = "inter-dc-throttle";
+    public static final String INTER_DC_THROTTLE_MEBIBYTES = "inter-dc-throttle-mib";
+    public static final String ENTIRE_SSTABLE_THROTTLE_MEBIBYTES = "entire-sstable-throttle-mib";
+    public static final String ENTIRE_SSTABLE_INTER_DC_THROTTLE_MEBIBYTES = "entire-sstable-inter-dc-throttle-mib";
```
yifan-c: @frankgh, can you also revert the renaming in this PR? THX

src/java/org/apache/cassandra/tools/LoaderOptions.java
yifan-c: I feel less worried about someone reading the field values... I can only image the use cases that use the option name constants to invoke the tool, or construct the option programmatically. The change does not affect those use cases.

PR #1662, opened 2022-06-02
CASSANDRA-17584

smiklosovic:


src/java/org/apache/cassandra/db/virtual/GossipInfoTable.java
```diff
@@ -19,8 +19,12 @@
 package org.apache.cassandra.db.virtual;
 
 import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.HashSet;
```
adelapena: Nit: unused imports

src/java/org/apache/cassandra/db/virtual/GossipInfoTable.java
```diff
@@ -76,10 +89,10 @@ final class GossipInfoTable extends AbstractVirtualTable
     public DataSet data()
     {
         SimpleDataSet result = new SimpleDataSet(metadata());
-        for (Map.Entry<InetAddressAndPort, EndpointState> entry : Gossiper.instance.endpointStateMap.entrySet())
+        for (Map.Entry<InetAddressAndPort, EndpointState> entry : endpointStateMapSupplier.get().entrySet())
         {
             InetAddressAndPort endpoint = entry.getKey();
-            EndpointState localState = entry.getValue();
+            EndpointState localState = new EndpointState(entry.getValue());
```
adelapena, line 96: I think we don't need to create a new `EndpointState` here, the `EndpointState` is already copied by the `testSelectAllWithStateTransitions` test.

test/unit/org/apache/cassandra/db/virtual/GossipInfoTableTest.java
```diff
@@ -33,64 +36,68 @@
...
+            final AtomicReference<Map<InetAddressAndPort, EndpointState>> stateRef = new AtomicReference<>(null);
+
+            Awaitility.await().until(() -> {
+                ConcurrentMap<InetAddressAndPort, EndpointState> endpointStateMap = Gossiper.instance.endpointStateMap;
+                stateRef.set(endpointStateMap);
+                return endpointStateMap.size() == 1;
+            });
+
+            Optional<Map.Entry<InetAddressAndPort, EndpointState>> entry = stateRef.get().entrySet().stream().findFirst();
+            assertThat(entry).isNotEmpty();
```
adelapena: This check is duplicated a few lines below.

test/unit/org/apache/cassandra/db/virtual/GossipInfoTableTest.java
```diff
@@ -33,64 +36,68 @@
...
+
+            InetAddressAndPort endpoint = entry.get().getKey();
+            EndpointState localState = new EndpointState(entry.get().getValue());
```
adelapena: This might be simplified to:
```java
ConcurrentMap<InetAddressAndPort, EndpointState> states = Gossiper.instance.endpointStateMap;
Awaitility.await().until(() -> !states.isEmpty());
Map.Entry<InetAddressAndPort, EndpointState> entry = states.entrySet().stream().findFirst()
                                                           .orElseThrow(AssertionError::new);
InetAddressAndPort endpoint = entry.getKey();
EndpointState localState = new EndpointState(entry.getValue());
```

adelapena (commented):
Overall the approach looks good to me, I have just left a few minor suggestions.

src/java/org/apache/cassandra/db/virtual/GossipInfoTable.java
smiklosovic: @adelapena well, but imagine that this is used in production, then we get that map from the gossip. However, is not it true that by time I am populating the table, that value might change or it might change in the middle of me processing it? This basically ensures that whatever happens in Gossiper, the table I am going to return basically captures one particular point in time.

adelapena: Ah, that makes sense. I thought this was related with the test fix, but it seems like a separate fix/improvement. I'm ok with it.

test/unit/org/apache/cassandra/db/virtual/GossipInfoTableTest.java
```diff
@@ -33,64 +36,58 @@
...
+            Map.Entry<InetAddressAndPort, EndpointState> entry = states.entrySet().stream().findFirst()
+                    .orElseThrow(AssertionError::new);
+            InetAddressAndPort endpoint = entry.getKey();
+            EndpointState localState = new EndpointState(entry.getValue());
 
-            assertThat(resultSet.size()).isEqualTo(1);
-            assertThat(Gossiper.instance.endpointStateMap.size()).isEqualTo(1);
+            Supplier<Map<InetAddressAndPort, EndpointState>> endpointStateMapSupplier = () -> new HashMap<InetAddressAndPort, EndpointState>() {{put(endpoint, localState);}};
 
-            Optional<Map.Entry<InetAddressAndPort, EndpointState>> entry = Gossiper.instance.endpointStateMap.entrySet()
```
adelapena, line 77: Nit: Now the import of `Optional` is not used anymore.

test/unit/org/apache/cassandra/db/virtual/GossipInfoTableTest.java
```diff
@@ -18,11 +18,14 @@
 
 package org.apache.cassandra.db.virtual;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicReference;
```
adelapena: Nit: unused import after the last changes.

src/java/org/apache/cassandra/db/virtual/GossipInfoTable.java
```diff
@@ -76,10 +87,10 @@ final class GossipInfoTable extends AbstractVirtualTable
...
```
adelapena, line 96: Maybe we can add a brief comment here saying that we are snapshotting the `EndpointState`s obtained from Gossiper because they can be concurrently modified by `LoadBroadcaster`.

adelapena (approved):
Looks good to me, left a couple of nits that can be addressed on commit if CI looks good.

test/unit/org/apache/cassandra/db/virtual/GossipInfoTableTest.java
smiklosovic: ah right, thanks, always forget to check that ...

src/java/org/apache/cassandra/db/virtual/GossipInfoTable.java
smiklosovic: ok i ll add it.

adelapena (approved):

PR #1665, opened 2022-06-03
CASSANDRA-17083 testsome target doesn't work with wildcards

bereng:


bereng:
CI [here](https://ci-cassandra.apache.org/view/all/job/Cassandra-devbranch/1753)

PR #1672, opened 2022-06-07
Refactor python code from bin to pylib

bschoening:
This splits main() and the majority of the cqlsh.py code into a new file, cqlshmain.py in the pylib directory.

What remains in cqlsh.py is mostly the zip loader and a few startup sanity checks.

Using sdiff with the original file works well to show the changes clearly.


jeffwidman (commented):
The overall idea makes sense, and this looks great.

it's sad that this loses `git blame` functionality, as that history can be very useful in a project like this.

Did you try doing it as a couple of commits? As long as it gets merged via merge and not squash, than that would preserve `git blame` history.

You'd probably want:
1. first commit that renames `bin/cqlsh.py` to `pylib/cqlshlib/cqlshmain.py`... this is what preserves the `git` history across the rename.
2. second commit create a new `bin/cqlsh/.py` file for the entrypoint, and also refactors the  `cqlshmain.py` file so it's usable as this entrypoint.

Again though, this only works if it's merged via a merge commit and not via squash.

bschoening:
@jeffwidman - good suggestion.  When this is ready to merge, I can update the patch.

PR #1673, opened 2022-06-07
Error while creating hotel.hotels table

codingswag998:
line 35 should be ```pois set<text>```


smiklosovic:
this work will be included in https://github.com/apache/cassandra/pull/2062

PR #1676, opened 2022-06-09
Cassandra 17531 Refactor python code from bin to pylib

bschoening:
This change splits main() and the majority of the cqlsh.py code into a new file, cqlshmain.py in the pylib directory.  What remains in cqlsh.py is mostly the zip loader and a few startup sanity checks.

Note, the first commit renames bin/cqlsh.py to pylib/cqlshlib/cqlshmain.py to preserves the git history across the rename.
The second commit create a new bin/cqlsh.py file as a startup script, and refactors the cqlshmain.py file so it's callable (as suggested by Jeff Widman)


jeffwidman:
@smiklosovic can you take a look at this? It'd be helpful to separate out the "entrypoint" from the "library logic", so that others can leverage the library logic more cleanly.

PR #1678, opened 2022-06-09
Cassandra 17658

jacek-lewandowski:


src/java/org/apache/cassandra/db/Keyspace.java
```diff
@@ -137,7 +140,10 @@ public static void setInitialized()
     @VisibleForTesting
     public static void unsetInitialized()
     {
-        initialized = false;
+        synchronized (Schema.instance)
```
adelapena, line 143: I think the method `Kesypace.unsetInitialized` isn't used anywhere, and it didn't seem to be used when it was introduced for the first time on CASSANDRA-17044. Should we remove it?

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -227,6 +227,7 @@ public Keyspace maybeAddKeyspaceInstance(String keyspaceName, Supplier<Keyspace>
         return keyspaceInstances.blockingLoadIfAbsent(keyspaceName, loadFunction);
     }
 
+    @VisibleForTesting
```
adelapena: The method `Schema.maybeRemoveKeyspaceInstance` doesn't seem used anymore, since it has lost its single caller, `Schema.dropKeyspace`. Should we remove it?

test/long/org/apache/cassandra/cql3/CorruptionTest.java
```diff
@@ -59,7 +60,11 @@ public class CorruptionTest extends SchemaLoader
     @BeforeClass()
     public static void setup() throws ConfigurationException, IOException
     {
-        Schema.instance.clear();
```
adelapena, line 62: Nit: removing this leaves an unused import of `Schema`.

test/long/org/apache/cassandra/cql3/CorruptionTest.java
```diff
@@ -59,7 +60,11 @@ public class CorruptionTest extends SchemaLoader
...
+        DatabaseDescriptor.daemonInitialization();
+
+        // clean-up is only needed if we run the test from IDE
+        ServerTestUtils.mkdirs();
+        ServerTestUtils.cleanup();
```
adelapena: This is always done before any instantiation of `EmbeddedCassandraService`, so we may make tests slightly briefer and less error prone if we encapsulate both things into a single utility method. Going a bit further, we could move the entire `EmbeddedCassandraService` class into the tests directory (it's only used for  tests), and place the cleanup of dirs into its constructor. wdyt?

jacek-lewandowski: I was thinking about that, I agree we can safely have a single method which includes `mkdirs` and `cleanup`., or even have a static methods which does that and creates ECS. However, I'm not sure we should move `EmbeddedCassandraService` to the tests directory because it is currently a user visible thing and some users may start Cassandra that way. When it was moved into the tests directory, it would not be a part of the `cassandra-all` artifact in public Maven repository. This is also why we should not put `mkdirs` and `cleanup` into the constructor

adelapena: Right, there could be someone using `EmbeddedCassandraService` from the Maven artifact. Then we can put the repeated block of code into a test method into a utility method, more or less [this way](https://github.com/adelapena/cassandra/commit/c6f25e16f548c929bc79e6e91d9efc8df6ad3124). I think that would save us some code duplication, wdyt?

src/java/org/apache/cassandra/schema/Schema.java
adelapena: After last changes the only caller of `Schema.maybeRemoveKeyspaceInstance` is 
 `Schema.dropKeyspace`, so we shouldn't add the `@VisibleForTesting` annotation. We could also make the method private.

ekaterinadimitrova2: I think this comment might have been missed

test/long/org/apache/cassandra/cql3/CorruptionTest.java
ekaterinadimitrova2: I agree with everything said here, maybe I would just rename the method as we already have start in `EmbeddedCassandraService` itself

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -533,13 +535,13 @@ private synchronized void updateVersion(UUID version)
     }
 
     /**
-     * Clear all KS/CF metadata and reset version.
+     * Clear all KS/CF metadata and reset version. Not to be ever used in production code
      */
-    public synchronized void clear()
+    @VisibleForTesting
+    public synchronized void clearUnsafe()
```
ekaterinadimitrova2: My IDE didn't find any usages of this method. Also, searching on the commit I didn't find any?

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -626,7 +628,9 @@ public void resetLocalSchema()
         updateHandler.clear();
 
         logger.debug("Clearing local schema keyspace instances...");
-        clear();
+        distributedKeyspaces.forEach(this::unload);
+        updateVersion(SchemaConstants.emptyVersion);
+        SchemaDiagnostics.schemaCleared(this);
```
ekaterinadimitrova2, line 622: Those were in a synchronized method before?

test/unit/org/apache/cassandra/ServerTestUtils.java
```diff
@@ -189,6 +190,16 @@ public static void cleanupSavedCaches()
         cleanupDirectory(DatabaseDescriptor.getSavedCachesLocation());
     }
 
+    public static EmbeddedCassandraService startEmbeddedCassandraService() throws IOException
```
ekaterinadimitrova2, line 193: I am wondering whether this is not a bit of a confusing name. We already have a start method in EmbeddedCassandraService.

test/unit/org/apache/cassandra/schema/SchemaTest.java
```diff
@@ -43,19 +44,16 @@
     public static void setupDatabaseDescriptor()
     {
         DatabaseDescriptor.daemonInitialization();
+        ServerTestUtils.prepareServer();
+        Schema.instance.loadFromDisk();
```
ekaterinadimitrova2, line 46: `loadFromDisk `is not exactly a DD setup, maybe it should be just a `setup `method and not `setupDatabaseDescriptor`?

test/unit/org/apache/cassandra/schema/SchemaTest.java
```diff
@@ -43,19 +44,16 @@
...
     }
 
     @Test
     public void testTransKsMigration() throws IOException
     {
-        CommitLog.instance.start();
```
ekaterinadimitrova2, line 51: Now we can clean the unused import import `org.apache.cassandra.db.commitlog.CommitLog;`

ekaterinadimitrova2 (commented):
Overall LGTM, left just a few small comments/questions

src/java/org/apache/cassandra/schema/Schema.java
jacek-lewandowski: good point, I need to revisit this

test/unit/org/apache/cassandra/ServerTestUtils.java
jacek-lewandowski: Why confusing? `start` is instance method in `EmbeddedCassandraService`. This is a different class, dedicated solely for setting up Cassandra instance for testing purposes

src/java/org/apache/cassandra/schema/Schema.java
jacek-lewandowski: We discussed on Slack, but I can apply this comment

ekaterinadimitrova2: I am not sure what was discussed tbh, I think it was with @adelapena probably

test/unit/org/apache/cassandra/ServerTestUtils.java
ekaterinadimitrova2: Because it doesn't do only the start from what I see, but at the same time I guess cleaning should be expected part of the start process in the tests. I am not super worried, just expressed opinion :-) I guess we can just add a Java doc, I will leave it to you

src/java/org/apache/cassandra/schema/Schema.java
ekaterinadimitrova2: Hmm, do we want the whole method now? I mean before it was considered that only those three operations were enough. Just trying to be careful with the price of synchronization but at the same time not to miss it :-)

src/java/org/apache/cassandra/schema/DefaultSchemaUpdateHandler.java
```diff
@@ -257,12 +255,11 @@ private synchronized SchemaTransformationResult reload()
...
-               ? reload()
-               : migrationCoordinator.pullSchemaFromAnyNode()
-                                     .flatMap(mutations -> ImmediateFuture.success(applyMutations(mutations)))
-                                     .awaitThrowUncheckedOnInterrupt()
-                                     .getNow();
+        if (local)
+            return reload();
+
+        Collection<Mutation> mutations = migrationCoordinator.pullSchemaFromAnyNode().awaitThrowUncheckedOnInterrupt().getNow();
+        return applyMutations(mutations);
```
ekaterinadimitrova2, line 262: Not sure why this change was needed

src/java/org/apache/cassandra/schema/Schema.java
jacek-lewandowski: Yes, I think it was a mistake that it wasn't synchronized. Synchronization does not really matter in terms of performance in this case because that method is called only from NodeTool and it is one time action. Basically what we need I think is that there are no other schema changes issued locally when we are restoring schema from the other node, not just during cleaning

src/java/org/apache/cassandra/schema/DefaultSchemaUpdateHandler.java
jacek-lewandowski: Yeah, so in the previous scenario, `reset` was called from `Schema.resetLocalSchema`, in a thread 1. In this method, we pull the schema from some node, which is actually processed asynchronously in thread 2 and returns `Future`. Hence we use `flatMap`, `applyMutations` was called in thread 2. It ended up in a deadlock because `Schema.resetLocalSchema`, which is now synchronized was waiting for the `reset` to finish, while `applyMutations` called from thread 2 was waiting for updating `Schema` in thread 2. 

Since it is blocking anyway, we firstly fetch the schema using whichever thread, and then continue with `applyMutations` in the main thread which initiated the schema changes

ekaterinadimitrova2: Oh, yeah, needed because now the whole method was synchronized, side effect... thanks :-)

PR #1683, opened 2022-06-14
CASSANDRA-14715 3.11

smiklosovic:


src/java/org/apache/cassandra/service/DataResolver.java
```diff
@@ -56,10 +58,16 @@ public class DataResolver extends ResponseResolver
     private final boolean enforceStrictLiveness;
 
     DataResolver(Keyspace keyspace, ReadCommand command, ConsistencyLevel consistency, int maxResponseCount, long queryStartNanoTime)
+    {
+        this(keyspace, command, consistency, maxResponseCount, queryStartNanoTime, command.metadata().enforceStrictLiveness());
+    }
+
+    @VisibleForTesting
+    DataResolver(Keyspace keyspace, ReadCommand command, ConsistencyLevel consistency, int maxResponseCount, long queryStartNanoTime, boolean enforceStrictLiveness)
```
smiklosovic, line 66: ReadCommand is quite hard to mock / initialize in test, it just needs too much stuff. When command is null, for the sake of the test to have something there for `DataResolver` constructor, it would fail on `command.metadata().enforceStrictLiveness`. So I did one more constructor which might be called from the test where we set it without going through (null) `command`.

src/java/org/apache/cassandra/service/DataResolver.java
```diff
@@ -248,13 +256,22 @@ private PartitionIterator resolveInternal(ResolveContext context,
                : Transformation.apply(counted, new EmptyPartitionsDiscarder());
     }
 
-    private class RepairMergeListener implements UnfilteredPartitionIterators.MergeListener
+    @VisibleForTesting
+    class RepairMergeListener implements UnfilteredPartitionIterators.MergeListener
```
smiklosovic, line 260: I had to make this to make an instance of it in the test.

src/java/org/apache/cassandra/service/DataResolver.java
```diff
@@ -248,13 +256,22 @@ private PartitionIterator resolveInternal(ResolveContext context,
...
     {
+        @VisibleForTesting
+        final MutableInt finishedRepairs = new MutableInt();
```
smiklosovic, line 263: This stuff is propagated there through constructor, I was not able to make this class static because then it broke other things.

src/java/org/apache/cassandra/service/DataResolver.java
```diff
@@ -293,22 +310,30 @@ private boolean isReversed(List<UnfilteredRowIterator> versions)
             return false;
         }
 
+        @VisibleForTesting
+        void waitForRepairResults() throws TimeoutException
+        {
+            FBUtilities.waitOnFutures(repairResults, DatabaseDescriptor.getWriteRpcTimeout(), finishedRepairs);
```
smiklosovic, line 316: I extracted this call to separate method so I can override its behavior in the test.

pmcfadin:
Closing as [CASSANDRA-17683](https://issues.apache.org/jira/browse/CASSANDRA-17683) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #1684, opened 2022-06-15
CASSANDRA-17151 trunk: Add guardrail for column value size

adelapena:


conf/cassandra.yaml
```diff
@@ -1752,6 +1752,17 @@ drop_compact_storage_enabled: false
...
 # write_consistency_levels_disallowed: []
 #
+# Guardrail to warn or fail when writing column values larger than threshold.
+# This guardrail is only applied to the values of regular columns because both the serialized partitions keys and the
+# values of the components of the clustering key already have a fixed, relatively small size limit of 65535 bytes, which
+# is probably lesser than the thresholds defined here.
+# This guardrail is different to max_value_size. max_value_size is checked when deserializing any value to detect
+# sstable corruption, whereas this guardrail is checked on the CQL layer at write time to reject regular user queries
+# inserting too large columns.
+# Default -1 to disable.
```
ekaterinadimitrova2: `# Default null to disable.`

conf/cassandra.yaml
```diff
@@ -1752,6 +1752,17 @@ drop_compact_storage_enabled: false
 # write_consistency_levels_warned: []
 # write_consistency_levels_disallowed: []
 #
+# Guardrail to warn or fail when writing column values larger than threshold.
+# This guardrail is only applied to the values of regular columns because both the serialized partitions keys and the
+# values of the components of the clustering key already have a fixed, relatively small size limit of 65535 bytes, which
```
ekaterinadimitrova2, line 1772: So I was not sure immediately where that value came from - 65535 bytes. 
So I found this in the QueryProcessor - 
```
// check that key can be handled by FBUtilities.writeShortByteArray
        if (key.remaining() > FBUtilities.MAX_UNSIGNED_SHORT)
```
The comment seems outdated as there is no `FBUtilities.writeShortByteArray` now. Shall we use the opportunity to clear this a bit? WDYT? I mean the whole explanation where this number comes from

test/unit/org/apache/cassandra/db/guardrails/GuardrailColumnValueSizeTest.java
```diff
@@ -0,0 +1,650 @@
...
+        // the size of primary key columns is not guarded because they already have a fixed limit of 65535B
+
+        testNoThreshold("INSERT INTO %s (k, v) VALUES (?, 0)");
+        testNoThreshold("UPDATE %s SET v = 1 WHERE k = ?");
+        testNoThreshold("DELETE v FROM %s WHERE k = ?");
+        testNoThreshold("DELETE FROM %s WHERE k = ?");
+    }
+
+    @Test
+    public void testComplexPartitionKey() throws Throwable
```
ekaterinadimitrova2: I think this was meant to be `testCompositePartitionKey`? No?

test/unit/org/apache/cassandra/db/guardrails/GuardrailColumnValueSizeTest.java
```diff
@@ -0,0 +1,650 @@
...
+    public void testSimplePartitionKey() throws Throwable
+    {
+        createTable("CREATE TABLE %s (k text PRIMARY KEY, v int)");
+
+        // the size of primary key columns is not guarded because they already have a fixed limit of 65535B
+
+        testNoThreshold("INSERT INTO %s (k, v) VALUES (?, 0)");
+        testNoThreshold("UPDATE %s SET v = 1 WHERE k = ?");
+        testNoThreshold("DELETE v FROM %s WHERE k = ?");
+        testNoThreshold("DELETE FROM %s WHERE k = ?");
```
ekaterinadimitrova2, line 67: I was wondering about the following case - someone added some column when the guardrail was disabled... Then operator enables it with certain threshold and someone wants to delete now the column created before which seems to be over the threshold, then we can't delete it anymore, the only way would be if we raise the threshold or disable? Did I get it right?

conf/cassandra.yaml
dcapwell: `Invalid data storage: value must be non-negative` will be thrown if you do that, as you use `DataStorageSpec.LongBytesBound` which doesn't allow `null`

dcapwell (commented):
overral LGTM, will do a closer review later

conf/cassandra.yaml
ekaterinadimitrova2: I have review in progress, just haven’t submitted it yet. This is just wording, he uses null internally :-) needs to be corrected only in the text

test/unit/org/apache/cassandra/db/guardrails/GuardrailColumnValueSizeTest.java
```diff
@@ -0,0 +1,650 @@
...
+import org.apache.cassandra.db.marshal.MapType;
+import org.apache.cassandra.db.marshal.SetType;
+
+import static java.lang.String.format;
+import static java.nio.ByteBuffer.allocate;
+
+/**
+ * Tests the guardrail for the size of column values, {@link Guardrails#columnValueSize}.
+ */
+public class GuardrailColumnValueSizeTest extends ThresholdTester
```
ekaterinadimitrova2, line 42: Wondering, because we use null value, shall we add a test that will always confirm the nulls are properly loading from yaml and map, even if we don't use converters. I mean in case the Config value is changed to something else in time and people forget we should be able to assign null (currently we can't if in Config we have it with some other value and it Is not marked @Nullable). More like regressions prevention if in time we decide to enable it with some value, WDYT? I am referring to these tests - https://github.com/apache/cassandra/blob/bd08798c1385ddb132701d0f5c200066e3329d77/test/unit/org/apache/cassandra/config/YamlConfigurationLoaderTest.java#L115-L153
We can change them not to be only where we use Converters, WDYT?

conf/cassandra.yaml
adelapena: Right, that comment should come from an older version of the patch. Changed to `The two thresholds default to null to disable`.

conf/cassandra.yaml
adelapena: That method was moved from `FBUtilities` to `ByteBufferUtil` by CASSANDRA-1714, in times of 0.7.0-rc4. I have updated the comment to point to the right methods.

As for the whole explanation about where that limitation in size comes from, I don't really now the full history but it seems it was introduced by CASSANDRA-767 for 0.7.0-beta1. That was 12 years ago, while adopting byte arrays for representing keys instead of Java strings.

test/unit/org/apache/cassandra/db/guardrails/GuardrailColumnValueSizeTest.java
adelapena: Yep, changed there and also in `testComplexClustering`.

test/unit/org/apache/cassandra/db/guardrails/GuardrailColumnValueSizeTest.java
adelapena: I think that regular columns over the threshold can be deleted because the required tombstone doesn't contain the value. The exception to this is deleting not-frozen set or map elements when that element is over the threshold. 

Sets and maps store their items in cell names. So deleting a set/map item would generate a tombstone containing the value of the item to be deleted. Such tombstone will be rejected by the guardrail, [here](https://github.com/apache/cassandra/blob/68c37ede1e1ade868ec6e6a9081b48e73db3f124/src/java/org/apache/cassandra/cql3/UpdateParameters.java#L147-L148). The only way to remove it would be either modifying the guardrail or deleting the entire collection. Deleting the entire row of the partition would also be possible.

If we wouldn't have that check for tombstone names it would be possible to insert large tombstones (over the threshold) into sets and maps, even if the element has never been written before. That is, users could write huge tombstones even if the guardrail has been active from day zero. This is tested [here](https://github.com/apache/cassandra/blob/68c37ede1e1ade868ec6e6a9081b48e73db3f124/test/unit/org/apache/cassandra/db/guardrails/GuardrailColumnValueSizeTest.java#L247),  [here](https://github.com/apache/cassandra/blob/68c37ede1e1ade868ec6e6a9081b48e73db3f124/test/unit/org/apache/cassandra/db/guardrails/GuardrailColumnValueSizeTest.java#L261) and also [here](https://github.com/apache/cassandra/blob/68c37ede1e1ade868ec6e6a9081b48e73db3f124/test/unit/org/apache/cassandra/db/guardrails/GuardrailColumnValueSizeTest.java#L295).

ekaterinadimitrova2: Yes, my bad, not for regular columns. Do we want to document it for sets and maps?

adelapena: Yes, I think we should mention it since the behaviour for sets and maps is a bit tricky. I have added a brief explanation in the yaml and in the place where the guardrail is checked.

test/unit/org/apache/cassandra/db/guardrails/GuardrailColumnValueSizeTest.java
adelapena: Makes sense. If we are going to extend those tests to cover not only special converters we should include quite a bunch of properties, not only the two new properties that are proposed here. So I think I would prefer to do it in a separate ticket covering all existing properties, and maybe also for 4.1 if we can do it on time.

By the way, it's not the same since it doesn't involve parsing, but `ThresholdTester` has some checks around the default value, [here](https://github.com/apache/cassandra/blob/5627ed1c8e9e264031675d3283de67b0f41af534/test/unit/org/apache/cassandra/db/guardrails/ThresholdTester.java#L273-L274).

ekaterinadimitrova2: Agreed, I totally didn’t mean tests for all properties here, only for the new 2. Actually, I do not think we will change defaults in 4.1 in a patch release, etc so my concern is more for trunk to add such tests. We normally change defaults in a major right? 
Happy to open a follow up ticket to add more tests as you suggested. Might be also nice for a new contributor maybe

test/unit/org/apache/cassandra/db/guardrails/GuardrailColumnValueSizeTest.java
ekaterinadimitrova2: LGTM, thanks!

conf/cassandra.yaml
```diff
@@ -1757,6 +1757,23 @@ drop_compact_storage_enabled: false
...
+# deleted element, independently on whether the element existed or not. That tombstone value is also guarded by this
+# guardrail, to prevent the insertion of tombstones over the threshold. The downside is that enabling or raising this
+# threshold can prevent users from deleting set/map elements that were written when the guardrail was disabled or with a
+# lower value. Deleting the entire column, row or partition is always allowed, since the tombstones created for those
+# operations don't contain the CQL column values.
+# This guardrail is different to max_value_size. max_value_size is checked when deserializing any value to detect
+# sstable corruption, whereas this guardrail is checked on the CQL layer at write time to reject regular user queries
+# inserting too large columns.
+# The two thresholds default to null to disable.
+# column_value_size_warn_threshold:
```
smiklosovic, line 1785: I had a hard time to realise what is the actual unit of size we are supporting here. I guess it is in bytes but it is not obvious from this comment. Maybe we might explicitly mention it?

ekaterinadimitrova2: All units, not only bytes. We have marked min unit only for old properties where we couldn't support anything less than the mentioned unit and the larger units. Considering I wrote a very detailed doc about config and what we support I do not think that we should add anything here really? We have it stated what units we support for Data Storage in the docs and we marked those which are exception due to legacy reasons in the yaml. I talk about this doc - https://cassandra.apache.org/doc/trunk/cassandra/new/configuration.html. There is also a blog post. Please let me know if you think there is anything unclear,  I will be making a few updates tomorrow to the mentioned doc and I can add/edit anything that might seem confusing. Thanks

ekaterinadimitrova2: Ha....I actually noticed I stated the min unit, but not explicitly in that doc - "these are all units supported per type of parameter". Good catch. I see only the Duration ones. They are in the ticket and other places but we need them listed explicitly in a table in that doc I think. Thanks! I will add it tomorrow

adelapena: Just added a comment about the min unit in the yaml.

dcapwell (approved):

PR #1687, opened 2022-06-16
CASSANDRA-15968 4.0

smiklosovic:


src/java/org/apache/cassandra/gms/Gossiper.java
```diff
@@ -2015,8 +2010,14 @@ public void addLocalApplicationStates(List<Pair<ApplicationState, VersionedValue
...
 
-    public void stop()
+    public synchronized void stop()
     {
+        if (isShuttingDown)
+        {
+            logger.info("Gossiper was already shutdown.");
+            return;
+        }
+        isShuttingDown = true;
```
dcapwell: `org.apache.cassandra.gms.Gossiper#start(int)` can be called via jmx and is a normal operational activity, same for `stop`.

We need to revert this config when `start` is called else this will break users

src/java/org/apache/cassandra/hints/HintsService.java
```diff
@@ -242,7 +242,10 @@ public void resumeDispatch()
     public synchronized void shutdownBlocking() throws ExecutionException, InterruptedException
     {
         if (isShutDown)
-            throw new IllegalStateException("HintsService has already been shut down");
+        {
+            logger.info("Hint service was already shut down.");
```
dcapwell, line 246: I think I handle this in Instance on trunk, but good to just not fail

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -529,6 +529,12 @@ public void listen()
         inboundSockets.open();
     }
 
+    public void unsafeResetListen()
+    {
+        inboundSockets.open();
+        isShuttingDown = false;
```
dcapwell: why not just move this to `listen`?  supporting a new method that is similar can be brittle

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -529,6 +529,12 @@ public void listen()
...
+    }
+
     public void waitUntilListening() throws InterruptedException
```
dcapwell, line 532: shouldn't we set false here as well?

dcapwell (changes requested):
mostly small changes, overall LGTM

pmcfadin:

PR #1695, opened 2022-06-17
CASSANDRA-17695: Fix AbstractCell#toString for tombstoned multicell CollectionTypes

frankgh:
Calling `AbstractCell#toString` on a tombstoned multi-cell `CollectionType` for the `date`, `duration`,
`smallint`, `time`, and `tinyint` types produces a `MarshalException`. The root cause of the issue
occurs when `AbstractCell#toString` in turns calls `AbstractType#getString(V value, ValueAccessor<V> acccesor)`.
The `TypeSerializer` performs validation of the value. However, because this value is tombstoned
the size of the value is 0 which doesn't match the type value. The failed check throws a `MarshalException`.
In this commit, we check whether the Cell is tombstoned and add the "<tombstone>" string to the
output of the `AbstractCell#toString` method for tombstoned cells.

### DIFF
--- a/test/unit/org/apache/cassandra/db/CellTest.java
+++ b/test/unit/org/apache/cassandra/db/CellTest.java
@@ -19,6 +19,7 @@
 package org.apache.cassandra.db;
 
 import java.nio.ByteBuffer;
+import java.util.ArrayList;
 import java.util.List;
 
 import com.google.common.collect.Lists;
@@ -29,6 +30,7 @@ import org.junit.Test;
 
 import org.apache.cassandra.SchemaLoader;
 import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.cql3.CQL3Type;
 import org.apache.cassandra.cql3.ColumnIdentifier;
 import org.apache.cassandra.cql3.FieldIdentifier;
 import org.apache.cassandra.db.marshal.*;
@@ -41,7 +43,7 @@ import org.apache.cassandra.serializers.MarshalException;
 import org.apache.cassandra.utils.ByteBufferUtil;
 import org.apache.cassandra.utils.FBUtilities;
 
-import static java.util.Arrays.*;
+import static java.util.Arrays.asList;
 
 public class CellTest
 {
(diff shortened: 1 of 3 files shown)


test/unit/org/apache/cassandra/db/CellTest.java
```diff
@@ -103,6 +105,26 @@ public void testConflictingTypeEquality()
...
+    public void testUnmarshallableInMulticellCollection()
+    {
+        List<CQL3Type.Native> unmarshallableTypes = new ArrayList<>();
+        for (CQL3Type.Native nativeType : CQL3Type.Native.values())
+        {
+            ColumnMetadata c = fakeColumn("c", MapType.getInstance(Int32Type.instance, nativeType.getType(), true));
+            BufferCell cell = BufferCell.tombstone(c, 0, 4, CellPath.create(ByteBufferUtil.bytes(4)));
+            try
+            {
+                Assert.assertTrue(cell.toString().contains("tombstone"));
```
yifan-c: nit: would be better if asserting on the format, instead of just contains.

yifan-c (commented):
LGTM

test/unit/org/apache/cassandra/db/CellTest.java
frankgh: thanks for the review, @yifan-c. I pushed a new commit addressing your feedback

maedhroz (approved):

test/unit/org/apache/cassandra/db/CellTest.java
```diff
@@ -103,6 +105,26 @@ public void testConflictingTypeEquality()
...
+            try
+            {
+                Assert.assertEquals("expected #toString failed for type " + nativeType, "[c[4]=<tombstone> ts=0 ldt=4]", cell.toString());
+            }
+            catch (MarshalException m)
+            {
+                unmarshallableTypes.add(nativeType);
+            }
+        }
+        Assert.assertTrue(unmarshallableTypes.isEmpty());
```
maedhroz, line 125: nit: Might get a better error message with AssertJ

```
assertThat(unmarshallableTypes).isEmpty();
```

frankgh: yeah , I thought about moving to AssertJ, but this class is already using `org.junit.Assert` so decided against it

PR #1696, opened 2022-06-20
CASSANDRA-17650: Add validation for batch_size_fail_threshold

akin-tekeoglu:
I also fixed the `batch_size_warn_threshold ` test. It was checking against `column_index_size` before.

[Task](https://issues.apache.org/jira/browse/CASSANDRA-17650)

### DIFF
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -559,7 +559,6 @@ public class DatabaseDescriptor
         checkValidForByteConversion(conf.column_index_size, "column_index_size");
         checkValidForByteConversion(conf.column_index_cache_size, "column_index_cache_size");
         checkValidForByteConversion(conf.batch_size_warn_threshold, "batch_size_warn_threshold");
-        checkValidForByteConversion(conf.batch_size_fail_threshold, "batch_size_fail_threshold");
 
         if (conf.native_transport_max_negotiable_protocol_version != null)
             logger.warn("The configuration option native_transport_max_negotiable_protocol_version has been deprecated " +
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -1638,7 +1638,7 @@ public class DatabaseDescriptor
 
     public static long getBatchSizeFailThreshold()
     {
-        return conf.batch_size_fail_threshold.toBytes();
+        return conf.batch_size_fail_threshold.toBytesInLong();
     }
 
     public static int getBatchSizeFailThresholdInKiB()
@@ -1658,13 +1658,7 @@ public class DatabaseDescriptor
 
     public static void setBatchSizeFailThresholdInKiB(int threshold)
     {
-        conf.batch_size_fail_threshold = createIntKibibyteBoundAndEnsureItIsValidForByteConversion(threshold,"batch_size_fail_threshold");
-    }
-
-    private static DataStorageSpec.IntKibibytesBound createIntKibibyteBoundAndEnsureItIsValidForByteConversion(int byteValue, String configName){
-        DataStorageSpec.IntKibibytesBound intKibibytesBound = new DataStorageSpec.IntKibibytesBound(byteValue);
-        checkValidForByteConversion(intKibibytesBound, configName);
-        return intKibibytesBound;
+        conf.batch_size_fail_threshold = new DataStorageSpec.IntKibibytesBound(threshold);
     }
 
     public static Collection<String> getInitialTokens()
@@ -3708,12 +3702,19 @@ public class DatabaseDescriptor
         commitLogSegmentMgrProvider = provider;
     }
 
+    private static DataStorageSpec.IntKibibytesBound createIntKibibyteBoundAndEnsureItIsValidForByteConversion(int kibibytes, String propertyName)
+    {
+        DataStorageSpec.IntKibibytesBound intKibibytesBound = new DataStorageSpec.IntKibibytesBound(kibibytes);
+        checkValidForByteConversion(intKibibytesBound, propertyName);
+        return intKibibytesBound;
+    }
+
     /**
      * Ensures passed in configuration value is positive and will not overflow when converted to Bytes
      */
     private static void checkValidForByteConversion(final DataStorageSpec.IntKibibytesBound value, String name)
     {
-        long valueInBytes = value.toBytes();
+        long valueInBytes = value.toBytesInLong();
         if (valueInBytes < 0 || valueInBytes > Integer.MAX_VALUE - 1)
         {
             throw new ConfigurationException(String.format("%s must be positive value <= %dB, but was %dB",
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -559,6 +559,7 @@ public class DatabaseDescriptor
         checkValidForByteConversion(conf.column_index_size, "column_index_size");
         checkValidForByteConversion(conf.column_index_cache_size, "column_index_cache_size");
         checkValidForByteConversion(conf.batch_size_warn_threshold, "batch_size_warn_threshold");
+        checkValidForByteConversion(conf.batch_size_fail_threshold, "batch_size_fail_threshold");
 
         if (conf.native_transport_max_negotiable_protocol_version != null)
             logger.warn("The configuration option native_transport_max_negotiable_protocol_version has been deprecated " +
@@ -1607,9 +1608,7 @@ public class DatabaseDescriptor
 
     public static void setColumnIndexSize(int val)
     {
-        DataStorageSpec.IntKibibytesBound memory = new DataStorageSpec.IntKibibytesBound(val);
-        checkValidForByteConversion(memory, "column_index_size");
-        conf.column_index_size = new DataStorageSpec.IntKibibytesBound(val);
+        conf.column_index_size =  createIntKibibyteBoundAndEnsureItIsValidForByteConversion(val,"column_index_size");
     }
 
     public static int getColumnIndexCacheSize()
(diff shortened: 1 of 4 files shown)


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1656,14 +1653,18 @@ public static int getUnloggedBatchAcrossPartitionsWarnThreshold()
...
+        conf.batch_size_warn_threshold = createIntKibibyteBoundAndEnsureItIsValidForByteConversion(threshold,"batch_size_warn_threshold");
     }
 
     public static void setBatchSizeFailThresholdInKiB(int threshold)
     {
-        conf.batch_size_fail_threshold = new DataStorageSpec.IntKibibytesBound(threshold);
+        conf.batch_size_fail_threshold = createIntKibibyteBoundAndEnsureItIsValidForByteConversion(threshold,"batch_size_fail_threshold");
+    }
+
+    private static DataStorageSpec.IntKibibytesBound createIntKibibyteBoundAndEnsureItIsValidForByteConversion(int byteValue, String configName){
```
adelapena: Nit: the final brace should be on a new line, K&R style:
```suggestion
    private static DataStorageSpec.IntKibibytesBound createIntKibibyteBoundAndEnsureItIsValidForByteConversion(int byteValue, String configName)
    {
```
Also, the `byteValue` argument is meant to contain a value in kibibytes, so we should probably name it `kibibyteValue`. Or maybe just `kibibytes`, as it's named in the`DataStorageSpec.IntKibibytesBound` constructor that we use immediately below:
```suggestion
    private static DataStorageSpec.IntKibibytesBound createIntKibibyteBoundAndEnsureItIsValidForByteConversion(int kibibytes, String configName)
    {
```
I would also rename the `configName` argument to `propertyName`, I think:
```suggestion
    private static DataStorageSpec.IntKibibytesBound createIntKibibyteBoundAndEnsureItIsValidForByteConversion(int kibibytes, String propertyName)
    {
```
I wonder if we should put this new method and the related `checkValidForByteConversion` closer to each other in the file, wdyt?

One last thought about this is that perhaps we should move this validation into `DataStorageSpec.IntKibibytesBound`, maybe with a validation method or with a static constructor performing the validation. But we don't have to do that during this ticket.

test/unit/org/apache/cassandra/config/DatabaseDescriptorTest.java
```diff
@@ -326,7 +326,24 @@ public void testExceptionsForInvalidConfigValues() {
             fail("Should have received a ConfigurationException batch_size_warn_threshold = 2GiB");
         }
         catch (ConfigurationException ignored) { }
-        Assert.assertEquals(4096, DatabaseDescriptor.getColumnIndexSize());
```
adelapena, line 329: Good catch!

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
ekaterinadimitrova2: batch_size_fail_threshold doesn't need the validation and I assumed this part of the patch will be removed.
Anyway, I am not against adding a nice utility method and agree with @adelapena's review comments.

> One last thought about this is that perhaps we should move this validation into DataStorageSpec.IntKibibytesBound, maybe with a validation method or with a static constructor performing the validation. But we don't have to do that during this ticket. 

Didn't we intentionally leave it out of those classes and opted out of custom static methods and constructors as it is a corner case for only a few parameters?  I thought we didn't want to encourage too much customizations in those classes to ensure we keep them clean, only in case we consider this will be a very common case.

adelapena: I think the widening that was used before instead of validation was broken by the adoption of `DataStorageSpec.IntKibibytesBound.toBytes`, which returns an `int`. So we have to either add new conversion method or also use validation here. Probably the former for compatibility, although I'm not sure.

As for moving validation logic into `DataStorageSpec`, I was thinking not in moving the entire method as it is but the logic to verify that the value can be converted to a lesser unit without losing precision. Perhaps a method returning a boolean saying whether the conversion to a unit is going to produce precision loss, so the validator on `DatabaseDescriptor` can use it. That would be a single utility method on `DataStorageSpec`, not extending to their child classes. But as I mentioned, I think we don't have to do that, if we do it, here.

ekaterinadimitrova2: Thankfully we don't need to handle the precision issue here as going to smaller unit we use multiplication. Overflow is the only issue we need to handle. 
Yes, both` DataStorageSpec.IntKibibytesBound` and `DataStorageSpec.IntMebibytesBound` will need something like `toBytesInLong`.

It seems that `batch_size_fail_threshold` is the only parameter of type `IntKibibytesBound` that suffers from this issue. 
Now, I went ahead and did a further check for the other properties that are widen to long as it seems this bug was introduced when we removed the static methods from `DataStorageSpec`.
We will need to add `toBytesInLong` also for `IntMebibytesBound` to cover
```
getMinFreeSpacePerDriveInBytes() / min_free_space_per_drive,
compaction_large_partition_warning_threshold / getMinFreeSpacePerDriveInBytes(), getMaxHintsFileSize() / max_hints_file_size
```

ekaterinadimitrova2: Not sure whether we want to cover all this in this one ticket or split

ekaterinadimitrova2: One more thing on my mind. I think further to adding comment it will be great if we add some unit tests for the bounds of those properties which will prevent people from regressing again in the future. It is confusing and easy to miss

akin-tekeoglu: @ekaterinadimitrova2 @adelapena I have a couple of questions

1. Why do we have `DataStorageUnit` as a constructor parameter in subclasses of `DataStorageSpec`? E.g., Isn't `IntKibibytesBound` represents values like 1 KiB, 100 KiB. Why do we have an option like this `new DataStorageSpec.IntKibibytesBound(10, MEBIBYTES)`

2. Shouldn't the `quantity` of the constructor of IntBound classes be int? It is long right now.

3. What is the upper limit of  `batch_size_fail_threshold`?

Either I missed the idea behind `DataStorageSpec`, or the class hierarchy is misleading.

ekaterinadimitrova2: This is part of the new config format where users can provide unit of their choice. Now, they cannot provide at the moment any unit smaller than the one used internally by Cassandra for a respective property because during conversion there can be lost precision. Also, we will run into more issues with the accepted upper bound. So in the case we call `DataStorageSpec.IntKibibytesBound(10, MEBIBYTES)` which says - the user wants/provides 10Mebibytes for a property which was type Int and it is used in Kibibytes by Cassandra internally. So it can't be more than Integer.MAX_VALUE-1 in kibibytes. Please refer to the Java Doc and this document - https://cassandra.apache.org/doc/trunk/cassandra/new/configuration.html

`batch_size_fail_threshold `- It can't be more than `Integer.MAX_VALUE` in kibibytes. So the long widening when we return it in bytes can accommodate this. But currently `getBatchSizeFailThreshold()` uses the` IntKibibytesBound.toBytes()` which would return Int and this is a bug. We will need to create a new method `IntKibibytesBound.toBytesInLong()` to be used by `getBatchSizeFailThreshold() `

akin-tekeoglu: I tried to fix the issues that you mentioned before as well. Please retake a look.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -559,6 +559,7 @@ else if (conf.repair_session_space.toMebibytes() > (int) (Runtime.getRuntime().m
         checkValidForByteConversion(conf.column_index_size, "column_index_size");
         checkValidForByteConversion(conf.column_index_cache_size, "column_index_cache_size");
         checkValidForByteConversion(conf.batch_size_warn_threshold, "batch_size_warn_threshold");
+        checkValidForByteConversion(conf.batch_size_fail_threshold, "batch_size_fail_threshold");
```
ekaterinadimitrova2: As we opted in for restoring the widening, I believe this check is not needed and expected anymore. It will also reduce the established upper bound from the previous versions.
It should have upper bound Integer.MAX_VALUE-1 in kibibytes, not bytes and the kibibytes upper value is already handled in the constructor.

test/unit/org/apache/cassandra/config/DatabaseDescriptorTest.java
```diff
@@ -326,7 +326,8 @@ public void testExceptionsForInvalidConfigValues() {
...
+        Assert.assertEquals(5120, DatabaseDescriptor.getBatchSizeWarnThreshold());
+
     }
```
ekaterinadimitrova2, line 338: I suggest we add below test in order to ensure no more regressions in the future in this area. We can also add a comment about that as Andres suggested on the ticket, to be more clear what is this and where it came from. (for legacy reasons we do widen to long instead of the check)

```
    @Test
    public void testWidenToLongInBytes() throws ConfigurationException
    {
        DatabaseDescriptor.setBatchSizeFailThresholdInKiB(2147483646);
        Assert.assertEquals((2147483646L * 1024), DatabaseDescriptor.getBatchSizeFailThreshold());
    }
```

ekaterinadimitrova2 (commented):
Overall looks good to me, you forgot to remove the check for batch_size_fail_threshold now when we figured it out that widening to long is what was chosen and used for this property in previous releases.

I also suggested one test that can be guarding against the same regression in the future.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
akin-tekeoglu: Removed the check

test/unit/org/apache/cassandra/config/DatabaseDescriptorTest.java
akin-tekeoglu: Added the test

PR #1697, opened 2022-06-20
fix typos and clarify "architecture" docs

zzantozz:
I'm reading the docs for the first time, and several typos stand
out. I've suggested mostly minor fixes here. I don't think any can
change the meaning of the content, but some of it was a bit ambiguous.


smiklosovic:
thank you! this work will be included in https://github.com/apache/cassandra/pull/2062

PR #1701, opened 2022-06-22
CASSANDRA-17677: Fix BulkLoader to load entireSSTableThrottle and entireSSTableInterDcThrottle

frankgh:
`entire_sstable_stream_throughput_outbound` and `entire_sstable_inter_dc_stream_throughput_outbound` were
introduced in CASSANDRA-17065. They were added to the LoaderOptions class but they are not being used in
the BulkLoader class. In this commit, we make use of the entire_sstable streaming options in BulkLoader.
In addittion, the `throttle` and `interDcThrottle` are being passed as MB/s but they should be MiB/s.
This commit fixes the verbiage as well as setting the correct value for those properties in BulkLoader.

### DIFF
--- a/doc/modules/cassandra/pages/operating/bulk_loading.adoc
+++ b/doc/modules/cassandra/pages/operating/bulk_loading.adoc
@@ -80,13 +80,23 @@ The following options are supported, with `-d,--nodes <initial hosts>` required:
 -d,--nodes <initial hosts>                                   Required.
                                                              Try to connect to these hosts (comma separated) initially for ring information
 
+-emib,--entire-sstable-throttle-mib <throttle-mib>           Entire SSTable throttle
+                                                             speed in MiB/s (default 0 for unlimited).
+
+-eidctmib,--entire-sstable-inter-dc-throttle-mib <inter-dc-throttle-mib>
+                                                             Entire SSTable inter-datacenter throttle
+                                                             speed in MiB/s (default 0 for unlimited).
+
 -f,--conf-path <path to config file>                         cassandra.yaml file path for streaming throughput and client/server SSL.
 
 -h,--help                                                    Display this help message
 
 -i,--ignore <NODES>                                          Don't stream to this (comma separated) list of nodes
 
--idct,--inter-dc-throttle <inter-dc-throttle>                Inter-datacenter throttle speed in Mbits (default unlimited)
+-idct,--inter-dc-throttle <inter-dc-throttle>                (deprecated) Inter-datacenter throttle speed in Mbits (default 0 for unlimited)
+                                                             Use -idctmib / --inter-dc-throttle-mib instead
+
+-idctmib,--inter-dc-throttle-mib <inter-dc-throttle-mib>     Inter-datacenter throttle speed in MiB/s (default 0 for unlimited)
 
 -k,--target-keyspace <target keyspace name>                  Target
                                                              keyspace name
@@ -111,8 +121,10 @@ The following options are supported, with `-d,--nodes <initial hosts>` required:
                                                              for TLS internode communication (default 7001)
 -st,--store-type <STORE-TYPE>                                Client SSL:
                                                              type of store
--t,--throttle <throttle>                                     Throttle
-                                                             speed in Mbits (default unlimited)
+-t,--throttle <throttle>                                     (deprecated) Throttle speed in Mbits (default 0 for unlimited)
+                                                             Use -tmib / --throttle-mib instead
+-tmib,--throttle-mib <throttle-mib>                          Throttle
+                                                             speed in MiB/s (default 0 for unlimited)
 -ts,--truststore <TRUSTSTORE>                                Client SSL:
                                                              full path to truststore
 -tspw,--truststore-password <TRUSTSTORE-PASSWORD>            Client SSL:
--- a/doc/modules/cassandra/pages/operating/bulk_loading.adoc
+++ b/doc/modules/cassandra/pages/operating/bulk_loading.adoc
@@ -80,10 +80,10 @@ The following options are supported, with `-d,--nodes <initial hosts>` required:
 -d,--nodes <initial hosts>                                   Required.
                                                              Try to connect to these hosts (comma separated) initially for ring information
 
--emib,--entire-sstable-throttle-mib <throttle-mib>           Entire SSTable throttle
+--entire-sstable-throttle-mib <throttle-mib>                 Entire SSTable throttle
                                                              speed in MiB/s (default 0 for unlimited).
 
--eidctmib,--entire-sstable-inter-dc-throttle-mib <inter-dc-throttle-mib>
+--entire-sstable-inter-dc-throttle-mib                       <inter-dc-throttle-mib>
                                                              Entire SSTable inter-datacenter throttle
                                                              speed in MiB/s (default 0 for unlimited).
 
@@ -96,7 +96,7 @@ The following options are supported, with `-d,--nodes <initial hosts>` required:
 -idct,--inter-dc-throttle <inter-dc-throttle>                (deprecated) Inter-datacenter throttle speed in Mbits (default 0 for unlimited)
                                                              Use -idctmib / --inter-dc-throttle-mib instead
 
--idctmib,--inter-dc-throttle-mib <inter-dc-throttle-mib>     Inter-datacenter throttle speed in MiB/s (default 0 for unlimited)
+--inter-dc-throttle-mib <inter-dc-throttle-mib>              Inter-datacenter throttle speed in MiB/s (default 0 for unlimited)
 
 -k,--target-keyspace <target keyspace name>                  Target
                                                              keyspace name
@@ -123,7 +123,7 @@ The following options are supported, with `-d,--nodes <initial hosts>` required:
                                                              type of store
 -t,--throttle <throttle>                                     (deprecated) Throttle speed in Mbits (default 0 for unlimited)
                                                              Use -tmib / --throttle-mib instead
--tmib,--throttle-mib <throttle-mib>                          Throttle
+--throttle-mib <throttle-mib>                                Throttle
                                                              speed in MiB/s (default 0 for unlimited)
 -ts,--truststore <TRUSTSTORE>                                Client SSL:
                                                              full path to truststore
--- a/doc/modules/cassandra/pages/operating/bulk_loading.adoc
+++ b/doc/modules/cassandra/pages/operating/bulk_loading.adoc
@@ -93,8 +93,8 @@ The following options are supported, with `-d,--nodes <initial hosts>` required:
 
 -i,--ignore <NODES>                                          Don't stream to this (comma separated) list of nodes
 
--idct,--inter-dc-throttle <inter-dc-throttle>                (deprecated) Inter-datacenter throttle speed in Mbits (default 0 for unlimited)
-                                                             Use -idctmib / --inter-dc-throttle-mib instead
+-idct,--inter-dc-throttle <inter-dc-throttle>                (deprecated) Inter-datacenter throttle speed in Mbits (default 0 for unlimited).
+                                                             Use --inter-dc-throttle-mib instead.
 
 --inter-dc-throttle-mib <inter-dc-throttle-mib>              Inter-datacenter throttle speed in MiB/s (default 0 for unlimited)
 
@@ -121,8 +121,8 @@ The following options are supported, with `-d,--nodes <initial hosts>` required:
                                                              for TLS internode communication (default 7001)
 -st,--store-type <STORE-TYPE>                                Client SSL:
                                                              type of store
--t,--throttle <throttle>                                     (deprecated) Throttle speed in Mbits (default 0 for unlimited)
-                                                             Use -tmib / --throttle-mib instead
+-t,--throttle <throttle>                                     (deprecated) Throttle speed in Mbits (default 0 for unlimited).
+                                                             Use --throttle-mib instead.
 --throttle-mib <throttle-mib>                                Throttle
                                                              speed in MiB/s (default 0 for unlimited)
 -ts,--truststore <TRUSTSTORE>                                Client SSL:
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -2099,6 +2099,11 @@ public class DatabaseDescriptor
         conf.stream_throughput_outbound = new DataRateSpec.LongBytesPerSecondBound(value, MEBIBYTES_PER_SECOND);
     }
 
+    public static void setStreamThroughputOutboundBytesPerSec(long value)
+    {
+        conf.stream_throughput_outbound = new DataRateSpec.LongBytesPerSecondBound(value, BYTES_PER_SECOND);
+    }
+
     public static void setStreamThroughputOutboundMegabitsPerSec(int value)
     {
         conf.stream_throughput_outbound = DataRateSpec.LongBytesPerSecondBound.megabitsPerSecondInBytesPerSecond(value);
@@ -2157,6 +2162,11 @@ public class DatabaseDescriptor
                                                Integer.MAX_VALUE + " in megabits/s");
 
         conf.inter_dc_stream_throughput_outbound = new DataRateSpec.LongBytesPerSecondBound(value, MEBIBYTES_PER_SECOND);
+    }
+
+    public static void setInterDCStreamThroughputOutboundBytesPerSec(long value)
+    {
+        conf.inter_dc_stream_throughput_outbound = new DataRateSpec.LongBytesPerSecondBound(value, BYTES_PER_SECOND);
     }
 
     public static void setInterDCStreamThroughputOutboundMegabitsPerSec(int value)
--- a/src/java/org/apache/cassandra/tools/BulkLoader.java
+++ b/src/java/org/apache/cassandra/tools/BulkLoader.java
@@ -23,20 +23,26 @@ import java.util.Set;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLEngine;
 
-import com.datastax.driver.core.AuthProvider;
-import com.datastax.driver.core.RemoteEndpointAwareJdkSSLOptions;
-import com.datastax.driver.core.SSLOptions;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.Multimap;
 import org.apache.commons.cli.Option;
 import org.apache.commons.cli.Options;
 
+import com.datastax.driver.core.AuthProvider;
+import com.datastax.driver.core.RemoteEndpointAwareJdkSSLOptions;
+import com.datastax.driver.core.SSLOptions;
 import com.datastax.shaded.netty.channel.socket.SocketChannel;
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.config.EncryptionOptions;
 import org.apache.cassandra.io.sstable.SSTableLoader;
 import org.apache.cassandra.security.SSLFactory;
-import org.apache.cassandra.streaming.*;
+import org.apache.cassandra.streaming.ProgressInfo;
+import org.apache.cassandra.streaming.SessionInfo;
+import org.apache.cassandra.streaming.StreamEvent;
+import org.apache.cassandra.streaming.StreamEventHandler;
+import org.apache.cassandra.streaming.StreamResultFuture;
+import org.apache.cassandra.streaming.StreamState;
+import org.apache.cassandra.streaming.StreamingChannel;
 import org.apache.cassandra.utils.FBUtilities;
 import org.apache.cassandra.utils.JVMStabilityInspector;
 import org.apache.cassandra.utils.NativeSSTableLoaderClient;
@@ -46,7 +52,7 @@ import static org.apache.cassandra.utils.Clock.Global.nanoTime;
 
 public class BulkLoader
 {
-    public static void main(String args[]) throws BulkLoadException
+    public static void main(String[] args) throws BulkLoadException
     {
         LoaderOptions options = LoaderOptions.builder().parseArgs(args).build();
         load(options);
@@ -67,9 +73,11 @@ public class BulkLoader
                         handler,
                         options.connectionsPerHost,
                         options.targetKeyspace);
-        DatabaseDescriptor.setStreamThroughputOutboundMegabitsPerSec(options.throttle);
-        DatabaseDescriptor.setInterDCStreamThroughputOutboundMegabitsPerSec(options.interDcThrottle);
-        StreamResultFuture future = null;
+        DatabaseDescriptor.setStreamThroughputOutboundBytesPerSec(options.throttleBytes);
+        DatabaseDescriptor.setInterDCStreamThroughputOutboundBytesPerSec(options.interDcThrottleBytes);
+        DatabaseDescriptor.setEntireSSTableStreamThroughputOutboundMebibytesPerSec(options.entireSSTableThrottleMebibytes);
+        DatabaseDescriptor.setEntireSSTableInterDCStreamThroughputOutboundMebibytesPerSec(options.entireSSTableInterDcThrottleMebibytes);
+        StreamResultFuture future;
 
         ProgressIndicator indicator = new ProgressIndicator();
         try
@@ -120,7 +128,7 @@ public class BulkLoader
     // Return true when everything is at 100%
     static class ProgressIndicator implements StreamEventHandler
     {
-        private long start;
+        private final long start;
         private long lastProgress;
         private long lastTime;
 
@@ -219,7 +227,7 @@ public class BulkLoader
                 }
                 sb.append(" (avg: ").append(FBUtilities.prettyPrintMemoryPerSecond(totalProgress, time - start)).append(")");
 
-                System.out.println(sb.toString());
+                System.out.println(sb);
             }
         }
 
@@ -241,7 +249,7 @@ public class BulkLoader
             sb.append(String.format("   %-24s: %-10s%n", "Total duration ", durationMS + " ms"));
             sb.append(String.format("   %-24s: %-10s%n", "Average transfer rate ", FBUtilities.prettyPrintMemoryPerSecond(lastProgress, end - start)));
             sb.append(String.format("   %-24s: %-10s%n", "Peak transfer rate ",  FBUtilities.prettyPrintMemoryPerSecond(peak)));
-            System.out.println(sb.toString());
+            System.out.println(sb);
         }
     }
 
--- a/src/java/org/apache/cassandra/tools/LoaderOptions.java
+++ b/src/java/org/apache/cassandra/tools/LoaderOptions.java
@@ -22,25 +22,37 @@ package org.apache.cassandra.tools;
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
-import java.net.*;
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
+import java.net.MalformedURLException;
+import java.net.UnknownHostException;
 import java.util.HashSet;
 import java.util.Set;
 
 import com.google.common.base.Throwables;
 import com.google.common.net.HostAndPort;
+import org.apache.commons.cli.CommandLine;
+import org.apache.commons.cli.CommandLineParser;
+import org.apache.commons.cli.GnuParser;
+import org.apache.commons.cli.HelpFormatter;
+import org.apache.commons.cli.Options;
+import org.apache.commons.cli.ParseException;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
-import org.apache.cassandra.config.*;
+import com.datastax.driver.core.AuthProvider;
+import com.datastax.driver.core.PlainTextAuthProvider;
+import org.apache.cassandra.config.Config;
+import org.apache.cassandra.config.DataRateSpec;
+import org.apache.cassandra.config.EncryptionOptions;
+import org.apache.cassandra.config.YamlConfigurationLoader;
 import org.apache.cassandra.exceptions.ConfigurationException;
 import org.apache.cassandra.io.util.File;
 import org.apache.cassandra.locator.InetAddressAndPort;
 import org.apache.cassandra.tools.BulkLoader.CmdLineOptions;
 
-import com.datastax.driver.core.AuthProvider;
-import com.datastax.driver.core.PlainTextAuthProvider;
-import org.apache.commons.cli.*;
-import org.apache.commons.lang3.StringUtils;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import static org.apache.cassandra.config.DataRateSpec.DataRateUnit.MEBIBYTES_PER_SECOND;
 
 public class LoaderOptions
 {
@@ -60,10 +72,25 @@ public class LoaderOptions
     public static final String IGNORE_NODES_OPTION = "ignore";
     public static final String CONNECTIONS_PER_HOST = "connections-per-host";
     public static final String CONFIG_PATH = "conf-path";
+
+    /**
+     * Throttle defined in megabits per second. CASSANDRA-10637 introduced a builder and is the preferred way to
+     * provide options instead of using these constant fields.
+     * @deprecated Use {@code throttle-mib} instead
+     */
+    @Deprecated
     public static final String THROTTLE_MBITS = "throttle";
+    public static final String THROTTLE_MEBIBYTES = "throttle-mib";
+    /**
+     * Inter-datacenter throttle defined in megabits per second. CASSANDRA-10637 introduced a builder and is the
+     * preferred way to provide options instead of using these constant fields.
+     * @deprecated Use {@code inter-dc-throttle-mib} instead
+     */
+    @Deprecated
     public static final String INTER_DC_THROTTLE_MBITS = "inter-dc-throttle";
-    public static final String ENTIRE_SSTABLE_THROTTLE_MBITS = "entire-sstable-throttle";
-    public static final String ENTIRE_SSTABLE_INTER_DC_THROTTLE_MBITS = "entire-sstable-inter-dc-throttle";
+    public static final String INTER_DC_THROTTLE_MEBIBYTES = "inter-dc-throttle-mib";
+    public static final String ENTIRE_SSTABLE_THROTTLE_MEBIBYTES = "entire-sstable-throttle-mib";
+    public static final String ENTIRE_SSTABLE_INTER_DC_THROTTLE_MEBIBYTES = "entire-sstable-inter-dc-throttle-mib";
     public static final String TOOL_NAME = "sstableloader";
     public static final String TARGET_KEYSPACE = "target-keyspace";
 
@@ -85,10 +112,10 @@ public class LoaderOptions
     public final String user;
     public final String passwd;
     public final AuthProvider authProvider;
-    public final int throttle;
-    public final int interDcThrottle;
-    public final int entireSSTableThrottle;
-    public final int entireSSTableInterDcThrottle;
+    public final long throttleBytes;
+    public final long interDcThrottleBytes;
+    public final int entireSSTableThrottleMebibytes;
+    public final int entireSSTableInterDcThrottleMebibytes;
     public final int storagePort;
     public final int sslStoragePort;
     public final EncryptionOptions clientEncOptions;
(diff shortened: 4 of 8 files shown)


test/unit/org/apache/cassandra/tools/LoaderOptionsTest.java
```diff
@@ -84,38 +84,158 @@ public void testEncryptionSettings() throws Exception
...
+    }
+
+    @Test
+    public void testDeprecatedThrottleSettingsWithLongSettingNames() throws IOException
+    {
+        // Default Cassandra config
+        File config = new File(Paths.get(".", "test", "conf", "cassandra.yaml").normalize());
+        String[] args = { "--throttle", "200", "--inter-dc-throttle", "400", "-d", "127.9.9.1", "-f", config.absolutePath(), sstableDirName("legacy_sstables", "legacy_ma_simple") };
+        LoaderOptions options = LoaderOptions.builder().parseArgs(args).build();
+        // converts from megabytes to mebibytes
```
ekaterinadimitrova2:
```suggestion
        // converts from megabits to mebibytes
```

test/unit/org/apache/cassandra/tools/LoaderOptionsTest.java
```diff
@@ -84,38 +84,158 @@ public void testEncryptionSettings() throws Exception
...
+    }
+
+    @Test
+    public void testThrottleSettings() throws IOException
+    {
+        // Default Cassandra config
+        File config = new File(Paths.get(".", "test", "conf", "cassandra.yaml").normalize());
+        String[] args = { "-tmib", "200", "-idctmib", "400", "-d", "127.9.9.1", "-f", config.absolutePath(), sstableDirName("legacy_sstables", "legacy_ma_simple") };
+        LoaderOptions options = LoaderOptions.builder().parseArgs(args).build();
+        // converts from megabytes to mebibytes
```
ekaterinadimitrova2:
```suggestion
        // converts from megabits to mebibytes
```

test/unit/org/apache/cassandra/tools/LoaderOptionsTest.java
```diff
@@ -84,38 +84,158 @@ public void testEncryptionSettings() throws Exception
...
+    }
+
+    @Test
+    public void testThrottleSettingsWithLongSettingNames() throws IOException
+    {
+        // Default Cassandra config
+        File config = new File(Paths.get(".", "test", "conf", "cassandra.yaml").normalize());
+        String[] args = { "--throttle-mib", "200", "--inter-dc-throttle-mib", "400", "-d", "127.9.9.1", "-f", config.absolutePath(), sstableDirName("legacy_sstables", "legacy_ma_simple") };
+        LoaderOptions options = LoaderOptions.builder().parseArgs(args).build();
+        // converts from megabytes to mebibytes
```
ekaterinadimitrova2:
```suggestion
        // converts from megabits to mebibytes
```

test/unit/org/apache/cassandra/tools/LoaderOptionsTest.java
```diff
@@ -84,38 +84,158 @@ public void testEncryptionSettings() throws Exception
...
+    }
+
+    @Test
+    public void testDeprecatedThrottleSettings() throws IOException
+    {
+        // Default Cassandra config
+        File config = new File(Paths.get(".", "test", "conf", "cassandra.yaml").normalize());
+        String[] args = { "-t", "200", "-idct", "400", "-d", "127.9.9.1", "-f", config.absolutePath(), sstableDirName("legacy_sstables", "legacy_ma_simple") };
+        LoaderOptions options = LoaderOptions.builder().parseArgs(args).build();
+        // converts from megabytes to mebibytes
```
ekaterinadimitrova2:
```suggestion
        // converts from megabits to mebibytes
```

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -494,24 +508,46 @@ public Builder parseArgs(String cmdArgs[])
...
-                if (cmd.hasOption(THROTTLE_MBITS))
+                if (cmd.hasOption(THROTTLE_MEGABITS) && cmd.hasOption(THROTTLE_MEBIBYTES))
+                {
+                    errorMsg(String.format("Both '%s' and '%s' were provided. Please only provide one of the two options", THROTTLE_MEGABITS, THROTTLE_MEBIBYTES), options);
+                }
+
+                if (cmd.hasOption(INTER_DC_THROTTLE_MEGABITS) && cmd.hasOption(INTER_DC_THROTTLE_MEBIBYTES))
+                {
+                    errorMsg(String.format("Both '%s' and '%s' were provided. Please only provide one of the two options", INTER_DC_THROTTLE_MEGABITS, INTER_DC_THROTTLE_MEBIBYTES), options);
+                }
```
ekaterinadimitrova2: Good call! :-)

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -494,24 +508,46 @@ public Builder parseArgs(String cmdArgs[])
...
+
+                if (cmd.hasOption(THROTTLE_MEGABITS))
+                {
+                    throttleMebibytes = DataRateSpec.IntMebibytesPerSecondBound.megabitsPerSecondInMebibytesPerSecond(Integer.parseInt(cmd.getOptionValue(THROTTLE_MEGABITS)))
+                                                                               .toMebibytesPerSecondAsInt();
```
ekaterinadimitrova2: I think that this part has an issue and breaks the compatibility. The conversions between megabits and mebibytes are double numbers internally. Something very annoying to keep backward compatibility, etc. 
So the issue with this approach here is that someone will set megabits but by using `.toMebibytesPerSecondAsInt()` to set the throttling we actually round the number. This can be caught if you try to add  `--throttle` being `5megabits` in `testBulkLoader_WithArgs6`. Then if you do `assertEquals(5, DatabaseDescriptor.getStreamThroughputOutboundMegabitsPerSec());` you actually get 8. 
As it is a tool and the difference doesn't seem super big might be ok. 
The bad part is that if someone tries super small number just for testing or something, they will actually get 0 and unthrottle... That is how I identified that annoying issue when we worked on the new config framework and we switched to `double` in DataRateSpec to be precise as the `RateLimiter` gets double anyway. At the same time, of course, we do not allow the users to input `double` numbers, just `int`. The `SetGetStreamThroughputTest` tests small numbers but we didn't have this for the tool and it is very tiny detail easy to miss. :-( 
I am thinking maybe we can introduce double variable to ensure precision and no surprises to people who continue using the megabits version? WDYT?
In general probably we can just document for people this behavior and leave it as is I guess but if we want to keep the compatibility and be precise everywhere, we will need to introduce `double`. I guess this shouldn't be hard so at this point of the release I guess I would prefer just to work our the precision to be correct.

CC @yifan-c

test/unit/org/apache/cassandra/tools/LoaderOptionsTest.java
```diff
@@ -84,38 +84,158 @@ public void testEncryptionSettings() throws Exception
         assertEquals("test.jks", options.clientEncOptions.keystore);
     }
 
+    @Test
+    public void testThrottleDefaultSettings()
+    {
+        LoaderOptions options = LoaderOptions.builder().build();
+        assertEquals(0, options.throttleMebibytes);
+        assertEquals(0, options.interDcThrottleMebibytes);
```
ekaterinadimitrova2: nit: shall we add assertions also for the other two?

test/unit/org/apache/cassandra/tools/LoaderOptionsTest.java
```diff
@@ -84,38 +84,158 @@ public void testEncryptionSettings() throws Exception
...
+    public void testDeprecatedThrottleSettings() throws IOException
+    {
+        // Default Cassandra config
+        File config = new File(Paths.get(".", "test", "conf", "cassandra.yaml").normalize());
+        String[] args = { "-t", "200", "-idct", "400", "-d", "127.9.9.1", "-f", config.absolutePath(), sstableDirName("legacy_sstables", "legacy_ma_simple") };
+        LoaderOptions options = LoaderOptions.builder().parseArgs(args).build();
+        // converts from megabytes to mebibytes
+        assertEquals(24, options.throttleMebibytes);
+        assertEquals(48, options.interDcThrottleMebibytes);
+    }
```
ekaterinadimitrova2, line 106: After we fix the precision issue, I think it will be correct to check the megabits here

test/unit/org/apache/cassandra/tools/LoaderOptionsTest.java
```diff
@@ -84,38 +84,158 @@ public void testEncryptionSettings() throws Exception
...
+    @Test
+    public void testDeprecatedThrottleSettingsWithLongSettingNames() throws IOException
+    {
+        // Default Cassandra config
+        File config = new File(Paths.get(".", "test", "conf", "cassandra.yaml").normalize());
+        String[] args = { "--throttle", "200", "--inter-dc-throttle", "400", "-d", "127.9.9.1", "-f", config.absolutePath(), sstableDirName("legacy_sstables", "legacy_ma_simple") };
+        LoaderOptions options = LoaderOptions.builder().parseArgs(args).build();
+        // converts from megabytes to mebibytes
+        assertEquals(24, options.throttleMebibytes);
+        assertEquals(48, options.interDcThrottleMebibytes);
```
ekaterinadimitrova2: After we fix the precision issue, I think it will be correct to check the megabits here

ekaterinadimitrova2 (commented):
Overall looks good and thank you for adding all those tests. I found only one issue that is easy to miss if you haven't before. 
Precision problem when we set the old deprecated throttle in megabits. More info in the comments
Also, I almost forgot we need to add News.txt entry probably under Deprecation

test/unit/org/apache/cassandra/tools/LoaderOptionsTest.java
frankgh: we have another test for entire sstable default settings `testEntireSSTableDefaultSettings`. Do we want to combine them?

src/java/org/apache/cassandra/tools/LoaderOptions.java
frankgh: let me think about this. I have some ideas, but one of the options is to use a `double` type instead of `int`. Another idea is to keep two variables, one for mibs and one for MiBs. I will probably end up going with the solution that produces the more accurate result in the tests.

test/unit/org/apache/cassandra/tools/LoaderOptionsTest.java
ekaterinadimitrova2: Sorry, I missed that one it seems. Maybe we can just combine them, it's up to you.

src/java/org/apache/cassandra/tools/LoaderOptions.java
ekaterinadimitrova2: As long as the user sees only ints and input matches output - I am good :-) Probably both versions should work

src/java/org/apache/cassandra/config/DataRateSpec.java
```diff
@@ -296,7 +296,7 @@ public IntMebibytesPerSecondBound(long mebibytesPerSecond)
         // which were in megabits per second in 4.0. Do not start using it for any new properties
         public static IntMebibytesPerSecondBound megabitsPerSecondInMebibytesPerSecond(long megabitsPerSecond)
         {
-            final double MEBIBYTES_PER_MEGABIT = 0.119209289550781;
+            final double MEBIBYTES_PER_MEGABIT = 15625D / 131072D;
```
frankgh: I made these conversions more explicit, let me know if this is okay @ekaterinadimitrova2

src/java/org/apache/cassandra/config/DataRateSpec.java
```diff
@@ -395,7 +395,7 @@ public double convert(double source, DataRateUnit sourceUnit)
         };
 
         static final double MAX = Long.MAX_VALUE;
-        static final double MEGABITS_PER_MEBIBYTE = 8.388608;
+        static final double MEGABITS_PER_MEBIBYTE = 131_072D / 15_625D;
```
frankgh: another explicit conversion

src/java/org/apache/cassandra/config/DataRateSpec.java
yifan-c: It is hard to see the reasoning behind the equations.

How about define the constants for MEGABIT, MEBIBYTES, etc. in bits and derive the conversion from the constants? For example,

```java
static final double MEBIBYTES_IN_BITS = Math.pow(2, 20) * 8;
static final double MEGABIT_IN_BITS = Math.pow(10, 6); 
static final double MEBIBYTES_PER_MEGABIT = MEGABIT_IN_BITS / MEBIBYTES_IN_BITS;
... more for the other size units...
```

frankgh: good point, I've updated the PR. The only difference is that I've defined `MEBIBYTES_IN_BITS` as:
```java
static final double MEBIBYTES_IN_BITS = Math.pow(2, 23);
```

src/java/org/apache/cassandra/config/DataRateSpec.java
```diff
@@ -38,6 +38,10 @@
      * The Regexp used to parse the rate provided as String in cassandra.yaml.
      */
     private static final Pattern UNITS_PATTERN = Pattern.compile("^(\\d+)(MiB/s|KiB/s|B/s)$");
+    static final double MEBIBYTES_IN_BITS = Math.pow(2, 23);
```
ekaterinadimitrova2:
```suggestion
    static final double MEBIBYTE_IN_BITS = Math.pow(2, 23);
```

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2024,7 +2024,12 @@ public static double getStreamThroughputOutboundMebibytesPerSec()
         return conf.stream_throughput_outbound.toMebibytesPerSecond();
     }
 
-    public static void setStreamThroughputOutboundMegabitsPerSec(int value)
+    public static int getStreamThroughputOutboundMebibytesPerSecAsInt()
+    {
+        return conf.stream_throughput_outbound.toMebibytesPerSecondAsInt();
+    }
+
+    public static void setStreamThroughputOutboundMegabitsPerSec(long value)
```
ekaterinadimitrova2: I think value should stay `int`

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2054,7 +2059,12 @@ public static double getInterDCStreamThroughputOutboundMebibytesPerSec()
         return conf.inter_dc_stream_throughput_outbound.toMebibytesPerSecond();
     }
 
-    public static void setInterDCStreamThroughputOutboundMegabitsPerSec(int value)
+    public static int getInterDCStreamThroughputOutboundMebibytesPerSecAsInt()
+    {
+        return conf.inter_dc_stream_throughput_outbound.toMebibytesPerSecondAsInt();
+    }
+
+    public static void setInterDCStreamThroughputOutboundMegabitsPerSec(long value)
```
ekaterinadimitrova2: I think `value` should stay `int`

src/java/org/apache/cassandra/tools/BulkLoader.java
```diff
@@ -67,9 +73,11 @@ public static void load(LoaderOptions options) throws BulkLoadException
                         handler,
                         options.connectionsPerHost,
                         options.targetKeyspace);
-        DatabaseDescriptor.setStreamThroughputOutboundMegabitsPerSec(options.throttle);
-        DatabaseDescriptor.setInterDCStreamThroughputOutboundMegabitsPerSec(options.interDcThrottle);
-        StreamResultFuture future = null;
+        DatabaseDescriptor.setStreamThroughputOutboundMegabitsPerSec(Double.valueOf(options.throttleMegabits).longValue());
+        DatabaseDescriptor.setInterDCStreamThroughputOutboundMegabitsPerSec(Double.valueOf(options.interDcThrottleMegabits).longValue());
```
ekaterinadimitrova2: Why don't we just leave the option in double? it is later casted to double anyway in `megabitsPerSecondInMebibytesPerSecond`?

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -224,27 +238,27 @@ public Builder authProvider(AuthProvider authProvider)
             return this;
         }
 
-        public Builder throttle(int throttle)
+        public Builder throttleMegabits(int throttleMegabits)
```
ekaterinadimitrova2: I am wondering whether we should also add builder methods for the mebibytes version we added? 
Also, considering all latest discussions around tools, are we allowed change the names of these Builder methods? Maybe better to deprecate the old name and point to the new ones?
Something like: 
```
public Builder throttleMegabits(int throttleMegabits)
        {
            this.throttleMegabits = throttleMegabits;
            return this;
        }

        
        @Deprecated 
        public Builder throttle(int throttle)
        {
            return throttleMegabits(throttle);
        }
```

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -494,24 +508,44 @@ public Builder parseArgs(String cmdArgs[])
...
+                }
+
+                if (cmd.hasOption(THROTTLE_MEGABITS))
+                {
+                    throttleMegabits = Integer.parseInt(cmd.getOptionValue(THROTTLE_MEGABITS));
+                }
+
+                if (cmd.hasOption(THROTTLE_MEBIBYTES))
+                {
+                    throttleMegabits = DataRateSpec.DataRateUnit.MEBIBYTES_PER_SECOND.toMegabitsPerSecond(Integer.parseInt(cmd.getOptionValue(THROTTLE_MEBIBYTES)));
```
ekaterinadimitrova2: Maybe new can add static import:
 `import static org.apache.cassandra.config.DataRateSpec.DataRateUnit.MEBIBYTES_PER_SECOND;`

ekaterinadimitrova2: throttleMegabits ---> Internally we save in mebibytes in DataRateSpec and megabits is used just on the parsing level, backward compatibility. So I feel we need to do this the other way around. - store in throttleMebibytes and if someone wants to use the megabits, that is converted to mebibytes. Otherwise we will do more calculations - the user gives mebibytes, we make it megabits here, but then we use those megabits to set it internally in mebibytes again in DataRateSpec

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -494,24 +508,44 @@ public Builder parseArgs(String cmdArgs[])
...
                 {
-                    throttle = Integer.parseInt(cmd.getOptionValue(THROTTLE_MBITS));
+                    interDcThrottleMegabits = Integer.parseInt(cmd.getOptionValue(INTER_DC_THROTTLE_MEGABITS));
                 }
 
-                if (cmd.hasOption(INTER_DC_THROTTLE_MBITS))
+                if (cmd.hasOption(INTER_DC_THROTTLE_MEBIBYTES))
                 {
-                    interDcThrottle = Integer.parseInt(cmd.getOptionValue(INTER_DC_THROTTLE_MBITS));
+                    interDcThrottleMegabits = DataRateSpec.DataRateUnit.MEBIBYTES_PER_SECOND.toMegabitsPerSecond(Integer.parseInt(cmd.getOptionValue(INTER_DC_THROTTLE_MEBIBYTES)));
```
ekaterinadimitrova2: same comment as for throttleMegabits

test/unit/org/apache/cassandra/streaming/StreamManagerTest.java
```diff
@@ -61,11 +61,11 @@ public static void setupClass()
     public void testUpdateStreamThroughput()
     {
         // Initialized value check
-        assertEquals(defaultStreamThroughputMebibytesPerSec * BYTES_PER_MEBIBYTE, StreamRateLimiter.getRateLimiterRateInBytes(), 0);
+        assertEquals(defaultStreamThroughputMebibytesPerSec * BYTES_PER_MEBIBYTE, StreamRateLimiter.getRateLimiterRateInBytes(), 0.01);
```
ekaterinadimitrova2: nit: I believe these changes can be skipped and we can be consistent if we change in the test class the constant MEBIBYTES_PER_MEGABIT to (same as DataRateSpec)

private static final double MEBIBYTES_IN_BITS = Math.pow(2, 23);
    private static final double MEGABIT_IN_BITS = Math.pow(10, 6);
    private static final double MEBIBYTES_PER_MEGABIT = MEGABIT_IN_BITS / MEBIBYTES_IN_BITS;

src/java/org/apache/cassandra/config/DataRateSpec.java
ekaterinadimitrova2: I am wondering what is the benefit from introducing these calculations? We can also add just a comment, at the same time the name already states - MEBIBYTES/MEGABIT.

test/unit/org/apache/cassandra/tools/LoaderOptionsTest.java
```diff
@@ -77,45 +78,165 @@ public void testNativePort() throws Exception {
...
         LoaderOptions options = LoaderOptions.builder().parseArgs(args).build();
         assertEquals("test.jks", options.clientEncOptions.keystore);
     }
 
+    @Test
+    public void testThrottleDefaultSettings()
+    {
+        LoaderOptions options = LoaderOptions.builder().build();
+        assertEquals(0, options.throttleMegabits, 0.01);
+        assertEquals(0, options.interDcThrottleMegabits, 0.01);
```
ekaterinadimitrova2: I would say 0 should be exactly 0, we should never deviate as it has special meaning - unthrottle. 
I think it is safe to leave the delta being 0 and be sure this never regress

test/unit/org/apache/cassandra/tools/BulkLoaderTest.java
```diff
@@ -151,4 +152,50 @@ public void testBulkLoader_WithArgs4() throws Throwable
...
     }
+
+    @Test(expected = NoHostAvailableException.class)
+    public void testBulkLoader_WithArgs5() throws Throwable
+    {
+        ToolResult tool = ToolRunner.invokeClass(BulkLoader.class,
+                                                 "-d",
+                                                 "127.9.9.1:9041",
+                                                 "--throttle",
+                                                 "10",
```
ekaterinadimitrova2, line 163: I would suggest to set something crazy like 3 which is less than 0.5 in mebibytes to ensure no one ever makes a change this to be rounded to 0 and unthrottle. Highly unlikely, but someone might be testing something, I like correctness :-)

src/java/org/apache/cassandra/config/DataRateSpec.java
yifan-c: It is only to make the reasoning more clear since the procedure is coded. Adding a comment with the calculations also works. Meanwhile a few constants should not introduce any overkill. I think either way is OK.

ekaterinadimitrova2: Yeah, while you are right I prefer we leave it in the comments and don't do any more changes in this area

src/java/org/apache/cassandra/tools/BulkLoader.java
frankgh: `DatabaseDescriptor.setStreamThroughputOutboundMegabitsPerSec` and `DatabaseDescriptor.setInterDCStreamThroughputOutboundMegabitsPerSec` take int as a param. So we need to cast it. In your previous comment you mentioned that in `public static void setStreamThroughputOutboundMegabitsPerSec(int value)` `value` should stay as `int`

ekaterinadimitrova2: You are right, that one is called directly in nodetool and we should keep it int as otherwise we will introduce a bug by extending to long or double.

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -60,10 +72,14 @@
     public static final String IGNORE_NODES_OPTION = "ignore";
     public static final String CONNECTIONS_PER_HOST = "connections-per-host";
     public static final String CONFIG_PATH = "conf-path";
-    public static final String THROTTLE_MBITS = "throttle";
-    public static final String INTER_DC_THROTTLE_MBITS = "inter-dc-throttle";
-    public static final String ENTIRE_SSTABLE_THROTTLE_MBITS = "entire-sstable-throttle";
-    public static final String ENTIRE_SSTABLE_INTER_DC_THROTTLE_MBITS = "entire-sstable-inter-dc-throttle";
+    @Deprecated
+    public static final String THROTTLE_MEGABITS = "throttle";
```
ekaterinadimitrova2: My OCD around the tools is on alert these days and I was wondering about changing this public static final String names but I guess people should be using the actual string, not the constants? @yifan-c ? Should we deprecate them with the old names?

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -85,10 +101,10 @@
     public final String user;
     public final String passwd;
     public final AuthProvider authProvider;
-    public final int throttle;
-    public final int interDcThrottle;
-    public final int entireSSTableThrottle;
-    public final int entireSSTableInterDcThrottle;
+    public final double throttleMegabits;
+    public final double interDcThrottleMegabits;
```
ekaterinadimitrova2: I had for a moment a concern whether we change how and where we demand only ints from the users for those two but testing in LoaderOptionsTest with double numbers seems it triggers the same NumberFormatException for all 4 of them so it seems we are good

test/unit/org/apache/cassandra/tools/LoaderOptionsTest.java
```diff
@@ -77,45 +78,165 @@ public void testNativePort() throws Exception {
...
+    @Test
+    public void testDeprecatedThrottleSettings() throws IOException
+    {
+        // Default Cassandra config
+        File config = new File(Paths.get(".", "test", "conf", "cassandra.yaml").normalize());
+        String[] args = { "-t", "200", "-idct", "400", "-d", "127.9.9.1", "-f", config.absolutePath(), sstableDirName("legacy_sstables", "legacy_ma_simple") };
+        LoaderOptions options = LoaderOptions.builder().parseArgs(args).build();
+        // converts from megabits to mebibytes
+        assertEquals(200, options.throttleMegabits, 0.01);
+        assertEquals(400, options.interDcThrottleMegabits, 0.01);
```
ekaterinadimitrova2:
```suggestion
        assertEquals(200, options.throttleMegabits, 0);
        assertEquals(400, options.interDcThrottleMegabits, 0);
```

test/unit/org/apache/cassandra/tools/LoaderOptionsTest.java
```diff
@@ -77,45 +78,165 @@ public void testNativePort() throws Exception {
...
+    @Test
+    public void testDeprecatedThrottleSettingsWithLongSettingNames() throws IOException
+    {
+        // Default Cassandra config
+        File config = new File(Paths.get(".", "test", "conf", "cassandra.yaml").normalize());
+        String[] args = { "--throttle", "200", "--inter-dc-throttle", "400", "-d", "127.9.9.1", "-f", config.absolutePath(), sstableDirName("legacy_sstables", "legacy_ma_simple") };
+        LoaderOptions options = LoaderOptions.builder().parseArgs(args).build();
+        // converts from megabits to mebibytes
+        assertEquals(200, options.throttleMegabits, 0.01);
+        assertEquals(400, options.interDcThrottleMegabits, 0.01);
```
ekaterinadimitrova2:
```suggestion
        assertEquals(200, options.throttleMegabits, 0);
        assertEquals(400, options.interDcThrottleMegabits, 0);
```

src/java/org/apache/cassandra/tools/BulkLoader.java
```diff
@@ -67,9 +73,11 @@ public static void load(LoaderOptions options) throws BulkLoadException
                         handler,
                         options.connectionsPerHost,
                         options.targetKeyspace);
-        DatabaseDescriptor.setStreamThroughputOutboundMegabitsPerSec(options.throttle);
-        DatabaseDescriptor.setInterDCStreamThroughputOutboundMegabitsPerSec(options.interDcThrottle);
-        StreamResultFuture future = null;
+        DatabaseDescriptor.setStreamThroughputOutboundMegabitsPerSec(Double.valueOf(options.throttleMegabits).intValue());
+        DatabaseDescriptor.setInterDCStreamThroughputOutboundMegabitsPerSec(Double.valueOf(options.interDcThrottleMegabits).intValue());
```
ekaterinadimitrova2: I guess it is not a big deal but I wanted to mention again that if we deprecate the megabits options and we go for mebibytes internally in DataRateSpec, probably better to add a setter in mebibytes and use mebibytes and do only one calculation megabits to mebibytes in the parking area for legacy reason in case someone wants to use the old ones.

ekaterinadimitrova2 (commented):
Overall I think I am +1, I left just a nit and we need to update also NEWS.txt for the new flags and methods and mark as deprecated the old ones. I have only one comment that I will leave to @yifan-c to express final preference in the BulkLoader class, line 76-77

src/java/org/apache/cassandra/tools/LoaderOptions.java
yifan-c: Good call! 
Cassandra is not provided as a library, but people do use it for the purpose. Compatibility is best effort. Those constants can be perfectly as "private", but declared as "public". See https://issues.apache.org/jira/browse/CASSANDRA-10637
I think the changes of the constants names are not strictly required. We can deprecate the original names and add a comment saying "MBITS" are "MEGAMITS". 
WDYT? @frankgh

frankgh: I think it's best to keep the original name, since Cassandra could sometimes be used as a library. I will add the comment and I will retain the `@Deprecated` annotation as well.

ekaterinadimitrova2: Thanks for sharing that ticket @yifan-c , I think we can even add it to the comments for other people like us, tempted to make changes here :-)

src/java/org/apache/cassandra/tools/BulkLoader.java
frankgh: I've added two setters `setStreamThroughputOutboundMebibytesPerSec`, `setInterDCStreamThroughputOutboundMebibytesPerSec`. However, I've decided to keep `throttle` and `interDcThrottle` in megabits, because the conversion from mebibytes to megabits produces a larger number, and it is less prone to loss of precision.

For example, if we have 0.1 and 0.3 mebibytes, the rounded conversion to megabits will be 1, and 3 megabits respectively. Because we cast these numbers to `int` in the setter, we get more precision in megabits. If we go the other way around, we lose precision, 3 and 1 both end up as 0 mebibytes.

frankgh:
Thanks for the review @ekaterinadimitrova2 and @yifan-c . I think this PR is in a good place for merging.

src/java/org/apache/cassandra/tools/BulkLoader.java
ekaterinadimitrova2: Considering you kept `throttleMegabits` and `interDcThrottleMegabits` being double, I guess this comment is outdated. 
I spent some time thinking and I think we should just make internally those 4 properties to be operated in bytes and change also `throttleMegabits` and `interDcThrottleMegabits` to being bytes. This will save a lot of calculations and headache. 
If we step back and look at the `StreaManager.StreamRateLimiter`, we will see that the rate limiter expects bytes.
Why this was actually not done when we were switching to the new types - because the whole effort was huge and switching internally parameters to smallest possible unit was left for follow up improvement tickets. But in this case I think this is at this point more like a bug considering how many twists and calculations the bulk loader will trigger. Also, division is more expensive operation so it is good if we can get rid of it. WDYT? CC @yifan-c

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -669,10 +755,12 @@ private static CmdLineOptions getCmdLineOptions()
...
         options.addOption("sp",  STORAGE_PORT_OPTION, "storage port", "port used for internode communication (default 7000)");
         options.addOption("ssp",  SSL_STORAGE_PORT_OPTION, "ssl storage port", "port used for TLS internode communication (default 7001), this option is deprecated, all communication goes through storage port which handles encrypted communication as well");
-        options.addOption("t", THROTTLE_MBITS, "throttle", "throttle speed in Mbits (default unlimited)");
-        options.addOption("idct", INTER_DC_THROTTLE_MBITS, "inter-dc-throttle", "inter-datacenter throttle speed in Mbits (default unlimited)");
-        options.addOption("e", ENTIRE_SSTABLE_THROTTLE_MBITS, "entire-sstable-throttle", "entire SSTable throttle speed in Mbits (default unlimited)");
-        options.addOption("eidct", ENTIRE_SSTABLE_INTER_DC_THROTTLE_MBITS, "entire-sstable-inter-dc-throttle", "entire SSTable inter-datacenter throttle speed in Mbits (default unlimited)");
+        options.addOption("t", THROTTLE_MBITS, "throttle", "throttle speed in Mbps (default 0 for unlimited), this option is deprecated, use \"throttle-mib\" instead");
```
ekaterinadimitrova2, line 753: nit: use `tmib`, while `throttle-mib` explains it all, the actual option is different

ekaterinadimitrova2: Same for the `inter-dc`

ekaterinadimitrova2 (approved):
I left only one comment about the description, otherwise I am +1. we need only NEWS.txt (deprecation section and to explain we have changed the new flags for entire-stable an this is not compatible with alpha1 version)

src/java/org/apache/cassandra/tools/LoaderOptions.java
frankgh: We discussed offline and decided to keep the _long name_, and additionally, we decided to remove the short names that we're going to be introduced in this PR. I have updated the code, documentation, and tests to reflect the removal of the short names

doc/modules/cassandra/pages/operating/bulk_loading.adoc
```diff
@@ -80,13 +80,23 @@ The following options are supported, with `-d,--nodes <initial hosts>` required:
...
+
 -f,--conf-path <path to config file>                         cassandra.yaml file path for streaming throughput and client/server SSL.
 
 -h,--help                                                    Display this help message
 
 -i,--ignore <NODES>                                          Don't stream to this (comma separated) list of nodes
 
--idct,--inter-dc-throttle <inter-dc-throttle>                Inter-datacenter throttle speed in Mbits (default unlimited)
+-idct,--inter-dc-throttle <inter-dc-throttle>                (deprecated) Inter-datacenter throttle speed in Mbits (default 0 for unlimited)
+                                                             Use -idctmib / --inter-dc-throttle-mib instead
```
yifan-c: remove `-idctmib`

doc/modules/cassandra/pages/operating/bulk_loading.adoc
```diff
@@ -111,8 +121,10 @@ The following options are supported, with `-d,--nodes <initial hosts>` required:
                                                              for TLS internode communication (default 7001)
 -st,--store-type <STORE-TYPE>                                Client SSL:
                                                              type of store
--t,--throttle <throttle>                                     Throttle
-                                                             speed in Mbits (default unlimited)
+-t,--throttle <throttle>                                     (deprecated) Throttle speed in Mbits (default 0 for unlimited)
+                                                             Use -tmib / --throttle-mib instead
```
yifan-c: remove `-tmib`

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -22,25 +22,37 @@
 
 import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
-import java.net.*;
+import java.net.InetAddress;
```
yifan-c, line 25: I am not against the changes. Those are good and more in line with the code style, but unrelated changes of the patch.

src/java/org/apache/cassandra/tools/LoaderOptions.java
yifan-c: I am kind of worried about the similarity of the names... It could be prone to oversights. 
For example, code auto-completion may put `throttleMegabits`, while the actual intent is to have `throttleMebibytes`. 
How about not introducing a new method with the similar name, but adding java docs to indicate the input `throttle` (better rename it to `throttleMegabits`) is in the megabits, and the method is deprecated. 
@ekaterinadimitrova2 , what do you think?

ekaterinadimitrova2: Oh so you suggest we remove the `throttleMegabits` and end up only with the old deprecated `throttle` and the new `throttleMebibytes`
That also should work. Honestly, people again can make the same mistake and ignore `throttle `was in megabits but I guess less methods is better. Also, if we decide to get rid of the megabits version for those properties on a major version.

`better rename it to throttleMegabits` --> I think any renaming here will be breaking change?

yifan-c: > better rename it to throttleMegabits

I was not clear on what is "it". I mean to say that it is better to just rename the _input parameter_ from `throttle` to `throttleMegabits`. Instead, we have 

```java
        @Deprecated 
        public Builder throttle(int throttleMegabits)
        {
           ...
        }
```

src/java/org/apache/cassandra/tools/LoaderOptions.java
frankgh: should I revert these changes?

PR #1704, opened 2022-06-27
CASSANDRA-15241 Add a virtual table that exposes currently running queries

maedhroz:


maedhroz:
https://app.circleci.com/pipelines/github/maedhroz/cassandra?branch=CASSANDRA-15241-trunk-v2&filter=all

src/java/org/apache/cassandra/concurrent/SEPWorker.java
```diff
@@ -102,8 +122,9 @@ public void run()
                     continue;
                 if (SET_THREAD_NAME)
                     Thread.currentThread().setName(assigned.name + "-" + workerId);
-                task = assigned.tasks.poll();
 
+                task = assigned.tasks.poll();
```
maedhroz, line 126: @clohfink Without setting the task here, it's possible to miss tasks, right?

src/java/org/apache/cassandra/concurrent/TaskFactory.java
```diff
@@ -128,7 +128,41 @@ protected LocalAware() {}
...
+                        return debuggable.startTimeNanos();
+                    }
+
+                    @Override
+                    public String description()
+                    {
+                        return debuggable.description();
+                    }
+                };
+            }
```
maedhroz: Before the simulator changes, this bit wouldn't have been necessary, as everything got wrapped up in `FutureTask`. Is there a better way?

CC @belliottsmith

src/java/org/apache/cassandra/db/virtual/QueriesTable.java
```diff
@@ -0,0 +1,92 @@
...
+        long now = approxTime.now();
+        
+        for (DebuggableTask.RunningDebuggableTask task : SharedExecutorPool.SHARED.runningTasks())
+        {
+            if (!task.hasTask()) continue;
+            
+            long creationTimeNanos = task.creationTimeNanos();
+            long startTimeNanos = task.startTimeNanos();
+            long queuedMicros = NANOSECONDS.toMicros(max((startTimeNanos > 0 ? startTimeNanos : now) - creationTimeNanos, 0));
+            long runningMicros = startTimeNanos > 0 ? NANOSECONDS.toMicros(now - startTimeNanos) : 0;
```
maedhroz, line 84: In the original patch, this looked like..

```
long runningMicros = approxTimeOfStart > 0 ? NANOSECONDS.toMicros(max(approxTimeOfStart - now, 0)) : 0;
```

It seemed like the `approxTimeOfStart - now` was reversed, and the max wasn't necessary if `now` should never be less than the start time. Although thinking more about it, we should probably just re-acquire `now` for every task...

CC @clohfink

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -2479,11 +2518,18 @@ public DroppableRunnable(Verb verb)
...
         public final void run()
         {
-            long approxCurrentTimeNanos = MonotonicClock.Global.approxTime.now();
+            approxStartTimeNanos = MonotonicClock.Global.approxTime.now();
+
+            try {
+                Thread.sleep(10);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
```
maedhroz: TODO: will be removed...was just there for testing

src/java/org/apache/cassandra/transport/Dispatcher.java
```diff
@@ -79,17 +80,67 @@ public Dispatcher(boolean useLegacyFlusher)
 
     public void dispatch(Channel channel, Message.Request request, FlushItemConverter forFlusher, Overload backpressure)
     {
-        requestExecutor.submit(() -> processRequest(channel, request, forFlusher, backpressure));
+        requestExecutor.submit(new RequestProcessor(channel, request, forFlusher, backpressure));
```
maedhroz, line 83: The fun thing for testing is that in-JVM dtests don't dispatch normally against the NTR stage pool. It might end up being bette to use `CQLTester` for this...

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -2479,11 +2518,18 @@ public DroppableRunnable(Verb verb)
 
         public final void run()
         {
-            long approxCurrentTimeNanos = MonotonicClock.Global.approxTime.now();
+            approxStartTimeNanos = MonotonicClock.Global.approxTime.now();
```
maedhroz, line 2532: Note: Do we need to make this `volatile` to ensure visibility for the start time for the thread querying the virtual table?

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -2518,11 +2565,12 @@ private static abstract class LocalMutationRunnable implements Runnable
         public final void run()
         {
             final Verb verb = verb();
-            long nowNanos = MonotonicClock.Global.approxTime.now();
+            approxStartTimeNanos = MonotonicClock.Global.approxTime.now();
```
maedhroz, line 2572: Note: Do we need to make this `volatile` to ensure visibility for the start time for the thread querying the virtual table?

src/java/org/apache/cassandra/transport/Dispatcher.java
```diff
@@ -79,17 +80,67 @@ public Dispatcher(boolean useLegacyFlusher)
...
+            this.channel = channel;
+            this.request = request;
+            this.forFlusher = forFlusher;
+            this.backpressure = backpressure;
+        }
+
+        @Override
+        public void run()
+        {
+            approxStartTimeNanos = MonotonicClock.Global.approxTime.now();
```
maedhroz, line 108: Note: Do we need to make this `volatile` to ensure visibility for the start time for the thread querying the virtual table?

src/java/org/apache/cassandra/concurrent/FutureTask.java
```diff
@@ -63,6 +77,7 @@ public void run()
         finally
         {
             call = null;
+            debuggable = null; // TODO: If we didn't do this, "debuggable" could be final. Do we actually need to?
```
maedhroz: @belliottsmith I'm not entirely sure why we null out `call` here. I could avoid reasoning about threads sharing `debuggable` if it was `final`...

src/java/org/apache/cassandra/concurrent/TaskFactory.java
maedhroz: It's similar to what we still do w/ `FutureTask` for NTR stage requests (and did for everything in the old patch).

src/java/org/apache/cassandra/service/StorageProxy.java
belliottsmith: We can if we want, but this will be visible either way, it's just a question of promptness (not remotely important from our perspective, but also not a big deal to provide)

src/java/org/apache/cassandra/concurrent/FutureTask.java
belliottsmith: To garbage collect - a `Future` can survive a lot longer than the work it was performing.

src/java/org/apache/cassandra/db/virtual/QueriesTable.java
maedhroz: It looks like everything that could generate these numbers is pulling from `MonotonicClock.Global.approxTime`. If we get something negative, I think it would be a sign of a bug somewhere. Resolving for now.

src/java/org/apache/cassandra/db/virtual/QueriesTable.java
```diff
@@ -0,0 +1,94 @@
...
+ * cqlsh> SELECT * FROM system_views.queries;
+ *
+ *  thread_id                   | queued_micros |  running_micros | task
+ * ------------------------------+---------------+-----------------+--------------------------------------------------------------------------------
+ *  Native-Transport-Requests-7 |         72923 |            7611 |                      QUERY select * from system_views.queries; [pageSize = 100]
+ *              MutationStage-2 |         18249 |            2084 | Mutation(keyspace='distributed_test_keyspace', key='000000f8', modifications...
+ *                  ReadStage-2 |         72447 |           10121 |                                         SELECT * FROM keyspace.table LIMIT 5000
+ * </pre>
+ */    
+public class QueriesTable extends AbstractVirtualTable
```
frankgh: for consistency with other vtables
```suggestion
final class QueriesTable extends AbstractVirtualTable
```

src/java/org/apache/cassandra/db/virtual/QueriesTable.java
```diff
@@ -0,0 +1,94 @@
...
+{
+    private static final String TABLE_NAME = "queries";
+    private static final String ID = "thread_id";
+    private static final String QUEUED = "queued_micros";
+    private static final String RUNNING = "running_micros";
+    private static final String DESC = "task";
+
+    QueriesTable(String keyspace)
+    {
+        super(TableMetadata.builder(keyspace, TABLE_NAME)
```
frankgh, line 55: Should we add a comment? It's not super consistent across all vtables, but it seems like good practice
```suggestion
        super(TableMetadata.builder(keyspace, TABLE_NAME)
                           .comment("......")
```

src/java/org/apache/cassandra/db/virtual/QueriesTable.java
```diff
@@ -0,0 +1,94 @@
...
+            
+            long creationTimeNanos = task.creationTimeNanos();
+            long startTimeNanos = task.startTimeNanos();
+            long now = approxTime.now();
+
+            long queuedMicros = NANOSECONDS.toMicros(max((startTimeNanos > 0 ? startTimeNanos : now) - creationTimeNanos, 0));
+            long runningMicros = startTimeNanos > 0 ? NANOSECONDS.toMicros(now - startTimeNanos) : 0;
+            
+            result.row(task.threadId())
+                  .column(QUEUED, queuedMicros)
```
frankgh, line 87: I'm curious about the micros resolution. Would nanos suffice? Maybe there's a good motivation for micros

maedhroz: This I copied from @clohfink 's original patch. I think most of the other metrics we expose around latencies/running times are in micros.

maedhroz:
Committed as https://github.com/apache/cassandra/commit/89f3978dcde958fbad191b8cf628fd89ace64d7a

PR #1705, opened 2022-06-28
CASSANDRA-17695: Fix AbstractCell#toString for tombstoned multicell CollectionTypes

frankgh:
Calling `AbstractCell#toString` on a tombstoned multi-cell `CollectionType` for the `date`, `duration`,
`smallint`, `time`, and `tinyint` types produces a `MarshalException`. The root cause of the issue
occurs when `AbstractCell#toString` in turns calls `AbstractType#getString(V value, ValueAccessor<V> acccesor)`.
The `TypeSerializer` performs validation of the value. However, because this value is tombstoned
the size of the value is 0 which doesn't match the type value. The failed check throws a `MarshalException`.
In this commit, we check whether the Cell is tombstoned and add the "<tombstone>" string to the
output of the `AbstractCell#toString` method for tombstoned cells.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,2168 +14,126 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
 
-version: 2
 jobs:
-  j8_jvm_upgrade_dtests:
-    docker:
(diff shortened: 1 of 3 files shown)


maedhroz (approved):

frankgh:
Merged via https://github.com/apache/cassandra/commit/a9725b681b948f2122f3d48b96a5c4e7403d2c39

PR #1706, opened 2022-06-28
CASSANDRA-17708 Fixing flaky tests in InternodeEncryptionEnforcementTest

jyothsnakonisa:
1. Fixed the flakiness in InternodeEncryptionEnforcementTests
2. Fixed a race condition in `AllowFirstAndRejectOtherOutboundAuthenticator` which was causing timeouts
3. Added a log message after outbound connections are interrupted in OutboundConnectionInitiator,  `testOutboundConnectionsAreInterruptedWhenAuthFails` was sometimes timing out waiting for the log message.
4. Configured authenticators used for testing a special case only on instance 1 of the cluster. This way we can deterministically check for logs on instance 1. Earlier sometimes waiting for logs on instance 2 was timing out as the logs are not present on instance 2.
5. Ran repeated tests on the class `InternodeEncryptionEnforcementTest `, CircleCI link https://app.circleci.com/pipelines/github/jyothsnakonisa/cassandra/132/workflows/182dd59d-cffe-4f6e-959e-0c32395b9088

### DIFF
--- a/src/java/org/apache/cassandra/net/OutboundConnectionInitiator.java
+++ b/src/java/org/apache/cassandra/net/OutboundConnectionInitiator.java
@@ -147,7 +147,8 @@ public class OutboundConnectionInitiator<SuccessType extends OutboundConnectionI
         {
             // interrupt other connections, so they must attempt to re-authenticate
             MessagingService.instance().interruptOutbound(settings.to);
-            return ImmediateFuture.failure(new IOException("authentication failed to " + settings.connectToId()));
+            logger.error("Authentication failed to " + settings.connectToId());
+            return ImmediateFuture.failure(new IOException("Authentication failed to " + settings.connectToId()));
         }
 
 
@@ -281,7 +282,7 @@ public class OutboundConnectionInitiator<SuccessType extends OutboundConnectionI
             {
                 // interrupt other connections, so they must attempt to re-authenticate
                 MessagingService.instance().interruptOutbound(settings.to);
-                logger.error("authentication failed to " + settings.connectToId());
+                logger.error("Authentication failed to " + settings.connectToId());
 
                 // To release all the pending buffered data, replace authentication handler with discard handler.
                 // This avoids pending inbound data to be fired through the pipeline
--- a/test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
+++ b/test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
@@ -17,6 +17,7 @@
  */
 package org.apache.cassandra.distributed.test;
 
+import java.io.Closeable;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -43,6 +44,7 @@ import org.apache.cassandra.exceptions.ConfigurationException;
 import org.apache.cassandra.net.InboundMessageHandlers;
 import org.apache.cassandra.net.MessagingService;
 import org.apache.cassandra.net.OutboundConnections;
+import org.awaitility.Awaitility;
 
 import static com.google.common.collect.Iterables.getOnlyElement;
 import static org.hamcrest.Matchers.containsString;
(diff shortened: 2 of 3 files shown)


src/java/org/apache/cassandra/net/OutboundConnectionInitiator.java
```diff
@@ -147,6 +147,7 @@ private Future<Result<SuccessType>> initiate(EventLoop eventLoop)
         {
             // interrupt other connections, so they must attempt to re-authenticate
             MessagingService.instance().interruptOutbound(settings.to);
+            logger.error("authentication failed to " + settings.connectToId());
```
jonmeredith: We normally capitalize log messages - `Authentication failed`

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -164,7 +149,9 @@ public void testOutboundConnectionsAreInterruptedWhenAuthFails() throws IOExcept
                 assertTrue(!outbound.small.isConnected() && !outbound.large.isConnected() && !outbound.urgent.isConnected());
             };
             cluster.get(1).runOnInstance(runnable);
+            cluster.get(1).shutdown();
```
jonmeredith: Why shutdown before the try block completes and closes the cluster?

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -432,7 +426,8 @@ public static class AllowFirstAndRejectOtherOutboundAuthenticator extends Reject
         @Override
         public boolean authenticate(InetAddress remoteAddress, int remotePort, Certificate[] certificates, InternodeConnectionDirection connectionType)
         {
-            if (connectionType == InternodeConnectionDirection.OUTBOUND)
+            if (connectionType == InternodeConnectionDirection.OUTBOUND ||
+                connectionType == InternodeConnectionDirection.OUTBOUND_PRECONNECT)
```
jonmeredith: why add `OUTBOUND_PRECONNECT`? Doesn't that mean you don't exercise the outbound auth failure path as it always preconnects before the real auth attempt.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -161,10 +147,11 @@ public void testOutboundConnectionsAreInterruptedWhenAuthFails() throws IOExcept
 
                 // There should be no outbound connections as authentication fails.
                 OutboundConnections outbound = getOnlyElement(MessagingService.instance().channelManagers.values());
-                assertTrue(!outbound.small.isConnected() && !outbound.large.isConnected() && !outbound.urgent.isConnected());
+                Awaitility.await().until(() -> !outbound.small.isConnected() && !outbound.large.isConnected() && !outbound.urgent.isConnected());
```
jonmeredith, line 149: Looks like the default wait is 10s - should probably be enough time.

jonmeredith (approved):

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
jyothsnakonisa: Out of 500 runs none failed, so It should be okay.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -86,8 +84,7 @@ public void testInboundConnectionsAreRejectedWhenAuthFails() throws IOException,
             // Wait for authentication to fail
             cluster.get(1).logs().watchFor("Unable to authenticate peer");
             cluster.get(1).runOnInstance(runnable);
-            cluster.get(2).logs().watchFor("Unable to authenticate peer");
-            cluster.get(2).runOnInstance(runnable);
```
yifan-c, line 90: nit: how about adding a watch for "Connection reset by peer" in instances 2's log?

```
            cluster.get(2).logs().watchFor("Connection reset by peer");
```

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -103,16 +100,12 @@ public void testOutboundConnectionsAreRejectedWhenAuthFails() throws IOException
             executorService.submit(() -> openConnections(cluster));
 
             /*
-             * instance (1) should not connect to instance (2) as authentication fails;
-             * instance (2) should not connect to instance (1) as authentication fails.
+             * Instance (1) should not be able to make outbound connections to instance (2) but Instance (2) should be
+             * accepting outbound connections from Instance (1)
```
yifan-c, line 104: The comment is confusing to me.. 
If instance 1 can never make outbound connections, how can instance 2 expect connections _from_ instance 1? Do you mean instance 2 can make outbound connections _to_ instance 1?

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
```diff
@@ -134,19 +125,14 @@ public void testOutboundConnectionsAreRejectedWhenAuthFails() throws IOException
     public void testOutboundConnectionsAreInterruptedWhenAuthFails() throws IOException, TimeoutException
     {
         Cluster.Builder builder = createCluster(AllowFirstAndRejectOtherOutboundAuthenticator.class);
+        final ExecutorService executorService = Executors.newSingleThreadExecutor();
+
         try (Cluster cluster = builder.start())
```
yifan-c: nit: it is safer to close the `ExecutorService` along with `cluster`. Otherwise, if the body of `try` throws, only `cluster` is closed, but `executorService.shutdown()` will not be called. 

You can have multiple `closable`s in the try-with-resources statement, and remove `executorService.shutdown()` at the end of the test.
```
try (Cluster cluster = builder.start(); Closeable es = executorService::shutdown)
```

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
jyothsnakonisa: I think we should watch for "Unable to authenticate peer" which is more specific in terms of why the connection reset happen. Also it is the log that we print after `authenticate` call fails.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
jyothsnakonisa: `RejectOutboundAuthenticator` is configured only on instance 1 of the cluster and `AllowAllAuthenticator` is configured on instance 2. Any outbound connections or inbound connections on instance 2 should be successful as `AllowAllAuthenticator` is used, where as on instance 1, Inbound connections will be accepted but OutboundConnections will be rejected.

test/distributed/org/apache/cassandra/distributed/test/InternodeEncryptionEnforcementTest.java
yifan-c: ok.

PR #1714, opened 2022-07-01
CASSANDRA-17307 4.1: Flaky MixedModeAvailability tests

adelapena:


test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeAvailabilityTestBase.java
```diff
@@ -18,77 +18,129 @@
...
+        })
         .runAfterNodeUpgrade((cluster, n) -> {
 
+            ICoordinator coordinator = cluster.coordinator(COORDINATOR);
+
             // using 0 to 2 down nodes...
-            for (int numNodesDown = 0; numNodesDown < NUM_NODES; numNodesDown++)
+            for (int i = 0; i < NUM_NODES; i++)
             {
+                final int numNodesDown = i;
```
bereng, line 103: Why do you need this? I feel like I am missing sthg here

test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeAvailabilityTestBase.java
```diff
@@ -18,77 +18,129 @@
...
+                        coordinator.execute(INSERT, writeConsistencyLevel, row1);
+                        coordinator.execute(INSERT, writeConsistencyLevel, row2);
+                    });
+
+                    wrote = true;
+
+                    // test read
+                    maybeFail(true, numNodesDown > maxNodesDown(readConsistencyLevel), () -> {
+                        Object[][] rows = coordinator.execute(SELECT, readConsistencyLevel, key);
+                        if (numNodesDown <= maxNodesDown(writeConsistencyLevel))
```
bereng, line 130: Should this be checking against the read CL instead?

test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeAvailabilityTestBase.java
adelapena: We need a `final` copy of the non-final `i`, so we can use it in the lambda that we pass to the second call to `maybeFail`, [here](https://github.com/apache/cassandra/blob/f3e082990e00cc54b9e24212c2ab6da8f937344d/test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeAvailabilityTestBase.java#L130).

test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeAvailabilityTestBase.java
adelapena: The read CL is used as the `shouldFail` argument of the call to `maybeFail`. That should check that reads succeed or fail depending on the CL and down nodes. Once we know that the query succeeds, this check verifies the read query contents. This is only done if the writes have succeeded and left something to read, and for that we need to check the write CL.

This part comes more or less unmodified from the original test: https://github.com/apache/cassandra/blob/trunk/test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeAvailabilityTestBase.java#L148-L149

test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeAvailabilityTestBase.java
bereng: sgtm right

PR #1737, opened 2022-07-18
17735 trunk

ekaterinadimitrova2:


src/java/org/apache/cassandra/config/Converters.java
```diff
@@ -71,8 +71,8 @@
                             DurationSpec.IntSecondsBound::inSecondsString,
                             o -> Long.toString(o.toSeconds())),
     MINUTES_DURATION(Integer.class, DurationSpec.IntMinutesBound.class,
-                     DurationSpec.IntMinutesBound::new,
-                     DurationSpec.IntMinutesBound::toMinutes),
+                     o -> o == -1 ? null : new DurationSpec.IntMinutesBound(o),
```
adelapena, line 78: Maybe we should rename this converter to `MINUTES_CUSTOM_DURATION` and add a comment, since this is a special way of converting due to the acceptance of `-1` as disabled value. That's what was done with `MILLIS_CUSTOM_DURATION`, although here we wouldn't have a regular/custom pair of converters.

src/java/org/apache/cassandra/io/sstable/IndexSummaryManager.java
```diff
@@ -100,6 +100,7 @@ public void setResizeIntervalInMinutes(int resizeIntervalInMinutes)
     {
         int oldInterval = this.resizeIntervalInMinutes;
         this.resizeIntervalInMinutes = resizeIntervalInMinutes;
+        DatabaseDescriptor.setIndexSummaryResizeIntervalInMinutes(resizeIntervalInMinutes);
```
adelapena, line 101: Since we want to keep `IndexSummaryManager.resizeIntervalInMinutes` synchronized with `Config.index_summary_resize_interval`, we might simplify by removing the local copy on `IndexSummaryManager` and always use the value stored in `Config`, this way: https://github.com/adelapena/cassandra/commit/87882f0c94e8f890e01becf5caaed6cbfd356971

adelapena (commented):
Looks good, I have just left a couple of suggestions.

src/java/org/apache/cassandra/io/sstable/IndexSummaryManager.java
ekaterinadimitrova2: I thought about it, it was probably made local long time ago when there were no Virtual Tables, but I was considering that in 4.1 using directly the getter will add one more operation multiplication if someone doesn't use minutes so stayed conservative :D  might be too conservative though :D

src/java/org/apache/cassandra/config/Converters.java
ekaterinadimitrova2: That and I actually got reminded that I need to update this converter unit test to include the "-1" part

PR #1738, opened 2022-07-19
Cassandra 17759

smiklosovic:


test/distributed/org/apache/cassandra/distributed/impl/RowUtil.java
```diff
@@ -52,7 +52,8 @@ public static SimpleQueryResult toQueryResult(ResultMessage res)
         }
         else
         {
-            return QueryResults.empty();
+            List<String> warnings = res.getWarnings();
```
smiklosovic: this was not covered before, DDL statements can emit warnings too.

src/java/org/apache/cassandra/locator/TokenMetadata.java
```diff
@@ -1039,6 +1041,13 @@ public Set<InetAddress> getAllEndpoints()
         }
     }
 
+    public Set<InetAddress> getAllMembers()
```
driftx, line 1043: I think since tMD is the source of truth for the gossiper, we should do something similar to isMember (which is how the gossiper determines membership) and consult the tokenToEndpointMap.

PR #1739, opened 2022-07-21
CASSANDRA-8877 trunk: Ability to read the TTL and WRITE TIME of an element in a collection

adelapena:


NEWS.txt
```diff
@@ -63,9 +63,11 @@ New features
       If this is set to false, streaming will be considerably faster however it's possible that, in extreme situations
       (losing > quorum # nodes in a replica set), you may have data in your SSTables that never makes it to the CDC log.
       The default is true/enabled. The configuration can be altered via JMX.
+    - Added support for reading the write times and TTLs of the elements of collections and UDTs, frozen or not. The CQL
```
yifan-c: nit: "frozen or not" --> "regardless being frozen or not"

NEWS.txt
```diff
@@ -63,9 +63,11 @@ New features
...
+      functions writetime, maxwritetime and ttl can now be applied to entire collections/UDTs, single collection/UDT
+      elements and slices of collection/UDT elements.
     - Added a new CQL function, maxwritetime. It shows the largest unix timestamp that the data was written, similar to
-      its sibling CQL function, writetime. Unlike writetime, maxwritetime can be applied to multi-cell data types, e.g.
-      non-frozen collections and UDT, and returns the largest timestamp. One should not to use it when upgrading to 4.2.
+      its sibling CQL function, writetime. One should not to use it when upgrading to 4.2.
```
yifan-c: This one is my bad.
`One should not to use it` --> `One should not use it`

src/java/org/apache/cassandra/cql3/selection/ColumnTimestamps.java
```diff
@@ -0,0 +1,381 @@
...
+         * @param timestamp the timestamp to serialize
+         * @return the bytes corresponding to the specified timestamp
+         */
+        abstract ByteBuffer toByteBuffer(long timestamp);
+    }
+
+    /**
+     * A {@link ColumnTimestamps} that doesn't contain any timestamps.
+     */
+    static ColumnTimestamps NO_TIMESTAMP = new ColumnTimestamps(null)
```
yifan-c: Should it be `final`?

src/java/org/apache/cassandra/cql3/selection/RowTimestamps.java
```diff
@@ -0,0 +1,105 @@
...
+     *
+     * @param index the column index
+     * @return the timestamp of the specified column
+     */
+    ColumnTimestamps get(int index);
+
+    /**
+     * A {@code RowTimestamps} that does nothing.
+     */
+    static RowTimestamps NOOP_ROW_TIMESTAMPS = new RowTimestamps()
```
yifan-c: `static` is redundant since the variable is defined in an interface.

src/java/org/apache/cassandra/cql3/selection/WritetimeOrTTLSelector.java
```diff
@@ -32,51 +32,57 @@
...
-                return kind.returnType;
+                AbstractType<?> type = kind.returnType;
+                return isMultiCell && !kind.aggregatesMultiCell() ? ListType.getInstance(type, false) : type;
             }
 
+            @Override
             protected void addColumnMapping(SelectionColumnMapping mapping, ColumnSpecification resultsColumn)
             {
-               mapping.addMapping(resultsColumn, def);
+                factory.addColumnMapping(mapping, resultsColumn);
```
yifan-c, line 79: nit: remove the extra heading space.

src/java/org/apache/cassandra/serializers/SetSerializer.java
```diff
@@ -248,4 +253,88 @@ public ByteBuffer getSliceFromSerialized(ByteBuffer collection,
...
+            {
+                ByteBuffer value = readValue(input, ByteBufferAccessor.instance, offset, ProtocolVersion.V3);
+                offset += sizeOfValue(value, ByteBufferAccessor.instance, ProtocolVersion.V3);
+                int comparison = comparator.compareForCQL(value, key);
+                if (comparison == 0)
+                    return i;
+                else if (comparison > 0)
+                    // since the set is in sorted order, we know we've gone too far and the element doesn't exist
+                    return -1;
+                // else, we're before the element so continue
```
yifan-c: nit: add curly brackets. I found it a bit difficult to read when being surrounded by comments.

src/antlr/Parser.g
```diff
@@ -414,12 +414,12 @@ simpleUnaliasedSelector returns [Selectable.Raw s]
...
 
 selectionFunction returns [Selectable.Raw s]
-    : K_COUNT '(' '\*' ')'                      { $s = Selectable.WithFunction.Raw.newCountRowsFunction(); }
-    | K_MAXWRITETIME '(' c=sident ')'           { $s = new Selectable.WritetimeOrTTL.Raw(c, Selectable.WritetimeOrTTL.Kind.MAX_WRITE_TIME); }
-    | K_WRITETIME '(' c=sident ')'              { $s = new Selectable.WritetimeOrTTL.Raw(c, Selectable.WritetimeOrTTL.Kind.WRITE_TIME); }
-    | K_TTL       '(' c=sident ')'              { $s = new Selectable.WritetimeOrTTL.Raw(c, Selectable.WritetimeOrTTL.Kind.TTL); }
-    | K_CAST      '(' sn=unaliasedSelector K_AS t=native_type ')' {$s = new Selectable.WithCast.Raw(sn, t);}
-    | f=functionName args=selectionFunctionArgs { $s = new Selectable.WithFunction.Raw(f, args); }
+    : K_COUNT        '(' '\*' ')'                                    { $s = Selectable.WithFunction.Raw.newCountRowsFunction(); }
+    | K_MAXWRITETIME '(' c=sident m=selectorModifier[c] ')'          { $s = new Selectable.WritetimeOrTTL.Raw(c, m, Selectable.WritetimeOrTTL.Kind.MAX_WRITE_TIME); }
```
yifan-c, line 418: now `maxwritetime` can be applied to a slice?

yifan-c (commented):
Skimmed through for the first time. Left some questions and nits.

src/java/org/apache/cassandra/cql3/selection/WritetimeOrTTLSelector.java
adelapena: I think that the previous line was missing a heading space, and now it's properly aligned.

NEWS.txt
adelapena: +1, changed to `regardless of being frozen or not`.

src/java/org/apache/cassandra/cql3/selection/ColumnTimestamps.java
adelapena: Right, it should be `final`.

src/java/org/apache/cassandra/cql3/selection/RowTimestamps.java
adelapena: Removed.

src/antlr/Parser.g
adelapena: Yes, `maxwritetime` can be applied to entire collections, slices and collection items :)

It's verified for all the test cases in `WritetimeOrTTLTest`, [here](https://github.com/apache/cassandra/blob/a1c6828d186ee7ac36e8965b70930341d1ea5196/test/unit/org/apache/cassandra/cql3/validation/entities/WritetimeOrTTLTest.java#L1086-L1091) and [here](https://github.com/apache/cassandra/blob/a1c6828d186ee7ac36e8965b70930341d1ea5196/test/unit/org/apache/cassandra/cql3/validation/entities/WritetimeOrTTLTest.java#L1116-L1122).

src/java/org/apache/cassandra/cql3/selection/ElementsSelector.java
```diff
@@ -226,9 +234,19 @@ public ByteBuffer getOutput(ProtocolVersion protocolVersion)
...
+    public void addInput(InputRow input)
+    {
+        selected.addInput(input);
+    }
+
+    protected int getElementIndex(ByteBuffer output, ByteBuffer key)
     {
-        selected.addInput(protocolVersion, input);
+        return type.getSerializer().getIndexFromSerialized(output, key, keyType(type));
+    }
```
yifan-c: This method is not used. It looks almost the same as the one in `ElementsSelector.ElementSelector#getElementIndex` at line 305

src/java/org/apache/cassandra/cql3/selection/ElementsSelector.java
```diff
@@ -43,12 +44,19 @@
  */
 abstract class ElementsSelector extends Selector
 {
+    /**
+     * An empty collection is composed of an int size of zero.
+     */
+    private static final ByteBuffer EMPTY_FROZEN_COLLECTION = ByteBufferUtil.bytes(0);
+
     protected final Selector selected;
+    protected final CollectionType<?> type;
```
yifan-c, line 53: With the `type` defined in the abstract class, I think the one in the implementation (line 276) can be removed.

src/java/org/apache/cassandra/cql3/selection/ColumnTimestamps.java
```diff
@@ -0,0 +1,381 @@
...
+        {
+            if (range.isEmpty())
+                return NO_TIMESTAMP;
+
+            int from = !range.hasLowerBound() ? 0
+                                              : range.lowerBoundType() == BoundType.CLOSED ? range.lowerEndpoint()
+                                                                                           : range.lowerEndpoint() + 1;
+            int to = !range.hasUpperBound() ? timestamps.size()
+                                            : range.upperBoundType() == BoundType.CLOSED ? range.upperEndpoint() + 1
+                                                                                         : range.upperEndpoint();
```
yifan-c: nit: just a different style to avoid the nested ternary operators. 
Can you also add comments to explain calculations for `from` and `to`?

```suggestion
            int from = 0;
            int to = timestamps.size();
            if (range.hasLowerBound())
            {
                from = range.lowerBoundType() == BoundType.CLOSED
                       ? range.lowerEndpoint()
                       : range.lowerEndpoint() + 1;
            }
            if (range.hasUpperBound())
            {
                to = range.upperBoundType() == BoundType.CLOSED
                     ? range.upperEndpoint() + 1
                     : range.upperEndpoint();
            }
```

test/unit/org/apache/cassandra/cql3/validation/entities/WritetimeOrTTLTest.java
```diff
@@ -248,25 +1166,31 @@ private void assertInvalidPrimaryKeySelection(String column) throws Throwable
...
         assertInvalidThrowMessage("Cannot use selection function ttl on PRIMARY KEY part " + column,
                                   InvalidRequestException.class,
-                                  String.format("SELECT TTL(%s) FROM %%s", column));
+                                  format("SELECT TTL(%s) FROM %%s", column));
+    }
+
+    private void assertInvalidListElementSelection(String column, String list) throws Throwable
+    {
+        String message = format("Element selection is only allowed on sets and maps, but %s is a list", list);
+        assertInvalidThrowMessage(message,
```
yifan-c, line 1224: add the assertion for `MAXWRITETIME` too.

test/unit/org/apache/cassandra/cql3/validation/entities/WritetimeOrTTLTest.java
```diff
@@ -248,25 +1166,31 @@ private void assertInvalidPrimaryKeySelection(String column) throws Throwable
...
     }
 
-    private void assertInvalidMultiCellSelection(String column, boolean isCollection) throws Throwable
+    private void assertInvalidListSliceSelection(String column, String list) throws Throwable
     {
-        String message = format("Cannot use selection function %%s on non-frozen %s %s",
-                                isCollection ? "collection" : "UDT", column);
-        assertInvalidThrowMessage(format(message, "writetime"),
+        String message = format("Slice selection is only allowed on sets and maps, but %s is a list", list);
+        assertInvalidThrowMessage(message,
```
yifan-c, line 1238: also add the assertion for `MAXWRITETIME`

test/unit/org/apache/cassandra/cql3/validation/entities/WritetimeOrTTLTest.java
```diff
@@ -248,25 +1166,31 @@ private void assertInvalidPrimaryKeySelection(String column) throws Throwable
     {
         assertInvalidThrowMessage("Cannot use selection function writetime on PRIMARY KEY part " + column,
                                   InvalidRequestException.class,
-                                  String.format("SELECT WRITETIME(%s) FROM %%s", column));
-        assertInvalidThrowMessage("Cannot use selection function maxwritetime on PRIMARY KEY part " + column,
-                                  InvalidRequestException.class,
-                                  String.format("SELECT MAXWRITETIME(%s) FROM %%s", column));
+                                  format("SELECT WRITETIME(%s) FROM %%s", column));
```
yifan-c, line 1212: add the assertion for `MAXWRITETIME` too.

src/java/org/apache/cassandra/serializers/MapSerializer.java
```diff
@@ -243,6 +247,96 @@ public ByteBuffer getSliceFromSerialized(ByteBuffer collection,
...
+            throw new MarshalException("Not enough bytes to read a map");
+        }
+    }
+
+    @Override
+    public Range<Integer> getIndexesRangeFromSerialized(ByteBuffer collection,
+                                                        ByteBuffer from,
+                                                        ByteBuffer to,
+                                                        AbstractType<?> comparator)
+    {
```
yifan-c: Add validation that `from <= to` if both are present. I tried to test with `s[2..0]` the query passes, which is a bit surprising.

src/java/org/apache/cassandra/serializers/SetSerializer.java
```diff
@@ -248,4 +253,92 @@ public ByteBuffer getSliceFromSerialized(ByteBuffer collection,
...
+            throw new MarshalException("Not enough bytes to read a set");
+        }
+    }
+
+    @Override
+    public Range<Integer> getIndexesRangeFromSerialized(ByteBuffer collection,
+                                                        ByteBuffer from,
+                                                        ByteBuffer to,
+                                                        AbstractType<?> comparator)
+    {
```
yifan-c: add `from <= to` if both present

src/java/org/apache/cassandra/serializers/MapSerializer.java
```diff
@@ -243,6 +247,96 @@ public ByteBuffer getSliceFromSerialized(ByteBuffer collection,
...
+            input.position(input.position() + sizeOfCollectionSize(n, ProtocolVersion.V3));
+            int start = from == ByteBufferUtil.UNSET_BYTE_BUFFER ? 0 : -1;
+            int end = to == ByteBufferUtil.UNSET_BYTE_BUFFER ? n : -1;
+
+            for (int i = 0; i < n; i++)
+            {
+                if (start >= 0 && end >= 0)
+                    break;
+                else if (i > 0)
+                    skipValue(input, ProtocolVersion.V3);
```
yifan-c: The logic is the same as the one in `SetSerializer` except those 2 lines. Maybe have a helper that combines both and adds a `valueSkipper` function, e.g. `Consumer<Input>`?

yifan-c (commented):
Took another pass today.
How is the forward compatibility? For example, the cluster is in the mixed mode and the old node (as the coordinator) receives the query with `WRITETIME` and it further queries replicas running both old and new versions.

src/java/org/apache/cassandra/cql3/selection/ElementsSelector.java
adelapena: Right, good catch, it should be a leftover. Removed.

src/java/org/apache/cassandra/cql3/selection/ElementsSelector.java
adelapena: Right, we can use the `type` in the parent class.

src/java/org/apache/cassandra/cql3/selection/ColumnTimestamps.java
adelapena: Changed to the proposed format and added some comments about the calculations.

src/java/org/apache/cassandra/serializers/MapSerializer.java
adelapena: Yep, those methods are derived from the previously existent `getSliceFromSerialized`. That method also has a considerable duplication between `SetSerializer` and `MapSerializer`.

I have moved the three `getSliceFromSerialized`, `getIndexFromSerialized` and `getIndexesRangeFromSerialized` methods to a common abstract superclass, since sets and maps have so much in common. That removes a good bunch of code duplication, although I'm not very sure about how to call that class. I have used `SetIndexedCollectionSerializer`, but any ideas are welcome.

src/java/org/apache/cassandra/serializers/MapSerializer.java
adelapena: That's the same behaviour that the regular selection of collection items has, independently of `writetime` and `ttl`. For example, with and without the proposed changes a query like `SELECT s[3..1] FROM t;` passes, and it always returns no data.

I'd say that type of intervals is strictly correct given that `[3, 1]` in interval notation just means `3 <= x <= 1`, which is useless but correct. However, I agree that we should consider throwing an error in that case because the user is probably giving those intervals by mistake. 

If we change the behaviour of collection ranges to reject that type of bounds, we should do it for all types of selections, not only `writetime` and `ttl`. And if we consider it as a bug we should backport the fix. So I think I'd keep the current behaviour on this ticket and would open a separate followup ticket. wdyt?

yifan-c: Make sense! I did not try the slice syntax with `SELECT`. It feels to me that a separate ticket is a better fit since it is on the syntax. It can help clear out some potential confusions.

src/java/org/apache/cassandra/serializers/SetIndexedCollectionSerializer.java
```diff
@@ -0,0 +1,209 @@
...
+            input.position(input.position() + sizeOfCollectionSize(n, ProtocolVersion.V3));
+            int start = from == ByteBufferUtil.UNSET_BYTE_BUFFER ? 0 : -1;
+            int end = to == ByteBufferUtil.UNSET_BYTE_BUFFER ? n : -1;
+
+            for (int i = 0; i < n; i++)
+            {
+                if (start >= 0 && end >= 0)
+                    break;
+                else if (hasValues && i > 0)
+                    skipValue(input, ProtocolVersion.V3);
```
yifan-c: Super weak nit: 
```java
    private void skipMapValue(ByteBuffer input, ProtocolVersion version)
    {
        if (hasValues)
            skipValue(input, version);
    }
```

src/java/org/apache/cassandra/serializers/SetIndexedCollectionSerializer.java
```diff
@@ -0,0 +1,209 @@
...
+
+import org.apache.cassandra.db.marshal.AbstractType;
+import org.apache.cassandra.db.marshal.ByteBufferAccessor;
+import org.apache.cassandra.transport.ProtocolVersion;
+import org.apache.cassandra.utils.ByteBufferUtil;
+
+/**
+ * Common superclass for {@link SetSerializer} and {@link MapSerializer}, considering a set as a map without values.
+ */
+abstract class SetIndexedCollectionSerializer<T> extends CollectionSerializer<T>
```
yifan-c: maybe `AbstractMapSerializer`? As you mentioned, a set is a map without values.

adelapena: Changed to `AbstractMapSerializer`.

adelapena:
> How is the forward compatibility? For example, the cluster is in the mixed mode and the old node (as the coordinator) receives the query with WRITETIME and it further queries replicas running both old and new versions.

In that case, if the column is not a collection the old coordinator would run the query as usual. Nothing has changed for regular columns so the new nodes would return the same. If the column is a collection the old coordinator would reject the query before that wasn't supported until now.

As for queries running on the new node, they would work even if there are not upgraded nodes, because everything happens on the coordinator's CQL layer. 

The problematic case would be aggregation queries such as `MAX(WRITETIME)` or `MAX(TTL)` running on a new node while there are not-upgraded nodes. Those queries won't work properly because the selector is serialized and sent to the not-upgraded nodes. So I think we should add a note advising against using these functions on not-frozen collections/UDTs when upgrading to 4.2, similar to the one that we added for `MAXWRITETIME`.

adelapena:
I have added checks to explicitly forbid:

- `MAXWRITETIME` on any column if the cluster contains nodes < 4.2.
- `WRITETIME` and `TTL` on multi-cell columns if the cluster contains nodes < 4.2.

Hopefully that will protect users ignoring the `NEWS.txt` entries.

I've also added a simple upgrade test for pre-upgrade, partial upgrade and full upgrade.

test/unit/org/apache/cassandra/cql3/validation/entities/WritetimeOrTTLTest.java
```diff
@@ -1050,6 +1052,47 @@ public void testFrozenNestedUDTs() throws Throwable
...
+        execute("UPDATE %s USING TIMESTAMP 20 AND TTL 2000 SET s = s + {20, 30} WHERE k = 1");
+
+        // Regular column
+        assertRows("SELECT min(v) FROM %s", row(0));
+        assertRows("SELECT max(v) FROM %s", row(1));
+        assertRows("SELECT writetime(v) FROM %s", row(10L), row(1L));
+        assertRows("SELECT min(writetime(v)) FROM %s", row(1L));
+        assertRows("SELECT max(writetime(v)) FROM %s", row(10L));
+        assertRows("SELECT min(maxwritetime(v)) FROM %s", row(1L));
+        assertRows("SELECT max(maxwritetime(v)) FROM %s", row(10L));
```
yifan-c, line 1071: hm... why `min` and `max` of `maxwritetime` are different? I'd expect `max(writetime(v)) == maxwritetime(v)`

adelapena: This is indeed confusing. The `max` function is not what perhaps we would like in this context, but an aggregation function that selects the row with the maximum column value. If it's applied to a collection column it doesn't select the max item within the collection, but the max collection among the selected rows. 

There is some followup on this in [CASSANDRA-17811](https://issues.apache.org/jira/browse/CASSANDRA-17811).

yifan-c: oh.. I see. 
maxwritetime(v) => [1, 10]
and max([1, 10]) => 10
and min([1, 10]) => 1
Confusing indeed, but nothing to do with this patch.

PR #1742, opened 2022-07-22
CASSANDRA-17575 3.11: Fix compaction token range bounds

adelapena:


src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -812,11 +812,20 @@ protected void runMayThrow()
...
+     *
+     * @param cfStore The column family store to be compacted.
+     * @param ranges The token ranges to be compacted, interpreted as closed intervals.
+     */
     public void forceCompactionForTokenRange(ColumnFamilyStore cfStore, Collection<Range<Token>> ranges)
     {
         Callable<Collection<AbstractCompactionTask>> taskCreator = () -> {
             Collection<SSTableReader> sstables = sstablesInBounds(cfStore, ranges);
-            if (sstables == null || sstables.isEmpty())
+            if (sstables.isEmpty())
```
ekaterinadimitrova2: This change is unrelated to the patch and my personal preference is to stick to the main patch in older versions

src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -265,8 +265,14 @@ public interface StorageServiceMBean extends NotificationEmitter
...
     public int relocateSSTables(int jobs, String keyspace, String ... cfnames) throws IOException, ExecutionException, InterruptedException;
+
     /**
-     * Forces major compaction of specified token range in a single keyspace
+     * Forces major compaction of specified token range in a single keyspace.
+     *
+     * @param keyspaceName the name of the keyspace to be compacted
+     * @param startToken the token at which the compaction range starts (inclusive)
+     * @param endToken the token at which compaction range ends (inclusive)
+     * @param tableNames the names of the tables to be compacted
```
ekaterinadimitrova2, line 275: Do we want also to add java doc to `forceKeyspaceCompactionForTokenRange` in `NodeProbe`, for completeness?

ekaterinadimitrova2 (approved):
Overall LGTM, left just 2 suggestions, thanks

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
adelapena: Fine with me. The unneeded null check isn't on trunk nor 4.1, so it will remain on 3.11 and 4.0.

src/java/org/apache/cassandra/service/StorageServiceMBean.java
adelapena: Makes sense, added.

PR #1748, opened 2022-07-25
17737 4.0

ekaterinadimitrova2:


src/java/org/apache/cassandra/service/GCInspector.java
```diff
@@ -55,7 +55,7 @@ public class GCInspector implements NotificationListener, GCInspectorMXBean
     public static final String MBEAN_NAME = "org.apache.cassandra.service:type=GCInspector";
     private static final Logger logger = LoggerFactory.getLogger(GCInspector.class);
     private volatile long gcLogThreshholdInMs = DatabaseDescriptor.getGCLogThreshold();
-    private volatile long gcWarnThreasholdInMs = DatabaseDescriptor.getGCWarnThreshold();
+    private volatile long gcWarnThresholdInMs = DatabaseDescriptor.getGCWarnThreshold();
```
adelapena: As in CASSANDRA-17735, I think I would remove `GCInspector`'s local copies of `Config`'s `gc_log_threshold_in_ms` and `gc_warn_threshold_in_ms`. Keeping those properties in a single place should easy synchronization. 

For example, `GCInspector#gcLogThreshholdInMs` and `GCInspector#gcWarnThresholdInMs` are marked as `volatile`, so changes to these properties via JMX should be visible for other JMX readers and `GCInspector` itself. However, `Config#gc_log_threshold_in_ms` and `Config#gc_warn_threshold_in_ms` aren't marked as `volatile`. Thus, changes done to those properties through JMX could not be visible for the settings virtual table. Conversely, we could see the opposite situation when we have writable virtual tables.

ekaterinadimitrova2: So what you suggest is just to make them volatile also in Config? And the long-int issue is already solved in the setter

adelapena: Setting them as `volatile` in `Config` should work, but I was thinking on removing the local copies on `GCInspector`, [this way](https://github.com/adelapena/cassandra/commit/cb1534ca0242c93277ea01511459604d03c86c9c), as we did in CASSANDRA-17735. wdyt?

ekaterinadimitrova2: Yes, sorry, I was not clear - that is exactly what I meant, making those in Config volatile further to removing the local copies

ekaterinadimitrova2:
Done

ekaterinadimitrova2:
Done

PR #1750, opened 2022-07-25
CASSANDRA-17776: When doing a host replacement, -Dcassandra.broadcast_interval_ms is used to know when to check the ring but checks that the ring wasn't changed in -Dcassandra.ring_delay_ms, changes to ring delay should not depend on when we publish load stats

dcapwell:


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1776,11 +1776,18 @@ public Collection<InetAddressAndPort> prepareForBootstrap(long schemaTimeoutMill
...
+                // who are aware of this relationship, this coupling should not be broken, but for most users this
+                // relationship isn't known and instead we should rely on the ring_delay.
+                // See CASSANDRA-17776
+                long sleepDelayMillis = Math.max(LoadBroadcaster.BROADCAST_INTERVAL, ringTimeoutMillis * 2);
                 try
                 {
                     // Sleep additionally to make sure that the server actually is not alive
                     // and giving it more time to gossip if alive.
-                    Thread.sleep(LoadBroadcaster.BROADCAST_INTERVAL);
+                    logger.info("Sleeping for {}ms waiting to make sure no new updates happen for {}", sleepDelayMillis, DatabaseDescriptor.getReplaceAddress());
```
jonmeredith: drive by: `updates happen` is vague, especially in StorageService - `gossip updates`, `metadata updates` or `token updates` might be more helpful to operators.

dcapwell: went with `gossip updates`

PR #1751, opened 2022-07-26
Cassandra 17774

jmckenzie-dev:


src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -2979,6 +2979,11 @@ public long[] getPerLevelSizeBytes()
         return compactionStrategyManager.getPerLevelSizeBytes();
     }
 
+    public int[] getSSTableCountPerTWCSBucket()
```
maedhroz, line 2990: nit: add `@Override`

src/java/org/apache/cassandra/db/compaction/CompactionStrategyManager.java
```diff
@@ -141,6 +146,8 @@ should update the compaction strategy in maybeReload() due to an ALTER.
     private volatile long maxSSTableSizeBytes;
     private volatile String name;
 
+    public static int TWCS_BUCKET_COUNT_MAX = 30;
```
maedhroz: I know we recommend 20-30 buckets, but if there are 31, are we just going to silently cut off the oldest one?

src/java/org/apache/cassandra/db/compaction/CompactionStrategyManager.java
```diff
@@ -610,6 +617,32 @@ public long[] getPerLevelSizeBytes()
...
     }
 
+    public int[] getSSTableCountPerTWCSBucket()
+    {
+        readLock.lock();
+        try
+        {
+            List<Map<Long, Integer>> countsByBucket = Stream.concat(
+                                                            StreamSupport.stream(repaired.allStrategies().spliterator(), false),
+                                                            StreamSupport.stream(unrepaired.allStrategies().spliterator(), false))
```
maedhroz: nit: Any way we can indent the concatenated streams here, so it's easier to read onward to `filter`?

src/java/org/apache/cassandra/db/compaction/TimeWindowCompactionStrategy.java
```diff
@@ -56,6 +58,7 @@ public class TimeWindowCompactionStrategy extends AbstractCompactionStrategy
     private final Set<SSTableReader> sstables = new HashSet<>();
     private long lastExpiredCheck;
     private long highestWindowSeen;
+    private Map<Long, Integer> sstableCountByBuckets = Collections.emptyMap();
```
maedhroz: if `sstableCountByBuckets` is accessed directly by a JMX thread, I think we want to make it volatile here.

src/java/org/apache/cassandra/db/ColumnFamilyStoreMBean.java
```diff
@@ -217,6 +217,12 @@ public List<String> importNewSSTables(Set<String> srcPaths,
      */
     public long[] getPerLevelSizeBytes();
 
+    /**
+     * @return sstable count for each bucket in TWCS. null unless time window compaction is used.
+     *         array index corresponds to bucket(int[0] is for most recent, ...).
```
maedhroz, line 222: Given its temporality, I might be inclined to want to read oldest to newest left to right. Is the thought here to emulate what we do w/ LCS where the levels move left to right? (i.e. The age of the data moves left to right?)

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
jmckenzie-dev: Will go ahead and fix the other MBean methods that are missing it as well. 👍

src/java/org/apache/cassandra/db/compaction/CompactionStrategyManager.java
jmckenzie-dev: Yes. We could up this, make it a configurable, both, neither, ? Nothing jumps out at me as an immediately convenient way to indicate to the end user that the value has been truncated when hitting up the JMX endpoint, and since we're not including it in the `tablestats` output we don't have that outlet to indicate truncation. You have any clever ideas here? :)

src/java/org/apache/cassandra/db/compaction/CompactionStrategyManager.java
jmckenzie-dev: Ew. Good catch.

src/java/org/apache/cassandra/db/compaction/TimeWindowCompactionStrategy.java
jmckenzie-dev: Subtle; good catch. This definitely opens this up to being accessed by multiple thread contexts. This tweak should give us slightly more up-to-date information in what we see via JMX if it just so happens to race w/the repair manager or compaction managers picking their next tasks. 👍

src/java/org/apache/cassandra/db/ColumnFamilyStoreMBean.java
jmckenzie-dev: I can't speak to the intent of the original author (will update commit message to reflect that it was Stefan's work; can check w/him), however the current shape of things is recency left to right, levels left to right, level 0 corresponding to "newest".

```
array index corresponds to level(int[0] is for level 0, ...)

array index corresponds to bucket(int[0] is for most recent, ...).
```

I don't really have an opinion on it excepting the current formatting provides logical symmetry between the two JMX calls and I like consistency even at the expense of intuition sometimes. 🤷

maedhroz: I'm fine with keeping it like this.

src/java/org/apache/cassandra/db/compaction/CompactionStrategyManager.java
maedhroz: The only easy improvement I can think of is to raise it to something that won't truncate w/ a slight misconfiguration. 128 maybe? Failing the JMX call because there are too many buckets seems like a bad option...

@krummas Thoughts?

maedhroz (approved):

PR #1754, opened 2022-07-26
17725 4.1

ekaterinadimitrova2:


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -906,6 +898,36 @@ else if (conf.max_value_size.toMebibytes() >= 2048)
         logInitializationOutcome(logger);
     }
 
+    @VisibleForTesting
+    static void validateUpperBoundStreamingConfig() throws ConfigurationException
+    {
+        // below 2 checks are needed in order to match the pre-CASSANDRA-15234 upper bound for those parameters which were still in megabits per second
+        if (conf.stream_throughput_outbound.toMegabitsPerSecond() >= Integer.MAX_VALUE)
+        {
+            throw new ConfigurationException("Invalid value of stream_throughput_outbound: " + conf.stream_throughput_outbound.toString(), false);
```
maedhroz, line 917: nit: It might be helpful to mention why something is invalid (i.e. the value overflows int bounds when converted to megabits) in the error message, but if this is only used for testing, I guess it's not a huge concern.

test/unit/org/apache/cassandra/db/KeyCacheTest.java
```diff
@@ -338,6 +340,24 @@ private void testKeyCache(String cf) throws ExecutionException, InterruptedExcep
         assertKeyCacheSize(noEarlyOpen ? 4 : 2, KEYSPACE1, cf);
     }
 
+    @Test
+    public void testKeyCacheLoadZeroCacheLoadTimeBackwardCompatibility() throws Exception
+    {
+        // Pre-4.1 users were asked to use negatives to disable. This test validates the backward compatibility
+        // in case a user still uses their old yaml format and negative values for this property.
+        DatabaseDescriptor.setCacheLoadTimeout(-1);
```
maedhroz, line 348: ...but doesn't `setCacheLoadTimeout()` here coerce the value to zero and, in fact, not test `NEGATIVE_SECONDS_DURATION` w/ `cache_load_timeout_seconds`? Seems like we'd want to reject negative values in `setCacheLoadTimeout()` but test w/ something like `cassandra-old.yaml` where `cache_load_timeout_seconds: -1` right?

maedhroz: Maybe like a special little version of `cassandra-old.yaml` that just looks at old default/disabled values in `LoadOldYAMLBackwardCompatibilityTest`?

src/java/org/apache/cassandra/config/DataRateSpec.java
```diff
@@ -285,27 +301,12 @@ public IntMebibytesPerSecondBound(double quantity, DataRateUnit unit)
         /**
          * Creates a {@code DataRateSpec.IntMebibytesPerSecondBound} of the specified amount in mebibytes per second.
          *
-         * @param mebibytesPerSecond where mebibytesPerSecond shouldn't be bigger than Long.MAX_VALUE-1
+         * @param mebibytesPerSecond where mebibytesPerSecond shouldn't be bigger than Integer.MAX_VALUE-1
```
maedhroz: nitty nit: We call this `max` in the downstream `validateQuantity()`, although it's actually one more than the max. I know I'm getting pedantic now, but perhaps we should just pass `Integer.MAX_VALUE - 1` and use `>` :D

src/java/org/apache/cassandra/config/DataRateSpec.java
```diff
@@ -252,6 +253,21 @@ public LongBytesPerSecondBound(long bytesPerSecond)
...
+        // which were in megabits per second in 4.0. Do not start using it for any new properties
+        public static LongBytesPerSecondBound megabitsPerSecondInBytesPerSecond(long megabitsPerSecond)
+        {
+            final double BYTES_PER_MEGABIT = 125_000;
+            double bytesPerSecond = (double) megabitsPerSecond * BYTES_PER_MEGABIT;
+
+            if (megabitsPerSecond >= Integer.MAX_VALUE)
+                throw new IllegalArgumentException("Invalid data rate: " + megabitsPerSecond + " megabits per second; " +
+                                                   "stream_throughput_outbound and inter_dc_stream_throughput_outbound" +
+                                                   " should be between 0 and " + Integer.MAX_VALUE + " in megabits per second");
```
maedhroz: `Integer.MAX_VALUE - 1`?

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -331,13 +331,13 @@ public MemtableOptions()
...
+    public volatile DataRateSpec.LongBytesPerSecondBound stream_throughput_outbound = new DataRateSpec.LongBytesPerSecondBound("24MiB/s");
+    @Replaces(oldName = "inter_dc_stream_throughput_outbound_megabits_per_sec", converter = Converters.MEGABITS_TO_BYTES_PER_SECOND_DATA_RATE, deprecated = true)
+    public volatile DataRateSpec.LongBytesPerSecondBound inter_dc_stream_throughput_outbound = new DataRateSpec.LongBytesPerSecondBound("24MiB/s");
 
-    public volatile DataRateSpec.IntMebibytesPerSecondBound entire_sstable_stream_throughput_outbound = new DataRateSpec.IntMebibytesPerSecondBound("24MiB/s");
-    public volatile DataRateSpec.IntMebibytesPerSecondBound entire_sstable_inter_dc_stream_throughput_outbound = new DataRateSpec.IntMebibytesPerSecondBound("24MiB/s");
+    public volatile DataRateSpec.LongBytesPerSecondBound entire_sstable_stream_throughput_outbound = new DataRateSpec.LongBytesPerSecondBound("24MiB/s");
+    public volatile DataRateSpec.LongBytesPerSecondBound entire_sstable_inter_dc_stream_throughput_outbound = new DataRateSpec.LongBytesPerSecondBound("24MiB/s");
```
maedhroz, line 342: So is `IntMebibytesPerSecondBound` now completely unused outside of tests?

maedhroz: The motivation here is that he rate limiter gets bytes at the end of the day, so we might as well just make sure this value is usable there?

src/java/org/apache/cassandra/tools/LoaderOptions.java
```diff
@@ -413,15 +413,25 @@ public Builder parseArgs(String cmdArgs[])
                     {
                         throw new ConfigurationException("Invalid value of inter_dc_stream_throughput_outbound: " + config.inter_dc_stream_throughput_outbound.toString(), false);
                     }
+
+                    if (config.entire_sstable_stream_throughput_outbound.toMebibytesPerSecond() >= Integer.MAX_VALUE)
+                    {
+                        throw new ConfigurationException("Invalid value of entire_sstable_stream_throughput_outbound: " + config.entire_sstable_stream_throughput_outbound.toString(), false);
```
maedhroz: nit: Might be helpful to mention that `config.entire_sstable_stream_throughput_outbound` is too large rather than just invalid?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1953,14 +1975,31 @@ public static int getCompactionThroughputMebibytesPerSecAsInt()
...
+        if (MEBIBYTES_PER_SECOND.toMebibytesPerSecond(value) >= Integer.MAX_VALUE)
+            throw new IllegalArgumentException("Invalid value of compaction_throughput: " + value);
+
+        conf.compaction_throughput = new DataRateSpec.LongBytesPerSecondBound(value);
+    }
+
     public static void setCompactionThroughputMebibytesPerSec(int value)
     {
-        conf.compaction_throughput = new DataRateSpec.IntMebibytesPerSecondBound(value);
+        if (MEBIBYTES_PER_SECOND.toMebibytesPerSecond(value) >= Integer.MAX_VALUE)
```
maedhroz: Isn't this a no-op? Like we could just say `value >= Integer.MAX_VALUE`?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1953,14 +1975,31 @@ public static int getCompactionThroughputMebibytesPerSecAsInt()
...
+
     public static double getCompactionThroughputMebibytesPerSec()
     {
         return conf.compaction_throughput.toMebibytesPerSecond();
     }
 
+    @VisibleForTesting // only for testing!
+    public static void setCompactionThroughputBytesPerSec(int value)
+    {
+        if (MEBIBYTES_PER_SECOND.toMebibytesPerSecond(value) >= Integer.MAX_VALUE)
```
maedhroz: Should this be `BYTES_PER_SECOND`? Looks like it would be a no-op currently...

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2016,59 +2055,115 @@ public static int getStreamThroughputOutboundMegabitsPerSec()
...
-        return conf.entire_sstable_stream_throughput_outbound.toMebibytesPerSecondAsInt();
+        return conf.stream_throughput_outbound.toMebibytesPerSecondAsInt();
+    }
+
+    public static void setStreamThroughputOutboundMebibytesPerSecAsInt(int value)
+    {
+        if (MEBIBYTES_PER_SECOND.toMegabitsPerSecond(value) >= Integer.MAX_VALUE)
+            throw new IllegalArgumentException("Invalid value of stream_throughput_outbound: " + value);
+
+        long valueInBytes = value * 1024L * 1024L;
```
maedhroz: nit: Could we use `MEBIBYTES_PER_SECOND.toBytesPerSecond(value)` here?

maedhroz: Same thing in `setCompactionThroughputMebibytesPerSec`, `setEntireSSTableStreamThroughputOutboundMebibytesPerSec`, `setInterDCStreamThroughputOutboundMebibytesPerSecAsInt`, and `setEntireSSTableInterDCStreamThroughputOutboundMebibytesPerSec`

src/java/org/apache/cassandra/config/DataRateSpec.java
```diff
@@ -212,7 +213,7 @@ public boolean equals(Object obj)
     @Override
     public String toString()
     {
-        return Math.round(quantity) + unit.symbol;
+        return DoubleMath.isMathematicalInteger(quantity) ? ((long)quantity + unit.symbol) : (quantity + unit.symbol);
```
maedhroz: nit: might be able to shorten slightly to...

```
return (DoubleMath.isMathematicalInteger(quantity) ? (long) quantity : quantity) + unit.symbol;
```

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -153,18 +153,18 @@ public CompactionMetrics getMetrics()
...
     }
 
     /**
      * Sets the rate for the rate limiter. When compaction_throughput is 0 or node is bootstrapping,
      * this sets the rate to Double.MAX_VALUE bytes per second.
-     * @param throughPutMiBPerSec throughput to set in MiB/s
+     * @param throughputBytesPerSec throughput to set in B/s
      */
-    public void setRate(final double throughPutMiBPerSec)
+    public void setRate(final double throughputBytesPerSec)
```
maedhroz: nit: maybe just change this back to `setRate(double throughput)`, ditch the local `throughput`, and let the doc comment clarify the units?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
maedhroz: ...and `setCompactionThroughputMbPerSec` in `StorageService`?

src/java/org/apache/cassandra/tools/nodetool/GetCompactionThroughput.java
```diff
@@ -17,17 +17,34 @@
...
+        double throughput = probe.getCompactionThroughputMebibytesAsDouble();
+
+        if (compactionThroughputAsDouble)
+            probe.output().out.println("Current compaction throughput: " + throughput + " MiB/s");
+        else
+        {
+            if (!DoubleMath.isMathematicalInteger(throughput))
+                throw new RuntimeException("You should use -d to get exact throughput in MiB/s");
+
+            probe.output().out.println("Current compaction throughput: " + probe.getCompactionThroughput() + " MB/s");
```
maedhroz, line 47: Leaving this `MB/s` for back-compat? Maybe we should just tolerate the risk and be consistent w/ everything else in this class.

src/java/org/apache/cassandra/tools/nodetool/GetCompactionThroughput.java
```diff
@@ -17,17 +17,34 @@
...
 import org.apache.cassandra.tools.NodeProbe;
 import org.apache.cassandra.tools.NodeTool.NodeToolCmd;
 
-@Command(name = "getcompactionthroughput", description = "Print the MiB/s throughput cap for compaction in the system")
+@Command(name = "getcompactionthroughput", description = "Print the MiB/s throughput cap for compaction in the system as a rounded number")
 public class GetCompactionThroughput extends NodeToolCmd
 {
+    @SuppressWarnings("UnusedDeclaration")
+    @Option(name = { "-d", "--precise-mib" }, description = "Print the MiB/s throughput cap for compaction in the system as a precise number (double)")
+    private boolean  compactionThroughputAsDouble;
```
maedhroz, line 33: It's unfortunate that existing tooling might be expecting an integer here, but I guess this is the best trade-off we can make for now. Perhaps we can deprecate the integral return value and just remove it in 5.0...

src/java/org/apache/cassandra/tools/nodetool/GetInterDCStreamThroughput.java
```diff
@@ -17,27 +17,71 @@
...
     @Override
     public void execute(NodeProbe probe)
     {
-        int throughput = entireSSTableThroughput ? probe.getEntireSSTableInterDCStreamThroughput() : probe.getInterDCStreamThroughput();
+        int throughput;
+        double throughputInDouble;
+
+        if (entireSSTableThroughput)
+        {
+            if(interDCStreamThroughputDoubleMbit || interDCStreamThroughputMiB)
```
maedhroz:
```suggestion
            if (interDCStreamThroughputDoubleMbit || interDCStreamThroughputMiB)
```

src/java/org/apache/cassandra/tools/nodetool/GetInterDCStreamThroughput.java
```diff
@@ -17,27 +17,71 @@
...
+                throw new IllegalArgumentException("You cannot use more than one flag with this command");
+
+            throughputInDouble = probe.getEntireSSTableInterDCStreamThroughput();
+            probe.output().out.printf("Current entire SSTable inter-datacenter stream throughput: %s%n",
+                                      throughputInDouble > 0 ? throughputInDouble + " MiB/s" : "unlimited");
+        }
+        else if (interDCStreamThroughputMiB)
+        {
+            if(interDCStreamThroughputDoubleMbit)
```
maedhroz:
```suggestion
            if (interDCStreamThroughputDoubleMbit)
```

src/java/org/apache/cassandra/tools/nodetool/GetInterDCStreamThroughput.java
```diff
@@ -17,27 +17,71 @@
...
+                throw new IllegalArgumentException("You cannot use more than one flag with this command");
+
+            throughputInDouble = probe.getInterDCStreamThroughputMibAsDouble();
+            probe.output().out.printf("Current inter-datacenter stream throughput: %s%n",
+                                      throughputInDouble > 0 ? throughputInDouble + " MiB/s" : "unlimited");
+
+        }
+        else if(interDCStreamThroughputDoubleMbit)
```
maedhroz:
```suggestion
        else if (interDCStreamThroughputDoubleMbit)
```

src/java/org/apache/cassandra/tools/nodetool/GetStreamThroughput.java
```diff
@@ -17,27 +17,70 @@
...
+            if(streamThroughputDoubleMbit || streamThroughputMiB)
+                throw new IllegalArgumentException("You cannot use more than one flag with this command");
+
+            throughputInDouble = probe.getEntireSSTableStreamThroughput();
+            probe.output().out.printf("Current entire SSTable stream throughput: %s%n",
+                                      throughputInDouble > 0 ? throughputInDouble + " MiB/s" : "unlimited");
+        }
+        else if (streamThroughputMiB)
+        {
+            if(streamThroughputDoubleMbit)
```
maedhroz:
```suggestion
            if (streamThroughputDoubleMbit)
```

src/java/org/apache/cassandra/tools/nodetool/GetStreamThroughput.java
```diff
@@ -17,27 +17,70 @@
...
     @Override
     public void execute(NodeProbe probe)
     {
-        int throughput = entireSSTableThroughput ? probe.getEntireSSTableStreamThroughput() : probe.getStreamThroughput();
+        int throughput;
+        double throughputInDouble;
+
+        if (entireSSTableThroughput)
+        {
+            if(streamThroughputDoubleMbit || streamThroughputMiB)
```
maedhroz:
```suggestion
            if (streamThroughputDoubleMbit || streamThroughputMiB)
```

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -315,7 +315,7 @@ public MemtableOptions()
     public Integer unlogged_batch_across_partitions_warn_threshold = 10;
     public volatile Integer concurrent_compactors;
     @Replaces(oldName = "compaction_throughput_mb_per_sec", converter = Converters.MEBIBYTES_PER_SECOND_DATA_RATE, deprecated = true)
-    public volatile DataRateSpec.IntMebibytesPerSecondBound compaction_throughput = new DataRateSpec.IntMebibytesPerSecondBound("16MiB/s");
```
frankgh, line 320: do we want to keep the `DataRateSpec.IntMebibytesPerSecondBound` class? I only see it being used in tests. Maybe remove early before someone starts relying on it.

src/java/org/apache/cassandra/config/DataRateSpec.java
frankgh: this changes behavior. `Math.round` of `999.999` will return -> `1000`. Whereas, `(long)999.999` returns `999`. Is this intended?

src/java/org/apache/cassandra/config/DataRateSpec.java
```diff
@@ -252,6 +253,21 @@ public LongBytesPerSecondBound(long bytesPerSecond)
         {
             this(bytesPerSecond, BYTES_PER_SECOND);
         }
+
+        // this one should be used only for backward compatibility for stream_throughput_outbound and inter_dc_stream_throughput_outbound
+        // which were in megabits per second in 4.0. Do not start using it for any new properties
+        public static LongBytesPerSecondBound megabitsPerSecondInBytesPerSecond(long megabitsPerSecond)
```
frankgh, line 259: should we annotate with `@Deprecated` if this is not intended to be used

src/java/org/apache/cassandra/config/DataRateSpec.java
```diff
@@ -252,6 +253,21 @@ public LongBytesPerSecondBound(long bytesPerSecond)
...
+        {
+            final double BYTES_PER_MEGABIT = 125_000;
```
frankgh: should we declare it as a field constant instead?

src/java/org/apache/cassandra/config/DataRateSpec.java
```diff
@@ -252,6 +253,21 @@ public LongBytesPerSecondBound(long bytesPerSecond)
...
+            double bytesPerSecond = (double) megabitsPerSecond * BYTES_PER_MEGABIT;
```
frankgh: Why use double here? long should be fine
```suggestion
            final long BYTES_PER_MEGABIT = 125_000L;
            long bytesPerSecond = megabitsPerSecond * BYTES_PER_MEGABIT;
```

src/java/org/apache/cassandra/config/DataRateSpec.java
```diff
@@ -252,6 +253,21 @@ public LongBytesPerSecondBound(long bytesPerSecond)
...
+
+            if (megabitsPerSecond >= Integer.MAX_VALUE)
```
frankgh, line 264: Did you mean `>` here?
```suggestion
            if (megabitsPerSecond > Integer.MAX_VALUE)
```

src/java/org/apache/cassandra/config/DataRateSpec.java
```diff
@@ -385,7 +386,7 @@ public double toMegabitsPerSecond(double d)
             {
                 if (d > MAX / (MEGABITS_PER_MEBIBYTE))
                     return MAX;
-                return Math.round(d * MEGABITS_PER_MEBIBYTE);
+                return d * MEGABITS_PER_MEBIBYTE;
```
frankgh, line 350: this also changes behavior but it's probably desired

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -906,6 +898,36 @@ else if (conf.max_value_size.toMebibytes() >= 2048)
         logInitializationOutcome(logger);
     }
 
+    @VisibleForTesting
+    static void validateUpperBoundStreamingConfig() throws ConfigurationException
+    {
+        // below 2 checks are needed in order to match the pre-CASSANDRA-15234 upper bound for those parameters which were still in megabits per second
+        if (conf.stream_throughput_outbound.toMegabitsPerSecond() >= Integer.MAX_VALUE)
```
frankgh, line 915: I'm confused why we exclude `Integer.MAX_VALUE`. Should this be `(conf.stream_throughput_outbound.toMegabitsPerSecond() > Integer.MAX_VALUE)` instead?

src/java/org/apache/cassandra/config/DataRateSpec.java
```diff
@@ -252,6 +253,21 @@ public LongBytesPerSecondBound(long bytesPerSecond)
...
+        {
+            final double BYTES_PER_MEGABIT = 125_000;
+            double bytesPerSecond = (double) megabitsPerSecond * BYTES_PER_MEGABIT;
+
+            if (megabitsPerSecond >= Integer.MAX_VALUE)
+                throw new IllegalArgumentException("Invalid data rate: " + megabitsPerSecond + " megabits per second; " +
+                                                   "stream_throughput_outbound and inter_dc_stream_throughput_outbound" +
+                                                   " should be between 0 and " + Integer.MAX_VALUE + " in megabits per second");
+
+            return new LongBytesPerSecondBound(bytesPerSecond, BYTES_PER_SECOND);
```
frankgh, line 269: `LongBytesPerSecondBound`'s constructor takes a double, but it should really take a long. We cast from double to long internally. I tried making `DataRateSpec`'s `quantity` a `long`, and tests are passing.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2016,59 +2055,115 @@ public static int getStreamThroughputOutboundMegabitsPerSec()
...
     {
         return conf.inter_dc_stream_throughput_outbound.toMebibytesPerSecond();
     }
 
+    public static double getInterDCStreamThroughputOutboundBytesPerSec()
+    {
+        return conf.inter_dc_stream_throughput_outbound.toBytesPerSecond();
+    }
+
+    public static int getInterDCStreamThroughputOutboundMebibytesPerSecAsInt()
```
frankgh, line 2147: is this only used for JMX calls?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -3186,7 +3281,10 @@ public static int getCacheLoadTimeout()
     @VisibleForTesting
     public static void setCacheLoadTimeout(int seconds)
     {
-        conf.cache_load_timeout = new DurationSpec.IntSecondsBound(seconds);
+        if (seconds < 0)
```
frankgh: minor nit: 
```suggestion
        if (seconds <= 0)
```

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2016,59 +2055,115 @@ public static int getStreamThroughputOutboundMegabitsPerSec()
...
+        return conf.entire_sstable_inter_dc_stream_throughput_outbound.toMebibytesPerSecond();
     }
 
     public static void setEntireSSTableInterDCStreamThroughputOutboundMebibytesPerSec(int value)
     {
-        conf.entire_sstable_inter_dc_stream_throughput_outbound = new DataRateSpec.IntMebibytesPerSecondBound(value);
+        if (MEBIBYTES_PER_SECOND.toMebibytesPerSecond(value) >= Integer.MAX_VALUE)
+            throw new IllegalArgumentException("Invalid value of entire_sstable_inter_dc_stream_throughput_outbound: " + value);
+
+        long valueInBytes = value * 1024L * 1024L;
```
frankgh: It feels like we do this a sufficient number of times that it might be worth defining a constant
` static final int ONE_MEBIBTE_IN_BYTES = 1024L * 1024L;`

src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -625,26 +625,71 @@ default int upgradeSSTables(String keyspaceName, boolean excludeCurrentVersion,
     public long getTruncateRpcTimeout();
 
     public void setStreamThroughputMbitPerSec(int value);
+    /**
+     * Please use the new getStreamThroughputMbitPerSecAsDouble method as this one will provide a rounded value
+     * @return stream_throughput_outbound in megabits
```
frankgh, line 629: maybe used the deprecated annotation for javadocs here?
```suggestion
     * @return stream_throughput_outbound in megabits
     * @deprecated Use getStreamThroughputMbitPerSecAsDouble instead
```

src/java/org/apache/cassandra/tools/nodetool/GetCompactionThroughput.java
```diff
@@ -17,17 +17,34 @@
...
     {
-        probe.output().out.println("Current compaction throughput: " + probe.getCompactionThroughput() + " MiB/s");
+        double throughput = probe.getCompactionThroughputMebibytesAsDouble();
+
+        if (compactionThroughputAsDouble)
+            probe.output().out.println("Current compaction throughput: " + throughput + " MiB/s");
+        else
+        {
+            if (!DoubleMath.isMathematicalInteger(throughput))
+                throw new RuntimeException("You should use -d to get exact throughput in MiB/s");
```
frankgh:
```suggestion
                throw new RuntimeException("Use the -d flag to quiet this error and get the exact throughput in MiB/s");
```

src/java/org/apache/cassandra/tools/nodetool/GetCompactionThroughput.java
frankgh: good point, when will we switch to MiB/s?

test/unit/org/apache/cassandra/tools/nodetool/SetGetEntireSSTableStreamThroughputTest.java
```diff
@@ -34,6 +34,8 @@
  */
 public class SetGetEntireSSTableStreamThroughputTest extends CQLTester
 {
+    private static final int MAX_INT_CONFIG_VALUE_MIB = Integer.MAX_VALUE - 1;
```
frankgh, line 37: ah ok, max value is `Integer.MAX_VALUE - 1`

frankgh (commented):
Thanks for the patch, I've left some comments

src/java/org/apache/cassandra/config/DataRateSpec.java
maedhroz: Yes, I think the point here was to avoid rounding, but still avoid printing a decimal point.

frankgh: should we just simplify to:
```suggestion
        return (long)quantity + unit.symbol;
```

src/java/org/apache/cassandra/config/DataRateSpec.java
maedhroz: If it's only used here, and should be removed at some point, I'm fine either way.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
maedhroz: See also https://github.com/apache/cassandra/pull/1754#discussion_r931567701

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
maedhroz: See also https://github.com/apache/cassandra/pull/1754#discussion_r931592327

src/java/org/apache/cassandra/config/DataRateSpec.java
maedhroz: We don't want to truncate either, right?

frankgh: yeah, good point

ekaterinadimitrova2: Yes, this should have been `Math.round(quantity) `and not `long(quantity)`, good catch, thanks!

src/java/org/apache/cassandra/tools/nodetool/GetCompactionThroughput.java
ekaterinadimitrova2: I was considering to do it in 5.0 in major, considering all the latest discussions on the ML

src/java/org/apache/cassandra/tools/nodetool/GetCompactionThroughput.java
ekaterinadimitrova2: Yeah, I will also document all this today in NEWS.txt and probably in other places to be more noisy for people to consider it and I hope we can get rid of it in 5.0

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
ekaterinadimitrova2: Yes, we can, yes, we should. I will change it, thanks for spotting it!

src/java/org/apache/cassandra/config/DataRateSpec.java
ekaterinadimitrova2: Yes, this one was desired and anyway now we don't use it when we switch to bytes

src/java/org/apache/cassandra/config/DataRateSpec.java
ekaterinadimitrova2: That is a good point if we consider we will have only BytesPerSecondBound and min units as the other new config classes but I think we want to get rid of the min units from now on. 
I guess we can just keep the doubles only in the DataRateSpec.DataRateUnit converters as we need doubles bytes -> mebibytes for example, see nodetool at the moment. 
Wondering whether we should make this change now or not, @maedhroz any thoughts?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
ekaterinadimitrova2: Yeah, makes sense not to do conversions, absolutely, thanks. I will also add more tests as @maedhorz suggested, I am actually working on special cases loading from yaml and from fromMap (old and new names) and checking the Settings Virtual Table for all Converters. And I think this is exactly what you are looking for here. I will share commit later today

src/java/org/apache/cassandra/config/DataRateSpec.java
maedhroz: I mean, truncating is fine if `isMathematicalInteger()`, right?

Otherwise, we could just always print the double.

ekaterinadimitrova2: Haha yes, definitely, sorry not my day :-( in that sense, both will work truncate and round but it doesn’t make sense to do round when w know already it is a round number…

test/unit/org/apache/cassandra/db/KeyCacheTest.java
ekaterinadimitrova2: Yes, I am writing better `Converters`'s tests to cover both loading from yaml and using `fromMap` method as part of another ticket. I will share it soon.

src/java/org/apache/cassandra/config/DataRateSpec.java
ekaterinadimitrova2: Correct, but let's not do it now, please... the patch became super big and If we change it here I think we have the same pattern in `DataStorageSpec` and `DurationSpec`...

src/java/org/apache/cassandra/config/Config.java
ekaterinadimitrova2: This is tested in the `SetGet*` tests, `assertSetGetValidThroughput `. if I didn't misunderstand you. There are those types of `asserts assertThat(StreamRateLimiter.getEntireSSTableInterDCRateLimiterRateInBytes()).isEqualTo(rateInBytes, withPrecision(0.01));`

maedhroz: Not asking for any new tests. Mostly just wondering if we need to keep `IntMebibytesPerSecondBound`.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
ekaterinadimitrova2: This whole method is messy... 
It should be changed to:
```
if (value == Integer.MAX_VALUE)
            throw new IllegalArgumentException("Invalid value of compaction_throughput: " + value);

        conf.compaction_throughput = new DataRateSpec.LongBytesPerSecondBound(value, MEBIBYTES_PER_SECOND);
```

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
ekaterinadimitrova2: Even better - `conf.stream_throughput_outbound = new DataRateSpec.LongBytesPerSecondBound(value, MEBIBYTES_PER_SECOND);`

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
ekaterinadimitrova2: This was actually originally throughputMbPerSec.
https://github.com/apache/cassandra/commit/5bb4bab12f8edfef95ed13cbabf8c0f377986065#diff-f7dd0237c343649f70b7ec9fefd7f6941a40b5164fd6063dce00fc09d2c234a7L163

It is not in the `CompactionManagerMbean` but it is public so not sure whether I have to worry about it actually...

src/java/org/apache/cassandra/config/DataRateSpec.java
ekaterinadimitrova2: It should work but I kept it consistent with bytesPerSecond which is double in the constructor.. which is the other question you had which I thinks falls back to weather we want to keep the mebibytes nested class here
@maedhroz any thoughts?

src/java/org/apache/cassandra/config/DataRateSpec.java
ekaterinadimitrova2: Nope, that is correct, in 4.1 we explicitly made Integer.MAX_VALUE and Long.MAX_VALUE illegal. This was done as they are anyway too big probably to be used and to differentiate from the overload handling in the units classes where in case of overflow we assign the MAX_VALUE anyway and we warned people that if they see it this will mean there is somewhere a bug and they need to raise a ticket

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
ekaterinadimitrova2: Yes

src/java/org/apache/cassandra/config/DataRateSpec.java
maedhroz: I'm fine w/ changing to `long` we won't have fractional bytes, right?

src/java/org/apache/cassandra/service/StorageServiceMBean.java
ekaterinadimitrova2: Good call, changed everywhere I had similar deprecation, thanks!

src/java/org/apache/cassandra/tools/nodetool/GetCompactionThroughput.java
ekaterinadimitrova2: Good call, I will change it also in the other getters, thanks!

src/java/org/apache/cassandra/config/DataRateSpec.java
ekaterinadimitrova2: Depends on whether we keep the `IntMebibytesBound`. I think probably we don't but we would want `IntBytesBound` at some point. I guess that one we can add in trunk so people can use it. We do not expect new properties in 4.1 anymore? The question is what do we want in this ticket... remove also IntMebibytesBound and switch to long in DataRateSpec, but not in the DataRateUnit? And people need to be clear they need to use double for any new getters that will return rounded value in bigger than bytes unit....

maedhroz: If there isn't anything that uses it, and we have no near-term idea of what might use it, I'd lean toward getting rid of `IntMebibytesBound`.

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
ekaterinadimitrova2: My OCD made me keep and deprecate `setRate `public method and add  `setRateInBytes` I will push soon the latest version post-review

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1962,14 +1985,35 @@ public static int getCompactionThroughputMebibytesPerSecAsInt()
...
+                                               " is too large; it should be less than " +
+                                               Integer.MAX_VALUE + " in MiB/s");
+
+        conf.compaction_throughput = new DataRateSpec.LongBytesPerSecondBound(value);
+    }
+
     public static void setCompactionThroughputMebibytesPerSec(int value)
     {
-        conf.compaction_throughput = new DataRateSpec.IntMebibytesPerSecondBound(value);
+        if (value == Integer.MAX_VALUE)
```
frankgh, line 2011: if this value overflows, it will be negative and  `DataRateSpec#validateQuantity` will throw an `IllegalArgException`. So this check is sufficient

ekaterinadimitrova2: My line of thinking is that we are expected to provide an int value so we cannot even provide something bigger to this method at first place. The JMX method does accept also int values only

PR #1755, opened 2022-07-26
Cassandra 17777 - Warn on unknown directories found in system keyspace directory rather than kill node during startup checks

jmckenzie-dev:


src/java/org/apache/cassandra/service/StartupChecks.java
```diff
@@ -571,6 +571,25 @@ public FileVisitResult visitFile(Path path, BasicFileAttributes attrs)
...
+
+                        if (tablePart.contains("-"))
+                            tablePart = tablePart.split("-")[0];
+
+                        // In very old versions of cassandra, we wouldn't necessarily delete sstables from dropped system tables
+                        // which were removed in various major version upgrades (e.g system.Versions in 1.2)
+                        if (ksPart.equals(SchemaConstants.SYSTEM_KEYSPACE_NAME) && !SystemKeyspace.ALL_TABLE_NAMES.contains(tablePart))
+                        {
+                            logger.warn("Found unknown system directory {}.{} at {} - this is likely left over from a previous " +
+                                        "version of cassandra and should be removed after inspection." , ksPart, tablePart, dir.toFile().getCanonicalPath());
```
smiklosovic: cassandra -> Cassandra

src/java/org/apache/cassandra/service/StartupChecks.java
```diff
@@ -571,6 +571,25 @@ public FileVisitResult visitFile(Path path, BasicFileAttributes attrs)
...
+                        String ksPart = nameParts[nameParts.length - 2];
+
+                        if (tablePart.contains("-"))
+                            tablePart = tablePart.split("-")[0];
+
+                        // In very old versions of cassandra, we wouldn't necessarily delete sstables from dropped system tables
+                        // which were removed in various major version upgrades (e.g system.Versions in 1.2)
+                        if (ksPart.equals(SchemaConstants.SYSTEM_KEYSPACE_NAME) && !SystemKeyspace.ALL_TABLE_NAMES.contains(tablePart))
+                        {
+                            logger.warn("Found unknown system directory {}.{} at {} - this is likely left over from a previous " +
```
smiklosovic: left over -> a leftover. I checked some dictionary and leftover is a proper noun.

src/java/org/apache/cassandra/service/StartupChecks.java
```diff
@@ -571,6 +571,25 @@ public FileVisitResult visitFile(Path path, BasicFileAttributes attrs)
...
+                    {
+                        String tablePart = nameParts[nameParts.length - 1];
+                        String ksPart = nameParts[nameParts.length - 2];
+
+                        if (tablePart.contains("-"))
+                            tablePart = tablePart.split("-")[0];
+
+                        // In very old versions of cassandra, we wouldn't necessarily delete sstables from dropped system tables
+                        // which were removed in various major version upgrades (e.g system.Versions in 1.2)
+                        if (ksPart.equals(SchemaConstants.SYSTEM_KEYSPACE_NAME) && !SystemKeyspace.ALL_TABLE_NAMES.contains(tablePart))
```
smiklosovic, line 585: @josh-mckenzie 

what happens when I have this on disk?

/var/lib/cassandra/data/demo/tb1-d26731900d2611edbc7601a9bc5ba989/snapshots/system/.tb1_blabla_idx/

This would be created when I take a snapshot with tag "system" on a table called tb1 which has an index on column called "blabla".

Is not it true that your check would detect this and emit the warning even though it is fine?

src/java/org/apache/cassandra/service/StartupChecks.java
jmckenzie-dev: I'm fine changing this but fwiw, searching through the codebase we're inconsistent with whether we capitalize or not.

src/java/org/apache/cassandra/service/StartupChecks.java
jmckenzie-dev: Think it's correct as "left over" as it's not directly modifying a noun: https://grammarist.com/spelling/leftover-left-over/#:~:text=The%20one%2Dword%2C%20unhyphenated%20leftover,e.g.%2C%20save%20the%20leftovers).

src/java/org/apache/cassandra/service/StartupChecks.java
jmckenzie-dev: Hm. It probably would yeah. We don't get any extra metadata in the preVisitDirectory callback to determine keyspace directory structure so no method to prevent this kind of "false positive" warning immediately arises. Maybe grepping for the word "snapshot" in the string and slightly tweaking the log message to indicate it could potentially be a snapshot if we find it?

smiklosovic: yeah something like that would be nice ... something like if that `if` is true and the third path "component" from the end is "snapshot", dont report it?

src/java/org/apache/cassandra/service/StartupChecks.java
smiklosovic: ok

src/java/org/apache/cassandra/service/StartupChecks.java
smiklosovic: as you wish

jmckenzie-dev: I'm more noodling on if we should open a JIRA to make this consistent across the codebase. It bothers me when I find inconsistent grammar on things (comments not capitalized, indentation, cassandra vs. Cassandra, etc)

src/java/org/apache/cassandra/service/StartupChecks.java
jmckenzie-dev: So another use-case that'd trip over this is if we create a snapshot of our system tables, which also happens to have "snapshots" as the 2nd path component from the end rather than 3rd if we have a -t system*. So I'm thinking we go with the couched phrasing (potentially, I see snapshot, etc) in the log message if we find the string "snapshot" in the path anywhere and otherwise go with the default message.

I think we should still report it but just phrase the log message differently; be more aggressive with saying "Hey, you can probably delete this after inspection" if it's not a snapshot and in system and we don't know it, or otherwise say "this is quite possibly a snapshot and legit, but still take a look"

jmckenzie-dev: Revised log message and level to info when includes the string `snapshot`.

PR #1756, opened 2022-07-26
CASSANDRA-17769 4.1 Remove usages of Path#toFile() in the snapshot apparatus

maedhroz:


maedhroz:
https://app.circleci.com/pipelines/github/maedhroz/cassandra?branch=CASSANDRA-17769-4.1

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -1197,6 +1197,11 @@ public static boolean isSecondaryIndexFolder(File dir)
         return dir.name().startsWith(SECONDARY_INDEX_NAME_SEPARATOR);
     }
 
+    public static boolean isSecondaryIndexFolder(Path dir)
+    {
+        return dir.getFileName().toString().startsWith(SECONDARY_INDEX_NAME_SEPARATOR);
```
dcapwell: `getFileName().toString()` really always hated that pattern =/

src/java/org/apache/cassandra/io/sstable/SSTableHeaderFix.java
```diff
@@ -301,7 +301,7 @@ void processFileOrDirectory(Path path)
               .filter(p -> {
                   try
                   {
-                      return Descriptor.fromFilenameWithComponent(new File(p.toFile())).right.type == Component.Type.DATA;
+                      return Descriptor.fromFilenameWithComponent(new File(p.toFile())).right.type == Component.Type.DATA; // checkstyle: permit this invocation
```
dcapwell: do we need to allow? `org.apache.cassandra.io.util.File#File(java.nio.file.Path)` we can just pass in the `p`?  this compiles for me locally

```
return Descriptor.fromFilenameWithComponent(new File(p)).right.type == Component.Type.DATA;
```

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
```diff
@@ -80,7 +80,7 @@ public Set<TableSnapshot> loadSnapshots()
         {
             try
             {
-                if (dataDir.toFile().exists())
+                if (Files.exists(dataDir))
```
dcapwell: When switching I always forget if there is a subtle difference with regard to bad disks...


`java.nio.file.Files#exists` does

```
} catch (IOException x) {
            // does not exist or unable to determine if file exists
            return false;
        }
```

and File calls `java.io.UnixFileSystem#getBooleanAttributes0`... which is native... it doesn't say it throws... so if disk is bad... does it fail or return false?

src/java/org/apache/cassandra/db/Directories.java
belliottsmith: `import static org.apache.cassandra.io.util.PathUtils.filename`

`filename(dir)`

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
belliottsmith: `java.io.File` rarely throws, and doesn't in this case, so the two are consistent behaviourally. It even goes so far as to return 0 for most methods that return a long in the case of error. 

Or you could simply do `new File(dataDir).exists()`

src/java/org/apache/cassandra/io/sstable/SSTableHeaderFix.java
maedhroz: That looks good. `SSTableHeaderFixTest` passes just fine w/ `new File(p)`.

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
maedhroz: I'll just use `o.a.c.io.util.File#exists()` here and in `BinLog`.

maedhroz:
@belliottsmith @dcapwell Made the suggested changes and kicked off the tests again. (The first run was 100% green.)

dcapwell (approved):

maedhroz:
Committed in https://github.com/apache/cassandra/commit/3a0936b61f63d8d7ab08bda411dd7d5963cafe0b

PR #1757, opened 2022-07-27
CASSANDRA-17780 - react on newer jemalloc implementations

smiklosovic:


bin/cassandra
```diff
@@ -166,7 +166,7 @@ case "`uname -s`" in
                 dirs="/lib64 /lib /usr/lib64 /usr/lib `cat /etc/ld.so.conf /etc/ld.so.conf.d/*.conf | grep '^/'`"
             fi
             dirs=`echo $dirs | tr " " ":"`
-            CASSANDRA_LIBJEMALLOC=$(find_library '.*/libjemalloc\.so\(\.1\)*' $dirs)
+            CASSANDRA_LIBJEMALLOC=$(find_library '.*/libjemalloc\.so\.*[1-9]*' $dirs)
```
smiklosovic: This will cover all libraries which are ending on a number, like:
````
/usr/lib/x86_64-linux-gnu/libjemalloc.so.1
/usr/lib/x86_64-linux-gnu/libjemalloc.so.2
...
````

File: bin/cassandra [hunk 2]
```diff
@@ -152,7 +152,7 @@ find_library()
     pattern=$1
     path=$(echo ${2} | tr ":" " ")
 
-    find $path -regex "$pattern" -print 2>/dev/null | head -n 1
+    find $path -regex "$pattern" -print 2>/dev/null | sort -r | head -n 1
```
smiklosovic: If we have more than one libjemalloc package installed, even by accident, like, both 1 and 2 versions, `sort -r` will sort it in descending order and head will take the first one so we always use the newest libjemalloc library.

bin/cassandra
```diff
@@ -132,21 +132,21 @@ find_library()
     pattern=$1
     path=$(echo ${2} | tr ":" " ")
 
-    find $path -regex "$pattern" -print 2>/dev/null | head -n 1
+    find $path -regex "$pattern" -print 2>/dev/null | sort -r | uniq | head -n 1
```
dcapwell, line 135: “sort -u” makes things uniq, so don’t need that other pipe

bin/cassandra
```diff
@@ -132,21 +132,21 @@ find_library()
...
 }
 case "`uname -s`" in
     Linux)
         if [ -z $CASSANDRA_LIBJEMALLOC ] ; then
             which ldconfig > /dev/null 2>&1
-            if [ $? = 0 ] ; then
+            if [ ! -f /etc/lsb-release ] ; then
```
dcapwell, line 141: The above check makes sure the command exists, would keep it unless you found a bug.

dcapwell (changes requested):
Once the if conditions is resolved I am +1

PR #1758, opened 2022-07-27
17737 4.1 sstable preemptive open interval in mb

ekaterinadimitrova2:


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -3074,12 +3074,17 @@ public static void setMigrateKeycacheOnCompaction(boolean migrateCacheEntry)
...
     {
+        if (conf.sstable_preemptive_open_interval == null)
+            return -1;
         return conf.sstable_preemptive_open_interval.toMebibytes();
     }
 
     public static void setSSTablePreemptiveOpenIntervalInMiB(int mib)
     {
-        conf.sstable_preemptive_open_interval = new DataStorageSpec.IntMebibytesBound(mib);
+        if (mib == -1)
```
jonmeredith:
```suggestion
        if (mib < 0)
```

conf/cassandra.yaml
```diff
@@ -992,6 +992,8 @@ compaction_throughput: 64MiB/s
 # are completely written, and used in place of the prior sstables for
 # any range that has been written. This helps to smoothly transfer reads 
 # between the sstables, reducing page cache churn and keeping hot rows hot
+# Set sstable_preemptive_open_interval to null for disabled which is equivalent to the sstable_preemptive_open_interval_in_mb
+# being negative
```
jonmeredith: Comment is a lot longer than the rest of the block, would you mind wrapping it early. I'm guessing we still refer to deprecated parameters in other comments and we can remove them if the deprecated settings are actually removed.

jonmeredith: How do you set to `null`? I tried 

`sstable_preemptive_open_interval: null`

but got this on startup.
```
Exception (org.apache.cassandra.exceptions.ConfigurationException) encountered during startup: Invalid yaml. Those properties [sstable_preemptive_open_interval] are not valid
ERROR [main] 2022-07-27 11:34:46,521 CassandraDaemon.java:896 - Exception encountered during startup: Invalid yaml. Those properties [sstable_preemptive_open_interval] are not valid
Invalid yaml. Those properties [sstable_preemptive_open_interval] are not valid
```

then
`sstable_preemptive_open_interval:`

and got
```Exception (org.apache.cassandra.exceptions.ConfigurationException) encountered during startup: Invalid yaml. Those properties [sstable_preemptive_open_interval] are not valid
ERROR [main] 2022-07-27 11:37:15,213 CassandraDaemon.java:896 - Exception encountered during startup: Invalid yaml. Those properties [sstable_preemptive_open_interval] are not valid
Invalid yaml. Those properties [sstable_preemptive_open_interval] are not valid```

and of course `-1` still fails as expected.

**Review by jonmeredith — COMMENTED** [review]
Checking the virtual tables, I used the deprecated `sstable_preemptive_open_interval_in_mb: -1` until I understand how to set the newer version. The backward compatibility doesn't handle converting back to `sstable_preemptive_open_interval_in_mb` and reports as `null`. I'm happy to live with it if it isn't an easy fix.

```jmeredith@Jons-MacBook-Pro j11-4.1-work % bin/cqlsh
Connected to Test Cluster at 127.0.0.1:9042
[cqlsh 6.1.0 | Cassandra 4.1-alpha2-SNAPSHOT | CQL spec 3.4.6 | Native protocol v5]
Use HELP for help.
cqlsh> SELECT * FROM system_views.settings WHERE name = 'sstable_preemptive_open_interval';

 name                             | value
----------------------------------+-------
 sstable_preemptive_open_interval |  null

(1 rows)
cqlsh> SELECT * FROM system_views.settings WHERE name = 'sstable_preemptive_open_interval_in_mb';

 name                                   | value
----------------------------------------+-------
 sstable_preemptive_open_interval_in_mb |  null
```
Then tried setting to 99 using JMX which worked fine.
```
(1 rows)
cqlsh> SELECT * FROM system_views.settings WHERE name = 'sstable_preemptive_open_interval_in_mb';

 name                                   | value
----------------------------------------+-------
 sstable_preemptive_open_interval_in_mb |    99

(1 rows)
cqlsh> SELECT * FROM system_views.settings WHERE name = 'sstable_preemptive_open_interval';

 name                             | value
----------------------------------+-------
 sstable_preemptive_open_interval | 99MiB

(1 rows)
```

conf/cassandra.yaml
ekaterinadimitrova2: Good catch, null is valid only if it is default value in the Config class which is not the case here :( 
I should have seen that. I will have to think about how to workaround it and add a startup test as now I double checked and it seems I made the same mistake yesterday with one more property so now I have to fix both of them. Thank you for catching it!

About the -1, it starts for me with sstable_preemptive_open_interval_in_mb = -1, then we see in the logs the converter loaded it as sstable_preemptive_open_interval = null; But I definitely null doesn't work.

ekaterinadimitrova2: Thinking out loud.... we can do probably a mix of what you suggested initially and this fix for this special case. We can add a flag parameter to enable/disable which will be used with the new property if we want to disable it. If someone uses the old one we should prohibit setting the new flag then in parallel on startup if the old one was found in the yaml, WDYT?

ekaterinadimitrova2:
About the VT, the conversion happens in the last method in the converter which was not handling the special case, I will fix it in a moment.
The reverseConverter in Converter is what handles this, it should be in this case `o -> o ==  null ? -1 : o.toMebibytes()),`

ekaterinadimitrova2:
Ha...about the VT, it seems that someone added null check in the unconvert() method and broke the special cases we have in the converter... I will see to fix it and add more tests to prevent that from happening in the future

conf/cassandra.yaml
ekaterinadimitrova2: I might be able to workaround that need of additional flag actually and special case those two properties so that they do not error out when set to null.

jonmeredith: I think a separate flag would be a good idea and cleaner than relying on magic values to control whether features are enabled or not.  Does the new config framework have the capability to drive to settings from the deprecated `sstable_preemptive_open_interval_in_mb`?

ekaterinadimitrova2: I already fixed it, but if we want a flag we can do that of course. The thing is the usage of null was accepted and documented long time ago, if we do it here differently, does this mean we need to change it in all other places we did use null?
In that case this would fix the issue with null and the VT as it seems that check is important for a work done in CASSANDRA-17166 as if I remove it VT throws currently null pointer exception. I have to check what @dcapwell did in that patch as I was not really working on that ticket. (CASSANDRA-17166)  

Looking back at the `Converters`, there are 4 other properties where we depend on `null`:
`credentials_update_interval, permissions_update_interval, roles_update_interval` - those are actually interesting in the context of this ticket as JMX uses the getters and returns their `*_validity` equivalent if they are null but VT shows null/-1... 
I will produce a patch for them separately. This is also a 4.0 bug but there we use -1, not null
The 4-th one dependent on the null value (which I just fixed to be able to start with null, seems I didn't fully fix it the other day...) is `index_summary_resize_interval`. I will fix it after we agree which direction we prefer - flags or the null value. Now the only uncomfortable situation I see is that people will have to maintain two properties for one old one and if we continue adding feature flags and not removing properties, our config becomes too big and scary. 

`Does the new config framework have the capability to drive to settings from the deprecated sstable_preemptive_open_interval_in_mb?`

If you set the old one with `sstable_preemptive_open_interval_in_mb=-1` in cassandra.yaml, it loads internally through the `@Replaces` annotation and the `Converters` class into `sstable_preemptive_open_interval=null`. That is how the backward compatibility works on the parsing level for cassandra.yaml. I hope I didn't misunderstood your question.

ekaterinadimitrova2: Oh, we actually have guardrails of the newly introduced configuration types and they use disabled = null, as they are enabled per guardrail, the big guardrails feature flag was removed and they are individually enabled. Adding an enable flag for every guardrail will be a pain so I think we should stay with the null option, at least for those.

ekaterinadimitrova2: The more I think, the more I am convinced enable/disable flags should work for features and not individual properties

jonmeredith: > I hope I didn't misunderstood your question.

I was asking if `sstable_preemptive_open_interval_in_mb` could set a new `sstable_preemptive_open_enabled` entry as well as `sstable_preemptive_open_interval`.

If you have a fix for setting `null` and don't need to create an `enabled` flag I'm fine with that. I thought you were saying you might need to create on to resolve this problem.

ekaterinadimitrova2: `I was asking if sstable_preemptive_open_interval_in_mb could set a new sstable_preemptive_open_enabled entry as well as sstable_preemptive_open_interval.`
No, currently we map one or more old names to only one new property. 

`If you have a fix for setting null and don't need to create an enabled flag I'm fine with that. I thought you were saying you might need to create on to resolve this problem.`
I created a quick one by special casing this particular property to be able to set it to null even if the default value is not null. So I will need to figure out then only the VT Converters

ekaterinadimitrova2: ...and more tests to prevent us from future regressions of course :-)

ekaterinadimitrova2: Oh, seems it was me not that ticket that broke it (that one just changes to return null and not 0)... I already have a fix, adding more tests, will push a patch later tonight

ekaterinadimitrova2:
@jonmeredith I just pushed review commits to address your feedback and follow up discussion plus improved the testing. The only current issue I see is with the `fromMap` in the configuration loader which does not accept currently null values for properties that are not null by default in the Config class. I will try to figure it out tomorrow but at least we have a workaround and I wanted to show you what I came up with so far. Currently that method is also only used by the in-jvm test framework. 
The Converter was broken when we introduced null post CASSANDRA-15234, I will add to the new tests for the rest of the properties and converters in a separate commit tomorrow. I didn't want to overwhelm this one. Only fixed the issue with `index_summary_resize_interval=null` which was the same as the one with `sstable_preemptive_open_interval=null` on startup

.build/build-rat.xml
```diff
@@ -58,6 +58,8 @@
                  <exclude NAME="**/doc/antora.yml"/>
                  <exclude name="**/test/conf/cassandra.yaml"/>
                  <exclude name="**/test/conf/cassandra-old.yaml"/>
+                 <exclude name="**/cassandra-converters-special-cases-old-names.yaml"/>
```
adelapena: I think this should be `<exclude name="**/test/conf/cassandra-converters-special-cases-old-names.yaml"/>`

NEWS.txt
```diff
@@ -165,6 +165,9 @@ New features
 
 Upgrading
 ---------
+    - `sstable_preemptive_open_interval_in_mb` being negatve for disabled is equivalent to `sstable_preemptive_open_interval`
```
adelapena:
```suggestion
    - `sstable_preemptive_open_interval_in_mb` being negative for disabled is equivalent to `sstable_preemptive_open_interval`
```

test/conf/cassandra-converters-special-cases-old-names.yaml
```diff
@@ -0,0 +1,3 @@
+sstable_preemptive_open_interval_in_mb: -1
```
adelapena, line 5: Maybe we could add a brief comment at the beginning of this file indicating what it's used for.

test/conf/cassandra-converters-special-cases.yaml
```diff
@@ -0,0 +1,3 @@
+sstable_preemptive_open_interval:
```
adelapena, line 5: Maybe we could add a brief comment at the beginning of this file indicating what it's used for.

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -423,7 +425,8 @@ public MemtableOptions()
     @Replaces(oldName = "trickle_fsync_interval_in_kb", converter = Converters.KIBIBYTES_DATASTORAGE, deprecated = true)
     public DataStorageSpec.IntKibibytesBound trickle_fsync_interval = new DataStorageSpec.IntKibibytesBound("10240KiB");
 
-    @Replaces(oldName = "sstable_preemptive_open_interval_in_mb", converter = Converters.MEBIBYTES_DATA_STORAGE_INT, deprecated = true)
+    @Nullable
+    @Replaces(oldName = "sstable_preemptive_open_interval_in_mb", converter = Converters.NEGATIVE_DATA_STORAGE_INT, deprecated = true)
```
adelapena: Should't this keep including `MEBIBYTES` in the name, as in `NEGATIVE_MEBIBYTES_DATA_STORAGE_INT` or `MEBIBYTES_CUSTOM_DATA_STORAGE_INT`?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -3074,12 +3074,17 @@ public static void setMigrateKeycacheOnCompaction(boolean migrateCacheEntry)
 
     public static int getSSTablePreemptiveOpenIntervalInMiB()
     {
+        if (conf.sstable_preemptive_open_interval == null)
```
adelapena, line 3078: I'd add some JavaDoc to the getter mentioning that it might return negative.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -3074,12 +3074,17 @@ public static void setMigrateKeycacheOnCompaction(boolean migrateCacheEntry)
...
+            return -1;
         return conf.sstable_preemptive_open_interval.toMebibytes();
     }
 
     public static void setSSTablePreemptiveOpenIntervalInMiB(int mib)
     {
-        conf.sstable_preemptive_open_interval = new DataStorageSpec.IntMebibytesBound(mib);
+        if (mib < 0)
```
adelapena, line 3086: I'd add some JavaDoc to this setter mentioning that it accepts negative values.

.build/build-rat.xml
ekaterinadimitrova2: Good point, it still works this way (tested locally) but it is inconsistent, I will  change it, thanks

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
ekaterinadimitrova2: This info is in the JavaDoc for the JMX method but I guess it doesn't hurt to add it here too.

ekaterinadimitrova2:
@adelapena I addressed your feedback. The new tests pass locally, but I started full CI #1823 https://app.circleci.com/pipelines/github/ekaterinadimitrova2/cassandra?branch=17737-4.1-sstable_preemptive_open_interval_in_mb

ekaterinadimitrova2:
CI shows some issue with @Nullable, I think I need to change the scope in build.xml for the annotation, I will work it now. I will let you know when I have the fix pushed

ekaterinadimitrova2:
com.google.code.findbugs needs to have default compile and not provided scope for this patch to work. 
Looking into its license https://opensource.org/licenses/BSD-3-Clause
 and reading here - https://www.apache.org/legal/resolved.html
I think this is fine. Do we need to bring it to the mailing list? I think only introducing brand new dependencies? WDYT? 
CI - https://app.circleci.com/pipelines/github/ekaterinadimitrova2/cassandra?branch=17737-4.1-sstable_preemptive_open_interval_in_mb #1824 still running

src/java/org/apache/cassandra/config/YamlConfigurationLoader.java
```diff
@@ -363,10 +365,12 @@ else if (result.getAnnotation(Deprecated.class) != null)
 
             return new ForwardingProperty(result.getName(), result)
             {
+                boolean allowsNull = result.getAnnotation(Nullable.class) != null;
+
                 @Override
                 public void set(Object object, Object value) throws Exception
                 {
-                    if (value == null && get(object) != null)
+                    if (value == null && get(object) != null && !allowsNull)
```
jonmeredith, line 374: It seems inconsistent to only require an `Nullable` annotation for configuration that can be set null, but is not defaulted to be null.  What do you think about dropping the `get(object) != null` check, and applying `@Nullable` to all cases in Config that can be legitimately set to `null`?

ekaterinadimitrova2: I thought about that yesterday but my concern is that there are too many of those and they are not only from the new types and not only new ones...
Maybe we can open a follow up ticket only on trunk as an improvement? And add more testing, etc... I am afraid we are getting into another rabbit hole.
Also, it is good if we can keep this case as an exception and not make people think that it is a good practice. Why? Because if they start adding other default values and do not test properly that the null value is handled (and I am sure someone will do that) this can be reaaaally bad. What do you think?

jonmeredith: Understand about rabbit holes. I'd be happy with opening a follow up JIRA and adding a comment referencing it in this method.

ekaterinadimitrova2: Ticket opened, I linked it to this ticket and added a TODO comment

src/java/org/apache/cassandra/config/Converters.java
```diff
@@ -69,44 +69,47 @@
...
                                 DataStorageSpec.LongMebibytesBound::new,
-                                DataStorageSpec.LongMebibytesBound::toMebibytes),
+                                o -> o == null ? null : o.toMebibytes()),
     MEBIBYTES_DATA_STORAGE_INT(Integer.class, DataStorageSpec.IntMebibytesBound.class,
                                DataStorageSpec.IntMebibytesBound::new,
-                               DataStorageSpec.IntMebibytesBound::toMebibytes),
+                               o -> o == null ? null : o.toMebibytes()),
+    NEGATIVE_MEBIBYTES_DATA_STORAGE_INT(Integer.class, DataStorageSpec.IntMebibytesBound.class,
+                              o -> o < 0 ? null : new DataStorageSpec.IntMebibytesBound(o),
+                              o -> o == null ? -1 : o.toMebibytes()),
```
adelapena, line 88: Trivial nit that can be addressed on commit: the parameters are misaligned after renaming the converter.

adelapena:
The last changes look good to me. It seems that CI hasn't been started. Also we'd need to apply the changes to the patch for trunk.

jonmeredith (approved):
Looks good to me - assuming tests are passing and changes  on this PR are propagated up to `trunk`

PR #1759, opened 2022-07-27
17737 trunk sstable preemptive open interval in mb

ekaterinadimitrova2:


test/unit/org/apache/cassandra/config/YamlConfigurationLoaderTest.java
```diff
@@ -311,6 +311,10 @@ public void converters()
         assertThatThrownBy(() -> from("stream_throughput_outbound_megabits_per_sec", -2).stream_throughput_outbound.toMegabitsPerSecondAsInt())
         .hasRootCauseInstanceOf(IllegalArgumentException.class)
         .hasRootCauseMessage("Invalid data rate: value must be non-negative");
+
+        // NEGATIVE_DATA_STORAGE_INT
+        assertThat(from("sstable_preemptive_open_interval_in_mb", "1").sstable_preemptive_open_interval.toMebibytes()).isEqualTo(1);
+        assertThat(from("sstable_preemptive_open_interval_in_mb", -2).sstable_preemptive_open_interval).isNull();
     }
```
Claudenw, line 318: Should there be a test here for 
`from("sstable_preemptive_open_interval_in_mb", "0")`

ekaterinadimitrova2: The idea of this test is to check the two branches of the respective converter, in this case:
`o -> o < 0 ? null : new DataStorageSpec.IntMebibytesBound(o)`

0 or 1 for the test it doesn't matter. IntMebibytesBound accepts non-negative numbers and 0 is not a special case.

Or is there anything special about 0 and that particular property?

Claudenw: Just that it sits on the boundary and the test would make it clear that 0 sets bound of 0 while -1 sets no bound, and 1 shows that the bound actually gets set to the proper value.

ekaterinadimitrova2: I am not sure I understand you. Both 0 and 1 are valid values and they can be set, as well as any other non-negative number up to Integer.MAX_VALUE.
There is nothing special about 0 here. It is the lowest number, but that’s all. We have extensive testing for IntMebibytesBound in the DataStorageSpec unit tests

Claudenw: It is just a matter of test point choices.  I would have chosen 0 and -1 to exhibit the change at the boundary.   But if you think the test is sufficient, I have no objections to proceeding.

ekaterinadimitrova2: By the way, I just realized that this PR is still open as I was having trouble to close any PRs (hitting some weird Github bug) but the work was already accepted and committed last week

ekaterinadimitrova2: `It is just a matter of test point choices. I would have chosen 0 and -1 to exhibit the change at the boundary. But if you think the test is sufficient, I have no objections to proceeding.`
It seems our last comments were posted at the same time.
Thank for your feedback. I can incorporate it later, this commit was already done but I will keep in mind to add it as part of the next one. I don’t see any particular lack of testing here but I understand your point to exhibit the exact boundary.

PR #1761, opened 2022-07-27
Cassandra 17781 - Add separate thread pool for Secondary Index building so it doesn't block compactions

jmckenzie-dev:


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1980,6 +1980,11 @@ public static int getConcurrentValidations()
         return conf.concurrent_validations;
     }
 
+    public static int getConcurrentIndexBuilds()
```
smiklosovic: the method should be probably called "getConcurrentIndexBuilders" to be aligned with the terminology we use for other methods around here. `conf.concurrent_index_builds` would be probably changed to `conf.concurrent_index_builders` as well.

jmckenzie-dev: Good catch on both; revised.

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -325,6 +325,9 @@ public MemtableOptions()
     public volatile int concurrent_materialized_view_builders = 1;
     public volatile int reject_repair_compaction_threshold = Integer.MAX_VALUE;
 
+    // The number of executors to use for building secondary indexes
+    public int concurrent_index_builders = 2;
```
maedhroz, line 329: @josh-mckenzie Should we make a hot-prop out of this?

maedhroz (approved):

src/java/org/apache/cassandra/config/Config.java
jmckenzie-dev: 🤔 I think we should have a JMX extension for it and promote it as a first class citizen in the .yaml. If we end up using this same thread pool for index building for SASI it's probably something that's going to see _a  lot_ more usage than it is today.

I originally kept it as the very primitive "let's just break the scheduling coupling" with the suspicion that https://issues.apache.org/jira/browse/CASSANDRA-11218 would make this work here moot, however now that I think on it a bit more there's no predicting when and if that other work will be completed.

PR #1769, opened 2022-08-05
CASSANDRA-17740 Avoid initializing schema via SystemKeyspace.getPreferredIP() with the BulkLoader tool (4.1)

maedhroz:


maedhroz:
https://app.circleci.com/pipelines/github/maedhroz/cassandra?branch=CASSANDRA-17740-4.1

test/unit/org/apache/cassandra/tools/ToolRunner.java
```diff
@@ -402,8 +400,11 @@ public Exception getException()
          */
         public void assertCleanStdErr()
         {
-            assertTrue("Failed because cleaned stdErr wasn't empty: " + getCleanedStderr(),
-                       getCleanedStderr().isEmpty());
+            String raw = getStderr();
+            String cleaned = getCleanedStderr();
+            assertTrue("Failed to clean stderr completely.\nRaw (length=" + raw.length() + "):\n" + raw + 
+                       "\nCleaned (length=" + cleaned.length() + "):\n" + cleaned,
+                       cleaned.trim().isEmpty());
```
maedhroz, line 407: Not strictly necessary for the fix, but it makes clean stderr issues easier to debug.

src/java/org/apache/cassandra/tools/BulkLoadConnectionFactory.java
```diff
@@ -70,4 +70,9 @@ private NettyStreamingChannel create(OutboundConnectionSettings template, int me
```
maedhroz, line 70: If a `connectTo` is always set in this last `create()` overload, the comment in the second `create()` is incorrect, given the preferred address is always overridden. From an API perspective, the `preferred` address is completely ignored, and the second `create()` overload is useless, and we could just remove it. (The `Factory` interface has a default method that ignores `preferred`.)

maedhroz: ...and just for the curious, I think we can't try to make this compatible w/ `EC2Snitch` either, given the whole point of this issue is that getting the preferred IP from the system keyspace is a no-no.

maedhroz, line 71: The comment in the second `create()` method is incorrect now. The preferred address is always ignored, not just ignored when encryption is enabled. We can't make this compatible w/ `EC2Snitch` (using the preferred address), so the only reason that overload exists is so we don't call the `NettyStreamingConnectionFactory` implementation.

jonmeredith (approved):

maedhroz:
Committed as https://github.com/apache/cassandra/commit/5e39c54678b0d70145b5c699593eb9ce7fa7706a

PR #1770, opened 2022-08-05
CASSANDRA-17792: Fix race condition on updating cdc size and advancing to next segment

yifan-c:


src/java/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDC.java
```diff
@@ -190,14 +189,20 @@ public CommitLogSegment.Allocation allocate(Mutation mutation, int size) throws
...
+    // - Non-blocking mode has just recently been enabled for CDC.
+    // - The CDC total space has droppped below the limit (e.g. CDC consumer cleans up).
+    private void permitSegmentMaybe(CommitLogSegment segment)
     {
-        if (!DatabaseDescriptor.getCDCBlockWrites() && segment.getCDCState() == CDCState.FORBIDDEN)
+        if (segment.getCDCState() != CDCState.FORBIDDEN)
+            return;
+
+        if (!DatabaseDescriptor.getCDCBlockWrites()
+            || cdcSizeTracker.sizeInProgress.get() < DatabaseDescriptor.getCDCTotalSpace())
```
jmckenzie-dev: We _may_ want to compare to DBD.getCDCTotalSpace - the size of a CL Segment. A hypothetical example of an edge case w/this logic here:
1. CommitLogSegment size 128mb
2. getCDCTotalSpace = 3.95G out of 4.0G
3. we see that getCDCTotalSpace is < max, so we flip the bit
4. Then we overflow by `CommitLogSegment size - (4.0 - 3.95)`

OR - we just leave it as is because this is likely not a big problem and maybe document it. 🤷

src/java/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDC.java
```diff
@@ -190,14 +189,20 @@ public CommitLogSegment.Allocation allocate(Mutation mutation, int size) throws
...
         {
             segment.setCDCState(CDCState.PERMITTED);
         }
+
```
jmckenzie-dev: nit: extra whitespace here

src/java/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDC.java
```diff
@@ -219,6 +224,8 @@ private void throwIfForbidden(Mutation mutation, CommitLogSegment segment) throw
     /**
      * On segment creation, flag whether the segment should accept CDC mutations or not based on the total currently
      * allocated unflushed CDC segments and the contents of cdc_raw
+     *
+     * Synchronized on this
```
jmckenzie-dev, line 232: This is only called in one place; recommend we forklift this code out and above to where it's called so it's clear the lock is held while these operations take place.

src/java/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDC.java
```diff
@@ -386,9 +388,6 @@ private void recalculateOverflowSize()
         {
             rateLimiter.acquire();
             calculateSize();
```
jmckenzie-dev: With this change we should probably go ahead and lift up the calculateSize() code body into the `submitOverflowSizeRecalculation` since it's only called in one place. Saves future maintainers the jump to trace operations.

test/unit/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDCTest.java
```diff
@@ -91,7 +91,7 @@ public void testCDCWriteFailure() throws Throwable
                 FileUtils.deleteWithConfirm(f);
 
             // Update size tracker to reflect deleted files. Should flip flag on current allocatingFrom to allow.
-            cdcMgr.updateCDCTotalSize();
+            long x = cdcMgr.updateCDCTotalSize();
```
jmckenzie-dev: unused variable declaration

jmckenzie-dev (changes requested):

src/java/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDC.java
yifan-c: The `CDCSizeTracker` sums up the size of all commit logs, including CDC and non-CDC files, since it creates hard links for all newly created files. (A side question is.. maybe we want to created the link unless the state is `FORBIDDEN`?). So the size is often an over-estimate when the workload contains both CDC and non-CDC traffic. 
Back to the scenario in `allocate`... if the state was previously `FORBIDDEN`, it means the total size has exceeded the limit at the time of creation. Now, the total size drops below the limit. The `CDCSizeTracker` must have run at least once. In this case, the size of the file (`allocatingFrom`) is already counted.

src/java/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDC.java
yifan-c: The method is an abstract one defined in the base class though. Alternatively, we can add `synchronized` to the method signature. Synchronize within a synchronized block should add trivial impact.

src/java/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDC.java
jmckenzie-dev: `The CDCSizeTracker must have run at least once. In this case, the size of the file (allocatingFrom) is already counted.` <- I didn't connect that. You're right.

`maybe we want to created the link unless the state is FORBIDDEN?`
🤔 We shouldn't hard link the file if the cdcState in there is FORBIDDEN should we?
```
// Hard link file in cdc folder for realtime tracking
FileUtils.createHardLink(segment.logFile, segment.getCDCFile());
```

src/java/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDC.java
jmckenzie-dev: Ah, you're right. Let's add a `@Override` then and leave it as is. 😀

src/java/org/apache/cassandra/db/commitlog/CommitLogSegmentManagerCDC.java
yifan-c: No. If the the state is FORBIDDEN, we do not create a link. The link is created only when the state is NOT FORBIDDEN (i.e. PERMITTED/CONTAINS). 
The current implementation is that a hardlink is created for all new segments, regardless of the state. Since each FORBIDDEN segment has a change to revert back to PERMITTED, this handling is easy but at the cost of slightly over-counting (the size of one file). 
I think, and as you mentioned, it likely not a big problem.

jmckenzie-dev: I'm confused. Here's what I'm seeing in the code:
```
    public CommitLogSegment createSegment()
    {
        CommitLogSegment segment = CommitLogSegment.createSegment(commitLog, this);

        // Hard link file in cdc folder for realtime tracking
        FileUtils.createHardLink(segment.logFile, segment.getCDCFile());

        cdcSizeTracker.processNewSegment(segment);
        return segment;
    }
```
I think the code would need to read something like this to prevent hard-linking files with cdc disabled:
```
    public CommitLogSegment createSegment()
    {
        CommitLogSegment segment = CommitLogSegment.createSegment(commitLog, this);

        if (segment.getCDCState() != CDCState.FORBIDDEN)
        {
            // Hard link file in cdc folder for realtime tracking
            FileUtils.createHardLink(segment.logFile, segment.getCDCFile());
        }

        cdcSizeTracker.processNewSegment(segment);
        return segment;
    }
```

Further, we'd need to add logic to hard link the file on transition from FORBIDDEN to PERMITTED if that happened later in its lifecycle.

That make sense?

yifan-c: It makes sense. The if statement is what I proposed in the earlier comment. The ease of handling in the current implementation is what I guessed by reading the code.

jmckenzie-dev (approved):
👍 after the changes.

PR #1773, opened 2022-08-05
CASSANDRA-17805: Check that the replacing node is alive during host r…

frankgh:
…eplacement

Add a new check during host replacement. Currently, during a node replacement, we check that the node
has not updated gossip for a configured ring_delay amount of time (defaults to 30 seconds). In CASSANDRA-17776,
the delay is calculated from the max value between the `BROADCAST_INTERVAL` and 2X the configured `ring_delay`.

If we see an update from the node that we are replacing in less than the calculated `sleep delay`, we throw a
`UnsupportedOperationException` with message `Cannot replace a live node....`. However, we never check whether
the node is reporting as alive or not alive. In this commit, we add the check to ensure that the node is in
fact reporting as alive before throwing the exception. Additionally, we add logging information with values for
the token, `updateTimestamp`, and `allowedDelay` values for better reporting.

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -1868,7 +1868,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
                         // if the node was updated within the ring delay or the node is alive, we should fail
                         if (updateTimestamp > allowedDelay || endpointStateForExisting.isAlive())
                         {
-                            logger.error("Unable to replace node for token={}. The node is reporting as {}alive with updateTimestamp={} which exceeds the allowedDelay={}",
+                            logger.error("Unable to replace node for token={}. The node is reporting as {}alive with updateTimestamp={}, allowedDelay={}",
                                          token, endpointStateForExisting.isAlive() ? "" : "not ", updateTimestamp, allowedDelay);
                             throw new UnsupportedOperationException("Cannot replace a live node... ");
                         }
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -1864,7 +1864,9 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
                         EndpointState endpointStateForExisting = Gossiper.instance.getEndpointStateForEndpoint(existing);
                         long updateTimestamp = endpointStateForExisting.getUpdateTimestamp();
                         long allowedDelay = nanoTime() - nanoDelay;
-                        if (updateTimestamp > allowedDelay && endpointStateForExisting.isAlive())
+
+                        // if the node was updated within the ring delay or the node is alive, we should fail
+                        if (updateTimestamp > allowedDelay || endpointStateForExisting.isAlive())
                         {
                             logger.error("Unable to replace node for token={}. The node is reporting as alive with updateTimestamp={} which exceeds the allowedDelay={}",
                                          token, updateTimestamp, allowedDelay);
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -1868,8 +1868,8 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
                         // if the node was updated within the ring delay or the node is alive, we should fail
                         if (updateTimestamp > allowedDelay || endpointStateForExisting.isAlive())
                         {
-                            logger.error("Unable to replace node for token={}. The node is reporting as alive with updateTimestamp={} which exceeds the allowedDelay={}",
-                                         token, updateTimestamp, allowedDelay);
+                            logger.error("Unable to replace node for token={}. The node is reporting as {}alive with updateTimestamp={} which exceeds the allowedDelay={}",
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1855,14 +1855,21 @@ public Collection<InetAddressAndPort> prepareForBootstrap(long schemaTimeoutMill
                 }
 
                 // check for operator errors...
+                long nanoDelay = MILLISECONDS.toNanos(ringTimeoutMillis);
```
frankgh, line 1858: no need to recalculate this value in every iteration of the loop. So moved outside of the for loop

frankgh: also using `MILLISECONDS.toNanos` here to avoid conversion errors

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1855,14 +1855,21 @@ public Collection<InetAddressAndPort> prepareForBootstrap(long schemaTimeoutMill
...
                 for (Token token : bootstrapTokens)
                 {
                     InetAddressAndPort existing = tokenMetadata.getEndpoint(token);
                     if (existing != null)
                     {
-                        long nanoDelay = ringTimeoutMillis * 1000000L;
-                        if (Gossiper.instance.getEndpointStateForEndpoint(existing).getUpdateTimestamp() > (nanoTime() - nanoDelay))
+                        EndpointState endpointStateForExisting = Gossiper.instance.getEndpointStateForEndpoint(existing);
+                        long updateTimestamp = endpointStateForExisting.getUpdateTimestamp();
+                        long allowedDelay = nanoTime() - nanoDelay;
```
dcapwell, line 1866: don't need to save to a local variable, can keep in the if statement

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1855,14 +1855,21 @@ public Collection<InetAddressAndPort> prepareForBootstrap(long schemaTimeoutMill
...
+                        if (updateTimestamp > allowedDelay && endpointStateForExisting.isAlive())
```
dcapwell: should be `||`. If it was updated within the last ring delay or we think its alive, then fail

dcapwell (changes requested):

src/java/org/apache/cassandra/service/StorageService.java
frankgh: we use it in the if statement as well as the log

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1855,14 +1855,23 @@ public Collection<InetAddressAndPort> prepareForBootstrap(long schemaTimeoutMill
...
-                        long nanoDelay = ringTimeoutMillis * 1000000L;
-                        if (Gossiper.instance.getEndpointStateForEndpoint(existing).getUpdateTimestamp() > (nanoTime() - nanoDelay))
+                        EndpointState endpointStateForExisting = Gossiper.instance.getEndpointStateForEndpoint(existing);
+                        long updateTimestamp = endpointStateForExisting.getUpdateTimestamp();
+                        long allowedDelay = nanoTime() - nanoDelay;
+
+                        // if the node was updated within the ring delay or the node is alive, we should fail
+                        if (updateTimestamp > allowedDelay || endpointStateForExisting.isAlive())
+                        {
+                            logger.error("Unable to replace node for token={}. The node is reporting as alive with updateTimestamp={} which exceeds the allowedDelay={}",
```
dcapwell: > The node is reporting as alive

This log can be confusing in the case that the endpoint is `isAlive` but the update <= allowedDelay, can you rework this to handle both cases?

frankgh: good catch, I have updated the log statement

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1855,14 +1855,23 @@ public Collection<InetAddressAndPort> prepareForBootstrap(long schemaTimeoutMill
...
-                        long nanoDelay = ringTimeoutMillis * 1000000L;
-                        if (Gossiper.instance.getEndpointStateForEndpoint(existing).getUpdateTimestamp() > (nanoTime() - nanoDelay))
+                        EndpointState endpointStateForExisting = Gossiper.instance.getEndpointStateForEndpoint(existing);
+                        long updateTimestamp = endpointStateForExisting.getUpdateTimestamp();
+                        long allowedDelay = nanoTime() - nanoDelay;
+
+                        // if the node was updated within the ring delay or the node is alive, we should fail
+                        if (updateTimestamp > allowedDelay || endpointStateForExisting.isAlive())
+                        {
+                            logger.error("Unable to replace node for token={}. The node is reporting as {}alive with updateTimestamp={} which exceeds the allowedDelay={}",
```
dcapwell: > which exceeds the allowedDelay={}

again, this may not be true. you can keep it simple and just log the values

dcapwell: spoke in slack `Unable to replace node for token={}. The node is reporting as {}alive with updateTimestamp={}, allowedDelay={}"` works for me!

dcapwell (approved):

PR #1774, opened 2022-08-08
CASSANDRA-16471: DataOutputBuffer#scratchBuffer is around 50% of all memory allocations

dcapwell:


src/java/org/apache/cassandra/io/util/DataOutputBuffer.java
```diff
@@ -221,6 +240,17 @@ public ByteBuffer buffer(boolean duplicate)
         return result;
     }
 
+    /**
+     * Gets the underlying ByteBuffer and calls {@link ByteBuffer#flip()}.  This method is "unsafe" in the sense that
+     * it returns the underlying buffer, which may be modified by other methods after calling this method.  If the calling
+     * logic knows that no new calls to this object will happen after calling this method, then this method can avoid the
+     * copying done in {@link #getData()} and {@link #asNewBuffer()}.
+     */
+    public ByteBuffer unsafeGetBufferAndFlip() {
```
maedhroz: super-nit: newline

src/java/org/apache/cassandra/io/util/DataOutputBuffer.java
```diff
@@ -221,6 +240,17 @@ public ByteBuffer buffer(boolean duplicate)
         return result;
     }
 
+    /**
+     * Gets the underlying ByteBuffer and calls {@link ByteBuffer#flip()}.  This method is "unsafe" in the sense that
+     * it returns the underlying buffer, which may be modified by other methods after calling this method.  If the calling
+     * logic knows that no new calls to this object will happen after calling this method, then this method can avoid the
+     * copying done in {@link #getData()} and {@link #asNewBuffer()}.
```
maedhroz: nit: I guess the `asNewBuffer()` part is no longer relevant, as we're sticking with that in `truncationAsMapEntry()`?

maedhroz: I think the original intent was also to avoid `buffer(true)`, which we no longer used outside `HintsWriter`.

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -291,7 +291,7 @@ public CommitLogPosition add(Mutation mutation) throws CDCWriteException
                 buffer.putInt((int) checksum.getValue());
 
                 // checksummed mutation
-                dos.write(dob.getData(), 0, size);
+                dos.write(dob.unsafeGetBufferAndFlip());
```
maedhroz, line 294: I'd revert this and the similar thing in `UnfilteredSerializer`.

src/java/org/apache/cassandra/io/util/DataOutputBuffer.java
dcapwell: `asNewBuffer` is still used in Describe and SystemKeyspace

`buffer` I feel is newer than when I first wrote it... there are duplicate methods in this class huh !

dcapwell: added buffer to the doc

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
dcapwell: why?

dcapwell: CommitLog does not hold ref after, `.write` copies the value into its local buffer.  

```
public byte[] getData()
    {
        assert buffer.arrayOffset() == 0;
        return buffer.array();
    }
```

this method is no longer safe as it may be direct now...

maedhroz: Makes sense. I was only thinking in terms of whether we copied or not.

maedhroz (approved):

PR #1781, opened 2022-08-11
CASSANDRA-17757

smiklosovic:
Testing it if links to CASSANDRA-17757

### DIFF
--- a/src/java/org/apache/cassandra/db/Keyspace.java
+++ b/src/java/org/apache/cassandra/db/Keyspace.java
@@ -303,20 +303,6 @@ public class Keyspace
     }
 
     /**
-     * Clear all the snapshots for a given keyspace.
-     *
-     * @param snapshotName the user supplied snapshot name. It empty or null,
-     *                     all the snapshots will be cleaned
-     */
-    public static void clearSnapshot(String snapshotName, String keyspace)
-    {
-        RateLimiter clearSnapshotRateLimiter = DatabaseDescriptor.getSnapshotRateLimiter();
-
-        List<File> tableDirectories = Directories.getKSChildDirectories(keyspace);
-        Directories.clearSnapshot(snapshotName, tableDirectories, clearSnapshotRateLimiter);
-    }
-
-    /**
      * @return A list of open SSTableReaders
      */
     public List<SSTableReader> getAllSSTables(SSTableSet sstableSet)
--- a/src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
+++ b/src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
@@ -54,7 +54,8 @@ public class SnapshotLoader extends SimpleFileVisitor<Path>
 {
     private static final Logger logger = LoggerFactory.getLogger(SnapshotLoader.class);
 
-    static final Pattern SNAPSHOT_DIR_PATTERN = Pattern.compile("(?<keyspace>\\w+)/(?<tableName>\\w+)\\-(?<tableId>[0-9a-f]{32})/snapshots/(?<tag>[\\w-]+)$");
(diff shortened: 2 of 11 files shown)


src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
```diff
@@ -79,15 +79,19 @@ public SnapshotLoader(Directories directories)
         this(directories.getCFDirectories().stream().map(File::toPath).collect(Collectors.toList()));
     }
 
-    public Set<TableSnapshot> loadSnapshots()
+    public Set<TableSnapshot> loadSnapshots(String keyspace)
     {
+        int maxDepth = keyspace == null ? 5 : 4;
```
pauloricardomg, line 87: Can you elaborate why this is needed?

src/java/org/apache/cassandra/db/Keyspace.java
```diff
@@ -299,15 +301,34 @@ public boolean snapshotExists(String snapshotName)
...
+                                                                  })
+                                                                  .collect(Collectors.toSet());
+
         RateLimiter clearSnapshotRateLimiter = DatabaseDescriptor.getSnapshotRateLimiter();
 
-        List<File> tableDirectories = Directories.getKSChildDirectories(keyspace);
-        Directories.clearSnapshot(snapshotName, tableDirectories, clearSnapshotRateLimiter);
+        for (TableSnapshot snapshot : snapshotsToClear)
+            for (File snapshotDirectory : snapshot.getDirectories())
+                Directories.removeSnapshotDirectory(clearSnapshotRateLimiter, snapshotDirectory);
```
pauloricardomg: It seems like this logic is duplicated on `SnapshotManager.clearSnapshot`. We also need to ensure expiring snapshots are properly removed from `SnapshotManager`. I think this logic is in `Keyspace` for historical reasons, now we can move it to `StorageService` so we can properly centralize snapshot clearing via `SnapshotManager`. I created [this PR](https://github.com/instaclustr/cassandra/pull/47) moving `Keyspace.clearSnapshot` to `StorageService.clearKeyspaceSnapshot`. Let me know what do you think.

src/java/org/apache/cassandra/db/Keyspace.java
```diff
@@ -299,15 +301,34 @@ public boolean snapshotExists(String snapshotName)
...
+     * @param snapshotName the user supplied snapshot name. If empty or null,
      *                     all the snapshots will be cleaned
+     * @param keyspace keyspace to remove snapshots for
      */
     public static void clearSnapshot(String snapshotName, String keyspace)
     {
+        Set<TableSnapshot> snapshotsToClear = new SnapshotLoader().loadSnapshots(keyspace)
+                                                                  .stream()
+                                                                  .filter(ts -> {
+                                                                      if (snapshotName == null || snapshotName.isEmpty())
```
pauloricardomg: I simplified this filtering a bit [here](https://github.com/instaclustr/cassandra/pull/47/files#diff-9bf2c26bc294ef9085e16bf287490223665eaa2eb8ec24bcf5bd8653c713644bR4213) by encapsulating the filtering logic on [TableSnapshot.shouldClearSnapshot](https://github.com/instaclustr/cassandra/pull/47/files#diff-7d6d1bafcad95c5715c91c9065a4a8c58c3d5c98d0699d9c913717f5c0086bb7R328). Let me know what do you think.

src/java/org/apache/cassandra/service/snapshot/SnapshotLoader.java
smiklosovic: I see that you had `5` in maxDepth on walking. If we supply `keyspace`, we basically move one level "down" so is not it true that your max depth of `5` suddenly becomes `4`?

src/java/org/apache/cassandra/db/Keyspace.java
```diff
@@ -302,20 +302,6 @@ public boolean snapshotExists(String snapshotName)
         return false;
     }
 
-    /**
-     * Clear all the snapshots for a given keyspace.
```
smiklosovic, line 306: test this comment links to CASSANDRA-17757

PR #1783, opened 2022-08-11
Cassandra 17812 - Rate-limit new client connection auth setup to avoid overwhelming bcrypt

jmckenzie-dev:


src/java/org/apache/cassandra/transport/Dispatcher.java
```diff
@@ -237,9 +267,12 @@ public static void shutdown()
         {
             requestExecutor.shutdown();
         }
+        if (authExecutor != null)
+        {
+            authExecutor.shutdown();
+        }
```
maedhroz, line 273: nit: Is it even possible for `requestExecutor` or `authExecutor` to be `null`? (They're both `static final`.)

src/java/org/apache/cassandra/transport/Dispatcher.java
```diff
@@ -80,7 +101,16 @@ public Dispatcher(boolean useLegacyFlusher)
...
-        requestExecutor.submit(new RequestProcessor(channel, request, forFlusher, backpressure));
+        // if native_transport_max_auth_threads is < 1, don't delegate to new pool on auth messages
+        if (DatabaseDescriptor.getNativeTransportMaxAuthThreads() > 0 && (request.type == Message.Type.AUTH_RESPONSE || request.type == Message.Type.CREDENTIALS))
+        {
+            // importantly will handle the AUTHENTICATE message which may be CPU intensive.
+            authExecutor.submit(new RequestProcessor(channel, request, forFlusher, backpressure));
+        }
+        else
+        {
+            requestExecutor.submit(new RequestProcessor(channel, request, forFlusher, backpressure));
```
maedhroz, line 112: nit: We could create the request processor before the `if` and avoid duplicating that code?

maedhroz (approved):
Just a few nits here and there.
LGTM

src/java/org/apache/cassandra/transport/Dispatcher.java
jmckenzie-dev: I don't believe it is. Tidied up both new and old executor. 👍

src/java/org/apache/cassandra/transport/Dispatcher.java
jmckenzie-dev: I think I prefer taking it a step further and just restructuring it entirely:
```
    public void dispatch(Channel channel, Message.Request request, FlushItemConverter forFlusher, Overload backpressure)
    {
        // if native_transport_max_auth_threads is < 1, don't delegate to new pool on auth messages
        boolean isAuthQuery = DatabaseDescriptor.getNativeTransportMaxAuthThreads() > 0 &&
                              (request.type == Message.Type.AUTH_RESPONSE || request.type == Message.Type.CREDENTIALS);

        // Importantly, the authExecutor will handle the AUTHENTICATE message which may be CPU intensive.
        LocalAwareExecutorPlus executor = isAuthQuery ? authExecutor : requestExecutor;

        executor.submit(new RequestProcessor(channel, request, forFlusher, backpressure));
        ClientMetrics.instance.markRequestDispatched();
    }
```

smiklosovic: +1 to that idea. I had the same idea as Josh to resolve the executor to submit that to.

PR #1788, opened 2022-08-14
CASSANDRA-17668 handle leak of non-standard Java types as clients using JMX cannot handle them

lmtrombone:
DatabaseDescriptor had changes where some methods threw non-standard exceptions.  This caused some of the JMX setter methods that called them to throw them as well which causes issues as the clients can't handle them.  This pull request addresses these issues by creating new APIs that catches and converts the non-standard exceptions to standard ones and deprecates the old setters.  For consistency, since the setters have been deprecated, new getters have been created as well.

### DIFF
--- a/src/java/org/apache/cassandra/service/ActiveRepairService.java
+++ b/src/java/org/apache/cassandra/service/ActiveRepairService.java
@@ -46,6 +46,7 @@ import org.apache.cassandra.concurrent.ExecutorPlus;
 import org.apache.cassandra.config.Config;
 import org.apache.cassandra.config.DurationSpec;
 import org.apache.cassandra.db.compaction.CompactionManager;
+import org.apache.cassandra.exceptions.ConfigurationException;
 import org.apache.cassandra.locator.AbstractReplicationStrategy;
 import org.apache.cassandra.locator.EndpointsByRange;
 import org.apache.cassandra.locator.EndpointsForRange;
@@ -297,14 +298,40 @@ public class ActiveRepairService implements IEndpointStateChangeSubscriber, IFai
         return DatabaseDescriptor.getRepairSessionSpaceInMiB();
     }
 
+    @Deprecated
     @Override
     public void setRepairSessionSpaceInMebibytes(int sizeInMebibytes)
     {
         DatabaseDescriptor.setRepairSessionSpaceInMiB(sizeInMebibytes);
     }
 
+    @Deprecated
     @Override
     public int getRepairSessionSpaceInMebibytes()
+    {
+        return DatabaseDescriptor.getRepairSessionSpaceInMiB();
+    }
+
+    @Override
+    public void setRepairSessionSpaceInMiB(int sizeInMebibytes)
+    {
+        try
+        {
+            DatabaseDescriptor.setRepairSessionSpaceInMiB(sizeInMebibytes);
+        }
+        catch (ConfigurationException e)
+        {
+            throw new IllegalArgumentException(e.getMessage());
+        }
+    }
+
+    /*
+     * In CASSANDRA-17668, JMX setters that did not throw standard exceptions were deprecated in favor of ones that do.
+     * For consistency purposes, the respective getter "getRepairSessionSpaceInMebibytes" was also deprecated and
+     * replaced by this method.
+     */
+    @Override
+    public int getRepairSessionSpaceInMiB()
     {
         return DatabaseDescriptor.getRepairSessionSpaceInMiB();
     }
--- a/src/java/org/apache/cassandra/service/ActiveRepairServiceMBean.java
+++ b/src/java/org/apache/cassandra/service/ActiveRepairServiceMBean.java
@@ -34,8 +34,19 @@ public interface ActiveRepairServiceMBean
     @Deprecated
     public int getRepairSessionSpaceInMegabytes();
 
+    /**
+     * @deprecated use setRepairSessionSpaceInMiB instead as it will not throw non-standard exceptions
+     */
+    @Deprecated
     public void setRepairSessionSpaceInMebibytes(int sizeInMebibytes);
+    /**
+     * @deprecated use getRepairSessionSpaceInMiB instead
+     */
+    @Deprecated
     public int getRepairSessionSpaceInMebibytes();
+
+    public void setRepairSessionSpaceInMiB(int sizeInMebibytes);
+    public int getRepairSessionSpaceInMiB();
 
     public boolean getUseOffheapMerkleTrees();
     public void setUseOffheapMerkleTrees(boolean value);
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -4554,12 +4554,40 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
                ImmutableList.<String>builder().add(pair.left.name()).addAll(pair.right).build();
     }
 
+    @Deprecated
+    @Override
     public void setRepairSessionMaxTreeDepth(int depth)
     {
         DatabaseDescriptor.setRepairSessionMaxTreeDepth(depth);
     }
 
+    @Deprecated
+    @Override
     public int getRepairSessionMaxTreeDepth()
+    {
+        return DatabaseDescriptor.getRepairSessionMaxTreeDepth();
+    }
+
+    @Override
+    public void setRepairSessionMaximumTreeDepth(int depth)
+    {
+        try
+        {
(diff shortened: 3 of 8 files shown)


src/java/org/apache/cassandra/service/ActiveRepairServiceMBean.java
```diff
@@ -34,9 +34,14 @@
     @Deprecated
     public int getRepairSessionSpaceInMegabytes();
 
+    /**
+     * @deprecated use setRepairSessionSpaceInMiB instead as it will not throw non-standard exceptions
+     */
```
ekaterinadimitrova2, line 39: We need to add @Deprecated here too, not only in Java doc and the actual implementation

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4525,6 +4525,7 @@ public List<String> getParentRepairStatus(int cmd)
                ImmutableList.<String>builder().add(pair.left.name()).addAll(pair.right).build();
     }
 
+    @Deprecated
```
ekaterinadimitrova2, line 4557: I think @Override was missed before and we can use the opportunity to add it maybe?

src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -835,13 +838,23 @@ default int upgradeSSTables(String keyspaceName, boolean excludeCurrentVersion,
 
     /** Returns the granularity of the collation index of rows within a partition **/
     public int getColumnIndexSizeInKiB();
-    /** Sets the granularity of the collation index of rows within a partition **/
+    /**
+     * @deprecated use setColumnIndexSizeInKiB instead as it will not throw non-standard exceptions
+     * Sets the granularity of the collation index of rows within a partition
+     */
     public void setColumnIndexSize(int columnIndexSizeInKB);
```
ekaterinadimitrova2, line 889: Same as before, let's add also @Deprecated

src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -835,13 +838,23 @@ default int upgradeSSTables(String keyspaceName, boolean excludeCurrentVersion,
...
+    /** Sets the granularity of the collation index of rows within a partition **/
+    public void setColumnIndexSizeInKiB(int columnIndexSizeInKiB);
 
     /** Returns the threshold for skipping the column index when caching partition info **/
     public int getColumnIndexCacheSize();
-    /** Sets the threshold for skipping the column index when caching partition info **/
+    /**
+     * @deprecated use setColumnIndexCacheSizeInKiB instead as it will not throw non-standard exceptions
+     * Sets the threshold for skipping the column index when caching partition info
+     */
```
ekaterinadimitrova2, line 901: @Deprecated to be added

src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -850,8 +863,13 @@ default int upgradeSSTables(String keyspaceName, boolean excludeCurrentVersion,
 
     /** Returns the threshold for warning queries due to a large batch size */
     public int getBatchSizeWarnThreshold();
-    /** Sets the threshold for warning queries due to a large batch size */
+    /**
+     * @deprecated use setBatchSizeWarnThresholdInKiB instead as it will not throw non-standard exceptions
+     * Sets the threshold for warning queries due to a large batch size
+     */
```
ekaterinadimitrova2, line 925: Same @Deprecated to be added

ekaterinadimitrova2 (commented):
Overall looks good on a quick skim, I left some tiny suggestions. Just to confirm, did you have the chance to verify whether there are other setters/getters in the DD that can suffer from this problem, out of the list/examples suggested on the ticket? 
Also, I would like to suggest to add some basic unit tests to DatabaseDescriptorTest class? 
Also, I think we need patches for previous versions and we normally merge them from lower affected version upwards to trunk - for more info https://cassandra.apache.org/_/development/patches.html (Bug Fixes section). There might be also the case we need different patches for the different versions as things diverged between them. I can run a full CI for you when we get things into shape.
One more thing is to check whether our docs need some relevant update, at minimum we need to update the Deprecation section in NEWS.txt. Thank you for your work!

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4525,16 +4525,39 @@ public List<String> getParentRepairStatus(int cmd)
...
+            DatabaseDescriptor.setRepairSessionMaxTreeDepth(depth);
+        }
+        catch (ConfigurationException e)
+        {
+            throw new IllegalArgumentException(e.getMessage());
+        }
+    }
+
+    @Override
+    public int getRepairSessionMaximumTreeDepth()
```
dcapwell, line 4590: can you add a comment explaining why this exists?  Fo someone reviewing the code its going to be asked "why is getRepairSessionMaxTreeDepth deprecated... its the same logic here...", so if we are going to depreciate the `get` we should explain why in the code

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -6146,29 +6169,71 @@ public void setCachedReplicaRowsFailThreshold(int threshold)
         logger.info("updated replica_filtering_protection.cached_rows_fail_threshold to {}", threshold);
     }
 
+    @Override
```
dcapwell, line 6250: same comment as above, can we document "why" the new method that has identical logic?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -6146,29 +6169,71 @@ public void setCachedReplicaRowsFailThreshold(int threshold)
...
 
+    @Deprecated
+    @Override
     public void setColumnIndexCacheSize(int cacheSizeInKB)
     {
         DatabaseDescriptor.setColumnIndexCacheSize(cacheSizeInKB);
         logger.info("Updated column_index_cache_size to {}", cacheSizeInKB);
     }
 
+    @Override
```
dcapwell, line 6300: same comment as above, can we document "why" the new method that has identical logic?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -6180,17 +6245,42 @@ public void setBatchSizeFailureThreshold(int threshold)
...
+    @Deprecated
+    @Override
     public void setBatchSizeWarnThreshold(int threshold)
     {
         DatabaseDescriptor.setBatchSizeWarnThresholdInKiB(threshold);
         logger.info("Updated batch_size_warn_threshold to {}", threshold);
     }
 
+    @Override
+    public int getBatchSizeWarnThresholdInKiB()
```
dcapwell, line 6352: same comment as above, can we document "why" the new method that has identical logic?

src/java/org/apache/cassandra/service/StorageService.java
ekaterinadimitrova2: Good point, we can also add the ticket number for clarity

src/java/org/apache/cassandra/service/StorageService.java
lmtrombone: this is actually an existing method that I added `@Override` to for consistency purposes / best practices.  If you still think it makes sense to add a comment here, let me know.

src/java/org/apache/cassandra/service/ActiveRepairService.java
```diff
@@ -297,18 +298,44 @@ public int getRepairSessionSpaceInMegabytes()
...
+    @Override
+    public void setRepairSessionSpaceInMiB(int sizeInMebibytes)
+    {
+        try
+        {
+            DatabaseDescriptor.setRepairSessionSpaceInMiB(sizeInMebibytes);
+        }
+        catch (ConfigurationException e)
+        {
+            throw new IllegalArgumentException(e.getMessage());
```
ekiauhce, line 324: It's better to use `IllegalArgumentException(Throwable cause)` constructor, because it will preserve `e` stacktrace.

ekaterinadimitrova2: This would be a precedent here I think, I can't imagine putting the Stacktrace as an error message in JConsole, @dcapwell ?

lmtrombone: Unless I'm missing something, I don't think this will be applicable here.  I tried it and JConsole did not respond well to it and had trouble properly displaying the stacktrace as is.

PR #1796, opened 2022-08-17
CASSANDRA-17828: RequestFailure thrown when RequestTimeout should have thrown

dcapwell:


src/java/org/apache/cassandra/service/reads/ReadCallback.java
```diff
@@ -120,6 +120,18 @@ public void awaitResults() throws ReadFailureException, ReadTimeoutException
...
+        // Need to distigues between a timeout and a failure (network, bad data, etc.), so store an extra field timeout
+        // to say that most failures were actually a timeout
+        // see CASSANDRA-17828
+        boolean timedout = false;
+        if (failed)
+        {
+            // is it actually a timeout?
+            long numTimeout = failureReasonByEndpoint.values().stream().filter(RequestFailureReason.TIMEOUT::equals).count();
+            long nonTimeout = failures - numTimeout;
+            timedout = nonTimeout > numTimeout;
```
dcapwell: should write use the same logic?  it would be good to be consistent I guess...

src/java/org/apache/cassandra/service/TruncateResponseHandler.java
```diff
@@ -61,24 +62,36 @@ public TruncateResponseHandler(int responseCount)
...
+            {
+                throw new TimeoutException("Truncate timed out - received only " + responses.get() + " responses");
+            }
+            else
+            {
+                StringBuilder sb = new StringBuilder("Truncate failed on replicas ");
+                for (Map.Entry<InetAddressAndPort, RequestFailureReason> e : failureReasonByEndpoint.entrySet())
+                    sb.append(e.getKey()).append(' ').append(e.getValue()).append(", ");
+                sb.setLength(sb.length() - 2);
+                throw new TruncateException(sb.toString());
```
dcapwell: this change is causing test failures... switching from `replica` to `replicas` and now including the reason... wondering if I should keep the old logic; just log 1

test/distributed/org/apache/cassandra/distributed/test/metrics/RequestTimeoutTest.java
```diff
@@ -0,0 +1,191 @@
...
+import org.apache.cassandra.utils.concurrent.Condition;
+import org.assertj.core.api.Assertions;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+import static net.bytebuddy.matcher.ElementMatchers.takesArguments;
+import static org.apache.cassandra.utils.AssertionUtils.isThrowable;
+
+public class RequestTimeoutTest extends TestBaseImpl
+{
+    private static AtomicInteger NEXT = new AtomicInteger(0);
```
maedhroz: nit: could be `final`

src/java/org/apache/cassandra/service/TruncateResponseHandler.java
```diff
@@ -61,24 +63,34 @@ public TruncateResponseHandler(int responseCount)
...
+            Map<InetAddressAndPort, RequestFailureReason> failureReasonByEndpoint = new HashMap<>(this.failureReasonByEndpoint);
+            int size = failureReasonByEndpoint.size();
+            long timeouts = failureReasonByEndpoint.values().stream().filter(RequestFailureReason.TIMEOUT::equals).count();
+            long nonTimeout = size - timeouts;
+            if (nonTimeout <= timeouts)
+                throw new TimeoutException("Truncate timed out - received only " + responses.get() + " responses");
+
+            StringBuilder sb = new StringBuilder("Truncate failed on ");
+            for (Map.Entry<InetAddressAndPort, RequestFailureReason> e : failureReasonByEndpoint.entrySet())
+                sb.append("replica ").append(e.getKey()).append(' ').append(e.getValue()).append(", ");
```
maedhroz: nit: maybe a color after the address would be helpful?
```suggestion
                sb.append("replica ").append(e.getKey()).append(': ').append(e.getValue()).append(", ");
```

src/java/org/apache/cassandra/service/TruncateResponseHandler.java
```diff
@@ -61,24 +63,34 @@ public TruncateResponseHandler(int responseCount)
...
+        if (!failureReasonByEndpoint.isEmpty())
         {
-            throw new TruncateException("Truncate failed on replica " + truncateFailingReplica);
+            // clone to make sure no race condition happens
+            Map<InetAddressAndPort, RequestFailureReason> failureReasonByEndpoint = new HashMap<>(this.failureReasonByEndpoint);
+            int size = failureReasonByEndpoint.size();
+            long timeouts = failureReasonByEndpoint.values().stream().filter(RequestFailureReason.TIMEOUT::equals).count();
+            long nonTimeout = size - timeouts;
+            if (nonTimeout <= timeouts)
+                throw new TimeoutException("Truncate timed out - received only " + responses.get() + " responses");
```
maedhroz, line 84: What if instead of throwing a `TimeoutException` when there are more timeouts than non-timeouts (okay >=), we just get rid of what's between lines 83 and 88 here? (i.e. We just throw a `TruncateException` with the per-endpoint failure summary snapshot, unless perhaps every single one of the failure reasons is TIMEOUT? Even then, we might want to be able to tell the difference between "all the replicas failed on TIMEOUT" and "we waited too long at the coordinator".)

src/java/org/apache/cassandra/service/reads/ReadCallback.java
```diff
@@ -120,6 +120,17 @@ public void awaitResults() throws ReadFailureException, ReadTimeoutException
          */
         int received = resolver.responses.size();
         boolean failed = failures > 0 && (blockFor > received || !resolver.isDataPresent());
+        // If all messages came back as a TIMEOUT then signaled=true and failed=true.
+        // Need to distigues between a timeout and a failure (network, bad data, etc.), so store an extra field.
```
maedhroz:
```suggestion
        // Need to distinguish between a timeout and a failure (network, bad data, etc.), so store an extra field.
```

test/distributed/org/apache/cassandra/distributed/test/metrics/RequestTimeoutTest.java
```diff
@@ -0,0 +1,191 @@
...
+                           .method(named("await").and(takesArguments(2)))
+                           .intercept(MethodDelegation.to(BB.class))
+                           .make()
+                           .load(cl, ClassLoadingStrategy.Default.INJECTION);
+        }
+
+        public static boolean await(long time, TimeUnit units, @This Awaitable self, @SuperMethod Method method) throws InterruptedException, InvocationTargetException, IllegalAccessException
+        {
+            // attempt to decouple the two usages of write_request_timeout: await(write_request_timeout) and message expire
+            return (boolean) method.invoke(self, time * 10, units);
```
maedhroz: Without this BB injection, the tests seem to pass. Is there a race somewhere that makes this ^ necessary?

test/distributed/org/apache/cassandra/distributed/test/metrics/RequestTimeoutTest.java
```diff
@@ -0,0 +1,191 @@
...
+    {
+        CLUSTER.filters().reset();
+    }
+
+    @Test
+    public void insert()
+    {
+        CLUSTER.filters().verbs(Verb.MUTATION_REQ.id).to(2).drop();
+        Assertions.assertThatThrownBy(() -> CLUSTER.coordinator(1).execute(withKeyspace("INSERT INTO %s.tbl (pk, v) VALUES (?, ?)"), ConsistencyLevel.ALL, NEXT.getAndIncrement(), NEXT.getAndIncrement()))
+                  .is(isThrowable(WriteTimeoutException.class));
```
maedhroz, line 101: A quick search for "Exception.class.getName()" shows a handful of places that could definitely be converted to use this. Perhaps a follow-up Jira?

test/unit/org/apache/cassandra/utils/AssertionUtils.java
```diff
@@ -0,0 +1,124 @@
...
+                return name;
+            }
+        };
+    }
+
+    public static <T extends Throwable> Condition<Throwable> isThrowable(Class<T> klass)
+    {
+        // org.assertj.core.api.AbstractAssert.is has <? super ? extends Throwable> which blocks <T>, so need to
+        // always return Throwable
+        return (Condition<Throwable>) (Condition<?>) is(klass);
```
maedhroz, line 57: nit: I think you can remove the second cast here if you rewrite `is`...

```
public static <T> Condition<T> is(Class<T> klass)
{
    String name = klass.getCanonicalName();
    return new Condition<T>() {
        @Override
        public boolean matches(T value)
        {
            return value.getClass().getCanonicalName().equals(name);
        }

        @Override
        public String toString()
        {
            return name;
        }
    };
}
```

src/java/org/apache/cassandra/service/TruncateResponseHandler.java
dcapwell: `:` may be an issue as that could imply port is missing... but we could do ` -> ` instead of the space?  I also want to make sure there is a space between the replica to avoid any existing log parsers freaking out.

src/java/org/apache/cassandra/service/TruncateResponseHandler.java
dcapwell: the bug reported was when 100% of the replies are timeout and we return the wrong exception type, so I wouldn't want to return a `TrucateException` in this case.

Now, if `signal=true && failed`, I guess we have a more fundmental question of "how many replicas timeout for us to call it a timeout"?  I would prefer we are consistent so would want all read/write/truncate to match the logic...

So, lets say we are dealing with a LOCAL_QUORUM read with rf=6, if failures=3 then we will fail, but lets say we have the following

R1: TIMEOUT
R2: TIMEOUT
R3: UNKNOWN

*should* this be a timeout, or should this be an error?  *should* this only ever be a timeout when 100% are timed out?  Personally I feel that if majority are timeout, we actually timed out.  @clohfink thoughts?

src/java/org/apache/cassandra/service/TruncateResponseHandler.java
maedhroz: `->` WFM

test/distributed/org/apache/cassandra/distributed/test/metrics/RequestTimeoutTest.java
dcapwell: The race is that we use the same timeout for messages and condition wait, so the message expire system needs to happen-before the signal timeout is done to hit the race that this patch deals with.  For me, most of the time signal timeout happens-before message expires, which then means `signal=false`, which means we timeout (and were before).  Now, we need to *FORCE* message expires to happen-before in order to hit the condition this patch is fixing.

dcapwell: I guess, you are calling out that the test may not be working, so we should validate that the expected branch happened... ill have to think through how to detect that... its the same exception from the caller so without changing that we don't have a clear way to know

test/unit/org/apache/cassandra/utils/AssertionUtils.java
dcapwell: I tried that, doesn't work with Assertions builder 😢 

The builder type is `? super ? extends T`, so only the actual value of `T` gets allowed... This means cases like the following don't work

```
Object o = "testing"
Assertions.assertThat(o).is(is(String.class)) // type checking fails, so code doesn't compile.
```

src/java/org/apache/cassandra/service/TruncateResponseHandler.java
maedhroz: If you pointed a gun to my head rn, for all three cases (read/write/truncate), I would just...
1.) throw a timeout if all the failure reasons in the map are TIMEOUT
2.) show the user the failure map by endpoint attached to a real error otherwise

maedhroz (approved):

src/java/org/apache/cassandra/service/TruncateResponseHandler.java
dcapwell: spoke with @clohfink and Caleb in slack and will go with 100% replies are timeout = timeout, this would then just make it so this race condition properly returns timeout and doesn't impact any other case.

test/distributed/org/apache/cassandra/distributed/test/metrics/RequestTimeoutTest.java
dcapwell: ok since now we have a single method to say when isTimeout=true, was able to use that to track times we returned `true` and assert that in the test.

dcapwell: pushed a patch that makes the tests fail if you remove `BB`, it uses reflection to monitor `isTimeout=true` and asserts that we saw at least one (seems reads can trigger multiple).  All but CAS tests fail with it disabled, and they all pass with it disabled.

src/java/org/apache/cassandra/service/TruncateResponseHandler.java
```diff
@@ -61,24 +63,31 @@ public TruncateResponseHandler(int responseCount)
...
-        {
+        if (!signaled)
             throw new TimeoutException("Truncate timed out - received only " + responses.get() + " responses");
-        }
 
-        if (truncateFailingReplica != null)
+        if (!failureReasonByEndpoint.isEmpty())
         {
-            throw new TruncateException("Truncate failed on replica " + truncateFailingReplica);
+            // clone to make sure no race condition happens
```
maedhroz, line 81: No math going on, so we might not strictly need the defensive copy of `failureReasonByEndpoint` now.

src/java/org/apache/cassandra/service/reads/ReadCallback.java
```diff
@@ -120,6 +121,12 @@ public void awaitResults() throws ReadFailureException, ReadTimeoutException
          */
         int received = resolver.responses.size();
         boolean failed = failures > 0 && (blockFor > received || !resolver.isDataPresent());
+        // If all messages came back as a TIMEOUT then signaled=true and failed=true.
+        // Need to distinguish between a timeout and a failure (network, bad data, etc.), so store an extra field.
+        // see CASSANDRA-17828
+        boolean timedout = !signaled;
+        if (failed)
+            timedout = RequestCallback.isTimeout(new HashMap<>(failureReasonByEndpoint));
```
maedhroz, line 129: nit: Defensive copy of `failureReasonByEndpoint` isn't strictly necessary now.

PR #1804, opened 2022-08-22
CASSANDRA-17819: Fix resetting the schema

jacek-lewandowski:


src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -235,6 +235,7 @@
     MEMTABLE_OVERHEAD_SIZE("cassandra.memtable.row_overhead_size", "-1"),
     MEMTABLE_OVERHEAD_COMPUTE_STEPS("cassandra.memtable_row_overhead_computation_step", "100000"),
     MIGRATION_DELAY("cassandra.migration_delay_ms", "60000"),
+    SCHEMA_PULL_INTERVAL("cassandra.schema_pull_interval_ms", "60000"),
```
ekaterinadimitrova2: How about some Javadoc?

src/java/org/apache/cassandra/schema/DefaultSchemaUpdateHandler.java
```diff
@@ -232,12 +240,15 @@ public synchronized SchemaTransformationResult apply(SchemaTransformation transf
...
+        if (!update.diff.isEmpty())
         {
-            migrationCoordinator.announce(update.after.getVersion());
+            this.schema = update.after;
+            logger.debug("Schema updated: {}", update);
+            updateCallback.accept(update, true);
+            if (!local)
+            {
+                migrationCoordinator.announce(update.after.getVersion());
+            }
```
ekaterinadimitrova2, line 264: Shall we add some logging in case no update has happened?

src/java/org/apache/cassandra/schema/DefaultSchemaUpdateHandler.java
```diff
@@ -253,20 +264,28 @@ private synchronized SchemaTransformationResult reload()
...
-    public SchemaTransformationResult reset(boolean local)
+    public void reset(boolean local)
     {
         if (local)
-            return reload();
-
-        Collection<Mutation> mutations = migrationCoordinator.pullSchemaFromAnyNode().awaitThrowUncheckedOnInterrupt().getNow();
-        return applyMutations(mutations);
+        {
+            reload();
```
ekaterinadimitrova2, line 287: Do we still need to return in `reload()`? Seems like this is the only usage

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -323,7 +372,7 @@ private boolean shouldPullFromEndpoint(InetAddressAndPort endpoint)
 
         if (!messagingService.versions.knows(endpoint))
         {
-            logger.debug("Not pulling schema from {} because their messaging version is unknown", endpoint);
+            logger.trace("Not pulling schema from {} because their messaging version is unknown", endpoint);
```
ekaterinadimitrova2: Why this was switched to trace?

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -403,16 +461,61 @@ private synchronized void removeEndpointFromVersion(InetAddressAndPort endpoint,
...
+            }
+        });
+    }
+
     synchronized void removeAndIgnoreEndpoint(InetAddressAndPort endpoint)
     {
+        logger.debug("Removing and ignoring endpoint {}", endpoint);
         Preconditions.checkArgument(endpoint != null);
+        // TODO The endpoint address is now ignored but when a node with the same address is added again later,
+        //  there will be no way to include it in schema synchronization other than restarting each other node
```
ekaterinadimitrova2: I think I mentioned in the previous version but for completeness - I think this TO DO deserves to be documented actually

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -611,18 +613,42 @@ public SchemaTransformationResult transform(SchemaTransformation transformation,
      * Clear all locally stored schema information and fetch schema from another node.
      * Called by user (via JMX) who wants to get rid of schema disagreement.
      */
-    public synchronized void resetLocalSchema()
```
ekaterinadimitrova2, line 614: Why this change?

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -611,18 +613,42 @@ public SchemaTransformationResult transform(SchemaTransformation transformation,
...
+    public void resetLocalSchema()
     {
         logger.debug("Clearing local schema...");
-        updateHandler.clear();
 
-        logger.debug("Clearing local schema keyspace instances...");
-        distributedKeyspaces.forEach(this::unload);
```
ekaterinadimitrova2, line 620: Where did the unload go?

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
```diff
@@ -95,21 +98,37 @@ private void selectSilent(Cluster cluster, String name)
...
             assertTrue(cluster.get(1).callOnInstance(() -> Schema.instance.getTableMetadata(KEYSPACE, "tbl") != null));
             assertTrue(cluster.get(2).callOnInstance(() -> Schema.instance.getTableMetadata(KEYSPACE, "tbl") != null));
 
+            // now we have a table tbl in the schema of both nodes
+
             cluster.get(2).shutdown().get();
 
-            // when schema is removed and there is no other node to fetch it from, node 1 should be left with clean schema
-            cluster.get(1).runOnInstance(() -> Schema.instance.resetLocalSchema());
+            // when there is no node to fetch the schema from, reset local schema should immediately fail
```
ekaterinadimitrova2, line 148: How does this work in previous versions?

src/java/org/apache/cassandra/schema/DefaultSchemaUpdateHandler.java
```diff
@@ -253,20 +264,28 @@ private synchronized SchemaTransformationResult reload()
...
+            {
+                logger.error("Timeout exceeded when waiting for schema from other nodes");
+            }
+        }
     }
 
     @Override
     public synchronized void clear()
     {
+        schema = DistributedSchema.EMPTY;
```
ekaterinadimitrova2: Why did you decide to move it up?

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -369,22 +418,31 @@ private synchronized boolean shouldApplySchemaFor(VersionInfo info)
...
 
         VersionInfo info = versionInfo.computeIfAbsent(version, VersionInfo::new);
         if (Objects.equals(schemaVersion.get(), version))
+        {
             info.markReceived();
+            logger.trace("Schema {} from {} has been marked as recevied because it is equal the local schema", version, endpoint);
+        }
         info.endpoints.add(endpoint);
-        info.requestQueue.addFirst(endpoint);
+        info.requestQueue.addFirst(endpoint); // TODO not sure if it is correct - given we've just marked this schema version as received, why do we add a request to receive it?
```
ekaterinadimitrova2: Have to think a bit about this TO DO

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -149,12 +153,27 @@ private static Set<InetAddressAndPort> getIgnoredEndpoints()
...
+        /**
+         * The set of endpoints containing this schema version
+         */
         final Set<InetAddressAndPort> endpoints           = Sets.newConcurrentHashSet();
+        /**
+         * The set of endpoints from which we are already fetching the schema
+         */
         final Set<InetAddressAndPort> outstandingRequests = Sets.newConcurrentHashSet();
+        /**
+         * The queue of endpoint from which we are going to fetch the schema
```
ekaterinadimitrova2: nit: endpoints

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -180,6 +199,18 @@ boolean wasReceived()
         {
             return receivedSchema;
         }
+
+        @Override
+        public String toString()
```
ekaterinadimitrova2, line 205: Do we actually use this one anywhere?

ekaterinadimitrova2 (commented):
Left some comments, primarily questions actually. Need to look at it more tomorrow. Seems like there are few things going on and I do not want to change any behavior

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
jacek-lewandowski: AFAIU it worked in the way that if we don't get a schema from some other node, we will be left in an inconsistent state, where we have `Keyspace` instances for keyspaces which are missing in the local schema. We would have to restart the node in such case. Also if some keyspaces were modified on the other node in the meantime, say a table was dropped - that table will not be dropped on the node were we called reset because from that node's view point, nothing has been dropped, and the stale data files will be left.

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
jacek-lewandowski: in log messages

ekaterinadimitrova2: Yes, but I didn't see it used, probably I missed it :-)

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -220,18 +251,23 @@ boolean wasReceived()
 
     void start()
     {
+        logger.info("Starting migration coordinator and scheduling pulling schema versions every 1 minute");
```
ekaterinadimitrova2: I think we need to make it print the SCHEMA_PULL_INTERVAL as it is mutable, not always 1 minute necessarily anymore

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -426,49 +529,19 @@ private Future<Void> scheduleSchemaPull(InetAddressAndPort endpoint, VersionInfo
         FutureTask<Void> task = new FutureTask<>(() -> pullSchema(endpoint, new Callback(endpoint, info)));
 
         if (shouldPullImmediately(endpoint, info.version))
+        {
+            logger.debug("Pulling {} immediatelly from {}", info, endpoint);
```
ekaterinadimitrova2: immediately

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -403,16 +461,61 @@ private synchronized void removeEndpointFromVersion(InetAddressAndPort endpoint,
...
+
+        this.endpointVersions.clear();
+
+        Iterator<Map.Entry<UUID, VersionInfo>> it = versionInfo.entrySet().iterator();
+        while (it.hasNext())
+        {
+            Map.Entry<UUID, VersionInfo> entry = it.next();
+            it.remove();
+            entry.getValue().waitQueue.signal();
+        }
```
ekaterinadimitrova2, line 488: This feels like it deserves its own method?

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -403,16 +461,61 @@ private synchronized void removeEndpointFromVersion(InetAddressAndPort endpoint,
...
+                if (state != null)
+                {
+                    UUID v = state.getSchemaVersion();
+                    if (v != null)
+                    {
+                        reportEndpointVersion(endpoint, v);
+                    }
+                }
+            }
+        });
```
ekaterinadimitrova2: This one too
Not that I am super opinionated but it just feels like it can improve readability that way

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
```diff
@@ -95,25 +107,54 @@ private void selectSilent(Cluster cluster, String name)
...
             assertTrue(cluster.get(2).callOnInstance(() -> Schema.instance.getTableMetadata(KEYSPACE, "tbl") != null));
 
+            // now we have a table tbl in the schema of both nodes
+
             cluster.get(2).shutdown().get();
 
-            // when schema is removed and there is no other node to fetch it from, node 1 should be left with clean schema
-            cluster.get(1).runOnInstance(() -> Schema.instance.resetLocalSchema());
+            Awaitility.await().atMost(Duration.ofSeconds(30)) // wait until node 1 notices that node 2 is dead
+                      .until(() -> cluster.get(1).callOnInstance(() -> Gossiper.instance.getLiveMembers().stream().allMatch(e -> e.equals(FBUtilities.getBroadcastAddressAndPort()))));
```
ekaterinadimitrova2: I think this can be split in a few lines as it becomes too long

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
```diff
@@ -95,25 +107,54 @@ private void selectSilent(Cluster cluster, String name)
...
+            cluster.get(1).runOnInstance(() -> {
+                Schema.instance.transform(current -> Schema.instance.distributedKeyspaces().without(KEYSPACE), false);
+             });
+
             assertTrue(cluster.get(1).callOnInstance(() -> Schema.instance.getTableMetadata(KEYSPACE, "tbl") == null));
 
+            // clear will wait until it receives schema from some other node
+            // also if we start the node2 first, schema of node2 will be synced to schema of node1 because node1 has the newest change (dropping the table)
+            // if we run clean on node1 first, it will advertise empty schema to node2 and apply the schema from it without merging
+            CompletableFuture<Boolean> clear1 = CompletableFuture.supplyAsync(() -> cluster.get(1).callOnInstance(() -> Schema.instance.updateHandler.clear().awaitUninterruptibly(1, TimeUnit.MINUTES)));
```
ekaterinadimitrova2: This line can also be split in a few lines so we don't have to scroll right

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
```diff
@@ -95,25 +107,54 @@ private void selectSilent(Cluster cluster, String name)
...
+            // this proves node1 reset schema works - the most recent change should be discarded because it receives
+            // the schema from node2 and applies it on a clean schema
             Awaitility.waitAtMost(Duration.ofMinutes(1))
                       .pollDelay(Duration.ofSeconds(1))
                       .until(() -> cluster.get(1).callOnInstance(() -> Schema.instance.getTableMetadata(KEYSPACE, "tbl") != null));
 
+            // now let's break schema locally and let it be reset
+            cluster.get(1).runOnInstance(() -> Schema.instance.getLocalKeyspaces()
+                                                              .get(SchemaConstants.SCHEMA_KEYSPACE_NAME)
+                                                              .get().tables.forEach(t -> ColumnFamilyStore.getIfExists(t.keyspace, t.name).truncateBlockingWithoutSnapshot()));
```
ekaterinadimitrova2, line 177: This line also seems too long

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
```diff
@@ -19,15 +19,27 @@
...
+import org.apache.cassandra.config.CassandraRelevantProperties;
+import org.apache.cassandra.db.ColumnFamilyStore;
 import org.apache.cassandra.distributed.Cluster;
 import org.apache.cassandra.distributed.api.ConsistencyLevel;
 import org.apache.cassandra.distributed.api.Feature;
+import org.apache.cassandra.gms.Gossiper;
 import org.apache.cassandra.schema.Schema;
+import org.apache.cassandra.schema.SchemaConstants;
+import org.apache.cassandra.schema.SchemaKeyspace;
+import org.apache.cassandra.service.StorageService;
```
ekaterinadimitrova2: Unused imports

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
```diff
@@ -19,15 +19,27 @@
...
+import org.apache.cassandra.utils.FBUtilities;
+import org.assertj.core.api.Assertions;
 import org.awaitility.Awaitility;
 
+import static org.apache.cassandra.schema.SchemaKeyspaceTables.ALL;
```
ekaterinadimitrova2: Unused import

ekaterinadimitrova2 (commented):
Left some small comments, I want to do one more pass and check further the MigrationCoordinatorTest as I just skimmed through it. I want to be sure I didn't miss anything. I also want to look further in 4.0. Not an area of the code that I have a lot of experience

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
adelapena: I'd add a brief JavaDoc line saying what the property does.

adelapena: Probably we should include the unit in the name, like in `SCHEMA_PULL_INTERVAL_MS`.

src/java/org/apache/cassandra/schema/DefaultSchemaUpdateHandler.java
```diff
@@ -85,8 +90,21 @@ public DefaultSchemaUpdateHandler(MigrationCoordinator migrationCoordinator,
...
+            {
+                if (requestedReset == null)
+                    applyMutations(msg.payload);
+            }
+        });
+        SchemaPullVerbHandler.instance.register(msg -> {
+            try
+            {
+                messagingService.send(msg.responseWith(getSchemaMutations()), msg.from());
+            } catch (RuntimeException ex) {
```
adelapena:
```suggestion
            }
            catch (RuntimeException ex)
            {
```

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -23,24 +23,24 @@
...
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.LongSupplier;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
```
adelapena: Nit: unused import

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -85,6 +85,10 @@
  * schema. It performs periodic checks and if there is a schema version mismatch between the current node and the other
  * node, it pulls the schema and applies the changes locally through the callback.
  *
+ * In particular the Migration Coordinator keeps track of all schema versions reported from each node in the cluster.
+ * As long as the certain version is advertised by some node, it is being tracked. As long as a version is tracked,
```
adelapena:
```suggestion
 * As long as a certain version is advertised by some node, it is being tracked. As long as a version is tracked,
```

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -180,6 +199,18 @@ boolean wasReceived()
...
         }
+
+        @Override
+        public String toString()
+        {
+            return "VersionInfo{" +
+                   "version=" + version +
+                   ", outstandingRequests=" + outstandingRequests +
+                   ", requestQueue=" + requestQueue +
+                   ", waitQueue=" + waitQueue.getWaiting() +
```
adelapena: Maybe it should be `waitQueueSize`?
```suggestion
                   ", waitQueueSize=" + waitQueue.getWaiting() +
```

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -611,18 +615,27 @@ public SchemaTransformationResult transform(SchemaTransformation transformation,
...
-        SchemaDiagnostics.schemaCleared(this);
+        if (Gossiper.instance.getLiveMembers().stream().allMatch(ep -> FBUtilities.getBroadcastAddressAndPort().equals(ep)))
+            throw new InvalidRequestException("Cannot reset local schema when there are no other live nodes");
 
-        updateHandler.reset(false);
-        logger.info("Local schema reset is complete.");
+        Awaitable clearCompletion = updateHandler.clear();
+        try
+        {
+            if (!clearCompletion.await(StorageService.SCHEMA_DELAY_MILLIS, TimeUnit.MILLISECONDS)) {
```
adelapena:
```suggestion
            if (!clearCompletion.await(StorageService.SCHEMA_DELAY_MILLIS, TimeUnit.MILLISECONDS))
            {
```

src/java/org/apache/cassandra/schema/SchemaUpdateHandler.java
```diff
@@ -63,14 +64,15 @@
...
      */
-    SchemaTransformationResult reset(boolean local);
+    void reset(boolean local);
 
     /**
      * Clears the locally stored schema entirely. After this operation the schema is equal to {@link DistributedSchema#EMPTY}.
      * The method does not execute any callback. It is indended to reinitialize the schema later using the method
      * {@link #reset(boolean)}.
+     *
+     * @return
```
adelapena: I think we should either add some description to the `@return` tag or remove it.

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
```diff
@@ -95,25 +107,54 @@ private void selectSilent(Cluster cluster, String name)
...
             cluster.get(2).shutdown().get();
 
-            // when schema is removed and there is no other node to fetch it from, node 1 should be left with clean schema
-            cluster.get(1).runOnInstance(() -> Schema.instance.resetLocalSchema());
+            Awaitility.await().atMost(Duration.ofSeconds(30)) // wait until node 1 notices that node 2 is dead
+                      .until(() -> cluster.get(1).callOnInstance(() -> Gossiper.instance.getLiveMembers().stream().allMatch(e -> e.equals(FBUtilities.getBroadcastAddressAndPort()))));
+
+            // when there is no node to fetch the schema from, reset local schema should immediately fail
+            Assertions.assertThatExceptionOfType(RuntimeException.class).isThrownBy(() -> {
+                cluster.get(1).runOnInstance(() -> Schema.instance.resetLocalSchema());
```
adelapena, line 150:
```suggestion
                cluster.get(1).runOnInstance(Schema.instance::resetLocalSchema);
```

adelapena: Please ignore this last suggestion, I forgot that method references don't work in the dtest framework.

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
adelapena: Still trying to understand why it's needed, but for sure a quick attempt at removing it breaks the tests.

ekaterinadimitrova2: The same logic is in `pullComplete`.
Seems to me we `markReceived` before we deal with `requestQueue` in `maybePullSchema`, I am still not sure why though.... looking again into that

adelapena: Actually, it hasn't necessarily marked are received, has it? It's only marked as received if it's equals to the local schema, I think.

ekaterinadimitrova2: Yes, reading back that is right but then it is confusing why in those cases we add to the requestQueue which seems to be a queue of endpoints from which we are going to fetch.... do we need still to fetch? :-)

ekaterinadimitrova2: Oh, that is covered in maybePullSchema, we check whether it was already marked received on the first line... It is just not immediately clear before you look into maybePullSchema... perhaps we can simplify or at least comment a bit...

ekaterinadimitrova2: At that point we  do not remove the endpoint from the queue, we seem to remove it next time we call maybePullSchema and we iterate the queue. Not sure why that approach was taken really and not just not adding it at all in case we markReceived... I will experiment

adelapena: So we add it to `requestQueue` because we might need to fetch it, unless it has been marked as received. I guess the simplified version would be adding the endpoint to the request queue only if hasn't been marked, something like:
```java
VersionInfo info = versionInfo.computeIfAbsent(version, VersionInfo::new);
info.endpoints.add(endpoint);
logger.trace("Added endpoint {} to schema {}: {}", endpoint, info.version, info);

if (Objects.equals(schemaVersion.get(), version))
{
    info.markReceived();
    logger.trace("Schema {} from {} has been marked as received because it is equal the local schema", version, endpoint);
}
else
{
    info.requestQueue.addFirst(endpoint);
}
```

ekaterinadimitrova2: Yes, that was my point but I have to test if it can have some other side effects that we might miss...

ekaterinadimitrova2: At least the `MigrationCoordinatorTest` and `SchemaTest` (both in-jvm and unit tests) pass for me locally with this change

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
jacek-lewandowski: > Please ignore this last suggestion, I forgot that method references don't work in the dtest framework.

heh... yes, I struggled this problem some time ago. `runElsewhere(Schema.instance::resetLocalSchema)` means more or less:
```java
Schema schema = Schema.instance;
runElsewhere(() -> schema.resetLocalSchema());
```
so it is something completely different :-)

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
jacek-lewandowski: I'll revert

jacek-lewandowski: but actually I considered this message as temporary situation - when the messaging version is eventually known, the schema is either pulled or it is logged that the messaging version is incompatible (at debug level)

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
jacek-lewandowski: Ok, but where do you want it to be documented? I think we should rather fix it

src/java/org/apache/cassandra/schema/Schema.java
jacek-lewandowski: Most general answer to that question is - because we are waiting in this method. The method simply mark a schema that is should be refreshed and wait until it happens. However, the schema would be unable to refresh if we hold a lock because we no changes could be applied then. Note that, the changes are applied from the MigrationCoordinator thread, usually a different to the called thread.

src/java/org/apache/cassandra/schema/SchemaUpdateHandler.java
jacek-lewandowski: ok, and btw. I suppose the more adequate names for those methods now would be reset -> sync, clear -> refresh

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
jacek-lewandowski: I will, but with the current formatting rules I can see no way to format it nicely

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
jacek-lewandowski: again, the formatting rules does not allow for make it nicely to my knowledge

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
jacek-lewandowski: we do not have to call `.toString()` explicitly in statements like `log.info("bla bla bla {}", info);` :)

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
jacek-lewandowski: exactly, that was what I've commented - "...given we've just marked this schema version as received..." :-)

jacek-lewandowski:
Thank you for thorough review, I've applied your suggestions

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
ekaterinadimitrova2: Then in a follow up ticket in order to be fixed and add the number in this comment? WDYT?

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
```diff
@@ -93,70 +99,98 @@ private void selectSilent(Cluster cluster, String name)
...
     }
 
+    /**
+     * The purpose of this test is to verify manual schema reset functinality.
+     * <p>
+     * There is a 2-node cluster and a table one created. The schema version is agreed on both nodes. Then the 2nd node
+     * is shutdown, and we introduce a disagreement by dropping table one and creating table two. Therefore, 1st node
+     * has a newer schema version with table two, while the shutdown 2nd node has older schema version with table one.
+     * <p>
+     * At this point, if we just start 2nd node, it synces its schema by getting the fresh mutations from 1st node which
```
ekaterinadimitrova2: typo: `syncs`

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
```diff
@@ -93,70 +99,98 @@ private void selectSilent(Cluster cluster, String name)
...
                 cause = cause.getCause();
             }
             assertTrue(causeIsUnknownColumn);
         }
     }
 
+    /**
+     * The purpose of this test is to verify manual schema reset functinality.
+     * <p>
+     * There is a 2-node cluster and a table one created. The schema version is agreed on both nodes. Then the 2nd node
```
ekaterinadimitrova2: I would use the names here probably `tbl_one` or so.

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
```diff
@@ -93,70 +99,98 @@ private void selectSilent(Cluster cluster, String name)
...
+            assertTrue(checkTables(cluster.get(2), true, false));
         }
     }
 
+    private static ConditionFactory await(int seconds)
+    {
+        return Awaitility.await().atMost(ofSeconds(seconds)).pollDelay(ofSeconds(1));
+    }
+
+    private static boolean checkTables(IInvokableInstance instance, boolean one, boolean two)
```
ekaterinadimitrova2: I would probably change this name to `checkTablesPropagated` or something

test/distributed/org/apache/cassandra/distributed/test/SchemaTest.java
```diff
@@ -93,70 +99,98 @@ private void selectSilent(Cluster cluster, String name)
...
             }).withMessageContaining("Cannot reset local schema when there are no other live nodes");
 
-            // now, let's make a disagreement, the shutdown node 2 has a definition of tbl, while the running node 1 does not
-            cluster.get(1).runOnInstance(() -> {
-                Schema.instance.transform(current -> Schema.instance.distributedKeyspaces().without(KEYSPACE), false);
-             });
-
-            assertTrue(cluster.get(1).callOnInstance(() -> Schema.instance.getTableMetadata(KEYSPACE, "tbl") == null));
+            // now, let's make a disagreement, the shutdown node 2 has a definition of table one, while
+            // the running node 2 will have definitions of table two
```
ekaterinadimitrova2: `a definition`

jacek-lewandowski:
Merged to cassandra-4.1 in d8bbeb9e398723e5efe13166e16372d104fb93ce

PR #1814, opened 2022-08-24
CASSANDRA-17823 Mark antora.yml as prelease:true

minho95:


minho95:
Jira ticket: https://issues.apache.org/jira/browse/CASSANDRA-17823

PR #1815, opened 2022-08-24
Fix typo from actie to active CASSANDRA-17845.

dallinb:
My first attempt at submitting a patch so apologies if incorrect.  This fixes the issue raised in [CASSANDRA-17845](https://issues.apache.org/jira/browse/CASSANDRA-17845) which pointed out that `actie` should be `active`.


smiklosovic:
this will be included in https://github.com/apache/cassandra/pull/2062

PR #1816, opened 2022-08-24
CASSANDRA-17857: upsert with adder support is not consistent with numbers and strings in LWT

dcapwell:


belliottsmith (commented):
So, I'm happy with the patch, but something occurs to me: these changes make us inconsistent with _counter_ fields, which (must, given how they are implemented) treat `NULL` as zero.

This is probably still the best outcome, but perhaps we should raise a DISCUSS thread on the mailing list to confirm? 

I think it is fine that operations over standard columns are treated differently to counter columns, which are a special data type not present in SQL. But since we are committing the project to a course of action here, it might be better to at least seek endorsement.

dcapwell:
> This is probably still the best outcome, but perhaps we should raise a DISCUSS thread on the mailing list to confirm?

Fair, for my use was I prefer `null + 42 = 42` rather than `null` (matching counter) but it is departing from SQL so should prob be more widely agreed too.

test/distributed/org/apache/cassandra/distributed/test/CASAddTest.java
```diff
@@ -62,6 +65,48 @@ public void testAddition() throws Throwable
...
 
+    @Test
+    public void testAdditionNotExists() throws Throwable
+    {
+        try (Cluster cluster = init(Cluster.create(3)))
+        {
+            cluster.schemaChange("CREATE TABLE " + KEYSPACE + ".tbl (pk int PRIMARY KEY, a int, b text, undefined blob)");
+
+            // n = n + value where n = null
+            cluster.coordinator(1).execute("UPDATE " + KEYSPACE + ".tbl SET a = a + 1, b = b + 'fail' WHERE pk = 1 IF undefined = NULL", ConsistencyLevel.QUORUM);
```
maedhroz: nit: If you don't want to resort to using the dummy blob, I think you could rewrite like this:

```
cluster.schemaChange("CREATE TABLE " + KEYSPACE + ".tbl (pk int PRIMARY KEY, a int, b text)");

// n = n + value where n = null
cluster.coordinator(1).execute("INSERT INTO " + KEYSPACE + ".tbl (pk) VALUES (1)", ConsistencyLevel.QUORUM);
cluster.coordinator(1).execute("UPDATE " + KEYSPACE + ".tbl SET a = a + 1, b = b + 'fail' WHERE pk = 1 IF EXISTS", ConsistencyLevel.QUORUM);
assertRows(cluster.coordinator(1).execute("SELECT * FROM " + KEYSPACE + ".tbl WHERE pk = 1", ConsistencyLevel.SERIAL), row(1, null, null));
...
// have cas add defaults when missing
cluster.coordinator(1).execute("UPDATE " + KEYSPACE + ".tbl SET a = a + 1, b = b + 'success' WHERE pk = 1 IF EXISTS", ConsistencyLevel.QUORUM);
assertRows(cluster.coordinator(1).execute("SELECT * FROM " + KEYSPACE + ".tbl WHERE pk = 1", ConsistencyLevel.SERIAL), row(1, 1, "success"));
```

maedhroz (approved):
I'm fine w/ the SQL-style "null is a black hole" behavior.

Just dropped a minor suggestion on the new test.

test/distributed/org/apache/cassandra/distributed/test/CASAddTest.java
dcapwell: from a code point of view partition/row not exist looks the same as column not exist... so that should be fine.

PR #1818, opened 2022-08-25
CASSANDRA-17668 handle leak of non-standard Java types as clients using JMX cannot handle them

lmtrombone:
DatabaseDescriptor had changes where some methods threw non-standard exceptions.  This caused some of the JMX setter methods that called them to throw them as well which causes issues as the clients can't handle them.  This pull request addresses these issues by creating new APIs that catches and converts the non-standard exceptions to standard ones and deprecates the old setters.  For consistency, since the setters have been deprecated, new getters have been created as well.


lmtrombone:
Closing as it as been decided that we will be making changes only to trunk.

PR #1828, opened 2022-08-30
CASSANDRA-16860

smiklosovic:


src/java/org/apache/cassandra/tools/nodetool/ClearSnapshot.java
```diff
@@ -43,34 +46,60 @@ public class ClearSnapshot extends NodeToolCmd
     @Option(title = "clear_all_snapshots", name = "--all", description = "Removes all snapshots")
     private boolean clearAllSnapshots = false;
 
+    @Option(title = "older_than", name = "--older-than", description = "Clear snapshots older than specified time period.")
+    private String olderThan;
+
+    @Option(title = "older_than_timestamp", name = "--older-than-timestamp", description = "Clear snapshots older than specified timestamp.")
+    private Long olderThanTimestamp;
```
pauloricardomg: Why is this a long and not a isodate (`2022-09-18T12:09:54+0000`) ?

test/unit/org/apache/cassandra/tools/nodetool/ClearSnapshotTest.java
```diff
@@ -52,60 +71,142 @@ public static void teardown() throws IOException
...
         tool.assertOnCleanExit();
         assertThat(tool.getStdout()).isNotEmpty();
         
-        Map<String, TabularData> snapshots_after = probe.getSnapshotDetails();
+        Map<String, TabularData> snapshots_after = probe.getSnapshotDetails(emptyMap());
         assertThat(snapshots_after).isEmpty();
     }
+
+    @Test
+    public void testClearSnapshotWithOlderThanFlag() throws Throwable
```
pauloricardomg, line 114: Are we testing `olderThanTimestamp` ?

test/unit/org/apache/cassandra/tools/nodetool/ClearSnapshotTest.java
```diff
@@ -19,21 +19,40 @@
...
+import static java.lang.String.format;
+import static java.time.temporal.ChronoUnit.HOURS;
+import static java.time.temporal.ChronoUnit.MINUTES;
+import static java.util.Collections.emptyMap;
+import static org.apache.cassandra.tools.ToolRunner.invokeNodetool;
+import static org.apache.cassandra.utils.Clock.Global.currentTimeMillis;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.awaitility.Awaitility.await;
 
 public class ClearSnapshotTest extends CQLTester
```
pauloricardomg, line 53: Can you add a test that it's not possible to use `--older-than` or `--older-than-timestamp` unless `--all` is specified?

src/java/org/apache/cassandra/tools/nodetool/ClearSnapshot.java
```diff
@@ -43,34 +46,60 @@ public class ClearSnapshot extends NodeToolCmd
...
+
     @Override
     public void execute(NodeProbe probe)
     {
-        if(snapshotName.isEmpty() && !clearAllSnapshots)
+        if (snapshotName.isEmpty() && !clearAllSnapshots)
             throw new RuntimeException("Specify snapshot name or --all");
 
-        if(!snapshotName.isEmpty() && clearAllSnapshots)
+        if (!snapshotName.isEmpty() && clearAllSnapshots)
```
pauloricardomg, line 64: I think it doesn't make sense to use `--older-than` or `--older-than-timestamp` unless `-all` is also specified. Should we error out if  `--older-than` or `--older-than-timestamp` is specified without `--all` ?

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
```diff
@@ -325,7 +325,7 @@ static File getLiveFileFromSnapshotFile(Path snapshotFilePath)
         return new File(liveDir.toString(), snapshotFilePath.getFileName().toString());
     }
 
-    public static Predicate<TableSnapshot> shouldClearSnapshot(String tag)
+    public static Predicate<TableSnapshot> shouldClearSnapshot(String tag, long olderThanTimestamp)
```
pauloricardomg, line 328: Should we add a specific unit test for this method in addition to the nodetool tests? I think this method is critical enough that deserves a specific unit test to ensure it's doing the right thing (and it will stay that way).

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4207,8 +4215,25 @@ public void clearSnapshot(String tag, String... keyspaceNames)
             }
         }
 
+        Object olderThan = options.get("older_than");
```
pauloricardomg, line 4225: I was wondering if we need to support both options in the server or just `clearOlderThanTimestamp`, since we can easily construct that in nodetool from `olderThan` and not need to parse it in the server. On the other hand, if people access via JMX directly they might lose this functionality. WDYT?

src/java/org/apache/cassandra/tools/nodetool/ClearSnapshot.java
smiklosovic: @pauloricardomg why is not it a good idea to be able to use it when "--all" is not specified? You might want to remove snapshots only per table or per keyspace and not touch the others.

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: Imagine there is unix timestamp 1000 at client but time is different on server, like 1050. So when I specify 3h on the command line, I would take the current time on the command line (1000) and subtract 3h from it, for example, just to have an example, it would be timestamp 800.

But timestamp 800 is not 3 hours back from the timestamp 1050 on the server. I want to compute "three hours back" not on the client but on the server.

Please correct me if my reasoning is wrong here.

src/java/org/apache/cassandra/tools/nodetool/ClearSnapshot.java
smiklosovic: No problem, I can change this to iso, the string provided would be parsed before calling jmx method so we know it is valid.

test/unit/org/apache/cassandra/tools/nodetool/ClearSnapshotTest.java
smiklosovic: We are, at the bottom of that test method. I added a dedicated test method for it so we do not mix two flags together.

src/java/org/apache/cassandra/tools/nodetool/ClearSnapshot.java
smiklosovic: Changed to String and parsing it to be sure it is valid Instant.

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
smiklosovic: added

src/java/org/apache/cassandra/service/StorageService.java
pauloricardomg: Makes sense. Thanks for clarifying.

src/java/org/apache/cassandra/tools/nodetool/ClearSnapshot.java
pauloricardomg: Ok, makes sense. Maybe we should just disallow older-than when a specific tag is specified? Can you add a test for deletion per table and keyspace? Thanks

smiklosovic: I added test for illegal arguments and their combinations to cover it all.

Btw I noticed that in the current clearsnapshot command we can not clear per table, it was always like that, only per keyspace. But still, we might be interested only in one keyspace and not the others on --all flag so the above logic still holds.

PR #1830, opened 2022-08-31
Remove frozen section link

tlblessing:
There is no section named frozen in this file. Either add one in or remove the link that does not work.


smiklosovic:
thank you, this will be in included under PR #2062

PR #1831, opened 2022-08-31
CASSANDRA-10383

smiklosovic:


src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -563,6 +565,11 @@ private static void addTableParamsToRowBuilder(TableParams params, Row.SimpleBui
         // in mixed operation with pre-4.1 versioned node during upgrades.
         if (params.memtable != MemtableParams.DEFAULT)
             builder.add("memtable", params.memtable.configurationKey());
+
+        // As above, only add the allow_auto_snapshot column if the value is not default (true) and
+        // auto-snapshotting is enabled, to avoid RTE in pre-4.2 versioned node during upgrades
+        if (DatabaseDescriptor.isAutoSnapshot() && !params.allowAutoSnapshot)
```
iamaleksey: I would eliminate the `DatabaseDescriptor.isAutoSnapshot()` condition here - I'm unsure why it was added here? Only the value of the table param is important here.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -955,6 +962,10 @@ private static TableMetadata fetchTable(String keyspaceName, String tableName, T
     static TableParams createTableParamsFromRow(UntypedResultSet.Row row)
     {
         return TableParams.builder()
+                          // allow_auto_snapshot column was introduced in 4.2
+                          .allowAutoSnapshot(row.has("allow_auto_snapshot") ?
+                                             row.getBoolean("allow_auto_snapshot", DatabaseDescriptor.isAutoSnapshot()) :
+                                             DatabaseDescriptor.isAutoSnapshot())
```
iamaleksey: Again, when it comes to schema things, `DatabaseDescriptor.isAutoSnapshot()` shouldn't be relevant. Only set the value on the builder at all if the column is present in the row. Otherwise let it pick up the default builder value of `true` and later defer to `ColumnFamilyStore.isAutoSnapshot()` to see if both the per-table setting and the node-local one are true.

src/java/org/apache/cassandra/cql3/UntypedResultSet.java
```diff
@@ -359,6 +359,12 @@ public boolean getBoolean(String column)
             return BooleanType.instance.compose(data.get(column));
         }
 
+        public boolean getBoolean(String column, boolean ifNull)
+        {
+            ByteBuffer bytes = data.get(column);
+            return bytes == null ? ifNull : BooleanType.instance.compose(bytes);
+        }
+
```
iamaleksey: The method will no longer be needed if the other comment in this review is followed.

iamaleksey (changes requested):

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -568,8 +568,7 @@ private static void addTableParamsToRowBuilder(TableParams params, Row.SimpleBui
 
         // As above, only add the allow_auto_snapshot column if the value is not default (true) and
         // auto-snapshotting is enabled, to avoid RTE in pre-4.2 versioned node during upgrades
-        if (DatabaseDescriptor.isAutoSnapshot() && !params.allowAutoSnapshot)
-            builder.add("allow_auto_snapshot", false);
+        builder.add("allow_auto_snapshot", params.allowAutoSnapshot);
```
iamaleksey: You still want to gate it on `!params.allowAutoSnapshot`, so it's only persisted if explicitly set to `false`. Just don't involve `DatabaseDescriptor.isAutoSnapshot()`.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -961,31 +960,33 @@ private static TableMetadata fetchTable(String keyspaceName, String tableName, T
...
+                                                 .additionalWritePolicy(row.has("additional_write_policy") ?
+                                                                        SpeculativeRetryPolicy.fromString(row.getString("additional_write_policy")) :
+                                                                        SpeculativeRetryPolicy.fromString("99PERCENTILE"))
+                                                 .cdc(row.has("cdc") && row.getBoolean("cdc"))
+                                                 .readRepair(getReadRepairStrategy(row));
+
+        // allow_auto_snapshot column was introduced in 4.2
+        if (row.has("allow_auto_snapshot")) {
+            builder.allowAutoSnapshot(row.getBoolean("allow_auto_snapshot"));
+        }
```
iamaleksey: Either put braces on a new line or don't do braces at all? According to our style guide. I would prefer the latter, no braces.

iamaleksey (changes requested):

src/java/org/apache/cassandra/cql3/statements/schema/TableAttributes.java
```diff
@@ -98,6 +98,9 @@ public static Set<String> allKeywords()
 
     private TableParams build(TableParams.Builder builder)
     {
+        if (hasOption(Option.ALLOW_AUTO_SNAPSHOT))
+            builder.allowAutoSnapshot(getBoolean(Option.ALLOW_AUTO_SNAPSHOT.toString(), false));
```
iamaleksey: I don't think this default value is correct. But more importantly, we know the option is present, so we shouldn't even be passing the default. Need to add a method that doesn't take a default value. And also do the same for `cdc` while here.

iamaleksey (approved):

iamaleksey (approved):
LGTM

PR #1832, opened 2022-09-01
CASSANDRA-17750

aratno:
This is the 3.0 backport - see #1725


aratno:
This change is going to be trunk-only; closing.

PR #1833, opened 2022-09-01
CASSANDRA-17874 only reload compaction strategies if disk boundaries change

krummas:


iamaleksey (commented):
Had some trouble understanding mostly pre-existing `reload()` code, so, since it's for trunk, came up with some cleanup suggestions that I think make it much easier to follow, here: https://github.com/iamaleksey/cassandra/commits/17874-review

iamaleksey (approved):

PR #1853, opened 2022-09-08
CASSANDRA-16791 Introduce auto_snapshot_ttl table parameter

smiklosovic:
patch by Stefan Miklosovic; reviewed by TODO for CASSANDRA-16791

### DIFF
--- a/pylib/cqlshlib/cql3handling.py
+++ b/pylib/cqlshlib/cql3handling.py
@@ -44,6 +44,7 @@ class Cql3ParsingRuleSet(CqlParsingRuleSet):
 
     columnfamily_layout_options = (
         ('allow_auto_snapshot', None),
+        ('auto_snapshot_ttl', None),
         ('bloom_filter_fp_chance', None),
         ('comment', None),
         ('gc_grace_seconds', None),
@@ -517,6 +518,8 @@ def cf_prop_val_completer(ctxt, cass):
         return [Hint('<\'none\'|\'blocking\'>')]
     if this_opt == 'allow_auto_snapshot':
         return [Hint('<boolean>')]
+    if this_opt == 'auto_snapshot_ttl':
+        return [Hint('<duration>')]
     return [Hint('<option_value>')]
 
 
--- a/pylib/cqlshlib/test/test_cqlsh_completion.py
+++ b/pylib/cqlshlib/test/test_cqlsh_completion.py
@@ -617,6 +617,7 @@ class TestCqlshCompletion(CqlshCompletionCase):
                             immediate='ITH ')
         self.trycompletions(prefix + ' new_table (col_a int PRIMARY KEY) WITH ',
                             choices=['allow_auto_snapshot',
+                                     'auto_snapshot_ttl',
                                      'bloom_filter_fp_chance', 'compaction',
                                      'compression',
                                      'default_time_to_live', 'gc_grace_seconds',
@@ -627,6 +628,7 @@ class TestCqlshCompletion(CqlshCompletionCase):
                                      'min_index_interval', 'speculative_retry', 'additional_write_policy', 'cdc', 'read_repair'])
         self.trycompletions(prefix + ' new_table (col_a int PRIMARY KEY) WITH ',
                             choices=['allow_auto_snapshot',
+                                     'auto_snapshot_ttl',
                                      'bloom_filter_fp_chance', 'compaction',
                                      'compression',
                                      'default_time_to_live', 'gc_grace_seconds',
@@ -675,9 +677,13 @@ class TestCqlshCompletion(CqlshCompletionCase):
                             choices=[';', 'AND'])
         self.trycompletions(prefix + " new_table (col_a int PRIMARY KEY) WITH compaction = "
                             + "{'class': 'SizeTieredCompactionStrategy'} AND ",
-                            choices=['allow_auto_snapshot', 'bloom_filter_fp_chance', 'compaction',
+                            choices=['allow_auto_snapshot',
+                                     'auto_snapshot_ttl',
+                                     'bloom_filter_fp_chance',
+                                     'compaction',
                                      'compression',
-                                     'default_time_to_live', 'gc_grace_seconds',
+                                     'default_time_to_live',
+                                     'gc_grace_seconds',
                                      'max_index_interval',
                                      'memtable_flush_period_in_ms',
                                      'CLUSTERING',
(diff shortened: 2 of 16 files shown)


pylib/cqlshlib/test/test_cqlsh_output.py
```diff
@@ -653,6 +653,7 @@ def test_describe_columnfamily_output(self):
                 varintcol varint
             ) WITH additional_write_policy = '99p'
                 AND allow_auto_snapshot = true
+                AND auto_snapshot_ttl = 0s
```
smiklosovic: default is 0s which means no TTL. There can not be negative values and 0 TTL does not make sense which makes 0s good candidate.

pylib/cqlshlib/cql3handling.py
```diff
@@ -517,6 +518,8 @@ def cf_prop_val_completer(ctxt, cass):
         return [Hint('<\'none\'|\'blocking\'>')]
     if this_opt == 'allow_auto_snapshot':
         return [Hint('<boolean>')]
+    if this_opt == 'auto_snapshot_ttl':
+        return [Hint('<duration>')]
```
smiklosovic: duration is seconds, minutes, hours or days. (s, m, h, d). This nicely relates to what DurationSpec is based on. Duration spec has resolution to `ns` but anything less than seconds for snapshot ttl is hardly applicable.

pylib/cqlshlib/test/test_cqlsh_completion.py
```diff
@@ -675,9 +677,13 @@ def create_columnfamily_table_template(self, name):
                             choices=[';', 'AND'])
         self.trycompletions(prefix + " new_table (col_a int PRIMARY KEY) WITH compaction = "
                             + "{'class': 'SizeTieredCompactionStrategy'} AND ",
-                            choices=['allow_auto_snapshot', 'bloom_filter_fp_chance', 'compaction',
```
smiklosovic: this started to be hard to visually parse so I put every option to new line.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -171,7 +171,7 @@
     private static boolean daemonInitialized;
 
     private static final int searchConcurrencyFactor = Integer.parseInt(System.getProperty(Config.PROPERTY_PREFIX + "search_concurrency_factor", "1"));
-    private static DurationSpec.IntSecondsBound autoSnapshoTtl;
+    private static DurationSpec.IntSecondsBound autoSnapshotTtl = new DurationSpec.IntSecondsBound(0);
```
smiklosovic: This is not null anymore, will be explained further.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -412,7 +412,7 @@ private static void applySimpleConfig()
         {
             try
             {
-                autoSnapshoTtl = new DurationSpec.IntSecondsBound(conf.auto_snapshot_ttl);
+                autoSnapshotTtl = new DurationSpec.IntSecondsBound(conf.auto_snapshot_ttl);
```
smiklosovic: fixed typo.

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -3078,6 +3078,16 @@ public boolean isAutoSnapshotEnabled()
         return metadata().params.allowAutoSnapshot && DatabaseDescriptor.isAutoSnapshot();
     }
 
+    public DurationSpec.IntSecondsBound getAutoSnapshotTtl()
+    {
+        // If it is not set as a table parameter (which is by default 0s), then take the default value from descriptor.
+        // It is possible that it is unset in descriptor, which again defaults to 0s.
+        // When SnapshotManifest receives ttl with null or 0s, it will not write expiration time to manifest file
+        // Hence, table parameter has precedence over descriptor value.
+        DurationSpec.IntSecondsBound ttlFromTable = metadata().params.autoSnapshotTtl;
```
smiklosovic: `autoSnapshotTtl` is not null because we need to have non-null value when e.g. describing a table. It is way easier to work with non-null values. If `DatabaseDescriptor.getAutoSnapshotTtl` was null, we would have non-null ttl from params which might be of value 0s or we would possibly return null from descriptor.

I think that this method should return non-null in every case and making special logic about nullity is cumbersome. Hence I prefer to have it non-null every time.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -567,9 +569,14 @@ private static void addTableParamsToRowBuilder(TableParams params, Row.SimpleBui
             builder.add("memtable", params.memtable.configurationKey());
 
         // As above, only add the allow_auto_snapshot column if the value is not default (true) and
-        // auto-snapshotting is enabled, to avoid RTE in pre-4.2 versioned node during upgrades
```
smiklosovic: this is leftover from the previous patch as we do not depend on what auto_snapshot is set to in cassandra.yaml so this piece of the comment is not relevant.

src/java/org/apache/cassandra/schema/TableParams.java
```diff
@@ -90,6 +94,9 @@ private TableParams(Builder builder)
     {
         comment = builder.comment;
         allowAutoSnapshot = builder.allowAutoSnapshot;
+        autoSnapshotTtl = builder.autoSnapshotTtl.toSeconds() == 0
```
smiklosovic: we can not get null here as getAutoSnapshotTtl is not null anymore.

src/java/org/apache/cassandra/schema/TableParams.java
```diff
@@ -325,6 +338,7 @@ public void appendCqlTo(CqlBuilder builder, boolean isView)
     {
         private String comment = "";
         private boolean allowAutoSnapshot = true;
+        private DurationSpec.IntSecondsBound autoSnapshotTtl = new DurationSpec.IntSecondsBound(0);
```
smiklosovic: default is 0s when not set.

src/java/org/apache/cassandra/service/snapshot/SnapshotManifest.java
```diff
@@ -63,7 +63,7 @@ public SnapshotManifest(List<String> files, DurationSpec.IntSecondsBound ttl, In
     {
         this.files = files;
         this.createdAt = creationTime;
-        this.expiresAt = ttl == null ? null : createdAt.plusSeconds(ttl.toSeconds());
+        this.expiresAt = (ttl == null || ttl.toSeconds() == 0) ? null : createdAt.plusSeconds(ttl.toSeconds());
```
smiklosovic: leaving `ttl == null` here is fine and desirable as there is other code path (e.g. by taking snapshot) when passing null in e.g. jmx is just fine.

PR #1855, opened 2022-09-13
Cassandra 17868 - Allow disabling hotness persistence for high sstable counts

jmckenzie-dev:
The persisting of the sstables hotness when there are 10s of thousands of sstables can have issues keeping up, and the rate limiter is hard coded. Another option may be nice to just completely disable the feature.

When sstables super backed up (from repairs) the hotness tracking tends to cause the STCS in L0 to make poor decisions, always grabbing the large sstables and skipping the tiny sstables that would benefit more.

### DIFF
--- a/test/unit/org/apache/cassandra/io/sstable/SSTableReaderTest.java
+++ b/test/unit/org/apache/cassandra/io/sstable/SSTableReaderTest.java
@@ -26,7 +26,7 @@ import java.util.concurrent.*;
 import java.util.stream.Stream;
 
 import com.google.common.collect.Sets;
-import org.apache.cassandra.io.util.File;
+
 import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.junit.Test;
@@ -34,7 +34,9 @@ import org.junit.rules.ExpectedException;
 
 import org.apache.cassandra.SchemaLoader;
 import org.apache.cassandra.Util;
+import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.cql3.Operator;
+import org.apache.cassandra.cql3.UntypedResultSet;
 import org.apache.cassandra.db.*;
 import org.apache.cassandra.db.compaction.CompactionManager;
 import org.apache.cassandra.db.compaction.OperationType;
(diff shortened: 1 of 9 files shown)


test/unit/org/apache/cassandra/io/sstable/SSTableReaderTest.java
```diff
@@ -247,6 +253,23 @@ public void testReadRateTracking()
 
         Util.getAll(Util.cmd(store, key).includeRow("0").build());
         assertEquals(2, sstable.getReadMeter().count());
+
+        // With persistence enabled, we should be able to retrieve the state of the meter.
+        sstable.maybePersistSSTableReadMeter();
+
+        UntypedResultSet meter = SystemKeyspace.readSSTableActivity(store.keyspace.getName(), store.name, sstable.descriptor.id);
+        // So this is the only issue atm; seems like it's empty
```
maedhroz, line 261: nit: What does this comment mean? Looks like the assertion verifies that it's _not_ empty?

test/unit/org/apache/cassandra/io/sstable/SSTableReaderTest.java
```diff
@@ -247,6 +253,23 @@ public void testReadRateTracking()
...
+        assertFalse(meter.isEmpty());
+
+        Util.getAll(Util.cmd(store, key).includeRow("0").build());
+        assertEquals(3, sstable.getReadMeter().count());
+
+        // After cleaning existing state and disabling persistence, there should be no meter state to read.
+        SystemKeyspace.clearSSTableReadMeter(store.keyspace.getName(), store.name, sstable.descriptor.id);
+        DatabaseDescriptor.setSStableReadRatePersistenceEnabled(false);
```
maedhroz, line 269: nit: Any risk to any of the other tests if we don't switch back to the original setting?

maedhroz (approved):

test/unit/org/apache/cassandra/io/sstable/SSTableReaderTest.java
jmckenzie-dev: Vestigial from interim debugging; removed.

test/unit/org/apache/cassandra/io/sstable/SSTableReaderTest.java
jmckenzie-dev: Good catch; I'm going to try/finally wrap it on commit just to make sure it reverts to whatever it entered with. Better safe than sorry.

PR #1863, opened 2022-09-18
fix typo

gruzilkin:
both case to both cases


smiklosovic:
thank you, this will be included in PR #2062

PR #1865, opened 2022-09-18
fix typo

gruzilkin:
missing closing parentheses


smiklosovic:
thank you, this will be included under PR #2062

PR #1866, opened 2022-09-18
add HISTORY command to CQLSH

larinachen:
JIRA #[CASSANDRA-15046] [Link](https://issues.apache.org/jira/browse/CASSANDRA-15046?jql=project%20%3D%20CASSANDRA%20AND%20resolution%20%3D%20Unresolved%20AND%20labels%20%3D%20ghc-lhf%20ORDER%20BY%20priority%20DESC%2C%20updated%20DESC)

### DIFF
--- a/.build/dependency-check-suppressions.xml
+++ b/.build/dependency-check-suppressions.xml
@@ -29,6 +29,15 @@
         <cve>CVE-2021-44521</cve>
     </suppress>
     <suppress>
+        <!--  https://issues.apache.org/jira/browse/CASSANDRA-17907 -->
+        <packageUrl regex="true">^pkg:maven/org\.yaml/snakeyaml@.*$</packageUrl>
+        <cve>CVE-2022-25857</cve>
+        <cve>CVE-2022-38749</cve>
+        <cve>CVE-2022-38750</cve>
+        <cve>CVE-2022-38751</cve>
+        <cve>CVE-2022-38752</cve>
+    </suppress>
+    <suppress>
         <!-- dependency checker identified this as a completely different package (wire) -->
         <packageUrl regex="true">^pkg:maven/net\.openhft/chronicle\-wire@.*$</packageUrl>
         <cpe>cpe:/a:wire:wire</cpe>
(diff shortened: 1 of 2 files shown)


bin/cqlsh.py
```diff
@@ -1881,6 +1881,21 @@ def do_clear(self, parsed):
...
 
+    def do_history(self,parsed):
+        """
+        history/HIS
+
+        Displays the most recent (up to 50) commands executed in cqlsh.
+        """
+        recent_history = ""
+        hisotry_length = readline.get_current_history_length()
+        for index in range(hisotry_length, hisotry_length-50, -1):
```
pauloricardomg: maybe make `hisotry_length-50` configurable as an environment variable

bin/cqlsh.py
```diff
@@ -1881,6 +1881,21 @@ def do_clear(self, parsed):
...
     do_cls = do_clear
 
+    def do_history(self,parsed):
+        """
+        history/HIS
+
+        Displays the most recent (up to 50) commands executed in cqlsh.
+        """
+        recent_history = ""
+        hisotry_length = readline.get_current_history_length()
```
pauloricardomg: fix typo `hisotry_length` -> `history_length`

bin/cqlsh.py
larinachen: What about making history_length-50 configurable via an optional input parameter instead? This way, the user doesn't have to re-configure an environmental variable every time they want to view more/less. Let me know what you think about the input parameter approach vs the environmental variable approach!

pauloricardomg: I think making the history length configurable via optional input parameter makes sense. We'd still want a configuration for the default length (ie. 50) when the parameter is not specified. This could be an environmental variable or a [cqlshrc](https://docs.datastax.com/en/dse/5.1/cql/cql/cql_reference/cqlsh_commands/cqlshCqlshrc.html) configuration. Perhaps we can start with an environment variable and later expand to allow the default size to be specified in `cqlshrc` ?

PR #1875, opened 2022-09-23
Fix typo

jbarotin:
Add a missing character return, without it, a bad HTML layout was generated : table is displayed incorporated in the note.


jbarotin:
Merged in https://github.com/apache/cassandra/pull/1909

PR #1894, opened 2022-10-01
CASSANDRA-17933 remove empty cq4 log files to prevent BinLog from failing to start

smiklosovic:
This patch also backports CASSANDRA-17595.

### DIFF
--- a/src/java/org/apache/cassandra/utils/binlog/BinLog.java
+++ b/src/java/org/apache/cassandra/utils/binlog/BinLog.java
@@ -20,7 +20,6 @@ package org.apache.cassandra.utils.binlog;
 
 import java.io.File;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
@@ -28,6 +27,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Function;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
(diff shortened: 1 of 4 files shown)


src/java/org/apache/cassandra/utils/binlog/BinLog.java
```diff
@@ -462,36 +468,61 @@ public BinLog build(boolean cleanDirectory)
...
-            return Throwables.merge(accumulate, new RuntimeException(String.format("%s is not a directory", directory)));
-        }
-        for (File f : directory.listFiles())
-        {
-            accumulate = deleteRecursively(f, accumulate);
-        }
+        return cleanDir(directory, accumulate, File::listFiles);
+    }
+
+    private static Throwable cleanDir(File directory, Throwable accumulate, Function<File, File[]> lister)
```
maedhroz: nit: Would it be clearer just to keep the name `cleanDirectory()`? (i.e. It's an overload, and shortening the name doesn't really buy us anything.)

test/unit/org/apache/cassandra/audit/AuditLoggerCleanupTest.java
```diff
@@ -0,0 +1,132 @@
...
+        // only valid files are present
+        assertLogFileExists();
+    }
+
+    private void assertLogFileExists() throws Exception
+    {
+        assertTrue(list(temporaryFolder.getRoot().toPath()).allMatch(p -> {
+            String fileName = p.getFileName().toString();
+            return fileName.endsWith("cq4") || fileName.endsWith("cq4t");
+        }));
```
maedhroz: nit: Getting a compiler warning around closing the `Stream`. I guess we could always do...

```
try (Stream<Path> stream = list(temporaryFolder.getRoot().toPath()))
{
    assertTrue(stream.allMatch(p -> {
        String fileName = p.getFileName().toString();
        return fileName.endsWith("cq4") || fileName.endsWith("cq4t");
    }));
}
```

test/unit/org/apache/cassandra/audit/AuditLoggerCleanupTest.java
```diff
@@ -0,0 +1,132 @@
...
+        emptyCq4File = temporaryFolder.newFile("20220928-12.cq4");
+
+        StorageService.instance.enableAuditLog(null, null, null, null, null, null, null, null);
+
+        assertTrue(StorageService.instance.isAuditLogEnabled());
+
+        // invalid file were removed again
+        assertFalse(emptyCq4File.exists());
+        // only valid files are present
+        assertLogFileExists();
```
maedhroz, line 109: nit: Would it make sense to "strengthen" the check in `assertLogFileExists()` to make sure files are non-empty?

src/java/org/apache/cassandra/utils/binlog/BinLog.java
```diff
@@ -462,36 +468,61 @@ public BinLog build(boolean cleanDirectory)
...
 
+    /**
+     * ChronicleQueue fails to start on cq4 files which are empty. Find such files in log dir and remove them.
+     */
+    private static Throwable cleanEmptyLogFiles(File directory, Throwable accumulate)
+    {
+        return cleanDir(directory, accumulate,
+                        (dir) -> dir.listFiles(file -> !file.isDirectory()
+                                                       && file.length() == 0
+                                                       && file.getName().endsWith(SingleChronicleQueue.SUFFIX)));
```
maedhroz: It would be nice if we make this visible in the logs, probably at WARN level, given it's not expected but also no catastrophic.

maedhroz (approved):
Thanks for addressing the nits!

PR #1895, opened 2022-10-04
CASSANDRA-17945 Update StorageService.getNativeaddress to handle IPv6 addresses

tolbertam:
[CASSANDRA-17945](https://issues.apache.org/jira/browse/CASSANDRA-17945)

StorageService.getNativeaddress does not currently correctly handle IPv6 addresses correctly when NATIVE_ADDRESS_AND_PORT are not present in that it simply concatenates the IP address with the default native port, e.g.:

0:0:0:0:0:0:5a:3:9042

This does not parse into an InetSocketAddress as the address and port can't be disambiguated.

Such a case would usually be present when there are 3.x nodes present in a cluster with 4.0 nodes.

Change updates RPC_ADDRESS and else case to create InetAddressAndPort instances with DatabaseDescriptor.getNativeTransportPort and returns the getHostAddress(withPort) which properly bracket encodes the address, e.g.:

[0:0:0:0:0:0:5a:3]:9042

which can be parsed as an InetSocketAddress.

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -2158,10 +2158,31 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
                 throw new RuntimeException(e);
             }
         }
-        else if (Gossiper.instance.getEndpointStateForEndpoint(endpoint).getApplicationState(ApplicationState.RPC_ADDRESS) == null)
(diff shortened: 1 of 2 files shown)


aweisberg:
Committed as [3bdd2caa22a0413929188536b41d8117177574fa|https://github.com/apache/cassandra/commit/3bdd2caa22a0413929188536b41d8117177574fa].

PR #1900, opened 2022-10-05
CASSANDRA-17948 Get logs via virtual table

smiklosovic:


test/distributed/org/apache/cassandra/distributed/impl/Instance.java
```diff
@@ -590,6 +591,7 @@ public void startup(ICluster cluster)
                 DistributedTestSnitch.assign(config.networkTopology());
 
                 DatabaseDescriptor.daemonInitialization();
+                ThreadAwareSecurityManager.install();
```
smiklosovic, line 594: this will call LoggingSupport stuff and will check the number of virtual appenders.

src/java/org/apache/cassandra/db/virtual/LogMessagesTable.java
```diff
@@ -0,0 +1,192 @@
...
+ */
+public final class LogMessagesTable extends AbstractMutableVirtualTable
+{
+    private static final Logger logger = LoggerFactory.getLogger(LogMessagesTable.class);
+
+    public static final int LOGS_VIRTUAL_TABLE_MIN_ROWS = 1000;
+    public static final int LOGS_VIRTUAL_TABLE_DEFAULT_ROWS = 50_000;
+    public static final int LOGS_VIRTUAL_TABLE_MAX_ROWS = 100_000;
+
+    public static final String TABLE_NAME = "logs";
```
driftx: I think we should be more descriptive and use something like "system_logs" in case there are other types in the future.

driftx (approved):

PR #1931, opened 2022-10-20
CASSANDRA-17870 nodetool/rebuild: Add flag to exclude nodes from local DC

sarankk:
During expansion by DC, when we issue nodetool/rebuild from new dc to rebuild data from other DCs. If src-dc is not passed explicitly, then C* tries to rebuild the data from the same (new) dc. 
We don’t exclude other nodes in the same DC. Only down sources and the local node itself are excluded.
```
 // We're filtering out local node and down sources
        addSourceFilter(new RangeStreamer.FailureDetectorSourceFilter(failureDetector));
        addSourceFilter(new RangeStreamer.ExcludeLocalNodeFilter());
```
We should fix nodetool/rebuild to exclude the local DC (from where we’re executing the command) while issuing nodetool/rebuild without passing src dc


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1290,7 +1295,18 @@ public void rebuild(String sourceDc, String keyspace, String tokens, String spec
             throw new IllegalStateException("Node is still rebuilding. Check nodetool netstats.");
         }
 
+<<<<<<< Updated upstream
```
frankgh: looks like there are some merge issues

sarankk: oh will remove those

src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -769,9 +769,22 @@ default int upgradeSSTables(String keyspaceName, boolean excludeCurrentVersion,
      * @param keyspace Name of the keyspace which to rebuild or null to rebuild all keyspaces.
      * @param tokens Range of tokens to rebuild or null to rebuild all token ranges. In the format of:
      *               "(start_token_1,end_token_1],(start_token_2,end_token_2],...(start_token_n,end_token_n]"
+     * @param specificSources list of sources that can be used for rebuilding. Mostly other nodes in the cluster.
```
yifan-c: nit: `list of sources that can be used for rebuilding. Must be other nodes in the cluster. The format of the string is comma separated values.`

src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -769,9 +769,22 @@ default int upgradeSSTables(String keyspaceName, boolean excludeCurrentVersion,
...
 
+    /**
+    * Same as {@link #rebuild(String)}, but only for specified keyspace and ranges. It excludes local data center nodes
+    *
+    * @param sourceDc Name of DC from which to select sources for streaming or null to pick any node
+    * @param keyspace Name of the keyspace which to rebuild or null to rebuild all keyspaces.
+    * @param tokens Range of tokens to rebuild or null to rebuild all token ranges. In the format of:
+    *               "(start_token_1,end_token_1],(start_token_2,end_token_2],...(start_token_n,end_token_n]"
+    * @param specificSources list of sources that can be used for rebuilding. Mostly other nodes in the cluster.
+    * @param excludeLocalDatacenterNodes Flag to indicate whether local data center nodes should be excluded as a source for streaming.
```
yifan-c: nit: `... be excluded as sources for streaming.`

yifan-c (commented):
LGTM with a few nits on the java doc.

test/unit/org/apache/cassandra/service/StorageServiceTest.java
```diff
@@ -298,12 +298,21 @@ public void testBatchSizeWarnThresholdInKiB()
...
             storageService.setBatchSizeWarnThresholdInKiB(previousBatchSizeWarnThreshold);
         }
     }
+
+    @Test(expected = IllegalArgumentException.class)
+    public void testLocalDatacenterNodesExcludedDuringRebuild()
+    {
+        StorageService service = StorageService.instance;
+        service.rebuild(DatabaseDescriptor.getLocalDataCenter(), "StorageServiceTest", null, null, true);
+    }
```
yifan-c: Prefer to use 
```
    @Rule
    ExpectedException expectedException = ExpectedException.none();
```

This way, you can assert the exception type and the exception message too.

frankgh: or we can use `org.assertj.core.api.Assertions.assertThrows(IllegalArgumentException.class, () -> service.rebuild(DatabaseDescriptor.getLocalDataCenter(), "StorageServiceTest", null, null, true));`

:)

yifan-c: Right, but the point is to assert the exception message too. `AssertJ` api can also do it.

frankgh: My vote is for `assertj` . I think we've started seeing more tests in the Cassandra codebase moving to `assertj` which is great!

frankgh: I believe we don't have assertThrows in the classpath :(

PR #1940, opened 2022-10-21
[CASSANDRA-17950] Enable dtest-offheap in CircleCI

dchenbecker:
The dtest-offheap test was only added to the Jenkins configuration, so this commit adds it to the CircleCI build for parity. Because dtest-offheap in Jenkins is configured to skip resource intensive tests, the dtest-offheap suite is added to all three sets (LOW, MEDIUM, HIGH).

Patch by Derek Chen-Becker; reviewed by Ekaterina Dimitrova (@ekaterinadimitrova2) for CASSANDRA-17950

### DIFF
--- a/.circleci/config-2_1.yml
+++ b/.circleci/config-2_1.yml
@@ -391,6 +391,17 @@ pre-commit_jobs: &pre-commit_jobs
     - j8_repeated_ant_test:
         requires:
           - build
+    # Java 8 off-heap tests
+    - start_offheap_dtest:
+        type: approval
+    - j8_dtests_offheap:
+        requires:
+          - start_offheap_dtest
+          - build
+    - j8_dtests_offheap_repeat:
+        requires:
+          - start_offheap_dtest
+          - build
 
 workflows:
     version: 2
@@ -654,6 +665,36 @@ jobs:
           upgrade: "true"
           stop_on_failure: ${REPEATED_TESTS_STOP_ON_FAILURE}
           count: ${REPEATED_UPGRADE_DTESTS_COUNT}
+
+  j8_dtests_offheap:
+    <<: *j8_par_executor
+    steps:
+      - attach_workspace:
+          at: /home/cassandra
+      - log_environment
+      - clone_dtest
+      - create_venv
+      - create_dtest_containers:
+          file_tag: j8_dtests_offheap
+          run_dtests_extra_args: "--use-vnodes --use-off-heap-memtables --skip-resource-intensive-tests"
+      - run_dtests:
+          file_tag: j8_dtests_offheap
+          pytest_extra_args: '--use-vnodes --num-tokens=16 --use-off-heap-memtables --skip-resource-intensive-tests'
+
+  j8_dtests_offheap_repeat:
+    <<: *j8_repeated_dtest_executor
+    steps:
+      - attach_workspace:
+          at: /home/cassandra
+      - clone_dtest
+      - create_venv
+      - run_repeated_dtest:
+          tests: ${REPEATED_DTESTS}
+          vnodes: "true"
+          upgrade: "false"
+          count: ${REPEATED_DTESTS_COUNT}
+          stop_on_failure: ${REPEATED_TESTS_STOP_ON_FAILURE}
+          extra_dtest_args: "--use-off-heap-memtables --skip-resource-intensive-tests"
 
 commands:
   log_environment:
@@ -1288,6 +1329,9 @@ commands:
         type: string
       stop_on_failure:
         type: string
+      extra_dtest_args:
+        type: string
+        default: ""
     steps:
       - run:
           name: Run repeated Python dtest
@@ -1342,7 +1386,7 @@ commands:
                 fi
 
                 # we need the "set -o pipefail" here so that the exit code that circleci will actually use is from pytest and not the exit code from tee
-                set -o pipefail && cd ~/cassandra-dtest && pytest $vnodes_args --count=$count $stop_on_failure_arg $upgrade_arg --log-cli-level=DEBUG --junit-xml=/tmp/results/dtests/pytest_result.xml -s --cassandra-dir=/home/cassandra/cassandra --keep-test-dir $tests_arg | tee /tmp/dtest/stdout.txt
+                set -o pipefail && cd ~/cassandra-dtest && pytest $vnodes_args --count=$count $stop_on_failure_arg $upgrade_arg --log-cli-level=DEBUG --junit-xml=/tmp/results/dtests/pytest_result.xml -s --cassandra-dir=/home/cassandra/cassandra --keep-test-dir <<parameters.extra_dtest_args>> $tests_arg | tee /tmp/dtest/stdout.txt
               fi
             fi
       - store_test_results:
--- a/.circleci/config-2_1.yml.mid_res.patch
+++ b/.circleci/config-2_1.yml.mid_res.patch
@@ -73,7 +73,7 @@
  
  separate_jobs: &separate_jobs
    jobs:
-@@ -500,7 +520,7 @@
+@@ -511,7 +520,7 @@
            target: stress-test
  
    j8_dtests_vnode:
@@ -82,7 +82,7 @@
      steps:
        - attach_workspace:
            at: /home/cassandra
-@@ -514,7 +534,7 @@
+@@ -525,7 +534,7 @@
            pytest_extra_args: '--use-vnodes --num-tokens=32 --skip-resource-intensive-tests'
  
    j8_dtests:
@@ -91,7 +91,7 @@
      steps:
        - attach_workspace:
            at: /home/cassandra
-@@ -528,7 +548,7 @@
+@@ -539,7 +548,7 @@
            pytest_extra_args: '--skip-resource-intensive-tests'
  
    j8_upgrade_dtests:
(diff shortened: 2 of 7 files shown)


.circleci/config-2_1.yml
```diff
@@ -391,6 +391,13 @@ pre-commit_jobs: &pre-commit_jobs
     - j8_repeated_ant_test:
         requires:
           - build
+    # Java 8 off-heap tests
+    - start_offheap_dtest:
+        type: approval
+    - j8_offheap_dtest:
+        requires:
+          - start_offheap_dtest
+          - build
```
ekaterinadimitrova2, line 404: We will need also repeat jobs similar to other jobs in order to fight any potential flakiness :-) Something similar this : https://github.com/apache/cassandra/pull/1940/commits/97ebcb5fe6ae290da0a24228177df8e47aefbaba

.circleci/config-2_1.yml
```diff
@@ -655,6 +662,21 @@ jobs:
...
+  j8_offheap_dtest:
+    <<: *j8_par_executor
+    steps:
+      - attach_workspace:
+          at: /home/cassandra
+      - log_environment
+      - clone_dtest
+      - create_venv
+      - create_dtest_containers:
+          file_tag: j8_with_vnodes
```
ekaterinadimitrova2: Shouldn't this be something like `j8_offheap_dtest` or so?

.circleci/config-2_1.yml
```diff
@@ -655,6 +662,21 @@ jobs:
...
+          run_dtests_extra_args: "--use-vnodes --use-off-heap-memtables --skip-resource-intensive-tests --pytest-options '-k not cql'"
+      - run_dtests:
+          file_tag: j8_with_vnodes
+          pytest_extra_args: '--use-vnodes --num-tokens=16 --use-off-heap-memtables --skip-resource-intensive-tests'
```
ekaterinadimitrova2, line 682: verified with this one https://github.com/apache/cassandra-builds/blob/trunk/build-scripts/cassandra-dtest-pytest.sh#L84

.circleci/config-2_1.yml
```diff
@@ -655,6 +662,21 @@ jobs:
...
+    <<: *j8_par_executor
+    steps:
+      - attach_workspace:
+          at: /home/cassandra
+      - log_environment
+      - clone_dtest
+      - create_venv
+      - create_dtest_containers:
+          file_tag: j8_with_vnodes
+          run_dtests_extra_args: "--use-vnodes --use-off-heap-memtables --skip-resource-intensive-tests --pytest-options '-k not cql'"
```
ekaterinadimitrova2: To be honest, I am not sure where this came from - `--pytest-options '-k not cql'"`

ekaterinadimitrova2 (commented):
- Tagging also @michaelsembwever as he was also reviewing the initial PR. And also 2 committers are needed to review every patch anyway

- I do not understand this sentence - "Because dtest-offheap in Jenkins is configured to skip resource intensive tests, the dtest-offheap suite is added to all three sets (LOW, MEDIUM, HIGH)." Is it just a mistake adding `,` instead of `.`? As LOW, MEDIUM and HIGH have nothing to do with whether we skip resource-intensive tests.

- I am not sure I mentioned that even if the patches can still be applied, we need to recreate them as otherwise we get warnings like this every time on using the `generate.sh` script

```
.circleci/generate.sh -a
Generating new config.yml file with low resources and LOWRES/MIDRES/HIGHRES templates from config-2_1.yml
patching file .circleci/config-2_1.yml
Hunk #2 succeeded at 527 (offset 7 lines).
Hunk #3 succeeded at 541 (offset 7 lines).
Hunk #4 succeeded at 555 (offset 7 lines).
patching file .circleci/config-2_1.yml
```

This is a script you can use directly to apply in order to recreate the patches and get rid of the warnings (you should run it from the `.circleci` dir):
```
# apply old patches (with hunk messages)
patch -o config-2_1.yml.MIDRES config-2_1.yml config-2_1.yml.mid_res.patch
patch -o config-2_1.yml.HIGHRES config-2_1.yml config-2_1.yml.high_res.patch

# generate new patches
diff -u config-2_1.yml config-2_1.yml.HIGHRES > config-2_1.yml.high_res.patch
diff -u config-2_1.yml config-2_1.yml.MIDRES > config-2_1.yml.mid_res.patch

# verify that the new patches generate the same files and cleanup
./generate.sh -a
```

- Actually how were MIDRES and HIGHRES created? They use the same j8_par_executor as the other dtests but I see different resources in Midres - medium and 25, when j8_dtests gets large and 50?  
- In 4.0+ we will need also JDK11, once we have the details around JDK8 one that will be easy to add. Just adding here so we do not forget it. 

- We will need also repeat jobs so we can fight the flakiness :-) Something like this : https://github.com/apache/cassandra/pull/1940/commits/97ebcb5fe6ae290da0a24228177df8e47aefbaba

Please address these comments and I can push it to CI after that. Thank you for all your work!

.circleci/config-2_1.yml
dchenbecker: Sorry, I was using j8_with_vnodes as a template since it seemed to be the closest. I'll change the tag

.circleci/config-2_1.yml
dchenbecker: This is copy/pasta from the j8_with_vnodes template I was using. Will remove

.circleci/config-2_1.yml
dchenbecker: I think you meant a different commit, since that hash is the commit I've submitted :)

.circleci/config-2_1.yml
dchenbecker: Yup, this is copypasta due to using j8_with_vnodes as a template since it seemed to be the closest setup. I'll change the tags

.circleci/config-2_1.yml
dchenbecker: I honestly don't know how that got there. I'll remove it

dchenbecker:
>     * I do not understand this sentence - "Because dtest-offheap in Jenkins is configured to skip resource intensive tests, the dtest-offheap suite is added to all three sets (LOW, MEDIUM, HIGH)." Is it just a mistake adding `,` instead of `.`? As LOW, MEDIUM and HIGH have nothing to do with whether we skip resource-intensive tests.

That's a leftover from my initial misunderstanding of how things worked. Initially I thought that line was true because the offheap tests explicitly set `--skip-resource-intensive-tests`, but I'll remove that comment from the commit message

> 
>     * I am not sure I mentioned that even if the patches can still be applied, we need to recreate them as otherwise we get warnings like this every time on using the `generate.sh` script

OK, I'll fix this. Off-topic, but patch gymnastics like this are something I hope we can eliminate with pipeline parameters. I'll write up a proposal this week (as an orthogonal concern, not blocking this PR).

>     * Actually how were MIDRES and HIGHRES created? They use the same j8_par_executor as the other dtests but I see different resources in Midres - medium and 25, when j8_dtests gets large and 50?

I edited config-2_1.yml to add the new test and then ran `generate.sh -a` per the README instructions. Did I miss a step?

> 
>     * In 4.0+ we will need also JDK11, once we have the details around JDK8 one that will be easy to add. Just adding here so we do not forget it.

Yup, I'll make sure to include j11 in the 4.0+ branches. 
> 
>     * We will need also repeat jobs so we can fight the flakiness :-) Something like this : [97ebcb5](https://github.com/apache/cassandra/commit/97ebcb5fe6ae290da0a24228177df8e47aefbaba)

That's the commit I'm PRing here, so did you mean a different commit?

I feel a lot closer now than I did last week :)

Cheers!

Derek

ekaterinadimitrova2:
`OK, I'll fix this. Off-topic, but patch gymnastics like this are something I hope we can eliminate with pipeline parameters. I'll write up a proposal this week (as an orthogonal concern, not blocking this PR).`
+1, looking forward to it
`I edited config-2_1.yml to add the new test and then ran generate.sh -a per the README instructions. Did I miss a step?`

It is correct in most cases, but when we add new job things can move and apply to wrong jobs; _I think_ this is probably what happened here. That is why we need to check the final result to be sure the patches are still applicable. They were created for a bit different content after all. It seems you need to create the patches again as the old ones are not applicable anymore. Let me know if you need help. 

`That's the commit I'm PRing here, so did you mean a different commit?`
Definitely. This is what I wanted to send - https://github.com/apache/cassandra/blob/cassandra-3.11/.circleci/config-2_1.yml#L264-L269. Sorry about that

`I feel a lot closer now than I did last week :)`
Yup, almost there

dchenbecker:
I had to change branches to save some other changes, so this is being closed in favor of https://github.com/apache/cassandra/pull/1953

.circleci/config-2_1.yml
ekaterinadimitrova2: Got it, cqlsh tests in 4.0+ branches are separate jobs. -k is a pytest option and we filter cql and non cql Python D-Tests using it. But not in earlier branches where we have single dtest job. we test pyton2 in 4.0 cqlsh. This was added as part of CASSANDRA-10190. You will need it for the next branches 
https://docs.pytest.org/en/latest/how-to/usage.html
They say: "pytest -k "MyClass and not method"
This will run tests which contain names that match the given string expression (case-insensitive), which can include Python operators that use filenames, class names and function names as variables. The example above will run TestMyClass.test_something but not TestMyClass.test_method_simple."

PR #1944, opened 2022-10-24
Handle timeouts when shutting down MessagingService

krummas:
Patch by marcuse; reviewed by X for CASSANDRA-17985

### DIFF
--- a/src/java/org/apache/cassandra/net/MessagingService.java
+++ b/src/java/org/apache/cassandra/net/MessagingService.java
@@ -17,6 +17,7 @@
  */
 package org.apache.cassandra.net;
 
+import java.io.IOError;
 import java.nio.channels.ClosedChannelException;
 import java.util.ArrayList;
 import java.util.List;
@@ -206,6 +207,9 @@ public final class MessagingService extends MessagingServiceMBeanImpl
     public static final int current_version = VERSION_40;
     static AcceptVersions accept_messaging = new AcceptVersions(minimum_version, current_version);
     static AcceptVersions accept_streaming = new AcceptVersions(current_version, current_version);
+
+    @VisibleForTesting
+    public static long shutdownTimeoutMinutes = 3;
 
     public enum Version
     {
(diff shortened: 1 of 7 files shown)


src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -452,9 +455,11 @@ public void shutdown(long timeout, TimeUnit units, boolean shutdownGracefully, b
             long deadline = System.nanoTime() + units.toNanos(timeout);
             maybeFail(() -> new FutureCombiner(closing).get(timeout, units),
                       () -> {
+                          logger.info("Shutting down inbound executors ({})...", inboundSockets.sockets() != null ? inboundSockets.sockets().size() : "NULL");
```
iamaleksey: I'm pretty certain this is not supposed to ever be null, and having a null check here might give readers the wrong idea. Can we omit the check here?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4573,6 +4566,47 @@ public void run()
         }
     }
 
+    /**
+     * Tries to gracefully shutdown MessagingService, if that fails, try ungracefully.
+     *
+     * Doesn't throw TimeoutException and IOError, only logs
+     */
+    private static void shutdownMessagingServiceWithRetry()
```
iamaleksey: The method should live inside MessagingService perhaps?

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -452,9 +455,11 @@ public void shutdown(long timeout, TimeUnit units, boolean shutdownGracefully, b
...
                           List<ExecutorService> inboundExecutors = new ArrayList<>();
                           inboundSockets.close(synchronizedList(inboundExecutors)::add).get();
                           ExecutorUtils.awaitTermination(1L, TimeUnit.MINUTES, inboundExecutors);
```
iamaleksey: Should this still be using the hardcoded 1 minute constant?

src/java/org/apache/cassandra/net/MessagingService.java
iamaleksey: Do we want to have the same logging in the non-graceful branch? And specify in the message if graceful or not perhaps?

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -452,9 +455,11 @@ public void shutdown(long timeout, TimeUnit units, boolean shutdownGracefully, b
...
+                          logger.info("Inbound executors shut down");
```
iamaleksey: Likewise, do we want to have the same logging in the non-graceful branch? And specify in the message if graceful or not perhaps?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4573,6 +4566,47 @@ public void run()
...
+                    MessagingService.instance().shutdown(MessagingService.shutdownTimeoutMinutes, MINUTES, false, true);
+                }
+                catch (Exception innerEx)
+                {
+                    if (innerEx.getCause() instanceof TimeoutException)
+                        logger.error("Timed out shutting down MessagingService", e);
+                    else
+                        throw innerEx;
+                }
+
```
iamaleksey: Stray newline :)

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4573,6 +4566,47 @@ public void run()
...
+     */
+    private static void shutdownMessagingServiceWithRetry()
+    {
+        try
+        {
+            MessagingService.instance().shutdown();
+        }
+        catch (IOError ioe)
+        {
+            logger.info("Failed to shutdown message service", ioe);
```
iamaleksey: Presumably should log at error too?

iamaleksey (changes requested):

iamaleksey (approved):

PR #1948, opened 2022-10-24
Feedback/cassandra 17103

dcapwell:
Addressing feedback from https://github.com/apache/cassandra/pull/1845

### DIFF
--- a/src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
+++ b/src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
@@ -33,8 +33,6 @@ import java.util.function.Supplier;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.Lists;
-import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -46,6 +44,8 @@ import accord.primitives.Ballot;
 import accord.primitives.Deps;
 import accord.primitives.Timestamp;
 import accord.primitives.TxnId;
+import accord.txn.Txn;
+import accord.txn.Writes;
 import accord.utils.DeterministicIdentitySet;
 import org.apache.cassandra.cql3.ColumnIdentifier;
 import org.apache.cassandra.cql3.UntypedResultSet;
@@ -80,9 +80,9 @@ import org.apache.cassandra.db.rows.Row;
 import org.apache.cassandra.db.rows.RowIterator;
 import org.apache.cassandra.db.transform.FilteredPartitions;
 import org.apache.cassandra.io.IVersionedSerializer;
+import org.apache.cassandra.io.LocalVersionedSerializer;
 import org.apache.cassandra.io.util.DataInputBuffer;
 import org.apache.cassandra.io.util.DataOutputBuffer;
-import org.apache.cassandra.net.MessagingService;
 import org.apache.cassandra.schema.ColumnMetadata;
 import org.apache.cassandra.schema.Functions;
 import org.apache.cassandra.schema.KeyspaceMetadata;
@@ -115,7 +115,6 @@ public class AccordKeyspace
     private static final Logger logger = LoggerFactory.getLogger(AccordKeyspace.class);
 
     public static final String COMMANDS = "commands";
-    public static final String COMMAND_SERIES = "command_series";
     public static final String COMMANDS_FOR_KEY = "commands_for_key";
 
     private static final String TIMESTAMP_TUPLE = "tuple<bigint, bigint, int, bigint>";
@@ -134,20 +133,14 @@ public class AccordKeyspace
               + format("txn_id %s,", TIMESTAMP_TUPLE)
               + "status int,"
               + "home_key blob,"
-              + "home_key_version int,"
               + "progress_key blob,"
-              + "progress_key_version blob,"
               + "is_globally_persistent boolean,"
-              + "txn_version int,"
               + "txn blob,"
               + format("execute_at %s,", TIMESTAMP_TUPLE)
               + format("promised_ballot %s,", TIMESTAMP_TUPLE)
               + format("accepted_ballot %s,", TIMESTAMP_TUPLE)
-              + "dependencies_version int,"
               + "dependencies blob,"
-              + "writes_version int,"
               + "writes blob,"
-              + "result_version int,"
               + "result blob,"
               + format("waiting_on_commit map<%s, blob>,", TIMESTAMP_TUPLE)
               + format("waiting_on_apply map<%s, blob>,", TIMESTAMP_TUPLE)
@@ -156,6 +149,20 @@ public class AccordKeyspace
               + format("blocking_apply_on set<%s>, ", TIMESTAMP_TUPLE)
               + "PRIMARY KEY((store_generation, store_index, txn_id))"
               + ')');
+
+    private static class CommandsSerializers
+    {
+        static final LocalVersionedSerializer<AccordKey> ACCORD_KEY_SERIALIZER = localSerializer(AccordKey.serializer);
+        static final LocalVersionedSerializer<Txn> TXN_SERIALIZER = localSerializer(CommandSerializers.txn);
+        static final LocalVersionedSerializer<Deps> DEPS_SERIALIZER = localSerializer(CommandSerializers.deps);
+        static final LocalVersionedSerializer<Writes> WRITES_SERIALIZER = localSerializer(CommandSerializers.writes);
+        static final LocalVersionedSerializer<AccordData> RESULT_DATA_SERIALIZER = localSerializer(AccordData.serializer);
+
+        private static <T> LocalVersionedSerializer<T> localSerializer(IVersionedSerializer<T> serializer)
+        {
+            return new LocalVersionedSerializer<>(AccordSerializerVersion.CURRENT, AccordSerializerVersion.serializer, serializer);
+        }
+    }
 
     private static ColumnMetadata getColumn(TableMetadata metadata, String name)
     {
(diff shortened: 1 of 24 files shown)


src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -252,7 +255,7 @@ else if (hasStaticChanges)
             else if (hasRegularChanges)
                 return justRegular;
             else
-                throw new IllegalArgumentException();
+                throw new IllegalArgumentException("CommandsForKey has_modifications=" + commandsForKey.hasModifications() + ", but no Static or Regular columns changed!");
```
dcapwell, line 262: this was useful during debugging, but also has the side benefit if we start calling when hasModification=false we see clearly

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -570,12 +553,12 @@ public static AccordCommand loadCommand(AccordCommandStore commandStore, TxnId t
         return command;
     }
 
-    private static <T> T deserializeWithVersionOr(UntypedResultSet.Row row, String dataColumn, String versionColumn, IVersionedSerializer<T> serializer, Supplier<T> defaultSupplier) throws IOException
+    private static <T> T deserializeWithVersionOr(UntypedResultSet.Row row, String dataColumn, LocalVersionedSerializer<T> serializer, Supplier<T> defaultSupplier) throws IOException
     {
-        if (!row.has(versionColumn))
+        if (!row.has(dataColumn))
```
dcapwell, line 545: This isn't a regression as `null` writes to data column are no longer a null but `version:null`, so the data column will exist where it didn't before.

test/simulator/main/org/apache/cassandra/simulator/SimulationRunner.java
```diff
@@ -376,6 +376,7 @@ protected void run(long seed, B builder) throws IOException
                 catch (Throwable t)
                 {
                     logger().error("Failed on seed 0x{}", Long.toHexString(seed), t);
+                    throw t;
```
dcapwell, line 379: this was added as the simulator JVM would always have rc=0...

test/simulator/main/org/apache/cassandra/simulator/SimulatorUtils.java
```diff
@@ -40,9 +40,10 @@ public static RuntimeException failWithOOM()
     public static void dumpStackTraces(Logger logger)
     {
         Map<Thread, StackTraceElement[]> threadMap = Thread.getAllStackTraces();
-        threadMap.forEach((thread, ste) -> {
-            logger.error("{}:\n   {}", thread, Threads.prettyPrint(ste, false, "   ", "\n", ""));
-        });
+        String prefix = "   ";
+        String delimiter = "\n" + prefix;
```
dcapwell, line 44: `delimiter` doesn't include the `prefix` so to make the logs clean, I had to manually add

test/simulator/main/org/apache/cassandra/simulator/paxos/AccordSimulationRunner.java
```diff
@@ -60,19 +60,27 @@ public static class Help extends HelpCommand<AccordClusterSimulation.Builder> {}
...
+               .withCommand(Help.class)
+               .withDefaultCommand(Help.class)
+               .build()
+               .parse(args)
+               .run(builder);
+            System.exit(0);
+        }
+        catch (Throwable t)
+        {
+            System.exit(1);
```
dcapwell: feel that all simulator runners need this... when an error is throw the `main` thread dies, but not the background non-daemon threads...

test/distributed/org/apache/cassandra/distributed/test/simulator/AccordSimulationTest.java
```diff
@@ -0,0 +1,109 @@
...
+                                  "-s", "30"));
+        ProcessBuilder pb = new ProcessBuilder();
+        pb.environment().put("CLASSPATH", classpath);
+        pb.command(args);
+        Process p = pb.inheritIO().start();
+        boolean completed = p.waitFor(5, TimeUnit.MINUTES);
+        if (!completed)
+            throw new TimeoutException("Timeout waiting on simulation");
+        int rc = p.waitFor();
+        if (rc != 0)
```
dcapwell: heh, without the logs the CI output looks lame

https://app.circleci.com/pipelines/github/dcapwell/cassandra/1665/workflows/65e737da-5269-40c5-a504-5bbd2701c6f1/jobs/13785

test/simulator/main/org/apache/cassandra/simulator/paxos/AccordSimulationRunner.java
dcapwell: Ok this is a problem for current tests: https://app.circleci.com/pipelines/github/dcapwell/cassandra/1666/workflows/7dd47b41-c2ed-4074-a9cd-6f9a8a72a4cc/jobs/13823

Rather than do system exit we need to throw... but at least when I was testing we had daemon threads which blocked the JVM from exiting when there was an issue...

src/java/org/apache/cassandra/service/accord/serializers/AcceptSerializers.java
```diff
@@ -104,22 +104,29 @@ public long serializedSize(Accept.Invalidate invalidate, int version)
         public void serialize(AcceptOk acceptOk, DataOutputPlus out, int version) throws IOException
         {
             CommandSerializers.txnId.serialize(acceptOk.txnId, out, version);
-            CommandSerializers.deps.serialize(acceptOk.deps, out, version);
+            boolean hasDeps = acceptOk.deps != null;
```
bdeggleston: can we use NullableSerializer here?

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -115,12 +115,12 @@
     private static final Logger logger = LoggerFactory.getLogger(AccordKeyspace.class);
 
     public static final String COMMANDS = "commands";
-    public static final String COMMAND_SERIES = "command_series";
     public static final String COMMANDS_FOR_KEY = "commands_for_key";
 
     private static final String TIMESTAMP_TUPLE = "tuple<bigint, bigint, int, bigint>";
     private static final TupleType TIMESTAMP_TYPE = new TupleType(Lists.newArrayList(LongType.instance, LongType.instance, Int32Type.instance, LongType.instance));
     private static final String KEY_TUPLE = "tuple<uuid, blob>";
+    private static final int CURRENT_VERSION = 1;
```
bdeggleston: this should be `MessagingService.current_version`

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -433,31 +424,27 @@ public static Mutation getCommandMutation(AccordCommand command, long timestampM
...
             int nowInSeconds = (int) TimeUnit.MICROSECONDS.toSeconds(timestampMicros);
-            int version = MessagingService.current_version;
-            ByteBuffer versionBytes = accessor.valueOf(version);
+
 
             if (command.status.hasModifications())
                 builder.addCell(live(CommandsColumns.status, timestampMicros, accessor.valueOf(command.status.get().ordinal())));
 
             if (command.homeKey.hasModifications())
             {
```
bdeggleston: we can drop the braces for the blocks that have become single lines

src/java/org/apache/cassandra/service/accord/async/AsyncWriter.java
```diff
@@ -238,7 +240,7 @@ private void denormalize(AccordCommand command, AsyncContext context, Object cal
         }
 
         // There won't be a txn to denormalize against until the command has been preaccepted
-        if (command.status().hasBeen(Status.PreAccepted) && AccordPartialCommand.WithDeps.serializer.needsUpdate(command))
+        if (command.status().hasBeen(Status.PreAccepted) && AccordPartialCommand.WithDeps.serializer.needsUpdate(command) && !(command.txn() == null && command.status().isInvalidated()))
```
bdeggleston, line 243: Not denormalizing an invalidated command may leave it in the uncommitted set, which could lead to incorrect deps being reported. That said, cfk denormalization may no longer be needed, and I'm going to try removing it in another branch, so hang tight

bdeggleston (changes requested):

src/java/org/apache/cassandra/service/accord/serializers/AcceptSerializers.java
dcapwell: ill try to read that and see how to adapt

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
dcapwell: why?  The point was to decouple the storage version from the messaging version; if messaging bumps that doesn't mean storage did

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
dcapwell: pushed, left statements that are single statements but long and spread cross multiple lines; example

```
if (command.storedListeners.hasModifications())
            {
                addStoredSetChanges(builder, CommandsColumns.listeners,
                                    timestampMicros, nowInSeconds, command.storedListeners,
                                    ListenerProxy::identifier);
            }
```

src/java/org/apache/cassandra/service/accord/async/AsyncWriter.java
dcapwell: if we do need to denormalize then we need these messages updated to pass along the `Keys`, as right now we don't have access

bdeggleston: yep makes sense, nevermind

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
bdeggleston: the version used in IVersionedSerializer is intended to be the messaging service version, to support serializer changes in mixed mode clusters, so if we're going to be passing this directly into serializers, it does need to map to a current messaging version. That said, decoupling accord serializers from the messaging service _is_ a good idea. I'm already doing this in AccordPartialCommand.PartialCommandSerializer.Version. Why not promote that enum to AccordSerializerVersion and use it instead? That way you can still pass valid message versions into downstream serializers

dcapwell: yeah good point... the serialized used in the end also work on network, so versions *has* to be network in order to avoid duplication...

>  I'm already doing this in AccordPartialCommand.PartialCommandSerializer.Version. Why not promote that enum to AccordSerializerVersion and use it instead?

so the current value is `VERSION_0(0, MessagingService.current_version);` which has the bug that we redefine if anyone bumps; so should become `VERSION_0(0, MessagingService.VERSION_40);`...

Now, lets say we do promote, then we would need to have something like an AccordSerializer which is a `IVersionedSerializer` but that maps `int version` to `enum AccordSerializerVersion` and delegates to copies of the methods that take the `AccordSerializerVersion`...

One thing also being pointed out here is that `PartialCommandSerializer` and `LocalVersionedSerializer` duplicate logic, so we should unify this.

Ill take a stab at unifying this logic...

dcapwell: ok push new commit 39c27064f085371718495e6fa557ef0ae3ff3f14 which attempts to unify the 2 code paths

bdeggleston (approved):
+1

PR #1949, opened 2022-10-25
CASSANDRA-16491 track and handle errors during nodetool bootstrap resume properly

lmtrombone:
There is some issue where the `nodetool bootstrap resume` command returns success successfully even when some error happens during the operation.  To address that, we now keep track of the errors when they occur in `BootstrapMonitor` and now throw them after signaling.  Also as part of this MR, I made a change so we don't send a `ProgressEventType.COMPLETE` event anymore when we see an error as that seems to be a bit misleading.

To reproduce a scenario, I killed one of the nodes involved in the bootstrap resume process and checked if errors were thrown in the terminal.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 4.2
+ * Nodetool bootstrap resume will now return an error if the operation fails (CASSANDRA-16491)
  * Disable resumable bootstrap by default (CASSANDRA-17679)
  * Include Git SHA in --verbose flag for nodetool version (CASSANDRA-17753)
  * Update Byteman to 4.0.20 and Jacoco to 0.8.8 (CASSANDRA-16413)
--- a/src/java/org/apache/cassandra/tools/BootstrapMonitor.java
+++ b/src/java/org/apache/cassandra/tools/BootstrapMonitor.java
@@ -34,6 +34,7 @@ public class BootstrapMonitor extends JMXNotificationProgressListener
     private final SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss,SSS");
     private final PrintStream out;
     private final Condition condition = newOneTimeCondition();
+    private volatile Exception error;
 
     public BootstrapMonitor(PrintStream out)
     {
@@ -66,7 +67,7 @@ public class BootstrapMonitor extends JMXNotificationProgressListener
     @Override
     public void handleConnectionFailed(long timestamp, String message)
     {
-        Exception error = new IOException(String.format("[%s] JMX connection closed. (%s)",
+        error = new IOException(String.format("[%s] JMX connection closed. (%s)",
                                               format.format(timestamp), message));
         out.println(error.getMessage());
         condition.signalAll();
@@ -82,9 +83,19 @@ public class BootstrapMonitor extends JMXNotificationProgressListener
             message = message + " (progress: " + (int)event.getProgressPercentage() + "%)";
         }
         out.println(message);
+        if (type == ProgressEventType.ERROR)
+        {
+            error = new RuntimeException(String.format("Bootstrap resume has failed with error: %s", message));
+            condition.signalAll();
+        }
         if (type == ProgressEventType.COMPLETE)
         {
             condition.signalAll();
         }
+    }
+
+    public Exception getError()
+    {
+        return error;
     }
 }
--- a/src/java/org/apache/cassandra/tools/NodeProbe.java
+++ b/src/java/org/apache/cassandra/tools/NodeProbe.java
@@ -2005,6 +2005,8 @@ public class NodeProbe implements AutoCloseable
             {
                 out.println("Resuming bootstrap");
                 monitor.awaitCompletion();
+                if (monitor.getError() != null)
+                    throw monitor.getError();
             }
             else
             {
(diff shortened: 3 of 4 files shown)


src/java/org/apache/cassandra/tools/BootstrapMonitor.java
```diff
@@ -34,6 +34,7 @@ public class BootstrapMonitor extends JMXNotificationProgressListener
     private final SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss,SSS");
     private final PrintStream out;
     private final Condition condition = newOneTimeCondition();
+    private Exception error;
```
maedhroz: I think this is going to need to be `volatile`, as it's written by one thread and accessed by another.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2074,7 +2074,6 @@ public void onFailure(Throwable e)
                     }
                     logger.error(message, e);
                     progressSupport.progress("bootstrap", new ProgressEvent(ProgressEventType.ERROR, 1, 1, message));
-                    progressSupport.progress("bootstrap", new ProgressEvent(ProgressEventType.COMPLETE, 1, 1, "Resume bootstrap complete"));
```
maedhroz: The thing that worries me about this is that it will hang older versions of `nodetool` when an error occurs. They will be waiting for a `COMPLETE` that will never arrive after printing the error message. The fact that we're changing the behavior of `nodetool` itself on error in this Jira probably means that we only need to patch trunk, but it's probably not worth it to remove the `COMPLETE` here.

src/java/org/apache/cassandra/tools/NodeProbe.java
```diff
@@ -1992,6 +1992,8 @@ public void resumeBootstrap(PrintStream out) throws IOException
             {
                 out.println("Resuming bootstrap");
                 monitor.awaitCompletion();
+                if (monitor.getError() != null)
+                    throw monitor.getError();
```
maedhroz, line 2009: This is obviously going to change the return contract for `resume`, but it's really just a bug fix. I don't see it requiring a `NEWS.txt` entry. (Just `CHANGES.txt`.)

maedhroz (changes requested):
I've left a couple comments inline we'll want to address. Outside those, the only other thing missing is a test of some kind. Luckily, I think there's a place we can basically insert a one-liner that only works w/ your patch in place...

In `BootstrapBinaryDisabledTest`, we use `nodetoolResult("bootstrap", "resume")` on line 112 to finish a stopped bootstrap after disabling the Byteman rule that would fail it. All we need to do is change...

```
RewriteEnabled.disable();
node.nodetoolResult("bootstrap", "resume").asserts().success();
```
...to...
```
node.nodetoolResult("bootstrap", "resume").asserts().failure();
RewriteEnabled.disable();
node.nodetoolResult("bootstrap", "resume").asserts().success();
```
...which will only work w/ you patch making sure the return code is correct.

src/java/org/apache/cassandra/tools/BootstrapMonitor.java
```diff
@@ -82,9 +83,19 @@ public void progress(String tag, ProgressEvent event)
             message = message + " (progress: " + (int)event.getProgressPercentage() + "%)";
         }
         out.println(message);
+        if (type == ProgressEventType.ERROR)
+        {
+            error = new RuntimeException(String.format("Bootstrap resume has failed with error: %s", message));
+            condition.signalAll();
```
lmtrombone, line 89: @maedhroz Since I reverted my changes where I removed the `ProgressEventType.COMPLETE` event that was being sent, I assume we can just rely on the logic below (lines 91-94) where it already calls `condition.signalAll();`?   When we encounter an issue, I noticed that we always both `ProgressEventType.ERROR` and `ProgressEventType.COMPLETE` events so it seems a little redundant and unnecessary now.

CHANGES.txt
```diff
@@ -68,6 +68,7 @@
  * Add new CQL function maxWritetime (CASSANDRA-17425)
  * Add guardrail for ALTER TABLE ADD / DROP / REMOVE column operations (CASSANDRA-17495)
  * Rename DisableFlag class to EnableFlag on guardrails (CASSANDRA-17544)
+ * Nodetool bootstrap resume will now return an error if the operation fails (CASSANDRA-16491)
```
maedhroz: Committer can fix this up on commit too, but make sure CHANGES bits go a the top of the appropriate section, not the bottom. (For a hint, look at the date progression of the `git blame` annotation in your IDE, etc.)

maedhroz (approved):
LGTM w/ latest changes

src/java/org/apache/cassandra/tools/BootstrapMonitor.java
maedhroz: I definitely wouldn't do it this way from scratch, but it's probably just not worth the compatibility headache. What we have now LGTM

maedhroz:
https://app.circleci.com/pipelines/github/maedhroz/cassandra?branch=CASSANDRA-16491-circle

maedhroz:
Committed in https://github.com/apache/cassandra/commit/8ec04361b9e098430023e4776baf1941be958475

pmcfadin:
Closing as [CASSANDRA-16491](https://issues.apache.org/jira/browse/CASSANDRA-16491) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #1950, opened 2022-10-26
WIP:  Initial implementation of cassandra-conf with nodetool example - CASSANDRA-17773

Claudenw:
This is a work in progress I am looking for comments

This change does 4 things:

1. Moves the standard boiler plate from the standard scripts into a single maintainable script to be sourced in the original scripts.
2. Creates a debug logging so that problem determination (as it CASSANDRA-17773) is easier.
3. Has code to preserve environment variables when needed (c.f. nodetool script).
4. Provides a verification method that will verify that all standard environment variables are set.

In practice this is a simple 2 line replacement for most boiler plate blocks.

```

patch by Claude Warren; reviewed by <Reviewers> for CASSANDRA-17773

```


Claudenw:
@mmuzaf You may be interested in this change.

bin/cassandra-conf.sh
```diff
@@ -0,0 +1,168 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# A script to include the cassandra.in.sh file, set NUMACTL, and load 
+#          $CASSANDRA_CONF/cassandra-env.sh"
+#
+# Creates a validate_env() function to verify criticla environment vars are set
```
smiklosovic: "_creates_"?

criticla -> critical

bin/cassandra-conf.sh
```diff
@@ -0,0 +1,168 @@
...
+# and display them if desired
+
+## Environment variables:
+# ENV_DEBUG: Show all var settings during the validate_env() call.
```
smiklosovic: maybe having them prefixed so it would be like `CASSANDRA_ENV_DEBUG` and similar is better? A variable called "ENV_SHOW" is quite non-telling if I just execute `$ env` command. I do not know what software that environment variable relates to.

bin/cassandra-conf.sh
```diff
@@ -0,0 +1,168 @@
...
+##
+## Validates that required environment variables are set. 
+## Required Environment Vars:
+##      CASSANDRA_INCLUDE
+##      CASSANDRA_HOME
+##      CASSANDRA_LOG_DIR
+## 
+## Environment Vars:
+##      ENV_DEBUG   Enables display of variables.
+##      ENV_SHOW    List of additional variables to display.
```
smiklosovic: `ENV_PRESERVE` is missing

bin/cassandra-conf.sh
```diff
@@ -0,0 +1,168 @@
...
+        echo "ENVIRONMENT"
+        echo "-----------"
+        echo "PWD=$PWD"
+        echo "CASSANDRA_CONF=$CASSANDRA_CONF"
+    fi
+    
+    if [ "x$CASSANDRA_INCLUDE" = "x" ]; then
+        echo ">>> cassandra.in.sh not found or CASSANDRA_INCLUDE not set" >&2
+        retval=1
+    else
```
smiklosovic, line 57: does not `sh` have something like `elif` ? So you can make this less verbose and save some lines.

bin/cassandra-conf.sh
```diff
@@ -0,0 +1,168 @@
...
+    if [ "x$CASSANDRA_LOG_DIR" = "x" ]; then
+        echo ">>> CASSANDRA_LOG_DIR not set" >&2
+        retval=1
+    else
+        if [ -n "$ENV_DEBUG" ] ; then
+            echo "CASSANDRA_LOG_DIR=$CASSANDRA_LOG_DIR"
+        fi
+    fi
+
+    if [ -n "$ENV_DEBUG" ] ; then
```
smiklosovic: should not this be explicitly equal to `1` or `yes` ?

bin/cassandra-conf.sh
```diff
@@ -0,0 +1,168 @@
...
+        if [ -z $NUMACTL ]; then
+            echo "numactl not found"
```
smiklosovic, line 83: `echo ">>> numactl not found"`

bin/cassandra-conf.sh
```diff
@@ -0,0 +1,168 @@
...
+        fi
+
+        while [ -n "$1" ] ; do
+            eval echo "$1=\$$1"
+            shift
+        done
+        
+        if [ -n "$ENV_SHOW" ]
```
smiklosovic: should not this be explicitly equal to `1` or `yes` ?

bin/cassandra-conf.sh
```diff
@@ -0,0 +1,168 @@
...
+        fi
+    fi
+
+    if [ -n "$ENV_DEBUG" ] ; then
+        if [ -z $NUMACTL ]; then
+            echo "numactl not found"
+        fi
+
+        while [ -n "$1" ] ; do
+            eval echo "$1=\$$1"
```
smiklosovic, line 87: what about printing these variables too? Not sure if it is secure as it may potentially contain credentials but I dont think vanilla Cassandra does that

bin/cassandra-conf.sh
```diff
@@ -0,0 +1,168 @@
...
+            shift
+        done
+        
+        if [ -n "$ENV_SHOW" ]
+        then
```
smiklosovic: once you put `then` after `if` at the same line, next time it is on new line like here, it should be consistent.

bin/nodetool
```diff
@@ -98,6 +71,8 @@ if [ "x$MAX_HEAP_SIZE" = "x" ]; then
     MAX_HEAP_SIZE="128m"
 fi
 
+validate_env JMX_PORT
```
smiklosovic, line 75: could you please remind me why we are checking just JMX_PORT here? Seems to be quite random but I guess it has some logic in it?

bin/cassandra-conf.sh
Claudenw: changed to define and fixed typo.

bin/cassandra-conf.sh
Claudenw: Added CASSANDRA_ before all ENV_ e.g. CASSANDRA_ENV_PRESERVE

bin/cassandra-conf.sh
Claudenw: added

bin/cassandra-conf.sh
Claudenw: It could, but this solution lists all the missing symbols in one run.

bin/cassandra-conf.sh
Claudenw: The documentation says if it is set.  Doesn't say anything about value.  I suppose it could be changed.

bin/cassandra-conf.sh
Claudenw: This is not an error, it is a message that only is displayed while debugging.  Not having it set was not an error before.

bin/cassandra-conf.sh
Claudenw: The CASSANDRA_ENV_SHOW is a list of variables to show so if it is set then we use it, if it is not set we don't display anything,

bin/cassandra-conf.sh
Claudenw: fixed

bin/nodetool
Claudenw: It is not checking just the JMX_PORT, however when debugging the JMX_PORT will be displayed, this is added because the JMX_PORT can be set above.  

Added a comment reflecting this.

Claudenw (commented):
@smiklosovic I updated and commented.

bin/cassandra-conf.sh
smiklosovic: I understand, but I think "to be set" is not enough. When somebody sets it to "fals" when they meant "false" and they made typo, this will evaluate it like "it is set" and it will be shown .... I am for explicitly setting this stuff every time.

bin/cassandra-conf.sh
```diff
@@ -0,0 +1,167 @@
...
+##      CASSANDRA_LOG_DIR
+## 
+## Environment Vars:
+##      CASSANDRA_ENV_DEBUG   Enables display of variables if this var is set.
+##      CASSANDRA_ENV_SHOW    List of additional variables to display.
+##
+validate_env() {
+    retval=0
+    if [ -n "$CASSANDRA_ENV_DEBUG" ] ; then
+        echo ""
```
smiklosovic: @Claudenw I dont think this is necessary. Same empty line after the closing "------" is redundant.

bin/cassandra-conf.sh
```diff
@@ -0,0 +1,167 @@
+#!/bin/sh -x
```
smiklosovic: @Claudenw is "-x" necessary?

bin/nodetool
```diff
@@ -22,35 +22,8 @@ if [ "`basename "$0"`" = 'nodeprobe' ]; then
...
-# Run cassandra-env.sh to pick up JMX_PORT
-if [ -f "$CASSANDRA_CONF/cassandra-env.sh" ]; then
-    JVM_OPTS_SAVE=$JVM_OPTS
-    MAX_HEAP_SIZE_SAVE=$MAX_HEAP_SIZE
-    . "$CASSANDRA_CONF/cassandra-env.sh"
-    MAX_HEAP_SIZE=$MAX_HEAP_SIZE_SAVE
-    JVM_OPTS="$JVM_OPTS_SAVE"
-fi
+ENV_PRESERVE="$ENV_PRESERVE MAX_HEAP_SIZE JVM_OPTS"
+. ./cassandra-conf.sh
```
smiklosovic:
````
if [ -z $CASSANDRA_HOME ]; then
  . $(dirname "$0")/../conf/cassandra-conf.sh
else
  . $CASSANDRA_HOME/conf/cassandra-conf.sh
fi
````

This should enable people to execute it without having `CASSANDRA_HOME` set. I think nobody is going to move nodetool script away from its current location so it is given that conf file will be in `../conf/cassanra-conf.sh` from `nodetool`'s perspective.

Comment (smiklosovic) [reply bin/nodetool]: This should also somehow honor `CASSANDRA_CONF` property if it is set.

File: bin/cassandra-conf.sh (hunk 12 above)
Comment (Claudenw) [reply bin/cassandra-conf.sh]: no, it is for testing.  It should be removed before merge.

File: bin/cassandra-conf.sh (hunk 11 above)
Comment (Claudenw) [reply bin/cassandra-conf.sh]: It makes a clear separation in the logs

File: bin/nodetool (hunk 13 above)
Comment (Claudenw) [reply bin/nodetool]: CASSANDRA_CONF is an existing environment variable, it is used to load cassendra-env.sh.  which I believe is a user editable file for configurations.   `cassandra-conf.sh` now strikes me as a very bad name as it is not really associated with CASSANDRA_CONF.  `cassandra-conf.sh`can not be dependant upon CASSANDRA_CONF as CASSANDRA_CONF may be set using `cassandra.in.sh` or whatever CASSANDRA_INCLUDE points to.

The line in question here points to the cassandra-conf.sh which is in the same directory as nodetool.

File: bin/nodetool [hunk 14]
```diff
@@ -22,35 +22,8 @@ if [ "`basename "$0"`" = 'nodeprobe' ]; then
...
-
-# Run cassandra-env.sh to pick up JMX_PORT
-if [ -f "$CASSANDRA_CONF/cassandra-env.sh" ]; then
-    JVM_OPTS_SAVE=$JVM_OPTS
-    MAX_HEAP_SIZE_SAVE=$MAX_HEAP_SIZE
-    . "$CASSANDRA_CONF/cassandra-env.sh"
-    MAX_HEAP_SIZE=$MAX_HEAP_SIZE_SAVE
-    JVM_OPTS="$JVM_OPTS_SAVE"
-fi
+ENV_PRESERVE="$ENV_PRESERVE MAX_HEAP_SIZE JVM_OPTS"
```
Claudenw: This should read `CASSANDRA_ENV_PRESERVE="$CASSANDRA_ENV_PRESERVE MAX_HEAP_SIZE JVM_OPTS"

bin/cassandra-conf.sh
smiklosovic: Dont we have "-----------" for separation already? What do we have "--------" for if we still need to separate it by an empty line?

bin/nodetool
smiklosovic: I tend to disagree how this is implemented but I do need to test this manually to see how this really behaves.

Claudenw: `cassandra_conf.sh` pulls the common code from all the existing scripts.  This is found at:
https://github.com/apache/cassandra/pull/1950/files#diff-61e866313476b2b21fa3d88d315567447c43de4f55066776bc649d9f898b3bf6R105-R140
also lines 151-153 and 165-167

nodetool then has special requirements to unset some vars so it needs to preserve and reset some around the cassandra-env.sh call.

the script `cassandra.in.sh` 

- can move and is located by various means.
- sets CASSANDRA_HOME and CASSANDRA_CONF if not set.
- sets JAVA_AGENT, CLASSPATH, JVM_OPTS, JVM_DEP_OPTS, and JAVA_VERSION
- is a file that may be edited by users to change the configuration.

so `cassandra_conf.sh` should not move.  It should always be in the same directory with the script(s) loading it.

any script using `cassandra_conf.sh` will have the lines
```. ./cassandra-conf.sh
validate_env
```

some, like `nodetool` my preserve vars and so set `CASSANDRA_ENV_PRESERVE` with a list of variables to preserve.

some like `nodetool` may set additional vars  for which debugging display is appropriate.  Those will list the additional env vars after the `validate_env` as `nodetool` does

```
validate_env JMX_PORT
```

Simply put `cassandra_conf.sh` removes the boilerplate from the scripts to create one place where all the configuration determination is performed, it has hooks to support current operations (like `nodetool`), it adds debugging of the environment variables to make problem determination easier and consistent across all the script based tooling.

I hope this overview helps with your testing.

smiklosovic: Imagine you are developing some 3rd party command-line tool which needs the very same boilerplate as any other command we ship in Cassandra distribution, in `tools` or `bin` or what have you.

While implementing the script which executes that main class of yours, you basically take some already existing script as a skeleton and then you tweak it in such a way that the java command at the end of that file executes your main class.

When we eventually plan to incorporate this `cassandra-conf.sh` stuff into every script, one can imagine that the very same script will be sourced in our custom script as well.

Here, we are making an assumption that `cassandra-conf.sh` will be in the same directory as `nodetool` script is.

But what if it is not? What if you have some 3rd party script which is located it a different directory from nodetool's one? What if that script is in `/tmp/mycustomscripts/mytool` ? Then sourcing `. cassandra-conf.sh` will not work, will it? 

So for that reason, the loading of `. cassandra-conf.sh` has to be a little bit smarter about that and we should source it in such a way that the script it is sourced in can be in theory wherever. That is what my suggestion about sourcing with above is about.

Claudenw: since `cassandra-conf.sh` is intended to be used "as is" without modification I think it should be referenced in-situ.  If one is writing a new script to source `cassandra-conf.sh` then the full path should be provided.  Any calculation of the location of `cassandra-conf.sh` is, in my mind, risky and potentially hard to diagnose when an error occurs. 

The text:
```
if [ -z $CASSANDRA_HOME ]; then
  . $(dirname "$0")/../conf/cassandra-conf.sh
else
  . $CASSANDRA_HOME/conf/cassandra-conf.sh
fi
```
won't work because CASSANDRA_HOME is not expected to be set until after `cassandra.in.sh` which is loaded by `cassandra-conf.sh`, also the `cassandra-conf.sh` file is expected to be in `bin/` not `conf/`.  But even if `bin/` were used it it would still be incorrect for any new script that was not in a sibling directory of `bin/`.

smiklosovic: _won't work because CASSANDRA_HOME is not expected to be set until after cassandra.in.sh which is loaded by cassandra-conf.sh_

Is this really the case? I am pretty much used to have the environment property `CASSANDRA_HOME` set way before I start Cassandra / execute any script. It is set in my environment variables in .bashrc / .bash_profile or so ... 

We would just move this

https://github.com/apache/cassandra/blob/trunk/bin/cassandra.in.sh#L17-L19

to the script itself. That logic just sets CASSANDRA_HOME when it is not set. But if I set it before I execute any script, it will take it into account.

It would not hurt if we left it in `cassandra.in.sh` anyway. It would just do nothing if we happen to set it in the script itself already.

Claudenw: One of the goals of this change is to move the boiler plate to a single location and call it.  It should not require a change is step order of an existing file (e.g. defining CASSANDRA_HOME earlier or even explicitly)

If someone writing a new script wants to source `cassandra_conf.sh` from a known CASSANDRA_HOME that is fine, but I don't think we should require it of the existing scripts that do not currently set CASSANDRA_HOME and that only check the value of CASSANDRA_HOME after `cassandra.in.sh` is called.

michaelsembwever:
Some questions…

1. isn't this a wrapper to either conf/cassandra-env.sh or bin/cassandra.in.sh ? the name cassandra-conf.sh is confusing…

2. scripts should be silent when successful, by default

3. you can't presume the file is in the same directory, or where anything is being executed from, 

wrt (1), couldn't we instead pull a cassandra-env.conf out of cassandra-env.sh so to split the logic and the user-editable variables into separate files …?

smiklosovic:
I fully agree with the third point, @michaelsembwever . That is what I was trying to communicate as well. I think that Claude reached the same conclusion about 1) and I dont have any objections with 2), definitely it should be silent when executed just fine to follow basic *nix principles. 

Not sure about your last question. Maybe Claude can try to model that, just for the sake of seeing how that would look like.

Claudenw:
@michaelsembwever 

1. I agree the name is confusing, does anyone have a better name?  The goal of this change is to simply remove the boilerplate from the other scripts while providing debugging capabilities to more easily detect the issue that caused CASSANDRA-17773.  I did not make any change to operation of the scripts.  Pulling a `cassandra-env.conf` out of `cassandra-env.sh` may be a reasonable thing to do but is outside the scope of this change.
2. Script is silent by default.  I think you are referring to the `-x` argument.  That was included in error and was a debugging switch.
3. This script has to run before before we can assume any environment variables are set as it sets many of them through the logic that pulls in the cassandra.in.sh script.  I make 2 assumptions for Cassandra provided scripts.  
    * The scripts in `'bin/` are not to be edited by end users
    * The scripts provided by Cassandra are all deployed in `bin/`.  Thus the `cassandra-conf.sh` will be in the same directory as `nodetool`.  However, I realise that there was an error in the way the file was included.  The code in nodetool now uses `realpath` and `dirname` to figure out where `cassandra-conf.sh` is.  This means that symbolic links to `nodetool` will work correctly.  (other cassandra provided scripts in `bin/` will be configured the same way)

For user created scripts I assume they know where `bin/` is and will modify their scripts to source `cassandra-conf.sh` from there.

smiklosovic:
I yet again want to reiterate the fact that environment property `CASSANDRA_HOME` might be set in user's environment and it is in fact set way before any script is executed so the third point is not entirely true.

Claudenw:
If CASSANDRA_HOME were set in every case then there would be no need for `cassandra.in.sh`  to set it, and yet it does.  
`cassandra-conf.sh` can not depend upon CASSANDRA_HOME being set.

In fact looking at an unedited `sstableutil` code you can see that CASSANDRA_HOME is not used and may be set by `cassandra.in.sh` 

https://github.com/apache/cassandra/blob/98a055e51c8eb6b788d3d0fe94f3d2576ab14d39/bin/sstableutil#L19-L33

Claudenw:
@smiklosovic , @michaelsembwever  any further comments on this?

Claudenw:
@smiklosovic , @michaelsembwever any further comments on this or should I go ahead an perform the changes on all the scripts?

michaelsembwever:
1)
where did we get with this? @smiklosovic seems to say you agree, while you say it is out of scope ? 

3) 
If `CASSANDRA_HOME` is set it must be honoured.

One example I know of, it needs to be possible to execute scripts in a random place against a CASSANDRA_HOME somewhere else. (tests do this)

--
otherwise… my concerns with this work is that we it requires a lot of eyes and testing, and that we won't get that before the August freeze.

smiklosovic:
@Claudenw I would rather focus on this PR bellow, we need to be smart about time in the context of incoming freeze. That one is a new feature, this is "just" an improvement. AFAIK new features in a freeze window are forbidden.

https://github.com/apache/cassandra/pull/2282

PR #1951, opened 2022-10-26
Partial replication

belliottsmith:


src/java/org/apache/cassandra/service/accord/async/AsyncWriter.java
```diff
@@ -201,7 +201,7 @@ AccordState<K> getForDenormalization(K key,
             return item;
 
         item = cache.getOrNull(key);
-        if (item != null)
+        if (item != null && !cache.hasLoadFuture(key))
```
aweisberg, line 202: Is this necessary to ensure we don't write to something that will end up being replaced when the cache loads?

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -203,7 +203,10 @@ public Command ifLoaded(TxnId txnId)
     {
         AccordCommand command = commandCache.getOrNull(txnId);
         if (command != null && command.isLoaded())
+        {
+            getContext().commands.add(command);
```
aweisberg, line 208: What issue is this fixing? The context otherwise wouldn't be populated along with the cache?

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -251,15 +284,51 @@ public void processBlocking(Runnable runnable)
...
+    }
+
+    @Override
+    public long latestEpoch()
+    {
+        return time.epoch();
+    }
+
+    @Override
+    public Timestamp maxConflict(Keys keys)
```
aweisberg, line 324: Isn't this identical to the function in InMemoryCommandStore? Does it need to be duplicated? A lot of these look identical to InMemoryCommandStore

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -692,7 +691,7 @@ private static DecoratedKey makeKey(AccordCommandsForKey cfk)
         return makeKey(cfk.commandStore(), cfk.key());
     }
 
-    public static Mutation getCommandsForKeyMutation(AccordCommandsForKey cfk, long timestampMicros)
+    public static Mutation getCommandsForKeyMutation(AccordCommandStore commandStore, AccordCommandsForKey cfk, long timestampMicros)
```
aweisberg, line 658: Unused?

src/java/org/apache/cassandra/service/accord/AccordCommand.java
```diff
@@ -633,40 +652,42 @@ private Future<Void> applyWithCorrectScope()
...
+    public Future<Data> read(SafeCommandStore safeStore)
     {
-        ReadFuture future = cache().getReadFuture(txnId);
+        AccordStateCache.Instance<TxnId, AccordCommand> cache = ((AccordCommandStore) safeStore).commandCache();
+        Future<Data> future = cache.getReadFuture(txnId);
         if (future != null)
-            return future.scope.equals(scope) ? future : super.read(scope);
-        future = new ReadFuture(scope, super.read(scope));
-        cache().setReadFuture(txnId, future);
+            return future;
```
aweisberg, line 690: Why can the scope check be omitted now?

src/java/org/apache/cassandra/db/marshal/ValueAccessor.java
```diff
@@ -315,6 +315,24 @@ default boolean getBoolean(V value, int offset)
     long toLong(V value);
     /** returns a long from offset {@param offset} */
     long getLong(V value, int offset);
+
+    default long getUpTo8Bytes(V value, int offset, int count)
```
aweisberg: Unit test? Is this endian agnostic or does it matter? I am a bit rusty.

src/java/org/apache/cassandra/db/marshal/ValueAccessor.java
```diff
@@ -315,6 +315,24 @@ default boolean getBoolean(V value, int offset)
...
+        {
+            default: throw new IllegalArgumentException();
+            case 0: return 0;
+            case 1: return getByte(value, offset);
+            case 2: return getShort(value, offset);
+            case 3: return (getShort(value, offset) << 8) | getByte(value, offset + 2);
+            case 4: return getInt(value, offset);
+            case 5: return ((long)getInt(value, offset) << 8) | getByte(value, offset + 4);
+            case 6: return ((long)getInt(value, offset) << 16) | getShort(value, offset + 4);
+            case 7: return ((long)getInt(value, offset) << 24) | (getShort(value, offset + 4) << 8) | getByte(value, offset + 5);
```
aweisberg: This gets two bytes from offset + 4 then one byte from offset + 5 which overlaps

src/java/org/apache/cassandra/dht/ByteOrderedPartitioner.java
```diff
@@ -144,8 +144,51 @@ public double size(Token next)
...
+            return new BytesToken(newToken);
+        }
+
+        @Override
+        public Token decreaseSlightly()
+        {
+            if (token.length == 0)
+                throw new IndexOutOfBoundsException("Cannot create a smaller token the MINIMUM");
+
+            // find first byte we can increment
```
aweisberg:
```suggestion
            // find first byte we can decrement
```

src/java/org/apache/cassandra/dht/RandomPartitioner.java
```diff
@@ -276,6 +276,14 @@ public Token increaseSlightly()
             return new BigIntegerToken(token.add(BigInteger.ONE));
         }
 
+        @Override
+        public Token decreaseSlightly()
+        {
+            if (token.equals(MINIMUM.token))
```
aweisberg, line 284: Increase doesn't have this protection and there is a maximum value

src/java/org/apache/cassandra/service/accord/AccordCommand.java
```diff
@@ -116,53 +119,52 @@ boolean isReadOnly()
...
-        executeAt = new StoredValue<>(kind());
-        deps = new StoredValue<>(kind());
-        writes = new StoredValue<>(kind());
-        result = new StoredValue<>(kind());
-        status = new StoredValue.HistoryPreserving<>(kind());
-        isGloballyPersistent = new StoredBoolean(kind());
-        waitingOnCommit = new StoredNavigableMap<>(kind());
-        waitingOnApply = new StoredNavigableMap<>(kind());
-        storedListeners = new StoredSet.DeterministicIdentity<>(kind());
+        homeKey = new StoredValue<>(rw());
```
aweisberg, line 140: Why hardcoding to `rw()`?

src/java/org/apache/cassandra/service/accord/AccordCommand.java
```diff
@@ -524,22 +535,29 @@ public Timestamp executeAt()
     }
 
     @Override
-    public void executeAt(Timestamp timestamp)
+    public Txn.Kind kind()
+    {
+        PartialTxn txn = partialTxn.get();
+        return txn == null ? null : txn.kind();
```
aweisberg: Huh, this is allowed to return `null`? Nullable annotation?

src/java/org/apache/cassandra/service/accord/AccordCommandStores.java
```diff
@@ -23,36 +23,45 @@
...
     }
 
-    @Override
-    protected CommandStore createCommandStore(int generation, int index, int numShards, Node node, Agent agent, DataStore store, ProgressLog.Factory progressLogFactory, CommandStore.RangesForEpoch rangesForEpoch)
+    private AccordCommandStores(int numShards, NodeTimeService time, Agent agent, DataStore store,
+                                ProgressLog.Factory progressLogFactory, ExecutorService[] executors)
+    {
+        super(numShards, time, agent, store, progressLogFactory,
+              (id, generation, index, numShards1, node1, agent1, store1, progressLogFactory1, rangesForEpoch)
+                -> new AccordCommandStore(id, generation, index, numShards1, node1, agent1, store1, progressLogFactory1, rangesForEpoch, executors[index]));
```
aweisberg: Should some of these parameters be bound into the lambda instead of being passed to the factory if they already known at this point?

src/java/org/apache/cassandra/service/accord/AccordCommandsForKey.java
```diff
@@ -364,13 +374,13 @@ public void updateSummaries(AccordCommand command)
             if (command.status.previous() == null || !command.status.previous().hasBeen(Status.Committed))
                 uncommitted.map.blindRemove(command.txnId());
 
-            ByteBuffer bb = AccordPartialCommand.WithDeps.serializer.serialize(command);
-            committedById.map.blindPut(command.txnId(), bb);
-            committedByExecuteAt.map.blindPut(command.executeAt(), bb);
+            ByteBuffer partialCommand = AccordPartialCommand.serializer.serialize(key, command);
+            committedById.map.blindPut(command.txnId(), partialCommand);
```
aweisberg, line 378: Will sharing the same command twice confuse memory accounting?

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -312,20 +320,20 @@ private static Map<ByteBuffer, ByteBuffer> serializeWaitingOn(Map<TxnId, ByteBuf
...
-        NavigableMap<TxnId, ByteBuffer> result = new TreeMap<>();
+        NavigableMap<Timestamp, TxnId> result = new TreeMap<>();
         for (Map.Entry<ByteBuffer, ByteBuffer> entry : serialized.entrySet())
-            result.put(deserializeTimestampOrNull(entry.getKey(), TxnId::new), entry.getValue());
+            result.put(deserializeTimestampOrNull(entry.getKey(), Timestamp::new), deserializeTimestampOrNull(entry.getValue(), TxnId::new));
         return result;
     }
 
-    private static NavigableMap<TxnId, ByteBuffer> deserializeWaitingOn(UntypedResultSet.Row row, String name)
+    private static NavigableMap<Timestamp, TxnId> deserializeWaitingOnApply(UntypedResultSet.Row row, String name)
```
aweisberg: Do we need the wrapper method if it is the only caller?

src/java/org/apache/cassandra/service/accord/AccordObjectSizes.java
```diff
@@ -42,6 +52,27 @@ public static long key(Key key)
...
+    {
+        return EMPTY_KEY_RANGE_SIZE + key(range.start()) + key(range.end());
+    }
+
+    private static final long EMPTY_KEY_RANGES_SIZE = ObjectSizes.measure(KeyRanges.of());
+    public static long ranges(KeyRanges ranges)
+    {
+        long size = EMPTY_KEY_RANGES_SIZE;
+        size += ObjectSizes.sizeOfReferenceArray(ranges.size());
+        for (int i = 0, mi = ranges.size() ; i < mi ; i++)
```
aweisberg, line 72: Worth/possible to optimize for fixed size key ranges where this can just be multiplication?

src/java/org/apache/cassandra/service/accord/AccordPartialCommand.java
```diff
@@ -244,26 +129,23 @@ public int hashCode()
...
-                        throw new IllegalArgumentException();
-                }
+                if (b == 0)
+                    return VERSION_0;
+                throw new IllegalArgumentException();
             }
         }
 
-        public void serialize(T command, DataOutputPlus out, Version version) throws IOException
+        public void serialize(AccordPartialCommand command, DataOutputPlus out, Version version) throws IOException
```
aweisberg: Could be private

src/java/org/apache/cassandra/service/accord/AccordPartialCommand.java
```diff
@@ -244,26 +129,23 @@ public int hashCode()
...
-            CommandSerializers.status.serialize(command.status(), out, version.msg_version);
-            serializeNullable(command.txn(), out, version.msg_version, CommandSerializers.txn);
             serializeNullable(command.executeAt(), out, version.msg_version, CommandSerializers.timestamp);
+            CommandSerializers.status.serialize(command.status(), out, version.msg_version);
+            serializeNullable(command.kind(), out, version.msg_version, CommandSerializers.kind);
+            CollectionSerializer.serializeCollection(CommandSerializers.txnId, command.deps, out, version.msg_version);
         }
 
-        public ByteBuffer serialize(T command)
+        public ByteBuffer serialize(AccordPartialCommand command)
```
aweisberg, line 125: Could be private

src/java/org/apache/cassandra/service/accord/TokenRange.java
```diff
@@ -20,33 +20,21 @@
...
 import org.apache.cassandra.io.util.DataInputPlus;
 import org.apache.cassandra.io.util.DataOutputPlus;
 import org.apache.cassandra.schema.TableId;
 import org.apache.cassandra.service.accord.api.AccordKey;
-import org.apache.cassandra.service.accord.api.AccordKey.SentinelKey;
-import org.apache.cassandra.service.accord.api.AccordKey.TokenKey;
+import org.apache.cassandra.service.accord.api.AccordRoutingKey;
+import org.apache.cassandra.service.accord.api.AccordRoutingKey.SentinelKey;
 
 public class TokenRange extends KeyRange.EndInclusive
```
aweisberg, line 33: Should be called AccordTokenRange since there is also one in C*?

src/java/org/apache/cassandra/service/accord/api/AccordKey.java
```diff
@@ -229,37 +79,38 @@ public boolean equals(Object o)
             if (this == o) return true;
```
aweisberg: Remove this equals method because the super class has it?

src/java/org/apache/cassandra/service/accord/async/AsyncWriter.java
```diff
@@ -226,21 +227,38 @@ private void denormalize(AccordCommand command, AsyncContext context, Object cal
...
+                }
+            }, cmd -> cmd.blockingApplyOn);
+        }
 
         if (command.shouldUpdateDenormalizedWaitingOn())
         {
-            ByteBuffer summary = AccordPartialCommand.serializer.serialize(command);
-            denormalizeWaitingOnSummaries(command, context, summary, cmd -> cmd.waitingOnCommit, cmd -> cmd.blockingCommitOn);
-            denormalizeWaitingOnSummaries(command, context, summary, cmd -> cmd.waitingOnApply, cmd -> cmd.blockingApplyOn);
+            denormalizeWaitingOnSummaries(command, context, cmd -> (txnId, ignore) -> cmd.waitingOnCommit.blindAdd(txnId), cmd -> cmd.blockingCommitOn);
```
aweisberg, line 252: I don't even know what it means to have multiple `->` chained. A lambda that returns a lambda?

src/java/org/apache/cassandra/service/accord/db/AbstractKeyIndexed.java
```diff
@@ -89,99 +113,126 @@ public boolean equals(Object o)
...
 
     public long estimatedSizeOnHeap()
     {
         long size = emptySizeOnHeap();
-        for (Map.Entry<PartitionKey, ByteBuffer> entry : serialized.entrySet())
-        {
-            size += entry.getKey().estimatedSizeOnHeap();
-            size += ByteBufferUtil.EMPTY_SIZE_ON_HEAP + ByteBufferAccessor.instance.size(entry.getValue());
-        }
+        for (Key key : keys) size += ((PartitionKey) key).estimatedSizeOnHeap();
```
aweisberg: Using something from `AccordObjectSizes`?

src/java/org/apache/cassandra/service/accord/db/AbstractKeyIndexed.java
```diff
@@ -89,99 +113,126 @@ public boolean equals(Object o)
...
+        for (ByteBuffer buffer : serialized) size += ByteBufferUtil.EMPTY_SIZE_ON_HEAP + ByteBufferAccessor.instance.size(buffer);
```
aweisberg: `ObjectSizes.sizeOnHeapOf`
Also I think this is missing the reference array?

I think there may even be a helper for size of an array of ByteBuffers.

src/java/org/apache/cassandra/service/accord/db/AccordData.java
```diff
@@ -47,35 +49,50 @@
...
-    private static final long EMPTY_SIZE = ObjectSizes.measureDeep(new AccordData());
+    private static final long EMPTY_SIZE = ObjectSizes.measureDeep(new AccordData(Collections.emptyList()));
 
     private static PartitionKey getKey(FilteredPartition partition)
     {
         return new PartitionKey(partition.metadata().id, partition.partitionKey());
     }
 
-    public AccordData()
+    public static class Builder
```
aweisberg: Unused?

src/java/org/apache/cassandra/service/accord/serializers/EnumSerializer.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public E deserialize(DataInputPlus in, int version) throws IOException
+    {
+        return values[(int)in.readUnsignedVInt()];
+    }
+
+    @Override
+    public long serializedSize(E t, int version)
+    {
+        return TypeSizes.sizeofUnsignedVInt(t.ordinal());
```
aweisberg, line 56: In practice most enums are 1 byte and this could be hard coded for many enums.

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
```diff
@@ -689,22 +695,13 @@ public long serializedBodySize(IncrementingUpdate update, int version)
...
-        Set<Key> keys = new HashSet<>();
-        for (AbstractUpdate update : updates)
-            keys.add(update.partitionKey());
-        for (UpdatePredicate predicate : predicates)
-            keys.add(new PartitionKey(predicate.table.id, predicate.key));
-
-        return new Keys(keys);
-    }
-
     public AccordUpdate(List<AbstractUpdate> updates, List<UpdatePredicate> predicates)
```
aweisberg, line 698: Modifies its inputs, which for efficiency is of course good. It's only used from `AccordTxnBuilder` so seems fine.

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
```diff
@@ -689,22 +695,13 @@ public long serializedBodySize(IncrementingUpdate update, int version)
...
     {
-        this.keys = keysFrom(updates, predicates);
-        this.updates = serialize(updates, AbstractUpdate.serializer);
-        this.predicates = serialize(predicates, predicateSerializer);
+        updates.sort(Comparator.comparing(AbstractUpdate::partitionKey));
+        predicates.sort(Comparator.comparing(UpdatePredicate::partitionKey));
+        this.keys = Keys.of(updates, AbstractUpdate::partitionKey).union(Keys.of(predicates, UpdatePredicate::partitionKey));
```
aweisberg, line 702: Linear merge without the copy?

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
```diff
@@ -766,7 +769,54 @@ public Write apply(Data data)
...
+            j = in.findNext(out.get(i), j);
+            result[i] = from[j];
+        }
+        return result;
+    }
+
+    @Override
+    public Update merge(Update update)
+    {
+        // TODO: special method for linear merging keyed and non-keyed lists simultaneously
```
aweisberg, line 794: Seen this comment several times, what are non-keyed lists?

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
```diff
@@ -867,27 +913,26 @@ public AccordUpdate deserialize(DataInputPlus in, int version) throws IOExceptio
...
             return size;
         }
     };
 
-    private static <T> ByteBuffer serialize(T item, IVersionedSerializer<T> serializer)
+    private static <T> ByteBuffer serialize(List<T> items, int start, int end, IVersionedSerializer<T> serializer, int version)
     {
-        int version = MessagingService.current_version;
-        long size = serializer.serializedSize(item, version) + TypeSizes.INT_SIZE;
+        long size = TypeSizes.sizeofUnsignedVInt(version) + TypeSizes.sizeofUnsignedVInt(end - start);
```
aweisberg, line 926: Version is another thing that is a constant size in practice.

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
```diff
@@ -896,20 +941,46 @@ private static <T> ByteBuffer serialize(T item, IVersionedSerializer<T> serializ
         }
     }
 
-    private static <T> ByteBuffer[] serialize(List<T> items, IVersionedSerializer<T> serializer)
+    private static <T> ByteBuffer[] serialize(Keys keys, List<T> items, Function<? super T, ? extends Key> toKey, IVersionedSerializer<T> serializer)
```
aweisberg: A bit odd this is called serialize when it's really populating the arrays and the arrays contain empty byte buffers when the same key appears multiple times.

src/java/org/apache/cassandra/service/accord/store/StoredSet.java
```diff
@@ -179,11 +179,17 @@ public long estimatedSizeOnHeap(ToLongFunction<T> measure)
         return size;
     }
 
-    public static class Navigable<T extends Comparable<?>> extends StoredSet<T, NavigableSet<T>>
+    public interface Changes<T>
```
aweisberg, line 182: Pretty weird interface to stash here when it is just used in AsyncWriter for AsyncWriter things.

src/java/org/apache/cassandra/utils/ByteArrayUtil.java
```diff
@@ -195,6 +195,53 @@ public static void putDouble(byte[] b, int off, double val) {
...
+                putInt(b, off, (int)(register >> 32));
+                b[off + 4] = (byte)(register >> 24);
+                break;
+            case 6:
+                putInt(b, off, (int)(register >> 32));
+                putShort(b, off + 4, (short)(register >> 16));
+                break;
+            case 7:
+                putInt(b, off, (int)(register >> 32));
+                putShort(b, off + 4, (short)(register >> 16));
```
aweisberg: This has the same off + 4 and 5 issue?

src/java/org/apache/cassandra/utils/ByteArrayUtil.java
```diff
@@ -195,6 +195,53 @@ public static void putDouble(byte[] b, int off, double val) {
...
     }
 
+    /**
+     * An efficient way to write the type {@code bytes} of a long
+     *
+     * @param register - the long value to be written
+     * @param bytes - the number of bytes the register occupies. Valid values are between 1 and 8 inclusive.
+     * @throws IOException
+     */
+    public static void putBytes(byte[] b, int off, long register, int bytes)
```
aweisberg: Unit test it works in all 9 cases?

src/java/org/apache/cassandra/utils/ByteBufferUtil.java
```diff
@@ -726,6 +726,53 @@ public static int getShortLength(ByteBuffer bb, int position)
...
+                b.putInt(off, (int)(register >> 32));
+                b.put(off + 4, (byte)(register >> 24));
+                break;
+            case 6:
+                b.putInt(off, (int)(register >> 32));
+                b.putShort(off + 4, (short)(register >> 16));
+                break;
+            case 7:
+                b.putInt(off, (int)(register >> 32));
+                b.putShort(off + 4, (short)(register >> 16));
```
aweisberg: Has the offset + 4 and 5 issue

src/java/org/apache/cassandra/utils/ByteBufferUtil.java
```diff
@@ -726,6 +726,53 @@ public static int getShortLength(ByteBuffer bb, int position)
...
     }
 
+    /**
+     * An efficient way to write the type {@code bytes} of a long
+     *
+     * @param register - the long value to be written
+     * @param bytes - the number of bytes the register occupies. Valid values are between 1 and 8 inclusive.
+     * @throws IOException
+     */
+    public static void putBytes(ByteBuffer b, int off, long register, int bytes)
```
aweisberg: Unit test it works in all 9 cases?

aweisberg (commented):
Some questions, nits, and maybe one or two bugs.

src/java/org/apache/cassandra/utils/ByteBufferUtil.java
belliottsmith: This should have been removed entirely, my mistake.

src/java/org/apache/cassandra/utils/ByteArrayUtil.java
belliottsmith: This should have been removed entirely, my mistake.

src/java/org/apache/cassandra/utils/ByteArrayUtil.java
belliottsmith: This should have been removed entirely, my mistake.

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
belliottsmith: That's only because we haven't made it any more efficient here. This _should_ be implementation defined, so a bit of minor duplication for now seems fine.

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
belliottsmith: In `NotifyWaitingOn` we assume that commands higher up the stack than us are accessible, but the Context only permits you to access those that are in its context. I don't want to revisit the Context semantics too much for now, so this seemed like an acceptable approach.

src/java/org/apache/cassandra/service/accord/async/AsyncWriter.java
belliottsmith: My understanding is that the existence of a load future means the command hasn't been loaded yet. I believe this is being refactored by Blake as we speak.

src/java/org/apache/cassandra/db/marshal/ValueAccessor.java
belliottsmith: Should not be there, sorry

src/java/org/apache/cassandra/db/marshal/ValueAccessor.java
belliottsmith: Should not be there, sorry

src/java/org/apache/cassandra/dht/RandomPartitioner.java
belliottsmith: True, there seems to be generally inconsistent protections here between different kinds of token as well. I can try to normalise this if you like - personally, I think we should throw exceptions in all increase/decrease methods where we cannot safely do so.

src/java/org/apache/cassandra/service/accord/AccordCommand.java
belliottsmith: This isn't a hardcoding, I've renamed `kind` to `rw` to avoid clashing with `Txn.kind` (and also `Routable.kind`)

belliottsmith: Perhaps we can come up with another name for the method, but it's a method returning what its `rw` status is

src/java/org/apache/cassandra/service/accord/AccordCommand.java
belliottsmith: Sure. It might also be reasonable to insist it is only invoked on commands where we know the transaction already with a `Precondition` test

src/java/org/apache/cassandra/service/accord/AccordCommandStores.java
belliottsmith: -Ah, this looks like a bad IntelliJ refactor.-

Hmm. Actually, looking at it again, not sure that makes so much sense. I'll have a think about what we might be able to do to make this cleaner, but as structured the parent class is going to pass these parameters in either way, and I'm not sure it's clearer to ignore those parameters and supply the ones we have, even if they'll end up the same.

src/java/org/apache/cassandra/service/accord/AccordCommandsForKey.java
belliottsmith: Probably, yes. I thought I may have even had a TODO somewhere about this (or perhaps some other memory accounting).

I'm not sure it is worth agonising over this too much since 1) this is going to be pessimistic; 2) it's hard to cleanly avoid; 3) this is likely to be modified further before long, in a manner that may obviate this

src/java/org/apache/cassandra/service/accord/AccordObjectSizes.java
belliottsmith: Lots of room for optimisation of this kind of thing. I'll introduce a dedicated TODO for this case, to remind us.

src/java/org/apache/cassandra/service/accord/TokenRange.java
belliottsmith: The one in C* is only used for `describeRing` purposes, so I'm not sure it's sufficiently conflicting to be worth the verbosity.

src/java/org/apache/cassandra/service/accord/async/AsyncWriter.java
belliottsmith: Yep

src/java/org/apache/cassandra/service/accord/db/AbstractKeyIndexed.java
belliottsmith: There is, but I only mechanically updated these - I wasn't trying to edit the existing code. I've made the change, anyway

src/java/org/apache/cassandra/service/accord/db/AccordData.java
belliottsmith: 🤷  - will update

src/java/org/apache/cassandra/service/accord/serializers/EnumSerializer.java
belliottsmith: Agreed. I decided not to overthink it for now, though, as we also need to find a clean way to not rely on `ordinal` and many other incremental improvements. And this is basically fine and universal.

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
belliottsmith: Ok

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
belliottsmith: i.e. we have an array of keys, and another array of non-keys that correspond to the keys.

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
belliottsmith: Sure, though in downgrading from 4 to variable I decided not to semantically change the flexibility.

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
belliottsmith: > and the arrays contain empty byte buffers when the same key appears multiple times.

Not sure I follow? Open to changing the name (I was just updating existing methods)

src/java/org/apache/cassandra/service/accord/store/StoredSet.java
belliottsmith: All of the `StoredSet` things are pretty tightly coupled to `AsyncWriter`, and the implementations are `StoredSet`?

I wouldn't agonise over this personally, though, as it probably won't survive Blake's refactors (or not for long)

src/java/org/apache/cassandra/utils/ByteArrayUtil.java
aweisberg: FYI still here

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
aweisberg: `constructValueArrray`?

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
aweisberg: Actually looks like Union already does this.

src/java/org/apache/cassandra/service/accord/AccordCommandStores.java
aweisberg: It's more that the interface defined by Accord supplies the factory with things it maybe doesn't need. It's a minor nit.

src/java/org/apache/cassandra/service/accord/AccordCommand.java
aweisberg: That makes sense.

src/java/org/apache/cassandra/service/accord/AccordPartialCommand.java
belliottsmith: Should be resolved

src/java/org/apache/cassandra/service/accord/AccordPartialCommand.java
belliottsmith: Should be resolved

src/java/org/apache/cassandra/service/accord/db/AccordData.java
belliottsmith: Should be resolved

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
belliottsmith: This should be resolved now

src/java/org/apache/cassandra/utils/ByteArrayUtil.java
belliottsmith: This should be resolved now

src/java/org/apache/cassandra/service/accord/AccordCommand.java
belliottsmith: We always read the same scope, so I don't believe the check was ever necessary. However, this is more explicit now, as we read only the transactions associated with the transaction on disk for this store.

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
belliottsmith: To facilitate a cleaner method handle reference in AsyncWriter

src/java/org/apache/cassandra/service/accord/AccordCommand.java
belliottsmith: Hmm, so thinking about this more carefully, there is another unexpected side-effect of permitting `Accept` without a preceding PreAccept, as we need to know the `kind` to correctly adopt dependencies. This is probably fine, as the transaction will be visible to a quorum from its PreAccept, but it feels weird that a node that hasAccepted may fail to report the transaction as a dependency as a result.

I need to give it a little thought, but I am inclined to pack the `Kind` into the `TxnId` as it is simple, cheap (a single bit), and potentially simplifies other dependency analysis in future (since this would also be transmitted in all deps)

src/java/org/apache/cassandra/dht/RandomPartitioner.java
belliottsmith: This should be resolved now

src/java/org/apache/cassandra/service/accord/AccordCommand.java
belliottsmith: So we shouldn't actually need this null check, since the only place this is used is from the CommandsForKey, and we don't register ourselves with this until we have been pre-accepted and know the transaction definition (in order to know the keys we need to register with). So I think it is fine to leave this alone for now, and simply remove the null check.

belliottsmith:
I have squashed, rebased and force-pushed. I have left a single follow-up commit that represents changes since Thursday.

belliottsmith:
I have squashed, rebased and force-pushed. I have left a single follow-up commit that represents changes since Thursday morning.

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
```diff
@@ -867,27 +913,26 @@ public AccordUpdate deserialize(DataInputPlus in, int version) throws IOExceptio
...
-            size += TypeSizes.sizeof(update.predicates.length);
             for (ByteBuffer buffer : update.predicates)
                 size += ByteBufferUtil.serializedSizeWithVIntLength(buffer);
-
             return size;
         }
     };
 
-    private static <T> ByteBuffer serialize(T item, IVersionedSerializer<T> serializer)
+    private static <T> ByteBuffer toSerializedValueArray(List<T> items, int start, int end, IVersionedSerializer<T> serializer, int version)
```
aweisberg:
```suggestion
    private static <T> ByteBuffer toSerializedValue(List<T> items, int start, int end, IVersionedSerializer<T> serializer, int version)
```

src/java/org/apache/cassandra/service/accord/db/AccordUpdate.java
```diff
@@ -896,20 +941,46 @@ private static <T> ByteBuffer serialize(T item, IVersionedSerializer<T> serializ
         }
     }
 
-    private static <T> ByteBuffer[] serialize(List<T> items, IVersionedSerializer<T> serializer)
+    private static <T> ByteBuffer[] toSerializedValueArrays(Keys keys, List<T> items, Function<? super T, ? extends Key> toKey, IVersionedSerializer<T> serializer)
```
aweisberg:
```suggestion
    private static <T> ByteBuffer[] toSerializedValueArray(Keys keys, List<T> items, Function<? super T, ? extends Key> toKey, IVersionedSerializer<T> serializer)
```

aweisberg (approved):
+1 TY

PR #1953, opened 2022-10-26
[CASSANDRA-17950] Enable dtest-offheap in CircleCI

dchenbecker:
The dtest-offheap test was only added to the Jenkins configuration, so this commit adds it to the CircleCI build for parity.

Patch by Derek Chen-Becker; reviewed by Ekaterina Dimitrova for CASSANDRA-17950

### DIFF
--- a/.circleci/config-2_1.yml
+++ b/.circleci/config-2_1.yml
@@ -299,6 +299,19 @@ separate_jobs: &separate_jobs
         requires:
           - start_j8_repeated_ant_test
           - build
+    # Java 8 off-heap tests
+    - start_j8_dtests_offheap:
+        type: approval
+    - j8_dtests_offheap:
+        requires:
+          - start_j8_dtests_offheap
+          - build
+    - start_j8_dtests_offheap_repeat:
+        type: approval
+    - j8_dtests_offheap_repeat:
+        requires:
+          - start_j8_dtests_offheap_repeat
+          - build
 
 pre-commit_jobs: &pre-commit_jobs
   jobs:
@@ -390,6 +403,17 @@ pre-commit_jobs: &pre-commit_jobs
     # Java 8 repeated utest
     - j8_repeated_ant_test:
         requires:
+          - build
+    # Java 8 off-heap tests
+    - start_j8_dtests_offheap:
+        type: approval
+    - j8_dtests_offheap:
+        requires:
+          - start_j8_dtests_offheap
+          - build
+    - j8_dtests_offheap_repeat:
+        requires:
+          - start_j8_dtests_offheap
           - build
 
 workflows:
@@ -654,6 +678,36 @@ jobs:
           upgrade: "true"
           stop_on_failure: ${REPEATED_TESTS_STOP_ON_FAILURE}
           count: ${REPEATED_UPGRADE_DTESTS_COUNT}
+
+  j8_dtests_offheap:
+    <<: *j8_par_executor
+    steps:
+      - attach_workspace:
+          at: /home/cassandra
+      - log_environment
+      - clone_dtest
+      - create_venv
+      - create_dtest_containers:
+          file_tag: j8_dtests_offheap
+          run_dtests_extra_args: "--use-vnodes --use-off-heap-memtables --skip-resource-intensive-tests"
+      - run_dtests:
+          file_tag: j8_dtests_offheap
+          pytest_extra_args: '--use-vnodes --num-tokens=16 --use-off-heap-memtables --skip-resource-intensive-tests'
+
+  j8_dtests_offheap_repeat:
+    <<: *j8_repeated_dtest_executor
+    steps:
+      - attach_workspace:
+          at: /home/cassandra
+      - clone_dtest
+      - create_venv
+      - run_repeated_dtest:
+          tests: ${REPEATED_DTESTS}
+          vnodes: "true"
+          upgrade: "false"
+          count: ${REPEATED_DTESTS_COUNT}
+          stop_on_failure: ${REPEATED_TESTS_STOP_ON_FAILURE}
+          extra_dtest_args: "--use-off-heap-memtables --skip-resource-intensive-tests"
 
 commands:
   log_environment:
@@ -1288,6 +1342,9 @@ commands:
         type: string
       stop_on_failure:
         type: string
+      extra_dtest_args:
+        type: string
+        default: ""
     steps:
       - run:
           name: Run repeated Python dtest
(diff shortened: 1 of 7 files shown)


dchenbecker:
For posterity, this was originally on https://github.com/apache/cassandra/pull/1940 but I had to change branches, meaning we needed a new PR (this one)

ekaterinadimitrova2:
You are almost there. The patches are not created correctly. 
You use the recipe I mentioned on the other PR which covers the case when the patches are still the same, applicable, just moving a few lines and unneeded warnings popping up. In your case when you add the new jobs things shift and you still get wrong resources assigned to your jobs. 
You will need to modify cassandra-2_1.yml so that it can generate correct MIDRES and HIGHRES files and then create a patch. The way we do patches for code. You do the changes, you get a patch from the diff.
This is explained in the readme and I explained in the very first patch. We always need to check the final results in the MIDRES and HIGHRES files to ensure resources didn't shift for some reason. 
Now in this case I can still see wrong resources. This is what I mentioned in the previous PR review, it is still applicable:
`Actually how were MIDRES and HIGHRES created? They use the same j8_par_executor as the other dtests but I see different resources in Midres - medium and 25, when j8_dtests gets large and 50?` Your resources are still wrong because you probably used the recipe for recreating patches which cover the case when resources are correctly assigned. 
So looking into the current MIDRES file for cassandra 3.11 you need to end up with LARGE and 50 for running all tests(https://github.com/apache/cassandra/blob/cassandra-3.11/.circleci/config.yml.MIDRES#L1728-L1734) and LARGE and 25 for running tests in a loop (https://github.com/apache/cassandra/blob/cassandra-3.11/.circleci/config.yml.MIDRES#L1057-L1063)

dchenbecker:
OK, I updated the midres patch. For what it's worth, the README doesn't discuss what's in the patches at all, or what the valus should be for various runs. For example, the MIDRES patch changes the executors used by that various dtests in addition to changing resource class and parallelism, but the HIGHRES only changes resource class and parallelism, but not the executors. I'm happy to add documentation to the README for the next person, but it's not clear at all why things are done the way they are here. Looking at  git blame ( :) ) I can see you've done a lot of work on this, so you're much more familiar with this than I am, but I think we need to be careful about confusing familiarity for clarity.

ekaterinadimitrova2:
`OK, I updated the midres patch` I will take a look soon. 
` For what it's worth, the README doesn't discuss what's in the patches at all, or what the valus should be for various runs.`
 What values ---> well, when we apply the patch we see what were the values (same as in code patching) and we need to ensure when we add new jobs the old jobs are not changed or affected and the new additions are as we want them.  I'd say in this case for the new job we agreed to match the values that were assigned to the dtests(those values are assigned in time through testing, there Is no direct matrix), that is how  I compared that in MIDRES file something is wrong and we do not have the values we wanted to have.Another way to make a check is by looking at the executor assigned to our job, the resources in MIDRES and in that executor should match. Maybe we need to say that patch is nothing else than a regular patch, not something circle ci specific? 
About the executor - well, this is circle ci terminology, I do not think we should copy paste from the CircleCI docs in our readme. A new executor is added to raise the resources to a level not applicable for LOWRES and HIGHRES. I think the main issue for new comers is that having Readme they might not go to CircleCI docs themselves and miss how things in CircleCI itself work. (That is what I also did when I started, the people who created the initial setup with generate.sh were not active anymore on the project :( )I do not want to add links to docs in the readme as links change with versions and also they can get broken. Maybe what we can do is to say "For information on executors please check CircleCI documentation" or refer to this section or something else (but by name, not link)  https://circleci.com/docs/reusing-config/ 

`Looking at git blame ( :) ) I can see you've done a lot of work on this, so you're much more familiar with this than I am, but I think we need to be careful about confusing familiarity for clarity.`

I do not mean to argue and I totally agree about familiarity, especially in the Cassandra world where CI is no exception for complexity considering how many test suites, config options, etc we have. I am with you here. One of the reasons why I gave you an advice to start with one job and not like for example Andres who deal with all unit tests variations at once as he knows the setup by hard already. What I wanted to point out is that patch is a patch no matter whether it is config file or Java code. We need to verify the result of that patch. Clarity - I think we just need to realize that Cassandra works with other tools, frameworks and libraries that have their own documentation. It takes time to get acquainted with everything and make the connections. It is its own process we all go through. But we cannot and we should not rely on a customer specific readme to cover how the actual tool works in detail IMHO. Again, I suggest we add a point on top  of the readme - "below information is cassandra specific usage of CircleCI, you might want to read through the CircleCI documentation if you are not familiar with the tool before moving forward." or something similar.
My concern is even when we further improve and automate things, there will be always corner cases and breakages and new suites and specifics so operating one of our primary tools as a black box without taking the time to learn it how it works is not sustainable. Personal opinion. Also, CircleCI as a product itself evolves so we might want to follow up on new opportunities that we can take advantage of.
I will review and post update in the next few hours. Thank you for the quick fix :-) I think this one job is a good first exercise and then adding others, being similar will be completely different experience. Thank you for your efforts

.circleci/generate.sh
```diff
@@ -268,6 +268,8 @@ fi
 if (! (echo "$env_vars" | grep -q "REPEATED_DTESTS=")); then
   delete_job "j8_dtests_repeat"
   delete_job "j8_dtests_vnode_repeat"
+  delete_job "start_j8_dtests_offheap_repeat"
```
adelapena, line 271: This line is not needed, the call to `delete_job` right below also deletes the approval job starting with `start_`.

.circleci/config-2_1.yml
```diff
@@ -299,6 +299,19 @@ separate_jobs: &separate_jobs
         requires:
           - start_j8_repeated_ant_test
           - build
+    # Java 8 off-heap tests
```
adelapena, line 302:
```suggestion
    # Java 8 off-heap dtests
```

.circleci/config-2_1.yml
```diff
@@ -391,6 +404,17 @@ pre-commit_jobs: &pre-commit_jobs
     - j8_repeated_ant_test:
         requires:
           - build
+    # Java 8 off-heap tests
```
adelapena, line 407:
```suggestion
    # Java 8 off-heap dtests
```

.circleci/generate.sh
ekaterinadimitrova2: Btw we have j11 jobs in 3.11 branch to be deleted? It is a noop though....

pmcfadin:
Closing as [CASSANDRA-17950](https://issues.apache.org/jira/browse/CASSANDRA-17950) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #1954, opened 2022-10-26
Cassandra 17998

tjake:
The [Cassandra Jira](https://issues.apache.org/jira/projects/CASSANDRA/issues/CASSANDRA-17998)


jmckenzie-dev:
Some top level questions (about to transfer line items from notes to the diff):
* Why would caching a mutation at any given size be particularly bad for GC? We're already paying the tax to ser or deser in the first place; how long do we expect these messages to hang around and live -> worst-case is timeout from replicas?
* Where does our initial value for the default on `cacheable_mutation_size_limit_bytes` come from?
* If we're doing ThreadLocal on the limitedScratchBuffer, how are we managing the implications of a Mutation `deserialize` call bouncing across threads? We'll end up with a deserialized copy of it living in memory on another thread should something jump to another Stage in the future etc right? Not a big deal as they should be short-lived eden and we'd be deserializing it anyway, just thinking of the tradeoffs between having a central cache of deserialized mutations w/the current "raciness" attributes vs. the threadlocal approach.

edit: ah, see it now: `Huge" serialized mutations can have a
                // bad impact to G1 GC, if the cached serialized mutation results in a
                // "humonguous object" and also frequent re-allocations of the scratch buffer(s).
                // I.e. large cached mutation objects cause GC pressure.`

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -64,6 +67,13 @@ public class Mutation implements IMutation, Supplier<Mutation>
 
     private final boolean cdcEnabled;
 
+    // Contains serialized representations of this mutation.
+    // Note: there is no functionality to clear/remove serialized instances, because a mutation must never
+    // be modified (e.g. calling add(PartitionUpdate)) when it's being serialized.
+    private static final int CACHED_SERIALIZATIONS = MessagingService.Version.values().length;
+    private static final int CACHEABLE_MUTATION_SIZE_LIMIT = Integer.getInteger(Config.PROPERTY_PREFIX + "cacheable_mutation_size_limit_bytes", 2 * 1024 * 1024) - 24;
```
jmckenzie-dev: What's the -24 here for? Maybe document it?

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -393,34 +388,130 @@ public static class MutationSerializer implements IVersionedSerializer<Mutation>
...
+         * See {@link org.apache.cassandra.service.StorageProxy#sendToHintedEndpoints(Mutation, WriteEndpoints, WriteHandler, Verb.AckedRequest)}.
+         */
+        @SuppressWarnings("JavadocReference")
+        public void prepareSerializedBuffer(Mutation mutation, int version)
+        {
+            serialization(mutation, version);
+        }
+
+        /**
+         * Retrieve the cached serialization of this mutation, or computed and cache said serialization if it doesn't
```
jmckenzie-dev: nit: "compute and cache"

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -393,34 +388,130 @@ public static class MutationSerializer implements IVersionedSerializer<Mutation>
...
+         * exists yet. Note that this method is _not_ synchronized even though it may (and will often) be called
```
jmckenzie-dev: nit: "exist yet"

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -393,34 +388,130 @@ public static class MutationSerializer implements IVersionedSerializer<Mutation>
...
+                    if (!serializeInternal(PartitionUpdate.serializer, mutation, dob, version,true))
+                    {
+                        serialization = new NonCacheableSerialization();
+                    }
+                    else
+                    {
+                        serialization = new CachedSerialization(dob.toByteArray());
+                    }
+                }
+                catch (DataOutputBuffer.BufferCapacityExceededException tooBig)
```
jmckenzie-dev: What are the implications of us using exception handling basically for control flow in the case of mutations that exceed our cacheable capacity vs. some other less invasive method? Wondering from a performance perspective as this is hot path and exception handling is notoriously Not Fast.

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -393,34 +388,130 @@ public static class MutationSerializer implements IVersionedSerializer<Mutation>
...
-            int size = (int)in.readUnsignedVInt();
-            assert size > 0;
-
-            PartitionUpdate update = PartitionUpdate.serializer.deserialize(in, version, flag);
-            if (size == 1)
-                return new Mutation(update);
-
-            ImmutableMap.Builder<TableId, PartitionUpdate> modifications = new ImmutableMap.Builder<>();
-            DecoratedKey dk = update.partitionKey();
+            return isPrepare && out.position() > CACHEABLE_MUTATION_SIZE_LIMIT;
```
jmckenzie-dev: Consider a `CassandraRelevantProperties` for this parameter.

src/java/org/apache/cassandra/io/util/TeeDataInputPlus.java
```diff
@@ -0,0 +1,216 @@
...
+    private boolean limitReached = false;
+
+    public TeeDataInputPlus(DataInputPlus source, DataOutputPlus teeBuffer)
+    {
+        this(source, teeBuffer, -1);
+    }
+
+    public TeeDataInputPlus(DataInputPlus source, DataOutputPlus teeBuffer, long limit)
+    {
+        this.source = source;
```
jmckenzie-dev, line 50: We may want to assert that neither input is null here. May not; just a thought.

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -298,29 +308,14 @@ public String toString(boolean shallow)
...
-                    serializedSize3014 = (int) serializer.serializedSize(this, VERSION_3014);
-                return serializedSize3014;
-            case VERSION_40:
-                if (serializedSize40 == 0)
-                    serializedSize40 = (int) serializer.serializedSize(this, VERSION_40);
-                return serializedSize40;
-            default:
-                throw new IllegalStateException("Unknown serialization version: " + version);
-        }
+        if (serializedSize == 0)
```
jmckenzie-dev: Looks like we're losing our safety check against a bad version being passed in. If we have confidence this is the case we should document this method w/the invariant of the version being valid and/or Precondition check it. Or just restore some variation of the prior switch that throws an `IllegalStateException`

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -393,34 +388,130 @@ public static class MutationSerializer implements IVersionedSerializer<Mutation>
     {
         public void serialize(Mutation mutation, DataOutputPlus out, int version) throws IOException
         {
+            serialization(mutation, version).serialize(PartitionUpdate.serializer, mutation, out, version);
+        }
+
+        /**
+         * Called early during request processing to prevent that {@link #serialization(Mutation)} is
+         * called concurrently.
+         * See {@link org.apache.cassandra.service.StorageProxy#sendToHintedEndpoints(Mutation, WriteEndpoints, WriteHandler, Verb.AckedRequest)}.
```
jmckenzie-dev: This is an invalid javadoc target; doesn't resolve and I can't find it in the codebase.

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -393,34 +388,130 @@ public static class MutationSerializer implements IVersionedSerializer<Mutation>
...
+                // We need to use a capacity-limited DOB here.
+                // If a mutation consists of one PartitionUpdate with one column that exceeds the
+                // "cacheable-mutation-size-limit", a capacity-limited DOB can handle that case and
+                // throw a BufferCapacityExceededException. "Huge" serialized mutations can have a
+                // bad impact to G1 GC, if the cached serialized mutation results in a
+                // "humonguous object" and also frequent re-allocations of the scratch buffer(s).
+                // I.e. large cached mutation objects cause GC pressure.
+                try (DataOutputBuffer dob = DataOutputBuffer.limitedScratchBuffer(CACHEABLE_MUTATION_SIZE_LIMIT))
+                {
+                    if (!serializeInternal(PartitionUpdate.serializer, mutation, dob, version,true))
```
jmckenzie-dev: A ternary operator on the construction of the serialization variable here may help tidy up clarity a bit.

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -393,34 +388,130 @@ public static class MutationSerializer implements IVersionedSerializer<Mutation>
...
+                }
+                catch (DataOutputBuffer.BufferCapacityExceededException tooBig)
+                {
+                    serialization = new NonCacheableSerialization();
+                }
+                catch (IOException e)
+                {
+                    throw new RuntimeException(e);
+                }
+                mutation.serializations[versionIndex] = serialization;
```
jmckenzie-dev: I was a little startled to see a `NonCacheableSerialization`... cached here. :) Might be worth renaming it to something like `SizeOnlyCachedSerialization` or something?

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -393,34 +388,130 @@ public static class MutationSerializer implements IVersionedSerializer<Mutation>
...
+            }
+            return serialization;
+        }
+
+        static boolean serializeInternal(PartitionUpdate.PartitionUpdateSerializer serializer,
```
jmckenzie-dev: Recommend we change this to `maybeSerializeInternal` to reflect that it might or might not actually complete and that's acceptable. The fact that we use it w/out parsing the return value and also w/parsing it implies that there's some flexibility in that usage.

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -393,34 +388,130 @@ public static class MutationSerializer implements IVersionedSerializer<Mutation>
...
+            {
+                serializer.serialize(partitionUpdate, out, version);
+                if (isCacheableMutationSizeLimit(out, isPrepare))
+                    return false;
+            }
+            return true;
         }
 
-        public Mutation deserialize(DataInputPlus in, int version, DeserializationHelper.Flag flag) throws IOException
+        private static boolean isCacheableMutationSizeLimit(DataOutputPlus out, boolean isPrepare)
```
jmckenzie-dev: Super nit: maybe we rename to `atCacheableMutationSizeLimit`? The name kind of confused me as to what it was doing; kind of assumed it was checking to see if the limit _existed_ not that we'd hit it.

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -212,6 +215,23 @@ public class MessagingService extends MessagingServiceMBeanImpl
...
+    /**
+     * This is an optimisation to speed up the translation of the serialization
+     * version to the {@link Version} enum.
+     *
+     * @param version the serialization version
+     * @return a {@link Version}
+     */
+    public static int getVersionIndex(int version)
+    {
+        if (versionIndexMap.containsKey(version))
```
jmckenzie-dev: I think we could just .get on the map and throw on value == null. Save us the lookup then retrieve steps.

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -212,6 +215,23 @@ public class MessagingService extends MessagingServiceMBeanImpl
...
+            return versionIndexMap.get(version);
+        throw new IllegalStateException("Unkown serialization version: " + version);
+    }
+
+    public final static boolean NON_GRACEFUL_SHUTDOWN = Boolean.getBoolean("cassandra.test.messagingService.nonGracefulShutdown");
```
jmckenzie-dev: This looks vestigial.

src/java/org/apache/cassandra/io/util/TeeDataInputPlus.java
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * DataInput that also stores the raw inputs into an output buffer
+ * This is useful for storing serialized buffers as they are deserialized
+ */
+public class TeeDataInputPlus implements DataInputPlus
+{
+    private final DataInputPlus source;
+    private final DataOutputPlus teeBuffer;
+
+    private final long limit;
```
jmckenzie-dev, line 39: Why do we use -1 to indicate no limit here vs. the 0 to mean no limit in `DataOutputBuffer.LimitingScratchBuffer`?

src/java/org/apache/cassandra/io/util/TeeDataInputPlus.java
```diff
@@ -0,0 +1,216 @@
...
+
+import java.io.EOFException;
+import java.io.IOException;
+
+import org.apache.cassandra.db.TypeSizes;
+import org.apache.cassandra.utils.Throwables;
+
+/**
+ * DataInput that also stores the raw inputs into an output buffer
+ * This is useful for storing serialized buffers as they are deserialized
```
jmckenzie-dev: Would add something to class level javadoc about the API -> that writes will stop buffering when you hit the limit and you can check that state with `isLimitReached`.

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -393,34 +388,130 @@ public static class MutationSerializer implements IVersionedSerializer<Mutation>
...
+                    modifications.put(update.metadata().id, update);
+                    for (int i = 1; i < size; ++i)
+                    {
+                        update = PartitionUpdate.serializer.deserialize(in, version, flag);
+                        modifications.put(update.metadata().id, update);
+                    }
+                    m = new Mutation(update.metadata().keyspace, dk, modifications.build(), approxTime.now());
+                }
+
+                if (!((TeeDataInputPlus)in).isLimitReached())
```
jmckenzie-dev: So higher level meta question: right now there's really no sign that the data in the `DataOutputBuffer` is incomplete except for checking the `isLimitReached` method inside `TeeDataInputPlus`. This leaves us with the case where a buffer could be floating around and accessed in an incomplete state (between write calls above and checking) - how could we better tie this together?

Full disclosure - I can't think of a good answer. I don't think decorating `DataOutputBuffer` with a bool indicating if it's valid or not makes any sense for 99+% of the other use-cases. We could consider wrapping access to the nested `teeBuffer` inside `TeeDataInputPlus` to return null or some other sentinel on limit reached but at that point it's just a worse version of this current API.

Anyway - this might be something worth calling out on the JavaDoc class level comment for TeeDataInputPlus and be clearly prescriptive about the intended use of the class.

src/java/org/apache/cassandra/io/util/TeeDataInputPlus.java
```diff
@@ -0,0 +1,216 @@
...
+    public void skipBytesFully(int n) throws IOException
+    {
+        source.skipBytesFully(n);
+        maybeWrite(n, () -> {
+            for (int i = 0; i < n; i++)
+                teeBuffer.writeByte(0);
+        });
+    }
+
+    public boolean isLimitReached()
```
jmckenzie-dev, line 221: JavaDoc this to indicate that a "true" return indicates the data in your `teeBuffer` can't be trusted.

test/unit/org/apache/cassandra/utils/TeeDataInputPlusTest.java
```diff
@@ -0,0 +1,141 @@
...
+        skipped = limitedTee.skipBytes(100);
+        assertEquals(23, skipped);
+
+        byte[] teeData = teeOut.toByteArray();
+        assertFalse(tee.isLimitReached());
+        assertTrue(Arrays.equals(testData, teeData));
+
+        byte[] limitedTeeData = limitedTeeOut.toByteArray();
+        assertTrue(limitedTee.isLimitReached());
+        assertTrue(Arrays.equals(Arrays.copyOf(testData, LIMITED_SIZE -1 ), limitedTeeData));
```
jmckenzie-dev: ... 2 integers together without a delimiter? I'm not familiar with this - what's this about?

src/java/org/apache/cassandra/io/util/DataOutputBuffer.java
```diff
@@ -61,32 +61,62 @@ private enum AllocationType { DIRECT, ONHEAP }
...
+    {
+        LimitingScratchBuffer dob = (LimitingScratchBuffer) scratchBuffer.get();
+        dob.limit = limit;
+        return dob;
+    }
+
+    public static class BufferCapacityExceededException extends RuntimeException
+    {}
+
+    private static final class LimitingScratchBuffer extends DataOutputBuffer
```
jmckenzie-dev: Similar to how size limiting is a first class citizen in `TeeDataInputPlus`, I _think_ we could do the same thing for DataOutputPlus by promoting the `initialSizeBuffer` and `limit` to top level variables and just not use them in the case where the "no limit" constructor is used. That'd give us consistency between the structures of the `DataOutputPlus` and `TeeDataInputPlus` and their treatment of limits and remove one layer of type indirection that exists here, as well as the casting on usage.

src/java/org/apache/cassandra/net/MessagingService.java
```diff
@@ -212,6 +215,23 @@ public class MessagingService extends MessagingServiceMBeanImpl
...
+    static Map<Integer, Integer> versionIndexMap = Arrays.stream(Version.values()).collect(Collectors.toMap(v -> v.value, v -> v.ordinal()));
+
+    /**
+     * This is an optimisation to speed up the translation of the serialization
+     * version to the {@link Version} enum.
+     *
+     * @param version the serialization version
+     * @return a {@link Version}
+     */
+    public static int getVersionIndex(int version)
```
jmckenzie-dev: Nit: Consider renaming to something very literal like `getVersionOrdinal` -> the current name (index) doesn't really clarify _what_ index it's retrieving for what context / what purpose / etc.

I've spent some time chewing on this; I really don't like it from a meta / aesthetic perspective but I think we're backed into a corner w/the original enum implementation and it's rather strange values that we're using for `Version` right now.

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -430,10 +521,71 @@ public Mutation deserialize(DataInputPlus in, int version) throws IOException
...
-            int size = TypeSizes.sizeofUnsignedVInt(mutation.modifications.size());
-            for (Map.Entry<TableId, PartitionUpdate> entry : mutation.modifications.entrySet())
-                size += PartitionUpdate.serializer.serializedSize(entry.getValue(), version);
+            return serialization(mutation, version).serializedSize(PartitionUpdate.serializer, mutation, version);
+        }
+    }
 
+    /**
+     * There are two implementations of this class. One that keeps the serialized representation on-heap for later
+     * reuse and one that doesn't. Keeping all sized mutations around may lead to "bad" GC pressure (G1 GC) due to humongous objects.
```
jmckenzie-dev, line 538: Think we should lift this up to the CACHEABLE_MUTATION_SIZE_LIMIT declaration as a javadoc so it's clear at declaration time why this limit exists.

jmckenzie-dev (changes requested):

src/java/org/apache/cassandra/io/util/DataOutputBuffer.java
```diff
@@ -61,32 +61,62 @@ private enum AllocationType { DIRECT, ONHEAP }
...
         }
     };
 
+    /**
+     * Returns a {@link DataOutputBuffer} that is limited to a capacity of {@code limit} bytes.
+     * A {@code limit} of 0 means unlimited.
+     */
+    public static DataOutputBuffer limitedScratchBuffer(long limit)
+    {
+        LimitingScratchBuffer dob = (LimitingScratchBuffer) scratchBuffer.get();
```
jmckenzie-dev: Been chewing on this - given the advice is to close out this DOB quickly to prevent heap pressure, what's the point of using the FastThreadLocal scratchBuffer instead of just a new DOB?

src/java/org/apache/cassandra/db/Mutation.java
tjake: Yeah, I'm reworking this, good call out

src/java/org/apache/cassandra/net/MessagingService.java
tjake: oops

test/unit/org/apache/cassandra/utils/TeeDataInputPlusTest.java
tjake: I'm not sure I follow, do you mean the `LIMITED_SIZE-1`?  Since the limit is `LIMITED_SIZE` the buffer stopped copying at `LIMITED_SIZE-1` bytes

tjake:
Thanks for the great review.  I reworked the code to avoid the whole limiting buffer/exception throwing. And I think addressed all the other issues you mentioned.

> Why would caching a mutation at any given size be particularly bad for GC? 

Since you can end up making [Humongous allocations](https://docs.oracle.com/javase/10/gctuning/garbage-first-garbage-collector.htm#JSGCT-GUID-AC383806-7FA7-4698-8B92-4FD092B9F368) 

This threshold is based on the size of the heap but 1MB is a good default since that's the low end and covers 99% of Mutations.

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -393,34 +411,115 @@ public static class MutationSerializer implements IVersionedSerializer<Mutation>
...
+        {
+            int versionOrdinal = MessagingService.getVersionOrdinal(version);
+            // Retrieves the cached version, or build+cache it if it's not cached already.
+            Serialization serialization = mutation.cachedSerializations[versionOrdinal];
+            if (serialization == null)
+            {
+                serialization = new SizeOnlyCacheableSerialization();
+                long sertializedSize = serialization.serializedSize(PartitionUpdate.serializer, mutation, version);
+
+                // Exessively large mutation objects cause GC pressure and huge allocations when serialized.
```
jmckenzie-dev: nit: Excessively

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -393,34 +411,115 @@ public static class MutationSerializer implements IVersionedSerializer<Mutation>
...
+        @SuppressWarnings("JavadocReference")
+        private Serialization serialization(Mutation mutation, int version)
+        {
+            int versionOrdinal = MessagingService.getVersionOrdinal(version);
+            // Retrieves the cached version, or build+cache it if it's not cached already.
+            Serialization serialization = mutation.cachedSerializations[versionOrdinal];
+            if (serialization == null)
+            {
+                serialization = new SizeOnlyCacheableSerialization();
+                long sertializedSize = serialization.serializedSize(PartitionUpdate.serializer, mutation, version);
```
jmckenzie-dev: nit: spelling. Should be "serializedSize"

test/unit/org/apache/cassandra/utils/TeeDataInputPlusTest.java
jmckenzie-dev: Oh. 🤦 This is a formatting issue - missing a whitespace between the - and the 1. I was literally reading
`[int] [-int]` instead of `[int] - [int]`.

So this should be
```
assertTrue(Arrays.equals(Arrays.copyOf(testData, LIMITED_SIZE - 1 ), limitedTeeData));
```

src/java/org/apache/cassandra/io/util/DataOutputBuffer.java
jmckenzie-dev: A lot clearer now w/the new init + overrides on scratchBuffer init. 👍

jmckenzie-dev:
Couple spelling nits and it looks like we dropped the new unit test in DataOutputTest when we dropped that unused import:
```
    @Test
    public void testDataOutputBufferLimitingScratchBuffer() throws IOException
    {
```

w/those fixed we're good to go. 👍

jmckenzie-dev (approved):

tjake:
> Couple spelling nits and it looks like we dropped the new unit test in DataOutputTest when we dropped that unused import:
> 
> ```
>     @Test
>     public void testDataOutputBufferLimitingScratchBuffer() throws IOException
>     {
> ```
> 
> w/those fixed we're good to go. +1

I removed the LimitingScratchBuffer altogether.  Fixing the typos thx

src/java/org/apache/cassandra/db/Mutation.java
```diff
@@ -64,6 +66,13 @@ public class Mutation implements IMutation, Supplier<Mutation>
 
     private final boolean cdcEnabled;
 
+    // Contains serialized representations of this mutation.
+    // Note: there is no functionality to clear/remove serialized instances, because a mutation must never
+    // be modified (e.g. calling add(PartitionUpdate)) when it's being serialized.
+    private static final int CACHED_SERIALIZATIONS = MessagingService.Version.values().length;
+    private static final int CACHEABLE_MUTATION_SIZE_LIMIT = Integer.getInteger(Config.PROPERTY_PREFIX + "cacheable_mutation_size_limit_bytes", 2 * 1024 * 1024) - 24;
```
smiklosovic: This should be moved to `CassandraRelevantProperties`

PR #1965, opened 2022-11-02
CASSANDRA-17679-Fix BootstrapTest failure on vnodes

jmckenzie-dev:
Fix Bootstrap test failure; RTE catch was swallowing the AssumptionViolatedException that we normally check to ensure a test doesn't run in a vnode config if the test explicitly calls to be run without it

### DIFF
--- a/test/distributed/org/apache/cassandra/distributed/test/ring/BootstrapTest.java
+++ b/test/distributed/org/apache/cassandra/distributed/test/ring/BootstrapTest.java
@@ -28,6 +28,7 @@ import java.util.stream.IntStream;
 
 import org.junit.After;
 import org.junit.Assert;
+import org.junit.AssumptionViolatedException;
 import org.junit.Before;
 import org.junit.Test;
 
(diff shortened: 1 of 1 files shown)


test/distributed/org/apache/cassandra/distributed/test/ring/BootstrapTest.java
```diff
@@ -196,6 +197,12 @@ public void bootstrapUnspecifiedFailsOnResumeTest() throws Throwable
                                bootstrap()),
                         newInstance.config().num());
         }
+        catch (AssumptionViolatedException ave)
+        {
+            // We get an AssumptionViolatedException if we're in a test job configured w/vnodes
+            if (ave.getMessage().contains("vnode is requested but not supported"))
```
dcapwell: remove this check, the type is very explicit "junit ignore this test", you trying to ignore that isn't good and will be brittle.

Also, add `throw ave` so it actually is skipped and not passing.

dcapwell (changes requested):

dcapwell (approved):

PR #1978, opened 2022-11-07
CASSANDRA-17913: Fix nested selection of reversed collections

masokol:


masokol:
Commited in https://github.com/apache/cassandra/commit/2fce3025c448316720d9cf7447fe7d1d34e17625

PR #1983, opened 2022-11-07
CASSANDRA-17964 4.0 ant

smiklosovic:
 
```

```

### DIFF
--- a/test/unit/org/apache/cassandra/cql3/validation/operations/CompactStorageTest.java
+++ b/test/unit/org/apache/cassandra/cql3/validation/operations/CompactStorageTest.java
@@ -50,7 +50,6 @@ import static org.apache.cassandra.utils.ByteBufferUtil.EMPTY_BYTE_BUFFER;
 import static org.apache.cassandra.utils.ByteBufferUtil.bytes;
 import static org.junit.Assert.assertArrayEquals;
 import static org.junit.Assert.fail;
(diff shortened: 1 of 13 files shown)


test/unit/org/apache/cassandra/cql3/validation/operations/CompactStorageTest.java
```diff
@@ -4369,19 +4368,9 @@ private void testUpdate(boolean forceFlush) throws Throwable
         execute("UPDATE %s SET value = ? WHERE partitionKey = ? AND clustering_1 = ?", null, 0, 0);
         flush(forceFlush);
 
-        if (isEmpty(compactOption))
```
smiklosovic, line 4372: I believe this was completely wrong because `isEmpty` was used from `org.reflections` and this started to fail to compile because that lib was updated and that method is not there anymore. But that `if` condition does not make sense as it was done originally because it tests `compactOption` which is `private final static` `String`. Hence, only `else` branch makes sense.

Utils.isEmpty method does internally this:

````
    public static boolean isEmpty(String s) {
        return s == null || s.length() == 0;
    }
````

Hence it always returned false.

dcapwell: switching `isEmpty` was done in trunk as well, so makes sense

PR #1984, opened 2022-11-07
Validate the comparison git branch when generating CircleCI configs

dchenbecker:
Allow the user to specify a flag or environment variable to specify a non-default comparison git branch when detetcing changed unit tests to repeat.

Tested locally by generating the various resource config yml files with both a valid and invalid base branch.

Patch by Derek Chen-Becker; reviewed by TBD for CASSANDRA-17997

### DIFF
--- a/.circleci/generate.sh
+++ b/.circleci/generate.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
@@ -30,12 +30,15 @@ die ()
 
 print_help()
 {
-  echo "Usage: $0 [-f|-p|-a|-e|-i]"
+  echo "Usage: $0 [-f|-p|-a|-e|-i|-b]"
   echo "   -a Generate the config.yml, config.yml.FREE and config.yml.PAID expanded configuration"
   echo "      files from the main config_template.yml reusable configuration file."
   echo "      Use this for permanent changes in config that will be committed to the main repo."
   echo "   -f Generate config.yml for tests compatible with the CircleCI free tier resources"
   echo "   -p Generate config.yml for tests compatible with the CircleCI paid tier resources"
+  echo "   -b Specify the base git branch for comparison when determining changed tests to"
+  echo "      repeat. Defaults to ${BASE_BRANCH}. Note that this option is not used when"
+  echo "      the '-a' option is specified."
   echo "   -e <key=value> Environment variables to be used in the generated config.yml, e.g.:"
   echo "                   -e DTEST_BRANCH=CASSANDRA-8272"
   echo "                   -e DTEST_REPO=https://github.com/adelapena/cassandra-dtest.git"
@@ -73,7 +76,7 @@ paid=false
 env_vars=""
 has_env_vars=false
 check_env_vars=true
-while getopts "e:afpi" opt; do
+while getopts "e:afpib:" opt; do
   case $opt in
       a ) all=true
           ;;
@@ -87,6 +90,8 @@ while getopts "e:afpi" opt; do
             env_vars="$env_vars|$OPTARG"
           fi
           has_env_vars=true
+          ;;
+      b ) BASE_BRANCH="$OPTARG"
           ;;
       i ) check_env_vars=false
           ;;
@@ -173,6 +178,14 @@ fi
 
 # add new or modified tests to the sets of tests to be repeated
 if (!($all)); then
+  # Sanity check that the referenced branch exists
+  if ! git show ${BASE_BRANCH} -- >&/dev/null; then
+    echo -e "\n\nUnknown base branch: ${BASE_BRANCH}. Unable to detect changed tests.\n"
+    echo    "Please use the '-b' option to choose an existing branch name"
+    echo    "(e.g. origin/${BASE_BRANCH}, apache/${BASE_BRANCH}, etc.)."
+    exit 2
+  fi
+
   add_diff_tests ()
   {
     dir="${BASEDIR}/../${2}"


.circleci/generate.sh
```diff
@@ -33,6 +37,10 @@ print_help()
   echo "   -a Generate the default config.yml using low resources and the three templates"
   echo "      (config.yml.LOWRES, config.yml.MIDRES and config.yml.HIGHRES). Use this for"
   echo "      permanent changes in config-2_1.yml that will be committed to the main repo."
+  echo "   -b Specify the base git branch for comparison when determining changed tests to"
+  echo "      repeat. You can also set the BASE_BRANCH environment variable to control this."
+  echo "      Defaults to ${BASE_BRANCH}. Note that this option is not used when the '-a' flag"
+  echo "      is specified."
```
adelapena: I'd put this below the -l/-m/-h flags, so we have the closely related -a/-l/-m/-h together.

.circleci/generate.sh
```diff
@@ -72,10 +80,12 @@ highres=false
 env_vars=""
 has_env_vars=false
 check_env_vars=true
-while getopts "e:almhf" opt; do
+while getopts "e:almhfb:" opt; do
   case $opt in
       a ) all=true
           ;;
+      b ) BASE_BRANCH="$OPTARG"
+          ;;
```
adelapena: As before, I'd put this after -l/-m/-h

.circleci/generate.sh
```diff
@@ -184,36 +194,43 @@ fi
...
-        env_vars=$(echo "$env_vars" | sed -e "s/${1}=/${1}=${test},/")
-      elif [ -z "$env_vars" ]; then
-        env_vars="${1}=${test}"
-      else
-        env_vars="$env_vars|${1}=${test}"
-      fi
-    done
-  }
+  # Sanity check that the referenced branch exists
+  if ! git show ${BASE_BRANCH} -- >&/dev/null; then
```
adelapena, line 182: This isn't POSIX compliant and doesn't work in, for example, Ubuntu's dash.

.circleci/generate.sh
```diff
@@ -184,36 +194,43 @@ fi
...
+    echo "Unknown base branch: ${BASE_BRANCH}. Skipping detection of changed tests."
+    echo "If you wish to detect changed tests, please use the '-b' flag or set the"
+    echo "BASE_BRANCH environment variable to an existing branch name (e.g. origin/${BASE_BRANCH})."
```
adelapena: We could use the `die` function that immediately returns with exit code 1, so we don't have to put the diff code in an `else` branch.

adelapena: Ah, the intention is continuing with the rest of the script, isn't it? I think I'd rather fail and demand the users to provide a base branch, so they can't skip the detection of changed tests. Otherwise, the warning message can easily be overlooked.

adelapena: Perhaps the example, `(e.g. origin/${BASE_BRANCH}).`, could also include some other remote, as in `(e.g. origin/${BASE_BRANCH}, apache/${BASE_BRANCH}, etc.).`

.circleci/generate.sh
```diff
@@ -18,7 +18,11 @@
 #
 
 BASEDIR=`dirname $0`
-BASE_BRANCH=cassandra-3.11
+
+# Allow the user to set an env variable to control this, but default to our version branch
+[ -n "${BASE_BRANCH}" ] || {
```
adelapena: Do we have a use case for setting the branch trough a previously set env var? I think no other options work that way. If we don't have a use case we can save us the additional complexity of having two ways of setting the branch and simplify both the help and error messages.

.circleci/generate.sh
dchenbecker: Ah, sorry, I need to merge the change to switch to bash

.circleci/generate.sh
dchenbecker: Sure, will move

.circleci/generate.sh
dchenbecker: I can remove it. My setup auto-sets env vars for a given work dir, so it's one less flag to have to remember to place

.circleci/generate.sh
dchenbecker: I don't really want it to `print_help` if the branch can't be found, since the message already explains proper usage. I can just `exit 2` here instead.

.circleci/generate.sh
```diff
@@ -184,6 +190,14 @@ fi
 
 # add new or modified tests to the sets of tests to be repeated
 if (!($all)); then
+  # Sanity check that the referenced branch exists
+  if ! git show ${BASE_BRANCH} -- >&/dev/null; then
+    echo -e "\n\nUnknown base branch: ${BASE_BRANCH}. Unable to detect changed tests.\n"
+    echo    "If you wish to detect changed tests, please use the '-b' flag to choose existing"
```
adelapena: Since we are making the detection of tests mandatory, I think that the `If you wish to detect changed tests,` part of the message is not needed anymore, or it should be rephrased. Also, it should probably be `use the '-b' option` instead of `please use the '-b' flag`.

dchenbecker: Fixed in latest commit

.circleci/generate.sh
```diff
@@ -184,6 +190,14 @@ fi
 
 # add new or modified tests to the sets of tests to be repeated
 if (!($all)); then
+  # Sanity check that the referenced branch exists
+  if ! git show ${BASE_BRANCH} -- >&/dev/null; then
+    echo -e "\n\nUnknown base branch: ${BASE_BRANCH}. Unable to detect changed tests.\n"
+    echo    "Please use the '-b' flag to choose an existing branch name"
```
adelapena:
```suggestion
    echo    "Please use the '-b' option to choose an existing branch name"
```

pmcfadin:
Closing as [CASSANDRA-17997](https://issues.apache.org/jira/browse/CASSANDRA-17997) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #1991, opened 2022-11-09
Accord: Transaction Expiration

belliottsmith:


src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -321,7 +315,7 @@ public long latestEpoch()
     }
 
     @Override
-    public Timestamp maxConflict(Keys keys)
+    public Timestamp preaccept(TxnId txnId, Keys keys)
```
aweisberg, line 318: Why does `preaccept` not include the same logic as the implementation in `InMemoryCommandStore` around sometimes returning a new `Timestamp`?

src/java/org/apache/cassandra/service/accord/api/AccordAgent.java
```diff
@@ -49,4 +50,10 @@ public void onHandledException(Throwable throwable)
     {
         // TODO: this
     }
+
+    @Override
+    public boolean isExpired(TxnId initiated, long now)
+    {
+        return false;
```
aweisberg, line 58: No check at all? This is TBD until it can be configured?

aweisberg (changes requested):
Had two questions

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
belliottsmith: Good point. This was a (very) bad rebase.

src/java/org/apache/cassandra/service/accord/api/AccordAgent.java
belliottsmith: Yeah, I think we also want to discriminate between write and read transactions here. I don't mind plumbing it in for now, but it's probably better to leave it until later - it's also not essential for correctness, it only applies to long running original coordinators. I vacillated over whether we should keep it at all, but decided it was probably better.

belliottsmith: As it happens, it looks like the whole Agent is to-be-implemented anyway

aweisberg (approved):
+1 TY

PR #1993, opened 2022-11-10
When updating ThreadLocalRandom used the old sun.misc.Unsafe rather than jdk.internal.misc.Unsafe

dcapwell:


src/java/org/apache/cassandra/utils/reflection/ReflectionUtils.java
```diff
@@ -0,0 +1,59 @@
...
+import java.lang.reflect.Method;
+
+public class ReflectionUtils
+{
+    private ReflectionUtils()
+    {
+
+    }
+
+    public static Field getModifiersField() throws NoSuchFieldException
```
ekaterinadimitrova2: Yup, I suggested this also in my branch for jdk 17 https://github.com/apache/cassandra/commit/4daf0b149bc1524dfb2212ab958f4d45feb79577#diff-929245717cd8c37ec78fb97bddc5b6839b2430c3e32f31f30e4aad5fcd580c71R376-R404 I will put a note to get rid of this in my WIP branch and use this utility class on next rebase :-) Thanks

build.xml
```diff
@@ -153,25 +153,52 @@
         <equals arg1="${ant.java.version}" arg2="1.8"/>
     </condition>
     <condition property="java.version.11">
-        <not><isset property="java.version.8"/></not>
+        <equals arg1="${ant.java.version}" arg2="11"/>
+    </condition>
+    <condition property="java.version.17">
```
ekaterinadimitrova2: why do we add now 17 to build xml?
? It can't even compile with it as Benjamin is not ready with the UDF refactoring.... Also, I already had the changes to make It 11 and 17, double work

build.xml
```diff
@@ -190,9 +217,17 @@
         <string>--add-opens java.base/jdk.internal.module=ALL-UNNAMED</string>
         <string>--add-opens java.base/jdk.internal.util.jar=ALL-UNNAMED</string>
         <string>--add-opens jdk.management/com.sun.management.internal=ALL-UNNAMED</string>
+
+        <string>--add-opens java.base/sun.nio.ch=ALL-UNNAMED</string>
```
ekaterinadimitrova2: Are those justified? I was looking for ways not to add openings.

test/simulator/asm/org/apache/cassandra/simulator/asm/ClassTransformer.java
```diff
@@ -137,12 +139,58 @@ private ClassTransformer(int api, ClassWriter classWriter, String className, Enu
...
+    {
+        this.updateVisibility = updateVisibility;
+    }
+
+    /**
+     * Java 11 changed the way that classes defined in the same source file get access to private state (see https://openjdk.org/jeps/181),
+     * rather than trying to adapt to this, this method attempts to make the field/method/class public so that access
+     * is not restricted.
+     */
+    private int makePublic(int access)
```
ekaterinadimitrova2, line 152: This will be only for the simulator ?

build.xml
belliottsmith: Perhaps we can add these on the simulator launch path only, to avoid taking on new add-opens just to unblock it.

build.xml
belliottsmith: We don't need to do as complete a job as you, as we want to get JDK17 (actually JDK19) supported for the simulator only.

test/simulator/asm/org/apache/cassandra/simulator/asm/ClassTransformer.java
belliottsmith: Yep, this is in the simulator tree so I don't believe it's even available at compilation or runtime to other parts of the codebase.

dcapwell: yes, this is part of the simulator; we rewrite classes so we can control execution order... but on JDK 11 how private fields/methods are shard with subclasses changed, so rather than trying to make things complex, the rewrite logic makes everything public

src/java/org/apache/cassandra/utils/reflection/ReflectionUtils.java
dcapwell: reverted for now as this is for JDK 17

test/simulator/asm/org/apache/cassandra/simulator/asm/ClassTransformer.java
ekaterinadimitrova2: Yup, that's what I thought, thanks. For testing and being only used in the simulator sounds reasonable to me

build.xml
```diff
@@ -153,16 +153,22 @@
...
         <isset property="java.version.8"/>
         <isset property="java.version.11"/>
     </or></not></condition></fail>
 
+    <condition property="arch_x86">
+      <equals arg1="${os.arch}" arg2="x86" />
+    </condition>
+    <condition property="jvm_xss" value="-Xss256k" else="-Xss512k">
+      <isset property="arch_x86" />
+    </condition>
```
ekaterinadimitrova2, line 172: Do we need to add some comment or something for this condition for completeness or I am tired and this will be obvious and I should leave you alone? :D

dcapwell: ill add comment

belliottsmith (approved):

.circleci/config-2_1.yml
```diff
@@ -1164,6 +1164,18 @@ j11_separate_jobs: &j11_separate_jobs
         requires:
           - start_j11_jvm_dtests_vnode_repeat
           - j11_build
+    - start_j11_simulator_dtests:
```
ekaterinadimitrova2, line 1167: Ok, so as we talked in Slack probably we can skip j8 build with j11 runtime and directly go for j11 build with j17 when we do the switch on trunk. I leave the decision to you whether you will add now j8/j11 or not. But don't we need this also added in the pre-commit workflow, same as all other test suites?

build.xml
```diff
@@ -153,16 +153,26 @@
         <equals arg1="${ant.java.version}" arg2="1.8"/>
     </condition>
     <condition property="java.version.11">
+      <!-- This includes every JDK other than 8; so JDK 9 is flagged as JDK 11, and JDK 17 is as well... at the moment this is desired behavior
+      and may be relooked at once JDK 8 support is dropped -->
```
ekaterinadimitrova2, line 157: Thanks for adding this, it will be a good reminder in time as people might miss it!

build.xml
```diff
@@ -224,9 +235,7 @@
     </resources>
     <pathconvert property="_jvm11_test_arg_items_concat" refid="_jvm11_test_arg_items" pathsep=" "/>
     <condition property="_std-test-jvmargs" value="${_jvm11_test_arg_items_concat}" else="${_jvm8_test_arg_items_concat}">
-        <not>
-            <equals arg1="${ant.java.version}" arg2="1.8"/>
-        </not>
+      <isset property="java.version.11" />
```
ekaterinadimitrova2: Why this change was needed?Isn't this the same issue as earlier? I would prefer if we skip changes that are not explicitly needed. Did I miss anything here?

build.xml
```diff
@@ -458,7 +467,7 @@
         <jvmarg value="-Dcassandra.reads.thresholds.coordinator.defensive_checks_enabled=true" /> <!-- enable defensive checks -->
         <jvmarg value="-javaagent:${build.lib}/jamm-${jamm.version}.jar" />
         <jvmarg value="-ea"/>
-        <jvmarg line="${java11-jvmargs}"/>
+        <jvmarg line="${jvmargs}"/>
```
ekaterinadimitrova2: Why this change of name?

test/simulator/asm/org/apache/cassandra/simulator/asm/ClassTransformer.java
```diff
@@ -137,12 +139,58 @@ private ClassTransformer(int api, ClassWriter classWriter, String className, Enu
...
+    {
+        if (!updateVisibility)
+            return access;
+        // leave non-user created methods alone
```
ekaterinadimitrova2: Not only methods? ACC_SYNTHETIC applies also to fields?

ekaterinadimitrova2 (commented):
Overall +1, just some nits. Main questions about CircleCI - adding the job also to the pre-commit even if it still requires approval step. All jobs are in both workflows - the separate jobs one and the precommit one.
And a change in build.xml that I guess it was just forgotten to revert maybe when reverting the other jdk17 stuff.

So I think this needs squash, rebase and final run. We need to test also that if jobs are added or changed, the generate.sh will pick them up from diff as others and they will be pushed in a loop in CI to check for flakies :-) As there were many circle ci config changes (adding all missing jobs) in preparation for 4.1, the circleci files might be annoying. Let me know if you want me to land a hand there

.circleci/config-2_1.yml
dcapwell: > But don't we need this also added in the pre-commit workflow, same as all other test suites?

:sigh: this build is getting too complicated... ill add

> we can skip j8 build with j11 runtime and directly go for j11 build with j17 

for now, since we support 8/11 ill keep the build doing 8/11; when we drop 8 and add 17 we can switch

dcapwell: I wrote a script to confirm the change and found that if I manually applied on my own I would have messed up!

Here is the full diff of the workflows

```
Diff for ('java8_separate_tests', 'java11_separate_tests') is
left:
- start_java_dtest_jars_build
- java_dtest_jars_build
- start_jvm_upgrade_dtests
- java_jvm_upgrade_dtests
- start_jvm_upgrade_dtests_repeat
- java_jvm_upgrade_dtests_repeat
- start_upgrade_dtests
- java_upgrade_dtests
- start_java_cqlsh_tests_offheap
- start_java_upgrade_dtests_repeat
- java_upgrade_dtests_repeat
mismatch:
  java_cqlsh_dtests_py38_offheap:
    left:
    - start_java_cqlsh_tests_offheap
    - java_build
    right:
    - start_java_cqlsh-dtests-offheap
    - java_build
  java_cqlsh_dtests_py3_offheap:
    left:
    - start_java_cqlsh_tests_offheap
    - java_build
    right:
    - start_java_cqlsh-dtests-offheap
    - java_build
right:
- start_java_cqlsh-dtests-offheap

Diff for ('java8_pre-commit_tests', 'java11_pre-commit_tests') is
left:
- start_jvm_upgrade_dtests
- java_dtest_jars_build
- java_jvm_upgrade_dtests
- java_jvm_upgrade_dtests_repeat
- start_upgrade_dtests
- java_upgrade_dtests
- java_upgrade_dtests_repeat
- start_java_cqlsh_dtests_offheap
mismatch:
  java_dtests_offheap_repeat:
    left:
    - start_java_dtests_offheap
    - java_build
    right:
    - start_java_dtests_offheap_repeat
    - java_build
right:
- start_java_dtests_offheap_repeat

Diff for ('java8_separate_tests', 'java8_pre-commit_tests') is
left:
- j8_repeated_ant_test
- j11_repeated_ant_test

Diff for ('java11_separate_tests', 'java11_pre-commit_tests') is
left:
- j11_repeated_ant_test
```

Excluding differences not caused by this patch

```
$ circleci-verify.py --exclude upgrade dtest_jars cqlsh repeated_ant dtests_offheap
$
```

Here is the diff without the latest patch

```
$ circleci-verify.py --exclude upgrade dtest_jars cqlsh repeated_ant dtests_offheap
Diff for ('java8_pre-commit_tests', 'java11_pre-commit_tests') is
left:
- java_simulator_dtests
- java_simulator_dtests_repeat

Diff for ('java11_separate_tests', 'java11_pre-commit_tests') is
left:
- j11_simulator_dtests
- j11_simulator_dtests_repeat

```

test/simulator/asm/org/apache/cassandra/simulator/asm/ClassTransformer.java
dcapwell: updated comment

dcapwell:
> And a change in build.xml that I guess it was just forgotten to revert maybe when reverting the other jdk17 stuff.

nope, build.xml is expected.  `arch_x86` is needed to build on j8 on M1 Macs, and the rest are cleanups (optional); stuff like adding better error message when you run on a unsupported JDK

dcapwell:
> So I think this needs squash, rebase and final run

pushed latest changes

> We need to test also that if jobs are added or changed, the generate.sh will pick them up from diff as others and they will be pushed in a loop in CI to check for flakies 

now that the repeat jobs got deleted by default I don't know how to run them so I never run them anymore... how do you run it now?

build.xml
dcapwell: its not needed, its duplicate work; why reinvent the wheel when we already have one?  I can revert if you desire; its only cleanup while I was here.

build.xml
dcapwell: optional cleanup, can revert if you want; this was because JDK 17 but left as I felt it was fine

.circleci/config-2_1.yml
ekaterinadimitrova2: The issue you caught is naming which is known (tests vs dtests and _ vs -) and planned to be fixed across the board at once in a separate ticket. The only thing I noticed is that one of the repeated tests jobs will always be started in pre-commit if tests to run in a loop are added for that suite which can be made to match the rest, it is inconsistency but not some huge bug considering how many legit bugs(and I do not mean just missing tests) I found and how many tests I added.

`:sigh: this build is getting too complicated... ill add`
Well, we finally have all possible tests in CI and thanks to that we should be able to release. Also, now when it is finally clear that no one anymore uses the HIGHRES as is, we can switch to two configs - paid and free. I am looking forward to those changes. Let's see when people will have the time.

build.xml
ekaterinadimitrova2: Please revert, it is more clear when the version is mentioned, thanks

build.xml
ekaterinadimitrova2: I don't follow...sorry... what is duplicate work?

.circleci/config-2_1.yml
dcapwell: > paid and free. I am looking forward to those changes. Let's see when people will have the time.

doesn't address the comment, but +1 to unify MID and HIGHER to a PAID

build.xml
dcapwell: ill revert

.circleci/config-2_1.yml
ekaterinadimitrova2: [CASSANDRA-18007](https://issues.apache.org/jira/browse/CASSANDRA-18007)
CircleCI improvements and parity - this is an epic which I hope it addresses your concerns, if it doesn't feel free to add whatever is missing to the epic. Thank you in advance

build.xml
```diff
@@ -173,6 +183,7 @@
         <string>-XX:+CMSParallelInitialMarkEnabled</string>
         <string>-XX:+CMSEdenChunksRecordAlways</string>
 
+
```
ekaterinadimitrova2, line 186: Unneeded new line, please, feel free to remove it on commit :-)

test/simulator/test/org/apache/cassandra/simulator/test/ShortPaxosSimulationTest.java
```diff
@@ -40,3 +40,4 @@ public void selfReconcileTest() throws IOException
         PaxosSimulationRunner.main(new String[] { "reconcile", "-n", "3..6", "-t", "1000", "-c", "2", "--cluster-action-limit", "2", "-s", "30", "--with-self" });
     }
 }
+
```
ekaterinadimitrova2, line 43: So this was the change to trigger the repeated tests, thanks. You can remove it on commit :-)

PR #1998, opened 2022-11-11
sstabledump errors when dumping data from index for CASSANDRA-17698

Maxwell-Guo:
sstabledump errors when dumping data from index for CASSANDRA-17698

### DIFF
--- a/src/java/org/apache/cassandra/db/marshal/AbstractCompositeType.java
+++ b/src/java/org/apache/cassandra/db/marshal/AbstractCompositeType.java
@@ -248,6 +248,7 @@ public abstract class AbstractCompositeType extends AbstractType<ByteBuffer>
     @Override
     public String toJSONString(ByteBuffer buffer, int protocolVersion)
     {
+        //TODO support toJSONString for AbstractCompositeType
         throw new UnsupportedOperationException();
     }
 
--- a/src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
+++ b/src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
@@ -88,7 +88,7 @@ public class PartitionerDefinedOrder extends AbstractType<ByteBuffer>
     @Override
     public String toJSONString(ByteBuffer buffer, int protocolVersion)
     {
-        assert baseType != null && !baseType.equals(this) : "The base type is null";
+        assert baseType != null && !baseType.equals(this) : "PartitionerDefinedOrder's toJSONString method need a baseType but now is null or with a not euqal type.";
         return baseType.toJSONString(buffer, protocolVersion);   
     }
 
--- a/src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
+++ b/src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
@@ -88,7 +88,7 @@ public class PartitionerDefinedOrder extends AbstractType<ByteBuffer>
     @Override
     public String toJSONString(ByteBuffer buffer, int protocolVersion)
     {
-        assert baseType != null && !baseType.equals(this);
+        assert baseType != null && !baseType.equals(this) : "The base type is null";
         return baseType.toJSONString(buffer, protocolVersion);   
     }
 
--- a/src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
+++ b/src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
@@ -18,18 +18,13 @@
 package org.apache.cassandra.db.marshal;
 
 import java.nio.ByteBuffer;
-import java.util.Iterator;
-import java.util.List;
 
-import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.cql3.Term;
 import org.apache.cassandra.db.PartitionPosition;
 import org.apache.cassandra.serializers.TypeSerializer;
 import org.apache.cassandra.serializers.MarshalException;
 import org.apache.cassandra.dht.IPartitioner;
 import org.apache.cassandra.utils.ByteBufferUtil;
-import org.apache.cassandra.utils.FBUtilities;
-import org.apache.cassandra.utils.Pair;
 
 /** for sorting columns representing row keys in the row ordering as determined by a partitioner.
  * Not intended for user-defined CFs, and will in fact error out if used with such. */
@@ -57,7 +52,7 @@ public class PartitionerDefinedOrder extends AbstractType<ByteBuffer>
         return parser.getPartitionerDefinedOrder();
     }
     
-    public AbstractType<?> withBaseType(AbstractType<?> baseType) 
+    public AbstractType<?> withBaseType(AbstractType<?> baseType)
     {
         return new PartitionerDefinedOrder(partitioner, baseType);
     }
--- a/src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
+++ b/src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
@@ -19,6 +19,7 @@ package org.apache.cassandra.db.marshal;
 
 import java.nio.ByteBuffer;
 import java.util.Iterator;
+import java.util.List;
 
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.cql3.Term;
@@ -28,29 +29,37 @@ import org.apache.cassandra.serializers.MarshalException;
 import org.apache.cassandra.dht.IPartitioner;
 import org.apache.cassandra.utils.ByteBufferUtil;
 import org.apache.cassandra.utils.FBUtilities;
+import org.apache.cassandra.utils.Pair;
 
 /** for sorting columns representing row keys in the row ordering as determined by a partitioner.
  * Not intended for user-defined CFs, and will in fact error out if used with such. */
 public class PartitionerDefinedOrder extends AbstractType<ByteBuffer>
 {
     private final IPartitioner partitioner;
-
+    private final AbstractType<?> baseType;
+    
     public PartitionerDefinedOrder(IPartitioner partitioner)
     {
         super(ComparisonType.CUSTOM);
         this.partitioner = partitioner;
+        this.baseType = null;
+    }
+
+    public PartitionerDefinedOrder(IPartitioner partitioner, AbstractType<?> baseType)
+    {
+        super(ComparisonType.CUSTOM);
+        this.partitioner = partitioner;
+        this.baseType = baseType;
     }
 
     public static AbstractType<?> getInstance(TypeParser parser)
     {
-        IPartitioner partitioner = DatabaseDescriptor.getPartitioner();
-        Iterator<String> argIterator = parser.getKeyValueParameters().keySet().iterator();
-        if (argIterator.hasNext())
-        {
-            partitioner = FBUtilities.newPartitioner(argIterator.next());
-            assert !argIterator.hasNext();
-        }
-        return partitioner.partitionOrdering();
+        return parser.getPartitionerDefinedOrder();
+    }
+    
+    public AbstractType<?> withBaseType(AbstractType<?> baseType) 
+    {
+        return new PartitionerDefinedOrder(partitioner, baseType);
     }
 
     @Override
@@ -84,7 +93,8 @@ public class PartitionerDefinedOrder extends AbstractType<ByteBuffer>
     @Override
     public String toJSONString(ByteBuffer buffer, int protocolVersion)
     {
-        throw new UnsupportedOperationException();
+        assert baseType != null && !baseType.equals(this);
+        return baseType.toJSONString(buffer, protocolVersion);   
     }
 
     public int compareCustom(ByteBuffer o1, ByteBuffer o2)
@@ -107,6 +117,38 @@ public class PartitionerDefinedOrder extends AbstractType<ByteBuffer>
     @Override
     public String toString()
     {
+        if(baseType != null)
+        {
+            return String.format("%s(%s:%s)", getClass().getName(), partitioner.getClass().getName(), baseType.toString());
+        }
         return String.format("%s(%s)", getClass().getName(), partitioner.getClass().getName());
+    }
+    
+    public AbstractType<?>  getBaseType()
+    {
+        return baseType; 
+    }
+
+    @Override
+    public boolean equals(Object obj)
+    {
+        if (this == obj)
+        {
+            return true;
+        }
+        if (obj instanceof PartitionerDefinedOrder)
+        {
+            PartitionerDefinedOrder other = (PartitionerDefinedOrder) obj;
+            if (baseType == null && other.baseType == null)
+            {
+                return this.partitioner.equals(other.partitioner);
+            }
+            else if (baseType != null && other.baseType != null) 
+            {
+                return this.baseType.equals(other.baseType) && this.partitioner.equals(other.partitioner);
+            }
+            return false;
+        }
+        return false;
     }
 }
--- a/src/java/org/apache/cassandra/db/marshal/TypeParser.java
+++ b/src/java/org/apache/cassandra/db/marshal/TypeParser.java
@@ -21,12 +21,20 @@ import java.lang.reflect.Field;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
 
 import com.google.common.base.Verify;
 import com.google.common.collect.ImmutableMap;
 
-import org.apache.cassandra.exceptions.*;
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.dht.IPartitioner;
+import org.apache.cassandra.exceptions.ConfigurationException;
+import org.apache.cassandra.exceptions.SyntaxException;
 import org.apache.cassandra.utils.ByteBufferUtil;
 import org.apache.cassandra.utils.FBUtilities;
 import org.apache.cassandra.utils.Pair;
@@ -130,6 +138,59 @@ public class TypeParser
             return getAbstractType(name);
     }
 
+
+    /**
+     * parse PartitionOrdering from old version of PartitionOrdering' string format 
+     * */
+    private static  AbstractType<?> defaultParsePartitionOrdering(TypeParser typeParser)
+    {
+        IPartitioner partitioner = DatabaseDescriptor.getPartitioner();
+        Iterator<String> argIterator = typeParser.getKeyValueParameters().keySet().iterator();
+        if (argIterator.hasNext()) {
+            partitioner = FBUtilities.newPartitioner(argIterator.next());
+            assert !argIterator.hasNext();
+        }
+        return partitioner.partitionOrdering();
+    }
+    
+    /**
+     * parse and return the real PartitionerDefinedOrder from the string variable str
+     * the str format can be like PartitionerDefinedOrder(<partitioner>) or
+     * PartitionerDefinedOrder(<partitioner>:<baseType>)
+     * */
+    public AbstractType<?> getPartitionerDefinedOrder()
+    {
+        int initIdx = idx;
+        skipBlank();
+        if (isEOS())
+            return defaultParsePartitionOrdering(this);
+        if (str.charAt(idx) != '(')
+            throw new IllegalStateException();
+        
+        ++idx; // skipping '('
+        skipBlank();
+        
+        String k = readNextIdentifier();
+        IPartitioner partitioner = FBUtilities.newPartitioner(k);
+        skipBlank();
+        if (str.charAt(idx) == ':')
+        {
+            ++idx;
+            skipBlank();
+            // must be PartitionerDefinedOrder 
+            PartitionerDefinedOrder tmp = (PartitionerDefinedOrder) partitioner.partitionOrdering();
+            return tmp.withBaseType(parse());
+        }
+        else if (str.charAt(idx) == ')')
+        {
+            idx = initIdx;
+            // if PartitionerDefinedOrder(<partitioner>) then use the original way of parse partitioner Order
+            // for may exist some place we do not know ? or we can just return partitioner.partitionOrdering() here with not initIdx set 
+            return defaultParsePartitionOrdering(this);
+        }
+        throw new SyntaxException("Syntax error parsing '" + str + ": for msg unexpected character '" + str.charAt(idx) + "'");
+    }
+    
     public Map<String, String> getKeyValueParameters() throws SyntaxException
     {
         if (isEOS())
--- a/src/java/org/apache/cassandra/db/marshal/TypeParser.java
+++ b/src/java/org/apache/cassandra/db/marshal/TypeParser.java
@@ -146,7 +146,8 @@ public class TypeParser
     {
         IPartitioner partitioner = DatabaseDescriptor.getPartitioner();
         Iterator<String> argIterator = typeParser.getKeyValueParameters().keySet().iterator();
-        if (argIterator.hasNext()) {
+        if (argIterator.hasNext()) 
+        {
             partitioner = FBUtilities.newPartitioner(argIterator.next());
             assert !argIterator.hasNext();
         }
--- a/src/java/org/apache/cassandra/db/marshal/TypeParser.java
+++ b/src/java/org/apache/cassandra/db/marshal/TypeParser.java
@@ -138,7 +138,6 @@ public class TypeParser
             return getAbstractType(name);
     }
 
-
     /**
      * parse PartitionOrdering from old version of PartitionOrdering' string format 
      * */
--- a/src/java/org/apache/cassandra/index/internal/CassandraIndex.java
+++ b/src/java/org/apache/cassandra/index/internal/CassandraIndex.java
@@ -766,10 +766,10 @@ public abstract class CassandraIndex implements Index
         ColumnDefinition indexedColumn = target.left;
         AbstractType<?> indexedValueType = utils.getIndexedValueType(indexedColumn);
 
-        AbstractType<?> indexedTablePartitionKeyType =  baseCfsMetadata.partitioner.partitionOrdering();
-        if(indexedTablePartitionKeyType instanceof PartitionerDefinedOrder)
+        AbstractType<?> indexedTablePartitionKeyType = baseCfsMetadata.partitioner.partitionOrdering();
+        if (indexedTablePartitionKeyType instanceof PartitionerDefinedOrder)
         {
-            PartitionerDefinedOrder tmp =  (PartitionerDefinedOrder)indexedTablePartitionKeyType;
+            PartitionerDefinedOrder tmp = (PartitionerDefinedOrder) indexedTablePartitionKeyType;
             indexedTablePartitionKeyType = tmp.withBaseType(baseCfsMetadata.getKeyValidator());
         }
 
--- a/src/java/org/apache/cassandra/index/internal/CassandraIndex.java
+++ b/src/java/org/apache/cassandra/index/internal/CassandraIndex.java
@@ -48,6 +48,7 @@ import org.apache.cassandra.db.lifecycle.View;
 import org.apache.cassandra.db.marshal.AbstractType;
 import org.apache.cassandra.db.marshal.CollectionType;
 import org.apache.cassandra.db.marshal.EmptyType;
+import org.apache.cassandra.db.marshal.PartitionerDefinedOrder;
 import org.apache.cassandra.db.partitions.PartitionIterator;
 import org.apache.cassandra.db.partitions.PartitionUpdate;
 import org.apache.cassandra.db.rows.*;
@@ -765,6 +766,13 @@ public abstract class CassandraIndex implements Index
         ColumnDefinition indexedColumn = target.left;
         AbstractType<?> indexedValueType = utils.getIndexedValueType(indexedColumn);
 
+        AbstractType<?> indexedTablePartitionKeyType =  baseCfsMetadata.partitioner.partitionOrdering();
+        if(indexedTablePartitionKeyType instanceof PartitionerDefinedOrder)
+        {
+            PartitionerDefinedOrder tmp =  (PartitionerDefinedOrder)indexedTablePartitionKeyType;
+            indexedTablePartitionKeyType = tmp.withBaseType(baseCfsMetadata.getKeyValidator());
+        }
+
         // Tables for legacy KEYS indexes are non-compound and dense
         CFMetaData.Builder builder = indexMetadata.isKeys()
                                      ? CFMetaData.Builder.create(baseCfsMetadata.ksName,
@@ -775,8 +783,8 @@ public abstract class CassandraIndex implements Index
 
         builder =  builder.withId(baseCfsMetadata.cfId)
                           .withPartitioner(new LocalPartitioner(indexedValueType))
-                          .addPartitionKey(indexedColumn.name, indexedColumn.type)
-                          .addClusteringColumn("partition_key", baseCfsMetadata.partitioner.partitionOrdering());
+                          .addPartitionKey(indexedColumn.name, utils.getIndexedPartitionKeyType(indexedColumn))
+                          .addClusteringColumn("partition_key", indexedTablePartitionKeyType);
 
         if (indexMetadata.isKeys())
         {
--- a/src/java/org/apache/cassandra/index/internal/CassandraIndexFunctions.java
+++ b/src/java/org/apache/cassandra/index/internal/CassandraIndexFunctions.java
@@ -20,13 +20,22 @@ package org.apache.cassandra.index.internal;
 
 import java.util.List;
 
+import com.datastax.driver.core.ColumnMetadata;
 import org.apache.cassandra.config.CFMetaData;
 import org.apache.cassandra.config.ColumnDefinition;
 import org.apache.cassandra.db.ColumnFamilyStore;
 import org.apache.cassandra.db.marshal.AbstractType;
 import org.apache.cassandra.db.marshal.CollectionType;
 import org.apache.cassandra.db.marshal.CompositeType;
-import org.apache.cassandra.index.internal.composites.*;
+import org.apache.cassandra.db.marshal.ListType;
+import org.apache.cassandra.db.marshal.MapType;
+import org.apache.cassandra.db.marshal.SetType;
+import org.apache.cassandra.index.internal.composites.ClusteringColumnIndex;
+import org.apache.cassandra.index.internal.composites.CollectionEntryIndex;
+import org.apache.cassandra.index.internal.composites.CollectionKeyIndex;
+import org.apache.cassandra.index.internal.composites.CollectionValueIndex;
+import org.apache.cassandra.index.internal.composites.PartitionKeyIndex;
+import org.apache.cassandra.index.internal.composites.RegularColumnIndex;
 import org.apache.cassandra.index.internal.keys.KeysIndex;
 import org.apache.cassandra.schema.IndexMetadata;
 
@@ -47,6 +56,11 @@ public interface CassandraIndexFunctions
      * @return
      */
     default AbstractType<?> getIndexedValueType(ColumnDefinition indexedColumn)
+    {
+        return indexedColumn.type;
+    }
+
+    default AbstractType<?> getIndexedPartitionKeyType(ColumnDefinition indexedColumn)
     {
         return indexedColumn.type;
     }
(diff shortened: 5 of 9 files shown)


src/java/org/apache/cassandra/db/marshal/TypeParser.java
```diff
@@ -130,6 +138,73 @@ public AbstractType<?> parse() throws SyntaxException, ConfigurationException
...
+        }
+        return partitioner.partitionOrdering();
+    }
+    
+    //the format is (partitioner:type)
+    public AbstractType<?> getPartitionerDefinedOrder()
+    {
+        if (isEOS())
+            return defaultParsePartitionOrdering(this);
+        skipBlank();
```
blambov: I would move this before the `isEOS` check.

src/java/org/apache/cassandra/db/marshal/TypeParser.java
```diff
@@ -130,6 +138,73 @@ public AbstractType<?> parse() throws SyntaxException, ConfigurationException
...
+            throw new IllegalStateException();
+        Pair<Boolean, AbstractType<?>>  result = null;
+        ++idx; // skipping '('
+
+        if (str.charAt(idx) == ')')
+        {
+            ++idx;
+            return  defaultParsePartitionOrdering(this);
+        }
+        skipBlank();
```
blambov, line 171: This also makes better sense before the check above.

src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
```diff
@@ -28,29 +29,38 @@
...
-        IPartitioner partitioner = DatabaseDescriptor.getPartitioner();
-        Iterator<String> argIterator = parser.getKeyValueParameters().keySet().iterator();
-        if (argIterator.hasNext())
-        {
-            partitioner = FBUtilities.newPartitioner(argIterator.next());
-            assert !argIterator.hasNext();
-        }
-        return partitioner.partitionOrdering();
+        TypeParser clone = parser.clone();
+        return clone.getPartitionerDefinedOrder();
```
blambov: There shouldn't be a need for `clone` here. (The clone might even break something, in the unlikely case of this being part of a composite/tuple.)

src/java/org/apache/cassandra/db/marshal/TypeParser.java
```diff
@@ -130,6 +138,73 @@ public AbstractType<?> parse() throws SyntaxException, ConfigurationException
...
+        String k = readNextIdentifier();
+        skipBlank();
+        if (str.charAt(idx) == ':')
+        {
+            ++idx;
+            skipBlank();
+        }
+        else if (str.charAt(idx) != ',' && str.charAt(idx) != ')')
```
blambov: Shouldn't we throw an error if the next part is neither a separator nor a closing bracket?

src/java/org/apache/cassandra/db/marshal/TypeParser.java
```diff
@@ -130,6 +138,73 @@ public AbstractType<?> parse() throws SyntaxException, ConfigurationException
...
+        IPartitioner partitioner = DatabaseDescriptor.getPartitioner();
+        Iterator<String> argIterator = typeParser.getKeyValueParameters().keySet().iterator();
+        if (argIterator.hasNext()) {
+            partitioner = FBUtilities.newPartitioner(argIterator.next());
+            assert !argIterator.hasNext();
+        }
+        return partitioner.partitionOrdering();
+    }
+    
+    //the format is (partitioner:type)
```
blambov: Could you expand on this? The actual format appears to permit more, at least `PartitionerDefinedOrder`, `PartitionerDefinedOrder(<partitioner>)` and `PartitionerDefinedOrder(<partitioner>:<baseType>)`.

src/java/org/apache/cassandra/db/marshal/TypeParser.java
```diff
@@ -130,6 +138,73 @@ public AbstractType<?> parse() throws SyntaxException, ConfigurationException
...
+        skipBlank();
+        if (str.charAt(idx) != '(')
+            throw new IllegalStateException();
+        Pair<Boolean, AbstractType<?>>  result = null;
+        ++idx; // skipping '('
+
+        if (str.charAt(idx) == ')')
+        {
+            ++idx;
+            return  defaultParsePartitionOrdering(this);
```
blambov: Couldn't we just use `DatabaseDescriptor.getPartitioner().partitionerOrdering()` here?

src/java/org/apache/cassandra/db/marshal/TypeParser.java
```diff
@@ -130,6 +138,73 @@ public AbstractType<?> parse() throws SyntaxException, ConfigurationException
...
+        {
+            PartitionerDefinedOrder tmp = (PartitionerDefinedOrder) partitioner.partitionOrdering();
+            ++idx;
+            try
+            {
+                type = tmp.withBaseType(parse());
+            }
+            catch (Throwable throwable)
+            {
+                Iterator<String> argIterator = this.getKeyValueParameters().keySet().iterator();
```
blambov: Shouldn't we just throw an error in this case?

src/java/org/apache/cassandra/db/marshal/TypeParser.java
```diff
@@ -130,6 +138,73 @@ public AbstractType<?> parse() throws SyntaxException, ConfigurationException
...
+            ++idx;
+            skipBlank();
+        }
+        else if (str.charAt(idx) != ',' && str.charAt(idx) != ')')
+        {
+            return defaultParsePartitionOrdering(this);
+        }
+        IPartitioner partitioner = FBUtilities.newPartitioner(k);
+        AbstractType<?> type = partitioner.partitionOrdering();
+        if (partitioner.partitionOrdering() instanceof PartitionerDefinedOrder)
```
blambov: Looks like this test should be checking if a `:` was seen instead.

test/unit/org/apache/cassandra/db/marshal/TypeParserTest.java
```diff
@@ -98,4 +98,24 @@ public void testParsePartitionerOrder() throws ConfigurationException, SyntaxExc
...
+        for (IPartitioner partitioner: new IPartitioner[] { Murmur3Partitioner.instance,
+                                                            ByteOrderedPartitioner.instance,
+                                                            RandomPartitioner.instance,
+                                                            OrderPreservingPartitioner.instance })
+        {
+            AbstractType<?> type = partitioner.partitionOrdering();
+            if (type instanceof PartitionerDefinedOrder)
+            {
+                PartitionerDefinedOrder tmp = (PartitionerDefinedOrder) type;
+                tmp.withBaseType(Int32Type.instance);
```
blambov: I believe this should be `type = ...`.

test/unit/org/apache/cassandra/db/marshal/TypeParserTest.java
```diff
@@ -98,4 +98,24 @@ public void testParsePartitionerOrder() throws ConfigurationException, SyntaxExc
...
+            }
+            System.out.println(type.toString());
+            assertSame(type, TypeParser.parse(type.toString()));
+        }
+        assertSame(DatabaseDescriptor.getPartitioner().partitionOrdering(), TypeParser.parse("PartitionerDefinedOrder"));
+    }
```
blambov, line 202: It would be helpful to test some explicit definitions to try all acceptable variations (default partitioner, partitioner without base type, partitioner with base type).

blambov (changes requested):

src/java/org/apache/cassandra/db/marshal/TypeParser.java
Maxwell-Guo: just want to using the old format type parse of PartitionerDefinedOrder to do a double check parse. 
if the string of type is %s(%s)

src/java/org/apache/cassandra/db/marshal/TypeParser.java
Maxwell-Guo: just want to using the old format type parse of PartitionerDefinedOrder to do a double check parse.
if the string of type is %s(%s)

test/unit/org/apache/cassandra/db/marshal/TypeParserTest.java
Maxwell-Guo: @blambov besides ,I have a question here ,why we use assertSame here ? for this patch if it is a PartitionerDefinedOrder  with a base type there will return a new PartitionerDefinedOrder object here and the memory address must be different . 
So I want to know if the place in the codebase that invoked PartitionerDefinedOrder class must keep the memory address be the same ?

src/java/org/apache/cassandra/db/marshal/TypeParser.java
Maxwell-Guo: +1

src/java/org/apache/cassandra/db/marshal/TypeParser.java
Maxwell-Guo: yes ,throw an error may be more efficient

test/unit/org/apache/cassandra/db/marshal/TypeParserTest.java
Maxwell-Guo: yes, it is a bug I have write ,fix it now.

src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
Maxwell-Guo: +1

test/unit/org/apache/cassandra/db/marshal/TypeParserTest.java
Maxwell-Guo: yes

test/unit/org/apache/cassandra/db/marshal/TypeParserTest.java
blambov: Many implementations of `AbstractType` keep a map of already created types to ensure that they don't create the same object more than once. See e.g. `SetType`. It makes some sense to do the same for `PartitionerDefinedOrder` as well, but it is not something I personally insist on -- either change the tests to check equality, or add a static map of instances and consult it when constructing a type.

src/java/org/apache/cassandra/db/marshal/TypeParser.java
Maxwell-Guo: ok

src/java/org/apache/cassandra/db/marshal/TypeParser.java
Maxwell-Guo: ok

src/java/org/apache/cassandra/db/marshal/TypeParser.java
Maxwell-Guo: @blambov 
 It may no need to do checking.
There will be two  possibilities when arrive here,1.PartitionerDefinedOrder(<partitioner>), then the final results will returned at the catch section from line 194 to line 205 .the partitioner type is get from k
2.PartitionerDefinedOrder(<partitioner>:<baseType>),when arriving here the base type can be parsed with function parse() at line 192 ,also the partitioner type is get with k.
For ByteOrderedPartitioner AND OrderPreservingPartitioner ,their partitionOrdering() will return ByteType.Instance and UTF8Type.Instance so we should only do for Random and mum3

src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
```diff
@@ -107,6 +117,30 @@ public TypeSerializer<ByteBuffer> getSerializer()
...
+    public boolean equals(Object obj)
+    {
+        if (this == obj)
+        {
+            return true;
+        }
+        if (obj instanceof PartitionerDefinedOrder)
+        {
+            PartitionerDefinedOrder other = (PartitionerDefinedOrder) obj;
+            return this.baseType.equals(other.baseType) && this.partitioner.equals(other.partitioner);
```
blambov: Does this not throw when `baseType` is null?

src/java/org/apache/cassandra/db/marshal/TypeParser.java
blambov: In this case we have processed "PartitionerDefinedOrder()". Do we expect other parameters after the closing bracket?

src/java/org/apache/cassandra/db/marshal/TypeParser.java
blambov: Why do we accept a comma?

src/java/org/apache/cassandra/db/marshal/TypeParser.java
blambov: I see what you mean, and it makes sense if we want to preserve existing poor parsing behavior.

However, I would treat an appearance of e.g. `PartitionerDefinedOrder(ByteOrderedPartitioner)` as an error, because in that case there exists an underlying type that can and will be used directly. If the type is specified as `PartitionerDefinedOrder`, the result of this construction must be an instance of that.

The strings we are parsing here are generated internally; I don't believe the user has any control over them and thus can't construct invalid-but-acceptable definitions such as this.

src/java/org/apache/cassandra/db/marshal/TypeParser.java
blambov: "PartitionerDefinedOrder(%s)" can be handled more easily by just checking if `:` was seen. If not, just return `type`. If yes, then `((PartitionerDefinedOrder)type).withBaseType(parse())` must succeed.

Are you trying to preserve existing bad behavior of the parser (e.g. accepting string such as "PartitionerDefinedOrder(Murmur3Partitioner=something)")?

src/java/org/apache/cassandra/db/marshal/TypeParser.java
Maxwell-Guo: for line 176 we processed case: PartitionerDefinedOrder(<partitioner>) and We do not expect other parameters after closing bracket.

src/java/org/apache/cassandra/db/marshal/TypeParser.java
Maxwell-Guo: Yes, a comma is not need here, remove it now .

src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
Maxwell-Guo: yes, null should be added

src/java/org/apache/cassandra/db/marshal/TypeParser.java
Maxwell-Guo: Some modification is made for this review. This time I just use what you have said. ""PartitionerDefinedOrder(%s)" can be handled more easily by just checking if : was seen. If not, just return type. If yes, then ((PartitionerDefinedOrder)type).withBaseType(parse()) must succeed."
For me ,at first I am not sure of the existence of different type's string to be parse. So I re-use the original  way of Typeparse, so if the string that is not in the format of PartitionerDefinedOrder(<partitioner>:<baseType>) , the original parse will be used.

src/java/org/apache/cassandra/db/marshal/TypeParser.java
Maxwell-Guo: I modify the code , and left some comment around this place . For format "PartitionerDefinedOrder(partitioner)" I set the idx with the initIdx and  re-use the original way of parse here . But at first a partitioner.partitionOrdering() was just retrun . Finally I change to the original way of parse for I am not sure if some other corner case may exist?

src/java/org/apache/cassandra/index/internal/CassandraIndex.java
```diff
@@ -765,6 +766,13 @@ public static final CFMetaData indexCfsMetadata(CFMetaData baseCfsMetadata, Inde
         ColumnDefinition indexedColumn = target.left;
         AbstractType<?> indexedValueType = utils.getIndexedValueType(indexedColumn);
 
+        AbstractType<?> indexTableMokePkType =  baseCfsMetadata.partitioner.partitionOrdering();
```
blambov: Nit: `MokePk` is rather unclear. Maybe call this `indexedTablePartitionKeyType`?

src/java/org/apache/cassandra/tools/JsonTransformer.java
```diff
@@ -349,7 +350,12 @@ private void serializeClustering(ClusteringPrefix clustering) throws IOException
                 }
                 else
                 {
-                    json.writeRawValue(column.cellValueType().toJSONString(clustering.get(i), Server.CURRENT_VERSION));
+                    AbstractType<?> type = column.cellValueType();
+                    if (type instanceof PartitionerDefinedOrder)
```
blambov: Is this special case necessary when `PartitionerDefinedOrder` implements `toJSONString`?

blambov:
The latest parser code looks much better, thank you for the patience.

src/java/org/apache/cassandra/index/internal/CassandraIndex.java
Maxwell-Guo: ok

src/java/org/apache/cassandra/tools/JsonTransformer.java
Maxwell-Guo: yeah，I think this should remove for PartitionerDefinedOrder have already got toJSONString

blambov (approved):
LGTM

src/java/org/apache/cassandra/index/internal/CassandraIndexFunctions.java
```diff
@@ -20,13 +20,22 @@
 
 import java.util.List;
 
+import com.datastax.driver.core.ColumnMetadata;
```
adelapena: Nit: unused import

src/java/org/apache/cassandra/index/internal/CassandraIndexFunctions.java
```diff
@@ -142,6 +156,22 @@ public AbstractType<?> getIndexedValueType(ColumnDefinition indexedColumn)
...
+            assert indexedColumn.type.isCollection() ;
+            switch (((CollectionType)indexedColumn.type).kind)
+            {
+                case LIST:
+                    return ((ListType)indexedColumn.type).getElementsType();
+                case SET:
+                    return ((SetType)indexedColumn.type).getElementsType();
+                case MAP:
+                    return ((MapType)indexedColumn.type).getKeysType();
+            }
```
adelapena:
```suggestion
            switch (((CollectionType<?>) indexedColumn.type).kind)
            {
                case LIST:
                    return ((ListType<?>) indexedColumn.type).getElementsType();
                case SET:
                    return ((SetType<?>) indexedColumn.type).getElementsType();
                case MAP:
                    return ((MapType<?, ?>) indexedColumn.type).getKeysType();
            }
```

src/java/org/apache/cassandra/index/internal/CassandraIndexFunctions.java
```diff
@@ -168,6 +198,24 @@ public CFMetaData.Builder addIndexClusteringColumns(CFMetaData.Builder builder,
...
+            assert indexedColumn.type.isCollection() ;
+            switch (((CollectionType)indexedColumn.type).kind)
+            {
+                case LIST:
+                    return ((ListType)indexedColumn.type).getElementsType();
+                case SET:
+                    return ((SetType)indexedColumn.type).getElementsType();
+                case MAP:
+                    return ((MapType)indexedColumn.type).getValuesType();
+            }
```
adelapena:
```suggestion
            switch (((CollectionType<?>) indexedColumn.type).kind)
            {
                case LIST:
                    return ((ListType<?>) indexedColumn.type).getElementsType();
                case SET:
                    return ((SetType<?>) indexedColumn.type).getElementsType();
                case MAP:
                    return ((MapType<?, ?>) indexedColumn.type).getValuesType();
            }
```

src/java/org/apache/cassandra/index/internal/CassandraIndex.java
```diff
@@ -765,6 +766,13 @@ public static final CFMetaData indexCfsMetadata(CFMetaData baseCfsMetadata, Inde
         ColumnDefinition indexedColumn = target.left;
         AbstractType<?> indexedValueType = utils.getIndexedValueType(indexedColumn);
 
+        AbstractType<?> indexedTablePartitionKeyType =  baseCfsMetadata.partitioner.partitionOrdering();
+        if(indexedTablePartitionKeyType instanceof PartitionerDefinedOrder)
```
adelapena:
```suggestion
        if (indexedTablePartitionKeyType instanceof PartitionerDefinedOrder)
```

src/java/org/apache/cassandra/index/internal/CassandraIndex.java
```diff
@@ -765,6 +766,13 @@ public static final CFMetaData indexCfsMetadata(CFMetaData baseCfsMetadata, Inde
...
+        {
+            PartitionerDefinedOrder tmp =  (PartitionerDefinedOrder)indexedTablePartitionKeyType;
```
adelapena:
```suggestion
            PartitionerDefinedOrder tmp = (PartitionerDefinedOrder) indexedTablePartitionKeyType;
```

src/java/org/apache/cassandra/db/marshal/TypeParser.java
```diff
@@ -130,6 +138,59 @@ public AbstractType<?> parse() throws SyntaxException, ConfigurationException
             return getAbstractType(name);
     }
 
+    /**
+     * parse PartitionOrdering from old version of PartitionOrdering' string format 
+     * */
+    private static  AbstractType<?> defaultParsePartitionOrdering(TypeParser typeParser)
```
adelapena, line 144:
```suggestion
     * Parse PartitionOrdering from old version of PartitionOrdering' string format.
     */
    private static AbstractType<?> defaultParsePartitionOrdering(TypeParser typeParser)
```

src/java/org/apache/cassandra/db/marshal/TypeParser.java
```diff
@@ -130,6 +138,59 @@ public AbstractType<?> parse() throws SyntaxException, ConfigurationException
...
+            assert !argIterator.hasNext();
+        }
+        return partitioner.partitionOrdering();
+    }
+    
+    /**
+     * parse and return the real PartitionerDefinedOrder from the string variable str
+     * the str format can be like PartitionerDefinedOrder(<partitioner>) or
+     * PartitionerDefinedOrder(<partitioner>:<baseType>)
+     * */
```
adelapena, line 160:
```suggestion
     * Parse and return the real {@link PartitionerDefinedOrder} from the string variable str
     * the str format can be like {@link PartitionerDefinedOrder#PartitionerDefinedOrder(IPartitioner)} or
     * {@link PartitionerDefinedOrder#PartitionerDefinedOrder(IPartitioner, AbstractType)}.
     */
```

src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
```diff
@@ -18,39 +18,43 @@
...
-import org.apache.cassandra.utils.FBUtilities;
 
 /** for sorting columns representing row keys in the row ordering as determined by a partitioner.
  * Not intended for user-defined CFs, and will in fact error out if used with such. */
 public class PartitionerDefinedOrder extends AbstractType<ByteBuffer>
 {
     private final IPartitioner partitioner;
-
+    private final AbstractType<?> baseType;
+
```
adelapena, line 35: Nit: extra whitespaces
```suggestion

```

src/java/org/apache/cassandra/db/marshal/AbstractCompositeType.java
```diff
@@ -248,6 +248,7 @@ public Term fromJSONObject(Object parsed)
     @Override
     public String toJSONString(ByteBuffer buffer, int protocolVersion)
     {
+        //TODO support toJSONString for AbstractCompositeType
```
adelapena, line 251:
```suggestion
        // TODO support toJSONString for AbstractCompositeType
```

test/unit/org/apache/cassandra/db/marshal/TypeParserTest.java
```diff
@@ -98,4 +102,173 @@ public void testParsePartitionerOrder() throws ConfigurationException, SyntaxExc
         }
         assertSame(DatabaseDescriptor.getPartitioner().partitionOrdering(), TypeParser.parse("PartitionerDefinedOrder"));
     }
+
+    @Test
+    public void testParsePartitionerOrderWithBaseType() throws ConfigurationException, SyntaxException
```
adelapena: This doesn't throw `ConfigurationException` nor `SyntaxException`. Also, there is some duplication in the method that can be mitigated with an auxiliary function:
```java
    @Test
    public void testParsePartitionerOrderWithBaseType()
    {
        // default partitioner
        assertSame(DatabaseDescriptor.getPartitioner().partitionOrdering(), TypeParser.parse("PartitionerDefinedOrder"));
        
        // PartitionerDefinedOrder's base type is not composite type
        testParsePartitionerOrderWithBaseType(Int32Type.instance);

        // PartitionerDefinedOrder's base type is composite type
        testParsePartitionerOrderWithBaseType(CompositeType.getInstance(Int32Type.instance, UTF8Type.instance));

        // PartitionerDefinedOrder's base type is tuple type
        testParsePartitionerOrderWithBaseType(new TupleType(Lists.newArrayList(Int32Type.instance, UTF8Type.instance)));

        // PartitionerDefinedOrder's base type is ReversedType
        testParsePartitionerOrderWithBaseType(ReversedType.getInstance(Int32Type.instance));

        // PartitionerDefinedOrder's base type is CollectionType
        testParsePartitionerOrderWithBaseType(MapType.getInstance(Int32Type.instance, UTF8Type.instance, false));
    }

    private static void testParsePartitionerOrderWithBaseType(AbstractType<?> baseType)
    {
        for (IPartitioner partitioner: new IPartitioner[] { Murmur3Partitioner.instance,
                                                            ByteOrderedPartitioner.instance,
                                                            RandomPartitioner.instance,
                                                            OrderPreservingPartitioner.instance })
        {
            AbstractType<?> type = partitioner.partitionOrdering();
            if (type instanceof PartitionerDefinedOrder)
            {
                type = ((PartitionerDefinedOrder) type).withBaseType(baseType);
            }
            assertEquals(type, TypeParser.parse(type.toString()));
        }
    }
```

test/unit/org/apache/cassandra/db/marshal/TypeParserTest.java
```diff
@@ -98,4 +102,173 @@ public void testParsePartitionerOrder() throws ConfigurationException, SyntaxExc
...
+                PartitionerDefinedOrder tmp = (PartitionerDefinedOrder) type;
+                type = tmp.withBaseType(MapType.getInstance(Int32Type.instance, UTF8Type.instance, false));
+            }
+            System.out.println(type.toString());
+            assertEquals(type, TypeParser.parse(type.toString()));
+        }
+    }
+
+    @Test
+    public void testParsePartitionerOrderMistMatch() throws ConfigurationException, SyntaxException
```
adelapena: This doesn't throw `ConfigurationException` nor `SyntaxException`.

test/unit/org/apache/cassandra/db/marshal/TypeParserTest.java
```diff
@@ -98,4 +102,173 @@ public void testParsePartitionerOrder() throws ConfigurationException, SyntaxExc
...
+                String tmpStr1 =  type.toString().replace(')', ',');
+                try
+                {
+                    TypeParser.parse(tmpStr1);
+                    fail();
+                }
+                catch (Throwable t)
+                {
+                    String tmp = t.getCause().getMessage();
+                    System.out.println(tmp);
```
adelapena: I'd remove the calls to `System.out.println`

src/java/org/apache/cassandra/tools/SSTableExport.java
```diff
@@ -119,6 +122,14 @@ public static CFMetaData metadataFromSSTable(Descriptor desc) throws IOException
         {
             builder.addClusteringColumn("clustering" + (i > 0 ? i : ""), header.getClusteringTypes().get(i));
         }
+
+        if (SecondaryIndexManager.isIndexColumnFamily(desc.cfname))
+        {
+            String index = SecondaryIndexManager.getIndexName(desc.cfname);
+            IndexMetadata indexMetadata =  IndexMetadata.fromSchemaMetadata(index, IndexMetadata.Kind.CUSTOM, null);
```
adelapena:
```suggestion
            IndexMetadata indexMetadata = IndexMetadata.fromSchemaMetadata(index, IndexMetadata.Kind.CUSTOM, null);
```

adelapena: We don't really know if the index is `CUSTOM`, but in principle that doesn't seem relevant for this code path. Nevertheless I'd add a comment about it.

src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
```diff
@@ -84,7 +88,8 @@ public Term fromJSONObject(Object parsed)
     @Override
     public String toJSONString(ByteBuffer buffer, int protocolVersion)
     {
-        throw new UnsupportedOperationException();
+        assert baseType != null && !baseType.equals(this);
```
adelapena: This assertion fails if we try to read a sstable that has been written before applying this patch

src/java/org/apache/cassandra/tools/SSTableExport.java
Maxwell-Guo: Yes, you are right, first time I do not know what kind of the index should be ,so I just set it with COUSTOM as I think this do not make any sense for the code path. But I think I can now get the index meta info from system table and return the finally result.

Maxwell-Guo: @adelapena .It seems hard for this tool to open a query for cql to select the index meta system table. So any advice on the kind of index ?

src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
Maxwell-Guo: if reading from a sstable generated from the version that do not contains this patch and dump index sstable using version contains this patch then the baseType will be null and a java.lang.AssertionError will return at this line . 

if using a version that do not contains this patch , then an UnsupportedOperationException will be throw.

Maxwell-Guo: Add some tests for this

src/java/org/apache/cassandra/tools/SSTableExport.java
adelapena: I think that the kind of index doesn't have consequences for the sstable export, and I don't see a way to get the kind of index from the sstable. So it's probably ok to leave it as it is but, since it's not strictly correct, I'd leave a brief comment on the code about it.

test/unit/org/apache/cassandra/db/marshal/TypeParserTest.java
```diff
@@ -98,4 +102,125 @@ public void testParsePartitionerOrder() throws ConfigurationException, SyntaxExc
...
+                // error format PartitionerDefinedOrder(org.apache.cassandra.dht.Murmur3Partitioner,
+                String tmpStr1 =  type.toString().replace(')', ',');
+                try
+                {
+                    TypeParser.parse(tmpStr1);
+                    fail();
+                }
+                catch (Throwable t)
+                {
+                    String tmp = t.getCause().getMessage();
```
adelapena: Unused var, the message is taken from the exception again in the line right below.

test/unit/org/apache/cassandra/db/marshal/TypeParserTest.java
```diff
@@ -98,4 +102,125 @@ public void testParsePartitionerOrder() throws ConfigurationException, SyntaxExc
...
+                String tmpStr3 =  type.toString().replace(')', ':');
+                try
+                {
+                    TypeParser.parse(tmpStr3);
+                    fail();
+                }
+                catch (Throwable t)
+                {
+                    String tmp = t.getCause().getMessage();
+                    System.out.println(tmp);
```
adelapena: I'd remove the `println`

test/unit/org/apache/cassandra/db/marshal/TypeParserTest.java
```diff
@@ -98,4 +102,125 @@ public void testParsePartitionerOrder() throws ConfigurationException, SyntaxExc
...
+                                                            RandomPartitioner.instance,
+                                                            OrderPreservingPartitioner.instance })
+        {
+            AbstractType<?> type = partitioner.partitionOrdering();
+            if (type instanceof PartitionerDefinedOrder)
+            {
+                PartitionerDefinedOrder tmp = (PartitionerDefinedOrder) type;
+                type = tmp.withBaseType(baseType);
+            }
+            System.out.println(type.toString());
```
adelapena: I'd remove the `println`

test/unit/org/apache/cassandra/db/marshal/PartitionerDefinedOrderTest.java
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.cassandra.utils.ByteBufferUtil;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+public class PartitionerDefinedOrderTest
+{
+    private String key = "key";
+    private AbstractType type = UTF8Type.instance;
```
adelapena:
```suggestion
    private static final String key = "key";
    private static final AbstractType<?> type = UTF8Type.instance;
```

test/unit/org/apache/cassandra/db/marshal/PartitionerDefinedOrderTest.java
```diff
@@ -0,0 +1,80 @@
...
+    
+    @Test
+    public void testToJsonStringWithBaseType()
+    {
+        for (IPartitioner partitioner: new IPartitioner[] { Murmur3Partitioner.instance,
+                                                            ByteOrderedPartitioner.instance,
+                                                            RandomPartitioner.instance,
+                                                            OrderPreservingPartitioner.instance })
+        {
+            if(partitioner.partitionOrdering() instanceof PartitionerDefinedOrder)
```
adelapena:
```suggestion
            if (partitioner.partitionOrdering() instanceof PartitionerDefinedOrder)
```

test/unit/org/apache/cassandra/db/marshal/PartitionerDefinedOrderTest.java
```diff
@@ -0,0 +1,80 @@
...
+    
+    @Test
+    public void testToJsonStringWithOutBaseType()
+    {
+        for (IPartitioner partitioner: new IPartitioner[] { Murmur3Partitioner.instance,
+                                                            ByteOrderedPartitioner.instance,
+                                                            RandomPartitioner.instance,
+                                                            OrderPreservingPartitioner.instance })
+        {
+            if(partitioner.partitionOrdering() instanceof PartitionerDefinedOrder)
```
adelapena:
```suggestion
            if (partitioner.partitionOrdering() instanceof PartitionerDefinedOrder)
```

test/unit/org/apache/cassandra/db/marshal/PartitionerDefinedOrderTest.java
```diff
@@ -0,0 +1,80 @@
...
+            {
+                PartitionerDefinedOrder partitionerDefinedOrder = (PartitionerDefinedOrder) partitioner.partitionOrdering();
+                assertEquals(null, partitionerDefinedOrder.getBaseType());
```
adelapena:
```suggestion
                assertNull(partitionerDefinedOrder.getBaseType());
```

test/unit/org/apache/cassandra/db/marshal/PartitionerDefinedOrderTest.java
```diff
@@ -0,0 +1,80 @@
...
+                try
+                {
+                    partitionerDefinedOrder.toJSONString(ByteBufferUtil.bytes(key), 4);
+                    fail();
+                }
+                catch (Throwable throwable)
+                {
+                    assertTrue(throwable.getMessage().contains("The base type is null"));
+                }
```
adelapena: Nit: we can simplify a bit with:
```java
Assertions.assertThatThrownBy(() -> partitionerDefinedOrder.toJSONString(ByteBufferUtil.bytes(key), 4))
          .hasMessageContaining("The base type is null");
```

test/unit/org/apache/cassandra/db/marshal/PartitionerDefinedOrderTest.java
```diff
@@ -0,0 +1,80 @@
...
+    {
+        for (IPartitioner partitioner: new IPartitioner[] { Murmur3Partitioner.instance,
+                                                            ByteOrderedPartitioner.instance,
+                                                            RandomPartitioner.instance,
+                                                            OrderPreservingPartitioner.instance })
+        {
+            if(partitioner.partitionOrdering() instanceof PartitionerDefinedOrder)
+            {
+                PartitionerDefinedOrder partitionerDefinedOrder = (PartitionerDefinedOrder) partitioner.partitionOrdering();
+                String jsonString = partitionerDefinedOrder.withBaseType(type).toJSONString(ByteBufferUtil.bytes(key), 4);
```
adelapena:
```suggestion
                String jsonString = partitionerDefinedOrder.withBaseType(type).toJSONString(UTF8Type.instance.decompose(key), 4);
```

src/java/org/apache/cassandra/db/marshal/PartitionerDefinedOrder.java
adelapena: Perhaps we should leave a comment mentioning this ticket, or extend the error message, so anyone hitting this know what's going on. That way, anyone trying to export a sstable that wasn't generated with the fixed version of C* will be aware of the situation. Alternatively, I wonder if it would make sense to print the serialized value of the clustering instead of failing. @smiklosovic @blambov wdyt?

Maxwell-Guo: +1

Maxwell-Guo: print serialized value seems not easy, for in PartitionerDefinedOrder class, lack of really data type. When  using fixed code to export older sstable.

PR #2004, opened 2022-11-11
sstabledump errors when dumping data from index for CASSANDRA-17698

Maxwell-Guo:
4.1

### DIFF
--- a/src/java/org/apache/cassandra/index/internal/CassandraIndexFunctions.java
+++ b/src/java/org/apache/cassandra/index/internal/CassandraIndexFunctions.java
@@ -18,17 +18,25 @@
 
 package org.apache.cassandra.index.internal;
 
-import java.util.List;
-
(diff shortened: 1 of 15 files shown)


src/java/org/apache/cassandra/index/internal/CassandraIndexFunctions.java
```diff
@@ -51,6 +59,8 @@ default AbstractType<?> getIndexedValueType(ColumnMetadata indexedColumn)
         return indexedColumn.type;
     }
 
+    public AbstractType<?> getIndexedPartitionKeyType(ColumnMetadata indexedColumn);
```
Claudenw, line 62: Add a default implementation for the standard cases and only override when the return values i not indexedColumn.type.

Claudenw (changes requested):

PR #2008, opened 2022-11-12
Restore custom param types over messaging system

michaelsembwever:


michaelsembwever:
@piotrmaslanka , would you mind trying this out with your work with jaeger. You can now add do
```
# add span
messageBuilder.withCustomParam("JAEGER", myJaegerSpan);
```
and
```
# get span
myJaegerSpan = message.customParams.get("JAEGER");
```

src/java/org/apache/cassandra/net/CustomParamsSerializer.java
```diff
@@ -0,0 +1,72 @@
...
+            out.writeUTF(e.getKey());
+            ByteBufferUtil.writeWithLength(e.getValue(), out);
+        }
+    }
+
+    @Override
+    public Map<String, ByteBuffer> deserialize(DataInputPlus in, int version) throws IOException
+    {
+        int entries = in.readInt();
+        Map<String, ByteBuffer> customParams = new HashMap<>(entries);
```
iamaleksey: You want to use `Maps.newHashMapWithExpectedSize()` here, to take load factor into account.

src/java/org/apache/cassandra/net/CustomParamsSerializer.java
```diff
@@ -0,0 +1,72 @@
...
+        }
+        return customParams;
+    }
+
+    @Override
+    public long serializedSize(Map<String, ByteBuffer> t, int version)
+    {
+        return TypeSizes.INT_SIZE
+                + t.keySet().stream().collect(Collectors.summingInt((k) -> TypeSizes.sizeof(k)))
+                + t.values().stream().collect(Collectors.summingInt((v) -> TypeSizes.sizeofWithLength(v)));
```
iamaleksey: No need for two streams here. One regular for-loop over the entry set will do better and read no less clearly, *plus* mimic `serialize()` more closely, which is a good practice for the method couple that needs to be kept in-sync.

src/java/org/apache/cassandra/net/CustomParamsSerializer.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.stream.Collectors;
+
+import org.apache.cassandra.db.TypeSizes;
+import org.apache.cassandra.io.IVersionedSerializer;
+import org.apache.cassandra.io.util.DataInputPlus;
+import org.apache.cassandra.io.util.DataOutputPlus;
+import org.apache.cassandra.utils.ByteBufferUtil;
+
+
+class CustomParamsSerializer implements IVersionedSerializer<Map<String,ByteBuffer>>
```
iamaleksey: Would *probably* consider something less heavy than `ByteBuffer` for values.

iamaleksey (changes requested):

src/java/org/apache/cassandra/net/CustomParamsSerializer.java
iamaleksey: It looks to me like in the case of jaeger-tracing it's a String to String map, under the hood? Could use that, or byte[].

src/java/org/apache/cassandra/net/CustomParamsSerializer.java
michaelsembwever: too obvious. thanks!

src/java/org/apache/cassandra/net/CustomParamsSerializer.java
michaelsembwever: yup, good catch.

src/java/org/apache/cassandra/net/CustomParamsSerializer.java
michaelsembwever: Zipkin uses byte[]. This isn't meant to be tracing specific either.

michaelsembwever: changed to `byte[]`

src/java/org/apache/cassandra/net/CustomParamsSerializer.java
```diff
@@ -0,0 +1,77 @@
...
+            customParams.put(name, bytes);
+        }
+        return customParams;
+    }
+
+    @Override
+    public long serializedSize(Map<String, byte[]> t, int version)
+    {
+        return TypeSizes.INT_SIZE
+                + t.entrySet().stream().collect(Collectors.summingInt(
```
iamaleksey: FWIW I meant not using streams at all here. No need to create avoidable garbage here. What we tend to do in absolute most of our serializers (open any) is to have `serialize()` and `serializedSize()` look almost the same, visually, except in the latter you just sum up the sizes instead of writing out.

iamaleksey (changes requested):

src/java/org/apache/cassandra/net/CustomParamsSerializer.java
```diff
@@ -0,0 +1,80 @@
...
+    public static final CustomParamsSerializer serializer = new CustomParamsSerializer();
+
+    @Override
+    public void serialize(Map<String, byte[]> t, DataOutputPlus out, int version) throws IOException
+    {
+        out.writeInt(t.size());
+        for (Map.Entry<String, byte[]> e : t.entrySet())
+        {
+            out.writeUTF(e.getKey());
+            out.writeInt(e.getValue().length);
```
iamaleksey: I would use a varint here, as four bytes per value is going to be mostly wasteful in this context.

src/java/org/apache/cassandra/net/Message.java
```diff
@@ -473,6 +482,13 @@ public Builder<T> withParam(ParamType type, Object value)
             return this;
         }
 
+        public Builder<T> withCustomParam(String name, byte[] value)
```
iamaleksey, line 485: May want to provide a method that takes a prebuilt map in its entirety for those users of the API who would prefer that to doing some unnecessary look-ups per param.

src/java/org/apache/cassandra/net/Message.java
```diff
@@ -473,6 +482,13 @@ public Builder<T> withParam(ParamType type, Object value)
...
+        {
+            params.putIfAbsent(ParamType.CUSTOM_MAP, new HashMap<String,byte[]>());
+            ((Map<String,byte[]>) params.get(ParamType.CUSTOM_MAP)).put(name, value);
```
iamaleksey: On that note, you are wasting work here by doing the lookup twice - once for `putIfAbsent()`, then again for `get()`.

Not only that, but you are also allocating an entire `HashMap` every time, even if it already exists in `params` map.

Could either use `computeIfAbsent()` and hold on to the return value, or write this as a manual `get()` + `put()` if null.

iamaleksey (changes requested):

src/java/org/apache/cassandra/net/CustomParamsSerializer.java
iamaleksey: Would throw the helper methods `writeWithVIntLength()`, `readWithVIntLength()`, and `sizeWithVIntLength()` into `ByteArrayUtil`, next to the existing ones for short/int. We have the equivalents for byte buffers in `ByteBufferUtil`, might as well add them for byte arrays.

src/java/org/apache/cassandra/net/Message.java
michaelsembwever: At the moment there's not a clear use case for this method either.

Tracing implementations will override `Tracing. addTraceHeaders(..)` which `Message.buildParams(..)` calls.
e.g. https://github.com/thelastpickle/cassandra-zipkin-tracing/blob/ffb9afb9bc876ac212d51305ac7594843d661d2a/src/main/java/com/thelastpickle/cassandra/tracing/ZipkinTracing.java#L208-L210

src/java/org/apache/cassandra/net/CustomParamsSerializer.java
```diff
@@ -0,0 +1,76 @@
...
+
+
+class CustomParamsSerializer implements IVersionedSerializer<Map<String,byte[]>>
+{
+    public static final CustomParamsSerializer serializer = new CustomParamsSerializer();
+
+    @Override
+    public void serialize(Map<String, byte[]> t, DataOutputPlus out, int version) throws IOException
+    {
+        out.writeInt(t.size());
```
iamaleksey: Can/should also be a vint (unsigned).

src/java/org/apache/cassandra/utils/ByteArrayUtil.java
```diff
@@ -230,6 +232,17 @@ public static void writeWithShortLength(byte[] buffer, DataOutput out) throws IO
         out.write(buffer);
     }
 
+    public static void writeWithVIntLength(byte[] bytes, DataOutputPlus out) throws IOException
+    {
+        out.writeVInt(bytes.length);
```
iamaleksey: `writeUnsignedVInt()`

src/java/org/apache/cassandra/utils/ByteArrayUtil.java
```diff
@@ -230,6 +232,17 @@ public static void writeWithShortLength(byte[] buffer, DataOutput out) throws IO
...
+        out.write(bytes);
+    }
+
+    public static int serializedSizeWithVIntLength(byte[] bytes)
+    {
+        return TypeSizes.sizeofVInt(bytes.length) + bytes.length;
```
iamaleksey: `sizeofUnsignedVInt()`

iamaleksey (changes requested):

iamaleksey (approved):

PR #2010, opened 2022-11-14
Avoid schema mismatch problems on memtable API misconfiguration, 4.1 branch

blambov:
Falls back to default on schema change carrying invalid memtable configuration.

### DIFF
--- a/src/java/org/apache/cassandra/schema/MemtableParams.java
+++ b/src/java/org/apache/cassandra/schema/MemtableParams.java
@@ -28,6 +28,8 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableMap;
 
+import org.slf4j.LoggerFactory;
+
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.config.InheritingClass;
 import org.apache.cassandra.config.ParameterizedClass;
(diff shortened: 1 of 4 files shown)


test/distributed/org/apache/cassandra/distributed/test/AlterTest.java
```diff
@@ -107,4 +120,73 @@ public void alteringKeyspaceOnInsufficientNumberOfReplicasFiltersOutGossppingOnl
...
+                cluster.schemaChange("ALTER TABLE " + KEYSPACE + ".tbl WITH memtable = 'testconfig'", false, node2);
+                fail("Expected ALTER to fail with unknown memtable configuration.");
+            }
+            catch (Throwable t)
+            {
+                // expected
+                logger.info("Expected: {}", t.getMessage());
+                Assert.assertTrue(Throwables.isCausedBy(t, x -> x.getMessage().matches("Memtable configuration.*not found.*")));
+            }
+            long mark = node2.logs().mark();
```
maedhroz: nit: `mark` unused?

maedhroz (approved):

src/java/org/apache/cassandra/schema/MemtableParams.java
```diff
@@ -110,6 +112,22 @@ public static MemtableParams get(String key)
...
 
+    public static MemtableParams getWithFallback(String key)
+    {
+        try
+        {
+            return get(key);
+        }
+        catch (ConfigurationException e)
+        {
+            LoggerFactory.getLogger(MemtableParams.class).error("Invalid memtable configuration \"" + key + "\" in schema. " +
```
maedhroz: nit: Any reason we create the logger inline? (Looks like it would be the only usage, but just curious...)

test/distributed/org/apache/cassandra/distributed/test/AlterTest.java
blambov: Fixed now, `mark` is used is finding the relevant error after this point in the logs below.

src/java/org/apache/cassandra/schema/MemtableParams.java
blambov: It's never used in other places, and this is a rare event which to me did not warrant always creating the logger object on its own.

PR #2020, opened 2022-11-16
CASSANDRA-18044

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
+++ b/src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
@@ -313,9 +313,24 @@ public enum CassandraRelevantProperties
 
     /** Used when running in Client mode and the system and schema keyspaces need to be initialized outside of their normal initialization path **/
     FORCE_LOAD_LOCAL_KEYSPACES("cassandra.schema.force_load_local_keyspaces"),
+
+    // commit log relevant properties
+    /**
+     * Entities to replay mutations for upon commit log replay, property is meant to contain
+     * comma-separated entities which are either names of keyspaces or keyspaces and tables or their mix.
+     * Examples:
+     * just keyspaces
+     * -Dcassandra.replayList=ks1,ks2,ks3
+     * specific tables
+     * -Dcassandra.replayList=ks1.tb1,ks2.tb2
+     * mix of tables and keyspaces
+     * -Dcassandra.replayList=ks1.tb1,ks2
+     *
+     * If only keyspaces are specified, mutations for all tables in such keyspace will be replayed
+     * */
+    COMMIT_LOG_REPLAY_LIST("cassandra.replayList", null)
+
     ;
-
-
 
     CassandraRelevantProperties(String key, String defaultVal)
     {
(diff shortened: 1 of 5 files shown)


src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -313,9 +313,24 @@
...
+     * just keyspaces
+     * -Dcassandra.replayList=ks1,ks2,ks3
+     * specific tables
+     * -Dcassandra.replayList=ks1.tb1,ks2.tb2
+     * mix of tables and keyspaces
+     * -Dcassandra.replayList=ks1.tb1,ks2
+     *
+     * If only keyspaces are specified, mutations for all tables in such keyspace will be replayed
+     * */
+    COMMIT_LOG_REPLAY_LIST("cassandra.replayList", null)
```
smiklosovic, line 331: I moved this one here from CommitLogReplayer when I was on it.

src/java/org/apache/cassandra/db/commitlog/CommitLogReplayer.java
```diff
@@ -376,25 +377,45 @@ public static CommitLogPosition firstNotCovered(Collection<IntervalSet<CommitLog
...
             // If no replaylist is supplied an empty array of strings is used to replay everything.
-            if (System.getProperty("cassandra.replayList") == null)
+
+            String replayList = CassandraRelevantProperties.COMMIT_LOG_REPLAY_LIST.getString();
+
+            if (replayList == null)
                 return new AlwaysReplayFilter();
 
             Multimap<String, String> toReplay = HashMultimap.create();
-            for (String rawPair : System.getProperty("cassandra.replayList").split(","))
```
smiklosovic, line 383: moved to CassandraRelevantProperties as it should be

src/java/org/apache/cassandra/db/commitlog/CommitLogReplayer.java
```diff
@@ -376,25 +377,45 @@ public static CommitLogPosition firstNotCovered(Collection<IntervalSet<CommitLog
...
+                    ColumnFamilyStore cfs = ks.getColumnFamilyStore(pair[1]);
+                    if (cfs == null)
+                        throw new IllegalArgumentException(String.format("Unknown table %s.%s", keyspaceName, pair[1]));
 
-                toReplay.put(pair[0], pair[1]);
+                    toReplay.put(keyspaceName, pair[1]);
+                }
             }
+
+            logger.info("Tables going to be replayed: {}", toReplay.asMap().toString());
```
blambov: The message does not sound right, maybe `Tables to be replayed`?

blambov (approved):

src/java/org/apache/cassandra/db/commitlog/CommitLogReplayer.java
smiklosovic: @blambov I think we should also cover the case when `cassandra.replayList` is not set which means that `toReplay` would be empty map so this would look like `Tables to be replayed: <<empty map string>>` which is a little bit misleading.

I suggest to do it like:

````
if (toReplay.isEmpty())
    logger.info("All mutations in commit logs will be replayed.");
else
    logger.info("Tables to be replayed: {}", toReplay.asMap().toString());
````

EDIT:

even better version:

````
if (toReplay.isEmpty())
    logger.info("All tables will be included in commit log replay.");
else
    logger.info("Tables to be replayed: {}", toReplay.asMap().toString());
````

The problem with first iteration was that if we say "all mutations in commit logs will be replayed", it is again not entirely true as we might do point-in-time restoration and in that case not _all_ mutations are in fact replayed. But all tables are included in such replaying. It just does not necessarily replay all mutations for them.

WDYT?

blambov (approved):

PR #2025, opened 2022-11-21
CASSANDRA-18042 - WIP DO NOT COMMIT

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/db/guardrails/EnableFlag.java
+++ b/src/java/org/apache/cassandra/db/guardrails/EnableFlag.java
@@ -32,6 +32,7 @@ import org.apache.cassandra.service.ClientState;
  */
 public class EnableFlag extends Guardrail
 {
+    private final Predicate<ClientState> warned;
     private final Predicate<ClientState> enabled;
     private final String featureName;
 
@@ -47,7 +48,31 @@ public class EnableFlag extends Guardrail
      */
     public EnableFlag(String name, @Nullable String reason, Predicate<ClientState> enabled, String featureName)
     {
+        this(name, reason, (state) -> false, enabled, featureName);
+    }
+
+    /**
+     * Creates a new {@link EnableFlag} guardrail.
+     *
+     * @param name        the identifying name of the guardrail
+     * @param reason      the optional description of the reason for guarding the operation
+     * @param warned      a {@link ClientState}-based supplier of boolean indicating whether warning should be
+     *                    emitted even guardrail as such has passed. If guardrail fails, the warning will not be
+     *                    emitted. This might be used for cases when we want to warn a user regardless of successful
+     *                    guardrail execution.
+     * @param enabled     a {@link ClientState}-based supplier of boolean indicating whether the feature guarded by this
+     *                    guardrail is enabled.
+     * @param featureName The feature that is guarded by this guardrail (for reporting in error messages), {@link
+     *                    EnableFlag#ensureEnabled(String, ClientState)} can specify a different {@code featureName}.
+     */
+    public EnableFlag(String name,
+                      @Nullable String reason,
+                      Predicate<ClientState> warned,
+                      Predicate<ClientState> enabled,
+                      String featureName)
+    {
         super(name, reason);
+        this.warned = warned;
         this.enabled = enabled;
         this.featureName = featureName;
     }
@@ -93,7 +118,16 @@ public class EnableFlag extends Guardrail
      */
     public void ensureEnabled(String featureName, @Nullable ClientState state)
     {
-        if (!isEnabled(state))
+        if (!enabled(state))
+            return;
+
+        if (!enabled.test(state))
+        {
             fail(featureName + " is not allowed", state);
+            return;
+        }
+
+        if (warned.test(state))
+            warn(featureName + " is not recommended");
     }
 }
--- a/src/java/org/apache/cassandra/db/guardrails/Guardrails.java
+++ b/src/java/org/apache/cassandra/db/guardrails/Guardrails.java
@@ -33,11 +33,13 @@ import org.apache.cassandra.config.DataStorageSpec;
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.config.GuardrailsOptions;
 import org.apache.cassandra.db.ConsistencyLevel;
+import org.apache.cassandra.db.compaction.TimeWindowCompactionStrategy;
 import org.apache.cassandra.locator.InetAddressAndPort;
 import org.apache.cassandra.service.disk.usage.DiskUsageBroadcaster;
 import org.apache.cassandra.utils.MBeanWrapper;
 
 import static java.lang.String.format;
+import static org.apache.cassandra.schema.TableParams.Option;
 
 /**
  * Entry point for Guardrails, storing the defined guardrails and providing a few global methods over them.
@@ -198,6 +200,20 @@ public final class Guardrails implements GuardrailsMBean
                    null,
                    state -> CONFIG_PROVIDER.getOrCreate(state).getCompactTablesEnabled(),
                    "Creation of new COMPACT STORAGE tables");
+
+    /**
+     * Guardrail disabling the creation of a new table when {@link Option#DEFAULT_TIME_TO_LIVE} is set to 0
+     * and compaction strategy is {@link TimeWindowCompactionStrategy} or its subclass. If this guardrail does not fail
+     * and such configuration combination is found, it will always warn.
+     */
+    public static final EnableFlag zeroDefaultTTLOnTWCSEnabled =
+    new EnableFlag("zero_default_ttl_on_twcs",
+                   "It is suspicious to use default_time_to_live set to 0 with such compaction strategy. " +
+                   "Please keep in mind that data will not start to automatically expire after they are older " +
+                   "than a respective compaction window unit of a certain size. ",
+                   state -> true,
+                   state -> CONFIG_PROVIDER.getOrCreate(state).getZeroDefaultTtlOnTimeWindowCompactionStrategyEnabled(),
+                   "0 default_time_to_live on a table with " + TimeWindowCompactionStrategy.class.getSimpleName() + " compaction strategy");
 
     /**
      * Guardrail on the number of elements returned within page.
@@ -1011,6 +1027,18 @@ public final class Guardrails implements GuardrailsMBean
     public void setMinimumReplicationFactorThreshold(int warn, int fail)
     {
         DEFAULT_CONFIG.setMinimumReplicationFactorThreshold(warn, fail);
+    }
+
+    @Override
+    public boolean getZeroDefaultTtlOnTimeWindowCompactionStrategyEnabled()
+    {
+        return DEFAULT_CONFIG.getZeroDefaultTtlOnTimeWindowCompactionStrategyEnabled();
+    }
+
+    @Override
+    public void setZeroDefaultTtlOnTimeWindowCompactionStrategyEnabled(boolean enabled)
+    {
+        DEFAULT_CONFIG.setZeroDefaultTtlOnTimeWindowCompactionStrategyEnabled(enabled);
     }
 
     private static String toCSV(Set<String> values)
--- a/test/unit/org/apache/cassandra/db/guardrails/GuardrailZeroDefaultTTLOnTWCSTest.java
+++ b/test/unit/org/apache/cassandra/db/guardrails/GuardrailZeroDefaultTTLOnTWCSTest.java
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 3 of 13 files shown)


src/java/org/apache/cassandra/db/guardrails/EnableFlag.java
```diff
@@ -95,5 +120,8 @@ public void ensureEnabled(String featureName, @Nullable ClientState state)
     {
         if (!isEnabled(state))
             fail(featureName + " is not allowed", state);
```
adelapena, line 126: Note that `GuardrailFail` can emit warnings without throwing an exception. That happens when the `ClientState` is unknown, and it's used for background processes. So we need to either add a `return` after the call to `fail`, or put the warning part in an `else` branch. For example:
```java
if (!isEnabled(state))
{
    fail(featureName + " is not allowed", state);
    return;
}
```

test/unit/org/apache/cassandra/db/guardrails/GuardrailsTest.java
```diff
@@ -452,6 +453,22 @@ public void testPredicatesUsers() throws Throwable
         assertValid(() -> guard.guard(101, superClientState));
     }
 
+    @Test
+    public void testEnableFlagWarn() throws Throwable
```
adelapena: I'd put this test method immediately below `testEnableFlag` and `testEnableFlagUsers`, to keep related things together.

adelapena: Since the new warn was producing duplicated warnings for `null` client states, I'd consider that case in this test:
```java
@Test
public void testEnableFlagWarn() throws Throwable
{
    EnableFlag disabledGuard = new EnableFlag("x", null, state -> false, state -> true, FEATURE);
    assertFails(() -> disabledGuard.ensureEnabled(null), false, FEATURE + " is not allowed");
    assertFails(() -> disabledGuard.ensureEnabled(userClientState), FEATURE + " is not allowed");
    assertValid(() -> disabledGuard.ensureEnabled(systemClientState));
    assertValid(() -> disabledGuard.ensureEnabled(superClientState));
    
    EnableFlag enabledGuard = new EnableFlag("x", null, state -> true, state -> true, FEATURE);
    assertWarns(() -> enabledGuard.ensureEnabled(null), FEATURE);
    assertWarns(() -> enabledGuard.ensureEnabled(userClientState), FEATURE);
    assertValid(() -> disabledGuard.ensureEnabled(systemClientState));
    assertValid(() -> disabledGuard.ensureEnabled(superClientState));
}
```

src/java/org/apache/cassandra/db/guardrails/EnableFlag.java
```diff
@@ -33,6 +33,7 @@
 public class EnableFlag extends Guardrail
 {
     private final Predicate<ClientState> enabled;
+    private final Predicate<ClientState> warned;
```
adelapena: All the other guardrails mention first the soft limit and then the hard limit. We could follow the same criteria here and reverse the order of the `enabled` and `warned` attributes. Same for the constructor arguments.

src/java/org/apache/cassandra/db/guardrails/EnableFlag.java
adelapena: Also, we can avoid the duplicated checks of the `ClientState` by just checking it once at the beginning of the method:
```java
public void ensureEnabled(String featureName, @Nullable ClientState state)
{
    if (!enabled(state))
        return;

    if (!enabled.test(state))
    {
        fail(featureName + " is not allowed", state);
        return;
    }

    if (warned.test(state))
        warn(featureName);
}
```

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -199,6 +200,19 @@ public final class Guardrails implements GuardrailsMBean
                    state -> CONFIG_PROVIDER.getOrCreate(state).getCompactTablesEnabled(),
                    "Creation of new COMPACT STORAGE tables");
 
+    /**
+     * Guardrail disabling the creation of a new table when default_time_to_live is set to 0 and
+     * compaction strategy is TimeWindowCompactionStrategy or its subclass. If this guardrail does not fail
+     * and such configuration combination is found, it will always warn.
+     */
```
adelapena: We could use JavaDoc links:
```suggestion
    /**
     * Guardrail disabling the creation of a new table when {@link TableParams.Option#DEFAULT_TIME_TO_LIVE} is set to 0
     * and compaction strategy is {@link TimeWindowCompactionStrategy} or its subclass. If this guardrail does not fail
     * and such configuration combination is found, it will always warn.
     */
```

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -199,6 +200,19 @@ public final class Guardrails implements GuardrailsMBean
...
+    public static final EnableFlag zeroDefaultTTLonTimeWindowCompactionStrategyEnabled =
+    new EnableFlag("zero_default_ttl_on_time_window_compaction_strategy_enabled",
+                   "It is suspicious to use default_time_to_live set to 0 with such compaction strategy. Please keep in mind " +
+                   "that data will not start to automatically expire after they are older than a respective compaction window unit of a certain size. ",
```
adelapena: Line width:
```suggestion
new EnableFlag("zero_default_ttl_on_time_window_compaction_strategy_enabled",
                   "It is suspicious to use default_time_to_live set to 0 with such compaction strategy. Please keep " +
                   "in mind that data will not start to automatically expire after they are older than a respective " +
                   "compaction window unit of a certain size. ",
```

test/unit/org/apache/cassandra/db/guardrails/GuardrailZeroDefaultTtlOnTimeWindowCompactionsStrategyTest.java
```diff
@@ -0,0 +1,67 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.db.guardrails;
+
+import org.junit.Test;
+
+public class GuardrailZeroDefaultTtlOnTimeWindowCompactionsStrategyTest extends GuardrailTester
```
adelapena: Most instances of `GuardrailTester` have a `testExcludedUsers` test method to verify behaviour for excluded users. We could do the same here:
```java
@Test
public void testExcludedUsers() throws Throwable
{
    for (boolean enabled : new boolean[]{ false, true })
    {
        setGuardrail(enabled);
        testExcludedUsers(() -> QUERY,
                          () -> VALID_QUERY_1,
                          () -> VALID_QUERY_2);
    }
}
```

src/java/org/apache/cassandra/db/guardrails/EnableFlag.java
smiklosovic: ok, I was looking at that alphabetically.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -199,6 +201,20 @@ public final class Guardrails implements GuardrailsMBean
                    state -> CONFIG_PROVIDER.getOrCreate(state).getCompactTablesEnabled(),
                    "Creation of new COMPACT STORAGE tables");
 
+    /**
+     * Guardrail disabling the creation of a new table when {@link Option#DEFAULT_TIME_TO_LIVE} is set to 0
+     * and compaction strategy is {@link TimeWindowCompactionStrategy} or its subclass. If this guardrail does not fail
+     * and such configuration combination is found, it will always warn.
+     */
+    public static final EnableFlag zeroDefaultTTLOnTWCSEnabled =
+    new EnableFlag("zero_default_ttl_on_time_window_compaction_strategy_enabled",
```
adelapena: This should be the name of the guardrail, not the name of the property.  So either`zero_default_ttl_on_time_window_compaction_strategy` or `zero_default_ttl_on_twcs`

test/unit/org/apache/cassandra/db/guardrails/GuardrailZeroDefaultTTLOnTWCSTest.java
```diff
@@ -0,0 +1,79 @@
...
+    public GuardrailZeroDefaultTTLOnTWCSTest()
+    {
+        super(Guardrails.zeroDefaultTTLOnTWCSEnabled);
+    }
+
+    @Test
+    public void testGuardrailEnabled() throws Throwable
+    {
+        setGuardrail(true);
+        assertWarns(QUERY, guardrail.reason);
```
adelapena: The second argument should be the warn message, `0 default_time_to_live on a table with TimeWindowCompactionStrategy compaction strategy`. The reason is already automatically checked by the test method.

test/unit/org/apache/cassandra/db/guardrails/GuardrailZeroDefaultTTLOnTWCSTest.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @Test
+    public void testGuardrailDisabled() throws Throwable
+    {
+        setGuardrail(false);
+        assertFails(QUERY, guardrail.reason);
```
adelapena: The second argument should be the warn message, `0 default_time_to_live on a table with TimeWindowCompactionStrategy compaction strategy is not allowed`. The reason is already automatically checked by the test method.

src/java/org/apache/cassandra/db/guardrails/EnableFlag.java
```diff
@@ -93,7 +118,16 @@ public void ensureEnabled(@Nullable ClientState state)
...
+            return;
+
+        if (!enabled.test(state))
+        {
             fail(featureName + " is not allowed", state);
+            return;
+        }
+
+        if (warned.test(state))
+            warn(featureName);
```
adelapena: Why not add something similar to what we have on the call to `fail`? For example:
```suggestion
            warn(featureName + " is not recommended");
```

test/unit/org/apache/cassandra/db/guardrails/GuardrailZeroDefaultTTLOnTWCSTest.java
smiklosovic: Interestingly enough, this string is `featureName` in EnableFlag if you check `ensureEnabled` more closely. So it would be nice if we can use that class property rather than copying it here. However, `guardrail` here is of type `Guardrail`, not `EnableFlag` so I would need to cast it. Do you think that is ok?

test/unit/org/apache/cassandra/db/guardrails/GuardrailZeroDefaultTTLOnTWCSTest.java
```diff
@@ -0,0 +1,81 @@
...
+    public GuardrailZeroDefaultTTLOnTWCSTest()
+    {
+        super(Guardrails.zeroDefaultTTLOnTWCSEnabled);
+    }
+
+    @Test
+    public void testGuardrailEnabled() throws Throwable
+    {
+        setGuardrail(true);
+        assertWarns(QUERY, "0 default_time_to_live on a table with " + TimeWindowCompactionStrategy.class.getSimpleName() + " compaction strategy");
```
smiklosovic: As commented previously, this string is in fact `featureName` but I have left it here as a simple string. Extracting `featureName` from`EnableFlag` is quite tricky / not so straightforward. This whole testing framework works with `Guardrail` so I would need to cast it etc etc. Maybe some getter EnableFlag for featureName as it is private etc ...

test/unit/org/apache/cassandra/db/guardrails/GuardrailZeroDefaultTTLOnTWCSTest.java
```diff
@@ -0,0 +1,83 @@
...
+    {
+        super(Guardrails.zeroDefaultTTLOnTWCSEnabled);
+    }
+
+    @Test
+    public void testGuardrailEnabled() throws Throwable
+    {
+        setGuardrail(true);
+        assertWarns(QUERY, "0 default_time_to_live on a table with " +
+                           TimeWindowCompactionStrategy.class.getSimpleName() + " compaction strategy");
```
adelapena:
```suggestion
                           TimeWindowCompactionStrategy.class.getSimpleName() + " compaction strategy is not recommended");
```

test/unit/org/apache/cassandra/db/guardrails/GuardrailZeroDefaultTTLOnTWCSTest.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Test
+    public void testGuardrailDisabled() throws Throwable
+    {
+        setGuardrail(false);
+        assertFails(QUERY, "0 default_time_to_live on a table with " +
+                           TimeWindowCompactionStrategy.class.getSimpleName() + " compaction strategy");
```
adelapena:
```suggestion
                           TimeWindowCompactionStrategy.class.getSimpleName() + " compaction strategy is not allowed");
```

PR #2026, opened 2022-11-21
(Accord) Invalidation fixes

belliottsmith:


src/java/org/apache/cassandra/service/accord/AccordCommand.java
```diff
@@ -802,10 +803,21 @@ public void removeWaitingOn(TxnId txnId, Timestamp executeAt)
     }
 
     @Override
-    public TxnId firstWaitingOnApply()
+    public boolean isWaitingOnDependency()
```
aweisberg, line 815: These two changes are cut and pasting from Accord. Do they need to be duplicated? They reference different maps, so maybe that's not reasonable, but the less duplication the better.

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -325,6 +325,12 @@ public Timestamp preaccept(TxnId txnId, Keys keys)
         return time.uniqueNow(max);
     }
 
+    @Override
+    public Timestamp uniqueNow(TxnId greaterThan)
+    {
+        return time.uniqueNow(greaterThan);
```
aweisberg: If node time service already factors this out why does it need to be in both the command store implementations?

aweisberg (changes requested):
Don't have much to add beyond there seems to be some things that could be in the abstract base CommandStore?

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
belliottsmith: Fair. I'll expose the time service instead.

src/java/org/apache/cassandra/service/accord/AccordCommand.java
belliottsmith: I'm not settled on the final API for this, but I'm also not sure we should force implementations to use separate maps - that's a detail that just happens to be shared between implementations today. It might be that a single array of TxnId is used with the status being stored in an adjacent array. I'd prefer to have value types and be able to ask for the earliest dependency of either kind with both its Timestamp and TxnId. It might be, as we improve efficiency C* side, we can improve this API further.

Which is all to say I don't mind all that much either way if you feel strongly about this at present.

aweisberg (approved):
+1 TY

PR #2031, opened 2022-11-28
CASSANDRA-18077: Add SpotBugs to the build

dcapwell:


michaelsembwever:
Can we separate commits between introducing the linter, to code fixes for the linter?

Can we bring over checkstyle and other linter stuff from build.xml to .build/build-linter.xml ? 

Unrelated, what's your thoughts about the .build/cassandra-build-deps-template.xml filenames? (Is template confusing and should they be called something like `.build/cassandra-build-deps-unfiltered.xml` instead?)

otherwise LGTM

dcapwell:
> Can we separate commits between introducing the linter, to code fixes for the linter?

What is the intent?  I don't plan do merge with multiple commits, so is this to help review?

> Can we bring over checkstyle and other linter stuff from build.xml to .build/build-linter.xml ?

I could, I just felt this patch was large already...

> Unrelated, what's your thoughts about the .build/[cassandra-build-deps-template](https://issues.apache.org/jira/browse/CASSANDRA-build-deps-template).xml filenames? (Is template confusing and should they be called something like .build/cassandra-build-deps-unfiltered.xml instead?)

I find `template` is better than `unfiltered` as `template` is more what this is... a file we modify using special keywords to generate our desired output...

dcapwell:
spoke on slack about multi-commit; 

> trivial, can be ignored. i suspect folk may desire that separation even in the merge

My stance is that I am not a fan, but I may be bias as most of the time I see people do this in this project, the commits have issues when not containing everything for the JIRA... I "could" make sure to avoid this, but I don't see the benefit

.build/parent-pom-template.xml
```diff
@@ -497,6 +497,12 @@
         <version>8.40</version>
         <scope>test</scope>
       </dependency>
+			<dependency>
```
smiklosovic, line 500: this seems to be not aligned with other dependencies, too much tabs?

dcapwell: so many tabs!

maoling:
So great work, C* really needs this static analysis tool. @dcapwell  are you still here?
If you're not free, I can pick up it and co-author with you

PR #2038, opened 2022-12-02
CASSANDRA-18086: Fix wait time returned to be in nanoseconds

marianne-manaog:
A bug fix for 'wait' time to be in nanoseconds for consistency instead of microseconds.

patch by @marianne-manaog, although fix suggested by @belliottsmith;

The [Cassandra Jira](https://issues.apache.org/jira/browse/CASSANDRA-18086)

### DIFF
--- a/src/java/org/apache/cassandra/service/paxos/ContentionStrategy.java
+++ b/src/java/org/apache/cassandra/service/paxos/ContentionStrategy.java
@@ -372,7 +372,7 @@ public class ContentionStrategy
         }
     }
 
-    private long computeWaitUntilForContention(int attempts, TableMetadata table, DecoratedKey partitionKey, ConsistencyLevel consistency, Type type)
+    long computeWaitUntilForContention(int attempts, TableMetadata table, DecoratedKey partitionKey, ConsistencyLevel consistency, Type type)
     {
         if (attempts >= traceAfterAttempts && !Tracing.isTracing())
         {
@@ -411,7 +411,7 @@ public class ContentionStrategy
         return nanoTime() + MICROSECONDS.toNanos(wait);
     }
 
-    private boolean doWaitForContention(long deadline, int attempts, TableMetadata table, DecoratedKey partitionKey, ConsistencyLevel consistency, Type type)
+    boolean doWaitForContention(long deadline, int attempts, TableMetadata table, DecoratedKey partitionKey, ConsistencyLevel consistency, Type type)
     {
         long until = computeWaitUntilForContention(attempts, table, partitionKey, consistency, type);
         if (until >= deadline)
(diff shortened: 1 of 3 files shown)


aweisberg (changes requested):
Can you add a unit test to `ContentionStrategyTest` that checks that `computeWaitUntilForContention` is producing values in the range we want?

A better fix is probably to only use microseconds in configuration where it makes expressing the configuration easier and convert to nanoseconds immediately upon reading the configuration. @belliottsmith  WDYT?

At a higher level a [`Duration`](https://github.com/airlift/units/blob/c25cb5704978c48f8ce966d0c06447e351cab609/src/main/java/io/airlift/units/Duration.java) would be safer. It works as both a return value and a parameter unlike `TimeUnit` which doesn't pair well with a value.

Or we can keep going with somewhat ambiguous types and just be careful about naming and converting things to nanoseconds as soon as possible.

belliottsmith:
Yes, we could definitely convert to nanos earlier and always use nanos. Perhaps we can do that in a follow-up patch.

belliottsmith:
Marianne, I think you intended to merge the 18097 PR in here, but it seems to have merged "cleanly" meaning none of the commits were adopted. Not sure what happened there.

marianne-manaog:
Thanks for your comment, @belliottsmith. I have updated this branch and PR to reflect all related changes required for review.

> Marianne, I think you intended to merge the 18097 PR in here, but it seems to have merged "cleanly" meaning none of the commits were adopted. Not sure what happened there.

belliottsmith (approved):
Thanks!

marianne-manaog:
Thanks for your comment, @aweisberg. @mfleming kindly added the required tests in [this commit](8ec9d77de6492189fdce92d4b09e9104c6626fae).

> Can you add a unit test to `ContentionStrategyTest` that checks that `computeWaitUntilForContention` is producing values in the range we want?
> 
> A better fix is probably to only use microseconds in configuration where it makes expressing the configuration easier and convert to nanoseconds immediately upon reading the configuration. @belliottsmith WDYT?
> 
> At a higher level a [`Duration`](https://github.com/airlift/units/blob/c25cb5704978c48f8ce966d0c06447e351cab609/src/main/java/io/airlift/units/Duration.java) would be safer. It works as both a return value and a parameter unlike `TimeUnit` which doesn't pair well with a value.
> 
> Or we can keep going with somewhat ambiguous types and just be careful about naming and converting things to nanoseconds as soon as possible.

aweisberg:
Committed as [f0ad7eadbeb3208e08a9339881931222fdab253b](https://github.com/apache/cassandra/commit/f0ad7eadbeb3208e08a9339881931222fdab253b)

PR #2043, opened 2022-12-06
Use checked casts when reading vints as ints

aweisberg:
patch by Ariel Weisberg; reviewed by ? and ? for CASSANDRA-18099


src/java/org/apache/cassandra/cql3/functions/types/TypeCodec.java
```diff
@@ -3071,8 +3072,8 @@ public Duration deserialize(ByteBuffer bytes, ProtocolVersion protocolVersion)
                 DataInput in = ByteStreams.newDataInput(Bytes.getArray(bytes));
                 try
                 {
-                    int months = (int) VIntCoding.readVInt(in);
-                    int days = (int) VIntCoding.readVInt(in);
+                    int months = checkedCast(VIntCoding.readVInt(in));
+                    int days = checkedCast(VIntCoding.readVInt(in));
```
dcapwell: should use the new `checked` methods rather than doing `checkedCast` here

dcapwell (approved):
+1 assuming we all agree on the name =). It sounds like `readUnsignedVInt32`, but I am cool with other names

aweisberg:
So I think one thing to be aware of is that we aren't reading an unsigned int32. There is maybe zero code that handles unsigned int32s? We are casting to a signed int and the checked cast is for a signed ints range of values.

belliottsmith:
For consistency with the existing API though the valid thing to do would be to simply check its populating at most 32 bits, and if the 32nd bit is set it’s a negative int that may (or may not) be interpreted by the caller as an unsigned value.

That said, we could instead introduce eg readNaturalInt32 which would expect at most 31bits to be set, which might be better.

aweisberg:
OK, this looks like it is passing tests now https://app.circleci.com/pipelines/github/aweisberg/cassandra?branch=cassandra-18099-trunk

src/java/org/apache/cassandra/utils/vint/VIntCoding.java
```diff
@@ -73,6 +73,21 @@ public byte[] initialValue()
 
     public static final int MAX_SIZE = 10;
 
+    /**
+     * Throw when attempting to decode a vint and the output type
+     * doesn't have enough space to fit the value that was decoded
+     */
+    public static class VIntOutOfRangeException extends RuntimeException
+    {
+        private final long value;
```
maedhroz: nit: `value` isn't used?

src/java/org/apache/cassandra/cql3/functions/types/TypeCodec.java
```diff
@@ -17,6 +17,13 @@
  */
 package org.apache.cassandra.cql3.functions.types;
 
+import com.google.common.io.ByteStreams;
+import com.google.common.reflect.TypeToken;
+import org.apache.cassandra.cql3.functions.types.exceptions.InvalidTypeException;
+import org.apache.cassandra.cql3.functions.types.utils.Bytes;
+import org.apache.cassandra.transport.ProtocolVersion;
+import org.apache.cassandra.utils.vint.VIntCoding;
```
maedhroz: nit: Do these imports need to move down?

src/java/org/apache/cassandra/db/columniterator/AbstractSSTableIterator.java
```diff
@@ -17,23 +17,26 @@
...
 import org.apache.cassandra.io.util.DataPosition;
+import org.apache.cassandra.io.util.FileDataInput;
 import org.apache.cassandra.io.util.FileHandle;
+import org.apache.cassandra.schema.TableMetadata;
 import org.apache.cassandra.utils.ByteBufferUtil;
 
+import java.io.IOException;
+import java.util.Comparator;
+import java.util.Iterator;
+import java.util.NoSuchElementException;
```
maedhroz: nit: Put these back in order?

src/java/org/apache/cassandra/db/partitions/PartitionUpdate.java
```diff
@@ -17,33 +17,35 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.EOFException;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
```
maedhroz: nit: Put these back in order?

src/java/org/apache/cassandra/db/rows/EncodingStats.java
```diff
@@ -17,17 +17,17 @@
...
 import org.apache.cassandra.cache.IMeasurableMemory;
 import org.apache.cassandra.db.*;
 import org.apache.cassandra.db.partitions.PartitionStatisticsCollector;
 import org.apache.cassandra.io.util.DataInputPlus;
 import org.apache.cassandra.io.util.DataOutputPlus;
 import org.apache.cassandra.utils.ObjectSizes;
 
+import java.io.IOException;
+import java.util.*;
+import java.util.function.Function;
```
maedhroz: nit: Put these back in order

src/java/org/apache/cassandra/db/rows/UnfilteredSerializer.java
```diff
@@ -17,20 +17,20 @@
...
 import org.apache.cassandra.db.rows.Row.Deletion;
 import org.apache.cassandra.io.util.DataInputPlus;
 import org.apache.cassandra.io.util.DataOutputBuffer;
 import org.apache.cassandra.io.util.DataOutputPlus;
 import org.apache.cassandra.io.util.FileDataInput;
+import org.apache.cassandra.schema.ColumnMetadata;
 import org.apache.cassandra.utils.SearchIterator;
 import org.apache.cassandra.utils.WrappedException;
 
+import java.io.IOException;
```
maedhroz: nit: import order

src/java/org/apache/cassandra/db/streaming/ComponentManifest.java
```diff
@@ -27,15 +27,11 @@
 import org.apache.cassandra.io.sstable.Descriptor;
 import org.apache.cassandra.io.util.DataInputPlus;
 import org.apache.cassandra.io.util.DataOutputPlus;
+import org.apache.cassandra.io.util.File;
 
 import java.io.IOException;
 import java.util.*;
```
maedhroz: nit: heh, these were out of order before your patch, but might as well fix that here :)

src/java/org/apache/cassandra/db/Columns.java
```diff
@@ -17,31 +17,30 @@
...
 import org.apache.cassandra.utils.btree.BTree;
-import org.apache.cassandra.utils.btree.BTreeSearchIterator;
 import org.apache.cassandra.utils.btree.BTreeRemoval;
+import org.apache.cassandra.utils.btree.BTreeSearchIterator;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.*;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
```
maedhroz: nit: import order

src/java/org/apache/cassandra/db/RowIndexEntry.java
```diff
@@ -45,6 +35,10 @@
 import org.apache.cassandra.utils.vint.VIntCoding;
 import org.github.jamm.Unmetered;
 
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.List;
```
maedhroz: nit: import order

src/java/org/apache/cassandra/utils/vint/VIntCoding.java
aweisberg: Ah, I meant for that to be public so the caller could inspect the out of range value.

src/java/org/apache/cassandra/utils/vint/VIntCoding.java
```diff
@@ -165,6 +190,27 @@ public static long readVInt(DataInput input) throws IOException
...
+    public static int readVInt32(DataInput input) throws IOException
+    {
+        return checkedCast(decodeZigZag64(readUnsignedVInt(input)));
+    }
+
+    /**
+     * Read up to a 32-bit unsigned integer. This uses a more efficient encoding
+     * for numbers that are mostly positive (avoids zigzag).
+     *
+     * @throws IllegalArgumentException Thrown if the vint doesn't fit a 32-bit integer
```
maedhroz: Should this be `VIntOutOfRangeException` now instead of `IllegalArgumentException`? (Happens in a few places, but mentioning it just here...)

maedhroz: super-nit: `fit a` -> `fit into a`?

src/java/org/apache/cassandra/db/SerializationHeader.java
```diff
@@ -40,6 +35,10 @@
 import org.apache.cassandra.schema.TableMetadata;
 import org.apache.cassandra.utils.ByteBufferUtil;
 
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.*;
```
maedhroz: nit: import order

src/java/org/apache/cassandra/cql3/functions/types/TypeCodec.java
aweisberg: No, it looks like the style for IntelliJ from here https://cassandra.apache.org/_/development/code_style.html#imports doesn't specify the correct import order.

src/java/org/apache/cassandra/hints/Hint.java
```diff
@@ -38,6 +31,11 @@
 import org.apache.cassandra.utils.vint.VIntCoding;
 import org.assertj.core.util.VisibleForTesting;
 
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.concurrent.TimeUnit;
```
maedhroz: nit: import order

src/java/org/apache/cassandra/hints/HintMessage.java
```diff
@@ -35,6 +28,12 @@
 import org.apache.cassandra.schema.TableId;
 import org.apache.cassandra.utils.UUIDSerializer;
 
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Objects;
+import java.util.UUID;
```
maedhroz: nit: import order

src/java/org/apache/cassandra/hints/HintsWriter.java
```diff
@@ -37,6 +27,16 @@
...
 import org.apache.cassandra.utils.Throwables;
 
+import java.io.IOException;
+import java.io.OutputStream;
+import java.nio.ByteBuffer;
+import java.nio.channels.FileChannel;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
+import java.util.zip.CRC32;
```
maedhroz: nit: import order

src/java/org/apache/cassandra/hints/HintsWriter.java
```diff
@@ -248,7 +248,10 @@ void append(Hint hint) throws IOException
                 updateChecksumInt(crc, hintSize);
                 out.writeInt((int) crc.getValue());
 
+                long startPosition = out.position();
                 Hint.serializer.serialize(hint, out, descriptor.messagingVersion());
+                long actualSize = out.position() - startPosition;
+                checkState(actualSize == hintSize);
```
maedhroz: Not directly related to the patch, I guess, but seems harmless.

src/java/org/apache/cassandra/io/util/DataInputPlus.java
```diff
@@ -30,11 +30,31 @@
 @Shared(scope = SIMULATION)
 public interface DataInputPlus extends DataInput
 {
+    /**
+     * Read a 64-bit integer back. This assumes it was originally written using writeVInt
+     * and uses zigzag encoding.
+     * @return
+     * @throws IOException
```
maedhroz: nit: remove `@throws IOException` unless it tells us more than the method signature already does

src/java/org/apache/cassandra/cql3/functions/types/TypeCodec.java
aweisberg: It's updated now.

src/java/org/apache/cassandra/io/util/DataInputPlus.java
```diff
@@ -30,11 +30,31 @@
 @Shared(scope = SIMULATION)
 public interface DataInputPlus extends DataInput
 {
+    /**
+     * Read a 64-bit integer back. This assumes it was originally written using writeVInt
+     * and uses zigzag encoding.
+     * @return
```
maedhroz: nit: Could remove `@return`...or potentially move the description into it? Maybe something like...

```
@return a 64-bit integer, assumed to originally have been written using {@link DataOutputPlus#writeVInt(long)}
```

src/java/org/apache/cassandra/utils/vint/VIntCoding.java
aweisberg: Made it public, let me know if you think that makes sense.

src/java/org/apache/cassandra/cql3/functions/types/TypeCodec.java
aweisberg: Updated my style and reformatted them. Now it just fixes one alphabetical issue.

src/java/org/apache/cassandra/io/util/DataInputPlus.java
```diff
@@ -30,11 +30,31 @@
...
     {
         return VIntCoding.readVInt(this);
     }
 
+    /**
+     * Read up to a 32-bit integer back, throwing IllegalArgumentException if
+     * the vint read is out of range for a 32-bit integer.
+     *
+     * This method assumes the integer was originally written using writeVInt32
+     * and uses zigzag encoding.
```
maedhroz: So in general, I don't care a ton how we document return stuff here, as long as it's consistent across this class. ex. We could replace the above w/ just...
```
@return a 32-bit integer, assumed to originally have been written using {@link DataOutputPlus#writeVInt32(int)}
```
...assuming you like that format.

src/java/org/apache/cassandra/io/util/RebufferingInputStream.java
```diff
@@ -231,11 +226,18 @@ public long readLong() throws IOException
             return readPrimitiveSlowly(8);
     }
 
+    @Override
     public long readVInt() throws IOException
     {
         return VIntCoding.decodeZigZag64(readUnsignedVInt());
     }
 
+    public int readVInt32() throws IOException
```
maedhroz, line 237: nit: `@Override`?

src/java/org/apache/cassandra/hints/HintsWriter.java
aweisberg: Ah, yeah was debugging something I broke and felt it was worth leaving in.

src/java/org/apache/cassandra/utils/vint/VIntCoding.java
```diff
@@ -165,6 +190,27 @@ public static long readVInt(DataInput input) throws IOException
         return decodeZigZag64(readUnsignedVInt(input));
     }
 
+    /**
+     * Read up to a signed 32-bit integer back, throwing IllegalArgumentException if
+     * the vint read is out of range for a 32-bit integer
+     * @throws IllegalArgumentException Thrown if the vint doesn't fit a 32-bit integer
```
maedhroz:
```suggestion
     * @throws VIntOutOfRangeException if the vint doesn't fit a 32-bit integer
```

src/java/org/apache/cassandra/db/columniterator/AbstractSSTableIterator.java
aweisberg: Fixed it, there were still a few imports out of order that are changed.

src/java/org/apache/cassandra/utils/vint/VIntCoding.java
```diff
@@ -101,14 +116,24 @@ public static void skipUnsignedVInt(DataInputPlus input) throws IOException
...
+     * Read up to a 32-bit integer back, using the unsigned (no zigzag) encoding.
+     *
      * Note this method is the same as {@link #readUnsignedVInt(DataInput)},
      * except that we do *not* block if there are not enough bytes in the buffer
      * to reconstruct the value.
      *
-     * WARNING: this method is only safe for vints we know to be representable by a positive long value.
-     *
-     * @return -1 if there are not enough bytes in the input to read the value; else, the vint unsigned value.
+     * @throws IllegalArgumentException If the vint doesn't fit a 32-bit integer
```
maedhroz:
```suggestion
     * @throws VIntOutOfRangeException if the vint doesn't fit a 32-bit integer
     * throws 
```

test/unit/org/apache/cassandra/db/streaming/ComponentManifestTest.java
```diff
@@ -42,20 +33,23 @@ public void testSerialization()
...
-
-        ComponentManifest.serializer.serialize(expected, out, MessagingService.VERSION_40);
-
-        buf.putInt(0, -100);
-
-        DataInputBuffer in = new DataInputBuffer(out.buffer(), false);
-        ComponentManifest actual = ComponentManifest.serializer.deserialize(in, MessagingService.VERSION_40);
-        assertNotEquals(expected, actual);
-    }
+    // Propose removing this test which now fails on VIntOutOfRange
```
maedhroz, line 36: WFM

maedhroz (approved):
My feedback is all pretty minor around documentation, import order, etc.

Otherwise, LGTM

src/java/org/apache/cassandra/utils/vint/VIntCoding.java
maedhroz: WFM

src/java/org/apache/cassandra/io/util/DataInputPlus.java
aweisberg: I think it's possibly too verbose? The return is just an integer. How it was encoded is more a matter of how the method works and where it gets it from? Seems like we shouldn't provide that information twice.

I vote for removing return.

test/unit/org/apache/cassandra/db/streaming/ComponentManifestTest.java
aweisberg: @dineshjoshi  WDYT? You added this test. What was the goal?

src/java/org/apache/cassandra/utils/vint/VIntCoding.java
aweisberg: Fixed

src/java/org/apache/cassandra/utils/vint/VIntCoding.java
aweisberg: Fixed

src/java/org/apache/cassandra/utils/vint/VIntCoding.java
aweisberg: Changed it

src/java/org/apache/cassandra/io/util/RebufferingInputStream.java
aweisberg: Added it.

src/java/org/apache/cassandra/io/util/DataInputPlus.java
aweisberg: Removed the return..

src/java/org/apache/cassandra/io/util/DataInputPlus.java
aweisberg: Removed it.

src/java/org/apache/cassandra/io/util/DataInputPlus.java
maedhroz: Yeah, I was more suggesting replacing the description, not providing the information twice. Anyway, just removing the `@return` is fine w/ me.

src/java/org/apache/cassandra/hints/HintsWriter.java
```diff
@@ -248,7 +249,10 @@ void append(Hint hint) throws IOException
...
```
dcapwell: can we have a useful error msg?  If this triggers we don't have enough info on what happened

src/java/org/apache/cassandra/io/util/DataOutputPlus.java
```diff
@@ -47,6 +47,17 @@ default void writeVInt(long i) throws IOException
         VIntCoding.writeVInt(i, this);
     }
 
+    @Deprecated
+    default void writeVInt(int i)
+    {
+        throw new UnsupportedOperationException("Use writeVInt32/readVInt32");
+    }
```
dcapwell, line 54: can we just remove?  This is a private interface right?

src/java/org/apache/cassandra/io/util/DataOutputPlus.java
```diff
@@ -59,6 +70,17 @@ default void writeUnsignedVInt(long i) throws IOException
         VIntCoding.writeUnsignedVInt(i, this);
     }
 
+    @Deprecated
+    default void writeUnsignedVInt(int i)
+    {
+        throw new UnsupportedOperationException("Use writeUnsignedVInt32/readUnsignedVInt32");
+    }
```
dcapwell, line 77: can we just remove?

dcapwell (changes requested):
can we address the interface comment?  Rest are nits

src/java/org/apache/cassandra/io/util/DataOutputPlus.java
dcapwell: spoke in slack, if we do remove `writeVInt(long)` will match =(

dcapwell (approved):
would be great to improve the error message of hint, but will leave that as a nit and not a blocker.

+1

src/java/org/apache/cassandra/io/util/DataOutputPlus.java
aweisberg: It's the same as with `writeVInt` where the long version matches if we don't.

PR #2045, opened 2022-12-08
Cassandra 8928 -- added sstable downgrader

Claudenw:
Fix for CASSANDRA-8928

This is a work in progress and I am looking for some feedback.

This fix appears to work correctly.  But I think the placement of the v3 directory is probably not the best and perhaps should be moved under the db/compaction directory where the Downgrader code is.

Suggestions appreciated.

## The changes are:

**added** 

- db/compaction/Downgrader
- tools/StandAloneDowngrader
- getNextsstableID to ColumnFamilyStore
- DOWNGRADE_SSTABLES to OperationType
- hasMaxCompressedLength to CompressionMetadata constructor and associated calls.
- V3 to SSTableFormat.Type
- added io/sstable/format/big/v3 directory containing BigFormatV3, BigTableReaderV3, BitTableScannerV3, BitTableWriterV3

**modified**

- CompressionMetadata to skip output of maxCompressedLength if not supported

**notes**

- io/sstable/format/big/v3 classes are the V3 classes modified as necessary to run within the V4 environment.


src/java/org/apache/cassandra/db/compaction/Downgrader.java
```diff
@@ -0,0 +1,182 @@
...
+
+    private final SSTableFormat.Type inputType;
+    private final SSTableFormat.Type outputType;
+    private final Version outputVersion;
+
+    private final OutputHandler outputHandler;
+
+    public static SSTableFormat.Type forWriting(String version)
+    {
+        String[] v3 =
```
jacek-lewandowski: v3 starts with MA

src/java/org/apache/cassandra/db/compaction/OperationType.java
```diff
@@ -26,6 +26,7 @@
     CLEANUP("Cleanup", true, 1),
     SCRUB("Scrub", true, 1),
     UPGRADE_SSTABLES("Upgrade sstables", true, 1),
+    DOWNGRADE_SSTABLES("Downgrade sstables", true, 1),
```
jacek-lewandowski: perhaps not important in this case, but we should always add enums at the end because otherwise, if they are serialized as numbers somewhere, that can break the serialization

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
```diff
@@ -43,7 +43,8 @@
     public static enum Type
     {
         //The original sstable format
-        BIG("big", BigFormat.instance);
+        BIG("big", BigFormat.instance),
+        V3("bigv3", BigFormatV3.instance);
```
jacek-lewandowski: Well, this is not a different format, it is a different version

Claudenw: That is true, but version and format seem to be tightly coupled, unless I am missing something.  The format of some of the files has changed.  And the writer can no longer write the earlier versions.

src/java/org/apache/cassandra/db/compaction/OperationType.java
Claudenw: fixed

src/java/org/apache/cassandra/db/compaction/Downgrader.java
Claudenw: fixed

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
tjake: Correct.  I think you need to better document the concept which is to "Freeze" the format from the last major release.  Similar to how C* supports the last major Messaging Version to support live upgrade and downgrade.

src/java/org/apache/cassandra/io/sstable/format/big/BigTableScannerV3.java
```diff
@@ -0,0 +1,497 @@
...
+import org.apache.cassandra.io.sstable.format.SSTableReadsListener;
+import org.apache.cassandra.io.util.FileUtils;
+import org.apache.cassandra.io.util.RandomAccessReader;
+import org.apache.cassandra.utils.ByteBufferUtil;
+
+import static org.apache.cassandra.dht.AbstractBounds.isEmpty;
+import static org.apache.cassandra.dht.AbstractBounds.maxLeft;
+import static org.apache.cassandra.dht.AbstractBounds.minRight;
+
+public class BigTableScannerV3 implements ISSTableScanner
```
tjake: Since our contract is to always support upgrading from prior versions we should be able to re-use the `BigTableScanner` meaning no need to copy this one

src/java/org/apache/cassandra/io/sstable/format/big/BigFormatV3.java
```diff
@@ -0,0 +1,309 @@
...
+        /**
+         * CASSANDRA-7066: compaction ancerstors are no longer used and have been removed.
+         */
+        private final boolean hasCompactionAncestors;
+        /**
+         * CASSANDRA-9067: 4.0 bloom filter representation changed (two longs just swapped)
+         * have no 'static' bits caused by using the same upper bits for both bloom filter and token distribution.
+         */
+        private final boolean hasOldBfFormat;
+        BigVersion(String version)
```
tjake, line 170: The logic here can be simplified to be just the flags from the `current_version`.  Otherwise this gives the impression it can write all these versions.  I would change the comments to make this clear

Claudenw:
@jacek-lewandowski @tjake This code is to downgrade from 4.x to 3.x  So I think it needs to be rebased to a 4.x branch.  However, I think that because it is added functionality it should be a new 4.2 not a 4.1 branch.  In any case I think I need to rebase this to 4.1 (rather than trunk).  Do you agree or should I do this differently?

PR #2047, opened 2022-12-10
Add compaction type output result for nodetool compactionhistory for CASSANDRA-18061

Maxwell-Guo:
Add compaction type output result for nodetool compactionhistory for CASSANDRA-18061

The [Cassandra 18061](https://issues.apache.org/jira/browse/CASSANDRA-18061)


src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryPrinter.java
```diff
@@ -29,6 +29,8 @@
 
 public class CompactionHistoryPrinter
 {
+    public static final String UNKNOW_TYPE = "UNKNOW";
```
smiklosovic: `UNKNOWN_TYPE = "UNKNOWN";` - Missing `N` at the end of `UNKNOW`.

This constant should be moved to `CompactionHistoryTabularData`

src/java/org/apache/cassandra/cql3/UntypedResultSet.java
```diff
@@ -351,7 +351,11 @@ public ByteBuffer getBlob(String column)
 
         public String getString(String column)
         {
-            return UTF8Type.instance.compose(data.get(column));
```
smiklosovic: this change is not necessary, leave that method as it was before

src/java/org/apache/cassandra/db/compaction/CompactionHistoryTabularData.java
```diff
@@ -74,10 +74,11 @@ public static TabularData from(UntypedResultSet resultSet) throws OpenDataExcept
             long bytesIn = row.getLong(ITEM_NAMES[4]);
             long bytesOut = row.getLong(ITEM_NAMES[5]);
             Map<Integer, Long> rowMerged = row.getMap(ITEM_NAMES[6], Int32Type.instance, LongType.instance);
-
+            String compactionType = row.getString(ITEM_NAMES[7]);
```
smiklosovic: if `row.getString` returns `null` here, `compactionType` should be set to `UNKNOWN` here.

src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryPrinter.java
```diff
@@ -65,14 +67,15 @@ public void print(CompactionHistoryHolder data, PrintStream out)
...
-                String[] obj = new String[7];
+                String[] obj = new String[8];
                 obj[0] = (String)value.get("id");
                 obj[1] = (String)value.get("keyspace_name");
                 obj[2] = (String)value.get("columnfamily_name");
                 obj[3] = (String)value.get("compacted_at");
                 obj[4] = value.get("bytes_in").toString();
                 obj[5] = value.get("bytes_out").toString();
                 obj[6] = (String)value.get("rows_merged");
+                obj[7] = value.get("compaction_type") == null ?  UNKNOW_TYPE : (String)value.get("compaction_type");
```
smiklosovic: this should be `obj[7] = value.get("compaction_type");` because it already contains type or `UNKNOWN`.

src/java/org/apache/cassandra/cql3/UntypedResultSet.java
Maxwell-Guo: I modify all other comment, but for this ,I should left it unchangede. When we upgrade from previous version that have not got the new added compaction_type column , the data.get(column) will return null ,and so a npe will appears, we can see the getInt and getUUID method below ,they have consider the case of null value.


Firstly , I considered changing all the methods of the corresponding type of get method  to consider the case where value is null, but I gave up later. I thought it would be more appropriate to open a new jira.

Maxwell-Guo: besides I have also tried upgrade my cluster from a previous version from this version for Cassandra-18062
if  leave that method as it was before a npe appears,

src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryPrinter.java
Maxwell-Guo: ok

src/java/org/apache/cassandra/db/compaction/CompactionHistoryTabularData.java
Maxwell-Guo: ok

src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryPrinter.java
Maxwell-Guo: +1

src/java/org/apache/cassandra/cql3/UntypedResultSet.java
smiklosovic: @Maxwell-Guo You can use `public boolean has(String column)` in that class to check if it is null and if it is not then you call `getString` otherwise you return null. It is not necessary to change this method.

It is ok to open new ticket to investigate if we should not solve this in more robust manner but that is not objective of this ticket.

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -90,7 +90,7 @@ public UpgradeableCluster.Builder builder()
     public static final Semver v30 = new Semver("3.0.0-alpha1", SemverType.LOOSE);
     public static final Semver v3X = new Semver("3.11.0", SemverType.LOOSE);
     public static final Semver v40 = new Semver("4.0-alpha1", SemverType.LOOSE);
-    public static final Semver v41 = new Semver("4.1-alpha1", SemverType.LOOSE);
+    public static final Semver v41 = new Semver("4.1", SemverType.LOOSE);
```
smiklosovic: @Maxwell-Guo why have you changed this?

Maxwell-Guo: @smiklosovic  Thank you for you review .There was something urgent yesterday at that time , so I seems that I have forgot to check this place when submitting.
I have changed it back.

test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
```diff
@@ -39,26 +39,22 @@ public void compactionHistorySystemTableTest() throws Throwable
...
             {
                 cluster.coordinator(1).execute("INSERT INTO " + KEYSPACE + ".tb (pk, c1, c2, c3) VALUES ('pk" + i + "', 'c1" + i + "', " + i + ',' + i + ')'  , ConsistencyLevel.ALL);
                 cluster.stream().forEach(node -> node.flush(KEYSPACE));
             }
-          
             // force compact
             cluster.stream().forEach(node -> node.forceCompact(KEYSPACE, "tb"));
-          
         }).runAfterNodeUpgrade((cluster, node) -> {
           String query = "SELECT compaction_type FROM " + KEYSPACE + ".tb where keyspace_name = '" + KEYSPACE + "' AND columnfamily_name = 'tb' ALLOW FILTERING";
```
smiklosovic: @Maxwell-Guo I do not think this is right. You are selecting from keyspace `KEYSPACE` and table `tb`. Should not that be `system.compaction_history`? Have you tried to execute this test?

test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
```diff
@@ -0,0 +1,77 @@
...
+          };
+          assertRows(cluster.coordinator(1).execute(query, ConsistencyLevel.ALL), expectedResult);
+          assertRows(cluster.coordinator(2).execute(query, ConsistencyLevel.ALL), expectedResult);
+
+          ToolRunner.ToolResult toolHistory = invokeNodetool("compactionhistory");
+          toolHistory.assertOnCleanExit();
+          String stdout = toolHistory.getStdout();
+          String [] resultArray = stdout.split("\n");
+          assertTrue(Arrays.stream(resultArray)
+              .anyMatch(result -> result.contains(UNKNOWN_TYPE)
```
smiklosovic: @Maxwell-Guo why do we expect UNKNOWN_TYPE here? Should not that be some concrete type?

smiklosovic: aha I see, we are testing the output from times it was not there yet.

src/java/org/apache/cassandra/db/compaction/CompactionHistoryTabularData.java
```diff
@@ -74,10 +74,14 @@ public static TabularData from(UntypedResultSet resultSet) throws OpenDataExcept
             long bytesIn = row.getLong(ITEM_NAMES[4]);
             long bytesOut = row.getLong(ITEM_NAMES[5]);
             Map<Integer, Long> rowMerged = row.getMap(ITEM_NAMES[6], Int32Type.instance, LongType.instance);
-
+            String compactionType = OperationType.UNKNOWN.type;
+            if (row.has(ITEM_NAMES[7]))
```
jacek-lewandowski: it does not seem like this check is really needed since the system table migrator should add that column before we ever execute the code here

test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+
+import static org.apache.cassandra.distributed.shared.AssertUtils.assertRows;
+import static org.apache.cassandra.distributed.shared.AssertUtils.row;
+import static org.apache.cassandra.tools.ToolRunner.invokeNodetoolJvmDtest;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(Parameterized.class)
+public class CompactionHistorySystemTableUpgradeTest extends UpgradeTestBase
```
jacek-lewandowski, line 38: I'm not sure, but such things are tested in system migrator test

jacek-lewandowski (changes requested):
I think that the additional column should be added in the SystemKeyspaceMigrator and test for that should be added to SystemKeyspaceMigratorTest. The column should be populated with the default value as part of the migration.

src/java/org/apache/cassandra/db/compaction/CompactionHistoryTabularData.java
Maxwell-Guo: Thanks @jacek-lewandowski 
Yes, if a migration for this system table  is added , This method is no longer need.

test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
Maxwell-Guo: If a migrate for compation_history is added in SystemKeyspaceMigrator , this method should be changed. 
I just want to doing some test on the result of upgrate ,So I think we can keep this test . right ?

src/java/org/apache/cassandra/db/compaction/CompactionHistoryTabularData.java
jacek-lewandowski: I think it should be, otherwise when this column is added?

test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
jacek-lewandowski: Depending on whether it provides some value of the test in migrator. If it provides some better coverage, then ok. But in general upgrade tests are quite heavy and if we can verify the same in a pretty simple unit/simple CQLTester based integration test I would go that way

Maxwell-Guo: Hi @jacek-lewandowski , I keep this test case  with the test of nodetool compactionhistory. I think the test case can not be covered by SystemKeyspaceMigratorTest

src/java/org/apache/cassandra/db/compaction/CompactionHistoryProperty.java
```diff
@@ -0,0 +1,63 @@
...
+    public static final String[] COMPACTION_PROPERTIES_KEYS = { "compaction_type" };
+    // define the map value for the compaction_properties column in system.compaction_history
+    public static final String[] DEFAULT_COMPACTION_PROPERTIES_VALUES = { OperationType.UNKNOWN.type };
+    
+    /**
+     * get the compaction history properties , if none value exists  a default value will return
+     * @param propertiesColumnValueInput input properties map
+     * @return the properties
+     * */
+    public static Map<String, String> getCompactionHistroyProperties(Map<String, String> propertiesColumnValueInput)
```
jacek-lewandowski: This method seems to be too complex to what it really does. What do you think about simply declaring an immutable map with defaults, and then in the method pre-populate the map you want to return with those defaults?

src/java/org/apache/cassandra/db/compaction/CompactionHistoryProperty.java
```diff
@@ -0,0 +1,63 @@
...
+import java.util.Map;
+import java.util.TreeMap;
+
+/**
+ * properties that for system.compaction_history, and more properties can be add in the future
+ * */
+public class CompactionHistoryProperty
+{
+    // define the map key for the compaction_properties column in system.compaction_history
+    public static final String[] COMPACTION_PROPERTIES_KEYS = { "compaction_type" };
```
jacek-lewandowski: would be good to have constants for certain property names

src/java/org/apache/cassandra/db/compaction/CompactionHistoryTabularData.java
```diff
@@ -24,16 +24,17 @@
...
 import org.apache.cassandra.db.marshal.Int32Type;
 import org.apache.cassandra.db.marshal.LongType;
+import org.apache.cassandra.db.marshal.UTF8Type;
 import org.apache.cassandra.utils.FBUtilities;
 
 public class CompactionHistoryTabularData
 {
     private static final String[] ITEM_NAMES = new String[]{ "id", "keyspace_name", "columnfamily_name", "compacted_at",
-                                                             "bytes_in", "bytes_out", "rows_merged", "compaction_type" };
+                                                             "bytes_in", "bytes_out", "rows_merged", "compaction_properties" };
```
jacek-lewandowski, line 33: what do you think about including all the keys defined in CompactionHstoryProperty?

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -244,7 +247,8 @@ public boolean apply(SSTableReader sstable)
             for (int i = 0; i < mergedRowCounts.length; i++)
                 totalSourceRows += mergedRowCounts[i] * (i + 1);
 
-            String mergeSummary = updateCompactionHistory(cfs.keyspace.getName(), cfs.getTableName(), mergedRowCounts, startsize, endsize, compactionType.type);
+            String mergeSummary = updateCompactionHistory(cfs.keyspace.getName(), cfs.getTableName(), mergedRowCounts, startsize, endsize,
+                ImmutableMap.of(CompactionHistoryProperty.COMPACTION_PROPERTIES_KEYS[0], compactionType.type));
```
jacek-lewandowski: This is a good place to use a constant for compaction type property name

jacek-lewandowski:
I have a question - where exactly the new column is added to the table schema?

src/java/org/apache/cassandra/db/compaction/CompactionHistoryProperty.java
Maxwell-Guo: +1 on declaring an immutable map with defaults, I can change the code

I think there may be two case :
1. when upgrate from lower version that do not contains the compaction_properties column , then all the default value can just return.

2.If some new properties was added in the compaction_properties column, such as for 4.2 we have compaction_type , but we may add new property for example compaction_time or resouce cost at 5.0 , then compaction_type is already exits but "compaction_time or resouce cost" may use a default value.

src/java/org/apache/cassandra/db/compaction/CompactionHistoryProperty.java
Maxwell-Guo: +1

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
Maxwell-Guo: +1

Maxwell-Guo:
The new column is added in [SystemKeyspace ](https://github.com/apache/cassandra/pull/2047/files#diff-f57518f964c71328146aeca95be5e697ca81a77261719eeef4dd4b1ed8daf63bR597)
In SystemKeyspace.java at the first time CompactionHistory's TableMeta parsed.

src/java/org/apache/cassandra/db/compaction/CompactionHistoryTabularData.java
Maxwell-Guo: 1. I just put all the properties together in a map as you said. The reason why I do not including all keys defined in CompactionHstoryProperty is that the keys may be so much , and for nodetool compactionhistory 's output it may be not easy to see. 
2. Besides, in the future if some one add new properties , he may mdofiy more than one place to suceesfully add a property, define the key and default value in CompactionHistoryProperty, and add property key here for ITEM_NAMES and ITEM_DESCS, and the ITEM_TYPES should be added with new property type.
And the out put result should also be modified to get the value from the map. 
But if use this compaction_properties , none of the modification described in part 2 is needed. And the out put result is just like rows_merged column ,which is also a map type , and the output is also looks better

src/java/org/apache/cassandra/db/SystemKeyspaceMigrator41.java
```diff
@@ -159,10 +162,36 @@ static void migrateSSTableActivity()
...
+                                                                    row.has("rows_merged") ? row.getMap("rows_merged", Int32Type.instance, LongType.instance) : null,
+                                                                    CompactionHistoryProperty.getCompactionHistroyProperties(row.getMap("compaction_properties", UTF8Type.instance, UTF8Type.instance))
+                     })
+        );
+    }
 
     /**
      * Perform table migration by reading data from the old table, converting it, and adding to the new table.
-     *
+     * if oldName and newName are same, means refresh data in table
```
smiklosovic: `If oldName and newName are same, it means data in the table will be refreshed.`

src/java/org/apache/cassandra/db/SystemKeyspaceMigrator41.java
```diff
@@ -159,10 +162,36 @@ static void migrateSSTableActivity()
...
+                     })
+        );
+    }
 
     /**
      * Perform table migration by reading data from the old table, converting it, and adding to the new table.
-     *
+     * if oldName and newName are same, means refresh data in table 
      * @param truncateIfExists truncate the existing table if it exists before migration; if it is disabled
      *                         and the new table is not empty, no migration is performed
```
smiklosovic: `and the new table is not empty and newName is not equal to oldName, no migration is performed`

smiklosovic: btw @jacek-lewandowski this line in this method may produce null:

````
UntypedResultSet rows = QueryProcessor.executeInternal(query);
````

I suggest to either assert it is not null or if it is we log on warn level and return from this method.

File: src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryHolder.java [hunk 16]
```diff
@@ -54,8 +54,9 @@ private static class CompactionHistoryRow implements Comparable<CompactionHistor
         private final long bytesIn;
         private final long bytesOut;
         private final String rowMerged;
+        private final String compactionType;
 
-        CompactionHistoryRow(String id, String ksName, String cfName, long compactedAt, long bytesIn, long bytesOut, String rowMerged)
+        CompactionHistoryRow(String id, String ksName, String cfName, long compactedAt, long bytesIn, long bytesOut, String rowMerged, String compactionType)
```
smiklosovic: @Maxwell-Guo you have not changed name of this property to `compactionProperties`.

src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryHolder.java
```diff
@@ -64,6 +65,7 @@ private static class CompactionHistoryRow implements Comparable<CompactionHistor
             this.bytesIn = bytesIn;
             this.bytesOut = bytesOut;
             this.rowMerged = rowMerged;
+            this.compactionType = compactionType;
```
smiklosovic: @Maxwell-Guo you have not changed name of this property to `compactionProperties`.

src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryHolder.java
```diff
@@ -83,6 +85,7 @@ private HashMap<String, Object> getAllAsMap()
             compaction.put("bytes_in", this.bytesIn);
             compaction.put("bytes_out", this.bytesOut);
             compaction.put("rows_merged", this.rowMerged);
+            compaction.put("compaction_type", this.compactionType);
```
smiklosovic: @Maxwell-Guo you have not changed name of this property to `compaction_property`.

src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryPrinter.java
```diff
@@ -65,14 +65,15 @@ public void print(CompactionHistoryHolder data, PrintStream out)
...
-                String[] obj = new String[7];
+                String[] obj = new String[8];
                 obj[0] = (String)value.get("id");
                 obj[1] = (String)value.get("keyspace_name");
                 obj[2] = (String)value.get("columnfamily_name");
                 obj[3] = (String)value.get("compacted_at");
                 obj[4] = value.get("bytes_in").toString();
                 obj[5] = value.get("bytes_out").toString();
                 obj[6] = (String)value.get("rows_merged");
+                obj[7] = (String)value.get("compaction_type");
```
smiklosovic: @Maxwell-Guo `compaction_properties`

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -322,7 +322,6 @@ public void run() throws Throwable
                 try (UpgradeableCluster cluster = init(UpgradeableCluster.create(nodeCount, upgrade.initial, configConsumer, builderConsumer)))
                 {
                     setup.run(cluster);
-
```
smiklosovic, line 325: @Maxwell-Guo put that line back please

src/java/org/apache/cassandra/db/compaction/CompactionHistoryTabularData.java
smiklosovic: @jacek-lewandowski I think map is just fine here with compaction_properties representing it. Everytime we add an entry we would need to update this. Was not the primary goal of map introduction to actually have something nice we can throw things in when we feel like that?

src/java/org/apache/cassandra/db/compaction/CompactionHistoryTabularData.java
```diff
@@ -44,13 +45,12 @@
     private static final CompositeType COMPOSITE_TYPE;
 
     private static final TabularType TABULAR_TYPE;
-
```
smiklosovic, line 47: @Maxwell-Guo why is this line removed?

src/java/org/apache/cassandra/db/compaction/CompactionHistoryTabularData.java
```diff
@@ -74,10 +74,11 @@ public static TabularData from(UntypedResultSet resultSet) throws OpenDataExcept
             long bytesIn = row.getLong(ITEM_NAMES[4]);
             long bytesOut = row.getLong(ITEM_NAMES[5]);
             Map<Integer, Long> rowMerged = row.getMap(ITEM_NAMES[6], Int32Type.instance, LongType.instance);
-
+            Map<String, String> compactionProperties = row.getMap(ITEM_NAMES[7], UTF8Type.instance, UTF8Type.instance);
             result.put(new CompositeDataSupport(COMPOSITE_TYPE, ITEM_NAMES,
                        new Object[]{ id.toString(), ksName, cfName, compactedAt, bytesIn, bytesOut,
-                                     "{" + FBUtilities.toString(rowMerged) + "}" }));
+                                     '{' + FBUtilities.toString(rowMerged) + '}' ,
```
smiklosovic: formatting, it should be

````
'{' + FBUtilities.toString(rowMerged) + '}',
'{' + FBUtilities.toString(compactionProperties) + '}' }));
````

1) there is redundant space at `'}' ,`
2) the second line is not aligned wit the first one

File: test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java [hunk 23]
```diff
@@ -0,0 +1,88 @@
...
+    @Parameterized.Parameter
+    public Semver version;
+
+    @Parameterized.Parameters()
+    public static ArrayList<Semver> versions()
+    {
+      return Lists.newArrayList(v30, v3X, v40, v41);
+    }
+
+  @Test
```
smiklosovic: @Maxwell-Guo `@Test` is not aligned. Do you use code formatter?

src/java/org/apache/cassandra/db/SystemKeyspaceMigrator41.java
Maxwell-Guo: ok

src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryHolder.java
Maxwell-Guo: ok

src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryHolder.java
Maxwell-Guo: thk

src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryHolder.java
Maxwell-Guo: ok

src/java/org/apache/cassandra/tools/nodetool/stats/CompactionHistoryPrinter.java
Maxwell-Guo: ok

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
Maxwell-Guo: ok

test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
Maxwell-Guo: thk,For I have more than one project to do, and they are of different format, so some times they conflict....
next time I will check the format,before push

src/java/org/apache/cassandra/db/compaction/CompactionHistoryProperty.java
Maxwell-Guo: @smiklosovic 
"Could you please change the type of compaction_properties to frozen<map<text, text>> ?

We were suggesting in above comments to have empty map inserted with no defaults in case the original value was null.

When that is done, when we migrate, we would insert empty map ({}) into compaction_properties and not map with one entry. It is very easy that way as we do not need any additional logic in CompactionHistoryProperty and I think that class can be completely removed."

my question is : if deafult value is not needed ? if deafult value is need, then we can come to  the two points above .
1.from 3.x to current version , the column value is null , so a complete default map is needed
2.form current version to some upper version in the future like 5.x , and we need to add more properties(such as 3 properties) to the column which already  have some properties (such as 2), so the 2 old properties should use their real value but the 3 new should use default.

test/unit/org/apache/cassandra/tools/nodetool/CompactionHistoryTest.java
```diff
@@ -0,0 +1,142 @@
...
+    public static void setup() throws Exception
+    {
+        requireNetwork();
+        startJMXServer();
+    }
+
+    @Test
+    public void testSystemCompactionHistoryTable() throws Throwable
+    {
+        createTable("CREATE TABLE %s (id text, value text, PRIMARY KEY ((id)))");
```
jacek-lewandowski: most of the code in each test looks very similar (or identical). Why don't you extract and reuse it or create a parameterized test?

test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
```diff
@@ -0,0 +1,84 @@
...
+    public static ArrayList<Semver> versions()
+    {
+      return Lists.newArrayList(v30, v3X, v40, v41);
+    }
+
+    @Test
+    public void compactionHistorySystemTableTest() throws Throwable
+    {
+        new TestCase()
+            .nodes(2)
```
jacek-lewandowski: why do you need 2 nodes? compaction is a local process though

test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
```diff
@@ -0,0 +1,84 @@
...
+            // generate sstables
+            for (int i = 0; i != 10; ++ i)
+            {
+                cluster.coordinator(1).execute("INSERT INTO " + KEYSPACE + ".tb (pk, c1, c2, c3) VALUES ('pk" + i + "', 'c1" + i + "', " + i + ',' + i + ')'  , ConsistencyLevel.ALL);
+                cluster.stream().forEach(node -> node.flush(KEYSPACE));
+            }
+            // force compact
+            cluster.stream().forEach(node -> node.forceCompact(KEYSPACE, "tb"));
+        }).runAfterClusterUpgrade((cluster) -> {
+          ToolRunner.ToolResult toolHistory = invokeNodetoolJvmDtest(cluster.get(1), "compactionhistory");
```
jacek-lewandowski: well, think about reusing the assertion against the output from `CompactionHistoryTest`

jacek-lewandowski:
also needs an update in NEWS.txt and CHANGES.txt

test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
smiklosovic: @Maxwell-Guo this is good question, why two nodes?

test/unit/org/apache/cassandra/tools/nodetool/CompactionHistoryTest.java
smiklosovic: @Maxwell-Guo good idea to do it as @jacek-lewandowski suggests.

test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
smiklosovic: @Maxwell-Guo I think that Jacek wants to say that you should check the result containing "{compaction_properties....}" as well.

src/java/org/apache/cassandra/db/compaction/CompactionHistoryProperty.java
smiklosovic: @Maxwell-Guo I am sorry, I do not understand.

test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
Maxwell-Guo: Yes one is ok for local compaction.
For the test is under a upgrade of  distributed dir （[org/apache/cassandra/distributed/upgrade/).And I think more than one node feels closer to the production system. In fact, I have three nodes at first. I think it can also cover the serial restart of nodes during the upgrade process,like gossip communctaion and so on.

smiklosovic: 1 node is enough as it is a table in the `system` keyspace which is local and it is not distributed anywhere.

test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
smiklosovic: you are still checking an empty map, not a map with an entry key with name `compaction_type` and some value

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
jacek-lewandowski: I think you should just rollback the change on this file

test/unit/org/apache/cassandra/tools/nodetool/CompactionHistoryTest.java
```diff
@@ -0,0 +1,133 @@
...
+        ColumnFamilyStore cfs = Keyspace.open(keyspace()).getColumnFamilyStore(currentTable());
+        cfs.disableAutoCompaction();
+        // write SSTables for the specific key
+        for (int i = 0; i < 10; i++)
+        {
+            execute("INSERT INTO %s (id, value) VALUES (?, ?)", "key" + i, "value" + i);
+            flush(keyspace());
+            if(compactionType.equals(OperationType.GARBAGE_COLLECT.type))
```
jacek-lewandowski: just curious, aren't deletions valid for other types of compaction?

test/unit/org/apache/cassandra/tools/nodetool/CompactionHistoryTest.java
```diff
@@ -0,0 +1,133 @@
...
+            {
+                execute("DELETE FROM %s WHERE id = ? ", "key" + i);
+                flush(keyspace());
+            }
+        }
+        
+        int sstableCount = compactionType.equals(OperationType.GARBAGE_COLLECT.type) ? 20 : 10;
```
jacek-lewandowski: whether 10 or 20 is a parameter, isn't it?

jacek-lewandowski (approved):
Great job,
I left some nits, but I'm +1

test/unit/org/apache/cassandra/tools/nodetool/CompactionHistoryTest.java
smiklosovic: @jacek-lewandowski yes you are right, I think that `garbagecollect` is used after cluster expansion and it is executed when data are not held by that respective node anymore as token assignment has changed. However, for the sake of testing this, it is ok.

What I do is that I will remove this logic and we will not need the below code either.

test/unit/org/apache/cassandra/tools/nodetool/CompactionHistoryTest.java
smiklosovic: This will be changed to 10 only based on the above comment.

jacek-lewandowski (approved):

PR #2049, opened 2022-12-12
CASSANDRA-18100 CAS and serial read support for Accord

aweisberg:
[CASSANDRA-18100](https://issues.apache.org/jira/browse/CASSANDRA-18100?filter=-3)

### DIFF
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -21,19 +21,17 @@ import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
 import java.util.function.Supplier;
-
 import javax.annotation.Nullable;
 
 import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Sets;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -1054,6 +1052,8 @@ public class Config
      */
     public volatile int paxos_repair_parallelism = -1;
 
+    public LegacyPaxosStrategy legacy_paxos_strategy = LegacyPaxosStrategy.migration;
+
     public volatile int max_top_size_partition_count = 10;
     public volatile int max_top_tombstone_partition_count = 10;
     public volatile DataStorageSpec.LongBytesBound min_tracked_partition_size = new DataStorageSpec.LongBytesBound("1MiB");
@@ -1147,6 +1147,29 @@ public class Config
         disabled,
         warn,
         exception
+    }
+
+    /*
+     * How to pick a consensus protocol for CAS
+     * and serial read operations. Transaction statements
+     * will always run on Accord. Legacy in this context includes PaxosV2.
+     */
+    public enum LegacyPaxosStrategy
+    {
+        /*
+         * Allow both Accord and PaxosV1/V2 to run on the same cluster
+         * Some keys and ranges might be running on Accord if they
+         * have been migrated and the rest will run on Paxos until
+         * they are migrated.
+         */
+        migration,
+
+        /*
+         * Everything will be run on Accord. Useful for new deployments
+         * that don't want to accidentally start using legacy Paxos
+         * requiring migration to Accord.
+         */
+        accord
     }
 
     private static final Set<String> SENSITIVE_KEYS = new HashSet<String>() {{
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -43,7 +43,6 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.regex.Pattern;
-
 import javax.annotation.Nullable;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -2836,6 +2835,11 @@ public class DatabaseDescriptor
     public static boolean paxoTopologyRepairStrictEachQuorum()
     {
         return conf.paxos_topology_repair_strict_each_quorum;
+    }
+
+    public static Config.LegacyPaxosStrategy getLegacyPaxosStrategy()
+    {
+        return conf.legacy_paxos_strategy;
     }
 
     public static void setNativeTransportMaxRequestDataInFlightPerIpInBytes(long maxRequestDataInFlightInBytes)
--- a/src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
+++ b/src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
@@ -17,24 +17,74 @@
  */
 package org.apache.cassandra.cql3.conditions;
 
+import java.io.IOException;
 import java.nio.ByteBuffer;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.SortedSet;
+import java.util.function.Supplier;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 
+import com.google.common.base.Suppliers;
 import com.google.common.collect.Iterators;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
 
-import org.apache.cassandra.cql3.*;
+import org.apache.cassandra.cql3.AbstractMarker;
+import org.apache.cassandra.cql3.ColumnSpecification;
+import org.apache.cassandra.cql3.Constants;
+import org.apache.cassandra.cql3.FieldIdentifier;
+import org.apache.cassandra.cql3.Lists;
+import org.apache.cassandra.cql3.Maps;
+import org.apache.cassandra.cql3.Operator;
+import org.apache.cassandra.cql3.QueryOptions;
+import org.apache.cassandra.cql3.Sets;
+import org.apache.cassandra.cql3.Term;
 import org.apache.cassandra.cql3.Term.Terminal;
+import org.apache.cassandra.cql3.Terms;
+import org.apache.cassandra.cql3.UserTypes;
+import org.apache.cassandra.cql3.VariableSpecifications;
 import org.apache.cassandra.cql3.functions.Function;
-import org.apache.cassandra.db.rows.*;
-import org.apache.cassandra.db.marshal.*;
+import org.apache.cassandra.db.marshal.AbstractType;
+import org.apache.cassandra.db.marshal.ByteBufferAccessor;
+import org.apache.cassandra.db.marshal.CollectionType;
+import org.apache.cassandra.db.marshal.CounterColumnType;
+import org.apache.cassandra.db.marshal.ListType;
+import org.apache.cassandra.db.marshal.MapType;
+import org.apache.cassandra.db.marshal.SetType;
+import org.apache.cassandra.db.marshal.UserType;
+import org.apache.cassandra.db.rows.Cell;
+import org.apache.cassandra.db.rows.CellPath;
+import org.apache.cassandra.db.rows.ComplexColumnData;
+import org.apache.cassandra.db.rows.Row;
+import org.apache.cassandra.io.IVersionedSerializer;
+import org.apache.cassandra.io.util.DataInputPlus;
+import org.apache.cassandra.io.util.DataOutputPlus;
 import org.apache.cassandra.schema.ColumnMetadata;
 import org.apache.cassandra.schema.TableMetadata;
 import org.apache.cassandra.transport.ProtocolVersion;
 import org.apache.cassandra.utils.ByteBufferUtil;
-import org.apache.commons.lang3.builder.ToStringBuilder;
-import org.apache.commons.lang3.builder.ToStringStyle;
 
-import static org.apache.cassandra.cql3.statements.RequestValidations.*;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.primitives.Ints.checkedCast;
+import static java.util.stream.Collectors.toList;
+import static org.apache.cassandra.cql3.Constants.UNSET_VALUE;
+import static org.apache.cassandra.cql3.statements.RequestValidations.checkFalse;
+import static org.apache.cassandra.cql3.statements.RequestValidations.invalidRequest;
+import static org.apache.cassandra.db.TypeSizes.sizeofUnsignedVInt;
+import static org.apache.cassandra.service.accord.AccordSerializers.columnMetadataSerializer;
+import static org.apache.cassandra.service.accord.AccordSerializers.deserializeCqlCollectionAsTerm;
+import static org.apache.cassandra.utils.ByteBufferUtil.UNSET_BYTE_BUFFER;
+import static org.apache.cassandra.utils.ByteBufferUtil.byteBufferSerializer;
+import static org.apache.cassandra.utils.ByteBufferUtil.nullableByteBufferSerializer;
+import static org.apache.cassandra.utils.CollectionSerializers.deserializeList;
+import static org.apache.cassandra.utils.CollectionSerializers.readCollectionSize;
+import static org.apache.cassandra.utils.CollectionSerializers.serializeList;
+import static org.apache.cassandra.utils.CollectionSerializers.serializedListSize;
 
 /**
  * A CQL3 condition on the value of a column or collection element.  For example, "UPDATE .. IF a = 0".
@@ -73,7 +123,7 @@ public abstract class ColumnCondition
         terms.collectMarkerSpecification(boundNames);
     }
 
-    public abstract ColumnCondition.Bound bind(QueryOptions options);
+    public abstract Bound bind(QueryOptions options);
 
     protected final List<ByteBuffer> bindAndGetTerms(QueryOptions options)
     {
@@ -107,7 +157,7 @@ public abstract class ColumnCondition
         {
             T value = values.get(i);
             // The value can be ByteBuffer or Constants.Value so we need to check the 2 type of UNSET
-            if (value != ByteBufferUtil.UNSET_BYTE_BUFFER && value != Constants.UNSET_VALUE)
+            if (value != UNSET_BYTE_BUFFER && value != UNSET_VALUE)
                 filtered.add(value);
         }
         return filtered;
@@ -129,7 +179,7 @@ public abstract class ColumnCondition
                 return new MultiCellCollectionBound(column, operator, bindTerms(options));
 
             if (column.type.isUDT() && column.type.isMultiCell())
-                return new MultiCellUdtBound(column, operator, bindAndGetTerms(options), options.getProtocolVersion());
+                return new MultiCellUdtBound(column, operator, bindAndGetTerms(options));
 
             return new SimpleBound(column, operator, bindAndGetTerms(options));
         }
@@ -210,13 +260,57 @@ public abstract class ColumnCondition
         return new UDTFieldCondition(column, udtField, op, terms);
     }
 
+    enum BoundKind
+    {
+        ELEMENT_ACCESS(0, ElementAccessBound.serializer),
+        MULTI_CELL_COLLECTION(1, MultiCellCollectionBound.serializer),
+        MULTI_CELL_UDT(2, MultiCellUdtBound.serializer),
+        SIMPLE(3, SimpleBound.serializer),
+        UDT_FIELD_ACCESS(4, UDTFieldAccessBound.serializer);
+
+        final int id;
+
+        final BoundSerializer serializer;
+
+        BoundKind(int id, BoundSerializer serializer)
+        {
+            this.id = id;
+            this.serializer = serializer;
+        }
+
+        @SuppressWarnings("rawtypes")
+        static BoundSerializer serializer(int id)
+        {
+            switch(id)
+            {
+                case 0:
+                    return ElementAccessBound.serializer;
+                case 1:
+                    return MultiCellCollectionBound.serializer;
+                case 2:
+                    return MultiCellUdtBound.serializer;
+                case 3:
+                     return SimpleBound.serializer;
+                case 4:
+                    return UDTFieldAccessBound.serializer;
+                default:
+                    throw new AssertionError("Shouldn't have an enum with no serializer");
+            }
+        }
+    }
+
     public static abstract class Bound
     {
+        @Nonnull
         public final ColumnMetadata column;
+
+        @Nonnull
         public final Operator comparisonOperator;
 
         protected Bound(ColumnMetadata column, Operator operator)
         {
+            checkNotNull(column);
+            checkNotNull(operator);
             this.column = column;
             // If the operator is an IN we want to compare the value using an EQ.
             this.comparisonOperator = operator.isIN() ? Operator.EQ : operator;
@@ -235,7 +329,7 @@ public abstract class ColumnCondition
         /** Returns true if the operator is satisfied (i.e. "otherValue operator value == true"), false otherwise. */
         public static boolean compareWithOperator(Operator operator, AbstractType<?> type, ByteBuffer value, ByteBuffer otherValue)
         {
-            if (value == ByteBufferUtil.UNSET_BYTE_BUFFER)
+            if (value == UNSET_BYTE_BUFFER)
                 throw invalidRequest("Invalid 'unset' value in condition");
 
             if (value == null)
@@ -257,6 +351,55 @@ public abstract class ColumnCondition
             }
             return operator.isSatisfiedBy(type, otherValue, value);
         }
+
+        void checkForUnsetValues(List<ByteBuffer> values)
+        {
+            for (ByteBuffer buffer : values)
+                if (buffer == UNSET_BYTE_BUFFER)
+                    throw invalidRequest("Invalid 'unset' value in condition");
+        }
+        protected abstract BoundKind kind();
+
+        public static final IVersionedSerializer<Bound> serializer = new IVersionedSerializer<Bound>()
+        {
+            @Override
+            @SuppressWarnings("unchecked")
+            public void serialize(Bound bound, DataOutputPlus out, int version) throws IOException
+            {
+                columnMetadataSerializer.serialize(bound.column, out, version);
+                bound.comparisonOperator.writeToUnsignedVInt(out);
+                BoundKind kind = bound.kind();
+                out.writeUnsignedVInt(kind.ordinal());
+                kind.serializer.serialize(bound, out, version);
+            }
+
+            @Override
+            public Bound deserialize(DataInputPlus in, int version) throws IOException
+            {
+                ColumnMetadata column = columnMetadataSerializer.deserialize(in, version);
+                Operator comparisonOperator = Operator.readFromUnsignedVInt(in);
+                int boundKind = checkedCast(in.readUnsignedVInt());
+                return BoundKind.serializer(boundKind).deserialize(in, version, column, comparisonOperator);
+            }
+
+            @Override
+            @SuppressWarnings("unchecked")
+            public long serializedSize(Bound bound, int version)
+            {
+                BoundKind kind = bound.kind();
+                return columnMetadataSerializer.serializedSize(bound.column, version)
+                       + bound.comparisonOperator.sizeAsUnsignedVInt()
+                       + sizeofUnsignedVInt(kind.ordinal())
+                       + kind.serializer.serializedSize(bound, version);
+            }
+        };
+    }
+
+    private interface BoundSerializer<T extends Bound>
+    {
+        void serialize(T bound, DataOutputPlus out, int version) throws IOException;
+        Bound deserialize(DataInputPlus in, int version, ColumnMetadata column, Operator operator) throws IOException;
+        long serializedSize(T condition, int version);
     }
 
     protected static final Cell<?> getCell(Row row, ColumnMetadata column)
@@ -317,6 +460,7 @@ public abstract class ColumnCondition
         private SimpleBound(ColumnMetadata column, Operator operator, List<ByteBuffer> values)
         {
             super(column, operator);
+            checkForUnsetValues(values);
             this.values = values;
         }
 
@@ -341,6 +485,34 @@ public abstract class ColumnCondition
             }
             return false;
         }
+
+        @Override
+        protected BoundKind kind()
+        {
+            return BoundKind.SIMPLE;
+        }
+
+        private static final BoundSerializer<SimpleBound> serializer = new BoundSerializer<SimpleBound>()
+        {
+            @Override
+            public void serialize(SimpleBound bound, DataOutputPlus out, int version) throws IOException
+            {
+                serializeList(bound.values, out, version, nullableByteBufferSerializer);
+            }
+
+            @Override
+            public SimpleBound deserialize(DataInputPlus in, int version, ColumnMetadata column, Operator operator) throws IOException
+            {
+                List<ByteBuffer> values = deserializeList(in, version, nullableByteBufferSerializer);
+                return new SimpleBound(column, operator, values);
+            }
+
+            @Override
+            public long serializedSize(SimpleBound bound, int version)
+            {
+                return serializedListSize(bound.values, version, nullableByteBufferSerializer);
+            }
+        };
     }
 
     /**
@@ -351,11 +523,13 @@ public abstract class ColumnCondition
         /**
          * The collection element
          */
+        @Nonnull
         private final ByteBuffer collectionElement;
 
         /**
          * The conditions values.
          */
+        @Nonnull
         private final List<ByteBuffer> values;
 
         public ElementAccessBound(ColumnMetadata column,
@@ -364,6 +538,11 @@ public abstract class ColumnCondition
                                    List<ByteBuffer> values)
         {
             super(column, operator);
+            checkForUnsetValues(values);
+
+            if (collectionElement == null)
+                throw invalidRequest("Invalid null value for %s element access",
+                                     column.type instanceof MapType ? "map" : "list");
 
             this.collectionElement = collectionElement;
             this.values = values;
@@ -373,9 +552,6 @@ public abstract class ColumnCondition
         public boolean appliesTo(Row row)
         {
             boolean isMap = column.type instanceof MapType;
-
-            if (collectionElement == null)
-                throw invalidRequest("Invalid null value for %s element access", isMap ? "map" : "list");
 
             if (isMap)
             {
@@ -445,6 +621,36 @@ public abstract class ColumnCondition
             checkFalse(idx < 0, "Invalid negative list index %d", idx);
             return idx;
         }
+
+        @Override
+        protected BoundKind kind()
+        {
+            return BoundKind.ELEMENT_ACCESS;
+        }
+
+        private static final BoundSerializer<ElementAccessBound> serializer = new BoundSerializer<ElementAccessBound>()
+        {
+            @Override
+            public void serialize(ElementAccessBound bound, DataOutputPlus out, int version) throws IOException
+            {
+                byteBufferSerializer.serialize(bound.collectionElement, out, version);
+                serializeList(bound.values, out, version, nullableByteBufferSerializer);
+            }
+
+            @Override
+            public ElementAccessBound deserialize(DataInputPlus in, int version, ColumnMetadata column, Operator operator) throws IOException
+            {
+                ByteBuffer collectionElement = byteBufferSerializer.deserialize(in, version);
+                List<ByteBuffer> values = deserializeList(in, version, nullableByteBufferSerializer);
+                return new ElementAccessBound(column, collectionElement, operator, values);
+            }
+
+            @Override
+            public long serializedSize(ElementAccessBound bound , int version)
+            {
+                return byteBufferSerializer.serializedSize(bound.collectionElement, version) + serializedListSize(bound.values, version, nullableByteBufferSerializer);
+            }
+        };
     }
 
     /**
@@ -452,13 +658,34 @@ public abstract class ColumnCondition
      */
     public static final class MultiCellCollectionBound extends Bound
     {
-        private final List<Term.Terminal> values;
+        @Nonnull
+        private final List<Terminal> values;
 
-        public MultiCellCollectionBound(ColumnMetadata column, Operator operator, List<Term.Terminal> values)
+        /**
+         * Avoid recomputing the serialized terminals for serialize and serializedSize
+         */
+        @Nullable
+        private Supplier<List<ByteBuffer>> serializedValues;
+
+        public MultiCellCollectionBound(ColumnMetadata column, Operator operator, List<Terminal> values)
         {
             super(column, operator);
+            checkNotNull(values);
             assert column.type.isMultiCell();
             this.values = values;
+        }
+
+        public Supplier<List<ByteBuffer>> serializedValues()
+        {
+            if (serializedValues != null)
+                return serializedValues;
+
+            serializedValues = Suppliers.memoize(() ->
+                                                 values.stream()
+                                                       .map(v -> v == null ? null : v.get(ProtocolVersion.CURRENT))
+                                                       .collect(toList()));
+
+            return serializedValues;
         }
 
         public boolean appliesTo(Row row)
@@ -466,12 +693,12 @@ public abstract class ColumnCondition
             return appliesTo(column, comparisonOperator, values, row);
         }
 
-        public static boolean appliesTo(ColumnMetadata column, Operator operator, List<Term.Terminal> values, Row row)
+        public static boolean appliesTo(ColumnMetadata column, Operator operator, List<Terminal> values, Row row)
         {
             CollectionType<?> type = (CollectionType<?>) column.type;
 
             // copy iterator contents so that we can properly reuse them for each comparison with an IN value
-            for (Term.Terminal value : values)
+            for (Terminal value : values)
             {
                 Iterator<Cell<?>> iter = getCells(row, column);
                 if (value == null)
@@ -495,7 +722,7 @@ public abstract class ColumnCondition
             return false;
         }
 
-        private static boolean valueAppliesTo(CollectionType<?> type, Iterator<Cell<?>> iter, Term.Terminal value, Operator operator)
+        private static boolean valueAppliesTo(CollectionType<?> type, Iterator<Cell<?>> iter, Terminal value, Operator operator)
         {
             if (value == null)
                 return !iter.hasNext();
@@ -507,10 +734,10 @@ public abstract class ColumnCondition
             {
                 case LIST:
                     List<ByteBuffer> valueList = ((Lists.Value) value).elements;
-                    return listAppliesTo((ListType<?>)type, iter, valueList, operator);
+                    return setOrListAppliesTo(((ListType<?>)type).getElementsType(), iter, valueList.iterator(), operator, false);
                 case SET:
-                    Set<ByteBuffer> valueSet = ((Sets.Value) value).elements;
-                    return setAppliesTo((SetType<?>)type, iter, valueSet, operator);
+                    SortedSet<ByteBuffer> valueSet = ((Sets.Value) value).elements;
+                    return setOrListAppliesTo(((SetType<?>)type).getElementsType(), iter, valueSet.iterator(), operator, true);
                 case MAP:
                     Map<ByteBuffer, ByteBuffer> valueMap = ((Maps.Value) value).map;
                     return mapAppliesTo((MapType<?, ?>)type, iter, valueMap, operator);
@@ -527,7 +754,10 @@ public abstract class ColumnCondition
 
                 // for lists we use the cell value; for sets we use the cell name
                 ByteBuffer cellValue = isSet ? iter.next().path().get(0) : iter.next().buffer();
-                int comparison = type.compare(cellValue, conditionIter.next());
+                ByteBuffer conditionValue = conditionIter.next();
+                if (conditionValue == null)
+                    conditionValue = ByteBufferUtil.EMPTY_BYTE_BUFFER;
+                int comparison = type.compare(cellValue, conditionValue);
                 if (comparison != 0)
                     return evaluateComparisonWithOperator(comparison, operator);
             }
@@ -537,18 +767,6 @@ public abstract class ColumnCondition
 
             // they're equal
             return operator == Operator.EQ || operator == Operator.LTE || operator == Operator.GTE;
-        }
-
-        private static boolean listAppliesTo(ListType<?> type, Iterator<Cell<?>> iter, List<ByteBuffer> elements, Operator operator)
-        {
-            return setOrListAppliesTo(type.getElementsType(), iter, elements.iterator(), operator, false);
-        }
-
-        private static boolean setAppliesTo(SetType<?> type, Iterator<Cell<?>> iter, Set<ByteBuffer> elements, Operator operator)
-        {
-            ArrayList<ByteBuffer> sortedElements = new ArrayList<>(elements);
-            Collections.sort(sortedElements, type.getElementsType());
-            return setOrListAppliesTo(type.getElementsType(), iter, sortedElements.iterator(), operator, true);
         }
 
         private static boolean mapAppliesTo(MapType<?, ?> type, Iterator<Cell<?>> iter, Map<ByteBuffer, ByteBuffer> elements, Operator operator)
@@ -562,13 +780,17 @@ public abstract class ColumnCondition
                 Map.Entry<ByteBuffer, ByteBuffer> conditionEntry = conditionIter.next();
                 Cell<?> c = iter.next();
 
+                ByteBuffer conditionEntryKey = conditionEntry.getKey();
+
                 // compare the keys
-                int comparison = type.getKeysType().compare(c.path().get(0), conditionEntry.getKey());
+                int comparison = type.getKeysType().compare(c.path().get(0), conditionEntryKey);
                 if (comparison != 0)
                     return evaluateComparisonWithOperator(comparison, operator);
 
+                ByteBuffer conditionEntryValue = conditionEntry.getValue();
+
                 // compare the values
-                comparison = type.getValuesType().compare(c.buffer(), conditionEntry.getValue());
+                comparison = type.getValuesType().compare(c.buffer(), conditionEntryValue);
                 if (comparison != 0)
                     return evaluateComparisonWithOperator(comparison, operator);
             }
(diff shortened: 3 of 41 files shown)


test/unit/org/apache/cassandra/cql3/conditions/ColumnConditionTest.java
```diff
@@ -611,5 +882,96 @@ public void testMapCollectionBoundIsSatisfiedByValue() throws InvalidRequestExce
...
+    {
+        // EQ
+        assertTrue(conditionUDTApplies(list(ONE), EQ, list(ONE)));
+        assertFalse(conditionUDTApplies(list(ONE), EQ, list(ZERO)));
+        assertFalse(conditionUDTApplies(list(ZERO), EQ, list(ONE)));
+        assertFalse(conditionUDTApplies(list(ONE, ONE), EQ, list(ONE)));
+        assertFalse(conditionUDTApplies(list(ONE), EQ, list(ONE, ONE)));
+        assertFalse(conditionUDTApplies(list(ONE), EQ, list()));
+
+        assertFalse(conditionUDTApplies(list(ONE), EQ, (List<ByteBuffer>)null));
```
maedhroz:
```suggestion
        assertFalse(conditionUDTApplies(list(ONE), EQ, null));
```
A few more redundant casts on `null` below, but will just mention it here. On the other hand, it is kind of a nice bit of "documentation" :)

test/unit/org/apache/cassandra/cql3/conditions/ColumnConditionTest.java
```diff
@@ -17,162 +17,421 @@
...
     {
-        Row.Builder builder = BTreeRow.sortedBuilder();
-        builder.newRow(Clustering.EMPTY);
-        long now = System.currentTimeMillis();
-        if (values != null)
+        AbstractType<?> type = definition.type;
+        if (type.isFrozenCollection())
+        {
+            ByteBuffer cellValue = ListSerializer.pack(values, values.size(), CURRENT);
+            Cell cell = new BufferCell(definition, 0L, Cell.NO_TTL, Cell.NO_DELETION_TIME, cellValue, null);
```
maedhroz: nit: 
```suggestion
            Cell<ByteBuffer> cell = new BufferCell(definition, 0L, Cell.NO_TTL, Cell.NO_DELETION_TIME, cellValue, null);
```
Could even use `<?>` if you just want the most compact way to avoid the warning.

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1149,6 +1149,31 @@ public static void setOverrideLoadConfig(Supplier<Config> loadConfig)
...
+     * How to pick a consensus protocol for CAS
+     * and serial read operations. Transaction statements
+     * will always run on Accord. Legacy in this context includes PaxosV2.
+     */
+    public enum LegacyPaxosStrategy
+    {
+        /*
+         * Default setting
+         *
+         * Allow both Accord and PaxoV1/V2 to run on the same cluster
```
maedhroz:
```suggestion
         * Allow both Accord and PaxosV1/V2 to run on the same cluster
```

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1149,6 +1149,31 @@ public static void setOverrideLoadConfig(Supplier<Config> loadConfig)
...
 
+    /*
+     * How to pick a consensus protocol for CAS
+     * and serial read operations. Transaction statements
+     * will always run on Accord. Legacy in this context includes PaxosV2.
+     */
+    public enum LegacyPaxosStrategy
+    {
+        /*
+         * Default setting
```
maedhroz: nit: I might not mention that this is the default right here. Could be a bit annoying to remember to change the doc later if the actual default changes.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2838,6 +2837,16 @@ public static boolean paxoTopologyRepairStrictEachQuorum()
         return conf.paxos_topology_repair_strict_each_quorum;
     }
 
+    public static Config.LegacyPaxosStrategy getLegacyPaxosStrategy()
+    {
+        return conf.legacy_paxos_strategy;
+    }
+
+    public static void setLegacyPaxosStrategy(Config.LegacyPaxosStrategy strategy)
```
maedhroz: Do we need to be able to set this at runtime? (Noticed it's `volatile`...)

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
```diff
@@ -19,42 +19,28 @@
...
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.SortedSet;
-import java.util.TreeSet;
+import java.util.*;
```
maedhroz: nit: Restore single imports?

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
```diff
@@ -19,42 +19,28 @@
...
 import accord.primitives.Txn;
+import javax.annotation.Nullable;
 import org.apache.cassandra.audit.AuditLogContext;
 import org.apache.cassandra.audit.AuditLogEntryType;
-import org.apache.cassandra.cql3.CQLStatement;
-import org.apache.cassandra.cql3.ColumnSpecification;
-import org.apache.cassandra.cql3.QueryOptions;
-import org.apache.cassandra.cql3.ResultSet;
-import org.apache.cassandra.cql3.VariableSpecifications;
+import org.apache.cassandra.cql3.*;
```
maedhroz: nit: Restore single imports?

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
```diff
@@ -64,22 +50,12 @@
...
-import org.apache.cassandra.service.accord.txn.TxnCondition;
-import org.apache.cassandra.service.accord.txn.TxnData;
-import org.apache.cassandra.service.accord.txn.TxnDataName;
-import org.apache.cassandra.service.accord.txn.TxnNamedRead;
-import org.apache.cassandra.service.accord.txn.TxnQuery;
-import org.apache.cassandra.service.accord.txn.TxnRead;
-import org.apache.cassandra.service.accord.txn.TxnReference;
-import org.apache.cassandra.service.accord.txn.TxnUpdate;
-import org.apache.cassandra.service.accord.txn.TxnWrite;
+import org.apache.cassandra.service.accord.txn.*;
```
maedhroz: nit: Restore single imports?

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -110,7 +109,14 @@ public static long nowInMicros()
         return TimeUnit.MILLISECONDS.toMicros(Clock.Global.currentTimeMillis());
     }
 
-    public TxnData coordinate(Txn txn, QueryOptions options)
+    /**
+     * Consistency level is just echoed back in timeouts, in the future it may be used for interoperability
+     * with non-Accord operations.
+     * @param txn
+     * @param consistencyLevel
+     * @return
```
maedhroz: nit: I'd just remove the `@param` and `@return` bits.

test/distributed/org/apache/cassandra/distributed/test/accord/AccordTestBase.java
```diff
@@ -183,6 +196,27 @@ public static void install(ClassLoader classLoader, Integer num)
...
+                           .make()
+                           .load(cl, ClassLoadingStrategy.Default.INJECTION);
+        }
+
+        public static TxnData coordinate(Txn txn, @SuperCall Callable<TxnData> actual) throws Exception
+        {
+            count.incrementAndGet();
+            return actual.call();
+        }
+    }
```
maedhroz, line 272: I have as one of the TODO items in [CASSANDRA-18107](https://issues.apache.org/jira/browse/CASSANDRA-18107) latency metrics recording in `AccordService#coordinate()`, probably as a "Transaction" item in `ClientRequestsMetricsHolder`. Maybe we could just throw that in here, avoid BB, and kill two birds w/ one stone?

test/distributed/org/apache/cassandra/distributed/test/accord/AccordCQLTest.java
```diff
@@ -31,10 +31,16 @@
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import org.apache.cassandra.cql3.CQLTester;
 
 import accord.primitives.Unseekables;
 import accord.topology.Topologies;
 import org.apache.cassandra.cql3.functions.types.utils.Bytes;
+import org.apache.cassandra.db.marshal.*;
```
maedhroz: nit: unused?

test/distributed/org/apache/cassandra/distributed/test/accord/AccordCQLTest.java
```diff
@@ -50,16 +56,17 @@
...
-import org.apache.cassandra.cql3.CQLTester;
-import org.apache.cassandra.distributed.api.ConsistencyLevel;
-import org.apache.cassandra.distributed.api.SimpleQueryResult;
-import org.apache.cassandra.service.accord.AccordService;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
```
maedhroz: nit: unused block?

test/distributed/org/apache/cassandra/distributed/test/accord/AccordCQLTest.java
```diff
@@ -2305,4 +2312,30 @@ public void testReferenceArithmeticInUpdate() throws Exception
...
+                assertRowEqualsWithPreemptedRetry(cluster, new Object[]{ false, 5 }, "UPDATE " + currentTable + " SET v = 5 WHERE id = 1 AND c = 2 IF s = 4");
+                assertRowSerial(cluster, "SELECT id, c, v, s FROM " + currentTable + " WHERE id = 1 AND c = 2", 1, 2, 4, 5);
+                assertRowEqualsWithPreemptedRetry(cluster, new Object[]{true}, "UPDATE " + currentTable + " SET v = 5 WHERE id = 1 AND c = 2 IF s = 5");
+                assertRowSerial(cluster, "SELECT id, c, v, s FROM " + currentTable + " WHERE id = 1 AND c = 2", 1, 2, 5, 5);
+                assertRowEqualsWithPreemptedRetry(cluster, new Object[]{true}, "UPDATE " + currentTable + " SET s = 6 WHERE id = 1 IF s = 5");
+                assertRowSerial(cluster, "SELECT id, c, v, s FROM " + currentTable + " WHERE id = 1 AND c = 2", 1, 2, 5, 6);
+                // Make sure all the consensus using queries actually were run on Accord
+                assertEquals( 11, getAccordCoordinateCount() - startingAccordCoordinateCount);
+        });
+    }
```
maedhroz, line 2388: Should we parameterize all the `CASTestBase` tests, `CASAddTest`, `CasWriteTest`, etc. on `legacy_paxos_strategy` to make sure we get enough coverage?

src/java/org/apache/cassandra/utils/ByteBufferUtil.java
```diff
@@ -23,37 +23,28 @@
  * afterward, and ensure the tests still pass.
  */
 
-import java.io.*;
```
maedhroz, line 26: nit: fix import order and wildcards?

src/java/org/apache/cassandra/utils/ByteBufferUtil.java
```diff
@@ -359,6 +350,17 @@ public static void writeWithVIntLength(ByteBuffer bytes, DataOutputPlus out) thr
         out.writeUnsignedVInt(bytes.remaining());
         out.write(bytes);
     }
+    public static void writeWithVIntLengthAndNull(ByteBuffer bytes, DataOutputPlus out) throws IOException
```
maedhroz, line 363: unused?

src/java/org/apache/cassandra/utils/ByteBufferUtil.java
```diff
@@ -389,6 +391,17 @@ public static ByteBuffer readWithVIntLength(DataInputPlus in) throws IOException
         return ByteBufferUtil.read(in, length);
     }
 
+    public static ByteBuffer readWithVIntLengthAndNull(DataInputPlus in) throws IOException
```
maedhroz: unused?

src/java/org/apache/cassandra/utils/ByteBufferUtil.java
```diff
@@ -401,6 +414,15 @@ public static int serializedSizeWithVIntLength(ByteBuffer buffer)
         return TypeSizes.sizeofUnsignedVInt(size) + size;
     }
 
+    public static int serializedSizeWithVIntLengthAndNull(ByteBuffer buffer)
```
maedhroz: unused?

src/java/org/apache/cassandra/db/marshal/ValueAccessor.java
```diff
@@ -130,6 +135,8 @@ default int sizeWithShortLength(V value)
      */
     default boolean isEmpty(V value)
     {
+        if (value == null)
+            System.out.println("oops");
```
maedhroz: nit: Maybe just an `assert` w/ a reasonable message?

src/java/org/apache/cassandra/db/ClusteringPrefix.java
```diff
@@ -38,6 +38,8 @@
 import org.apache.cassandra.utils.bytecomparable.ByteComparable.Version;
 import org.apache.cassandra.utils.bytecomparable.ByteSource;
 
+import static com.google.common.base.Preconditions.checkNotNull;
```
maedhroz: unused?

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -135,39 +60,29 @@
...
 
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
-import static java.util.concurrent.TimeUnit.NANOSECONDS;
+import java.nio.ByteBuffer;
+import java.util.*;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.Function;
+import java.util.stream.Collectors;
```
maedhroz: TODO: revert the import order changes and wildcards

src/java/org/apache/cassandra/service/CASRequest.java
```diff
@@ -44,4 +47,8 @@
      * are passed as argument.
      */
     public PartitionUpdate makeUpdates(FilteredPartition current, ClientState clientState, Ballot ballot) throws InvalidRequestException;
+
+    public Txn toAccordTxn(ClientState clientState, int nowInSecs);
```
maedhroz: nit: `public` is redundant here, and I'd probably remove it from all these methods

src/java/org/apache/cassandra/service/accord/txn/TxnRead.java
```diff
@@ -51,6 +54,9 @@
 
 public class TxnRead extends AbstractKeySorted<TxnNamedRead> implements Read
 {
+    // There is only potentially one partition in a CAS and SERIAL/LOCAL_SERIAL read
+    public static final String DUMMY_NAME = "";
+    public static final TxnDataName DUMMY = TxnDataName.user(DUMMY_NAME);
```
maedhroz: nit/naming: What about something more specific like "SERIAL_READ" (that's also not just an empty string)?

src/java/org/apache/cassandra/service/accord/serializers/PreacceptSerializers.java
```diff
@@ -35,9 +31,10 @@
 import org.apache.cassandra.io.util.DataOutputPlus;
 import org.apache.cassandra.service.accord.serializers.TxnRequestSerializer.WithUnsyncedSerializer;
 
-import static org.apache.cassandra.utils.NullableSerializer.deserializeNullable;
-import static org.apache.cassandra.utils.NullableSerializer.serializeNullable;
-import static org.apache.cassandra.utils.NullableSerializer.serializedSizeNullable;
+import javax.annotation.Nullable;
+import java.io.IOException;
```
maedhroz: TODO: move these back up ^

src/java/org/apache/cassandra/cql3/statements/CQL3CasRequest.java
```diff
@@ -17,32 +17,40 @@
...
+import org.apache.cassandra.service.accord.txn.*;
 import org.apache.cassandra.service.paxos.Ballot;
 import org.apache.cassandra.utils.Pair;
 import org.apache.cassandra.utils.TimeUUID;
-
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
 
+import java.nio.ByteBuffer;
+import java.util.*;
```
maedhroz: TODO: restore proper import order

src/java/org/apache/cassandra/cql3/statements/CQL3CasRequest.java
```diff
@@ -399,11 +425,72 @@ public boolean appliesTo(FilteredPartition current) throws InvalidRequestExcepti
...
     }
     
     @Override
     public String toString()
     {
         return ToStringBuilder.reflectionToString(this, ToStringStyle.SHORT_PREFIX_STYLE);
     }
+
+    @Override
+    public Txn toAccordTxn(ClientState clientState, int nowInSecs) {
```
maedhroz: nit: newline before `{`

src/java/org/apache/cassandra/cql3/statements/CQL3CasRequest.java
```diff
@@ -399,11 +425,72 @@ public boolean appliesTo(FilteredPartition current) throws InvalidRequestExcepti
...
+        return fragments;
+    }
+
+    @Override
+    public RowIterator toCasResult(TxnData txnData)
+    {
+        FilteredPartition partition = txnData.get(TxnRead.DUMMY);
+        if (partition != null)
+            return partition.rowIterator();
+        return null;
```
maedhroz: Alternate if you're feeling one-linery...

```
return partition != null ? partition.rowIterator() : null;
```

src/java/org/apache/cassandra/cql3/statements/CQL3CasRequest.java
```diff
@@ -399,11 +425,72 @@ public boolean appliesTo(FilteredPartition current) throws InvalidRequestExcepti
...
+    {
+        List<TxnCondition> txnConditions = new ArrayList<>();
+        if (staticConditions != null)
+        {
+            txnConditions.add(staticConditions.asTxnCondition());
+        }
+        conditions.values()
+                  .stream()
+                  .map(RowCondition::asTxnCondition)
+                  .forEach(txnConditions::add);
```
maedhroz: nit: Having been burned many times by the object creation hidden here, I can only suggest...

```
for (RowCondition condition : conditions.values())
    txnConditions.add(condition.asTxnCondition());
```
...although in this case, it feels slightly more readable as well.

src/java/org/apache/cassandra/service/accord/txn/TxnData.java
```diff
@@ -75,6 +75,7 @@ public Set<Map.Entry<TxnDataName, FilteredPartition>> entrySet()
         return data.entrySet();
     }
 
+    // This is inadequate
```
aweisberg: I need to remove this. Accord does single replica reads so no need to merge duplicates.

For live migration Cassandra will probably coordinate quorum reads for Accord so the merging will already be done.

src/java/org/apache/cassandra/service/accord/txn/TxnCondition.java
```diff
@@ -509,7 +581,6 @@ public long serializedSize(BooleanGroup condition, int version)
 
     public static final IVersionedSerializer<TxnCondition> serializer = new IVersionedSerializer<TxnCondition>()
     {
-        @SuppressWarnings("unchecked")
```
maedhroz: nit: I still get a warning here without the suppression

src/java/org/apache/cassandra/service/accord/txn/TxnDataName.java
```diff
@@ -221,6 +243,7 @@ public long serializedSize(TxnDataName t, int version)
             int size = Byte.BYTES + sizeofUnsignedVInt(t.parts.length);
             for (String part : t.parts)
                 size += TypeSizes.sizeof(part);
+            size += serializedNullableSize(t.clustering, version, clusteringPrefixSerializer);
```
maedhroz: Aside: One of the items I have in the Alpha -> v1 TODO list is changing the internal representation of `parts` to not force converting everything to `String`

src/java/org/apache/cassandra/service/accord/txn/TxnQuery.java
```diff
@@ -49,17 +54,47 @@ public Result compute(TxnId txnId, Data data, @Nullable Read read, @Nullable Upd
...
+    public static final TxnQuery CONDITION = new TxnQuery()
+    {
+        @Override
+        protected byte type()
+        {
+            return 3;
+        }
+
+        @Override
+        public Result compute(TxnId txnId, Data data, @Nullable Read read, @Nullable Update update)
```
maedhroz: `update` isn't nullable for CAS, so what happens on 82 is fine?

src/java/org/apache/cassandra/service/accord/txn/TxnRead.java
```diff
@@ -67,6 +73,12 @@ public TxnRead(List<TxnNamedRead> items, Keys txnKeys)
         this.txnKeys = txnKeys;
     }
 
+    public static TxnRead createRead(SinglePartitionReadCommand readCommand)
```
maedhroz: nit/naming/non-binding: I'd be fine w/ `create()` too, given it's a static factory

src/java/org/apache/cassandra/service/accord/txn/TxnReference.java
```diff
@@ -18,29 +18,21 @@
...
 import org.apache.cassandra.io.IVersionedSerializer;
 import org.apache.cassandra.io.util.DataInputPlus;
 import org.apache.cassandra.io.util.DataOutputPlus;
 import org.apache.cassandra.schema.ColumnMetadata;
 import org.apache.cassandra.transport.ProtocolVersion;
 import org.apache.cassandra.utils.ByteBufferUtil;
 
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Objects;
```
maedhroz: TODO: restore import order/revert wildcards

src/java/org/apache/cassandra/service/accord/txn/TxnUpdate.java
```diff
@@ -47,14 +38,15 @@
 import org.apache.cassandra.utils.ByteBufferUtil;
 import org.apache.cassandra.utils.ObjectSizes;
 
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.*;
+import java.util.function.Function;
```
maedhroz: TODO: restore import order

src/java/org/apache/cassandra/service/accord/txn/TxnUpdate.java
```diff
@@ -185,6 +178,7 @@ public Write apply(Data data)
         return new TxnWrite(updates);
     }
 
+    // Should we serialize the conditionResult?
```
maedhroz: If we serialize the result, does that also remove the need for the `checkCondition()` method below? (i.e. The condition should already have been applied somewhere and populated...and `null` would mean something went wrong?) If so, +1

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -49,6 +42,12 @@
 import org.apache.cassandra.utils.concurrent.Future;
 import org.apache.cassandra.utils.concurrent.UncheckedInterruptedException;
 
+import java.util.Arrays;
+import java.util.Random;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
```
maedhroz: TODO: restore import order

src/java/org/apache/cassandra/service/accord/AccordSerializers.java
```diff
@@ -171,4 +177,73 @@ public long serializedSize(TableMetadata metadata, int version)
             return TableId.serializedSize();
         }
     };
-}
+
+    public static final IVersionedSerializer<Clustering<?>> clusteringPrefixSerializer = new IVersionedSerializer<Clustering<?>>()
```
maedhroz: nit/naming: `clusteringSerializer` ?

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
```diff
@@ -17,24 +17,44 @@
...
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
 
-import static org.apache.cassandra.cql3.statements.RequestValidations.*;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.*;
+import java.util.function.Supplier;
```
maedhroz: TODO: restore proper import order

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
```diff
@@ -257,6 +315,55 @@ else if (otherValue == null)
             }
             return operator.isSatisfiedBy(type, otherValue, value);
         }
+
+        void checkForUnsetValues(List<ByteBuffer> values)
```
maedhroz, line 355: With this in place, do we still need the per-value checks in `compareWithOperator()`? I think the only thing we're not currently double-validating is the usage in `TxnCondition#applies()`.

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
```diff
@@ -210,13 +230,51 @@ public static ColumnCondition condition(ColumnMetadata column, FieldIdentifier u
...
+                    return MultiCellUdtBound.serializer;
+                case SIMPLE:
+                     return SimpleBound.serializer;
+                case UDT_FIELD_ACCESS:
+                    return UDTFieldAccessBound.serializer;
+                default:
+                    throw new AssertionError("Shouldn't have an enum with no serializer");
+            }
+        }
+    }
```
maedhroz, line 300: Might be simpler to do something like...

```
enum BoundKind
{
    ELEMENT_ACCESS(ElementAccessBound.serializer),
    MULTI_CELL_COLLECTION(MultiCellCollectionBound.serializer),
    MULTI_CELL_UDT(MultiCellUdtBound.serializer),
    SIMPLE(SimpleBound.serializer),
    UDT_FIELD_ACCESS(UDTFieldAccessBound.serializer);
    
    private final BoundSerializer<Bound> serializer;
    
    @SuppressWarnings("unchecked")
    BoundKind(BoundSerializer<? extends Bound> serializer)
    {
        this.serializer = (BoundSerializer<Bound>) serializer;
    }

    BoundSerializer<Bound> serializer()
    {
        return serializer;
    }
}
```

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
```diff
@@ -210,13 +230,51 @@ public static ColumnCondition condition(ColumnMetadata column, FieldIdentifier u
         return new UDTFieldCondition(column, udtField, op, terms);
     }
 
+    /*
+     * Don't change or remove entries (and ordinals) or it will break cross version compatibility
+     * as the ordinals are sent in messages and stored by Accord for in flight transactions
+     */
```
maedhroz: If it's really important, why not just give each kind an explicit identifier, sort of like what we have in `cql3.Operator`?

```
static BoundKind fromID(int id)
{
    switch(id)
    {
        case 1: return ELEMENT_ACCESS;
        case 2: return MULTI_CELL_COLLECTION;
        case 3: return MULTI_CELL_UDT;
        case 4: return SIMPLE;
        case 5: return UDT_FIELD_ACCESS;
        default: throw new AssertionError("Identifier " + id + " does not match a kind!");
    }
}
```

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
```diff
@@ -364,6 +502,11 @@ public ElementAccessBound(ColumnMetadata column,
                                    List<ByteBuffer> values)
         {
             super(column, operator);
+            checkForUnsetValues(values);
+
+            boolean isMap = column.type instanceof MapType;
```
maedhroz: nit: might be worth inlining so we don't have to do anything if the element isn't `null`

src/java/org/apache/cassandra/utils/ByteBufferUtil.java
```diff
@@ -927,6 +949,19 @@ private static boolean startsWith(ByteBuffer src, ByteBuffer prefix, int offset)
...
+            serializer.serialize(value, dob, version);
+            return dob.buffer();
+        }
+        catch (IOException e)
+        {
+            throw new RuntimeException(e);
+        }
+    }
+
     public static final IVersionedSerializer<ByteBuffer> vintSerializer = new IVersionedSerializer<ByteBuffer>()
```
maedhroz: naming: Maybe something like `bytesSerializer`, given the vint part is an implementation detail (i.e. how we write the length)?

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
```diff
@@ -546,6 +737,7 @@ private static boolean listAppliesTo(ListType<?> type, Iterator<Cell<?>> iter, L
 
         private static boolean setAppliesTo(SetType<?> type, Iterator<Cell<?>> iter, Set<ByteBuffer> elements, Operator operator)
         {
+            // This is redundant since it is already a sorted set?
```
maedhroz: It certainly looks that way. `Value` always stores a `SortedSet` using the element type as a comparator. Probably worth streamlining.

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
```diff
@@ -445,33 +585,81 @@ private static int getListIndex(ByteBuffer collectionElement)
...
      * A condition on an entire collection column.
      */
     public static final class MultiCellCollectionBound extends Bound
     {
-        private final List<Term.Terminal> values;
+        @Nonnull
+        private final List<Terminal> values;
+
+        @Nullable
+        private Supplier<List<ByteBuffer>> serializedValues;
```
maedhroz, line 668: The idea here is that we avoid calculating the serialized form in both `serializedSize()` and `serialize()`?

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
```diff
@@ -579,6 +775,43 @@ private static boolean mapAppliesTo(MapType<?, ?> type, Iterator<Cell<?>> iter,
...
+            }
+
+            @Override
+            public MultiCellCollectionBound deserialize(DataInputPlus in, int version, ColumnMetadata column, Operator operator) throws IOException
+            {
+                List<ByteBuffer> values = deserializeList(in, version, vintNullableSerializer);
+                List<Terminal> terminals;
+                if (operator.isContains() || operator.isContainsKey())
+                {
+                    terminals = values.stream().map(Constants.Value::new).collect(toList());
```
maedhroz: nit/performance: Might be less garbage if this is just..

```
List<Terminal> terminals = new ArrayList<>(values.size());
...
for (ByteBuffer value : values)
    terminals.add(new Constants.Value(value));
```

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
```diff
@@ -647,6 +885,7 @@ private ByteBuffer rowValue(Row row)
         {
             UserType userType = (UserType) column.type;
 
+
```
maedhroz: nit: extra newline

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
```diff
@@ -730,6 +1005,37 @@ public String toString()
...
+                serializeList(bound.values, out, version, vintNullableSerializer);
+                out.writeUnsignedVInt(bound.protocolVersion.asInt());
+            }
+
+            @Override
+            public MultiCellUdtBound deserialize(DataInputPlus in, int version, ColumnMetadata column, Operator operator) throws IOException
+            {
+                List<ByteBuffer> values = deserializeList(in, version, vintNullableSerializer);
+                int protocolVersion = checkedCast(in.readUnsignedVInt());
+                // Does decode actually do what we want?
```
maedhroz, line 1047: It might not, given what it looks like I did in the `TxnCondition.Value` serializer (which itself was unnecessary, given where [CASSANDRA-18114](https://issues.apache.org/jira/browse/CASSANDRA-18114) is going).

test/unit/org/apache/cassandra/cql3/conditions/ColumnConditionTest.java
```diff
@@ -17,162 +17,421 @@
...
+    private boolean conditionUDTApplies(List<ByteBuffer> rowValue, Operator op, List<ByteBuffer> conditionValue)
+    {
+        ColumnMetadata definition = ColumnMetadata.regularColumn(KEYSPACE, maybeCreateTable(udtType, false), "c", udtType);
+        Term term = conditionValue == null ? Constants.NULL_VALUE : new UserTypes.Value(udtType, conditionValue.toArray(new ByteBuffer[0]));
+        ColumnCondition condition = ColumnCondition.condition(definition, op, Terms.of(term));
+        ColumnCondition.Bound bound = condition.bind(QueryOptions.DEFAULT);
+        return testRoundtripped(bound, newRow(definition, rowValue));
+    }
+
+    public void beforeTest() throws Throwable
```
maedhroz, line 413: nit: `@Override`

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
```diff
@@ -687,16 +957,21 @@ public static final class MultiCellUdtBound extends Bound
...
 
         public MultiCellUdtBound(ColumnMetadata column, Operator op, List<ByteBuffer> values, ProtocolVersion protocolVersion)
         {
             super(column, op);
+            checkNotNull(values);
+            checkNotNull(protocolVersion);
+            checkForUnsetValues(values);
             assert column.type.isMultiCell();
             this.values = values;
             this.protocolVersion = protocolVersion;
```
maedhroz: nit: `rowValue()` doesn't need to be `final` on 986...already `private`

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
```diff
@@ -622,16 +855,21 @@ private static final class UDTFieldAccessBound extends Bound
...
+        @Nonnull
         private final List<ByteBuffer> values;
 
         private UDTFieldAccessBound(ColumnMetadata column, FieldIdentifier field, Operator operator, List<ByteBuffer> values)
         {
             super(column, operator);
+            checkNotNull(field);
+            checkNotNull(values);
+            checkForUnsetValues(values);
             assert column.type.isUDT() && field != null;
```
maedhroz: nit: `field` can't be `null` here anymore

test/unit/org/apache/cassandra/cql3/conditions/ColumnConditionTest.java
aweisberg: I think there were cases where the cast makes explicit which function to call that would otherwise be ambiguous. Some were definitely not needed and I removed those.

test/unit/org/apache/cassandra/cql3/conditions/ColumnConditionTest.java
aweisberg: Added ByteBuffer to the three instances of this

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: Removed the extra null check on field

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: Removed it

test/unit/org/apache/cassandra/cql3/conditions/ColumnConditionTest.java
aweisberg: Added

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: Protocol version is null for user type. I could remove this and pass null and it would continue to work and might be slightly clearer.

aweisberg: Removed it

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: Removed

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: Now you make me wonder if streams can magically see that it could also preallocate. Or it the entire stream you think is on the slow side?

aweisberg: If you really want to think performance... the array list is just a list of objects and if it were mutable we could just replace each element with what we want :-P

The mistake is actually that deserialize list is not a stream (and streams are not efficient in their own right).

aweisberg: And really the List is itself a wasted allocation and indirection for the array and now I am just sad about Java.

aweisberg: Anyways this is what I did
```
                int numTerminals = readCollectionSize(in, version);
                List<Terminal> terminals = new ArrayList<>(numTerminals);
                if (operator.isContains() || operator.isContainsKey())
                {
                    for (int i = 0; i < numTerminals; i++)
                        terminals.add(new Constants.Value(vintNullableSerializer.deserialize(in, version)));
                }
                else
                {
                    for (int i = 0; i < numTerminals; i++)
                        terminals.add(deserializeCqlCollectionAsTerm(vintNullableSerializer.deserialize(in, version), column.type, ProtocolVersion.CURRENT));
                }
```

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: Yes, I think that was the reason. I will add a comment.

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: I removed the extra sort, I also removed the method containing the sort, and also removed the same wrapper method for list which didn't do much. LMK if it looks good.

src/java/org/apache/cassandra/utils/ByteBufferUtil.java
aweisberg: Renamed to `byteBufferSerializer` and `nullableByteBufferSerializer`

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: Inlined.

Optimizer probably already does the inlining, but who knows.

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: I changed things up a bit and now it uses explicit IDs.

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: Yeah changed to this as part of the previous thing. I didn't do this because it requires expressing the relationship twice. Once for ID and once for when you already have an enum value.

The ordinal gives you that for free, but then it's unsafe.

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: It's probably redundant, but that is such a fast thing to check at runtime I think it's fine to do it in both?

I added this check and moved it earlier because I was unsure if we had those unset values in practice and in which types of bounds. It's a little easier to find out the unset value is unset when the bound is constructed rather than later when it is applied.

Don't feel strongly either way.

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: Looks like it is fixed after touching it again.

src/java/org/apache/cassandra/service/accord/AccordSerializers.java
aweisberg: Renamed.

src/java/org/apache/cassandra/service/accord/AccordService.java
aweisberg: Touched the file

src/java/org/apache/cassandra/service/accord/AccordService.java
aweisberg: IDEA keeps auto-adding that and I keep needing to remove it. Will update my config to stop doing that. Removed.

Also why not remove consistency level as well?

src/java/org/apache/cassandra/service/accord/txn/TxnUpdate.java
aweisberg: The conclusion here was that we don't need to. At the coordinator it will not be recomputed in memory because it is memoized and on recovery it is safe for a recovery coordinator to recompute it.

src/java/org/apache/cassandra/service/accord/txn/TxnUpdate.java
aweisberg: Updated the imports

src/java/org/apache/cassandra/service/accord/txn/TxnReference.java
aweisberg: We don't have a threshold for when to use wildcards. Depending on the setting it will either remove wildcard imports or add wildcard imports.

I don't see an option to avoid fighting on that particular issue, even something like "don't change untouched imports"

src/java/org/apache/cassandra/service/accord/txn/TxnRead.java
aweisberg: Renamed to `SERIAL_READ_NAME`, and `SERIAL_READ` and the value is now `SERIAL_READ`

src/java/org/apache/cassandra/service/accord/txn/TxnRead.java
aweisberg: How about `createSerialRead` because this specifically creates one for that purpose?

src/java/org/apache/cassandra/service/accord/txn/TxnQuery.java
aweisberg: Huh, yeah copy paste error from the others. Accord says that `Data` is also nullable in the interface.

I changed it and added null preconditions.

src/java/org/apache/cassandra/service/accord/txn/TxnCondition.java
aweisberg: Huh, not sure why I touched that line. Added it back.

src/java/org/apache/cassandra/cql3/statements/CQL3CasRequest.java
aweisberg: Got it. Dial back the streams. 😁

aweisberg: Changed it and also sized the array list.

src/java/org/apache/cassandra/cql3/statements/CQL3CasRequest.java
aweisberg: Camel is my spirit animal. Changed.
image

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
aweisberg: Ran optimize imports with the right style (hopefully for real this time!)

src/java/org/apache/cassandra/cql3/statements/CQL3CasRequest.java
aweisberg: Added newline

src/java/org/apache/cassandra/cql3/statements/CQL3CasRequest.java
aweisberg: Optimized imports

src/java/org/apache/cassandra/service/accord/serializers/PreacceptSerializers.java
aweisberg: Optimized imports

src/java/org/apache/cassandra/service/CASRequest.java
aweisberg: Removed them

src/java/org/apache/cassandra/service/StorageProxy.java
aweisberg: Optimized imports

src/java/org/apache/cassandra/db/ClusteringPrefix.java
aweisberg: Removed

src/java/org/apache/cassandra/db/marshal/ValueAccessor.java
aweisberg: Oh that isn't supposed to be there at all... it's for settings breakpoint. Removed.

src/java/org/apache/cassandra/utils/ByteBufferUtil.java
aweisberg: Optimized imports

test/distributed/org/apache/cassandra/distributed/test/accord/AccordCQLTest.java
aweisberg: I was leaning on ColumnConditionTest to check that we could actually roundtrip the already tested implementation of CAS and CQL semantics and then end to end test with AccordCQLTest we could actually read/write data running through Accord.

CASTestBase requires functionality I don't think we have yet in accord like adding and removing nodes. CASWriteTest has tests that are pretty specifically testing the Paxos implementation.

I think running with simulator is more interesting, but it was not quite working when I went to run it although I do have a branch with it https://github.com/aweisberg/cassandra/tree/cassandra-18100-simulator

There are kind of two things to test in my mind. Does any CQL syntax/semantics break with CAS on Accord, and does actually running requests on Accord work. I think the latter is pretty well covered and simulator is the real proof, but also if Accord can pass simulator for other transaction types it should apply to CAS transactions.

The former is kind of unit tested by ColumnConditionTest in that if ColumnCondition correctly implements the CQL and we can roundtrip it then it should also work for CAS. A lot of the code is existing code.

The hole there is I am not sure we test CAS syntax methodically anywhere?

test/distributed/org/apache/cassandra/distributed/test/accord/AccordCQLTest.java
aweisberg: I am not even sure how that can happen with my settings. Optimized the imports.

test/distributed/org/apache/cassandra/distributed/test/accord/AccordTestBase.java
aweisberg: Seems reasonable. Do we want separate metrics for transactions that are write only since they generally have a very different latency profile?

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
aweisberg: Optimized imports

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
aweisberg: It's not a hot prop so no we don't. I'll remove the volatile.

aweisberg: Should I also remove the setter? It is unused.

src/java/org/apache/cassandra/config/Config.java
aweisberg: Removed mentioning the default.

src/java/org/apache/cassandra/config/Config.java
aweisberg: Changed

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
maedhroz: Might as well.

test/distributed/org/apache/cassandra/distributed/test/accord/AccordTestBase.java
maedhroz: > Do we want separate metrics for transactions that are write only

I'm fine w/ having separate read-only, write-only, and "mixed" versions. It's nicer to have more granularity when we don't need it than less when we do :p

test/distributed/org/apache/cassandra/distributed/test/accord/AccordCQLTest.java
maedhroz: > CASTestBase requires functionality I don't think we have yet in accord like adding and removing nodes. CASWriteTest has tests that are pretty specifically testing the Paxos implementation.

Good points.

> The former is kind of unit tested by ColumnConditionTest in that if ColumnCondition correctly implements the CQL and we can roundtrip it then it should also work for CAS.

True. If the serialization is solid, there's not much else to go wrong in `ColumnConditionsAdapter`.

> I think running with simulator is more interesting, but it was not quite working when I went to run it

Agreed. Getting simulator working is probably the best use of further testing energy at this point.

In any case, feel free to mark this as resolved.

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
maedhroz: > It's a little easier to find out the unset value is unset when the bound is constructed rather than later when it is applied.

Would it make sense to have the `value == UNSET_BYTE_BUFFER` check in the `TxnCondition.Value` constructor? We'd find out earlier I guess.

In terms of whether we'd want to remove the duplicate checks in `compareWithOperator()`, I'll leave it up to you. If I'm able to make it private again when I finish the CQL integration, perhaps I'd remove them, but I'm fine either way at this point.

maedhroz (approved):
LGTM

src/java/org/apache/cassandra/cql3/conditions/ColumnCondition.java
aweisberg: For `CQL3CasRequest` this would be later not earlier? The bounds are created when the request is first created from the parser/planner bits.

test/distributed/org/apache/cassandra/distributed/test/accord/AccordTestBase.java
aweisberg: This looks more complicated then just latency since it's counting a bunch of different things. Abort, timeout, unavailable, mutation size. I think we need to think a little harder about metrics for Accord transactions.

maedhroz: Agreed, although we already handle similar concerns in `CASClientWriteRequestMetrics` -> `CASClientRequestMetrics` -> `ClientRequestMetrics`, where we distinguish between those things.

maedhroz: I'm fine if you want to defer to [CASSANDRA-18107](https://issues.apache.org/jira/browse/CASSANDRA-18107)

aweisberg:
Merged as [a3ccd4d93d1e1ed3ca65e2c3ad3e0126a2c8bf8f](https://github.com/apache/cassandra/commit/56d7e6914399f35ccdc47cb4cfc4891bbbbc81f6)

PR #2050, opened 2022-12-13
CASSANDRA-18113 fqltool dump NPE fix.

nvharikrishna:
If fullquerylog is enabled and a prepared query executed with null value binded, then fqltool dump is running into NPE. Checking value for null before wrapping it.

The [CASSANDRA-18113](https://issues.apache.org/jira/browse/CASSANDRA-18113)

With this change, "null" is printed when a value is null. Sample output:
```
Type: single-query
Query start time: 1670955522494
Protocol version: 5
Generated timestamp:-9223372036854775808
Generated nowInSeconds:1670955522
Query: INSERT INTO ks1.t1 (id, value) VALUES (?, ?)
Values: 
00000000 00 00 00 0A                                      ····             
null
-----
```

### DIFF
--- a/tools/fqltool/src/org/apache/cassandra/fqltool/commands/Dump.java
+++ b/tools/fqltool/src/org/apache/cassandra/fqltool/commands/Dump.java
@@ -26,6 +26,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
+import com.google.common.annotations.VisibleForTesting;
+
 import io.airlift.airline.Arguments;
 import io.airlift.airline.Command;
 import io.airlift.airline.Option;
@@ -164,7 +166,8 @@ public class Dump implements Runnable
         }
     }
 
-    private static void dumpQuery(QueryOptions options, WireIn wireIn, StringBuilder sb)
+    @VisibleForTesting
+    static void dumpQuery(QueryOptions options, WireIn wireIn, StringBuilder sb)
     {
         sb.append("Query: ")
           .append(wireIn.read(FullQueryLogger.QUERY).text())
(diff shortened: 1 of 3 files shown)


tools/fqltool/src/org/apache/cassandra/fqltool/commands/Dump.java
```diff
@@ -219,12 +222,19 @@ private static void appendValuesToStringBuilder(List<ByteBuffer> values, StringB
         boolean first = true;
         for (ByteBuffer value : values)
         {
-            Bytes bytes = Bytes.wrapForRead(value);
-            long maxLength2 = Math.min(1024, bytes.readLimit() - bytes.readPosition());
-            toHexString(bytes, bytes.readPosition(), maxLength2, sb);
-            if (maxLength2 < bytes.readLimit() - bytes.readPosition())
+            if(null == value)
```
smiklosovic: `if (null == value)` - space after `if`.

nvharikrishna: done.

nvharikrishna:
On a different note, printing "------" is skipped for the first value: https://github.com/apache/cassandra/blob/trunk/tools/fqltool/src/org/apache/cassandra/fqltool/commands/Dump.java#L230-L237.

Not sure why it is done that way. But when there are multiple values, it is slightly difficult to differentiate the fist and second value. I can remove the "first" check too, if agreed.

smiklosovic:
@nvharikrishna can you please paste here the output so we clearly see what you mean by that "--------" comment?

I think we should preserve the output and we should not change it. Just fixing the null issue should be enough here.

nvharikrishna:
If I create a table something like this:

`CREATE TABLE t1 ( id int PRIMARY KEY , v1 int, v2 int, v3 int) ;`

and inserted a row using:

```
try (CqlSession cqlSession = CqlSession.builder().build()) {
    PreparedStatement preparedStatement = cqlSession.prepare("INSERT INTO ks1.t1 (id, v1, v2, v3) VALUES (?, ?, ?, ?)");
    cqlSession.execute(preparedStatement.bind(1, 1, 1, 1));
}
```

The output of fqltool looks something like this:

```
Type: single-query
Query start time: 1673373829119
Protocol version: 5
Generated timestamp:-9223372036854775808
Generated nowInSeconds:1673373829
Query: INSERT INTO ks1.t1 (id, v1, v2, v3) VALUES (?, ?, ?, ?)
Values: 
00000000 00 00 00 01                                      ····             
00000000 00 00 00 01                                      ····             
-----
00000000 00 00 00 01                                      ····             
-----
00000000 00 00 00 01                                      ····             
-----
```

`-----` is not printed between the first and second value
```
Values: 
00000000 00 00 00 01                                      ····             
00000000 00 00 00 01                                      ····             
```

smiklosovic:
@nvharikrishna thanks, interesting, that seems like a bug. Sorry it is taking so long to review this more closely. I should return to this soon.

smiklosovic:
I am looking at this. Thanks for the patience.

PR #2051, opened 2022-12-15
Cassandra 18118 3.11

bereng:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/db/Memtable.java
+++ b/src/java/org/apache/cassandra/db/Memtable.java
@@ -66,6 +66,7 @@ public class Memtable implements Comparable<Memtable>
     private static final Logger logger = LoggerFactory.getLogger(Memtable.class);
 
     public static final MemtablePool MEMORY_POOL = createMemtableAllocatorPool();
+    public static final long NO_MIN_TIMESTAMP = -1;
 
     private static MemtablePool createMemtableAllocatorPool()
     {
@@ -159,6 +160,16 @@ public class Memtable implements Comparable<Memtable>
         this.cfs = null;
         this.allocator = null;
         this.columnsCollector = new ColumnsCollector(metadata.partitionColumns());
+    }
+
+    @VisibleForTesting
+    public Memtable(CFMetaData metadata, long minTimestamp)
+    {
+        this.initialComparator = metadata.comparator;
+        this.cfs = null;
+        this.allocator = null;
+        this.columnsCollector = new ColumnsCollector(metadata.partitionColumns());
+        this.minTimestamp = minTimestamp;
     }
 
     public MemtableAllocator getAllocator()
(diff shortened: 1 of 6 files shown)


bereng:
CI [3.11](https://app.circleci.com/pipelines/github/bereng/cassandra?branch=CASSANDRA-18118-3.11)

src/java/org/apache/cassandra/db/Memtable.java
```diff
@@ -387,9 +397,16 @@ public Partition getPartition(DecoratedKey key)
...
+     * Returns the minTS if one available, otherwise -1.
+     *
+     * EncodingStats uses a synthetic epoch TS at 2015. We don't want to leak that (CASSANDRA-18118) so we return -1 instead.
+     *
+     * @return The minTS or -1 if none available
+     */
     public long getMinTimestamp()
     {
-        return minTimestamp;
+        return minTimestamp != EncodingStats.NO_STATS.minTimestamp ? minTimestamp : -1;
```
maedhroz: @bereng Looks like the general form of the solution is not to treat `EncodingStats.NO_STATS.minTimestamp` like a real timestamp in `CompactionController`. Any reason not to just leave `getMinTimestamp()` alone and check `EncodingStats.NO_STATS.minTimestamp` or `EncodingStats.TIMESTAMP_EPOCH` there instead?

bereng: Hi @maedhroz 

thx for reply. I am currently on xmas holidays so advanced apologies for delayed responses.

That is an option but ``Memtable.getMinTimestamp()` is a public method so any future new consumer could easily be leaking the 2015 TS again. I am thinking along the lines of leaving the 2015 TS as an encapsulated implementation detail.

If we found a better way to prevent future leaks that'd be perfect.

maedhroz: Right, I guess what I'm saying is that `-1` is also a bit arbitrary/a magic value. As long as callers have to check _something_, might as well check `EncodingStats.TIMESTAMP_EPOCH`?

bereng: Icwym but a TS in 2015, while debugging or if it got logged somewhere, reads to the eyes just like any other valid TS. A `-1` is clearly an 'out of band' value and it would have cut my debugging time by 10!. It would have immediately triggered me to look and identified the problem and I want to spare that pain to any future developers. Also, historically, `-1` has been used widely in returns values to signal similar things imo. It's a 'better' arbitrary value if such a thing exists imo.

Having said that, I also considered other alternatives such as a `checkTSisValid()` method just to raise awareness in the API about the 2015 TS but this being a hot path, I didn't want to add an extra method call.

Given my experience debugging this little devil and trying to prevent that pain to future devs, I can only think of the `-1` or even raise an exception to make it even more obvious. If you think about it, the nasty tricky bugs that can arise from this little TS may even grant an exception. Wdyt?

maedhroz: > while debugging or if it got logged somewhere, reads to the eyes just like any other valid TS

Yeah, makes sense. In that case, my only ask is that we pull out a constant to use across the new test, `getFullyExpiredSSTables()`, and `getMinTimestamp()`...maybe call it `NO_TIMESTAMP`. WDYT?

maedhroz (approved):
LGTM, modulo pulling out a const for the "no timestamp" timestamp ;)

bereng:
CI [3.11](https://app.circleci.com/pipelines/github/bereng/cassandra/832/workflows/9572df36-5bf8-4b15-a33c-7bb1bfeadcf4)

Please notice [this](https://github.com/apache/cassandra/pull/2051/files#diff-b4515f390c3b40e1f64feda8fd8746647b43cda26f8019ec96fc2dd4320fa96cR716) I somehow missed in my previous PR, pretty non-controversial imo.

bereng:
Squashed before commit.

PR #2056, opened 2022-12-16
CASSANDRA-17112: CEP-15: (C*) Strict Serializability verification

dcapwell:


build.xml
```diff
@@ -1063,7 +1063,7 @@
         <jvmarg value="-ea"/>
         <jvmarg value="-Djava.io.tmpdir=${tmp.dir}"/>
         <jvmarg value="-Dcassandra.debugrefcount=true"/>
-        <jvmarg value="-Xss256k"/>
+        <jvmarg value="-Xss384k"/>
```
dcapwell, line 1066: this is a weaker version of 

```
commit 4941b279bceb3a0fbe388be4305d25ec771f7b78
Author: David Capwell <dcapwell@apache.org>
Date:   Tue Dec 13 13:37:40 2022 -0800

    CEP-10: Simulator Java11 Support

    patch by David Capwell; reviewed by Benedict Elliott Smith, Ekaterina Dimitrova for CASSANDRA-17178
```

I needed as I test on M1 Mac

test/simulator/main/org/apache/cassandra/simulator/ActionList.java
```diff
@@ -41,6 +41,13 @@ public class ActionList extends AbstractCollection<Action>
     public static ActionList empty() { return EMPTY; }
     public static ActionList of(Action action) { return new ActionList(new Action[] { action }); }
     public static ActionList of(Stream<Action> action) { return new ActionList(action.toArray(Action[]::new)); }
+    public static ActionList of(Stream<Action> action, Stream<Action>... actions)
+    {
+        Stream<Action> accm = action;
+        for (Stream<Action> a : actions)
```
belliottsmith: probably cleaner (and more efficient) to use `Stream.of(actions).flatMap(a->a)`

test/simulator/main/org/apache/cassandra/simulator/paxos/AbstractPairOfSequencesPaxosSimulation.java
```diff
@@ -227,7 +149,7 @@ public Action get()
                     {
                         if (simulated.time.nanoTime() >= untilNanos)
                         {
-                            available.add(next);
+                            IntStream.of(partitions).mapToObj(Integer::valueOf).forEach(available::add);
```
belliottsmith: `.boxed()?`

build.xml
dcapwell: I can remove, or we can fix-up when we rebase against trunk

test/distributed/org/apache/cassandra/distributed/api/Row.java
```diff
@@ -170,26 +282,75 @@ public UUID getUUID(String name)
...
     {
         return get(name);
     }
 
+    public <T> Set<T> getSet(String name, Set<T> defaultValue)
+    {
+        return get(name, defaultValue);
+    }
+
+    // HERE
```
belliottsmith: ?

test/simulator/main/org/apache/cassandra/simulator/paxos/AbstractPairOfSequencesPaxosSimulation.java
```diff
@@ -124,101 +122,25 @@ public ActionPlan plan()
...
-                ActionList.of(
-                    cluster.stream().map(i -> simulated.run("Create Accord Epoch", i, () -> AccordService.instance().createEpochFromConfigUnsafe()))
-                )
-//            ).andThen(
-//                // TODO (now): this is temporary until we have parameterisation of simulation
-//                ActionList.of(
-//                    cluster.stream().map(i -> simulated.run("Disable Accord Cache", i, () -> AccordService.instance.setCacheSize(0)))
-//                )
+                cluster.stream().map(i -> simulated.run("Insert Partitions", i, executeForPrimaryKeys(preInsertStmt(), primaryKeys))),
+                cluster.stream().map(i -> simulated.run("Create Accord Epoch", i, () -> AccordService.instance().createEpochFromConfigUnsafe()))
```
belliottsmith, line 196: The TODO is still valid?

test/simulator/main/org/apache/cassandra/simulator/paxos/AbstractPairOfSequencesPaxosSimulation.java
```diff
@@ -124,101 +122,25 @@ public ActionPlan plan()
...
         List<Integer> available = IntStream.range(0, primaryKeys.length).boxed().collect(Collectors.toList());
         Action stream = Actions.infiniteStream(concurrency, new Supplier<Action>() {
             @Override
             public Action get()
             {
-                int i = simulated.random.uniform(0, available.size());
-                int next = available.get(i);
-                available.set(i, available.get(available.size() - 1));
-                available.remove(available.size() - 1);
+                int[] partitions = consume(simulated.random, available);
```
belliottsmith: `primaryKeys`? or `selectedPrimaryKeys`? To keep nomenclature consistent

test/simulator/main/org/apache/cassandra/simulator/paxos/HistoryChecker.java
```diff
@@ -127,16 +127,22 @@ Event setById(int id, Event event)
...
         int eventPosition = witnessSequence.length;
-        int eventId = eventPosition == 0 ? -1 : witnessSequence[eventPosition - 1];
+        int eventId = eventId(witnessSequence, eventPosition);
         setById(witness.id, new Event(witness.id)).log.add(new VerboseWitness(witness.id, start, end, witnessSequence));
         Event event = get(eventPosition, eventId);
         recordWitness(event, witness, witnessSequence);
         recordVisibleBy(event, end);
         recordVisibleUntil(event, start);
 
+
```
belliottsmith: ?

test/simulator/main/org/apache/cassandra/simulator/paxos/Observation.java
```diff
@@ -18,21 +18,38 @@
 
 package org.apache.cassandra.simulator.paxos;
 
+import org.apache.cassandra.distributed.api.SimpleQueryResult;
+
 class Observation implements Comparable<Observation>
 {
     final int id;
-    final Object[][] result;
+    private final SimpleQueryResult result;
```
belliottsmith: Either make them all private or none? It's package private, so doesn't seem necessary to encapsulate, but don't mind which you pick.

test/simulator/main/org/apache/cassandra/simulator/paxos/PairOfSequencesAccordSimulation.java
```diff
@@ -47,90 +56,34 @@
...
 import org.apache.cassandra.simulator.systems.SimulatedSystems;
 import org.apache.cassandra.simulator.utils.IntRange;
+import org.apache.cassandra.utils.concurrent.UncheckedInterruptedException;
 
-import static org.apache.cassandra.distributed.api.ConsistencyLevel.ANY;
 import static org.apache.cassandra.simulator.paxos.HistoryChecker.fail;
-import static org.apache.cassandra.utils.ByteBufferUtil.bytes;
+import static org.apache.cassandra.utils.AssertionUtils.hasCause;
+import static org.apache.cassandra.utils.AssertionUtils.hasCauseAnyOf;
+import static org.assertj.core.api.Assertions.anyOf;
```
belliottsmith: unused

test/simulator/main/org/apache/cassandra/simulator/paxos/HistoryValidator.java
```diff
@@ -0,0 +1,42 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.simulator.paxos;
+
+import javax.annotation.Nullable;
+
+public interface HistoryValidator
+{
+    Checker start(Observation observation);
```
belliottsmith: `start` is perhaps a confusing name when the operation has completed, and we refer to start/end times of an observation. `begin` seems less confusing, but perhaps `log` or `enter` or something else to distinguish it would be better.

I also don't think we should be saving these as registers inside the checkers themselves - if we want this API (and I'm fine with it) we should allocate a temporary object. Efficiency here isn't important, and it is much clearer and less prone to misuse.

test/simulator/main/org/apache/cassandra/simulator/paxos/HistoryValidator.java
```diff
@@ -0,0 +1,42 @@
...
+
+    void log(@Nullable Integer pk);
+
+    interface Checker extends AutoCloseable
+    {
+        void witnessRead(int pk, int count, int[] seq);
+        void witnessWrite(int pk);
```
belliottsmith: I'm not 100% keen on this, as it disguises the fact that the `id` is what we're logging here. I think perhaps(?) I would rather accept the `id` redundantly and verify it is the same, or some other change I haven't thought of yet.

test/simulator/main/org/apache/cassandra/simulator/paxos/StrictSerializabilityValidator.java
```diff
@@ -0,0 +1,105 @@
...
+    {
+        if (index.containsKey(pk))
+            return index.get(pk);
+        throw new IllegalArgumentException("Unknown pk=" + pk);
+    }
+
+    @Override
+    public void witnessRead(int pk, int count, int[] seq)
+    {
+        convertHistoryViolation(() -> verifier.witnessRead(get(pk), seq));
```
belliottsmith: I don't believe this call can throw a violation? Nor the write one.

test/simulator/main/org/apache/cassandra/simulator/paxos/PairOfSequencesAccordSimulation.java
```diff
@@ -281,21 +211,151 @@ protected String preInsertStmt()
...
+        return () -> execute(createAccordTxn(id, partitions, readOnly), "pk", "count", "seq");
     }
 
-    @Override
-    Operation modifying(int operationId, IInvokableInstance instance, int primaryKey, HistoryChecker historyChecker)
+    public class ReadWriteOperation extends Operation
     {
-        return new ModifyingOperation(operationId, instance, ANY, serialConsistency, primaryKey, historyChecker);
+        private final IntSet allPartitions;
+        private final IntSet readOnlySet;
```
belliottsmith: we should also have `writeOnly`?

belliottsmith (changes requested):
Looks good, mostly minor feedback, and I think the accord simulation coverage can be improved with a minor feature enhancement.

test/simulator/main/org/apache/cassandra/simulator/paxos/PairOfSequencesAccordSimulation.java
dcapwell: `allPartitions` contains all read/write partitions.  `readOnlySet` is the set that only did a read. Because we have access to the read accord did, all write operations have a read as well, so currently we do not get write only partitions.

I am not 100% sure how to do a write only txn as the validation logic tracks history; so if a txn doesn't load that history then its doing a blind write and could trigger a violation that wasn't a real violation.  My first thought would be that we periodically truncate the history, but do wonder if that would just make things more complex (if we have not observed all writes and did a truncate, may miss a violation)

test/distributed/org/apache/cassandra/distributed/api/Row.java
dcapwell: I honestly don't remember; removed

test/simulator/main/org/apache/cassandra/simulator/paxos/AbstractPairOfSequencesPaxosSimulation.java
dcapwell: This confused me when first working on it as its not the keys but the index of the keys.  The usage uses `primaryKeyIndex`, how about that?  

This is what the accord side does

```
return (simulated, primaryKeyIndex) -> {
  int[] primaryKeys = IntStream.of(primaryKeyIndex).map(i -> this.primaryKeys[i]).toArray();
...
```

test/simulator/main/org/apache/cassandra/simulator/paxos/HistoryChecker.java
dcapwell: the extra space?  I believe I made a change that I later removed, removing this extra space

test/simulator/main/org/apache/cassandra/simulator/paxos/Observation.java
dcapwell: removed `private` and removed the getter for it; usage accesses the field directly

test/simulator/main/org/apache/cassandra/simulator/paxos/PairOfSequencesAccordSimulation.java
dcapwell: when we rebase against trunk this will become a check style error... looking forward to that!

test/simulator/main/org/apache/cassandra/simulator/paxos/PairOfSequencesAccordSimulation.java
belliottsmith: Well, the list append can be truly blind (and should be), so we’re testing additional functionality. If it isn’t today, that’s something we should address (it is for Paxos, and we shouldn’t regress).

We shouldn’t also witness any reads performed internally, and if we do that’s also a bug - and for now at least we should filter them if it isn’t trivial to fix properly.

I’m not sure what you mean about false violations but the verifier supports blind writes just fine?

test/simulator/main/org/apache/cassandra/simulator/paxos/AbstractPairOfSequencesPaxosSimulation.java
belliottsmith: Sure, sounds good

test/simulator/main/org/apache/cassandra/simulator/paxos/HistoryValidator.java
dcapwell: > begin seems less confusing, but perhaps log or enter or something else to distinguish it would be better.

I like `log` so went with that.

> we should allocate a temporary object

done.

test/simulator/main/org/apache/cassandra/simulator/paxos/HistoryValidator.java
dcapwell: Not a fan as that is captured by `log(Observation)`, so would allow users to provide the wrong thing and make usage more verbose.

test/simulator/main/org/apache/cassandra/simulator/paxos/StrictSerializabilityValidator.java
dcapwell: it doesn't now, but figured it "may" in the future, so safer to guard against

test/simulator/main/org/apache/cassandra/simulator/paxos/HistoryValidator.java
belliottsmith: They couldn't provide the wrong thing if you validate it?

As it stands, it is unclear what it is we are logging as written, which wasn't the case before. It's implicit knowledge that it is the observation id we are writing, which is especially obtuse since the id exists whether or not we write. 

I'd be inclined then to just have `LinearizabilityValidator` implement an API similar to the one provided by the StrictSerializabilityVerifier, and introduce a common interface that we call directly, as it was IMO clearer what was being reported (but not prescriptive - if you have a better idea by all means)

dcapwell: > As it stands, it is unclear what it is we are logging as written, which wasn't the case before. It's implicit knowledge that it is the observation id we are writing, which is especially obtuse since the id exists whether or not we write.

I understand your point.  If we add the id then I question why `log` should see `Observation`?  Maybe the following is better?

```
public interface HistoryValidator {
  Checker log(int start, int end);
  void log(@Nullable Integer pk);

  interface Checker extends AutoCloseable {
    void witnessRead(int pk, int count, int[] sequence);
    void witnessWrite(int pk, int id);
  }
}
```

The main issue with this interface is that `LinearizabilityValidator` doesn't see `Observation` so would need a refactor in `HistoryChecker`; not a large one but would be needed (issue is in `org.apache.cassandra.simulator.paxos.HistoryChecker#witness`), we did speak about dropping `LinearizabilityValidator`, maybe this is better motivation?

I personally prefer the existing interface as you can't mess up and put event id in pk and pk in event id, but feel the above interface is a compromise?  Thoughts?

test/simulator/main/org/apache/cassandra/simulator/paxos/PairOfSequencesAccordSimulation.java
dcapwell: > Well, the list append can be truly blind (and should be), so we’re testing additional functionality. If it isn’t today, that’s something we should address (it is for Paxos, and we shouldn’t regress).

Makes sense, we can't do for `String` but *should* be able to do for `list`, so could leverage that.

> We shouldn’t also witness any reads performed internally, and if we do that’s also a bug 

Its an internal detail, normal CQL is unable to do this.  If you work at the `Txn` level you see the reads, if you work at the CQL level you only see what you asked for.

> I’m not sure what you mean about false violations but the verifier supports blind writes just fine?

its more if you are working with a type that requires read to update, then to avoid a read you need to ignore the history; simple example

```
create table foo (pk int PRIMARY KEY, seq text)

txn(INSERT INTO foo(pk, seq) VALUES (?, "lacking history...")
```

belliottsmith: Have you looked at how the Paxos one works? It uses `+=`, and can manage blind writes just fine for both registers. Obviously under the hood the implementation has to perform a read for the string based `+=`, so it is not a blind write in implementation. But at the CQL level it is a blind write, and at the implementation level these are treated slightly differently.

Saying that, it would probably also be good to randomise whether `seq1` as maintained/unmaintained so we can separately track the correctness of truly blind writes, as with Accord they are treated differently.

test/simulator/main/org/apache/cassandra/simulator/paxos/HistoryValidator.java
belliottsmith: yeah, that works for me. Though looking at it written there, the two different `log` methods probably aren't ideal. Perhaps rename the prior `log` method to `print` to match the underlying impls?

test/simulator/main/org/apache/cassandra/simulator/paxos/StrictSerializabilityValidator.java
belliottsmith: I think that is confusing, as it leads the reader to believe the method can throw - which is surprising if you understand the API involves building a collection of changes to apply together. Defensive programming is good, but I prefer it when it communicates the right thing to the reader.

test/simulator/main/org/apache/cassandra/simulator/paxos/HistoryValidator.java
dcapwell: pushed, but had to make slight changes

```
void witnessRead(int pk, int id, int count, int[] seq);
void witnessWrite(int pk, int id, boolean success);
```

id is needed in reader, and we need to know if the write was success or not.

test/simulator/main/org/apache/cassandra/simulator/paxos/StrictSerializabilityValidator.java
dcapwell: removed

test/simulator/main/org/apache/cassandra/simulator/paxos/PairOfSequencesAccordSimulation.java
dcapwell: > Have you looked at how the Paxos one works? It uses +=, and can manage blind writes just fine for both registers. Obviously under the hood the implementation has to perform a read for the string based +=, so it is not a blind write in implementation. But at the CQL level it is a blind write, and at the implementation level these are treated slightly differently.


Yep, looked closely at it before doing this patch.  A blind write there still performs a read, so I took advantage of the fact we have access to the reads.  If we want to remove this direct access, then I can have `writeOnly` as the current writes are in fact blind...

If we are cool keeping this access, I did confirm `org.apache.cassandra.cql3.Lists.Appender#doAppend` doesn't do a read, but we need to make sure not flatten else it looks to just override the celll (aka does *not* append, it *sets*)... given that I can switch/add list to add blind writes.

> Saying that, it would probably also be good to randomise whether seq1 as maintained/unmaintained so we can separately track the correctness of truly blind writes, as with Accord they are treated differently.

I don't know what paxos does, can you explain?  Given `seq1 text` I assume this is doing a `set` rather than `append`?  So would need to extend the validators to handle this truncate of history?  How does paxos handle this differently?

test/simulator/main/org/apache/cassandra/simulator/paxos/HistoryValidator.java
belliottsmith: Splitting hairs here, so feel free to ignore, but I think it might be confusing to a reader that the antonym of success in this context is unknown. Instead of accepting a `success` boolean parameter, it might be nice to have `witnessWriteSuccess` and `witnessWriteTimeout`. 

Also, another brief suggestion to accept/discard: perhaps the method to create the `Checker` should be `witness` and we can call it `Witness witness = validator.witness()` and invoke `witness.read(...)` and `witness.writeSuccess(...)`? `witness.writeTimeout(...)`?

test/simulator/main/org/apache/cassandra/simulator/paxos/PairOfSequencesAccordSimulation.java
belliottsmith: It does not do a read at the CQL layer, only internally. From the CQL perspective it is a blind write for Paxos.

We _shouldn't_ have access to the read in Accord if we perform a blind write either, as it shouldn't be returned to us since we didn't ask for it. We should be verifying this behaviour, even if it is only a modest divergence. Perhaps a small minority of writes should be run as blind at the CQL layer, but we should run at least some.

> I don't know what paxos does, can you explain? Given seq1 text I assume this is doing a set rather than append? So would need to extend the validators to handle this truncate of history? How does paxos handle this differently?

In this instance I am proposing additional functionality compared to the Paxos test, namely for some test runs we might want to randomly disable `seq1` entirely for the test duration, operating only on `seq2`, i.e. maintaining only seq2 and reading only seq2. This would permit us to run entirely on _true_ blind writes, to confirm they behave correctly, as they diverge more in implementation than with Paxos.

In the future, we might want to modify these tests not to rely on the reified history, by at least permitting history to be truncated. This is definitely possible with the strict serialisability verifier as it is designed, if we were to also request timestamps on read. But that's for another time.

test/simulator/main/org/apache/cassandra/simulator/paxos/HistoryValidator.java
dcapwell: pushed.  I like `witness`, I do feel its nicer.  I think `write` is still better in the current usage, but cool with default methods...

I prefer `checker.write(pk, outcome.id, outcome.isSuccess());` in the current use than

```
if (outcome.isSuccess()) checker.writeSuccess(pk, outcome.id);
else checker.writeUnknownFailure(pk, outcome.id);
```

I also prefer `UnknownFailure` than `Timeout` as it includes timeout + other issues

test/simulator/main/org/apache/cassandra/simulator/paxos/PairOfSequencesAccordSimulation.java
dcapwell: > We shouldn't have access to the read in Accord if we perform a blind write either, as it shouldn't be returned to us since we didn't ask for it. We should be verifying this behaviour, even if it is only a modest divergence.

I will update, I went this way as reading multiple rows is really annoying in the CQL layer atm...  If we had multi select statement support this feature would be nicer...

dcapwell: Created https://issues.apache.org/jira/browse/CASSANDRA-18154 to support moving to CQL rather than always bypass...  This will allow the following options: read-only, write-only, read-write... Ill work on this change now.

dcapwell: Moved to CQL which let me have : `writeOnly`, `readOnly`, `readWrite`

ide/idea/workspace.xml
```diff
@@ -167,7 +167,40 @@
       <option name="MAIN_CLASS_NAME" value="" />
       <option name="METHOD_NAME" value="" />
       <option name="TEST_OBJECT" value="class" />
-      <option name="VM_PARAMETERS" value="-Dcassandra.config=file://$PROJECT_DIR$/test/conf/cassandra.yaml -Dlogback.configurationFile=file://$PROJECT_DIR$/test/conf/logback-test.xml -Dcassandra.logdir=$PROJECT_DIR$/build/test/logs -Djava.library.path=$PROJECT_DIR$/lib/sigar-bin -Dlegacy-sstable-root=$PROJECT_DIR$/test/data/legacy-sstables -Dinvalid-legacy-sstable-root=$PROJECT_DIR$/test/data/invalid-legacy-sstables -Dcassandra.ring_delay_ms=1000 -Dcassandra.skip_sync=true -ea -XX:MaxMetaspaceSize=384M -XX:SoftRefLRUPolicyMSPerMB=0 -XX:HeapDumpPath=build/test -Dcassandra.strict.runtime.checks=true -Dlegacy-sstable-root=$PROJECT_DIR$/test/data/legacy-sstables -Dinvalid-legacy-sstable-root=$PROJECT_DIR$/test/data/invalid-legacy-sstables -Dmigration-sstable-root=$PROJECT_DIR$/test/data/migration-sstables -Dcassandra.ring_delay_ms=1000 -Dcassandra.tolerate_sstable_size=true -Dcassandra.skip_sync=true -Dcassandra.reads.thresholds.coordinator.defensive_checks_enabled=true" />
+      <option name="VM_PARAMETERS" value="
```
dcapwell, line 170: this causes every unit test to run with the configs for simulator... in testing a random sample of tests this looked fine...

src/java/org/apache/cassandra/utils/logging/ClassNameFilter.java
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.utils.logging;
+
+import ch.qos.logback.classic.spi.ILoggingEvent;
+import ch.qos.logback.core.filter.AbstractMatcherFilter;
+import ch.qos.logback.core.spi.FilterReply;
+
+public class ClassNameFilter extends AbstractMatcherFilter<ILoggingEvent>
```
dcapwell, line 25: this was added for the simulator log file, it takes the verifiers input and writes to a file

test/simulator/main/org/apache/cassandra/simulator/paxos/StrictSerializabilityValidator.java
```diff
@@ -0,0 +1,111 @@
...
+
+import accord.verify.StrictSerializabilityVerifier;
+import com.carrotsearch.hppc.IntIntHashMap;
+import com.carrotsearch.hppc.IntIntMap;
+import com.carrotsearch.hppc.cursors.IntIntCursor;
+
+public class StrictSerializabilityValidator implements HistoryValidator
+{
+    private final StrictSerializabilityVerifier verifier;
+    private final IntIntMap pkToIndex, indexToPk;
```
belliottsmith: not a big deal, but could just store the array for `indexToPk`?

test/simulator/main/org/apache/cassandra/simulator/paxos/PairOfSequencesAccordSimulation.java
```diff
@@ -281,21 +157,152 @@ protected String preInsertStmt()
...
+
+                        if (!seen.add(pk))
+                            throw new IllegalStateException("Duplicate partition key " + pk);
+                        // every partition was read, but not all were written to... need to verify each partition
+                        if (seq.length != count)
+                            throw fail(pk, "%d != #%s", count, seq);
+
+                        checker.read(pk, outcome.id, count, seq);
+
+                        if (writes.contains(pk))
```
belliottsmith: Can just leave this to the write loop?

belliottsmith:
minor feedback to consider, but +1

belliottsmith (approved):

test/simulator/main/org/apache/cassandra/simulator/paxos/StrictSerializabilityValidator.java
dcapwell: changed

test/simulator/main/org/apache/cassandra/simulator/paxos/PairOfSequencesAccordSimulation.java
dcapwell: I think this is a bug... this was true when every txn contained a read, but that is no longer true...

PR #2058, opened 2022-12-19
[CASSANDRA-16325] Update streaming metrics incrementally

isaacreath:
[CASSANDRA-16325] Update streaming metrics incrementally

Currently the inbound and outbound streamed bytes metrics are incremented after each file is streamed, what doesn't represent the current number of bytes streamed since it can take a long time for a large file to be streamed. This patch changes the `StreamSession` class to increment the total bytes streamed every time the `StreamSesson` makes progress instead of when a file finishes streaming. 

patch by Isaac Reath and Dejan Gvozdenac ; reviewed by Paulo Motta for CASSANDRA-16325

Co-authored-by: Isaac Reath
Co-authored-by: Dejan Gvozdenac


test/distributed/org/apache/cassandra/distributed/test/metrics/StreamingMetricsTest.java
```diff
@@ -145,6 +151,92 @@ public void testMetricsWithRepairAndStreamingToTwoNodes() throws Exception
...
+            cluster.get(1).flush(KEYSPACE);
+            cluster.get(2).flush(KEYSPACE);
+        }
+
+        drop1to3.off();
+
+        ExecutorService nodetoolExecutor = new ThreadPoolExecutor(1, 1, 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<>());
+        checkThatNoStreamingOccuredBetweenTheThreeNodes(cluster);
+        Future<Integer> streamingOperationExecution = nodetoolExecutor.submit(streamingOperation);
+        Thread.sleep(500); // Wait for some streaming to happen.
```
isaacreath: If we could find a way to get a notification instead of sleep that would be nice.

src/java/org/apache/cassandra/streaming/StreamSession.java
```diff
@@ -1027,9 +1022,30 @@ public void receive(IncomingStreamMessage message)
...
     {
         ProgressInfo progress = new ProgressInfo(peer, index, filename, direction, bytes, total);
+        updateMetricsOnProgress(progress);
         streamResult.handleProgress(progress);
     }
 
+    private void updateMetricsOnProgress(ProgressInfo progress)
+    {
+        ProgressInfo.Direction direction = progress.direction;
+        long lastSeenBytesStreamedForProgress = lastSeenBytesStreamed.getOrDefault(progress, 0L);
```
smiklosovic: does this mean that `lastSeenBytesStreamed` map will be ever growing only as I dont see any place where we are removing it from that map?

src/java/org/apache/cassandra/streaming/StreamSession.java
```diff
@@ -1027,9 +1022,30 @@ public void receive(IncomingStreamMessage message)
...
+        long newBytesStreamed = progress.currentBytes - lastSeenBytesStreamedForProgress;
+        if (direction == ProgressInfo.Direction.OUT)
+        {
+            StreamingMetrics.totalOutgoingBytes.inc(newBytesStreamed);
+            metrics.outgoingBytes.inc(newBytesStreamed);
+        }
+
```
smiklosovic: nit: empty line redundant

test/distributed/org/apache/cassandra/distributed/test/metrics/StreamingMetricsTest.java
```diff
@@ -145,10 +155,159 @@ public void testMetricsWithRepairAndStreamingToTwoNodes() throws Exception
         testMetricsWithStreamingToTwoNodes(true);
     }
 
-    private int getNumberOfSSTables(Cluster cluster, int node) {
+    @Test
+    public void testMetricsUpdateIncrementallyWithRepairAndStreamingBetweenNodes() throws Exception
+    {
+        try(Cluster cluster = init(Cluster.build(2)
```
smiklosovic: nit: `try (Cluster` - there should be space.

test/distributed/org/apache/cassandra/distributed/test/metrics/StreamingMetricsTest.java
```diff
@@ -145,10 +155,159 @@ public void testMetricsWithRepairAndStreamingToTwoNodes() throws Exception
...
+                                          .start(), 2))
+        {
+            runStreamingOperationAndCheckIncrementalMetrics(cluster, () -> cluster.get(2).nodetool("repair", "--full"));
+        }
+    }
+
+    @Test
+    public void testMetricsUpdateIncrementallyWithRebuildAndStreamingBetweenNodes() throws Exception
+    {
+        try(Cluster cluster = init(Cluster.build(2)
```
smiklosovic: nit: `try (Cluster` - there should be space.

test/distributed/org/apache/cassandra/distributed/test/metrics/StreamingMetricsTest.java
```diff
@@ -145,10 +155,159 @@ public void testMetricsWithRepairAndStreamingToTwoNodes() throws Exception
...
+    {
+        assertThat(cluster.size())
+            .describedAs("The minimum cluster size to check streaming metrics is 2 nodes.")
+            .isEqualTo(2);
+
+        // Create table with compression disabled so we can easily compute the expected final sstable size
+        cluster.schemaChange(String.format("CREATE TABLE %s.cf (k text PRIMARY KEY, c1 text) " +
+                                           "WITH compaction = {'class': '%s', 'enabled': 'false'} " +
+                                           "AND compression = {'enabled':'false'};",
+                                           KEYSPACE, "SizeTieredCompactionStrategy"));
```
smiklosovic: I fail to see why there is a placeholder for `class` when it is a constant. Why is it extracted?

test/distributed/org/apache/cassandra/distributed/test/metrics/StreamingMetricsTest.java
```diff
@@ -145,10 +155,159 @@ public void testMetricsWithRepairAndStreamingToTwoNodes() throws Exception
...
+
+        // Check that node 1 only has 1 sstable after flush + compaction
+        assertThat(getNumberOfSSTables(cluster, 1)).isEqualTo(1);
+        // Node 2 should have 0 sstables since messages from node1 were dropped
+        assertThat(getNumberOfSSTables(cluster, 2)).isEqualTo(0);
+
+        // Disable dropping of messages from node1 to node2
+        drop1to2.off();
+
+        ExecutorService nodetoolExecutor = new ThreadPoolExecutor(1, 1, 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<>());
```
smiklosovic: should not we shutdown this executor after we are done?

src/java/org/apache/cassandra/streaming/StreamSession.java
```diff
@@ -1027,9 +1022,30 @@ public void receive(IncomingStreamMessage message)
...
+        else if (direction == ProgressInfo.Direction.IN)
+        {
+            StreamingMetrics.totalIncomingBytes.inc(newBytesStreamed);
+            metrics.incomingBytes.inc(newBytesStreamed);
+        }
+
+        lastSeenBytesStreamed.put(progress, lastSeenBytesStreamedForProgress + newBytesStreamed);
```
smiklosovic: here we are just adding (presumably new) `progress`-es again and again. What does this mean in practice? If we are still adding the very same `progress` object, why does it have to be a map? Hence, I guess `progress` _is_ new every now and then. So the underlying map grows beyond any limit until streaming session is finished? What if we have thousands streaming sessions at once?

src/java/org/apache/cassandra/streaming/StreamSession.java
```diff
@@ -1027,9 +1022,30 @@ public void receive(IncomingStreamMessage message)
...
+        ProgressInfo.Direction direction = progress.direction;
+        long lastSeenBytesStreamedForProgress = lastSeenBytesStreamed.getOrDefault(progress, 0L);
+        long newBytesStreamed = progress.currentBytes - lastSeenBytesStreamedForProgress;
+        if (direction == ProgressInfo.Direction.OUT)
+        {
+            StreamingMetrics.totalOutgoingBytes.inc(newBytesStreamed);
+            metrics.outgoingBytes.inc(newBytesStreamed);
+        }
+
+        else if (direction == ProgressInfo.Direction.IN)
```
smiklosovic: ProgressInfo.Direction can be only IN or OUT. If it is not OUT, it is IN, so we do not need `else if`, is not `else` good enough? Or direction can be null?  I would `assert` that direction is not null and then the `else` branch does not need `else if`.

src/java/org/apache/cassandra/streaming/StreamSession.java
isaacreath: In practice we will be adding a new progress object for every file streamed in each direction by this `StreamSession` (see: [ProgressInfo::hashCode](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/streaming/ProgressInfo.java#L98)). The worst case is the number of total entries is equal to number of files on the local node * number of files on the remote node for each `StreamSession`. 

A simple optimization I can add here would be to remove the progress object from the map once we've completed streaming. In this case when `progress.currentBytes == progress.totalBytes`.  This would clean things up as each file completes and probably improve memory utilization in the average case. This wouldn't handle the worst case where all files complete streaming at the same time.

test/distributed/org/apache/cassandra/distributed/test/metrics/StreamingMetricsTest.java
isaacreath: This just keeps consistent with the way we create a schema in the existing tests (see: https://github.com/apache/cassandra/blob/trunk/test/distributed/org/apache/cassandra/distributed/test/metrics/StreamingMetricsTest.java#L71). If it's better to be simple, we change it in both places.

src/java/org/apache/cassandra/streaming/StreamSession.java
smiklosovic: _ The worst case is the number of total entries is equal to number of files on the local node * number of files on the remote node for each StreamSession._

Is it really _times_ (multiplication)? So thousand files on each side is 1 million progress objects? Or you meant _plus_ ?

Anyway I guess removing it would help indeed, we probably need to be cautions what to do when a streaming session fails in the middle and it is just garbage collected or something similar ... 


btw why do we need to hold whole progress objects? the only thing needed is `currentBytes` from it. If there is something unique in `ProgressInfo`, why not to use it as a key in that map?

isaacreath: Good catch, yeah `+`.

smiklosovic: also, would it be enough to pass just `bytes` used upon ProgressInfo instantiation into `updateMetricsOnProgress` method as another argument? So you would not need to bypass that object at all, probably. I am just trying to find some workaround.

smiklosovic: I think you ultimately need just file name, direction and bytes and you might probably just hash file name and direction to have it unique but I am just figuring this out in my mind, not tested nor anything ... 

Hashing just these two things is probably not necessary either if you dedicate two separate maps for each direction so key in each map would be just a file name.

smiklosovic:
````
        
        Map<String, long[]> ins = new HashMap<>();
        Map<String, long[]> outs = new HashMap<>();

        // at the beginning we would put it in there based on direction        
        ins.put(progress.fileName, new long[] {0, progress.totalBytes});
````

So later we could check when to remove it from that map as we have access to `totalBytes` we previously put there.

Comment (isaacreath) [reply src/java/org/apache/cassandra/streaming/StreamSession.java]: Yes, what we really care about to identify the uniqueness is the `(fileName, direction)` tuple, so two maps could work.

I'm not really sure about why you're looking to store the `progress.totalBytes` value in the `ins` and `outs` maps.  The call to `StreamSession::progress` _should_ always set the same value for a given `(fileName, direction)` pair.

File: src/java/org/apache/cassandra/streaming/StreamSession.java (hunk 9 above)
Comment (isaacreath) [reply src/java/org/apache/cassandra/streaming/StreamSession.java]: In theory `direction` can be null, but in practice, it doesn't appear to be (all calls to `StreamSession::progress` seem to set a value).

Instead of adding the assert, what about using the `@Nonnull` annotation to prevent anyone from passing null in the future. Then we can just use `else`.

File: src/java/org/apache/cassandra/streaming/StreamSession.java (hunk 8 above)
Comment (smiklosovic) [reply src/java/org/apache/cassandra/streaming/StreamSession.java]: too hard to elaborate on that here and now, lets see what you come up with and then we will see

Comment (smiklosovic) [reply src/java/org/apache/cassandra/streaming/StreamSession.java]: I think that main reason for putting there totalBytes was to know when you are allowed to remove that entry from the map based on your `progress.currentBytes == progress.totalBytes` suggestion.

we should really double check what happens when a streaming session fails - what happens with this whole class and how it behaves under such circumstances.

Comment (isaacreath) [reply src/java/org/apache/cassandra/streaming/StreamSession.java]: I tried out removing from the map when `progress.currentBytes == progress.totalBytes` and found that this function is called in [CassandraCompressedStreamReader::read](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/db/streaming/CassandraCompressedStreamReader.java#L96) multiple times (seemingly once per partition written as a file). What this results in is `progress` being called multiple times where `currentBytes == totalBytes`.  I'm guessing this is a behavior we'd like to keep instead of moving the call to progress outside of the `while` loop in `CassandraCompressedStreamReader::read`.

Given that, I'm going to spend some time thinking of another way to clean up this map when file streaming is done.

**Comment by isaacreath** [comment]
@smiklosovic After thinking about this issue for a little bit, @pauloricardomg  and I came up with a different approach. 

We moved the logic which increments the incoming / outgoing bytes metrics from the `StreamSession` to a dedicated class called the `FileStreamMetricsListener`. `FileStreamMetricsListener` objects will then be created by each `CassandraIncomingFile` and `CassandraOutgoingFile`. These then pass the `FileStreamMetricsListener` to the appropriate stream reader / writer. As those reader / writers transfer bytes, `FileStreamMetricsListener::onBytesTransferred` will be called to update the metrics. When file transfer is complete, we call `FileStreamMetricsListener::onStreamSuccessful` to verify that we've tracked the size appropriately (thus ensuring that any new stream reader / writers are calling `onBytesTransferred`). 

This approach addresses the concerns of having an ever-growing map of `ProgressInfo` objects and it neatly encapsulates the logic for updating these counter objects. There are some aspects of this approach of which I am not a fan, namely the fact that there is little ensuring that new Stream reader / writer objects actually create a `FileStreamMetricsListener` and update it properly; but, from looking at the code, there is some refactoring which needs to be done to unify all of the stream reader / writer classes in a single hierarchy that should probably be handled outside the scope of this ticket before we could consider cleaning up this approach.

File: src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java [hunk 10]
```diff
@@ -0,0 +1,96 @@
...
+import org.apache.cassandra.metrics.StreamingMetrics;
+import org.apache.cassandra.streaming.ProgressInfo;
+import org.apache.cassandra.utils.FBUtilities;
+
+/**
+ * Used to keep track of the bytes streamed for a given file as bytes of the file are transferred and report them to StreamingMetrics.
+ */
+public class FileStreamMetricsListener
+{
+    public static final FileStreamMetricsListener NO_OP = new FileStreamMetricsListener(0L, FBUtilities.getLocalAddressAndPort(), ProgressInfo.Direction.IN){
```
smiklosovic: Should not be the curly bracket at the new line too? There is also a space missing `){` vs `) {`. 

It is also a little bit counter-intuitive to have a direction `IN` used for NO_OP listener. Why not `OUT`? I think it would be preferrable to define a `Direction` called `NO_OP` or `NA` (Not Available) for this purpose.

However, by calling this constructor, we are running through the logic which interacts with `StreamingMetrics` etc. I do not think that is necessary at all. We could create an empty constructor for `FileStreamMetricsListener` and initialise class properties with nulls for counters and 0 for totalSize in it. That way we avoid messing with `Direction` too.

That no-arg constructor should be private.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
```diff
@@ -0,0 +1,96 @@
...
+            //no-op
+        }
+    };
+
+    private final long totalSize;
+    private final Counter totalOutgoingBytesStreamed;
+    private final Counter peerOutgoingBytesStreamed;
+    private long lastSeenBytes = 0L;
+
+    public FileStreamMetricsListener(long totalSize, InetAddressAndPort peer, ProgressInfo.Direction direction)
```
smiklosovic: non null annotation for `direction` maybe as it was done previously? If we introduce `Direction.NA` then `else` would need to be changed to `else if` for `IN`.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
```diff
@@ -0,0 +1,96 @@
...
+        {
+            this.totalOutgoingBytesStreamed = StreamingMetrics.totalOutgoingBytes;
+            this.peerOutgoingBytesStreamed = StreamingMetrics.get(peer).outgoingBytes;
+        }
+        else
+        {
+            this.totalOutgoingBytesStreamed = StreamingMetrics.totalIncomingBytes;
+            this.peerOutgoingBytesStreamed = StreamingMetrics.get(peer).incomingBytes;
+        }
+
```
smiklosovic: nit: redundant empty line

src/java/org/apache/cassandra/db/streaming/CassandraStreamReader.java
```diff
@@ -109,7 +111,6 @@ public CassandraStreamReader(StreamMessageHeader header, CassandraStreamHeader s
     public SSTableMultiWriter read(DataInputPlus inputPlus) throws Throwable
     {
         long totalSize = totalSize();
-
```
smiklosovic: just leave it as it was?

src/java/org/apache/cassandra/db/streaming/CassandraOutgoingFile.java
```diff
@@ -172,10 +174,11 @@ public void write(StreamSession session, StreamingDataOutputPlus out, int versio
             out.flush();
 
             CassandraStreamWriter writer = header.isCompressed() ?
-                                           new CassandraCompressedStreamWriter(sstable, header, session) :
-                                           new CassandraStreamWriter(sstable, header, session);
+                                           new CassandraCompressedStreamWriter(sstable, header, session, fileStreamMetricsListener) :
+                                           new CassandraStreamWriter(sstable, header, session, fileStreamMetricsListener);
             writer.write(out);
         }
+        fileStreamMetricsListener.onStreamSuccessful();
```
smiklosovic: what if `onStreamSuccessful` returned boolean so it would be called `isStreamSuccessful` and if it is not it would throw an exception? This method already throws `IOException` so it seems to be quite logical to me that if streaming was not successful we would throw as well. Right now we are doing `assert` in it which can throw `AssertionError` but that is not subclass of `IOException`.

Another option is to throw IOException directly from `onStreamSuccessful` with accompanying error message but that feels strange.

src/java/org/apache/cassandra/db/streaming/CassandraEntireSSTableStreamReader.java
```diff
@@ -95,7 +97,6 @@ public SSTableMultiWriter read(DataInputPlus in) throws Throwable
 
         ComponentManifest manifest = header.componentManifest;
         long totalSize = manifest.totalSize();
-
         logger.debug("[Stream #{}] Started receiving sstable #{} from {}, size = {}, table = {}",
```
smiklosovic: probably leave it as it was?

src/java/org/apache/cassandra/db/streaming/CassandraCompressedStreamReader.java
```diff
@@ -57,7 +57,6 @@ public CassandraCompressedStreamReader(StreamMessageHeader header, CassandraStre
     public SSTableMultiWriter read(DataInputPlus inputPlus) throws Throwable
     {
         long totalSize = totalSize();
-
```
smiklosovic: probably leave it as it was?

src/java/org/apache/cassandra/streaming/StreamSession.java
```diff
@@ -200,6 +203,8 @@ public class StreamSession implements IEndpointStateChangeSubscriber
 
     private final TimeUUID pendingRepair;
     private final PreviewKind previewKind;
+    private final Map<String, Long> lastSeenIncomingBytesStreamed;
```
smiklosovic: this does not seem to be necessary anymore with all other changes in this class?

src/java/org/apache/cassandra/streaming/StreamSession.java
```diff
@@ -36,13 +36,16 @@
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
+import javax.annotation.Nonnull;
```
smiklosovic: check unused imports maybe?

smiklosovic (commented):
Thanks @isaacreath . I left couple comments again.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
isaacreath: Agree that is much cleaner.

src/java/org/apache/cassandra/db/streaming/CassandraOutgoingFile.java
pauloricardomg: I don't think `FileStreamMetricsListener` should be responsible for determining whether a stream is successful, since this class is meant to be a dumb consumer. For instance, I think we should update the number of incoming files streamed on this method `(metrics.countStreamedIn(isEntireSSTable))`, since we can assume the stream is completed successfully when `onStreamSuccessful()` is called. The assertion `assert lastSeenBytes == totalSize;` on `onStreamSuccessful()` is just to verify that the behavior of the class `FileStreamMetricsListener` is correct, not to check whether a stream is successful or not.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
pauloricardomg: I was wondering if we should make `FileStreamMetricsListener NO_OP` an explicit testing method instead of a static field (ie. `static FileStreamMetricsListener forTesting()` ) to indicate this is only valid for testing and should not be used by db code - wdyt?

src/java/org/apache/cassandra/db/streaming/CassandraOutgoingFile.java
isaacreath: I tried making the change to `isStreamSuccessful` from `onStreamSuccessful` and it seems to make the `FileStreamMetricsListener` a very leaky abstraction to provide a proper error message and we end up duplicating the checking logic between the `CassandraIncomingFile` and `CassandraOutgoingFile` which I'm not a huge fan of (and again, seems like a leaky abstraction).

I could instead throw an `IOException` instead of running the assert; but, like you say, that feels strange as no IO is actually happening in `onStreamSuccessful`.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
smiklosovic: It should be package protected and annotation `@VisibleForTesting` should be added on that method.

However, creating a method for that seems to be like unnecessary exercise. Nothing wrong with having it like it is, package protected and annotated as we usually do. That serves the same purpose as having "forTesting()" method.

pauloricardomg: SGTM 👍

src/java/org/apache/cassandra/db/streaming/CassandraOutgoingFile.java
smiklosovic: Yes I get that but code-wise that method might throw AssertionError but the method that method is called in throws IOException and that is (technically) just wrong.

src/java/org/apache/cassandra/db/streaming/CassandraStreamReader.java
smiklosovic: @isaacreath maybe you forgot this?

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
pauloricardomg: hmm actually some tests outside the package are using `NO_OP`, so it will probably need to remain public.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
```diff
@@ -0,0 +1,113 @@
...
+    {
+        this.totalSize = totalSize;
+        this.totalOutgoingBytesStreamed = totalOutgoingBytesStreamed;
+        this.peerOutgoingBytesStreamed = peerOutgoingBytesStreamed;
+    }
+
+    /**
+     * Used for NO-OP listener.
+     */
+    private FileStreamMetricsListener(){
```
smiklosovic: @isaacreath put `{` on new line please.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * onStreamSuccessful lets the caller of the FileStreamMetricsHandler ensure that the bytes reported from the
+     * FileStreaMetricsHandler is equal to the declared size of the file that was going to be transferred at the creation
+     * of the FileStreamMetricsHandler.
+     */
+    public void onStreamSuccessful()
+    {
+        assert lastSeenBytes == totalSize;
+    }
+
```
smiklosovic: @isaacreath empty line not necessary.

src/java/org/apache/cassandra/db/streaming/CassandraOutgoingFile.java
pauloricardomg: > that method might throw AssertionError but the method that method is called in throws IOException and that is (technically) just wrong.

I don't see what is wrong about this, as far as I understand assertions can be added anywhere in the code to validate assumptions, without requiring methods to throw `AssertionError` explicitly. Am I missing something?

smiklosovic: OK Paulo.

May I ask, what happens when it actually throws that exception (AssertionError). What happens in runtime? Where it propagates and what eventually fails?

Just honestly asking as I have never checked out this code in IDE to dig deeper, sorry ...

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
smiklosovic: ah right, that is fine if we remain to have that annotation on it I guess?

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
```diff
@@ -0,0 +1,113 @@
...
+        }
+        else
+        {
+            this.totalOutgoingBytesStreamed = StreamingMetrics.totalIncomingBytes;
+            this.peerOutgoingBytesStreamed = StreamingMetrics.get(peer).incomingBytes;
+        }
+    }
+
+   @VisibleForTesting
+   public FileStreamMetricsListener(long totalSize, Counter totalOutgoingBytesStreamed, Counter peerOutgoingBytesStreamed)
```
smiklosovic: @isaacreath just checking if this needs to be explicitly `public` or it may be package protected.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
```diff
@@ -0,0 +1,113 @@
...
+    private final Counter totalOutgoingBytesStreamed;
+    private final Counter peerOutgoingBytesStreamed;
+    private long lastSeenBytes = 0L;
+
+    public FileStreamMetricsListener(long totalSize, InetAddressAndPort peer, @Nonnull ProgressInfo.Direction direction)
+    {
+        this.totalSize = totalSize;
+        if (direction == ProgressInfo.Direction.OUT)
+        {
+            this.totalOutgoingBytesStreamed = StreamingMetrics.totalOutgoingBytes;
```
smiklosovic: @isaacreath  I am not big fan of `this`. There is nothing but _this_ `totalOutgoingBytesStreamed`, it can not happen we need `super` one etc ... I think `this` is just redundant here, same for all other cases in constructors.

This class does not extend anything either so we would need to specifically refer to `this`.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
```diff
@@ -0,0 +1,113 @@
...
+     * Report to the FileStreamMetricsHandler the total number of bytes transferred so far.
+     * @param cumulativeBytesTransferred The total number of bytes transferred for a file so far. We use this and calculate
+     *                                   the delta from the last call ourselves as this is the number which the Stream Reader
+     *                                   and Writer classes track
+     */
+    public void onStreamingBytesTransferred(long cumulativeBytesTransferred)
+    {
+        long delta = cumulativeBytesTransferred - lastSeenBytes;
+        assert delta >= 0;
+        this.totalOutgoingBytesStreamed.inc(delta);
```
smiklosovic: @isaacreath why `this` ?

src/java/org/apache/cassandra/db/streaming/CassandraOutgoingFile.java
pauloricardomg: > May I ask, what happens when it actually throws that exception (AssertionError). What happens in runtime? Where it propagates and what eventually fails?

The way I see it is that assertions shouldn't happen in runtime unless the code is broken, as they're errors targeted for developers to detect broken code assumptions. Unfortunately in Cassandra we overload assertions a bit and use it for error handling (which IMO is a bad smell), but this is not what is being done here.

The assertion `assert lastSeenBytes == totalSize` on `FileStreamMetricsListener.onStreamSuccessful` is a runtime safety check to ensure that the behavior of the class `FileStreamMetricsListener` is working as expected and we don't report more streamed bytes than the expected amount. It's a future-proof safety check for programming errors, but should not be expected to fail during correct operation.  If this assertion is ever thrown, it means that there is a bug in the code that must be fixed.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
```diff
@@ -0,0 +1,113 @@
...
+    };
+
+    private final long totalSize;
+    private final Counter totalOutgoingBytesStreamed;
+    private final Counter peerOutgoingBytesStreamed;
+    private long lastSeenBytes = 0L;
+
+    public FileStreamMetricsListener(long totalSize, InetAddressAndPort peer, @Nonnull ProgressInfo.Direction direction)
+    {
+        this.totalSize = totalSize;
```
smiklosovic: in this case, `this` is indeed justified.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
isaacreath: Removed extraneous `this` usage.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
isaacreath: I can't imagine any reason it needs to be used outside of the specific unit tests for this class. Removed it.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
isaacreath: Removed extraneous `this` usage.

src/java/org/apache/cassandra/db/streaming/CassandraOutgoingFile.java
isaacreath: For a sanity check, we can look into whether or not this can happen during valid operation before we commit this assertion. As far as I can tell it can't, but it's worth digging into further.

src/java/org/apache/cassandra/db/streaming/FileStreamMetricsListener.java
```diff
@@ -0,0 +1,113 @@
...
+        this.totalOutgoingBytesStreamed = totalOutgoingBytesStreamed;
+        this.peerOutgoingBytesStreamed = peerOutgoingBytesStreamed;
+    }
+
+    /**
+     * Used for NO-OP listener.
+     */
+    private FileStreamMetricsListener()
+    {
+        this.totalSize = 0L;
```
smiklosovic: @isaacreath `this` in this constructor is not necessary either.

test/microbench/org/apache/cassandra/test/microbench/ZeroCopyStreamingBenchmark.java
```diff
@@ -145,7 +145,6 @@ public void setupBenchmark() throws IOException
                                      .withFirstKey(sstable.first)
                                      .withTableId(sstable.metadata().id)
                                      .build();
-
```
smiklosovic: @isaacreath I know it is super annoying but if you manage to put these lines back (on line 171 too), this file would stayed unchanged and your patch would be smaller (in total 3 files less when all my comments are reflected!) I totally get that this seems like a lot of red tape but we need to contain the dragon / lower the entropy as much as we can.

src/java/org/apache/cassandra/db/streaming/CassandraOutgoingFile.java
```diff
@@ -170,7 +170,6 @@ public void write(StreamSession session, StreamingDataOutputPlus out, int versio
             // legacy streaming is not affected by stats metadata mutation and index sumary redistribution
             CassandraStreamHeader.serializer.serialize(header, out, version);
             out.flush();
-
```
smiklosovic: @isaacreath this change is not necessary, please move it back so this file is not marked as changed in git.

src/java/org/apache/cassandra/db/streaming/CassandraIncomingFile.java
```diff
@@ -68,7 +68,6 @@ public synchronized void read(DataInputPlus in, int version) throws Throwable
         CassandraStreamHeader streamHeader = CassandraStreamHeader.serializer.deserialize(in, version);
         logger.debug("Incoming stream entireSSTable={} components={}", streamHeader.isEntireSSTable, streamHeader.componentManifest);
         session.countStreamedIn(streamHeader.isEntireSSTable);
-
```
smiklosovic: @isaacreath same

test/microbench/org/apache/cassandra/test/microbench/ZeroCopyStreamingBenchmark.java
smiklosovic: cc @pauloricardomg

PR #2066, opened 2023-01-05
Cassandra 18110

dcapwell:


src/java/org/apache/cassandra/streaming/StreamSession.java
```diff
@@ -1187,10 +1189,10 @@ public synchronized void prepareReceiving(StreamSummary summary)
             receivers.put(summary.tableId, new StreamReceiveTask(this, summary.tableId, summary.files, summary.totalSize));
     }
 
-    private void startStreamingFiles(boolean notifyPrepared)
+    private void startStreamingFiles(@Nullable PrepareType prepareType)
     {
-        if (notifyPrepared)
-            streamResult.handleSessionPrepared(this);
+        if (prepareType != null)
```
dcapwell: not sure why this was here before... it was `(true)` at every call-site.

src/java/org/apache/cassandra/streaming/StreamManagerMBean.java
```diff
@@ -29,4 +29,14 @@ public interface StreamManagerMBean extends NotificationEmitter
      * Returns the current state of all ongoing streams.
      */
     Set<CompositeData> getCurrentStreams();
+
+    /**
+     * @return weather or not the streaming virtual table should collect stats while streaming is running
```
aratno: nit: whether

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -875,6 +875,8 @@ public static void setClientMode(boolean clientMode)
     public volatile DurationSpec.LongNanosecondsBound streaming_state_expires = new DurationSpec.LongNanosecondsBound("3d");
     public volatile DataStorageSpec.LongBytesBound streaming_state_size = new DataStorageSpec.LongBytesBound("40MiB");
 
+    public volatile boolean streaming_stats_enabled = true;
```
aratno, line 878: Can you add a comment about why someone might want to disable this, the potential conflict with TCP_USER_TIMEOUT (via internode_streaming_tcp_user_timeout), and the consequences of disabling it?

src/java/org/apache/cassandra/streaming/StreamSession.java
```diff
@@ -158,6 +158,8 @@ public class StreamSession implements IEndpointStateChangeSubscriber
 {
     private static final Logger logger = LoggerFactory.getLogger(StreamSession.class);
 
+    public enum PrepareType { SEND, ACK }
```
aratno: nit: calling this type feels ambiguous to me - maybe PrepareDirection?

test/distributed/org/apache/cassandra/distributed/test/streaming/StreamingStatsDisabledTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+            cluster.get(2).nodetoolResult("rebuild", "--keyspace", KEYSPACE).asserts().success();
+            for (int nodeId : Arrays.asList(1, 2))
+                QueryResultUtil.assertThat(cluster.get(nodeId).executeInternalWithResult("SELECT * FROM system_views.streaming")).isEmpty();
+
+            // trigger streaming again
+            cluster.get(1).executeInternal(withKeyspace("INSERT INTO %s.users(user_id) VALUES ('trigger streaming')"));
+            // mimic JMX
+            cluster.get(2).runOnInstance(() -> StreamManager.instance.setStreamingStatsEnabled(true));
+            cluster.get(2).nodetoolResult("repair", KEYSPACE).asserts().success();
```
aratno, line 59: Why rebuild to trigger streaming the first time but repair the second time?

test/distributed/org/apache/cassandra/distributed/test/streaming/StreamingStatsDisabledTest.java
```diff
@@ -0,0 +1,65 @@
...
+
+            QueryResultUtil.assertThat(cluster.get(1).executeInternalWithResult("SELECT * FROM system_views.streaming")).isEmpty();
+            QueryResultUtil.assertThat(cluster.get(2).executeInternalWithResult("SELECT * FROM system_views.streaming")).hasSize(1);
```
aratno, line 62: Will this flake if streaming finishes too quickly?

src/java/org/apache/cassandra/streaming/StreamingState.java
```diff
@@ -54,18 +57,14 @@ public class StreamingState implements StreamEventHandler
...
     private final TimeUUID id;
     private final boolean follower;
     private final StreamOperation operation;
-    private Set<InetSocketAddress> peers = null;
-    private Sessions sessions = Sessions.EMPTY;
+    private final Set<InetSocketAddress> peers = Collections.newSetFromMap(new ConcurrentHashMap<>());
+    @GuardedBy("this")
+    private final ObjectLongMap<Pair<InetAddressAndPort, String>> activeFiles = new ObjectLongHashMap<>();
+    @GuardedBy("this")
+    private final Sessions sessions = new Sessions();
```
aratno: Thinking out loud: this shouldn't cause deadlock since "this" / synchronized will use the same reentrant lock

src/java/org/apache/cassandra/streaming/StreamingState.java
```diff
@@ -357,41 +333,9 @@ public static String columns()
...
-            return new Sessions(bytesToReceive, bytesReceived,
-                                bytesToSend, bytesSent,
-                                filesToReceive, filesReceived,
-                                filesToSend, filesSent);
-        }
-
         public boolean isEmpty()
         {
-            return this == EMPTY;
+            return bytesToReceive == 0 && bytesToSend == 0 && filesToReceive == 0 && filesToSend == 0;
```
aratno, line 331: This would consider a Sessions empty if it was not empty but complete, since it doesn't consider already {sent,received} {files,bytes}

aratno (commented):
Would you be willing to also include a NoSpamLogger message when stream events take longer than a fixed threshold? I was thinking of a 1ms threshold, with a 1 minute spam duration. That would gives users the feedback required for them to disable streaming_stats.

src/java/org/apache/cassandra/streaming/StreamingState.java
dcapwell: correct, shouldn't be able to deadlock doe to the reentrant nature

src/java/org/apache/cassandra/streaming/StreamingState.java
dcapwell: if the stream has no work to do, this is `empty` which will not populate the data in the vtable; so `null` rather than `0`.  We "could" solve this but think the vtable having `null` is fine in that case.

test/distributed/org/apache/cassandra/distributed/test/streaming/StreamingStatsDisabledTest.java
dcapwell: because copy/paste =D. You can't rebuild twice, but you can repair... so I needed to trigger a second stream and didn't know how to do that other than repair

test/distributed/org/apache/cassandra/distributed/test/streaming/StreamingStatsDisabledTest.java
dcapwell: we are only checking once we know the stream is complete, so no race.

src/java/org/apache/cassandra/config/Config.java
dcapwell: we don't do comments in this code, but I can add to the yaml

dcapwell: >  the potential conflict with TCP_USER_TIMEOUT

my doc doesn't go into that as that isn't related.  If something goes wrong or you just don't want you can disable; the fat that we had timeouts was a bug, so the JIRA should be what document this IMO.

dcapwell: check out comments added to `conf/cassandra.yaml`

aratno (approved):

test/distributed/org/apache/cassandra/distributed/test/streaming/RebuildStreamingTest.java
```diff
@@ -37,18 +38,21 @@ public class RebuildStreamingTest extends TestBaseImpl
     @Test
     public void test() throws IOException
     {
+        ByteBuffer blob = ByteBuffer.wrap(new byte[1 << 16]);
         try (Cluster cluster = init(Cluster.build(2)
                                            .withConfig(c -> c.with(Feature.values()).set("stream_entire_sstables", false))
                                            .start()))
         {
```
jonmeredith: Can we make this a parametrized tests and check with entire sstable streaming active too? The only change you have to make is the expected number of files increases by 7x in the assertion as entire sstable streaming counts each component it sends. Alternately we could modify the accounting to only increment the file count on the DATA file?

src/java/org/apache/cassandra/streaming/StreamingState.java
```diff
@@ -252,52 +239,55 @@ public synchronized void handleStreamEvent(StreamEvent event)
...
     }
 
     private void streamProgress(StreamEvent.ProgressEvent event)
     {
-        SessionInfo info = streamProgress.get(event.progress.peer);
-        if (info != null)
+        ProgressInfo info = event.progress;
+        Pair<InetAddressAndPort, String> key = Pair.create(info.peer, info.fileName);
+        long seenBytes = activeFiles.getOrDefault(key, 0);
+        long delta = info.currentBytes - seenBytes;
```
jonmeredith: nit: is it worth a check for a non-negative delta?

src/java/org/apache/cassandra/streaming/StreamingState.java
```diff
@@ -252,52 +239,55 @@ public synchronized void handleStreamEvent(StreamEvent event)
...
         }
         else
         {
-            logger.warn("[Stream #{}} ID#{}] Recieved stream progress before prepare; peer={}", id, event.progress.sessionIndex, event.progress.peer);
+            // sending
+            sessions.bytesSent += delta;
+            if (info.isCompleted())
+                sessions.filesSent++;
         }
+        activeFiles.put(key, info.currentBytes);
```
jonmeredith: Is it ok that `activeFiles` gradually grows to the total size of streams transferred and includes completed files once they are transferred. I suppose we need this if we want to check for non-negative deltas (or even better positive, but I'm not sure that's guaranteed).

src/java/org/apache/cassandra/streaming/StreamResultFuture.java
```diff
@@ -219,6 +226,10 @@ synchronized void fireStreamEvent(StreamEvent event)
                 logger.warn("Unexpected exception in listern while calling handleStreamEvent", t);
             }
         }
+        long totalNanos = nanoTime() - startNanos;
+        if (totalNanos > slowEventsLogTimeoutNanos)
+            NoSpamLogger.log(logger, NoSpamLogger.Level.WARN, 1, TimeUnit.MINUTES, "Handling streaming events took longer than {}; took {}",
```
jonmeredith, line 231: nit: include unit of nanos in the message?  Very low level nit as anybody investigating thing will be able to work it out from code.

test/distributed/org/apache/cassandra/distributed/test/streaming/RebuildStreamingTest.java
jonmeredith: also, is it worth lowering `streaming_slow_events_timeout` so we check that messages get logged if slow?

jonmeredith:
+1

Looks pretty solid with a couple of nits, and also noticed that
`org.apache.cassandra.streaming.StreamingState.Sessions#receivedBytesPercent` and `org.apache.cassandra.streaming.StreamingState.Sessions#sentBytesPercent` are not used anywhere.

Also interesting that `div(0,0)` returns `BigDecimal.ZERO`, not `BigDecimal.ONE`... as in this usage you're 100% done sending 0 files.

dcapwell:
> and also noticed that
org.apache.cassandra.streaming.StreamingState.Sessions#receivedBytesPercent and org.apache.cassandra.streaming.StreamingState.Sessions#sentBytesPercent are not used anywhere.

Removed the dead code

> Also interesting that div(0,0) returns BigDecimal.ZERO, not BigDecimal.ONE... as in this usage you're 100% done sending 0 files.

`div` is only used by `progress`, which is 

```
return div(bytesSent + bytesReceived, bytesToSend + bytesToReceive);
```

There are 2 cases where `bytesToSend + bytesToReceive == 0`

1) not seen any PREPARE acks; we don't know what is going to be streamed yet
2) nothing to stream; not sure how common this is... Maybe preview repair hits this, else it would be instance doesn't have anything for the range?

I left a comment to Abe (in here or slack don't remember); the fields are set to `0` on init, so can't tell if #1 has happened yet, so can't tell if we are done (nothing to stream) or not started yet...

Now, how critical is this?  The only usage of `progress` is 

```
// StreamingState
public float progress()
    {
        switch (status)
        {
            case INIT:
                return 0;
            case START:
                return Math.min(0.99f, sessions().progress().floatValue());
            case SUCCESS:
            case FAILURE:
                return 1;
            default:
                throw new AssertionError("unknown state: " + status);
        }
    }
```

So, we only call `sessions().progress()` when we know the stream is started and not completed.  This means that the empty case should be SUCCESS and return `1`, so the only case is #1...

Given all this, the correct value to return is `0` as we are not started yet... should `div` handle this?  I can always move higher up if desired... but `0` is still the correct answer

src/java/org/apache/cassandra/streaming/StreamingState.java
dcapwell: > Is it ok that activeFiles gradually grows to the total size

nope, I forgot to `clear` on success/failure... I used to in the old buffer state but forgot after changing... made such a change...

now, if we should remove when `info.isComplete()` I guess this depends on how defensive we want to be on duplicate events... if we get duplicate events `delta` should be 0 so we wouldn't update those metrics but could double count files...

I am in favor of clearing the key when complete as its not enough to block such corruption

src/java/org/apache/cassandra/streaming/StreamResultFuture.java
dcapwell: > nit: include unit of nanos in the message?

I use `Duration` which makes the nano human readable

jonmeredith (approved):
approved - thanks for the tweaks

src/java/org/apache/cassandra/db/streaming/CassandraCompressedStreamReader.java
```diff
@@ -89,11 +89,15 @@ public SSTableMultiWriter read(DataInputPlus inputPlus) throws Throwable
...
+                long lastBytesRead = 0;
                 while (in.getBytesRead() < sectionLength)
                 {
                     writePartition(deserializer, writer);
                     // when compressed, report total bytes of compressed chunks read since remoteFile.size is the sum of chunks transferred
-                    session.progress(filename + '-' + fileSeqNum, ProgressInfo.Direction.IN, cis.chunkBytesRead(), totalSize);
+                    long bytesRead = cis.chunkBytesRead();
+                    long bytesDelta = bytesRead - lastBytesRead;
+                    lastBytesRead = bytesRead;
+                    session.progress(filename + '-' + fileSeqNum, ProgressInfo.Direction.IN, bytesRead, bytesDelta, totalSize);
```
jonmeredith: Could pull `filename + '-' + fileSeqNum` out of the while loop too rather than hope the compiler is smart enough. Looks true for most of the other call points too.

src/java/org/apache/cassandra/db/streaming/CassandraStreamWriter.java
```diff
@@ -113,7 +113,7 @@ public void write(StreamingDataOutputPlus out) throws IOException
                     start += lastBytesRead;
                     bytesRead += lastBytesRead;
                     progress += (lastBytesRead - transferOffset);
-                    session.progress(sstable.descriptor.filenameFor(Component.DATA), ProgressInfo.Direction.OUT, progress, totalSize);
+                    session.progress(sstable.descriptor.filenameFor(Component.DATA), ProgressInfo.Direction.OUT, progress, lastBytesRead - transferOffset, totalSize);
```
jonmeredith: Could DRY it up 
```
long delta = lastBytesRead - transferOffset;
progress += delta;
session.progress(sstable.descriptor.filenameFor(Component.DATA), ProgressInfo.Direction.OUT, progress, delta, totalSize);
```

src/java/org/apache/cassandra/streaming/StreamingState.java
```diff
@@ -252,52 +237,59 @@ public synchronized void handleStreamEvent(StreamEvent event)
...
-        if (info != null)
+        ProgressInfo info = event.progress;
+
+        long delta = info.deltaBytes;
+        if (delta < 0)
         {
-            info.updateProgress(event.progress);
+            NoSpamLogger.log(logger, NoSpamLogger.Level.WARN, 1, TimeUnit.MINUTES,
+                             "[id={}, key={}] Stream event reported a negative delta ({})",
+                             this.id, Pair.create(info.peer, info.fileName), delta);
```
jonmeredith: Not sure this is still helpful - if you keep, you can remove the Pair.create/import and just log the two components directly. 

Could also move to the ProgressInfo constructor as it should never be created with a negative delta.

dcapwell: Moved to StreamSession.progress as its a better fit

src/java/org/apache/cassandra/db/streaming/CassandraStreamWriter.java
```diff
@@ -112,8 +112,9 @@ public void write(StreamingDataOutputPlus out) throws IOException
                     long lastBytesRead = write(proxy, validator, out, start, transferOffset, toTransfer, bufferSize);
                     start += lastBytesRead;
                     bytesRead += lastBytesRead;
-                    progress += (lastBytesRead - transferOffset);
-                    session.progress(sstable.descriptor.filenameFor(Component.DATA), ProgressInfo.Direction.OUT, progress, totalSize);
+                    long delta = lastBytesRead - transferOffset;
+                    progress += delta;
+                    session.progress(sstable.descriptor.filenameFor(Component.DATA), ProgressInfo.Direction.OUT, progress, delta, totalSize);
```
jonmeredith: incredibly `sstable.descriptor.filenameFor(Component.DATA)` goes off string building too

jonmeredith (approved):

src/java/org/apache/cassandra/db/streaming/CassandraEntireSSTableStreamWriter.java
```diff
@@ -93,7 +93,7 @@ public void write(StreamingDataOutputPlus out) throws IOException
             long bytesWritten = out.writeFileToChannel(channel, limiter);
             progress += bytesWritten;
 
-            session.progress(sstable.descriptor.filenameFor(component), ProgressInfo.Direction.OUT, bytesWritten, length);
+            session.progress(sstable.descriptor.filenameFor(component), ProgressInfo.Direction.OUT, bytesWritten, bytesWritten, length);
```
isaacreath, line 96: I think the first usage of `bytesWritten` may be incorrect (everywhere else uses what is calculated as `progress` above). Doesn't have to block this ticket because it's already done this way, but it might be worth looking at in the future.

dcapwell: can you explain?  This is whole file streaming, so `bytesWritten` should map to the file length.

Progress is only used for logging, showing how much was written

dcapwell: we spoke in Slack, the summary is:

There is the field `progress` that tracks all the bytes written regardless of files, but `session.progress` is relative to the file.  @isaacreath thought that we should do

```
session.progress(sstable.descriptor.filenameFor(component), ProgressInfo.Direction.OUT, progress, bytesWritten, length);
```

This was a misunderstanding and chatting about it cleared things up; we are relative to the file so can't use `progress` in this context

PR #2075, opened 2023-01-10
(Accord) Integrate accord-core range transaction support (but do not support them within C* yet)

belliottsmith:


src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -55,11 +64,312 @@
...
+                    if (slice.contains(key))
+                        forEach.accept(commandsForKey(key));
+                    break;
+                case Range:
+                    // TODO:
+                    throw new UnsupportedOperationException();
+            }
+        }
+
+        public <O> O mapReduceForKey(Routables<?, ?> keysOrRanges, Ranges slice, BiFunction<CommandsForKey, O, O> map, O accumulate, O terminalValue)
```
aweisberg, line 194: Make private?

aweisberg: Make private?

src/java/org/apache/cassandra/service/accord/TokenRange.java
```diff
@@ -36,9 +37,9 @@ public TokenRange(AccordRoutingKey start, AccordRoutingKey end)
         super(start, end);
     }
 
-    public static TokenRange fullRange(TableId tableId)
+    public static TokenRange fullRange(String keyspace, TableId tableId)
```
aweisberg: `TableId` is unused?

aweisberg (commented):
Two minor things. About half way through.

test/distributed/org/apache/cassandra/distributed/test/accord/AccordTestBase.java
```diff
@@ -188,9 +188,9 @@ protected SimpleQueryResult executeWithRetry(Cluster cluster, String check, Obje
     {
         // is this method safe?
         cluster.get(1).runOnInstance(() -> {
-            TransactionStatement stmt = AccordTestUtils.parse(check);
-            if (!isIdempotent(stmt))
-                throw new AssertionError("Unable to retry txn that is not idempotent: cql=" + check);
+//            TransactionStatement stmt = AccordTestUtils.parse(check);
```
aweisberg: Why remove this?

src/java/org/apache/cassandra/service/accord/txn/TxnNamedRead.java
```diff
@@ -114,8 +117,13 @@ public PartitionKey key()
...
         SinglePartitionReadCommand command = (SinglePartitionReadCommand) get();
-        AccordCommandsForKey cfk = (AccordCommandsForKey) safeStore.commandsForKey(key);
-        int nowInSeconds = cfk.nowInSecondsFor(executeAt, isForWriteTxn);
+        // TODO (required, safety): before release, double check reasoning that this is safe
+//        AccordCommandsForKey cfk = ((SafeAccordCommandStore)safeStore).commandsForKey(key);
+//        int nowInSeconds = cfk.nowInSecondsFor(executeAt, isForWriteTxn);
+        // It's fine for our nowInSeconds to lag slightly our insertion timestamp, as to the user
+        // this simply looks like the transaction witnessed TTL'd data and the data then expired
+        // immediately after the transaction executed, and this simplifies things a great deal
+        int nowInSeconds = (int) TimeUnit.MICROSECONDS.toSeconds(executeAt.hlc());
```
aweisberg, line 123: I am trying to figure out how `nowInSeconds` and insert timestamps aren't `executeAt` timestamps. Are we mapping from the domain of Accord txnids to the domain of Cassandra timestamps using CommandsForKey?


I thought we already went to the trouble in Accord of crafting an `executeAt` time that is > than any timestamp created by any of the dependency transactions so this work being done in `CommandsForKey`  should have been done before `Read` and `Update`?

aweisberg (commented):
Finished with the C* side. Not a ton in here that is controversial except the questions about how `CommandsForKey` has to track and resolve HLC collisions from`TxnId`. I thought this was already being done as part of `PreAccept` and determining `executeAt`

src/java/org/apache/cassandra/service/accord/txn/TxnNamedRead.java
belliottsmith: The `executeAt` is more than 8 bytes so we cannot _guarantee_ being able to fit it into the normal C* register, so we have to either support timestamps larger than 8 bytes, or we have to be able to compress them into 8 bytes in a manner that is consistent across replicas for a given key. 

It is pretty much out of scope of this patch to worry about this, but as you mentioned I believe in the other tickets, and as I have as an action item, this should be _mostly_ solvable with in-memory summaries. My currently vaguely formulated approach would be to track for contiguous ranges the maximum witnessed _evicted_ hlc we have applied to disk, and to bring a given record's state back into memory if we are to apply a transaction within the range with a hlc equal or lower to this. Since we will mostly not evict from cache until older timestamps have expired this should avoid any additional work.

In general, our caching needs lots of work, both to make it hyper lean and to have several layers of varying degrees of summary, but I do not believe we can readily address that until the immutable state refactor lands.

The `nowInSeconds` I think just does not need to worry and can always use the `hlc`. We don't care about overlaps there, as it only affects TTL expiration and tombstone collection (the latter of which we probably won't even use gc grace for eventually, since we have perfect consistency). I made this change partially in this patch I think, which was probably poor hygiene, but was envisaged as part of the asynchrony work for `calculateDeps` that we are punting to another patch, so I can roll those back too.

test/distributed/org/apache/cassandra/distributed/test/accord/AccordTestBase.java
belliottsmith: Ah, this was a compilation issue in trunk. My bad for not restoring it, I think it should have been fixed in one of the earlier patches I would have rebased against.

src/java/org/apache/cassandra/service/accord/txn/TxnNamedRead.java
belliottsmith: Regarding timestamps, I think medium term there are better solutions than this, too. Perhaps actually we should do them nearer term? 

I had been planning to address this much later as part of providing MVCC (should we ever get the budget to invest in that), as I could permit larger than 8-byte timestamps until we are consistent. But there are other options, at least if we ignore inconsistent writes (which anyway are a danger for this, and cannot be properly solved): 

Since we only write to the base table in ascending timestamp order, sstables could simply track per-partition the flush sstable generation, or even the actual `executeAt` - or indeed any arbitrary sequence of bytes. This wouldn't need to be materialised per cell, just used in the rare event of timestamp ties to break the tie between two sstables, with the larger of the tie-breakers copied to the next sstable.

This is a relatively modest change in sstable format, and probably quite tractable. Perhaps no more complicated than introducing a summary index for this information, and requiring no additional memory to maintain.

aweisberg: I can +1 this change in this patch.

For timestamp you mentioned tracking whether a range is dirty to decide if the HLC could be used. What does it mean for a range to be dirty?

belliottsmith: I think I may have meant whether we know of any transaction in the range that might conflict with its HLC - if we can say for a few seconds look back what approximate maximum timestamp has been witnessed for a given range we know if we have a candidate conflict. As time passes we can degrade our accuracy and merge with wider ranges, so hopefully memory use would be quite small, and number of necessary accesses of the accurate per key state kept minimal.

I did experiment today with introducing TieBreaker tokens to cell resolution, but our storage engine has become far too convoluted. It’s doable, but far from simple. I may plug away at it every so often though.

aweisberg (approved):
+1 LGTM. Thanks!

PR #2081, opened 2023-01-11
Fixes ordering in StandaloneUpgrader

Claudenw:
Fixed issue where sstables were not updated in order by sorting on SSTableId

patch by Claude Warren; reviewed by <Reviewers> for CASSANDRA-18143

### DIFF
--- a/test/unit/org/apache/cassandra/io/sstable/LegacySSTableTest.java
+++ b/test/unit/org/apache/cassandra/io/sstable/LegacySSTableTest.java
@@ -690,33 +690,35 @@ public class LegacySSTableTest
      * Finds the first point at which the HashSet inserts a generation before a previous
      * generation in the set. This is used in testing to verfy the sequence of tables where
      * the generation order is significant.
+     *
      * @param legacyVersion the legacy version to work with.
      * @return the generation that causes of the inversion.
      */
-    public static int findInversion(String legacyVersion) {
+    public static int findInversion(String legacyVersion)
+    {
         String ksname = "legacy_tables";
-        String cfname = String.format("legacy_%s_multiple", legacyVersion);;
+        String cfname = String.format("legacy_%s_multiple", legacyVersion);
+        ;
         SSTableFormat.Type formatType = SSTableFormat.Type.BIG;
         int generation = 0;
         SequenceBasedSSTableId id = new SequenceBasedSSTableId(generation);
-        File directory = new File( System.getProperty("java.io.tmpdir"));
+        File directory = new File(System.getProperty("java.io.tmpdir"));
         Version version = formatType.info.getVersion(legacyVersion);
-        Descriptor descriptor = new Descriptor( version,  directory,  ksname,  cfname, id, formatType);
+        Descriptor descriptor = new Descriptor(version, directory, ksname, cfname, id, formatType);
         HashSet<Descriptor> set = new HashSet<>();
-        set.add( descriptor );
+        set.add(descriptor);
         Iterator<Descriptor> iter;
         do
         {
             id = new SequenceBasedSSTableId(++generation);
-            descriptor = new Descriptor(version , directory , ksname , cfname , id , formatType);
+            descriptor = new Descriptor(version, directory, ksname, cfname, id, formatType);
             set.add(descriptor);
             // set has 2 elements and the order should be the old one then the new one.  This test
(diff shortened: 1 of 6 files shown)


test/unit/org/apache/cassandra/io/sstable/LegacySSTableTest.java
```diff
@@ -686,4 +689,77 @@ private static void copyFile(File cfDir, File file) throws IOException
...
+            {
+                byte[] buf = new byte[65536];
+                if (file.isFile())
+                {
+                    String[] fileNameParts = file.name().split("-");
+                    String targetName = String.format( "%s-%s-%s-%s", legacyVersion, start+i, fileNameParts[2], fileNameParts[3]);
+                    logger.info("creating legacy sstable {}", targetName);
+                    File target = new File(cfDir, targetName);
+                    int rd;
+                    try (FileInputStreamPlus is = new FileInputStreamPlus(file);
```
jacek-lewandowski: I believe there is some `FileUtils.copy` method

test/unit/org/apache/cassandra/io/sstable/LegacySSTableTest.java
```diff
@@ -686,4 +689,77 @@ private static void copyFile(File cfDir, File file) throws IOException
...
+        String cfname = String.format("legacy_%s_multiple", legacyVersion);;
+        SSTableFormat.Type formatType = SSTableFormat.Type.BIG;
+        int generation = 0;
+        SequenceBasedSSTableId id = new SequenceBasedSSTableId(generation);
+        File directory = new File( System.getProperty("java.io.tmpdir"));
+        Version version = formatType.info.getVersion(legacyVersion);
+        Descriptor descriptor = new Descriptor( version,  directory,  ksname,  cfname, id, formatType);
+        HashSet<Descriptor> set = new HashSet<>();
+        set.add( descriptor );
+        Object[] arry = set.toArray();
```
jacek-lewandowski: I suppose it should be a generic method somewhere in the utility class. Something which take a function, for example `<T> T generateUntilUnordered(Collection<T> collection, Function<T, T> generator)` where `generator` accepts previously generated value, and initially `null`.

test/unit/org/apache/cassandra/tools/StandaloneUpgraderOnSStablesTest.java
```diff
@@ -138,22 +138,58 @@ public void testUpgrade() throws Throwable
...
+        System.out.println( tool.getStdout() );
+        Assertions.assertThat(tool.getStdout()).contains("Found 3 sstables that need upgrading.");
+        Assertions.assertThat(tool.getStdout()).contains("legacy_tables/" + tableName);
+        int[] loc = new int[3];
+        for (int i=0;i<3;i++) {
+            String fileName = String.format( "%s-%s-big-Data.db", legacyId, startGeneration+i);
+            Assertions.assertThat(tool.getStdout()).contains(fileName);
+            loc[i] = tool.getStdout().lastIndexOf(fileName);
+        }
+        Assert.assertTrue( "Files processed out of sequence", loc[0] < loc[1] );
```
jacek-lewandowski: Use `Assertions.assertThat(loc).isSorted` or something like that from assertj

test/unit/org/apache/cassandra/io/sstable/LegacySSTableTest.java
Claudenw: I don't think this will work here.  The issue is that the HashSet wraps the values of the hash long before Hash collides. The method literally has to insert into the Hash table and then verify that the last item in the Hash table was the one inserted.  I have simplified the code so that it only retains two Descriptors at a time.

jacek-lewandowski: I was thinking about something like this:

```java
    static <T> T findFirstUnordered(java.util.function.Function<T, T> provider)
    {
        HashSet<T> set = new HashSet<>(2);
        T first = null;
        while (true)
        {
            set.clear();
            T second = provider.apply(first);
            if (first != null)
            {
                set.add(first);
                set.add(second);
                if (set.iterator().next() != first)
                    return second;
            }
            first = second;
        }
    }
```

smiklosovic:
this was delivered, closing

PR #2088, opened 2023-01-12
CASSANDRA-18154: CEP-15: Enhance returning SELECT to allow partition and clustering IN clauses to return multiple partitions/rows

dcapwell:


src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
```diff
@@ -177,9 +178,16 @@ TxnNamedRead createNamedRead(NamedSelect namedSelect, QueryOptions options)
         SinglePartitionReadQuery.Group<SinglePartitionReadCommand> selectQuery = (SinglePartitionReadQuery.Group<SinglePartitionReadCommand>) readQuery;
 
         if (selectQuery.queries.size() != 1)
-            throw new IllegalArgumentException("Within a transaction, SELECT statements must select a single partition; found " + selectQuery.queries.size() + " partitions");
+        {
+            if (!TxnDataName.returning().equals(namedSelect.name))
+                throw new IllegalArgumentException("Within a transaction, SELECT statements must select a single partition; found " + selectQuery.queries.size() + " partitions");
```
maedhroz:
```suggestion
                throw new IllegalArgumentException("Within a transaction, implicit reads and reads within LET statements must select a single partition; found " + selectQuery.queries.size() + " partitions");
```

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
```diff
@@ -177,9 +178,16 @@ TxnNamedRead createNamedRead(NamedSelect namedSelect, QueryOptions options)
...
+            // multi partitions on the same table are only allowed for the returning clause
+            return IntStream.range(0, selectQuery.queries.size())
+                            .mapToObj(i -> new TxnNamedRead(TxnDataName.returning(i), selectQuery.queries.get(i)))
+                            .collect(Collectors.toList());
```
maedhroz: This should be a pretty hot path, so I'd have two potential concerns.

1.) I think we'll create less overall garbage w/ something like...

```
List<TxnNamedRead> list = new ArrayList<>(selectQuery.queries.size());
for (int i = 0; i < selectQuery.queries.size(); i++)
    list.add(new TxnNamedRead(TxnDataName.returning(i), selectQuery.queries.get(i)));
```

2.) Given how many times we'll only have one query, I'd leave `createNamedRead()` alone and create a new `createNamedReads()` to house the new logic here. No point in creating even a singleton list if we don't have to.

maedhroz (changes requested):
Left a few minor performance notes, but once those are fixed, this looks like the least invasive way we could possibly do this. Good stuff.

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
```diff
@@ -182,6 +182,24 @@ TxnNamedRead createNamedRead(NamedSelect namedSelect, QueryOptions options)
         return new TxnNamedRead(namedSelect.name, Iterables.getOnlyElement(selectQuery.queries));
     }
 
+    List<TxnNamedRead> createNamedReads(NamedSelect namedSelect, QueryOptions options)
+    {
+        SelectStatement select = namedSelect.select;
+        ReadQuery readQuery = select.getQuery(options, 0);
+
+        // We reject reads from both LET and SELECT that do not specify a single row.
```
maedhroz, line 190: nit: can remove now

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
```diff
@@ -195,9 +213,10 @@ private List<TxnNamedRead> createNamedReads(QueryOptions options, Consumer<Key>
 
         if (returningSelect != null)
         {
-            TxnNamedRead read = createNamedRead(returningSelect, options);
-            keyConsumer.accept(read.key());
-            reads.add(read);
+            for (TxnNamedRead read : createNamedReads(returningSelect, options)) {
```
maedhroz: nit: newline before brace?

maedhroz (approved):

PR #2103, opened 2023-01-18
Support asynchronous CommandStore message processing

belliottsmith:
While messages are submitted to a CommandStore asynchronously, the processing of a message is then expected to be synchronous. However, it would be beneficial particularly for large range transactions, but also for recovery, to be able to defer some work that might involve IO to after the command store processing completes, e.g. for calculating dependencies.


aweisberg (approved):
+1

PR #2105, opened 2023-01-18
CASSANDRA-18155 Coordinator level metrics for read response and mutation row and column counts

maedhroz:


maedhroz:
https://app.circleci.com/pipelines/github/maedhroz/cassandra?branch=CASSANDRA-18155

src/java/org/apache/cassandra/cql3/restrictions/ClusteringColumnRestrictions.java
```diff
@@ -228,5 +228,9 @@ private boolean handleInFilter(SingleRestriction restriction, int index)
     {
         return restriction.isContains() || restriction.isLIKE() || index != restriction.getFirstColumn().position();
     }
-
+    
+    public RestrictionSet getRestrictionSet()
```
bereng: Could this be more useful in `RestrictionSetWrapper` ? At the end of the day it's like calling `getColumnDefs()` and iterating while calling `getRestrictions(ColumnMetadata columnDef)`, which all live in the super.

src/java/org/apache/cassandra/cql3/restrictions/RestrictionSet.java
```diff
@@ -104,6 +104,17 @@ public List<ColumnMetadata> getColumnDefs()
     {
         return new ArrayList<>(restrictions.keySet());
     }
+    
+    @Override
+    public Collection<ColumnMetadata> getColumnDefinitions()
```
bereng, line 112: Why not use the already existing `getColumnDefs()`?

src/java/org/apache/cassandra/cql3/restrictions/RestrictionSet.java
```diff
@@ -104,6 +104,17 @@ public List<ColumnMetadata> getColumnDefs()
...
+    {
+        return restrictions.keySet();
+    }
+
+    public Map<ColumnMetadata, SingleRestriction> getRestrictions()
```
bereng: We're already iterable do we need this?

src/java/org/apache/cassandra/cql3/restrictions/RestrictionSetWrapper.java
```diff
@@ -57,6 +58,12 @@ public List<ColumnMetadata> getColumnDefs()
         return restrictions.getColumnDefs();
     }
 
+    @Override
+    public Collection<ColumnMetadata> getColumnDefinitions()
```
bereng, line 62: Same, am I missing sthg?

src/java/org/apache/cassandra/cql3/restrictions/StatementRestrictions.java
```diff
@@ -390,6 +390,53 @@ public boolean isColumnRestrictedByEq(ColumnMetadata columnDef)
...
+    /**
+     * This method determines whether a specified column is restricted on equality or something equivalent, like IN.
+     * It can be used in conjunction with the columns selected by a query to determine which of those columns is 
+     * already bound by the client and not necessarily retrieved by the database.
+     *
+     * @param column a column specification from the same table these restrictions are against
+     *
+     * @return true if the given column is restricted on equality
+     */
+    public boolean isEqualityRestricted(ColumnSpecification column)
```
bereng: Here you're potentially traversing all columns on each call. Given you're on a hot path, could you be feeding instead this method with `ColumnMetada`, pulled from the restrictions, and relaying on `KIND` to spare a couple expensive loops?

src/java/org/apache/cassandra/db/partitions/PartitionUpdate.java
```diff
@@ -466,6 +466,56 @@ public List<CounterMark> collectCounterMarks()
...
+    {
+        // If there is a partition-level deletion, we intend to delete at least the columns of one row.
+        if (!partitionLevelDeletion().isLive())
+            return metadata().regularAndStaticColumns().size();
+
+        int count = 0;
+
+        // Each range delete should correspond to at least one intended row deletion, and with it, its regular columns.
+        if (deletionInfo().hasRanges())
+            count += deletionInfo().rangeCount() * metadata().regularColumns().size();
```
bereng, line 507: +stattics? +counters?

src/java/org/apache/cassandra/cql3/restrictions/RestrictionSet.java
maedhroz: It actually looks like this is simply unused. Removing it...

src/java/org/apache/cassandra/cql3/restrictions/RestrictionSet.java
maedhroz: I think the worry there (for performance) was what appeared to be a needless defensive copy in `getColumnDefs()`. I didn't want to audit the 12 other existing callers of `getColumnDefs()`. Certainly open to naming changes...

src/java/org/apache/cassandra/cql3/restrictions/ClusteringColumnRestrictions.java
maedhroz: Makes sense...I'll pull it up.

src/java/org/apache/cassandra/cql3/restrictions/RestrictionSetWrapper.java
maedhroz: see above (tl;dr avoid defensive copy)

src/java/org/apache/cassandra/db/partitions/PartitionUpdate.java
maedhroz: I'm not 100% sure on counters. They are just regular columns, correct? (Going up a level, this is really a pretty conservative lower bound estimate, given range deletes can delete more than one row. There just isn't a way to figure that out without actually doing a read, etc.)

Good catch on the statics. They wouldn't be covered by range deletes, but I'm still not accounting for them properly in either `affectedRowCount()` or `affectedColumnCount()`. Working on a couple new tests, and should have something up shortly...

src/java/org/apache/cassandra/cql3/restrictions/StatementRestrictions.java
maedhroz: Let me mull this one over. In the meantime though, note that we can disable all of this w/ `client_request_size_metrics_enabled` in the YAML.

src/java/org/apache/cassandra/cql3/restrictions/RestrictionSet.java
maedhroz: Also, we may be able to delete these entirely, depending on what we do w/ https://github.com/apache/cassandra/pull/2105#discussion_r1083887949

src/java/org/apache/cassandra/cql3/restrictions/StatementRestrictions.java
bereng: You can disable it yes, but when enabled you want it as performant as possible imo.

src/java/org/apache/cassandra/cql3/restrictions/RestrictionSet.java
bereng: Ohhh icwym with the defensive copy. Then I would either document it, rename it or add a flag to choose the desired behavior. Anything that makes it obvious why there are 2 options to choose.

maedhroz:
@bereng I've pushed changes to properly account for statics on the write side and make `isEqualityRestricted()` less expensive. Let me know what you think...

src/java/org/apache/cassandra/cql3/restrictions/RestrictionSet.java
maedhroz: Agreed. Will at least JavaDoc that for posterity...

src/java/org/apache/cassandra/cql3/restrictions/RestrictionSet.java
```diff
@@ -105,6 +105,15 @@ public List<ColumnMetadata> getColumnDefs()
         return new ArrayList<>(restrictions.keySet());
     }
 
+    /**
+     * @return a direct reference to the key set from {@link #restrictions} with no defenseive copying
```
bereng, line 109: nit: `a` -> `A direct reference...`

src/java/org/apache/cassandra/cql3/restrictions/StatementRestrictions.java
```diff
@@ -390,6 +390,54 @@ public boolean isColumnRestrictedByEq(ColumnMetadata columnDef)
...
+                    if (restricted.name.equals(column.name))
+                        return true;
+        }
+        else if (column.kind == ColumnMetadata.Kind.CLUSTERING)
+        {
+            if (hasClusteringColumnsRestrictions())
+            {
+                for (SingleRestriction restriction : clusteringColumnsRestrictions.getRestrictionSet())
+                {
+                    if (restriction.isEQ() || restriction.isIN())
```
bereng: Did we forget about `CONTAINS` here?

bereng: Maybe you want to collect the result in a variable and return at the end. That might help on branch prediction by avoiding early returns, although iirc most modern cpus are already optimized on that front.

src/java/org/apache/cassandra/cql3/restrictions/RestrictionSet.java
maedhroz: see https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#format

tl;dr I think the spirit of the official formatting style there is that the `@return` is sort of like the first word in a "sentence"

src/java/org/apache/cassandra/cql3/restrictions/StatementRestrictions.java
maedhroz: In the sense that `CONTAINS` and `CONTAINS_KEY` both deal w/ information provided by the user, I think it makes sense to include them if we're looking at whether or not the metric counts them as "read". I'll throw together a couple tests. While I'm at it, something like `isEqualityBased()` on `SingleRestriction` to avoid a messy/long if condition here probably also makes sense.

src/java/org/apache/cassandra/cql3/restrictions/RestrictionSet.java
bereng: TIL. Works for me.

maedhroz:
Committed as https://github.com/apache/cassandra/commit/4892331eb49f72d2e18432a2af56538d11c7c5d1

PR #2109, opened 2023-01-23
CASSANDRA-18187 - Add unit tests for per-row TTL and Timestamp usage in CQLSSTableWriter

JeetKunDoug:
CQLSSTableWriter supports per-row setting of both timestamp and TTL values, but it’s not tested or documented today. Add tests to cover setting both TTL and Timestamp values for rows using the CQLSSTableWriter.

### DIFF
--- a/src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
@@ -58,7 +58,8 @@ import org.apache.cassandra.schema.*;
 import org.apache.cassandra.service.ClientState;
 import org.apache.cassandra.transport.ProtocolVersion;
 import org.apache.cassandra.utils.ByteBufferUtil;
-import org.apache.cassandra.utils.Clock;
+
+import static org.apache.cassandra.utils.Clock.Global.currentTimeMillis;
 
 /**
  * Utility to write SSTables.
@@ -246,7 +247,7 @@ public class CQLSSTableWriter implements Closeable
         ClientState state = ClientState.forInternalCalls();
         List<ByteBuffer> keys = modificationStatement.buildPartitionKeyNames(options, state);
 
-        long now = Clock.Global.currentTimeMillis();
+        long now = currentTimeMillis();
         // Note that we asks indexes to not validate values (the last 'false' arg below) because that triggers a 'Keyspace.open'
         // and that forces a lot of initialization that we don't want.
         UpdateParameters params = new UpdateParameters(modificationStatement.metadata,
--- a/src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
@@ -58,8 +58,7 @@ import org.apache.cassandra.schema.*;
 import org.apache.cassandra.service.ClientState;
 import org.apache.cassandra.transport.ProtocolVersion;
 import org.apache.cassandra.utils.ByteBufferUtil;
-
-import static org.apache.cassandra.utils.Clock.Global.currentTimeMillis;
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
```diff
@@ -247,7 +246,7 @@ public CQLSSTableWriter rawAddRow(List<ByteBuffer> values)
         ClientState state = ClientState.forInternalCalls();
         List<ByteBuffer> keys = modificationStatement.buildPartitionKeyNames(options, state);
 
-        long now = currentTimeMillis();
+        long now = Clock.Global.currentTimeMillis();
```
dcapwell: please revert this change, the common style is to static import `Global`, so should remove this style change

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
```diff
@@ -1116,6 +1118,122 @@ public void testFrozenSetTypeCustomOrdered() throws Exception
...
+    public void testWriteWithTimestamps() throws Exception
+    {
+        long now = currentTimeMillis();
+        long then = now - 1000;
+        final String schema = "CREATE TABLE " + qualifiedTable + " ("
+                              + "  k int,"
+                              + "  c1 int,"
+                              + "  c2 int,"
+                              + "  v text,"
+                              + "  PRIMARY KEY (k)"
```
dcapwell: this is confusing, `c1` and `c2` normally means "clustering column" when you read the code, but this does not include it; can you rename these to v1, v2, and v3?  or change the test to make these clustering columns?

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
```diff
@@ -1116,6 +1118,122 @@ public void testFrozenSetTypeCustomOrdered() throws Exception
...
+    }
+    @Test
+    public void testWriteWithTtl() throws Exception
+    {
+        final String schema = "CREATE TABLE " + qualifiedTable + " ("
+                              + "  k int,"
+                              + "  c1 int,"
+                              + "  c2 int,"
+                              + "  v text,"
+                              + "  PRIMARY KEY (k)"
```
dcapwell: this is confusing, `c1` and `c2` normally means "clustering column" when you read the code, but this does not include it; can you rename these to v1, v2, and v3?  or change the test to make these clustering columns?

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
```diff
@@ -1116,6 +1118,122 @@ public void testFrozenSetTypeCustomOrdered() throws Exception
...
+    }
+    @Test
+    public void testWriteWithTimestampsAndTtl() throws Exception
+    {
+        final String schema = "CREATE TABLE " + qualifiedTable + " ("
+                              + "  k int,"
+                              + "  c1 int,"
+                              + "  c2 int,"
+                              + "  v text,"
+                              + "  PRIMARY KEY (k)"
```
dcapwell: this is confusing, `c1` and `c2` normally means "clustering column" when you read the code, but this does not include it; can you rename these to v1, v2, and v3?  or change the test to make these clustering columns?

dcapwell (changes requested):
would also be nice to add tests for statics and collections, that is not a requirement (these tests are fine) but would be nice

src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
JeetKunDoug: 👍

PR #2110, opened 2023-01-24
CASSANDRA-18068 trunk: Allow to attach native masking functions to table columns

adelapena:
This should be merged to a feature branch. We shouldn't include DDM with attached columns into trunk until we have CASSANDRA-18069 and, possibly, CASSANDRA-18070

### DIFF
--- a/doc/cql3/CQL.textile
+++ b/doc/cql3/CQL.textile
@@ -252,8 +252,10 @@ bc(syntax)..
                           '(' <column-definition> ( ',' <column-definition> )* ')'
                           ( WITH <option> ( AND <option>)* )?
 
-<column-definition> ::= <identifier> <type> ( STATIC )? ( PRIMARY KEY )?
+<column-definition> ::= <identifier> <type> ( STATIC )? ( <column_mask> )? ( PRIMARY KEY )?
                       | PRIMARY KEY '(' <partition-key> ( ',' <identifier> )* ')'
+
+<column-mask> ::= MASKED WITH ( DEFAULT | <function> '(' ( <term> (',' <term>)* )?  ')' )
 
 <partition-key> ::= <identifier>
                   | '(' <identifier> (',' <identifier> )* ')'
@@ -413,11 +415,15 @@ __Syntax:__
 bc(syntax).. 
 <alter-table-stmt> ::= ALTER (TABLE | COLUMNFAMILY) (IF NOT EXISTS)? <tablename> <instruction>
 
-<instruction> ::= ADD (IF NOT EXISTS)? ( <identifier> <type> ( , <identifier> <type> )* )
+<instruction> ::= ADD (IF NOT EXISTS)? ( <identifier> <type> ( <column-mask> )?
+                                       ( , <identifier> <type> ( <column-mask> )? )* )
                 | DROP  (IF EXISTS)? ( <identifier> ( , <identifier> )* )
+                | ALTER <identifier> ( <column-mask> | DROP MASKED )
                 | RENAME (IF EXISTS)? <identifier> to <identifier> (AND <identifier> to <identifier>)*
                 | DROP COMPACT STORAGE
                 | WITH  <option> ( AND <option> )*
+
+<column-mask> ::= MASKED WITH ( DEFAULT | <function> '(' ( <term> (',' <term>)* )?  ')' )
 p. 
 __Sample:__
 
@@ -437,6 +443,7 @@ If the table does not exist, the statement will return an error, unless @IF EXIS
 The @<tablename>@ is the table name optionally preceded by the keyspace name.  The @<instruction>@ defines the alteration to perform:
 * @ADD@: Adds a new column to the table. The @<identifier>@ for the new column must not conflict with an existing column. Moreover, columns cannot be added to tables defined with the @COMPACT STORAGE@ option. If the new column already exists, the statement will return an error, unless @IF NOT EXISTS@ is used in which case the operation is a no-op.
 * @DROP@: Removes a column from the table. Dropped columns will immediately become unavailable in the queries and will not be included in compacted sstables in the future. If a column is readded, queries won't return values written before the column was last dropped. It is assumed that timestamps represent actual time, so if this is not your case, you should NOT readd previously dropped columns. Columns can't be dropped from tables defined with the @COMPACT STORAGE@ option. If the dropped column does not already exist, the statement will return an error, unless @IF EXISTS@ is used in which case the operation is a no-op.
+* @ALTER@: Alters an existing column. It can be used to set its data mask, or to set it as unmasked. The data mask is any function meant to obscure the real values of the column.
 * @RENAME@ a primary key column of a table. Non primary key columns cannot be renamed. Furthermore, renaming a column to another name which already exists isn't allowed. It's important to keep in mind that renamed columns shouldn't have dependent secondary indexes. If the renamed column does not already exist, the statement will return an error, unless @IF EXISTS@ is used in which case the operation is a no-op.
 * @DROP COMPACT STORAGE@: Removes Thrift compatibility mode from the table.
 * @WITH@: Allows to update the options of the table. The "supported @<option>@":#createTableOptions (and syntax) are the same as for the @CREATE TABLE@ statement except that @COMPACT STORAGE@ is not supported. Note that setting any @compaction@ sub-options has the effect of erasing all previous @compaction@ options, so you  need to re-specify all the sub-options if you want to keep them. The same note applies to the set of @compression@ sub-options.
@@ -2489,6 +2496,7 @@ CQL distinguishes between _reserved_ and _non-reserved_ keywords. Reserved keywo
 | @LIST@         | no  |
 | @LOGIN@        | no  |
 | @MAP@          | no  |
+| @MASKED@       | no  |
 | @MATERIALIZED@ | yes |
 | @MBEAN@        | yes |
 | @MBEANS@       | yes |
@@ -2553,7 +2561,7 @@ CQL distinguishes between _reserved_ and _non-reserved_ keywords. Reserved keywo
 | @WHERE@        | yes |
 | @WITH@         | yes |
 | @WRITETIME@    | no  |
-| @MAXWRITETIME@    | no  |
+| @MAXWRITETIME@ | no  |
 
 h2(#appendixB). Appendix B: CQL Reserved Types
 
--- a/src/java/org/apache/cassandra/cql3/functions/masking/ColumnMask.java
+++ b/src/java/org/apache/cassandra/cql3/functions/masking/ColumnMask.java
@@ -1,0 +1,231 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.cql3.functions.masking;
+
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+import com.google.common.collect.ImmutableList;
+import org.apache.commons.lang3.StringUtils;
+
+import org.apache.cassandra.cql3.AssignmentTestable;
+import org.apache.cassandra.cql3.CQL3Type;
+import org.apache.cassandra.cql3.ColumnIdentifier;
+import org.apache.cassandra.cql3.CqlBuilder;
+import org.apache.cassandra.cql3.Term;
+import org.apache.cassandra.cql3.Terms;
+import org.apache.cassandra.cql3.functions.Function;
+import org.apache.cassandra.cql3.functions.FunctionName;
+import org.apache.cassandra.cql3.functions.FunctionResolver;
+import org.apache.cassandra.cql3.functions.ScalarFunction;
+import org.apache.cassandra.db.marshal.AbstractType;
+import org.apache.cassandra.db.marshal.ReversedType;
+import org.apache.cassandra.transport.ProtocolVersion;
+
+import static java.lang.String.format;
+import static org.apache.cassandra.cql3.statements.RequestValidations.invalidRequest;
+
+/**
+ * Dynamic data mask that can be applied to a schema column.
+ * <p>
+ * It consists on a partial application of a certain {@link MaskingFunction} to the values of a column, with the
+ * precondition that the type of any masked column is compatible with the type of the first argument of the function.
+ * <p>
+ * This partial application is meant to be associated to specific columns in the schema, acting as a mask for the values
+ * of those columns. It's associated to queries such as:
+ * <pre>
+ *    CREATE TABLE %t (k int PRIMARY KEY, v int MASKED WITH mask_inner(1, 1));
+ *    ALTER TABLE t ALTER v MASKED WITH mask_inner(2, 1);
+ *    ALTER TABLE t ALTER v DROP MASKED;
+ * </pre>
+ * Note that in the example above we are referencing the {@code mask_inner} function with two arguments. However, that
+ * CQL function actually has three arguments. The first argument is always ommitted when attaching the function to a
+ * schema column. The value of that first argument is always the value of the masked column, in this case an int.
+ */
+public class ColumnMask
+{
+    /** The CQL function used for masking. */
+    public final ScalarFunction function;
+
+    /** The values of the arguments of the partially applied masking function. */
+    public final List<ByteBuffer> partialArgumentValues;
+
+    public ColumnMask(ScalarFunction function, List<ByteBuffer> partialArgumentValues)
+    {
+        assert function.argTypes().size() == partialArgumentValues.size() + 1;
+        this.function = function;
+        this.partialArgumentValues = partialArgumentValues;
+    }
+
+    /**
+     * @return The types of the arguments of the partially applied masking function.
+     */
+    public List<AbstractType<?>> partialArgumentTypes()
+    {
+        List<AbstractType<?>> argTypes = function.argTypes();
+        return argTypes.size() == 1
+               ? Collections.emptyList()
+               : argTypes.subList(1, argTypes.size());
+    }
+
+    /**
+     * @return A copy of this mask for a version of its masked column that has its type reversed.
+     */
+    public ColumnMask withReversedType()
+    {
+        AbstractType<?> reversed = ReversedType.getInstance(function.argTypes().get(0));
+        List<AbstractType<?>> args = ImmutableList.<AbstractType<?>>builder()
+                                                  .add(reversed)
+                                                  .addAll(partialArgumentTypes())
+                                                  .build();
+        Function newFunction = FunctionResolver.get(function.name().keyspace, function.name(), args, null, null, null);
+        assert newFunction != null;
+        return new ColumnMask((ScalarFunction) newFunction, partialArgumentValues);
+    }
+
+    /**
+     * @param protocolVersion the used version of the transport protocol
+     * @param value           a column value to be masked
+     * @return the specified value after having been masked by the masked function
+     */
+    public ByteBuffer mask(ProtocolVersion protocolVersion, ByteBuffer value)
+    {
+        List<ByteBuffer> args = new ArrayList<>(partialArgumentValues.size() + 1);
+        args.add(value);
+        args.addAll(partialArgumentValues);
+        return function.execute(protocolVersion, args);
+    }
+
+    @Override
+    public boolean equals(Object o)
+    {
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
+        ColumnMask mask = (ColumnMask) o;
+        return function.name().equals(mask.function.name())
+               && partialArgumentValues.equals(mask.partialArgumentValues);
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(function.name(), partialArgumentValues);
+    }
+
+    @Override
+    public String toString()
+    {
+        List<AbstractType<?>> types = partialArgumentTypes();
+        List<String> arguments = new ArrayList<>(types.size());
+        for (int i = 0; i < types.size(); i++)
+        {
+            CQL3Type type = types.get(i).asCQL3Type();
+            ByteBuffer value = partialArgumentValues.get(i);
+            arguments.add(type.toCQLLiteral(value, ProtocolVersion.CURRENT));
+        }
+        return format("%s(%s)", function.name(), StringUtils.join(arguments, ", "));
+    }
+
+    public void appendCqlTo(CqlBuilder builder)
+    {
+        builder.append(" MASKED WITH ").append(toString());
+    }
+
+    /**
+     * A parsed but not prepared column mask.
+     */
+    public final static class Raw
+    {
+        public final FunctionName name;
+        public final List<Term.Raw> rawPartialArguments;
+
+        public Raw(FunctionName name, List<Term.Raw> rawPartialArguments)
+        {
+            this.name = name;
+            this.rawPartialArguments = rawPartialArguments;
+        }
+
+        public ColumnMask prepare(String keyspace, String table, ColumnIdentifier column, AbstractType<?> type)
+        {
+            ScalarFunction function = findMaskingFunction(keyspace, table, column, type);
+
+            List<ByteBuffer> partialArguments = preparePartialArguments(keyspace, function);
+
+            return new ColumnMask(function, partialArguments);
+        }
+
+        private ScalarFunction findMaskingFunction(String keyspace, String table, ColumnIdentifier column, AbstractType<?> type)
+        {
+            List<AssignmentTestable> args = new ArrayList<>(rawPartialArguments.size() + 1);
+            args.add(type);
+            args.addAll(rawPartialArguments);
+
+            Function function = FunctionResolver.get(keyspace, name, args, keyspace, table, type);
+
+            if (function == null)
+                throw invalidRequest("Unable to find masking function for %s, " +
+                                     "no declared function matches the signature %s",
+                                     column, this);
+
+            if (function.isAggregate())
+                throw invalidRequest("Aggregate function %s cannot be used for masking table columns", this);
+
+            if (!function.isNative())
+                throw invalidRequest("User defined function %s cannot be used for masking table columns", this);
+
+            if (!(function instanceof MaskingFunction))
+                throw invalidRequest("Not-masking function %s cannot be used for masking table columns", this);
+
+            if (!function.returnType().equals(type))
+                throw invalidRequest("Masking function %s return type is %s. " +
+                                     "This is different to the type of the masked column %s of type %s. " +
+                                     "Masking functions can only be attached to table columns " +
+                                     "if they return the same data type as the masked column.",
+                                     this, function.returnType().asCQL3Type(), column, type.asCQL3Type());
+
+            return (ScalarFunction) function;
+        }
+
+        private List<ByteBuffer> preparePartialArguments(String keyspace, ScalarFunction function)
+        {
+            // Note that there could be null arguments
+            List<ByteBuffer> arguments = new ArrayList<>(rawPartialArguments.size());
+
+            for (int i = 0; i < rawPartialArguments.size(); i++)
+            {
+                String term = rawPartialArguments.get(i).toString();
+                AbstractType<?> type = function.argTypes().get(i + 1);
+                arguments.add(Terms.asBytes(keyspace, term, type));
+            }
+
+            return arguments;
+        }
+
+        @Override
+        public String toString()
+        {
+            return format("%s(%s)", name, StringUtils.join(rawPartialArguments, ", "));
+        }
+    }
+}
--- a/src/java/org/apache/cassandra/cql3/selection/Selection.java
+++ b/src/java/org/apache/cassandra/cql3/selection/Selection.java
@@ -103,15 +103,20 @@ public abstract class Selection
      */
     public Integer getOrderingIndex(ColumnMetadata c)
     {
-        if (!isJson)
-            return getResultSetIndex(c);
-
         // If we order post-query in json, the first and only column that we ship to the client is the json column.
         // In that case, we should keep ordering columns around to perform the ordering, then these columns will
         // be placed after the json column. As a consequence of where the colums are placed, we should give the
         // ordering index a value based on their position in the json encoding and discard the original index.
         // (CASSANDRA-14286)
-        return orderingColumns.indexOf(c) + 1;
+        if (isJson)
+            return orderingColumns.indexOf(c) + 1;
+
+        // If the column is masked it might appear twice, once masked in the selected column and once unmasked in
+        // the ordering columns. For ordering we are interested in that second unmasked value.
+        if (c.isMasked())
+            return columns.lastIndexOf(c);
+
+        return getResultSetIndex(c);
     }
 
     public ResultSet.ResultMetadata getResultMetadata()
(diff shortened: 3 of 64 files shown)


src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -137,6 +139,11 @@ private SchemaKeyspace()
               + "kind text,"
               + "position int,"
               + "type text,"
+              + "mask_keyspace text,"
+              + "mask_name text,"
+              + "mask_argument_types frozen<list<text>>,"
+              + "mask_argument_values frozen<list<blob>>,"
+              + "mask_argument_nulls frozen<list<boolean>>," // arguments that are null
```
adelapena: The schema produced by this will be rejected by Cassandra versions without this patch. So this will break downgrades. This is ok if this lands in a major (5.0), but not if it goes to a minor (4.2).

As an alternative, we can just put the masks on a separate table, this way: https://github.com/adelapena/cassandra/commit/cc9983911c6a99954a1e6ef0aeffa8f439f9b888

That separate table for masks will just be ignored by older versions during a downgrade. It replicates the primary key of `system_schema.columns`, which is not ideal unless that's done for supporting downgrades. @blerer wdyt?

doc/cql3/CQL.textile
```diff
@@ -2552,8 +2560,8 @@ CQL distinguishes between _reserved_ and _non-reserved_ keywords. Reserved keywo
 | @VIEW@         | yes |
 | @WHERE@        | yes |
 | @WITH@         | yes |
-| @WRITETIME@    | no  |
-| @MAXWRITETIME@    | no  |
+| @WRITETIME@    | yes |
```
bereng: Is this change unrelated to this PR/ticket? just for my understanding

src/java/org/apache/cassandra/cql3/functions/masking/ColumnMask.java
```diff
@@ -0,0 +1,252 @@
...
+/**
+ * Dynamic data mask that can be applied to a schema column.
+ * <p>
+ * It consists on a partial application of a certain {@link MaskingFunction} to the values of a column, with the
+ * precondition that the type of any masked column is compatible with the type of the first argument of the function.
+ * <p>
+ * This partial application is meant to be associated to specific columns in the schema, acting as a mask for the values
+ * of those columns. It's associated to queries such as:
+ * <pre>
+ *    CREATE TABLE %t (k int PRIMARY KEY, v int MASKED WITH mask_inner(1, 1);
```
bereng: missing closing `) `?

doc/cql3/CQL.textile
adelapena: It's unrelated, I fixed the alignement because the `WITHOUT` keyword is next to it without the additional, optional change mentioned [here](https://issues.apache.org/jira/browse/CASSANDRA-18068?focusedCommentId=17682967&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17682967).

bereng: But why writetime changes from being non reserved to being reserved? Sorry if I am being thick.

src/java/org/apache/cassandra/cql3/functions/masking/ColumnMask.java
```diff
@@ -0,0 +1,252 @@
...
+            if (function.isAggregate())
+                throw invalidRequest("Aggregate function %s cannot be used for masking table columns", this);
+
+            if (!function.isNative())
+                throw invalidRequest("User defined function %s cannot be used for masking table columns", this);
+
+            if (!(function instanceof MaskingFunction))
+                throw invalidRequest("Not-masking function %s cannot be used for masking table columns", this);
+
+            if (!function.returnType().equals(type))
```
bereng, line 200: Note to self: will this fail on reversed types?

src/java/org/apache/cassandra/cql3/selection/Selection.java
```diff
@@ -187,7 +187,9 @@ public static Selection fromSelectors(TableMetadata table,
                                                                             factories,
                                                                             isJson);
 
-        return (processesSelection(selectables) || selectables.size() != selectedColumns.size() || hasGroupBy)
+        boolean hasMaskedColumns = selectedColumns.stream().anyMatch(ColumnMetadata::isMasked);
```
bereng: Shall we avoid streams for perf reasons and move to a classic loop?

src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
```diff
@@ -1153,10 +1153,14 @@ private Selection prepareSelection(TableMetadata table,
...
                 Guardrails.groupByEnabled.ensureEnabled(state);
 
+            boolean isJson = parameters.isJson;
+            boolean returnStaticContentOnPartitionWithNoRows = restrictions.returnStaticContentOnPartitionWithNoRows();
+
             if (selectables.isEmpty()) // wildcard query
             {
-                return hasGroupBy ? Selection.wildcardWithGroupBy(table, boundNames, parameters.isJson, restrictions.returnStaticContentOnPartitionWithNoRows())
-                                  : Selection.wildcard(table, parameters.isJson, restrictions.returnStaticContentOnPartitionWithNoRows());
+                return hasGroupBy || table.columns().stream().anyMatch(ColumnMetadata::isMasked)
```
bereng: Avoid streams?

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
bereng: The separate table solution seems superior right? It has no disadvantage and a clear advantage.

doc/cql3/CQL.textile
adelapena: Oh, right, I was thinking on `MAXWRITETIME ` instead of `WRITETIME `. Fixed.

src/java/org/apache/cassandra/cql3/selection/Selection.java
adelapena: This is run only once per query, so not in the hot path. It's relatively common to use streams while preparing the query.

src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
adelapena: Same as before, it's not in the hot path.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
adelapena: Using two tables has disadvantages in terms of complexity, space used and performance (we duplicate reads and writes when loading and storing the schema). However, the table is small and we don't access it very often, so the benefit of not breaking backwards compatibility could very well outweigh the costs, especially since we want to put this on current trunk, which is still 4.2.

src/java/org/apache/cassandra/cql3/selection/Selection.java
bereng: I disagree it's not in the hot path. It's not for a _specific query_ as yes, it gets prepared only once. But it's in the hot path in terms of how many total prepare requests we can handle. So in low prepared query scenarios _reusage_ it will have an impact.

IOW prepare once and query 1K times we're ok. Prepare once and reuse a handful of times we're not ok. High env traffic env where the prepared stmnts cache rotates often we're not ok. Nodes handling multi-tenant traffic may be handling a high number of prepares given the many different loads/apps they may support and we're not ok.

Maybe food for another ticket? I would still avoid it here if possible.

adelapena: I understand that the hot path where we avoid streams consists on that areas of the code that are repeated for every row or, more probably, column. The number of times those are hit is several orders of magnitude greater than in things that are used only to prepare the query, thereof the preference for performance over readability.

It seems to me that at the moment we are not avoiding streams in query building. Just taking a quick look at the CQL package there are multiple usages of streams, for example:
* [o.a.c.cql3/CQL3Type.java#L840](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/CQL3Type.java#L840)
* [o.a.c.cql3/CQL3Type.java#L876](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/CQL3Type.java#L876)
* [o.a.c.cql3/Lists.java#L133](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/Lists.java#L133)
* [o.a.c.cql3/Sets.java#L123](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/Sets.java#L123)
* [o.a.c.cql3/MultiColumnRelation.java#L225](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/MultiColumnRelation.java#L225)
* [o.a.c.cql3/QueryProcessor.java#L481](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/QueryProcessor.java#L481)
* [o.a.c.cql3/TokenRelation.java#L138](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/TokenRelation.java#L138)
* [o.a.c.cql3/conditions/ColumnConditions.java#L75](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/conditions/ColumnConditions.java#L75)
* [o.a.c.cql3/functions/AbstractFunction.java#L69](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/functions/AbstractFunction.java#L69)
* [o.a.c.cql3/functions/FunctionResolver.java#L203](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/functions/FunctionResolver.java#L203)
* [o.a.c.cql3/functions/UDAggregate.java#L97](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/functions/UDAggregate.java#L97)
* [o.a.c.cql3/restrictions/StatementRestrictions.java#L380](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/restrictions/StatementRestrictions.java#L380)
* [o.a.c.cql3/selection/AbstractFunctionSelector.java#L72](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/selection/AbstractFunctionSelector.java#L72)
* [o.a.c.cql3/selection/MapSelector.java#L113](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/selection/MapSelector.java#L113)
* [o.a.c.cql3/selection/Selectable.java#L708](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/selection/Selectable.java#L708)
* [o.a.c.cql3/selection/Selectable.java#L793](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/selection/Selectable.java#L793)
* [o.a.c.cql3/selection/Selectable.java#L886](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/selection/Selectable.java#L886)
* [o.a.c.cql3/selection/Selectable.java#L954](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/selection/Selectable.java#L954)
* [o.a.c.cql3/selection/Selectable.java#L1024](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/selection/Selectable.java#L1024)
* [o.a.c.cql3/statements/SelectStatement.java#L1278](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/statements/SelectStatement.java#L1278)

We should also check if queries are calling packages other than `o.a.c.cql3` that also use streams.

I'd say that getting rid of those streams used on query building leans towards premature optimization. But, if we are concerned about the performance of streams on query building we should probably open a followup ticket to audit where streams are used.

src/java/org/apache/cassandra/cql3/functions/masking/ColumnMask.java
adelapena: I think it won't fail. We have several tests that use masking with reversed types, but I'm adding some more explicit tests.

Nevertheless, this works because the masking functions added by CASSANDRA-17941 take care of returning a reversed type if their input is reversed. Those functions are the only ones that we allow here, so probably we are fine. However, we might need to revisit this when we add support for using UDFs as masking functions in CASSANDRA-18071.

src/java/org/apache/cassandra/cql3/selection/Selection.java
bereng: I don't think readability is impacted if streams are dropped, they would become quite simple loops. Given it's a cheap and immediate optimization I would still avoid streams here if it were my PR but I'll defer to your preference being the author. It's something that in multi tenancy env might be a thing and will make a nice LHF ticket. I'll be happy to create it at least for discussion [CASSANDRA-18248](https://issues.apache.org/jira/browse/CASSANDRA-18248)

bereng:
Nice one on splitting long tests

src/java/org/apache/cassandra/cql3/functions/masking/ColumnMask.java
bereng: > take care of returning a reversed type if their input is reversed

^ That is what I was wondering thx

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
```diff
@@ -158,6 +161,73 @@ public KeyspaceMetadata apply(KeyspaceMetadata keyspace, TableMetadata table)
...
+
+            if (column == null)
+            {
+                if (!ifColumnExists)
+                    throw ire("Column with name '%s' doesn't exist on table '%s'", columnName, tableName);
+
+                return keyspace;
+            }
+
+            ColumnMask columnMask = mask == null ? null : mask.prepare(keyspace.name, table.name, columnName, column.type);
```
bereng: Do we need to proceed if `columnMask` is null? The later `tableBuilder.alterColumnMask()` would throw anyway iiuc

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -688,15 +695,58 @@ private static void addColumnToSchemaMutation(TableMetadata table, ColumnMetadat
...
                .add("column_name_bytes", column.name.bytes)
                .add("kind", column.kind.toString().toLowerCase())
                .add("position", column.position())
                .add("clustering_order", column.clusteringOrder().toString().toLowerCase())
                .add("type", type.asCQL3Type().toString());
+
+        // Dynamic data masking functions shouldn't be attached to columns during rolling upgrades
+        // to avoid sending mutations with columns that are unknown to the old nodes.
+        ColumnMask mask = column.getMask();
+        if (ColumnMask.clusterSupportsMaskedColumns())
```
bereng: As I am reading the PR I wonder if `clusterSupportsMaskedColumns()` should be moved to `QueryProcessor` or some other 'higher'/clusterish class as being buried in `ColumnMask` sounds weird to me.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -688,15 +695,58 @@ private static void addColumnToSchemaMutation(TableMetadata table, ColumnMetadat
...
+        {
+            if (mask == null)
+            {
+                rowBuilder.delete("mask_keyspace")
```
bereng: `mask_keyspace` & friends could be constants

test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeColumnMaskingTest.java
```diff
@@ -0,0 +1,113 @@
...
+                assertColumnValue(cluster, "t", "v", 7);
+            }
+        }).runAfterClusterUpgrade(cluster -> {
+
+            // create table with masked column
+            cluster.schemaChange(withKeyspace("CREATE TABLE %s.t2 (k int PRIMARY KEY, v int MASKED WITH DEFAULT)"));
+            cluster.coordinator(1).execute(withKeyspace("INSERT INTO %s.t2(k, v) VALUES (0, 7)"), ALL);
+            assertColumnValue(cluster, "t2", "v", 0);
+
+            // mask existing column
```
bereng: Comment is wrong?

test/unit/org/apache/cassandra/cql3/functions/masking/ColumnMaskInAnyPositionTester.java
```diff
@@ -0,0 +1,262 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import org.apache.cassandra.db.marshal.AbstractType;
+
+import static java.lang.String.format;
+
+/**
+ * {@link ColumnMaskTester} verifying that masks can be applied to columns in any position (partition key columns,
+ * clustering key columns, static columns and regular columns). The columns of any depending materialized views should
```
bereng: Should we consider counters as well?

test/unit/org/apache/cassandra/cql3/functions/masking/ColumnMaskInAnyPositionTester.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    @Test
+    public void testAlterTableWithMaskedColumns() throws Throwable
+    {
+        // Create the table to be altered
+        createTable(format("CREATE TABLE %%s (k %s, c %<s, r1 %<s, r2 %<s MASKED WITH %s, r3 %s, s %<s static, " +
+                           "PRIMARY KEY (k, c))", type, mask, type));
+        assertTableColumnsAreMasked("r2");
+
```
bereng: No `assertTableColumnsAreNotMasked()`?

test/unit/org/apache/cassandra/cql3/functions/masking/ColumnMaskNativeTypesTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Test
+    public void testNativeDataTypes() throws Throwable
+    {
+        String def = format("%s MASKED WITH DEFAULT", type);
+        String keyDef = type == CQL3Type.Native.COUNTER || type == CQL3Type.Native.DURATION
+                        ? "int MASKED WITH DEFAULT" : def;
+
+        // Create table with masks
+        String table = createTable(format("CREATE TABLE %%s (k %s, c %<s, r %s, s %<s, PRIMARY KEY(k, c))", keyDef, def));
```
bereng: Is `s` supposed to be static?

src/java/org/apache/cassandra/cql3/functions/masking/ColumnMask.java
```diff
@@ -0,0 +1,252 @@
...
+                arguments.add(Terms.asBytes(keyspace, term, type));
+            }
+
+            return arguments;
+        }
+
+        @Override
+        public String toString()
+        {
+            return name.toString() + '(' + JOINER.join(rawPartialArguments) + ')';
```
blerer: nit: Do we really need to optimize by having a static JOINER here? It seems more confusing that anything else the method is not called on production and is not really optimized either.

src/java/org/apache/cassandra/cql3/selection/Selection.java
blerer: For masked column. `ColumnMetadata.processSelection` should return `true`. That will ensure that  `processesSelection(selectables)` returns `true` for masked columns and avoid the need for `hasMaskedColumns`.

Regarding performance my suggestion would be to use a profiler and see if the use of streams is visible or not. Otherwise it will end up as a long and inefficient discussion.

src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
blerer: I would have expected to have a `table.hasMaskedColumn() ` method to avoid code duplication each time people need that information and to make the code more readable.

test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeColumnMaskingTest.java
```diff
@@ -0,0 +1,113 @@
...
+import org.assertj.core.api.Assertions;
+
+import static org.apache.cassandra.distributed.api.ConsistencyLevel.ALL;
+import static org.apache.cassandra.distributed.api.Feature.GOSSIP;
+import static org.apache.cassandra.distributed.shared.AssertUtils.assertRows;
+import static org.apache.cassandra.distributed.shared.AssertUtils.row;
+
+/**
+ * Tests that dynamic data masking (DDM) functions can be attached to table columns during a rolling upgrade involving
+ * nodes that don't include DDM.
```
blerer: can/cannot

test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeColumnMaskingTest.java
adelapena: It is. Or, more exactly, the block of code immediately below it is swapped with the next block of code :) 

The intention is first masking an unmasked column, then unmasking it. Fixed.

test/unit/org/apache/cassandra/cql3/functions/masking/ColumnMaskInAnyPositionTester.java
adelapena: This test class is focused on positions in the table, regardless of the column type. Counters as a data type are already tested in `ColumnMaskNativeTypesTest`.

test/unit/org/apache/cassandra/cql3/functions/masking/ColumnMaskInAnyPositionTester.java
adelapena: Right, just added them, here and in the block immediately below.

test/unit/org/apache/cassandra/cql3/functions/masking/ColumnMaskNativeTypesTest.java
adelapena: Indeed, good catch! Just fixed it.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
adelapena: In other cases yes, but here I would prefer to keep the style of `SchemaKeyspace`, where column names are always referred to with direct strings instead of constants. There are so many of these that probably the class would lose readability if we made all constants.

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
adelapena: We do need to proceed is `columnMask` is `null`, that's used for `DROP MASKED`. However, that makes me think that we can take a shortcut if the new mask (null or not) is equals to the old mask. Just added it.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
adelapena: I don't know what that class would be. `QueryProcessor` is more CQL related and more related to the coordinator than to the cluster, so I don't see it there. Maybe in `Gossiper`, since the method calls it twice? But that would couple `Gossiper` with masking, which is not ideal.

blerer (commented):
First round

src/java/org/apache/cassandra/cql3/functions/masking/ColumnMask.java
adelapena: We don't, changed to `StringUtils.join`.

src/java/org/apache/cassandra/cql3/selection/Selection.java
adelapena: Makes sense, fixed `ColumnMetadata#processSelection` so the stream/loop is not needed.

src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
adelapena: Added that convenience `TableMetadata#hasMaskedColumns()` method, which is way cleaner. Since that's now a general purpose method that might called in any path I'm changing it to a loop.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
adelapena: @blerer what do you think of this? Should we go with the separate table?

src/java/org/apache/cassandra/cql3/selection/Selection.java
bereng: As I try to argue in CASSANDRA-18248 I think it boils down to deciding if we see streams avoidance as a free lunch or not. Profiling would mean getting the actual change in iiuc.

test/unit/org/apache/cassandra/cql3/functions/masking/ColumnMaskTest.java
```diff
@@ -207,22 +207,22 @@ public void testColumnMaskingWithNotNativeFunction() throws Throwable
     }
 
     @Test
+    @SuppressWarnings("resource")
```
bereng: What's the point of of replacing the auto-closable by this?

bereng:
LGTM +1

test/unit/org/apache/cassandra/cql3/functions/masking/ColumnMaskTest.java
adelapena: The sessions provided by `CQLTester#sessionNet()` are cached in `CQLTester.sessions`. If we close the session (manually or with try-with-resources) then the next call to `CQLTester#sessionNet()` will receive a closed session. 

Note that the multiple direct uses of `CQLTester#sessionNet()` on other tests never try to close the session, and all the flavours of `CQLTester#sessionNet` don't close it either. That seems to indicate that not closing the reusable session is the intended use, even if the warning suppression is not commonly used.

The sessions are later closed in `CQLTester.tearDownClass`.

bereng: Thx!

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
bereng: Just for the record it was commented yesterday 5.0 will be the choice granted by other changes such as removal of scripted UFFs.

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
```diff
@@ -158,6 +162,77 @@ public KeyspaceMetadata apply(KeyspaceMetadata keyspace, TableMetadata table)
...
     }
 
+    /**
+     * ALTER TABLE [IF EXISTS] <table> ALTER [IF EXISTS] <column> ( MASKED WITH <newMask> | DROP MASKED )
+     */
+    public static class MaskColumn extends AlterTableStatement
+    {
+        private final ColumnIdentifier columnName;
+        @Nullable
+        private final ColumnMask.Raw mask;
```
blerer: Nit: We should may be name it `rawMask`. T prepare call  on the `mask` variable is a bit confusing otherwise.

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
```diff
@@ -595,6 +684,10 @@ public static final class Raw extends CQLStatement.Raw
         // ADD
         private final List<AddColumns.Column> addedColumns = new ArrayList<>();
 
+        // ALTER MASK
+        private ColumnIdentifier maskedColumn = null;
+        private ColumnMask.Raw maskedColumnMask = null;
```
blerer: `rawMask `instead of `maskedColumnMask `?

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
blerer: I agree.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -688,15 +695,58 @@ private static void addColumnToSchemaMutation(TableMetadata table, ColumnMetadat
...
+                          .delete("mask_name")
+                          .delete("mask_argument_types")
+                          .delete("mask_argument_values")
+                          .delete("mask_argument_nulls");
+            }
+            else
+            {
+                FunctionName maskFunctionName = mask.function.name();
+
+                // Some arguments of the masking function can be null, but the CQL's list type that stores them doesn't
```
blerer: Why not storing the elements as their CQL string? It will make the table more readable and allow us to store null within the same collection if I am not missing anything.

blerer (commented):
The patch looks good. I raised a few minor points.
As we are moving to 5.0 it seems to me that we can simplify the code related to mix cluster version.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
adelapena: In light of the current discussions about downgradability it might be prudent to use a separate table for the masks, so the schema can be opened after an aborted upgrade. Otherwise, the new columns on the table column would fail at startup. wdyt?

bereng: Yep I agree

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
adelapena: Makes sense, renamed to `rawMask`.

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
adelapena: Renamed to `rawMask`.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
adelapena: That's a great idea, it didn't occur to me. Changed to CQL strings.

adelapena: Just realized that we have methods to make the conversion from byte buffer column values to CQL literals (`AbstractType#toCQLString`, `CQL3Type#toCQLLiteral`), but we don't have methods for the opposite conversion. 

If we want `mask_argument_values` to be able to represent nulls we would need to add the reverse parsing, so the argument values are stored with quotes, escaping, etc. That would allow us to distinguish between, for example, `'null'` and `'''null'''`. That parsing is currently done in antlr and `Constants.Literal`. Bringing that parsing to `AbstractType` would probably require some not entirely trivial refactoring, given how the conversion functions are spread across abstract types, CQL types, serializers and antlr.

Not sure if that would be worth compared to the current approach, which is more focused on overcoming the limitations of lists than on how we represent values. I think that if we had support for nulls on lists we wouldn't store quoted CQL values but just nulls.

However, storing the values as strings seems a win for readability, so I'm changing that part to store the argument values as strings, but keeping the boolean list of nulls. wdyt?

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
adelapena: Just moved the masks to a separate table, so the new columns don't prevent downgrades.

blerer (approved):

src/java/org/apache/cassandra/cql3/functions/masking/ColumnMask.java
```diff
@@ -0,0 +1,231 @@
...
+        public Raw(FunctionName name, List<Term.Raw> rawPartialArguments)
+        {
+            this.name = name;
+            this.rawPartialArguments = rawPartialArguments;
+        }
+
+        public ColumnMask prepare(String keyspace, String table, ColumnIdentifier column, AbstractType<?> type)
+        {
+            ScalarFunction function = findMaskingFunction(keyspace, table, column, type);
+
```
bereng, line 172: nit: why the extra blank lines?

bereng:
Latest changes LGTM +1

src/java/org/apache/cassandra/cql3/functions/masking/ColumnMask.java
adelapena: I'll remove them on commit.

blerer (approved):

pmcfadin:
Closing as [CASSANDRA-18068](https://issues.apache.org/jira/browse/CASSANDRA-18068) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #2114, opened 2023-01-26
CASSANDRA-18204: CEP-15: (C*) Add git submodule for Accord

dcapwell:


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "accord"]
+	path = accord
+	url = https://github.com/dcapwell/cassandra-accord.git
```
dcapwell: using my PR branch, once that branch is merged this switches to apache

dev-support/git-hooks/post-checkout/100-update-submodules.sh
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-#
```
dcapwell: this file is 100% new, so GitHub thinking its a move is a bad way to review... should review the whole file and not this diff.

build.xml
```diff
@@ -283,10 +286,20 @@
...
+        <!-- Need to call clean as a version change with a jar that didn't have code changes will not procude a new jar with the new version... so need clean to make sure build is stable -->
+        <arg value="clean" />
+        <arg value="build" />
+        <arg value="install" />
+        <arg value="-x" />
+        <arg value="test" />
+        <arg value="-x" />
+        <arg value="rat" />
+        <arg value="-Paccord_group=org.apache.cassandra" />
+        <arg value="-Paccord_artifactId=cassandra-accord" />
```
dcapwell: TODO should these be a ant property?

build.xml
```diff
@@ -371,6 +384,7 @@
         <mkdir dir="${build.dir.lib}"/>
         <mkdir dir="${jacoco.export.dir}"/>
         <mkdir dir="${jacoco.partials.dir}"/>
+        <exec executable="${basedir}/dev-support/install-git-hooks.sh" osfamily="unix" dir="${basedir}" logError="true" failonerror="true" failifexecutionfails="true" />
```
dcapwell: this noop if the hooks are already installed, they check for existence and exit early

.gitmodules
dcapwell: I switched from trunk to this branch and had to do a few steps, ill need to document this for developers

dev-support/git-hooks/post-checkout/100-update-submodules.sh
```diff
@@ -24,35 +25,17 @@ set -o nounset
...
+  # In case the usage happens at a different layer, make sure to cd to the toplevel
+  local root_dir
+  root_dir="$(git rev-parse --show-toplevel)"
+  cd "$root_dir"
+
+  if [[ ! -e .gitmodules ]]; then
+    # nothing to see here, look away!
+    return 0
   fi
+  git submodule update --init --recursive
```
dcapwell: originally didn't have `--init` but this is needed if someone adds a new submodule... `--init` becomes a noop if the submodule is already init, so safe to include.

We don't need `--recursive` right now, but since the hooks no-op if they exist figured if we ever go that route best to add now...

michaelsembwever: this is failing for me, on my initial switch to the branch i'm getting.
```
❯ git sw accord-submodules
Updating files: 100% (3001/3001), done.
Switched to branch 'accord-submodules'
Your branch is up to date with 'dcapwell/accord-submodules'.
Cloning into '/Users/mick/src/apache/cassandra/accord'...
fatal: transport 'file' not allowed
fatal: Fetched in submodule path 'accord', but it did not contain fa73c60fd132b817fe594e8e42bdd1329628b626. Direct fetching of that commit failed.
```

if i first do (learnt from reading https://vielmetti.typepad.com/logbook/2022/10/git-security-fixes-lead-to-fatal-transport-file-not-allowed-error-in-ci-systems-cve-2022-39253.html )
```
❯ git config --global protocol.file.allow always
```
then the first fatal message is removed, but it still does not work…

michaelsembwever: just me. fetching first the `dcapwell` remote fixed it. (odd error msgs though)

michaelsembwever:
can we move the `dev-support/` contents into `.build/git-hooks/` …?

michaelsembwever:
Some testing…

1. the accord directory is trying to be deleted but not…
```
~/src/apache/casSANDRA ± accord-submodules
❯ git sw trunk
warning: unable to rmdir 'accord': Directory not empty
Switched to branch 'trunk'
Your branch is up to date with 'origin/trunk'.

~/src/apache/casSANDRA ± ● trunk
```

2. trying to figure out how to switch between branches that each have their own modifications…
```
❯ git sw -c mck/accord-submodules
Switched to a new branch 'mck/accord-submodules'

~/src/apache/casSANDRA ± mck/accord-submodules
❯ cd accord

~/src/apache/casSANDRA/accord ± remotes/origin/CASSANDRA-18204
❯ git sw -c mck/CASSANDRA-18204
Switched to a new branch 'mck/CASSANDRA-18204'

~/src/apache/casSANDRA/accord ± mck/CASSANDRA-18204
❯ git st
## mck/CASSANDRA-18204
 M accord-core/src/main/java/accord/primitives/Timestamp.java

~/src/apache/casSANDRA/accord ± ● mck/CASSANDRA-18204
❯ cd ..

~/src/apache/casSANDRA ± ● mck/accord-submodules
❯ git st
## mck/accord-submodules
 m accord

~/src/apache/casSANDRA ± ● mck/accord-submodules
❯ git sw accord-submodules
M	accord
Switched to branch 'accord-submodules'
Your branch is up to date with 'dcapwell/accord-submodules'.

~/src/apache/casSANDRA ± ● accord-submodules
❯ git st
## accord-submodules...dcapwell/accord-submodules
 m accord

~/src/apache/casSANDRA ± ● accord-submodules
❯ cd accord

~/src/apache/casSANDRA/accord ± ● mck/CASSANDRA-18204
❯ git st
## mck/CASSANDRA-18204
 M accord-core/src/main/java/accord/primitives/Timestamp.java
```
i would have thought the accord branch switches with the cassandra branch…?  
It didn't switch without the modifications either.

3. apache-cassandra-4.2-src.tar.gz contains class files

dev-support/git-hooks/post-checkout/100-update-submodules.sh
dcapwell: we spoke in slack, I sadly can't repo and @michaelsembwever and I have the same `git` version... tested different files named `accord` and didn't work..

I have 1 theory based off more testing, if C* commits an accord sha but that sha was never pushed up, then you can hit the last line....  to help prevent the cases I detected I added git hooks to make sure your SHA is pushed before you commit.

dev-support/git-hooks/post-switch
```diff
@@ -0,0 +1 @@
+post-checkout
```
dcapwell: I created a symlink so we don't have to maintain 2 copies

dev-support/git-hooks/pre-commit/100-verify-submodules-pushed.sh
```diff
@@ -0,0 +1,82 @@
...
+# Status Table
+# A         Added
+# C         Copied
+# D         Deleted
+# M         Modified
+# R         Renamed
+# T         Type Changed (i.e. regular file, symlink, submodule, …<200b>)
+# U         Unmerged
+# X         Unknown
+# B         Broken
```
dcapwell: dead code for now, I don't have a good reason to use, but I do like having this documented in a easy manner to check quickly.

dev-support/git-hooks/pre-commit/100-verify-submodules-pushed.sh
```diff
@@ -0,0 +1,82 @@
...
+_main() {
+  # In case the usage happens at a different layer, make sure to cd to the toplevel
+  local root_dir
+  root_dir="$(git rev-parse --show-toplevel)"
+  cd "$root_dir"
+
+  if [[ ! -e .gitmodules ]]; then
+    # nothing to see here, look away!
+    return 0
+  fi
```
michaelsembwever:
```suggestion
[[ ! -e .gitmodules ]] && return 0
```
🤷🏻‍♀️ (the style is used below, i find it easy enough to read)

.build/build-resolver.xml
```diff
@@ -165,7 +165,7 @@
         <property name="resolver-ant-tasks.initialized" value="true"/>
     </target>
 
-    <target name="resolver-retrieve-build" depends="resolver-init,write-poms">
+    <target name="resolver-retrieve-build" depends="resolver-init,write-poms,_build_subprojects">
```
michaelsembwever, line 168: i'm hesitant about this. i'm curious if it is better this is added one-level up to those that call `resolver-retrieve-build`.

there's the nit wrt encapsulation. but i'm more concerned about other scripts that call this target solely to cache the network downloading (e.g. our docker images). if they start doing actual building that's a bit confusing…

build.xml
```diff
@@ -283,10 +286,21 @@
...
+        <arg value="copyDependencies" /> <!-- For IDE support -->
+        <arg value="publishToMavenLocal" />
+        <arg value="-x" />
+        <arg value="test" />
+        <arg value="-x" />
+        <arg value="rat" />
+        <arg value="-Paccord_group=org.apache.cassandra" />
+        <arg value="-Paccord_artifactId=cassandra-accord" />
+        <arg value="-Paccord_version=${version}" />
+      </exec>
```
michaelsembwever: nit: what about moving all accord ant xml stuff into `.build/build-accord.xml` ? 

(there's two things happening in this PR: 1. git submodule stuff, and 2. accord stuff)

dev-support/git-hooks/pre-commit/100-verify-submodules-pushed.sh
```diff
@@ -0,0 +1,82 @@
...
+      git_sub_dir="${file}/.git"
+      branch="$(git --git-dir "${git_sub_dir}" rev-parse --abbrev-ref HEAD)"
+      if [[ "$branch" == "HEAD" ]]; then
+        # can't find the remote... so push all
+        git --git-dir "${git_sub_dir}" push --atomic
+      else
+        default_remote="$(git --git-dir "${git_sub_dir}" config --local --get branch."${branch}".remote || true)"
+        remote="${default_remote:-origin}"
+
+        git --git-dir "${git_sub_dir}" push --atomic "$remote" "$branch"
```
michaelsembwever: not sure about this auto pushing (here and in `pre-push/`)… 
but i'm not entirely clear on the intent of both scripts – can we add comments after the license explaining when and what and why the scripts are doing what they are doing…

.build/parent-pom-template.xml
```diff
@@ -675,14 +675,14 @@
...
+        <artifactId>cassandra-accord</artifactId>
+        <version>@version@</version>
       </dependency>
       <dependency>
-        <groupId>accord</groupId>
-        <artifactId>accord</artifactId>
-        <version>1.0-SNAPSHOT</version>
+        <groupId>org.apache.cassandra</groupId>
+        <artifactId>cassandra-accord</artifactId>
+        <version>@version@</version>
```
maedhroz, line 685: So for now we'll just keep this in lock-step w/ the C* version?

build.xml
```diff
@@ -283,10 +286,21 @@
         </fileset>
     </path>
 
-    <target name="_build-accord">
-      <echo unless:set="env.ACCORD_LOCAL_DEV">building Accord, set ACCORD_LOCAL_DEV environment variable to skip</echo>
-      <exec unless:set="env.ACCORD_LOCAL_DEV" executable=".build/include-accord.sh" dir="${basedir}" logError="true" failonerror="true" failifexecutionfails="true" />
-      <echo if:set="env.ACCORD_LOCAL_DEV">ACCORD_LOCAL_DEV environment variable is set, skipping pull/install</echo>
+    <target name="_build-accord" depends="init">
+      <exec executable="${basedir}/accord/gradlew" dir="${basedir}/accord" logError="true" failonerror="true" failifexecutionfails="true">
+        <!-- Need to call clean as a version change with a jar that didn't have code changes will not procude a new jar with the new version... so need clean to make sure build is stable -->
```
maedhroz:
```suggestion
        <!-- Need to call clean as a version change with a jar that didn't have code changes will not produce a new jar with the new version... so need clean to make sure build is stable -->
```

build.xml
```diff
@@ -481,6 +496,7 @@
     <!--
         The build target builds all the .class files
     -->
+    <target name="_build_subprojects" depends="_build-accord" description="Builds all subprojects needed for the build" />
```
maedhroz, line 485: Hopefully we can add the dtest-api to this soon :)

dev-support/install-git-defaults.sh
```diff
@@ -0,0 +1,115 @@
...
+
+_git_config_set() {
+  local -r name="$1"
+  # only care about rc
+  git config --local --get "$name" &> /dev/null
+}
+
+_install_configs() {
+  # when doing pull, this makes sure submodules are updated
+  _git_config_set cassandra.defaults-set && return 0
```
maedhroz: nit: Is setting the config items not idempotent?

dev-support/install-git-defaults.sh
```diff
@@ -0,0 +1,115 @@
...
+}
+
+_git_config_set() {
+  local -r name="$1"
+  # only care about rc
+  git config --local --get "$name" &> /dev/null
+}
+
+_install_configs() {
+  # when doing pull, this makes sure submodules are updated
```
maedhroz: nit: Minor thing, but I'd move this comment down one to the line it applies to.

dev-support/install-git-defaults.sh
```diff
@@ -0,0 +1,115 @@
...
+  local -r trigger_on_install=$3
+
+  mkdir "${d_dir}" &> /dev/null || true
+  local -r script_name="${hooks_dir}/${name}"
+  local installed=true
+  if [[ -e "$script_name" ]]; then
+    # was the script already installed?
+    if ! grep "This script is autogenerated by the Apache Cassandra build" "$script_name" &> /dev/null ; then
+      # user already has a script... move it!
+      #TODO should we bail and ask user to deal with this?
```
maedhroz: > #TODO should we bail and ask user to deal with this?

My vote would be to bail and make the user deal w/ it, rather than making the script more complicated (including inline documentation of the additional behaviors) to deal w/ it. (It seems like we'd mostly be expecting not to be in this state.)

maedhroz (approved):
There are a couple outstanding TODOs, but overall, this looks pretty good. We've tried to handle developer ergonomics and possible pitfalls in reasonable ways, not forced too much esoteric `git` knowledge onto people, etc.

CROSS-CUTTING-CONTRIBUTING.md
```diff
@@ -0,0 +1,46 @@
...
+
+# Creating a development environment
+
+When cross cutting changes are needed, it is best to work in a personal branch, but would need to replicate this to each submodule impacted.
+
+```
+$ dev-support/development-switch.sh --jira CASSANDRA-<number>
+```
+
+To commit changes, submodules most be updated first before committing those changes to Apache Cassandra.
```
smiklosovic: most -> must

CROSS-CUTTING-CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+# limitations under the License.
+#
+-->
+
+# Creating a development environment
+
+When cross cutting changes are needed, it is best to work in a personal branch, but would need to replicate this to each submodule impacted.
+
+```
+$ git checkout -b CASSANDRA-<number>
```
maedhroz: It's assumed we're starting in the project root here?

CROSS-CUTTING-CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+# Change the URL to point to your fork
+$ git submodule set-url accord "git@github.com:$USER/cassandra-accord.git"
+# Change the branch to your development branch
+$ (branch="$(git rev-parse --abbrev-ref HEAD)"; cd accord && git checkout -b "$branch")
+```
+
+To better automate this a utility script exists, so can do
+
+```
+$ dev-support/development-switch.sh --jira CASSANDRA-<number>
```
maedhroz: I know my GH user is different than my `$USER`. Might be worth just mentioning `--github-user` here.

dev-support/bump-accord.sh
```diff
@@ -0,0 +1,37 @@
...
+
+_main() {
+  git submodule status accord
+  echo "Is this the correct SHA? [y/n; default=y]"
+  read correct
+  if [[ "${correct:-y}" != "y" ]]; then
+    echo "Please update Accord's SHA and try again"
+    exit 1
+  fi
+  git commit -m "Change Accord to $(cd accord; git log -1  --format='%h: %B')" accord
```
maedhroz: Worth mentioning the pre-commit hook will push the submodule in a comment here?

CROSS-CUTTING-CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
+<!--
```
maedhroz: nit: For a title what about `WORKING-WITH-SUBMODULES.md` ?

CROSS-CUTTING-CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+# Creating a development environment
+
+When cross cutting changes are needed, it is best to work in a personal branch, but would need to replicate this to each submodule impacted.
```
maedhroz:
```suggestion
When changes to a submodule are needed, it is best to work in a personal branch, but would need to replicate this to each submodule impacted.
```
Does this make sense, given at least this first step will be pretty common even if there aren't real changes on the C* side?

CROSS-CUTTING-CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+-->
+
+# Creating a development environment
```
maedhroz:
```suggestion
# Local Development
```

CROSS-CUTTING-CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+
+```
+$ git checkout -b CASSANDRA-<number>
+# Change the URL to point to your fork
+$ git submodule set-url accord "git@github.com:$USER/cassandra-accord.git"
+# Change the branch to your development branch
+$ (branch="$(git rev-parse --abbrev-ref HEAD)"; cd accord && git checkout -b "$branch")
+```
+
+To better automate this a utility script exists, so can do
```
maedhroz:
```suggestion
You can also just use the following script:
```

CROSS-CUTTING-CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
...
+
+```
+$ dev-support/development-switch.sh --jira CASSANDRA-<number>
+```
+
```
maedhroz: At this point in the narrative, a couple things we could do:

1.) Make it clear for readers who aren't familiar w/ submodules that any commits they make to Accord will be visible to C* immediately, and local testing will be possible. (Let's say this ends the "Local Development" section.)

2.) We could start a new section, let's call it "Remote Testing", that covers `bump-accord.sh`, which, because of the pre-commit hook, starts pushing things to remotes. (Then pushing the C* bits makes remote testing on Circle or wherever possible, etc.)

CROSS-CUTTING-CONTRIBUTING.md
maedhroz: If you like it, there are the cases below too...

CROSS-CUTTING-CONTRIBUTING.md
michaelsembwever: `CONTRIBUTING-SUBMODULES.md` ? 

or just putting the content into the existing `CONTRIBUTING.md` ? 

🤷🏻‍♀️

CROSS-CUTTING-CONTRIBUTING.md
michaelsembwever: When changes to submodules

.build/build-resolver.xml
dcapwell: this is what I see

```
$ grep -r resolver-retrieve-build build.xml .build/*
build.xml:    <target name="build" depends="resolver-retrieve-build,build-project,checkstyle" description="Compile Cassandra classes"/>
build.xml:    <target name="_build_unsafe" depends="resolver-retrieve-build,build-project" description="Compile Cassandra classes without checks"/>
build.xml:  <target name="init-checkstyle" depends="resolver-retrieve-build,build-project" unless="no-checkstyle">
build.xml:  <target name="checkstyle-test" depends="init-checkstyle,resolver-retrieve-build,build-project" description="Run custom checkstyle code analysis on tests" if="java.version.8" unless="no-checkstyle">
.build/build-resolver.xml:    <target name="resolver-retrieve-build" depends="resolver-init,write-poms,_build_subprojects">
.build/build-resolver.xml:    <target name="resolver-dist-lib" depends="resolver-retrieve-build">
```

ill test this out

dcapwell: pushed the change will see how CI works; if everything is good will resolve this (and we can cherry pick which tasks do not depend)

CROSS-CUTTING-CONTRIBUTING.md
maedhroz: It's so small now, I'm fine w/ just putting it in `CONTRIBUTING.md`

CROSS-CUTTING-CONTRIBUTING.md
maedhroz: oops, smh

dev-support/git-hooks/pre-commit/100-verify-submodules-pushed.sh
dcapwell: ill add a comment, but ill also explain here.

If you are doing development in both git projects, you will create a sequence of SHAs.  Now, if you rewrite the accord history, you will have a C* commit pointing to an accord SHA that does not exist.   To make sure all commits can build, we push the accord changes up, so if anyone checks out C* it can build.  

This is why you had C* commits that failed to find a SHA, its because the sha was never pushed up to GitHub and history was rewritten removing it.

dcapwell: pushed a comment explaining why this is done, and improved the logging to show users how to disable.

.build/parent-pom-template.xml
dcapwell: correct, as this is the cassandra release.  We can migrate off building and use an already published jar later on if we want.

The one negative to this is that if we migrate to the jar, the group/artifact/versions are different, so this "may" impact users of -all jar.

build.xml
dcapwell: yep, this is why I went this route, to make life easier later on... been trying to make most of this generic with jvm-dtest in mind

dev-support/install-git-defaults.sh
dcapwell: setting should be idempotent but if a user disables, we should not override; so the case is `submodule.recurse = false`, we should not set it to `true`

dcapwell: now... do we need `cassandra.defaults-set`?  nah... ill remove

dev-support/install-git-defaults.sh
dcapwell: agree, I prefer fail fast, so I now log why I am failing and how to fix.

CROSS-CUTTING-CONTRIBUTING.md
dcapwell: yeah, that's how I wrote it... some things work just fine but never bothered tested if you do this in sub-directories

.build/git/install-git-defaults.sh
```diff
@@ -0,0 +1,117 @@
...
+  local -r d_dir="${hooks_dir}/${name}.d"
+  local -r trigger_on_install=$3
+
+  mkdir "${d_dir}" &> /dev/null || true
+  local -r script_name="${hooks_dir}/${name}"
+  local installed=true
+  if [[ -e "$script_name" ]]; then
+    # was the script already installed?
+    if ! grep "This script is autogenerated by the Apache Cassandra build" "$script_name" &> /dev/null ; then
+      echo "$script_name found, but was not generated by the Apache Cassandra build; please remove or move to ${d_dir}/000-original.sh; creating and moving to ${d_dir} will cause it to run as expected, but wont conflict with hooks this build adds" 1>&2
```
maedhroz:
```suggestion
      echo "$script_name found, but was not generated by the Apache Cassandra build; please remove or move to ${d_dir}/000-original.sh; creating and moving to ${d_dir} will cause it to run as expected, but won't conflict with hooks this build adds" 1>&2
```

.build/sh/development-switch.sh
```diff
@@ -0,0 +1,150 @@
...
+  # can it be inferred from remote?
+  local current_branch
+  current_branch="$(git rev-parse --abbrev-ref HEAD)"
+  local url
+  url="$(git config --local --get branch."${current_branch}".url || true)"
+  local inferred_name="$USER"
+  if [[ ! -z "${url:-}" ]]; then
+    inferred_name="$(echo "$url" | awk -F/ '{print $(NF-1)}' | awk -F: '{print $NF}')"
+  fi
+  echo "What is the github user/org used?  Inferred ${inferred_name}, enter new name or empty to accept" 1>&2
```
maedhroz:
```suggestion
  echo "What is the github user/org used?  Inferred ${inferred_name}, enter new name or empty to accept:" 1>&2
```

dev-support/git-hooks/pre-commit/100-verify-submodules-pushed.sh
dcapwell: I guess another option is to detect that local accord is further ahead than remote and block commit until resolved?  I could always make it a config so you choose which solution you like...

in development I like it pushing for me...

dev-support/bump-accord.sh
dcapwell: I don't feel so, its also something you can opt-out of...

CROSS-CUTTING-CONTRIBUTING.md
dcapwell: I didn't as I don't think it's the avg case.  The script now asks for your GH user, so `--github-user` now only exists for cases where you want to override, such as if I want to work on Blakes project I do `bdeggleston` (though https will have issues...)

CROSS-CUTTING-CONTRIBUTING.md
dcapwell: will move there

CROSS-CUTTING-CONTRIBUTING.md
dcapwell: > that any commits they make to Accord will be visible to C* immediately

I guess I assumed that by us having it in the repo... why have it in the repo if you can't see it?

Do we need to call out `Remote Testing`?  The defaults are setup to work in Circle/Jenkins out of the box, so there is nothing to actually do to enable; you push like normal and CI runs like normal.

>  because of the pre-commit hook, starts pushing things to remotes

I feel like that is tangental, we push to avoid bad commits to Cassandra... If we don't push you could create a commit that points to a bad accord SHA.  We *could* document this, but is this something people need to think about?  There is a very specific edge case I see and that is if an committer is committing to `trunk`... they might not know that they try to push that change to `apache` (unless they updated to their repo)... since people tend to work in feature branches, I am not sure how much of a concern this should be

CROSS-CUTTING-CONTRIBUTING.md
belliottsmith: https://stackoverflow.com/questions/40841882/automatically-access-git-submodules-via-ssh-or-https

Not sure if this simplifies things, but if we say the module must reside in the same source server as the c* branch referencing it we could use relative URLs and perhaps solve the user and https problem…

PR #2117, opened 2023-01-30
Add a virtual table to list snapshots for CASSANDRA-18102

Maxwell-Guo:
Add a virtual table to list snapshots for CASSANDRA-18102

```
patch by maxwellguo; reviewed by <Reviewers> for CASSANDRA-18102
```

The [CASSANDRA-18102](https://issues.apache.org/jira/browse/CASSANDRA-18102)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 5.0
+ * Add MaxSSTableSize and MaxSSTableDuration metrics and propagate them together with local read/write ratio to tablestats (CASSANDRA-18283)
  * Add more logging around CompactionManager operations (CASSANDRA-18268)
  * Reduce memory allocations of calls to ByteBufer.duplicate() made in org.apache.cassandra.transport.CBUtil#writeValue (CASSANDRA-18212)
  * CEP-17: SSTable API (CASSANDRA-17056)
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -126,7 +126,10 @@ New features
       normally require it, except `system_views.system_logs`.
     - More accurate skipping of sstables in read path due to better handling of min/max clustering and lower bound;
       SSTable format has been bumped to 'nc' because there are new fields in stats metadata\
-    - SSTal
+    - Added MaxSSTableSize and MaxSSTableDuration metrics to TableMetrics. The former returns the size of the biggest 
+      SSTable of a table or 0 when there is not any SSTable. The latter returns the maximum duration, computed as 
+      `maxTimestamp - minTimestamp`, effectively non-zero for SSTables produced by TimeWindowCompactionStrategy.
+    - Added local read/write ratio to tablestats.
 
 Upgrading
 ---------
--- a/doc/modules/cassandra/pages/operating/metrics.adoc
+++ b/doc/modules/cassandra/pages/operating/metrics.adoc
@@ -144,6 +144,13 @@ this table (in bytes).
 |TotalDiskSpaceUsed |Counter |Total disk space used by SSTables
 belonging to this table, including obsolete ones waiting to be GC'd.
 
+|MaxSSTableSize |Gauge<Long> |Maximum size of SSTable of this table -
+the physical size on disk of all components for such SSTable in bytes. Equals to
+zero if there is not any SSTable on disk.
+
+|MaxSSTableDuration |Gauge<Long> |Maximum duration in milliseconds of an SSTable for this table,
+computed as `maxTimestamp - minTimestamp`. Equals to zero if min or max timestamp is `Long.MAX_VALUE`.
+
 |MinPartitionSize |Gauge<Long> |Size of the smallest compacted partition
 (in bytes).
 
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -3436,6 +3436,18 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean, Memtable.Owner
     }
 
     @Override
+    public long getMaxSSTableSize()
+    {
+        return metric.maxSSTableSize.getValue();
+    }
+
+    @Override
+    public long getMaxSSTableDuration()
+    {
+        return metric.maxSSTableDuration.getValue();
+    }
+
+    @Override
     public Map<String, Long> getTopSizePartitions()
     {
         if (topPartitions == null)
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStoreMBean.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStoreMBean.java
@@ -299,4 +299,23 @@ public interface ColumnFamilyStoreMBean
     public Long getTopSizePartitionsLastUpdate();
     public Map<String, Long> getTopTombstonePartitions();
     public Long getTopTombstonePartitionsLastUpdate();
+
+    /**
+     * Returns the size of the biggest SSTable of this table.
+     *
+     * @return (physical) size of the biggest SSTable of this table on disk or 0 if no SSTable is present
+     */
+    public long getMaxSSTableSize();
+
+    /**
+     * Returns the longest duration of an SSTable, in milliseconds, of this table,
+     * computed as {@code maxTimestamp - minTimestamp}.
+     *
+     * It returns 0 if there are no SSTables or if {@code maxTimestamp} or {@code minTimestamp} is
+     * equal to {@code Long.MAX_VALUE}. Effectively non-zero for tables on {@code TimeWindowCompactionStrategy}.
+     *
+     * @return the biggest {@code maxTimestamp - minTimestamp} among all SSTables of this table
+     * or 0 if no SSTable is present
+     */
+    public long getMaxSSTableDuration();
 }
--- a/src/java/org/apache/cassandra/db/compaction/TimeWindowCompactionStrategyOptions.java
+++ b/src/java/org/apache/cassandra/db/compaction/TimeWindowCompactionStrategyOptions.java
@@ -39,11 +39,11 @@ public final class TimeWindowCompactionStrategyOptions
     protected static final int DEFAULT_EXPIRED_SSTABLE_CHECK_FREQUENCY_SECONDS = 60 * 10;
     protected static final Boolean DEFAULT_UNSAFE_AGGRESSIVE_SSTABLE_EXPIRATION = false;
 
-    protected static final String TIMESTAMP_RESOLUTION_KEY = "timestamp_resolution";
-    protected static final String COMPACTION_WINDOW_UNIT_KEY = "compaction_window_unit";
-    protected static final String COMPACTION_WINDOW_SIZE_KEY = "compaction_window_size";
-    protected static final String EXPIRED_SSTABLE_CHECK_FREQUENCY_SECONDS_KEY = "expired_sstable_check_frequency_seconds";
-    protected static final String UNSAFE_AGGRESSIVE_SSTABLE_EXPIRATION_KEY = "unsafe_aggressive_sstable_expiration";
+    public static final String TIMESTAMP_RESOLUTION_KEY = "timestamp_resolution";
+    public static final String COMPACTION_WINDOW_UNIT_KEY = "compaction_window_unit";
+    public static final String COMPACTION_WINDOW_SIZE_KEY = "compaction_window_size";
+    public static final String EXPIRED_SSTABLE_CHECK_FREQUENCY_SECONDS_KEY = "expired_sstable_check_frequency_seconds";
+    public static final String UNSAFE_AGGRESSIVE_SSTABLE_EXPIRATION_KEY = "unsafe_aggressive_sstable_expiration";
 
     static final String UNSAFE_AGGRESSIVE_SSTABLE_EXPIRATION_PROPERTY = Config.PROPERTY_PREFIX + "allow_unsafe_aggressive_sstable_expiration";
 
--- a/src/java/org/apache/cassandra/metrics/TableMetrics.java
+++ b/src/java/org/apache/cassandra/metrics/TableMetrics.java
@@ -126,6 +126,10 @@ public class TableMetrics
     public final Gauge<Integer> liveSSTableCount;
     /** Number of SSTables with old version on disk for this CF */
     public final Gauge<Integer> oldVersionSSTableCount;
+    /** Maximum duration of an SSTable for this table, computed as maxTimestamp - minTimestamp*/
+    public final Gauge<Long> maxSSTableDuration;
+    /** Maximum size of SSTable of this table - the physical size on disk of all components for such SSTable in bytes*/
+    public final Gauge<Long> maxSSTableSize;
     /** Disk space used by SSTables belonging to this table */
     public final Counter liveDiskSpaceUsed;
     /** Uncompressed/logical disk space used by SSTables belonging to this table */
@@ -631,6 +635,35 @@ public class TableMetrics
                     if (!sstable.descriptor.version.isLatestVersion())
                         count++;
                 return count;
+            }
+        });
+        maxSSTableDuration = createTableGauge("MaxSSTableDuration", new Gauge<Long>()
+        {
+            @Override
+            public Long getValue()
+            {
+                return cfs.getTracker()
+                          .getView()
+                          .liveSSTables()
+                          .stream()
+                          .filter(sstable -> sstable.getMinTimestamp() != Long.MAX_VALUE && sstable.getMaxTimestamp() != Long.MAX_VALUE)
+                          .map(ssTableReader -> ssTableReader.getMaxTimestamp() - ssTableReader.getMinTimestamp())
+                          .max(Long::compare)
+                          .orElse(0L);
+            }
+        });
+        maxSSTableSize = createTableGauge("MaxSSTableSize", new Gauge<Long>()
+        {
+            @Override
+            public Long getValue()
+            {
+                return cfs.getTracker()
+                          .getView()
+                          .liveSSTables()
+                          .stream()
+                          .map(SSTableReader::bytesOnDisk)
+                          .max(Long::compare)
+                          .orElse(0L);
             }
         });
         liveDiskSpaceUsed = createTableCounter("LiveDiskSpaceUsed");
--- a/src/java/org/apache/cassandra/tools/NodeProbe.java
+++ b/src/java/org/apache/cassandra/tools/NodeProbe.java
@@ -1802,6 +1802,8 @@ public class NodeProbe implements AutoCloseable
                 case "EstimatedPartitionCount":
                 case "KeyCacheHitRate":
                 case "LiveSSTableCount":
+                case "MaxSSTableDuration":
+                case "MaxSSTableSize":
                 case "OldVersionSSTableCount":
                 case "MaxPartitionSize":
                 case "MeanPartitionSize":
--- a/src/java/org/apache/cassandra/tools/nodetool/TableStats.java
+++ b/src/java/org/apache/cassandra/tools/nodetool/TableStats.java
@@ -60,7 +60,8 @@ public class TableStats extends NodeToolCmd
                         + "memtable_off_heap_memory_used, memtable_switch_count, number_of_partitions_estimate, "
                         + "off_heap_memory_used_total, pending_flushes, percent_repaired, read_latency, reads, "
                         + "space_used_by_snapshots_total, space_used_live, space_used_total, "
-                        + "sstable_compression_ratio, sstable_count, table_name, write_latency, writes)")
+                        + "sstable_compression_ratio, sstable_count, table_name, write_latency, writes, " +
+                          "max_sstable_size, local_read_write_ratio, twcs_max_duration)")
     private String sortKey = "";
 
     @Option(title = "top",
--- a/src/java/org/apache/cassandra/tools/nodetool/stats/StatsTable.java
+++ b/src/java/org/apache/cassandra/tools/nodetool/stats/StatsTable.java
@@ -31,6 +31,7 @@ public class StatsTable
     public boolean isLeveledSstable = false;
     public Object sstableCount;
     public Object oldSSTableCount;
+    public Long maxSSTableSize;
     public String spaceUsedLive;
     public String spaceUsedTotal;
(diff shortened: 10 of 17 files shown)


src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -1848,7 +1848,7 @@ public void run()
                 }
             }
         };
-
+
```
smiklosovic: this is not necessary, please return it back as it was.

src/java/org/apache/cassandra/db/virtual/SnapshotsTable.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final String TRUE_SIZE = "true_size";
+    private static final String SIZE_ON_DISK = "size_on_disk";
+    private static final String CREATE_TIME = "created_at";
+    private static final String EXPIRATION_TIME = "expires_at";
+    private static final String EPHEMERAL = "ephemeral";
+    
+    SnapshotsTable(String keyspace)
+    {
+        super(TableMetadata.builder(keyspace, "snapshots")
+                           .comment("tables snapshots")
```
smiklosovic: I would change this to "available snapshots" or "current snapshots" or "present snapshots"

src/java/org/apache/cassandra/db/virtual/SnapshotsTable.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+  
+    @Override
+    public DataSet data() 
+    {
+        SimpleDataSet result = new SimpleDataSet(metadata());
+        
+        // include snapshots with ttl and ephemeral
+        Map<String, TabularData> snapshotDetails = StorageService.instance.getSnapshotDetails(ImmutableMap.of("no_ttl", "false", "include_ephemeral", "true"));
+        for(Map.Entry<String, TabularData> entry : snapshotDetails.entrySet())
```
smiklosovic: `for (`

src/java/org/apache/cassandra/db/virtual/SnapshotsTable.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final String EXPIRATION_TIME = "expires_at";
+    private static final String EPHEMERAL = "ephemeral";
+    
+    SnapshotsTable(String keyspace)
+    {
+        super(TableMetadata.builder(keyspace, "snapshots")
+                           .comment("tables snapshots")
+                           .kind(TableMetadata.Kind.VIRTUAL)
+                           .partitioner(new LocalPartitioner(UTF8Type.instance))
+                           .addPartitionKeyColumn(SNAPSHOT_NAME, UTF8Type.instance)
```
smiklosovic: I would change the key. Partition key would be keyspace and clustering column would be table and name of the snapshot. That way, you can ask:

````
use system_views;
select * from snapshots;
select * from snapshots where keyspace = 'ks1';
select * from snapshots where keyspace = 'ks1' and table = 'tb1';
select * from snapshots where keyspace = 'ks1' and table = 'tb1' and name = "mysnapshot";
select * from snapshots where name = "mysnapshot" allow filtering;
````

In practice, I think it is more common to ask questions like "what snapshots this keyspace / table" has? Instead of querying that table by name of the particular snapshot. What if I do not know the name of the snapshot in advance?

@pauloricardomg what do you think? I think this is the most important aspect of this PR.

File: src/java/org/apache/cassandra/db/virtual/SnapshotsTable.java [hunk 5]
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.cassandra.db.marshal.BooleanType;
+import org.apache.cassandra.db.marshal.LongType;
+import org.apache.cassandra.db.marshal.UTF8Type;
+import org.apache.cassandra.dht.LocalPartitioner;
+import org.apache.cassandra.schema.TableMetadata;
+import org.apache.cassandra.service.StorageService;
+
+public class SnapshotsTable extends AbstractVirtualTable
+{
+    private static final String SNAPSHOT_NAME = "snapshot_name";
```
smiklosovic: isnt "name" enough?

src/java/org/apache/cassandra/db/virtual/SnapshotsTable.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final String KEYSPACE_NAME = "keyspace_name";
```
smiklosovic: isnt "keyspace" enough?

src/java/org/apache/cassandra/db/virtual/SnapshotsTable.java
```diff
@@ -0,0 +1,93 @@
...
+    private static final String COLUMNFAMILY_NAME = "columnfamily_name";
```
smiklosovic: I would rename "columnfamily" to "table" so "columnfamily_name" to "table" only?

smiklosovic (changes requested):

src/java/org/apache/cassandra/db/virtual/SnapshotsTable.java
Maxwell-Guo: +1  on what you said

src/java/org/apache/cassandra/db/virtual/SnapshotsTable.java
Maxwell-Guo: I hope that the  selectstatement' output result of snapshot virtual table is similar to that of listsnapshot, otherwise the user may have some confusion.
The description of snapshot name is "snapshot_name", "keyspace_name", "columnfamilt_name" and so on.
Though I think table or table name is better than "columnfamily_name" as what you said below.

src/java/org/apache/cassandra/db/virtual/SnapshotsTable.java
Maxwell-Guo: same with the upper comment

src/java/org/apache/cassandra/db/virtual/SnapshotsTable.java
Maxwell-Guo: same with the upper comment

src/java/org/apache/cassandra/db/virtual/SnapshotsTable.java
smiklosovic: hi @Maxwell-Guo , any chance to rework this as discussed above?

Maxwell-Guo: hi @smiklosovic ，thanks so much. Busy these days so I feel sorry for the delay in replying to the message.
I have already finished the above work.
Besides I hope you can take a look at the reply of naming below.Thanks

smiklosovic: @pauloricardomg what do you think about column names? Should we follow the naming in nodetool's output? I think we should not be tied by it.

pauloricardomg: Sorry for the delay - have been busy in the past couple of weeks but will try to be faster on next iterations so we can wrap this up soon.

I agree with with @smiklosovic comment about using more succint column names (ie. `keyspace` vs `keyspace_name` and `table` vs `columnfamily_name`), but I think we should use the column name "tag" and not "name" to refer to a snapshot identifier, and keep this as table partition key. We should enable allow filtering on this virtual table (with [CASSANDRA-18271](https://issues.apache.org/jira/browse/CASSANDRA-18271)) to allow querying by non partition keys (ie. `keyspace` or `table`). This would allow the following queries:

```sql
use system_views;
select * from snapshots;
select * from snapshots where tag = "mysnapshot" # query by partition key
select * from snapshots where keyspace = 'ks1'; # query by non-partition keys
select * from snapshots where keyspace = 'ks1' and table = 'tb1';
select * from snapshots where keyspace = 'ks1' and table = 'tb1' and tag = "mysnapshot";
```

What do you think?

pauloricardomg: > I think we should use the column name "tag" and not "name" to refer to a snapshot identifier

Even though we have been using `tag` to refer to snapshot ids, this name is a bit misleading since a resource could in theory have multiple tags (but just one ID). @smiklosovic I wonder if it would make sense to just call it snapshot `id` and update internal usages to refer to snapshots by `id` and not `name/tag` ?

Updated queries with `id` colum name for snapshot identifier:
```sql
use system_views;
select * from snapshots;
select * from snapshots where id = "mysnapshot" # query by partition key
select * from snapshots where keyspace = 'ks1'; # query by non-partition keys
select * from snapshots where keyspace = 'ks1' and table = 'tb1';
select * from snapshots where keyspace = 'ks1' and table = 'tb1' and id = "mysnapshot";
```

Maxwell-Guo: Hi @pauloricardomg  and @smiklosovic 
I found that the word keyspace is a reserved word for cql 
https://github.com/apache/cassandra/blob/trunk/src/antlr/Parser.g#L1898
So and error will occurs when doing select when specified keyspace = xxx or select keyspace  where xxx
I think that's why other system table using keyspace_name instand of keyspace
https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/db/SystemKeyspace.java#L238

And I am ok if we change snapshot_name from name to tag, And i have already finish this.

smiklosovic: yes it is very unfortunate that keyspace is a reserved word. It may be workarounded by 'keyspace' (putting that to quotes) but that is quite offputting and not intuitive for users to deal with this ... So next best thing is "keyspace_name".

Or we fix cql grammar to not fail on that.

I dont object using 'id' but doing that internally might be dedicated to a separate ticket for the sake of simplicity in this ticket.

Filtering will be enabled by default on vtables when not explicitly disallowed so nothing to cover there.

Maxwell-Guo: Thanks @smiklosovic , So what about using keyspace_name and table_name ? Because I found that some of the system tables just using keyspace_name and table_name , although virtual table is not a real system table, it is similar in a sense. 

I think open a new ticket for change the tag/name to id internally is ok, if we all agree on this I can open a ticket for this and finish the task.
@pauloricardomg  WDYT?

smiklosovic: I think it is fine to have keyspace_name and table_name.

smiklosovic: Please change the code to include my suggestion I posted above:

_Partition key would be `keyspace_name` and clustering columns would be `table_name` and `id` of the snapshot._

pauloricardomg: > Partition key would be keyspace_name and clustering columns would be table_name and id of the snapshot.

Why do you think they should be primary keys?  I think primary key of snapshot should be `id`, since this is the primary snapshot identifier (ie. different keyspaces cannot have the same snapshot id). It's still possible to query by `keyspace_name` and `table_name` if needed after CASSANDRA-18238.

Maxwell-Guo: In my mind ,I think the scope of snapshot name is bigger than keyspace as different keyspace may have same snapshot name, so first time I treate id as partition key and the output format is same with snapshot detail information.But latter I am +1 on @smiklosovic , for the first perspective of a user may be the keyspace, and they want to know a certain snapshot under the keyspace when writing a cql.I want to make sure it's what you think。 @smiklosovic

smiklosovic: @pauloricardomg 

Regardless of whether we can query without allow filtering after 18238 I think it is still a good practice to act as we would normally do when modeling the schema.

As explained above and what Maxwell just explained, I think keyspace > table > snapshot id.

We can have same snapshot names after all, no? So if keyspace and table are clustering columns, with two snapshots of the same name, we would have this partition:

    snapshotName1 | keyspace1 | table1
    snapshotName1 | keyspace2 | table2

primary key would be `((snapshotName), keyspaceName, tableName)`

Is not this counter-intuitive to have a partition which is logically coupling different keyspaces and names under the same snapshot name for logically different snapshots? That does not make sense to me. 

However, on the other hand, it is possible to do this:

    ./bin/nodetool snapshot --kt-list ks.tb,system.local -t mysnapshot

So `listsnapshots` will do this:

    mysnapshot  system  local  1.16 KiB  21.47 KiB    2023-03-02T13:19:13.757Z
    mysnapshot  ks      tb     1.02 KiB   6.08 KiB    2023-03-02T13:19:13.757Z

But then I can do this as well, again:

    ./bin/nodetool snapshot --kt-list ks.tb2 -t mysnapshot

Which would print it like:

    mysnapshot     ks       tb2        1.16 KiB  21.47 KiB    2023-03-02T13:19:42.140Z                
    mysnapshot     ks       tb         1.02 KiB  6.08 KiB     2023-03-02T13:19:13.757Z                
    mysnapshot     system   local      107 bytes 6.98 KiB     2023-03-02T13:19:13.757Z 

So, with the primary key `((id), keyspace, table)`, the advantage is that we would be able to visually see what all tables were snapshotted in that one logical snapshot based on the same timestamp. Here, we see, from the timestamp, that ks.tb and system.local were snapshotted "together".

So from this perspective it is better if snapshot id is partition key.

I would go so far to include timestamp into primary key as well: `((id), keyspace, table, timestamp)`

This way we would have them ordered too and it does not need to be specified when querying.

If we made it like already suggested: (keyspace, table, snapshotid), we would lose the information, it might be like:

    ks1 tb1 snapshot1 2023-03-02T13:19:42.757Z        
    ks3 tb2 snapshot1 2023-03-02T13:19:13.757Z 
    ks3 tb3 snapshot1 2023-03-02T13:19:42.757Z 
    ks4 tb4 snapshot1 2023-03-02T13:19:13.757Z 

But here it is not so simple to see that "ks4.tb4" and "ks3.tb2" are forming one logical snapshot and then "ks1.tb1" and "ks3.tb3" another.

So, utimately, it should be like this:

((id), ts, ks, tb))

Timestamp has to go first to have it ordered, same timestamp for some snapshot id would mean that all keyspaces and tables belong to that particular snapshot name which was taken at that exact time.

    cqlsh> select * from ks.test ;
     id | ts | ks | tb
    ----+----+----+----
      1 |  2 |  3 |  4      first snapshot
      1 |  2 |  3 |  5      first snapshot
      1 |  2 |  3 |  6      first snapshot
      1 |  3 |  1 |  2      second snapshot
      1 |  3 |  1 |  3      second snapshot
      2 |  5 |  2 |  6      third snapshot

pauloricardomg: > Is not this counter-intuitive to have a partition which is logically coupling different keyspaces and names under the same snapshot name for logically different snapshots? That does not make sense to me.

Agreed, thanks for the clarification.

> But then I can do this as well, again: `./bin/nodetool snapshot --kt-list ks.tb2 -t mysnapshot`

Even though this is allowed I think this is an inconsistency between [takeSnapshot](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/service/StorageService.java#L4162) and [takeMultipleTableSnapshot](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/service/StorageService.java#L4208). In the first method [it's not possible](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/service/StorageService.java#L4183) to create a snapshot if a snapshot with the same id/tag already exists in any keyspace, while in the second method it's only checked if a snapshot with the same name does not exist [in the same table](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/service/StorageService.java#L4235).

I believe the behavior of `takeSnapshot` is the correct one, because it was implemented first and avoids confusion if there are multiple snapshots with the same name taken at different times for different tables. I think we should unify these methods and make this consistent on [CASSANDRA-18271](https://issues.apache.org/jira/browse/CASSANDRA-18271), so it's only possible to have a single logical snapshot with the same id.

If we do that, we should make the partition key `((snapshotId), keyspaceName, tableName)` and not include the timestamp in the primary key, since logical snapshots taken at the same time will be grouped together by id - in this case only "legacy" snapshots (taken before  [CASSANDRA-18271](https://issues.apache.org/jira/browse/CASSANDRA-18271)) will be unordered which is not a big deal. WDYT?

Maxwell-Guo: If so,It seems that there is no need for me to change the code logic ?

smiklosovic: @pauloricardomg I think we should bring this to ML as we are changing the behavior suddenly and some snapshots will not be possible anymore (snapshots with same name).

pauloricardomg: ok I will send a message shortly to the ML to see what others think. In any case we should make the behavior consistent whatever we agree to (either allow `takeSnapshot` to take snapshot with existing id on different table, or restrict `takeMultipleTableSnapshot` to not allow snapshot with existing id).

test/unit/org/apache/cassandra/db/virtual/SnapshotsTableTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+        // query all from snapshots virtual table
+        UntypedResultSet result = execute("SELECT id, keyspace_name, table_name, created_at, expires_at, ephemeral FROM vts.snapshots");
+        assertRows(result,
+                row(SNAPSHOT_EPHEMERAL, CQLTester.KEYSPACE, currentTable(), createTimeStr, null, true),
+                row(SNAPSHOT_NO_TTL, CQLTester.KEYSPACE, currentTable(), createTimeStr, null, false),
+                row(SNAPSHOT_TTL, CQLTester.KEYSPACE, currentTable(), createTimeStr, createTime.plusSeconds(ttl.toSeconds()).toString(), false));
+
+        // query with conditions
+        result = execute("SELECT id, keyspace_name, table_name, created_at, expires_at, ephemeral FROM vts.snapshots where ephemeral = true allow filtering");
```
smiklosovic: `allow filtering` is not necessary

test/unit/org/apache/cassandra/db/virtual/SnapshotsTableTest.java
```diff
@@ -0,0 +1,103 @@
...
+        assertRows(result,
+                row(SNAPSHOT_EPHEMERAL, CQLTester.KEYSPACE, currentTable(), createTimeStr, null, true));
+
+        result = execute("SELECT id, keyspace_name, table_name, created_at, expires_at, ephemeral FROM vts.snapshots where id = ? allow filtering", SNAPSHOT_TTL);
```
smiklosovic: `allow filtering` is not necessary

test/unit/org/apache/cassandra/db/virtual/SnapshotsTableTest.java
```diff
@@ -0,0 +1,103 @@
...
+public class SnapshotsTableTest extends CQLTester
+{
+    private static final String KS_NAME = "vts";
+    private final String SNAPSHOT_TTL = "snapshotTtl";
+    private final String SNAPSHOT_NO_TTL = "snapshotNoTtl";
+    private final String SNAPSHOT_EPHEMERAL = "snapshotEphemeral";
+    private final DurationSpec.IntSecondsBound ttl = new DurationSpec.IntSecondsBound ("4h");
+    
+    @BeforeClass
+    public static void setUpClass()
```
smiklosovic: is this method really necessary? Is not it executed anyway?

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -680,6 +684,60 @@ public void cleanupCache()
...
+    public void snapshot(String snapshotName, boolean ephemeral, boolean skipMemtable, DurationSpec.IntSecondsBound ttl, RateLimiter rateLimiter, Instant creationTime)
+    {
+        ColumnFamilyStore store = getCurrentColumnFamilyStore();
+        if(store != null)
+        {
+            store.snapshot(snapshotName, null, ephemeral, skipMemtable, ttl, rateLimiter, creationTime);
+        }
+    }
+    
+    public void clearSnapshot(String keyspace)
```
smiklosovic: `clearSnapshotForKeyspace` is better, I would think that paramter for this method is name of the snapshot, not name of keyspace.

src/java/org/apache/cassandra/db/virtual/SnapshotsTable.java
smiklosovic: @pauloricardomg  based on the thread in ML it seems like we are ok with what is here. I put some more comments to PR, @Maxwell-Guo , could you please address them? So we might approach the CI.

Maxwell-Guo: thanks @smiklosovic @pauloricardomg , I will fix them later today.

test/unit/org/apache/cassandra/cql3/CQLTester.java
Maxwell-Guo: This method ("clearSnapshot(String keyspace)")is not used, what I used is "void clearSnapshot(String snapshotName, String keyspace)" ,but I think some one may need it in the future so I left it here.
Besides I think we can also implement some other method like  clearSnapshot(String id) in the future too.

test/unit/org/apache/cassandra/db/virtual/SnapshotsTableTest.java
Maxwell-Guo: It's needed  here when  whereclause is ephemeral = true , ephemeral is not partition key or clustering key

smiklosovic: @Maxwell-Guo you really do not need that. Please check CASSANDRA-18238.

Maxwell-Guo: It seems that I should pull to get the newest code

Maxwell-Guo:
I am so sorry that this pr is discard due to an operation error, And I have create a new one . https://github.com/apache/cassandra/pull/2205

PR #2119, opened 2023-01-30
CASSANDRA-17561 rebased against trunk

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/repair/consistent/SyncStatSummary.java
+++ b/src/java/org/apache/cassandra/repair/consistent/SyncStatSummary.java
@@ -21,13 +21,18 @@ package org.apache.cassandra.repair.consistent;
 import java.net.InetSocketAddress;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 
 import com.google.common.collect.Lists;
 
+import org.apache.cassandra.dht.Range;
+import org.apache.cassandra.dht.Token;
 import org.apache.cassandra.repair.RepairResult;
 import org.apache.cassandra.repair.RepairSessionResult;
 import org.apache.cassandra.repair.SyncStat;
@@ -50,7 +55,7 @@ public class SyncStatSummary
 
         int files = 0;
         long bytes = 0;
-        long ranges = 0;
+        Set<Range<Token>> ranges = new HashSet<>();
 
         Session(InetSocketAddress src, InetSocketAddress dst)
         {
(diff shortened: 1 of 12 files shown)


src/java/org/apache/cassandra/repair/consistent/SyncStatSummary.java
```diff
@@ -147,21 +152,36 @@ public String toString()
             }
             StringBuilder output = new StringBuilder();
 
-            output.append(String.format("%s.%s - %s ranges, %s sstables, %s bytes\n", keyspace, table, ranges, files, FBUtilities.prettyPrintMemory(bytes)));
+            output.append(String.format("%s.%s - %s ranges, %s sstables, %s bytes\n", keyspace, table, ranges.size(), files, FBUtilities.prettyPrintMemory(bytes)));
```
beobal, line 155: I'm unclear what the process is for modifying output of nodetool subcommands, but this will potentially change the console output of `nodetool repair`. IMO external tooling shouldn't be depending on the exact strings that nodetool emits, so this is fine with me but are we ok with having no option to force it to the old output?

src/java/org/apache/cassandra/utils/RangesSerializer.java
```diff
@@ -0,0 +1,74 @@
...
+    public static final RangesSerializer serializer = new RangesSerializer();
+
+    @Override
+    public void serialize(Collection<Range<Token>> ranges, DataOutputPlus out, int version) throws IOException
+    {
+        out.writeInt(ranges.size());
+        for (Range<Token> r : ranges)
+        {
+            Token.serializer.serialize(r.left, out, version);
+            Token.serializer.serialize(r.right, out, MessagingService.current_version);
```
beobal: Not that it currently matters as the token serializer disregards `version` but this looks like a mistake?

beobal (changes requested):
Mostly LGTM: 1 question, 1 minor issue.

src/java/org/apache/cassandra/repair/consistent/SyncStatSummary.java
smiklosovic: @beobal good point in general but I think this change is justifiable. It seems to be like overkill to me to offer the old format too.

src/java/org/apache/cassandra/utils/RangesSerializer.java
smiklosovic: @beobal you mean like this line is redundant? There is the very same thing on line 46 so this seems to be a leftover from coding ...
 
edit: ah sorry, there is `r.right`, so you say it should be like this?

    Token.serializer.serialize(r.right, out, version);

beobal: yep, I mean I don't see a reason to hardcode the version here.

src/java/org/apache/cassandra/repair/consistent/SyncStatSummary.java
beobal: I agree, just wanted to mention it in case we are in the minority on this.

src/java/org/apache/cassandra/utils/RangesSerializer.java
smiklosovic: done.

PR #2120, opened 2023-01-30
 CASSANDRA-18195 Feature Flag for Accord Transactions

maedhroz:
patch by Caleb Rackliffe; reviewed by David Capwell for CASSANDRA-18195


maedhroz:
https://app.circleci.com/pipelines/github/maedhroz/cassandra?branch=CASSANDRA-18195

maedhroz:
This patch adds a `cassandra.yaml` flag that, by default, does the following:

1.) At startup, prevents...
a.) ...Accord virtual tables from being created
b.) ...the Accord system keyspace/tables from being created
c.) ...the `AccordService` from being initialized, along w/ the threads it owns. (The implementation involving a no-op `IAccordService` might be up for discussion, but the idea seems to have been agreed to on the Jira already.)
d.) ...operators from enabling Accord-based CAS operations while Accord itself is disabled.

2.) In general, prevents any CQL Accord transaction from executing.

src/java/org/apache/cassandra/service/accord/IAccordService.java
```diff
@@ -0,0 +1,46 @@
...
+
+    void createEpochFromConfigUnsafe();
+
+    TxnData coordinate(Txn txn, ConsistencyLevel consistencyLevel);
+
+    long currentEpoch();
+
+    void setCacheSize(long kb);
+
+    TopologyManager topology();
```
dcapwell, line 43: I kinda wish we could just return `Node`, but since that is what owns the threads... I doubt we can =(

test/conf/cassandra.yaml
```diff
@@ -53,6 +53,7 @@ materialized_views_enabled: true
 drop_compact_storage_enabled: true
 file_cache_enabled: true
 auto_hints_cleanup_enabled: true
+accord_transactions_enabled: true
```
dcapwell, line 56: don't you need to change this config for JVM dtest as well?

dcapwell (changes requested):
one small comment about jvm-default, but overall LGTM +1

test/conf/cassandra.yaml
maedhroz: I changed it in `AccordTestBase`, of course, but outside that, my goal was to keep the rest of the in-JVM test landscape running w/ the default, which right now means `accord_transactions_enabled: false`. Right trade-off, or am I thinking about it wrong?

maedhroz:
Fixed a couple test failures related to the simulator and LOCAL_SYSTEM_KEYSPACE_NAMES...tests running again...

test/conf/cassandra.yaml
maedhroz: You know what...I think you're right. It's probably better to have all of this active in a testing context so we ferret out problems earlier. Changing...

dcapwell (approved):
+1

maedhroz:
Committed as https://github.com/apache/cassandra/commit/2e680a33c03ce66d4b1358e1a1cc11cf4ee0189f

PR #2122, opened 2023-01-31
CASSANDRA-18212: Reduce memory allocations of calls to ByteBufer.duplicate() made in org.apache.cassandra.transport.CBUtil#writeValue

NateAdere:
Used HeapByteBuffers and ThreadLocal to allocate a ByteBuffer for every thread. 
 * Reduced 20% memory allocation when doing mostly memory based queries
 * Reduced 4% memory allocation when doing disk-based queries
  
```
patch by Natnael Adere; reviewed by <Reviewers> for CASSANDRA-#####
```

The [Cassandra Jira](https://issues.apache.org/jira/browse/CASSANDRA-18212)


test/unit/org/apache/cassandra/transport/DuplicateHeapBufferTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private void directTest(int capacity, boolean readOnly)
+    {
+        ByteBuffer hollowBuffer = CBUtil.getLocalBuffer();
+        ByteBuffer bb = readOnly ? ByteBuffer.allocateDirect(capacity).asReadOnlyBuffer() : ByteBuffer.allocateDirect(capacity);
+
+        int size = bb.capacity();
+        buf = allocator.directBuffer(size);
+
+        try {
```
smiklosovic: braces on new line in this whole try catch

test/unit/org/apache/cassandra/transport/DuplicateHeapBufferTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+    private void heapTest(int capacity, boolean readOnly)
+    {
+        ByteBuffer hollowBuffer = CBUtil.getLocalBuffer();
+        ByteBuffer bb = readOnly ? ByteBuffer.allocate(capacity).asReadOnlyBuffer() : ByteBuffer.allocate(capacity);
+
+        int size = bb.capacity();
+        buf = allocator.heapBuffer(size);
+
+        try {
```
smiklosovic: same

test/unit/org/apache/cassandra/transport/DuplicateHeapBufferTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+    @Test
+    public void CBUtilWriteValueTest()
+    {
+        ByteBuffer bb = ByteBuffer.allocate(50);
+        int size = bb.capacity();
+        buf = allocator.heapBuffer(size);
+
+        CBUtil.writeValue(bb, buf);
+        Assert.assertEquals(0, buf.readerIndex());
```
smiklosovic: cant you import it statically? saves some time to read it

test/unit/org/apache/cassandra/transport/DuplicateHeapBufferTest.java
```diff
@@ -0,0 +1,107 @@
...
+        ByteBuffer bb = readOnly ? ByteBuffer.allocateDirect(capacity).asReadOnlyBuffer() : ByteBuffer.allocateDirect(capacity);
+
+        int size = bb.capacity();
+        buf = allocator.directBuffer(size);
+
+        try {
+            MemoryUtil.duplicateHeapByteBuffer(bb, hollowBuffer);
+        }
+        catch (Error e) {
+            Assert.assertEquals(new AssertionError().getClass(), e.getClass());
```
smiklosovic: why do you need to create an instance of AssertionError to call getClass() on it? Does not `AssertionError.class.getClass()` do the job?

test/unit/org/apache/cassandra/transport/DuplicateHeapBufferTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+        try {
+            ByteBuffer temp = MemoryUtil.duplicateHeapByteBuffer(bb, hollowBuffer);
+            Assert.assertEquals(bb.position(), temp.position());
+            Assert.assertEquals(bb.limit(), temp.limit());
+            Assert.assertEquals(bb.capacity(), temp.capacity());
+            Assert.assertEquals(bb.array(), temp.array());
+        }
+        catch (Exception e){
+            Assert.assertEquals(new ReadOnlyBufferException().getClass(), e.getClass());
```
smiklosovic: `ReadOnlyBufferException.class.getClass()` maybe?

src/java/org/apache/cassandra/transport/CBUtil.java
```diff
@@ -46,7 +46,7 @@
 import org.apache.cassandra.utils.Pair;
 import org.apache.cassandra.utils.TimeUUID;
 import org.apache.cassandra.utils.UUIDGen;
-
+import org.apache.cassandra.utils.memory.MemoryUtil;
```
smiklosovic, line 48: should not be there some free line after imports and before the beginning of comments? This feels strange.

src/java/org/apache/cassandra/transport/CBUtil.java
```diff
@@ -614,4 +623,9 @@ private static byte[] readRawBytes(ByteBuf cb, int length)
         return bytes;
     }
 
+    public static ByteBuffer getLocalBuffer()
+    {
+        return localBuffer.get();
+    }
+
```
smiklosovic: empty line probably redundant?

test/unit/org/apache/cassandra/transport/DuplicateHeapBufferTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.ByteBufAllocator;
+import io.netty.buffer.PooledByteBufAllocator;
+import org.apache.cassandra.utils.memory.MemoryUtil;
+
+public class DuplicateHeapBufferTest
```
dcapwell: spoke already, we should remove as you moved away from heap buffer copying

src/java/org/apache/cassandra/utils/memory/MemoryUtil.java
```diff
@@ -34,19 +34,20 @@
 
     private static final Unsafe unsafe;
     private static final Class<?> DIRECT_BYTE_BUFFER_CLASS, RO_DIRECT_BYTE_BUFFER_CLASS;
-    private static final long DIRECT_BYTE_BUFFER_ADDRESS_OFFSET;
```
dcapwell: can you revert all changes?  as you migrated away from the heap buffer duplicate logic this logic is no longer needed to be changed

src/java/org/apache/cassandra/transport/CBUtil.java
```diff
@@ -614,4 +653,9 @@ private static byte[] readRawBytes(ByteBuf cb, int length)
         return bytes;
     }
 
+    public static ByteBuffer getLocalBuffer()
```
dcapwell: I would leave private for now

dcapwell: also can you rename to `getLocalDirectBuffer`?  so callers are more clear

src/java/org/apache/cassandra/transport/CBUtil.java
```diff
@@ -478,7 +487,37 @@ public static void writeValue(ByteBuffer bytes, ByteBuf cb)
         cb.writeInt(remaining);
 
         if (remaining > 0)
-            cb.writeBytes(bytes.duplicate());
+        {
+            cb.writeBytes(MemoryUtil.duplicateHeapByteBuffer(bytes, getLocalBuffer()));
+        }
+    }
+
+    public static void writeValue(ByteBuffer src, ByteBuf dest)
```
dcapwell: should not copy/paste an existing method...

You "should" add a new method that can write a `ByteBuffer` to a `ByteBuf` without allocating, then have the one line `cb.writeBytes(bytes.duplicate());` call that method instead.

src/java/org/apache/cassandra/transport/CBUtil.java
```diff
@@ -466,7 +475,7 @@ public static void writeValue(byte[] bytes, ByteBuf cb)
         cb.writeBytes(bytes);
     }
 
-    public static void writeValue(ByteBuffer bytes, ByteBuf cb)
+    public static void writeValu(ByteBuffer bytes, ByteBuf cb)
```
dcapwell: please revert

src/java/org/apache/cassandra/transport/CBUtil.java
```diff
@@ -69,6 +69,15 @@ protected CharsetDecoder initialValue()
         }
     };
 
+    private final static FastThreadLocal<ByteBuffer> localBuffer = new FastThreadLocal<ByteBuffer>()
```
dcapwell: would be best to call it `localDirectBuffer`

src/java/org/apache/cassandra/transport/CBUtil.java
```diff
@@ -478,7 +487,37 @@ public static void writeValue(ByteBuffer bytes, ByteBuf cb)
...
+            dest.writeInt(-1);
+            return;
+        }
+
+        int length = src.remaining();
+        dest.writeInt(length);
+
+        if (src.hasArray())
+        {
+            byte[] array = src.array();
```
dcapwell, line 506: can you add a comment explaining why we do this when direct duplicates the buffer?  You found a JDK bug, we should expose that to anyone reading so they know that there may be dragons to replicate

dcapwell (changes requested):

src/java/org/apache/cassandra/transport/CBUtil.java
```diff
@@ -478,7 +487,35 @@ public static void writeValue(ByteBuffer bytes, ByteBuf cb)
         cb.writeInt(remaining);
 
         if (remaining > 0)
-            cb.writeBytes(bytes.duplicate());
+        {
```
dcapwell: remove the `{}`

src/java/org/apache/cassandra/transport/CBUtil.java
```diff
@@ -478,7 +487,35 @@ public static void writeValue(ByteBuffer bytes, ByteBuf cb)
...
+            addBytes(bytes, cb);
+        }
+    }
+
+    public static void addBytes(ByteBuffer src, ByteBuf dest)
+    {
+        if (src == null)
```
dcapwell: why the null check?  at least in the single usage this doesn't allow null... I prefer if we remove and allow the NPE if a user provides null

test/unit/org/apache/cassandra/utils/Generators.java
```diff
@@ -299,6 +299,16 @@ private static char[] createDNSDomainPartDomain()
...
+    {
+        return bytes(min, max, SourceDSL.arbitrary().constant(BBCases.HEAP));
+    }
+
+    public static Gen<ByteBuffer> bytesAnyType(int min, int max)
+    {
+        return bytes(min, max, SourceDSL.arbitrary().enumValues(BBCases.class));
+    }
+
+    public static Gen<ByteBuffer> bytes(int min, int max, Gen<BBCases> cases)
```
dcapwell: have to make this private, or fix `BBCases`.  You can't have a public method expose a private class (the build should be failing right now...).

If you choose to keep this method (I would not) then you should rename `BBCases` to be more clear to readers...

test/unit/org/apache/cassandra/utils/Generators.java
```diff
@@ -314,11 +324,32 @@ public static Gen<ByteBuffer> bytes(int min, int max)
...
+            switch (cases.generate(rnd))
+            {
+                case HEAP: return ByteBuffer.wrap(LazySharedBlob.SHARED_BYTES, offset, size);
+                case READ_ONLY_HEAP: return ByteBuffer.wrap(LazySharedBlob.SHARED_BYTES, offset, size).asReadOnlyBuffer();
+                case DIRECT:
+                {
+                    ByteBuffer bb = ByteBuffer.allocateDirect(size);
+                    bb.put(LazySharedBlob.SHARED_BYTES, offset, size);
+                    bb.flip();
+                    return bb;
```
dcapwell: can you make this a function?  best not to copy/paste

test/unit/org/apache/cassandra/transport/WriteBytesDuplicateTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import org.apache.cassandra.utils.Generators;
+import org.assertj.core.api.Assertions;
+
+import static org.quicktheories.QuickTheory.qt;
+
+
+public class WriteBytesDuplicateTest
```
dcapwell: why `Duplicate`?  You are "writing"

dcapwell (changes requested):

src/java/org/apache/cassandra/transport/CBUtil.java
dcapwell: @NateAdere  why was this comment resolved?

src/java/org/apache/cassandra/transport/CBUtil.java
```diff
@@ -478,7 +487,32 @@ public static void writeValue(ByteBuffer bytes, ByteBuf cb)
...
+    }
+
+    public static void addBytes(ByteBuffer src, ByteBuf dest)
+    {
+        if (src.remaining() == 0)
+            return;
+
+        int length = src.remaining();
+
+        //heap buffers are copied this way in order to avoid JVM crashing
```
dcapwell: please give more detail, this isn't enough for anyone to understand why this code is this way.  You should also move it within the `src.hasArray` block as that comment only makes sense there.

test/unit/org/apache/cassandra/utils/Generators.java
```diff
@@ -314,11 +324,36 @@ public static Gen<ByteBuffer> bytes(int min, int max)
...
+        switch (cases.generate(rnd))
+        {
+            case HEAP: return ByteBuffer.wrap(LazySharedBlob.SHARED_BYTES, offset, size);
+            case READ_ONLY_HEAP: return ByteBuffer.wrap(LazySharedBlob.SHARED_BYTES, offset, size).asReadOnlyBuffer();
+            case DIRECT:
+            {
+                ByteBuffer bb = ByteBuffer.allocateDirect(size);
+                bb.put(LazySharedBlob.SHARED_BYTES, offset, size);
+                bb.flip();
+                return bb;
```
dcapwell: when I said you should avoid copy/paste and make this a method, I mean 

```
ByteBuffer bb = ByteBuffer.allocateDirect(size);
                 bb.put(LazySharedBlob.SHARED_BYTES, offset, size);
                 bb.flip();
```

and not this new `handleCases`...

test/unit/org/apache/cassandra/utils/Generators.java
```diff
@@ -314,11 +324,36 @@ public static Gen<ByteBuffer> bytes(int min, int max)
...
+                bb.put(LazySharedBlob.SHARED_BYTES, offset, size);
+                bb.flip();
+                return bb.asReadOnlyBuffer();
+            }
+            default: throw new AssertionError("cann't wait for jdk 17!");
+        }
     }
 
-    /**
+     /**
```
dcapwell, line 351: please revert

dcapwell (changes requested):

src/java/org/apache/cassandra/transport/CBUtil.java
NateAdere: Is this comment descriptive enough?

test/unit/org/apache/cassandra/utils/Generators.java
NateAdere: the new handleCases helps with readability so my plan is to keep that method and add "ByteBuffer bb = directBufferFromSharedBlob(offset, size);" to address the copy/paste code. Any opinions on removing or keeping the method?

src/java/org/apache/cassandra/transport/CBUtil.java
dcapwell: can you flesh this out more?  Scott showed you the bug.

it is also not an "improper" way, it was a bug in CMS

test/unit/org/apache/cassandra/utils/Generators.java
```diff
@@ -314,11 +324,39 @@ public static Gen<ByteBuffer> bytes(int min, int max)
...
+    private static ByteBuffer handleCases(Gen<BBCases> cases, RandomnessSource rnd, int offset, int size) {
+        switch (cases.generate(rnd))
+        {
+            case HEAP: return ByteBuffer.wrap(LazySharedBlob.SHARED_BYTES, offset, size);
+            case READ_ONLY_HEAP: return ByteBuffer.wrap(LazySharedBlob.SHARED_BYTES, offset, size).asReadOnlyBuffer();
+            case DIRECT:
+            {
+                ByteBuffer bb = directBufferFromSharedBlob(offset, size);
+                return bb;
+            }
```
dcapwell: nit: cleanup, you can just do `return directBufferFromSharedBlob(offset, size);` which lets you get rid of 3 lines of code

test/unit/org/apache/cassandra/utils/Generators.java
```diff
@@ -314,11 +324,39 @@ public static Gen<ByteBuffer> bytes(int min, int max)
...
+            case READ_ONLY_DIRECT:
+            {
+                ByteBuffer bb = directBufferFromSharedBlob(offset, size);
+                return bb.asReadOnlyBuffer();
```
dcapwell: nit: cleanup, you can just do `return directBufferFromSharedBlob(offset, size);` which lets you get rid of 3 lines of code

test/unit/org/apache/cassandra/utils/Generators.java
```diff
@@ -314,11 +324,39 @@ public static Gen<ByteBuffer> bytes(int min, int max)
...
+            }
+            default: throw new AssertionError("cann't wait for jdk 17!");
```
dcapwell: fix typo plz

dcapwell (approved):

michaelsembwever:
Committed with https://github.com/apache/cassandra/commit/35f8da66f9b05d18b4177f8d2e1b86c772ad2221

PR #2123, opened 2023-02-01
Add cache type information for maximum memory usage warning message

yongj:
Add cache type information for maximum memory usage warning message

Tested on one server by setting the networking_cache_size to 1MiB. The new INFO message looks like below in the system log:
`INFO  [Messaging-EventLoop-3-1] 2023-02-01 09:00:47,673 NoSpamLogger.java:105 - Maximum memory usage reached (1.000MiB) for networking, cannot allocate chunk of 8.000MiB`

patch by Yong Jiang; reviewed by <Reviewers> for CASSANDRA-18184

### DIFF
--- a/src/java/org/apache/cassandra/utils/memory/BufferPool.java
+++ b/src/java/org/apache/cassandra/utils/memory/BufferPool.java
@@ -441,7 +441,7 @@ public class BufferPool
                 {
                     if (memoryUsageThreshold > 0)
                     {
-                        noSpamLogger.info("Maximum memory usage reached ({}) for {}, cannot allocate chunk of {}",
+                        noSpamLogger.info("Maximum memory usage reached ({}) for {} buffer pool, cannot allocate chunk of {}",
                                           readableMemoryUsageThreshold, name, READABLE_MACRO_CHUNK_SIZE);
                     }
                     return null;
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/utils/memory/BufferPool.java
```diff
@@ -441,8 +441,8 @@ private Chunk allocateMoreChunks()
                 {
                     if (memoryUsageThreshold > 0)
                     {
-                        noSpamLogger.info("Maximum memory usage reached ({}), cannot allocate chunk of {}",
-                                          readableMemoryUsageThreshold, READABLE_MACRO_CHUNK_SIZE);
+                        noSpamLogger.info("Maximum memory usage reached ({}) for {}, cannot allocate chunk of {}",
```
Maxwell-Guo: see line 461 , may be " for {} buffer pool " is better and consistent with line 461‘s log message.

yongj: Thanks for the suggestion. That will provide more context. I will make the change. Do I need to creat a new pull request or make the change based on current one?

Maxwell-Guo: you can just modify the change on the current code, and doing a new git push to you branch ,then we will see the change here,

yongj: sounds good. I have pushed the change to my branch.

PR #2132, opened 2023-02-02
use equals() rather than reference equality for excluding ourselves from the dependency builder

maedhroz:
patch by Benedict; reviewed by Caleb Rackliffe for CASSANDRA-18218

### DIFF
--- a/.build/include-accord.sh
+++ b/.build/include-accord.sh
@@ -25,7 +25,7 @@ set -o nounset
 bin="$(cd "$(dirname "$0")" > /dev/null; pwd)"
 
 accord_repo='https://github.com/apache/cassandra-accord.git'
-accord_branch='1230eceb077c928123e6bf848a103964fe90c9f7'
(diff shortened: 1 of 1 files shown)


.build/include-accord.sh
```diff
@@ -24,8 +24,8 @@ set -o nounset
 
 bin="$(cd "$(dirname "$0")" > /dev/null; pwd)"
 
-accord_repo='https://github.com/apache/cassandra-accord.git'
-accord_branch='1230eceb077c928123e6bf848a103964fe90c9f7'
+accord_repo='https://github.com/maedhroz/cassandra-accord.git'
+accord_branch='349281bcb5a870312af2fe1c97601a1b494e22b7'
```
maedhroz: Just for testing...this will revert back to a trunk commit when the accord side merges.

maedhroz:
https://app.circleci.com/pipelines/github/maedhroz/cassandra?branch=CASSANDRA-18218

PR #2133, opened 2023-02-02
CASSANDRA-15254 Support update statements over the SettingsTable virtual table

Mmuzaf:
 
```

```


src/java/org/apache/cassandra/config/ConfigFields.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.config;
+
+/**
+ * This class is autogenerated bases on the {@link Config} class. It contains all the configuration property
+ * names as constants.
+ */
+public class ConfigFields
```
dcapwell: I don't follow this logic... why is there another config system rather than just updating `Config`?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -3711,7 +3718,13 @@ public static boolean getCDCBlockWrites()
         return conf.cdc_block_writes;
     }
 
+    @Deprecated
```
dcapwell: why?  This isn't a public class, we can make breaking changes here

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,371 @@
...
+    public ConfigurationRegistry()
+    {
+        this(DatabaseDescriptor.getRawConfig());
+    }
+
+    public ConfigurationRegistry(Config config)
+    {
+        registry = getProperties().entrySet()
+                                  .stream()
+                                  .collect(Collectors.toMap(Map.Entry::getKey, e -> new ConfigurationUnit<>(config, e.getKey(), e.getValue())));
```
dcapwell: you shouldn't hold the ref, this can change on you causing this class to drift.  `DD` is what owns `Config`, we should reference it and not cache

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,371 @@
...
+    {
+        ConfigurationUnit<?> unit = registry.get(name);
+
+        if (unit == null)
+            throw new ConfigurationException(String.format("Property with name '%s' is not availabe.", name));
+
+        if (unit.setter == null)
+            throw new ConfigurationException(String.format("Property with name '%s' is not writable.", name));
+
+        unit.lock.lock();
```
dcapwell: why do we need locks?

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,371 @@
...
+
+    /**
+     * settings table was released in 4.0 and attempted to support nested properties for a few hand selected properties.
+     * The issue is that 4.0 used '_' to seperate the names, which makes it hard to map back to the yaml names; to solve
+     * this 4.1+ uses '.' to avoid possible conflicts, this class provides mappings from old names to the '.' names.
+     *
+     * There were a handle full of properties which had custom names, names not present in the yaml, this map also
+     * fixes this and returns the proper (what is accessable via yaml) names.
+     */
+    private static Map<String, String> getBackwardsCompatableNames()
```
dcapwell: should not rely on copy/paste, the 2 may drift and will be hard to maintain

src/java/org/apache/cassandra/config/ConfigFields.java
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.config;
+
+/**
+ * This class is autogenerated bases on the {@link Config} class. It contains all the configuration property
```
dcapwell: > This class is autogenerated

Where?  I don't see anything in this patch that generates it, `build.xml` wasn't updated and this is not in a `gen-java` dir

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -384,6 +386,8 @@ public StorageService()
         jmxObjectName = "org.apache.cassandra.db:type=StorageService";
 
         sstablesTracker = new SSTablesGlobalTracker(SSTableFormat.Type.current());
+
+        ConfigurationRegistry.instance.addPropertyChangeListener(REPAIR_REQUEST_TIMEOUT, this::repairRequestTimeoutListener, DurationSpec.LongMillisecondsBound.class);
```
dcapwell: this doesn't need to exist... All "mutators" via JMX log the mutation, this may be in DD or in the Mean itself (such as this case), there is no reason to trigger this as all this does is update `Config`.

the vtable can produce a similar log, so don't need to duplicate

src/java/org/apache/cassandra/db/virtual/SettingsTable.java
```diff
@@ -59,102 +56,83 @@ final class SettingsTable extends AbstractVirtualTable
...
-        if (BACKWARDS_COMPATABLE_NAMES.containsKey(name))
-            ClientWarn.instance.warn("key '" + name + "' is deprecated; should switch to '" + BACKWARDS_COMPATABLE_NAMES.get(name) + "'");
-        if (PROPERTIES.containsKey(name))
-            result.row(name).column(VALUE, getValue(PROPERTIES.get(name)));
-        return result;
+        converterRegistry.put(Boolean.class, CassandraRelevantProperties.BOOLEAN_CONVERTER);
+        converterRegistry.put(boolean.class, CassandraRelevantProperties.BOOLEAN_CONVERTER);
+        converterRegistry.put(Integer.class, CassandraRelevantProperties.INTEGER_CONVERTER);
+        converterRegistry.put(int.class, CassandraRelevantProperties.INTEGER_CONVERTER);
+        converterRegistry.put(DurationSpec.LongMillisecondsBound.class,
```
dcapwell: why are you using reflection when you can do `DurationSpec.LongMillisecondsBound::new`?

src/java/org/apache/cassandra/db/virtual/SettingsTable.java
```diff
@@ -59,102 +56,83 @@ final class SettingsTable extends AbstractVirtualTable
...
-        SimpleDataSet result = new SimpleDataSet(metadata());
-        for (Map.Entry<String, Property> e : PROPERTIES.entrySet())
-            result.row(e.getKey()).column(VALUE, getValue(e.getValue()));
-        return result;
+        Class<?> setterType = configurationRegistry.getPropertyType(name);
+        PropertyConverter<?> converter = converterRegistry.get(setterType);
+        if (converter == null)
+            throw invalidRequest("Unknown converter for property with name '%s' and type '%s'", name, setterType);
+
+        configurationRegistry.update(name, value == null ? null : converter.convert(value));
```
dcapwell: I don't follow why this is needed.  You have a map of "real" type to "PropertyConverter", so can't you just do

```
Property p = PROPERTIES.get(name);
if (p == null) throw unknownProperty(name);
p.set(DatabaseDescriptor.getRawConfig(), getConverter(p.getType()).convert(value));
```

If there is a missing converter that means we have a bug, we don't support all types we support...

src/java/org/apache/cassandra/db/virtual/SettingsTable.java
```diff
@@ -59,102 +56,83 @@ final class SettingsTable extends AbstractVirtualTable
...
-        names.put("server_encryption_options_cipher_suites", "server_encryption_options.cipher_suites");
-        names.put("server_encryption_options_enabled", "server_encryption_options.enabled");
-        names.put("server_encryption_options_internode_encryption", "server_encryption_options.internode_encryption");
-        names.put("server_encryption_options_optional", "server_encryption_options.optional");
-        names.put("transparent_data_encryption_options_chunk_length_kb", "transparent_data_encryption_options.chunk_length_kb");
-        names.put("transparent_data_encryption_options_cipher", "transparent_data_encryption_options.cipher");
-        names.put("transparent_data_encryption_options_enabled", "transparent_data_encryption_options.enabled");
-        names.put("transparent_data_encryption_options_iv_length", "transparent_data_encryption_options.iv_length");
-
-        return names;
```
dcapwell: why are you removing?  These strings only make sense to this class, so pulling out to a config registry is not safe, only this class should know about config names it created in the past

src/java/org/apache/cassandra/db/virtual/SettingsTable.java
```diff
@@ -59,102 +56,83 @@ final class SettingsTable extends AbstractVirtualTable
...
+
+        configurationRegistry.update(name, value == null ? null : converter.convert(value));
     }
 
-    private String getValue(Property prop)
+    private static @Nullable String getProperty(ConfigurationRegistry registry, String name)
     {
-        Object value = prop.get(config);
+        Object value = registry.get(name);
         return value == null ? null : value.toString();
```
dcapwell: an issue this patch doesn't look to address is that the `input` and `output` may not match.

For basic types like int/boolean this is fine as they are simple, but cases like `hinted_handoff_disabled_datacenters` are not correct as that's java collection toString, which we won't have a property converter for...  it's also weird for users to generate java's collection toString.  There are many types that fall into this bucket.

You *should* have a property that w/e is returned from this table is accepted by this table; which is not true with this patch.

test/distributed/org/apache/cassandra/distributed/test/jmx/JMXGetterCheckTest.java
```diff
@@ -119,6 +120,16 @@ public void test() throws Exception
...
+
+        long ts = currentTimeMillis();
+        while (!Thread.interrupted())
+        {
+            if (ts + 5000 >= currentTimeMillis())
+                continue;
+
+            ts = currentTimeMillis();
+            System.out.println(">>>> Running: " + ts);
+        }
```
dcapwell: why is this here?  Also should avoid `System.out`

src/java/org/apache/cassandra/config/DatabaseDescriptorWalker.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.config;
+
+import org.apache.cassandra.config.registry.PropertySetterWalker;
+
+/**
+ * This class is autogenerated bases on given {@link DatabaseDescriptor} class. It contains all the configuration
```
dcapwell: > This class is autogenerated

Where?  I don't see such a thing in this patch

src/java/org/apache/cassandra/config/DurationSpec.java
```diff
@@ -135,6 +135,25 @@ public TimeUnit unit()
         return unit;
     }
 
+    public static <T extends DurationSpec> T from(String value, Class<T> clazz)
```
dcapwell: can we remove?  The only usage can be replaced with `DurationSpec.IntMillisecondsBound::new`

src/java/org/apache/cassandra/config/PropertyConverter.java
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.config;
+
+/**
+ * A converter that converts a string to a specific type.
+ * @param <T> the type to convert to
+ */
+public interface PropertyConverter<T>
```
dcapwell: you moved this out of the `CassandraRelevantProperties`, so are you making a assertion that system properties *must* match the same input as the vtable?  I am not saying they shouldn't, I am calling out that is what this code is asserting.  The current usage are all primitives, so this is fine, but we should call out that we don't actual handle `ConfigurationException` in the vtable, so user will get an error saying C* failed, and not that we had bad input

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,371 @@
...
+import org.apache.cassandra.config.Replacements;
+import org.apache.cassandra.exceptions.ConfigurationException;
+import org.apache.cassandra.utils.AbstractIterator;
+
+
+/**
+ * This is a simple configuration property registry that stores all the {@link Config} settings.
+ * It is a singleton and can be accessed via {@link #instance} field.
+ */
+public class ConfigurationRegistry extends AbstractMap<String, Object>
```
dcapwell: should *not* extend `AbstractMap<String, Object>`, this can be error prone as there are 2 other maps and from an API everything takes a `String`, so what does `ConfigurationRegistry.get` actually mean?  What about `ConfigurationRegistry.put` or `ConfigurationRegistry.removeKey`?

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,371 @@
...
+//        Class<T> type = (Class<T>) ((ParameterizedType) listener.getClass().getGenericInterfaces()[0]).getActualTypeArguments()[0];
+        if (!unit.setterType.equals(listenerType))
+            throw new ConfigurationException(String.format("Property with name '%s' expects type '%s', but got '%s'.", name, unit.setterType, listenerType));
+
+        listeners.put(name, new PropertyChangeListenerWrapper<>(listener, listenerType));
+    }
+
+    /** {@inheritDoc} */
+    @Override
+    public Set<Entry<String, Object>> entrySet()
```
dcapwell: this is very confusing, this boils down to the state in `registry`, but if I do `cr.put("does not exist", null)` I won't see that if I call `entrySet`

src/java/org/apache/cassandra/config/registry/PropertyChangeListener.java
```diff
@@ -0,0 +1,34 @@
...
+public interface PropertyChangeListener<T>
+{
+    /**
+     * Called before a configuration property has been changed.
+     *
+     * @param name the name of the property.
+     * @param oldValue the old value of the property.
+     * @param newValue the new value of the property.
+     */
+    void onBeforeChange(String name, T oldValue, T newValue);
```
dcapwell: based off name, should there also be `onAfterChange`?

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,371 @@
...
+import org.apache.cassandra.utils.AbstractIterator;
+
+
+/**
+ * This is a simple configuration property registry that stores all the {@link Config} settings.
+ * It is a singleton and can be accessed via {@link #instance} field.
+ */
+public class ConfigurationRegistry extends AbstractMap<String, Object>
+{
+    public static final List<PropertySetterWalker> PROPERTY_SETTERS_LIST = ImmutableList.of(new DatabaseDescriptorWalker());
```
dcapwell: I don't understand the need for `PropertySetterWalker` and this field?  There exists a `Property` for every key (nested and top level), so I think we could remove a lot of code by just calling `Property.set` like we do in YAML

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -122,6 +135,9 @@ private static CommitLog construct()
 
         // register metrics
         metrics.attach(executor, segmentManager);
+
+        ConfigurationRegistry.instance.addPropertyChangeListener(CDC_BLOCK_WRITES, this::cdcBlockWritesChangeListener, Boolean.class);
```
dcapwell: Looks like the majority of complexity of this patch is to add listeners for this single listener?  There maybe more cases as you flesh this out, but at least in the current PR this is the only real listener (the others just log you mutated, so don't need to listen).

I feel we could simplify to

```ConfigurationRegistry.instance.register("cdc_block_writes", this:: setCDCBlockWrites0)```

Where `setCDCBlockWrites0` is just the logic in `cdcBlockWritesChangeListener`

I don't like the `Boolean.class` as it's brittle and can drift; in fact this patch provides the wrong type!  You want `Boolean.TYPE`.

File: src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java [hunk 21]
```diff
@@ -0,0 +1,371 @@
...
+            throw new ConfigurationException(String.format("Property with name '%s' is not availabe for resitering handler.", name));
+
+        if (unit.setterType == null)
+            throw new ConfigurationException(String.format("Property with name '%s' is not writable.", name));
+
+//        Class<T> type = (Class<T>) ((ParameterizedType) listener.getClass().getGenericInterfaces()[0]).getActualTypeArguments()[0];
+        if (!unit.setterType.equals(listenerType))
+            throw new ConfigurationException(String.format("Property with name '%s' expects type '%s', but got '%s'.", name, unit.setterType, listenerType));
+
+        listeners.put(name, new PropertyChangeListenerWrapper<>(listener, listenerType));
```
dcapwell: what happens if multiple different places need to listen to the same config key?

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,371 @@
...
+
+        /** {@inheritDoc} */
+        @Override
+        public void onBeforeChange(String name, Object oldValue, Object newValue)
+        {
+            T castedOldValue = type.cast(oldValue);
+            T castedNewValue = type.cast(newValue);
+            listener.onBeforeChange(name, castedOldValue, castedNewValue);
+        }
+    }
```
dcapwell: is this needed?  If the user defines the call site as `boolean` the JVM will add the cost for you, so should be able to drop this whole class

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,371 @@
...
+            ConfigurationUnit<?> old = registry.get(name);
+            registry.replace(name, new ConfigurationUnit<>(old.source, name, old.getter, clazz, setter));
+        }
+    }
+
+    /**
+     * Configuration unit is a single configuration entry that can be read and written.
+     * @param <V> type of the configuration setter value.
+     */
+    private static class ConfigurationUnit<V> implements Map.Entry<String, V>
```
dcapwell: im not seeing why we have this class... it looks like it just duplicates `Property`.

dcapwell (changes requested):

dcapwell:
One thing I want to say is thanks for trying to tackle this work, looking forward to the progress!

src/java/org/apache/cassandra/auth/AuthConfig.java
```diff
@@ -56,9 +60,9 @@ public static void applyAuth()
         // work with PasswordAuthenticator, so log a message if some other authenticator
         // is in use and non-default values are detected
         if (!(authenticator instanceof PasswordAuthenticator)
-            && (conf.credentials_update_interval != null
-                || conf.credentials_validity.toMilliseconds() != 2000
-                || conf.credentials_cache_max_entries != 1000))
+            && (registry.get(DurationSpec.IntMillisecondsBound.class, ConfigFields.CREDENTIALS_UPDATE_INTERVAL) != null
```
dcapwell, line 68: why is config validation going through `registry`?

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -80,45 +82,54 @@ public static Set<String> splitCommaDelimited(String src)
...
-    public volatile DurationSpec.IntMillisecondsBound roles_update_interval = null;
-    public volatile boolean roles_cache_active_update = false;
-    @Replaces(oldName = "credentials_validity_in_ms", converter = Converters.MILLIS_DURATION_INT, deprecated = true)
-    public volatile DurationSpec.IntMillisecondsBound credentials_validity = new DurationSpec.IntMillisecondsBound("2s");
-    public volatile int credentials_cache_max_entries = 1000;
-    @Replaces(oldName = "credentials_update_interval_in_ms", converter = Converters.MILLIS_CUSTOM_DURATION, deprecated = true)
-    public volatile DurationSpec.IntMillisecondsBound credentials_update_interval = null;
-    public volatile boolean credentials_cache_active_update = false;
+    @Mutable @Replaces(oldName = "permissions_validity_in_ms", converter = Converters.MILLIS_DURATION_INT, deprecated = true)
+    public DurationSpec.IntMillisecondsBound permissions_validity = new DurationSpec.IntMillisecondsBound("2s");
```
dcapwell: why did you remove `volatile`?  this is a regression

src/java/org/apache/cassandra/config/Mutable.java
```diff
@@ -0,0 +1,33 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation denoting that a configuration field in the {@link Config} is mutable with a live update.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.FIELD)
```
dcapwell, line 30: we do support methods as well, think maybe 1-2 use it and not sure if they are mutable... but it is supported

src/java/org/apache/cassandra/config/StringConverters.java
```diff
@@ -0,0 +1,104 @@
...
+     * or default conversion should be used instead.
+     */
+    public static StringConverters fromType(Class<?> type)
+    {
+        for (StringConverters converter : values())
+            if (converter.type.equals(type))
+                return converter;
+
+        return null;
+    }
```
dcapwell, line 103: this highlights to me that we don't want an enum, we really want a map.  `O(n)` for every config accessed is a problem as we would need to do this on the write side as well, so its `O(n * m)` due to this... 

We need a `Map<Class<?>, Converter>`

src/java/org/apache/cassandra/config/registry/ConfigurationConstraint.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.config.registry;
+
+/**
+ * Interface validating configuration property's value.
+ */
+public interface ConfigurationConstraint<T>
```
dcapwell, line 24: what is this?  looks like dead code?

src/java/org/apache/cassandra/config/registry/ConfigurationListener.java
```diff
@@ -0,0 +1,41 @@
...
+     * Called on configuration change property event occurr.
+     *
+     * @param name     the name of the property.
+     * @param oldValue the old value of the property.
+     * @param newValue the new value of the property.
+     */
+    void onUpdate(String name, T oldValue, T newValue);
+
+    /** Type of property change. */
+    enum ChangeType
```
dcapwell, line 36: this is dead code

src/java/org/apache/cassandra/config/registry/TypeConverter.java
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Converts configuration value from one type to another, you can use {@link org.apache.cassandra.config.StringConverters}
+ * if your input type is String and you want to convert it to an appropriate configuration object type.
+ *
+ * @param <T> Type to convert to.
+ *
+ * @see Registry
+ * @see org.apache.cassandra.config.StringConverters
+ */
+public interface TypeConverter<T>
```
dcapwell, line 33: this API isn't what we need, we need

```
T1 -> T2
T2 -> T1
```

The reason is that w/e we do when we return *must* be accepted as valid input when the user writes, so both directions are required.

Also, using `Object` is a problem as we can use the wrong type and wouldn't ever know

src/java/org/apache/cassandra/db/virtual/SettingsTable.java
```diff
@@ -157,4 +174,85 @@ private static Map<String, String> getBackwardsCompatableNames()
...
+        }
+
+        @Override
+        public void set(String name, @Nullable Object value)
+        {
+            Replacement replacement = replacements.get(name);
+            if (replacement == null)
+                registry.set(name, value);
+            else
+                throw new ConfigurationException(String.format("Unable to set '%s' as it is deprecated and is read only; use '%s' instead", name, replacement.newName));
```
dcapwell, line 205: this is a breaking change, so can't allow this... the point of `Replacement` is that we can make changes without forcing users, this logic is now forcing users.

src/java/org/apache/cassandra/db/virtual/SettingsTable.java
```diff
@@ -68,54 +91,48 @@ public DataSet data(DecoratedKey partitionKey)
         SimpleDataSet result = new SimpleDataSet(metadata());
         String name = UTF8Type.instance.compose(partitionKey.getKey());
         if (BACKWARDS_COMPATABLE_NAMES.containsKey(name))
-            ClientWarn.instance.warn("key '" + name + "' is deprecated; should switch to '" + BACKWARDS_COMPATABLE_NAMES.get(name) + "'");
-        if (PROPERTIES.containsKey(name))
-            result.row(name).column(VALUE, getValue(PROPERTIES.get(name)));
+            ClientWarn.instance.warn("key '" + name + "' is deprecated; should switch to '" + BACKWARDS_COMPATABLE_NAMES.get(name) + '\'');
+        if (registry.contains(name))
```
dcapwell, line 95: how does the registry know about names defined in `BACKWARDS_COMPATABLE_NAMES`?  Those are not valid configs, those are bugs with this specific API, and must only live with this API

src/java/org/apache/cassandra/db/virtual/SettingsTable.java
```diff
@@ -68,54 +91,48 @@ public DataSet data(DecoratedKey partitionKey)
...
+            runExceptionally(() -> result.row(name).column(VALUE, registry.getString(name)),
```
dcapwell, line 96: nit: `try/catch` is easier to fallow.

src/java/org/apache/cassandra/db/virtual/SettingsTable.java
```diff
@@ -68,54 +91,48 @@ public DataSet data(DecoratedKey partitionKey)
...
+                                         e -> invalidRequest("Invalid configuration request during searching by key; cause: '%s'", e.getMessage()));
```
dcapwell, line 97: should include the config key

src/java/org/apache/cassandra/db/virtual/SettingsTable.java
```diff
@@ -68,54 +91,48 @@ public DataSet data(DecoratedKey partitionKey)
...
 
     @Override
     public DataSet data()
     {
         SimpleDataSet result = new SimpleDataSet(metadata());
-        for (Map.Entry<String, Property> e : PROPERTIES.entrySet())
-            result.row(e.getKey()).column(VALUE, getValue(e.getValue()));
+        for (String name : registry.keys())
+            runExceptionally(() -> result.row(name).column(VALUE, registry.getString(name)),
+                                         e -> invalidRequest("Invalid configuration request; cause: '%s'", e.getMessage()));
```
dcapwell, line 107: missing config key

src/java/org/apache/cassandra/db/virtual/SettingsTable.java
```diff
@@ -157,4 +174,85 @@ private static Map<String, String> getBackwardsCompatableNames()
...
     }
+
+    /**
+     * This class is used to provide backwards compatable support for the settings table in case the {@link Config}
+     * metadata changes. This class will provide the old names for the properties, but will use the new name to
+     * get the value from the {@link Config} object.
+     * <p>
+     * Updating a configuration property object will throw an exception if you will try to update a deprecated property.
+     */
+    private static class BackwardsCompatableRegistry implements Registry
```
dcapwell, line 185: this code is more verbose than just handling the `Replacement`... before we would create a valid `Property` for each `Replacement` and that would handle the conversions and call the "real" `Property`

It looks like this class makes invalid assumptions, such as old names are not allowed... they are 100% supported, which makes this more verbose it seems...

Is there a reason to migrate away from the existing logic that makes old/new properties look the same?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1536,7 +1540,6 @@ else if (tokens == null)
     public void setRpcTimeout(long value)
     {
         DatabaseDescriptor.setRpcTimeout(value);
-        logger.info("set rpc timeout to {} ms", value);
```
dcapwell, line 1192: this is a regression, people may depend on these logs

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -972,6 +974,8 @@ public void runMayThrow() throws InterruptedException, ExecutionException, IOExc
             logger.info("Not joining ring as requested. Use JMX (StorageService->joinRing()) to initiate ring joining");
         }
 
+        getConfigRegistry().addPropertyConstraint(REPAIR_REQUEST_TIMEOUT,
+                                                  DurationSpec.LongMillisecondsBound.class, (newVal) -> Preconditions.checkState(newVal.toMilliseconds() > 0));
```
dcapwell, line 995: why does this exist?

`DurationSpec.LongMillisecondsBound` adds restrictions on the input, so your logic looks to duplicate?

You are also doing it *AFTER* configs have been accepted, so this is far too late to add.

You also do not add a useful error message, so this gets hard for users

src/java/org/apache/cassandra/utils/ConstantFieldsGenerateUtil.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.TreeSet;
+import java.util.regex.Pattern;
+
+import org.apache.cassandra.config.Config;
+import org.apache.cassandra.config.Mutable;
+
+/**
+ * This class is used to generate constants class of provided class.
+ */
+public class ConstantFieldsGenerateUtil
```
dcapwell, line 40: this is dead code.

Also, adding in code gen is a problem as it makes this solution far too complex.  Can we remove and avoid code gen all together?  I personally want to remove the `ConfigFields` class as well as that is a problem to maintain.

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -137,7 +150,13 @@ synchronized public CommitLog start()
             segmentManager.start();
             executor.start();
             started = true;
-        } catch (Throwable t)
+
+            DatabaseDescriptor.getConfigRegistry().addPropertyConstraint(CDC_BLOCK_WRITES, Boolean.TYPE, (newVal) -> ensureCDCEnabled());
```
dcapwell, line 151: if you add this to start don't you need to remove on shutdown?

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -137,7 +150,13 @@ synchronized public CommitLog start()
...
+            DatabaseDescriptor.getConfigRegistry().addPropertyConstraint(CDC_ON_REPAIR_ENABLED, Boolean.TYPE, (newVal) -> ensureCDCEnabled());
```
dcapwell, line 152: nit: cleaner to save `DatabaseDescriptor.getConfigRegistry()` to a variable, makes the code less verbose

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -137,7 +150,13 @@ synchronized public CommitLog start()
...
+            DatabaseDescriptor.getConfigRegistry().addPropertyChangeListener(CDC_BLOCK_WRITES, ConfigurationListener.ChangeType.BEFORE,
+                                                                             this::cdcBlockWritesBeforeChangeListener, Boolean.TYPE);
```
dcapwell, line 154: why does this API flip type and function?  others are type, function; this is function, type

src/java/org/apache/cassandra/db/ConsistencyLevel.java
```diff
@@ -264,4 +264,9 @@ private void requireNetworkTopologyStrategy(AbstractReplicationStrategy replicat
             throw new InvalidRequestException(String.format("consistency level %s not compatible with replication strategy (%s)",
                                                             this, replicationStrategy.getClass().getName()));
     }
+
+    public static ConsistencyLevel fromStringIgnoreCase(String s)
```
dcapwell, line 268: why?  Is this now trying to make even more things valid?  Is this needed for this patch?

src/java/org/apache/cassandra/config/registry/PrimitiveUnaryConverter.java
```diff
@@ -0,0 +1,84 @@
...
+ * Converter do type conversion from the specified Object  value to the given {@code T}. If the class is
+ * a primitive type (e.g. Boolean.TYPE, Long.TYPE etc), the value returned will use the corresponding
+ * wrapper type (Long.class, Boolean.class, etc).
+ *
+ * @see TypeConverter
+ * @see org.apache.cassandra.config.StringConverters
+ */
+public class PrimitiveUnaryConverter
+{
+    private static Object to(Class<?> cls, @Nonnull Object value)
```
dcapwell, line 35: this feels unneeded, if you follow the feedback I gave and use typed classes, then any violation is blocked during the cast that java adds...

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,401 @@
...
+        try
+        {
+            if (initialized)
+                return;
+            properties = ImmutableMap.copyOf(defaultLoader()
+                                             .flatten(Config.class)
+                                             .entrySet()
+                                             .stream()
+                                             .map(e -> new AbstractMap.SimpleEntry<>(e.getKey(), new PropertyAdapter(configProvider, e.getValue())))
+                                             .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
```
dcapwell, line 89: why is this in `lazyInit`? This is static and can be done during class loading... we shouldn't need to make the locks this costly

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,401 @@
...
+            Set<String> leftConstraints = new HashSet<>(constraints.keySet());
```
dcapwell, line 90: `com.google.common.collect.Sets#difference` is cheaper

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,401 @@
...
+            leftConstraints.removeAll(properties.keySet());
+            if (!leftConstraints.isEmpty())
+                throw new ConfigurationException("Constraints are defined for non-existing properties:" + leftConstraints);
```
dcapwell, line 93: properties are static and immutable, so why even allow this to happen?  This feels like trying to be lazy caused this, but there is no reason to do that and you can just make properties static and loaded during class loading, then you can block such cases from happening *at the call site*, which is far easier to work with

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,401 @@
...
+            Set<String> leftListeners = propertyChangeListeners.values().stream()
+                                                               .map(l -> l.wrappers.keySet())
+                                                               .flatMap(Collection::stream)
+                                                               .collect(Collectors.toSet());
+            leftListeners.removeAll(properties.keySet());
+            if (!leftListeners.isEmpty())
+                throw new ConfigurationException("Listeners are defined for non-existing properties:" + leftListeners);
```
dcapwell, line 100: properties are static and immutable, so why even allow this to happen?  This feels like trying to be lazy caused this, but there is no reason to do that and you can just make properties static and loaded during class loading, then you can block such cases from happening *at the call site*, which is far easier to work with

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,401 @@
...
+        finally
+        {
+            rwLock.writeLock().unlock();
+        }
+    }
+
+    @Override public void set(String name, @Nullable Object value)
+    {
+        lazyInit();
+        PropertyAdapter property = properties.get(name);
```
dcapwell, line 112: concurrency bug.. because you are trying to use these locks to defer logic, you need to access these via the locks as well.

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,401 @@
...
+            Object convertedValue = value;
+            // Do conversion if the value is not null and the type is not the same as the property type.
+            if (sourceType != null && !primitiveToWrapper(originalType).equals(sourceType))
+            {
+                StringConverters converter;
+                if (sourceType.equals(String.class) && (converter = StringConverters.fromType(originalType)) != null)
+                    convertedValue = converter.fromString((String) value, originalType);
+                else
+                    throw new IllegalArgumentException(String.format("No converter found for type '%s'", originalType.getName()));
+            }
```
dcapwell, line 138: this section isn't needed.  Why are we trying to support saving the wrong type?  This adds a lot of complexity and possible bugs, we should keep this super simple.

```
Object oldValue = property.getValue();
             propertyChangeListeners.get(ConfigurationListener.ChangeType.BEFORE).fireTypeCast(property.getName(), oldValue, convertedValue);
             property.setValue(convertedValue);
             propertyChangeListeners.get(ConfigurationListener.ChangeType.AFTER).fireTypeCast(property.getName(), oldValue, convertedValue);
             // This potentially may expose the values that are not safe to see in logs on production.
             logger.info("Property '{}' updated from '{}' to '{}'.", property.getName(), oldValue, convertedValue);
```

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,401 @@
...
+            // Do validation first for converted new value.
+            List<TypedConstraintAdapter<?>> constraintsList = constraints.getOrDefault(property.getName(), Collections.emptyList());
+            for (TypedConstraintAdapter<?> typed : constraintsList)
+                typed.validateTypeCast(convertedValue);
```
dcapwell, line 142: Why do we need this?  We have been moving towards types doing their own validation, so custom validators adds more complexity if they are "lazy".

I am actually ok if this exists and is done via the following

```
// Config.java
@Mutable
@Min(42) @Max(57)
public volatile int some_config = 42;
```

but custom validation has been something we are trying to move away from... I don't see this style in this patch, so questioning...

Also, this concept duplicates `propertyChangeListeners.get(ConfigurationListener.ChangeType.BEFORE).fireTypeCast(property.getName(), oldValue, convertedValue);`, so you could create a much simpler model by having any custom validation done as a before listener

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,401 @@
...
+
+
+/**
+ * This is a simple configuration property registry that stores all the {@link Config} settings, it doesn't
+ * take into account any configuration changes that might happen during properties replacement between releases.
+ */
+public class ConfigurationRegistry implements Registry
+{
+    private static final Logger logger = LoggerFactory.getLogger(ConfigurationRegistry.class);
+    private final ReadWriteLock rwLock = new ReentrantReadWriteLock();
```
dcapwell, line 59: this class is more complex than it needs to due to these locks... I strongly believe that we know all properties/"validations" at object creation, so those should be final and immutable... this makes it so the only *mutable* field is the listeners, which can use standard concurrent collections to avoid needing to manage our locks ourselves; they also make it so we can have stronger concurrent access to this class, right now contention can add up slowing down the system.

dcapwell (changes requested):

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -80,45 +82,54 @@ public static Set<String> splitCommaDelimited(String src)
...
-    @Replaces(oldName = "roles_update_interval_in_ms", converter = Converters.MILLIS_CUSTOM_DURATION, deprecated = true)
-    public volatile DurationSpec.IntMillisecondsBound roles_update_interval = null;
-    public volatile boolean roles_cache_active_update = false;
-    @Replaces(oldName = "credentials_validity_in_ms", converter = Converters.MILLIS_DURATION_INT, deprecated = true)
-    public volatile DurationSpec.IntMillisecondsBound credentials_validity = new DurationSpec.IntMillisecondsBound("2s");
-    public volatile int credentials_cache_max_entries = 1000;
-    @Replaces(oldName = "credentials_update_interval_in_ms", converter = Converters.MILLIS_CUSTOM_DURATION, deprecated = true)
-    public volatile DurationSpec.IntMillisecondsBound credentials_update_interval = null;
-    public volatile boolean credentials_cache_active_update = false;
+    @Mutable @Replaces(oldName = "permissions_validity_in_ms", converter = Converters.MILLIS_DURATION_INT, deprecated = true)
```
dcapwell: by putting `@Mutable` on the same line as `@Replaces` this diff gets hard to review as I have to check every character in `@Replaces` to make sure nothing changed... if you keep then in different lines the diff is much cleaner

src/java/org/apache/cassandra/config/DataStorageSpec.java
```diff
@@ -207,6 +207,14 @@ public long toBytes()
...
             return unit().toBytes(quantity());
         }
+
+        /**
+         * @return the amount of data storage in mebibytes.
+         */
+        public long toMebibytes()
+        {
+            return unit().toMebibytes(quantity());
+        }
```
dcapwell: unrelated changes must be removed as per our style guide

src/java/org/apache/cassandra/config/registry/Registry.java
```diff
@@ -0,0 +1,95 @@
...
+    <T> T get(Class<T> cls, String name);
+
+    /**
+     * Get property's value by name and convert it to the String type. The exception will be thrown if the property
+     * is not present in the registry.
+     *
+     * @param name Property name.
+     * @return Property's value converted to String.
+     */
+    String getString(String name);
```
dcapwell, line 67: we should remove this API as it is 100% only for SettingsTable, leave this problem for the SettingsTable

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
```diff
@@ -0,0 +1,401 @@
...
+        return properties.get(name).isWritable();
+    }
+
+    /**
+     * Adds a listener for the property with the given name.
+     * @param name property name to listen to.
+     * @param listener listener to add.
+     * @param <T> type of the property.
+     */
+    public <T> void addPropertyChangeListener(String name, ConfigurationListener.ChangeType type, ConfigurationListener<T> listener, Class<T> listenerType)
```
dcapwell, line 256: if you listen to a property that is `@Replaced` what *should* happen?  The common case would be a bug (we forgot to update), but there *might* be a case where you want to see if this is happening (don't know of one)...

Maybe we should detect this and fail fast, force the caller to say "I know what I am doing, I want the old name!"?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1504,137 +1599,131 @@ public static void setRoleManager(IRoleManager roleManager)
 
     public static int getPermissionsValidity()
     {
-        return conf.permissions_validity.toMilliseconds();
+        return getProperty(DurationSpec.IntMillisecondsBound.class, ConfigFields.PERMISSIONS_VALIDITY).toMilliseconds();
```
dcapwell: please revert, `getProperty` is very expensive, and this is all for internal access...

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1504,137 +1599,131 @@ public static void setRoleManager(IRoleManager roleManager)
...
     }
 
     public static int getRolesCacheMaxEntries()
     {
-        return conf.roles_cache_max_entries;
+        return getProperty(Integer.TYPE, ConfigFields.ROLES_CACHE_MAX_ENTRIES);
     }
 
-    public static int setRolesCacheMaxEntries(int maxEntries)
+    public static void setRolesCacheMaxEntries(int maxEntries)
```
dcapwell, line 1845: please revert, unrelated code changes goes against our style guide

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -3105,12 +3193,14 @@ public static Set<String> hintedHandoffDisabledDCs()
 
     public static boolean useDeterministicTableID()
     {
-        return conf != null && conf.use_deterministic_table_id;
+        if (confRegistry == null)
```
dcapwell: please revert

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -4676,4 +4729,41 @@ public static void setClientRequestSizeMetricsEnabled(boolean enabled)
...
+                return new IllegalStateException(rt.getMessage());
+            else if ((rt = cause(e, UnsupportedOperationException.class)) != null)
+                return new UnsupportedOperationException(rt.getMessage());
+            else
+            {
+                logger.error("Unexpected exception", e);
+                return new RuntimeException(e.getMessage());
+            }
+        }
+    }
```
dcapwell, line 4986: Why does this exist?

dcapwell (changes requested):

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
Mmuzaf: It might be true, but according to my understanding the publication of configuration variables is safe for the following reasons:
- we read the volatile `initialized` variable each time the `lazyInit();` accessed;
- we do a safe publication for the `properties` variable as the ImmutableMap and double-check locking are used for; 

We have `happens-before` here and will see the non-null `PropertyAdapter property` (if it exists).

src/java/org/apache/cassandra/config/registry/PrimitiveUnaryConverter.java
Mmuzaf: We have a collection of `Config` properties. Some of these properties can be of a `boolean` type, while the others are of a `Boolean` (or we can take any pair from primitives and their wrappers). We always return a wrapper object from the `Registry` when the `<T> T get(Class<T> cls, String name)` method is called, regardless of whether the primitive or wrapper for the property is set.

For both of the cases primitive or wrapper, we have to be sure that we can cast to the right type when the `get` method is called  - and this is the issue hiding under the hood and the purpose of this method.

src/java/org/apache/cassandra/db/virtual/SettingsTable.java
Mmuzaf: It can't break anything because the SettingsTable is not updatable now, so it's up to us in this patch to decide which set of properties to make updatable :-)

AFAIU, the `Replacement` is used to support backward compatibility between the releases so that users can use their old configs with new releases smoothly keeping in mind that they have to migrate to a new version of the `cassandra.yaml` (e.g. export the real config from the cluster to the yaml file).

So showing a replaced name in the SettingsTable might be OK, but if we make it updatable it might cause confusion - because both the replaced name and a new name are responsible for the same single thing,  and this does not lead the user to understand which is the _right_ property in the release he uses.

If I'm missing something, just let me know and I'll fix it (it is easy).

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
Mmuzaf: The problem I was trying to solve with lazy initialization of the _ConfigurationRegistry_ is that when the `DatabaseDescriptor#toolInitialization(boolean)` is called it will initialize almost all the components affected by access through `defaultLoader().flatten(Config.class)` which pulls dozens of classes to be load. 
This was highlighted by the `DatabaseDescriptorRefTest` test.

src/java/org/apache/cassandra/config/Config.java
Mmuzaf: As we decided to leave access DatabaseDescriptor's getters 'as is' for performance reasons, no need to remove the `volatile` keyword for configuration properties. I'll do revert these changes.

src/java/org/apache/cassandra/config/Mutable.java
Mmuzaf: We can do it, but without real use it might look strange, so I suggest we extend it when we come up with such methods.

src/java/org/apache/cassandra/config/registry/ConfigurationConstraint.java
Mmuzaf: As we have discussed, we need to find a robust way of validating the configuration fields. This interface will provide a way to build a custom validation for each field that will be used to check the input, regardless of what type of user interface is being used: JMX, Yaml, SettingsTable. 
I will start a public discussion on the dev list and try to find common ground.

src/java/org/apache/cassandra/config/StringConverters.java
Mmuzaf: I think you're right here, we don't need such complexity with `Enum`s and the `Map` is the easiest solution for us.

src/java/org/apache/cassandra/db/virtual/SettingsTable.java
Mmuzaf: There are a few reasons under the hood to move away from the `ForwardingProperty` logic for the `SettingsTable`:
- The design assumes that we have a central point of access to the `Config' instance - the `ConfigurationRegistry' - to perform mutate configuration operations, ensuring that they are changed in the same way for each public interface;
- We want to restrict the places where the `Config' instance is accessed - the `DatabaseDescriptor' and the `ConfigurationRegistry' are enough to force the `SettingsTable` to do what it needs to do;
- the `Replacement` provides us with the name and type conversions for properties and as there are no usages of the old names in the source code we are able to perform name conversions for the `SettingsTable` needs;

Maybe I'm missing something, but I see the following access path as more verbose than giving short cuts from the `SettingsTable' directly to the `Config':
Config -> DatabaseDescription -> ConfigurationRegistry -> ReplacementRegistry -> SettingsTable;

src/java/org/apache/cassandra/service/StorageService.java
Mmuzaf: I tend to agree here, but my concern here is that someone is doing something wrong if someone relies on log messages. 

The `ConfigurationRegistry` will log any changes to the property's values, but the log message will look slightly different. Anyway, a user will still be able to see it in the logs.

src/java/org/apache/cassandra/service/StorageService.java
Mmuzaf: I agree with you - this is not the right place to have property constraints, so we will move it to the validation subsystem once the discussion on the ML is finished.

The reason why it exists is simple - the type `DurationSpec.LongMillisecondsBound` itself provides only `<= 0` guarantees, however, the property requires strickly `> 0`.

src/java/org/apache/cassandra/config/registry/Registry.java
Mmuzaf: This is actually not true. The `SettingsTable` is not the only system that does type conversions to and from the `String` type. 

For example, the same is also performed for JMX:

```
    public String getPaxosOnLinearizabilityViolations()
    {
        return DatabaseDescriptor.paxosOnLinearizabilityViolations().toString();
    }

    public void setPaxosOnLinearizabilityViolations(String v)
    {
        DatabaseDescriptor.setPaxosOnLinearizabilityViolations(Config.PaxosOnLinearizabilityViolation.valueOf(v));
        logger.info("paxos on linearizability violations {} via jmx", v);
    }
```

src/java/org/apache/cassandra/config/registry/ConfigurationRegistry.java
Mmuzaf: In my view, if you change the name of a property, you must also update the corresponding listener usage. I propose to enforce this rule and avoid supporting different names for the same properties in internal code, so only limited subsystems will work with the `Replaced` annotation and know the old names.

The `fail fast` rule will work once the `addPropertyChangeListener` is fired with an incorrect property name.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
Mmuzaf: Agree, let's move this and all such changes out of the scope.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
Mmuzaf: The reason for this exception conversion is that, for example, the JMX client may not know about Cassandra classes (missing in its classpath), so it will fail if the `ConfigurationRegistry` throws an exception that has them. 

Here is an example of a potential issues:
https://issues.apache.org/jira/browse/CASSANDRA-17527

I just re-think the usage once will find a common ground for the properties validation process on the ML.

PR #2143, opened 2023-02-06
[CASSANDRA-17979] Remove invalid handshake test assertion

dchenbecker:
The call to hasPending is not dependent on the isConnected call, and
the test is to ensure that handshake (e.g. connection) works, not that
messages are flushed. Instead, we loop on hasPending after connection
to ensure that messages are flushed. This affects both the
"testOutboundConnectionDoesntFallbackWhenErrorIsNotSSLRelated" and
"testOutboundFallbackOnSSLHandshakeFailure" tests which have similar
semantics.

Fixes CASSANDRA-17979

patch by Derek Chen-Becker; reviewed by Brandon Williams for CASSANDRA-17979

### DIFF
--- a/test/unit/org/apache/cassandra/net/HandshakeTest.java
+++ b/test/unit/org/apache/cassandra/net/HandshakeTest.java
@@ -19,6 +19,8 @@
 package org.apache.cassandra.net;
 
 import java.nio.channels.ClosedChannelException;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


test/unit/org/apache/cassandra/net/HandshakeTest.java
```diff
@@ -377,24 +378,35 @@ private void testOutboundFallbackOnSSLHandshakeFailure(SslFallbackConnectionType
...
-        while (!outboundConnection.isConnected() && System.currentTimeMillis() - startTime < 60000)
+        Instant endTime = Instant.now().plus(DELIVERY_WAIT_DURATION);
+
+        while (!outboundConnection.isConnected() && Instant.now().isBefore(endTime))
         {
             Thread.sleep(1000);
         }
+
+        // Now that we're connected, we also want to ensure that the delivery thread succeeded
+        while (outboundConnection.hasPending() && Instant.now().isBefore(endTime)) {
```
jyothsnakonisa, line 403: I think we just want to verify that the connection is established and we don't have to wait for pending connections to be flushed. We might be waiting indefinitely for the pending connections to be flushed and the test might fail because of timeout.

dchenbecker: I had made that argument in my original commit. @ekaterinadimitrova2 had pushed back and asked to also confirm delivery, so I made the change. Can you discuss in the thread and see what we can all agree on?

ekaterinadimitrova2: `We might be waiting indefinitely`
@jyothsnakonisa  Can you elaborate a bit, please? I suspect I might be missing something

dchenbecker:
This was merged separately, so I'm closing this PR

PR #2146, opened 2023-02-07
CASSANDRA-18070 trunk: Add SELECT_MASKED permission

adelapena:
Add new SELECT_MASKED permission. It allows to run SELECT queries filtering the clear values of masked columns.

Superusers have it by default, whereas regular users don't have it by default.

Users without UNMASK nor SELECT_MASKED permission won't be authorised to use masked columns on the WHERE part of a SELECT query.

### DIFF
--- a/doc/modules/cassandra/pages/cql/cql_singlefile.adoc
+++ b/doc/modules/cassandra/pages/cql/cql_singlefile.adoc
@@ -3611,6 +3611,7 @@ or not.
 |`LIST` |no
 |`LOGIN` |no
 |`MAP` |no
+|`MASKED` |no
 |`MATERIALIZED` |yes
 |`MBEAN` |yes
 |`MBEANS` |yes
--- a/doc/modules/cassandra/pages/cql/cql_singlefile.adoc
+++ b/doc/modules/cassandra/pages/cql/cql_singlefile.adoc
@@ -2262,6 +2262,7 @@ The full set of available permissions is:
 * `AUTHORIZE`
 * `DESCRIBE`
 * `EXECUTE`
+* `UNMASK`
 
 Not all permissions are applicable to every type of resource. For
 instance, `EXECUTE` is only relevant in the context of functions or
@@ -2420,6 +2421,12 @@ use of function in `CREATE AGGREGATE` | | |
 wildcard pattern | | |
 
 |`EXECUTE` |`MBEAN` |Execute operations on named mbean | | |
+
+|`UNMASK` |`ALL KEYSPACES` |See the clear contents of masked columns on any table | | |
+
+|`UNMASK` |`KEYSPACE` |See the clear contents of masked columns on any table in keyspace | | |
+
+|`UNMASK` |`TABLE` |See the clear contents of masked columns on the specified table | | |
 |===
 
 [[grantPermissionsStmt]]
@@ -2430,7 +2437,7 @@ _Syntax:_
 bc(syntax).. +
 ::= GRANT ( ALL ( PERMISSIONS )? | ( PERMISSION )? ) ON TO
 
-::= CREATE | ALTER | DROP | SELECT | MODIFY | AUTHORIZE | DESRIBE |
+::= CREATE | ALTER | DROP | SELECT | MODIFY | AUTHORIZE | DESRIBE | UNMASK
 EXECUTE
 
 ::= ALL KEYSPACES +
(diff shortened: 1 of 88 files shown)


src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
```diff
@@ -238,7 +239,21 @@ public void authorize(ClientState state) throws InvalidRequestException, Unautho
...
             state.ensurePermission(Permission.EXECUTE, function);
 
-        unmask = state.hasUnmaskPermission(table);
+        unmask = state.hasTablePermission(table, Permission.UNMASK);
+
+        if (!unmask && !state.hasTablePermission(table, Permission.SELECT_MASKED))
+        {
+            List<ColumnMetadata> queriedMaskedColumns = table.columns()
+                                                             .stream()
+                                                             .filter(ColumnMetadata::isMasked)
```
bereng, line 251: Would it be better to filter `isRestricted` first?

test/unit/org/apache/cassandra/cql3/functions/masking/QueryMaskedPermissionTest.java
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    @Test
+    public void testPartitionKeyColumn() throws Throwable
+    {
+        alterTable("ALTER TABLE %s ALTER k MASKED WITH DEFAULT");
+        Object[] maskedRow = row(0, 7, 7, 7);
+
+        // test queries with default permissions (no UNMASK nor SELECT_MASKED)
+        assertWithAuthSpin(() -> {
```
bereng: If my eyes aren't fooling me aren't we just repeating the same queries over and over again? Did you do this intentionally or would you consider some type of looping to avoid the repeats?

src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
adelapena: [`StatementRestrictions#isRestricted`](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/restrictions/StatementRestrictions.java#L341-L346) involves building a collection [in most cases](https://github.com/apache/cassandra/blob/cassandra-4.1.0/src/java/org/apache/cassandra/cql3/restrictions/SingleColumnRestriction.java#L60) and then running a `List#contains`. In contrast, {{ColumnMetadata#isMasked}} just returns a boolean. So I think `isMasked` is faster than is `isRestricted`.

test/unit/org/apache/cassandra/cql3/functions/masking/QueryMaskedPermissionTest.java
adelapena: I think that the queries tested of every block are different to each other. And the blocks are different to each other on whether they return a masked or unmasked row, or an authorization failure. The only blocks that are repeated are the ones for only `UNMASK` permission and for both `UNMASK` and `SELECT_MASKED` permissions, and the ones for defaults and manually removed permissions. I can put those pairs on blocks on separate methods to save some duplication, although I'm not sure that actually improves readability. 

As for the differences test to test, they might look similar but they aren't identical, since different schemas allow different queries.

src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
bereng: sgtm

doc/modules/cassandra/pages/cql/cql_singlefile.adoc
```diff
@@ -2427,6 +2428,12 @@ wildcard pattern | | |
 |`UNMASK` |`KEYSPACE` |See the clear contents of masked columns on any table in keyspace | | |
 
 |`UNMASK` |`TABLE` |See the clear contents of masked columns on specified table | | |
+
+|`SELECT_MASKED` | `ALL KEYSPACES` | `SELECT` restricting masked columns on any table | | |
+
+|`SELECT_MASKED` | `KEYSPACE` | `SELECT` restricting masked columns on any table in specified keyspace | | |
+
+|`SELECT_MASKED` | `TABLE` | `SELECT` restricting masked columns on specified table | | |
```
bereng: Wording?

blerer (approved):

pmcfadin:
Closing as [CASSANDRA-18070](https://issues.apache.org/jira/browse/CASSANDRA-18070) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #2154, opened 2023-02-10
18252 trunk 4

ekaterinadimitrova2:
CASSANDRA-18252 Remove Scripted UDFs from trunk to enable easier JDK17 testing on current trunk


src/java/org/apache/cassandra/cql3/functions/UDFunction.java
```diff
@@ -266,13 +266,7 @@ public static UDFunction create(FunctionName name,
     {
         assertUdfsEnabled(language);
 
-        switch (language)
-        {
-            case "java":
-                return new JavaBasedUDFunction(name, argNames, argTypes, returnType, calledOnNullInput, body);
-            default:
-                return new ScriptBasedUDFunction(name, argNames, argTypes, returnType, calledOnNullInput, language, body);
```
michaelsembwever, line 274: should the `src/java/org/apache/cassandra/cql3/functions/ScriptBasedUDFunction.java` file also be deleted?

test/conf/cassandra-murmur.yaml
```diff
@@ -39,6 +39,6 @@ compaction_throughput: 0MiB/s
 row_cache_class_name: org.apache.cassandra.cache.OHCProvider
 row_cache_size: 16MiB
 user_defined_functions_enabled: true
-scripted_user_defined_functions_enabled: true
+scripted_user_defined_functions_enabled: false
```
michaelsembwever, line 42: should we just delete all these `scripted_user_defined_functions_enabled: false` lines from test cassandra*.yaml files?
(since the deprecated default is `false` )

src/java/org/apache/cassandra/cql3/functions/UDFunction.java
ekaterinadimitrova2: Totally... just it was forgotten... I will take care, thanks

test/conf/cassandra-murmur.yaml
ekaterinadimitrova2: We can but I wasn't completely sure how things will go in the other ticket as this is only deprecated now in Config class. (And in general there is this tendancy to leave config in the Config class deprecated, maybe even forever? I hear different opinions)
Also, wanted to be sure during testing we don't get sabotaged getting the value from somewhere else that I miss, so flipping at first to be sure they are disabled and fail sounded as a good first choice

michaelsembwever: ok to leave. i trust your judgement.

ekaterinadimitrova2: Thanks, I will put a note on the other ticket that depending on how it goes we might want to remove them (we should if there are no more reasons to keep them, decluttering is important)

michaelsembwever (approved):

PR #2163, opened 2023-02-16
CASSANDRA-18012 trunk: CircleCI: Remove -l/-m/-h flags and have two options: free or paid

adelapena:
Remove -l/-m/-h flags in CircleCI config generation script and replace them by -f/-p, meant to be used with CircleCI's free/paid tiers:
* -f is equivalent to the previous -l
* -p is equivalent to the previous -m
* The previous -h is just gone, together with the HIGHRES config.
* The previous -f flag is replaced by -i (ignore unknown env vars)


.circleci/generate.sh
```diff
@@ -30,13 +30,12 @@ die ()
...
 {
-  echo "Usage: $0 [-l|-m|-h|-f|-e]"
-  echo "   -a Generate the default config.yml using low resources and the three templates"
-  echo "      (config.yml.LOWRES, config.yml.MIDRES and config.yml.HIGHRES). Use this for"
-  echo "      permanent changes in config-2_1.yml that will be committed to the main repo."
-  echo "   -l Generate config.yml using low resources"
-  echo "   -m Generate config.yml using mid resources"
-  echo "   -h Generate config.yml using high resources"
+  echo "Usage: $0 [-f|-p|-a|-e|-i]"
+  echo "   -a Generate the default config.yml using resources compatible with CircleCI's free tier"
```
jmckenzie-dev: nit: referring to `config.yml` on -a, -f, and -p muddies it up and it's not immediately clear how the relationship works. Noodling on this a bit, maybe something like:
```
-a Generate a new default config.yml to be committed after changes to the template <etc>
-f Generate a config.yml for tests compatible with the CircleCI free tier resources
-f Generate a config.yml for tests compatible with the CircleCI paid tier resources
```

It's subtle, but forklifting the "hey, we're generating a new default config to be committed" up front as well as the slight massaging on -f and -p imply the relationship between these params and the workflow for the config in a way the phrasing here doesn't.

Maybe. Take it or leave it. 😀 I know how it works and had to squint a bit and think hard to make sure I grokked the difference though.

.circleci/generate.sh
```diff
@@ -145,52 +141,38 @@ if $has_env_vars && $check_env_vars; then
...
-elif $midres; then
-  ($all || $lowres || $highres) && die "Cannot use option -m with options -a, -l or -h"
-  echo "Generating new config.yml file with middle resources from config-2_1.yml"
-  patch -o $BASEDIR/config-2_1.yml.MIDRES $BASEDIR/config-2_1.yml $BASEDIR/config-2_1.yml.mid_res.patch
-  circleci config process $BASEDIR/config-2_1.yml.MIDRES > $BASEDIR/config.yml.MIDRES.tmp
-  cat $BASEDIR/license.yml $BASEDIR/config.yml.MIDRES.tmp > $BASEDIR/config.yml
-  rm $BASEDIR/config-2_1.yml.MIDRES $BASEDIR/config.yml.MIDRES.tmp
+if $free; then
+  ($all || $paid) && die "Cannot use option -f with options -a or -p"
+  echo "Generating new config.yml file for free tier from config-2_1.yml"
```
jmckenzie-dev: For my own learning: why is this file named `config-2_1.yml`. 2.1 of what?

.circleci/generate.sh
```diff
@@ -145,52 +141,38 @@ if $has_env_vars && $check_env_vars; then
...
 elif $all; then
-  ($lowres || $midres || $highres || $has_env_vars) && die "Cannot use option -a with options -l, -m, -h or -e"
-  echo "Generating new config.yml file with low resources and LOWRES/MIDRES/HIGHRES templates from config-2_1.yml"
-
-  # setup lowres
-  circleci config process $BASEDIR/config-2_1.yml > $BASEDIR/config.yml.LOWRES.tmp
-  cat $BASEDIR/license.yml $BASEDIR/config.yml.LOWRES.tmp > $BASEDIR/config.yml.LOWRES
-  rm $BASEDIR/config.yml.LOWRES.tmp
+  ($free || $paid || $has_env_vars) && die "Cannot use option -a with options -f, -p or -e"
+  echo "Generating new default config.yml file for free tier and FREE/PAID templates from config-2_1.yml"
```
jmckenzie-dev: May want to echo out the whole "make sure you commit the newly generated files <list_files_here> after running this command if you want them to persist"

.circleci/readme.md
```diff
@@ -25,22 +25,27 @@ The file `config.yml` is the configuration file that is read by CircleCI. This f
 automatically generated by the `generate.sh` script from the `config-2_1.yml` file.
 
 The provided `config.yml` file uses low resources so users of the CircleCI free tier can
```
jmckenzie-dev, line 27: nit... ish?

Maybe we could just phrase this as something like:
```
The provided default `config.yaml` file targets a subset of tests that succeed on the free tier of CircleCI; this is a duplicate of the config.yml.FREE file for generating purposes. There is also an optional config.yml.PAID file that includes more test suites that require the higher resources available on premium CircleCI accounts, which we notably also use for validating new releases.
```

.circleci/readme.md
```diff
@@ -149,44 +154,37 @@ default medium config and Large containers.
 ## Updating the config
 For configuration changes meant to be permanent in the Apache repo you should never edit
 the `config.yml` file manually. Instead, you should edit the `config-2_1.yml` file and then
```
jmckenzie-dev: So to my earlier question, if we could name this something like `config-template.yml` it'd make the relationship between things a lot more innately clear I think.

.circleci/generate.sh
ekaterinadimitrova2: I was asking myself the same question the other day. I _think_ it might be coming from which version in C* it was introduced and it kept rolling as... why not? :D Kind of just left alone...

adelapena: I don't know for sure, but I think it refers to the version of CircleCI's config that it contains, which is 2.1 ([first line of the file](https://github.com/apache/cassandra/blob/trunk/.circleci/config-2_1.yml#L19)). 

The `config.yml` file contains the expanded config that is automatically generated by `circle config process`. That file however [uses version 2.0](https://github.com/apache/cassandra/blob/trunk/.circleci/config.yml#L19) instead of 2.1, don't ask me why. It has to be named just `config.yml` so Circle is able to identify it.

.circleci/generate.sh
adelapena: Makes sense. I have tried to rephrase using CircleCI terminology, where `config.yml` is the "expanded config" and `config-2_1.yml` is the "reusable config".

.circleci/generate.sh
adelapena: Added a line about that.

.circleci/readme.md
adelapena: +1 to renaming the file, even if it's only because I never remember where to put the hyphen and where the underscore in `config-2_1.yml`.

Regarding the new name for the file, I'm not sure if calling it a template is clear enough, and we also refer to the FREE/PAID configs as templates. Maybe it's the main config file, or the manually-generated file? In CircleCI terminology, `config.yml` is the "expanded config" and `config-2_1.yml` is the "reusable config". For us, `config-2_1.yml` is the file that we manually edit, and `config.yml` is the one that we automatically generate.

.circleci/readme.md
adelapena: I think we can update this when we actually remove the tests that require paid resources from `config.yml.FREE`. I was thinking on doing that on a separate ticket, so we get rid of HIGHRES as soon and painlessly as possible, since that eases any further improvement. wdyt?

jmckenzie-dev (approved):

jmckenzie-dev (approved):

.circleci/readme.md
jmckenzie-dev: > even if it's only because I never remember where to put the hyphen and where the underscore in config-2_1.yml.
This. 1000x this.

Don't have any ideas beyond referring to it as "template". Could also refer to all three with .template in them or something?

.circleci/readme.md
jmckenzie-dev: 👍

.circleci/readme.md
adelapena: I cannot think on a better name either, and "template" is definitively better that "2_1", so I'm renaming it.

PR #2170, opened 2023-02-17
CASSANDRA-18129 review PR

aweisberg:
A number of dependencies need to land before live migration such as transactional metadata, range transactions in Cassandra, immutable state refactor, so for review purposes this PR is against the sync point work that Benedict did that is also blocked.


aweisberg:
Replaced by https://github.com/apache/cassandra/pull/2237

PR #2173, opened 2023-02-22
CASSANDRA-18204: Switched away from https endpoints for git submodules in favor of relative paths during development

dcapwell:


.build/sh/change-submodule.sh
```diff
@@ -23,6 +23,11 @@ set -o nounset
 _usage() {
   cat <<EOF
 Usage: $(basename $0) [submodule path] [git repo URL] [git branch]
+
+Examples:
+
+Switch to branch immutable-state
```
maedhroz: nitty nit: Maybe quotes around `immutable-state` for clarity...not super important

maedhroz (approved):

PR #2177, opened 2023-02-25
docs: add link to PHP extension

maxm86545:
- Outdated drivers for PHP are removed.
- Added a link to actual PHP extension.

polandll: Fix required for trunk as well, I suspect.


bschoening:
Closing this because it is 2 years old and needs to be fixed in trunk.  @maxm86545 if wish to pickup [CASSANDRA-20573](https://issues.apache.org/jira/browse/CASSANDRA-20573) it would be great to fix with those other issue.

PR #2186, opened 2023-03-02
Improve memtable allocator accounting when updating AtomicBTreePartition

jonmeredith:
See CASSANDRA-18125

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -570,7 +570,7 @@
             <exclusion groupId="org.slf4j" artifactId="slf4j-api"/>
             <exclusion groupId="org.codehaus.jackson" artifactId="jackson-mapper-asl"/>
           </dependency>
-          <dependency groupId="net.java.dev.jna" artifactId="jna" version="5.6.0"/>
+          <dependency groupId="net.java.dev.jna" artifactId="jna" version="5.13.0"/>
 
           <dependency groupId="org.jacoco" artifactId="org.jacoco.agent" version="${jacoco.version}" scope="test"/>
           <dependency groupId="org.jacoco" artifactId="org.jacoco.ant" version="${jacoco.version}" scope="test"/>
--- a/src/java/org/apache/cassandra/db/rows/NativeCell.java
+++ b/src/java/org/apache/cassandra/db/rows/NativeCell.java
@@ -72,7 +72,7 @@ public class NativeCell extends AbstractCell<ByteBuffer>
                       CellPath path)
     {
         super(column);
-        long size = simpleSize(value.remaining());
+        long size = offHeapSizeWithoutPath(value.remaining());
 
         assert value.order() == ByteOrder.BIG_ENDIAN;
         assert column.isComplex() == (path != null);
@@ -105,7 +105,7 @@ public class NativeCell extends AbstractCell<ByteBuffer>
         }
     }
 
-    private static long simpleSize(int length)
+    private static long offHeapSizeWithoutPath(int length)
     {
         return VALUE + length;
     }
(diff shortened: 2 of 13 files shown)


test/unit/org/apache/cassandra/db/partitions/AtomicBTreePartitionMemtableAccountingTest.java
```diff
@@ -0,0 +1,398 @@
...
+        updateRowBuilder.addCell(makeCell(regular, updateTS, updateTTL, updateLDT, updateValueBB, null));
+        if (updateComplexDeletionTime != DeletionTime.LIVE)
+            updateRowBuilder.addComplexDeletion(complex, updateComplexDeletionTime);
+
+        // Make multiple update cells to make any issues more pronounced
+        cellPath = 1000;
+        for (int i = 0; i < numC2UpdateCells; i++)
+            updateRowBuilder.addCell(makeCell(complex, updateTS, updateTTL, updateLDT,
+                                              ByteBufferUtil.EMPTY_BYTE_BUFFER,
+                                              CellPath.create(ByteBufferUtil.bytes(cellPath++))));
```
jonmeredith: @blerer   https://github.com/apache/cassandra/commit/79f46d2e4035f05678aa98f2ab909117d613ff70#r102899481
> nits: It seems that the 2 blocks are the same if I am not mistaken. They could be refactored as a method

The difference is the direction the cellPaths go to make some distinct and some overlapping elements. It could be factored out but felt like a lot more passing settings around.

src/java/org/apache/cassandra/db/rows/NativeCell.java
```diff
@@ -171,4 +171,16 @@ public long unsharedHeapSizeExcludingData()
         return EMPTY_SIZE;
     }
 
+    public long offHeapSize()
+    {
+        long size = simpleSize(MemoryUtil.getInt(peer + LENGTH));
```
jonmeredith: @blerer https://github.com/apache/cassandra/commit/79f46d2e4035f05678aa98f2ab909117d613ff70#r102910076
> This `simpleSize` name is really confusing. I think we should inline the method. It would be much clearer.

Also used in the constructor so want to keep it DRY. Could rename to `offHeapSizeWithoutCellPath`?

test/unit/org/apache/cassandra/db/partitions/AtomicBTreePartitionMemtableAccountingTest.java
```diff
@@ -0,0 +1,398 @@
...
+    {
+        if (localDeletionTime != Cell.NO_DELETION_TIME) // never a ttl for a tombstone
+        {
+            ttl = Cell.NO_TTL;
+            value = ByteBufferUtil.EMPTY_BYTE_BUFFER;
+        }
+        return new BufferCell(column, timestamp, ttl, localDeletionTime, value, path);
+    }
+
+    void testCase(int initialTS, int initialTTL, int initialLDT, DeletionTime initialCDT, int numC2InitialCells,
```
jonmeredith: @blerer https://github.com/apache/cassandra/commit/79f46d2e4035f05678aa98f2ab909117d613ff70#r102899871
> nit: Would it not be more readable to have some object to bundle all those information together.

build.xml
```diff
@@ -570,7 +570,7 @@
             <exclusion groupId="org.slf4j" artifactId="slf4j-api"/>
             <exclusion groupId="org.codehaus.jackson" artifactId="jackson-mapper-asl"/>
           </dependency>
-          <dependency groupId="net.java.dev.jna" artifactId="jna" version="5.6.0"/>
+          <dependency groupId="net.java.dev.jna" artifactId="jna" version="5.13.0"/>
```
jonmeredith, line 573: will drop this from the patch - accidentally committed

src/java/org/apache/cassandra/db/rows/NativeCell.java
blerer: Yes, it sounds much better.

build.xml
blerer: I wanted to ask about it.

jonmeredith:
Merged.

PR #2187, opened 2023-03-02
Improve memtable allocator accounting when updating AtomicBTreePartition

jonmeredith:
See CASSANDRA-18125


jonmeredith:
Merged.

PR #2189, opened 2023-03-02
CASSANDRA-18294 3.0 Fix disk_failure_policy die to stop gossip, native transport and JVM

Runtian:
To reproduce the issue, switch to disk_failure_policy die

add some code to trigger FSError.

trigger the FSError and Cassandra process will not stop.

With this fix, Cassandra process will stop.

```
Fix default file system error handler for disk_failure_policy die

patch by Runtian Liu; reviewed by TBD for CASSANDRA-18294

```

The [Cassandra Jira](https://issues.apache.org/jira/browse/CASSANDRA-18294)

### DIFF
--- a/src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java
+++ b/src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java
@@ -44,6 +44,7 @@ public class DefaultFSErrorHandler implements FSErrorHandler
 
         switch (DatabaseDescriptor.getDiskFailurePolicy())
         {
+            case die:
             case stop_paranoid:
                 // exception not logged here on purpose as it is already logged
                 logger.error("Stopping transports as disk_failure_policy is " + DatabaseDescriptor.getDiskFailurePolicy());
@@ -60,6 +61,7 @@ public class DefaultFSErrorHandler implements FSErrorHandler
 
         switch (DatabaseDescriptor.getDiskFailurePolicy())
         {
+            case die:
             case stop_paranoid:
             case stop:
                 // exception not logged here on purpose as it is already logged
--- a/test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
+++ b/test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
@@ -77,7 +77,8 @@ public class DefaultFSErrorHandlerTest
 
     public DefaultFSErrorHandlerTest(Config.DiskFailurePolicy policy,
                                      boolean gossipRunningFSError,
-                                     boolean gossipRunningCorruptedSStableException) {
+                                     boolean gossipRunningCorruptedSStableException)
+    {
         this.testDiskPolicy = policy;
         this.gossipRunningFSError = gossipRunningFSError;
         this.gossipRunningCorruptedSStableException = gossipRunningCorruptedSStableException;
--- a/test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
+++ b/test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
@@ -1,0 +1,120 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.service;
+
+import java.io.IOException;
+import java.util.Arrays;
+import java.util.Collection;
+
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import org.apache.cassandra.SchemaLoader;
+import org.apache.cassandra.config.Config;
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.exceptions.ConfigurationException;
+import org.apache.cassandra.gms.Gossiper;
+import org.apache.cassandra.io.FSErrorHandler;
+import org.apache.cassandra.io.FSReadError;
+import org.apache.cassandra.io.sstable.CorruptSSTableException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(Parameterized.class)
+public class DefaultFSErrorHandlerTest
+{
+    private FSErrorHandler handler = new DefaultFSErrorHandler();
+    Config.DiskFailurePolicy oldDiskPolicy;
+    Config.DiskFailurePolicy testDiskPolicy;
+    private boolean gossipRunningFSError;
+    private boolean gossipRunningCorruptedSStableException;
+
+    @BeforeClass
+    public static void defineSchema() throws ConfigurationException
+    {
+        SchemaLoader.prepareServer();
+        CassandraDaemon daemon = new CassandraDaemon();
+        daemon.completeSetup(); //startup must be completed, otherwise FS error will kill JVM regardless of failure policy
+        StorageService.instance.registerDaemon(daemon);
+        StorageService.instance.initServer();
+    }
+
+    @AfterClass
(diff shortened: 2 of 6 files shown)


test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.junit.Test;
+
+import org.apache.cassandra.SchemaLoader;
+import org.apache.cassandra.config.Config;
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.exceptions.ConfigurationException;
+import org.apache.cassandra.gms.Gossiper;
+import org.apache.cassandra.io.FSErrorHandler;
+import org.apache.cassandra.io.FSReadError;
+import org.apache.cassandra.io.sstable.CorruptSSTableException;
```
Maxwell-Guo, line 40: remove ununsed import

test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
```diff
@@ -0,0 +1,121 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class DefaultFSErrorHandlerTest
+{
+    FSErrorHandler handler = new DefaultFSErrorHandler();
```
Maxwell-Guo: can they be private ?

test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
```diff
@@ -0,0 +1,121 @@
...
+    public void teardown()
+    {
+        DatabaseDescriptor.setDiskFailurePolicy(oldDiskPolicy);
+    }
+
+    @Test
+    public void testFSErrorPolicyIgnore()
+    {
+        assertEquals(DatabaseDescriptor.getDiskFailurePolicy(), Config.DiskFailurePolicy.ignore);
+        handler.handleFSError(new FSReadError(new IOException(), "blah"));
```
Maxwell-Guo: what does blah mean ?

test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
```diff
@@ -0,0 +1,121 @@
...
+        // error ignored, gossip not disabled
+        assertTrue(Gossiper.instance.isEnabled());
+    }
+
+    @Test
+    public void testFSErrorPolicyStop()
+    {
```
Maxwell-Guo, line 89: as all your test case's code logic are same with different DiskFailurePolicy 
I think it is better to use one test case with different Parameterized.Parameters [stop, die, ignore, stop_paranoid]

test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
```diff
@@ -0,0 +1,121 @@
...
+import org.apache.cassandra.gms.Gossiper;
+import org.apache.cassandra.io.FSErrorHandler;
+import org.apache.cassandra.io.FSReadError;
+import org.apache.cassandra.io.sstable.CorruptSSTableException;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+public class DefaultFSErrorHandlerTest
```
Maxwell-Guo, line 46: it is better to rename the class name to DefaultFSErrorHandlerForGossipTest 
if this test is only for gossip's status

test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
Runtian: just use the same exception from JVMStabilityInspectorTest.java

test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
Runtian: This handler should also stop rpc server as well as native transport if they are running.
But it is hard to start the mentioned services in unit tests.

CommitLogFailurePolicyTest.testCommitFailurePolicy_stop is also only verifying gossip.

test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
smiklosovic: Yes, I agree.

test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java
smiklosovic: Looking into `FSReadError` constructors, the second parameter is meant to represent "path" to a file which caused the exception.

src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java
```diff
@@ -60,6 +60,7 @@ public void handleFSError(FSError e)
 
         switch (DatabaseDescriptor.getDiskFailurePolicy())
         {
+            case die:
```
smiklosovic, line 64: the order of these enums in this switch is different from the order in `handleStartupFSError`

smiklosovic: Is not it true that when policy is `die` and we reach this method, instead of killing the jvm, you will only stop transports? There is this in `cassandra.yaml`:

````
# die
#   shut down gossip and client transports and kill the JVM for any fs errors or
#   single-sstable errors, so the node can be replaced.
````

As I read it, we should just kill whole JVM, no?

Comment (smiklosovic) [reply src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java:64]: Also, do not we miss `die` policy in `handleCorruptSSTable` method in this class? There is only `stop_paranoid`.

Comment (Maxwell-Guo) [reply src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java:64]: https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/utils/JVMStabilityInspector.java#L142

@smiklosovic I think at this step we will kill jvm when policy is die with FSerror or CorruptSSTableException

Comment (smiklosovic) [reply src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java:64]: ah right ...

Comment (smiklosovic) [reply src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java:64]: @Runtian can you actually explain what this change does? Because to my understanding it will just stop transports? But the objective is to kill the whole jvm? What are we stopping the transports for when we kill the jvm anyway?

It is here to not reach "default" so IllegalStateException is not thrown?

Comment (Runtian) [reply src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java:64]: Right, it is here to not reach the default so IllegalStateException will not be thrown.

We can also just add die with ignore to do nothing, but I think it is better to add die with stop and stop_paranoid so it tries to stop the transport and gossip etc. Also added the similar change to `handleCorruptSSTable`

File: test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java (hunk 5 above)
Comment (smiklosovic) [reply test/unit/org/apache/cassandra/service/DefaultFSErrorHandlerTest.java:46]: @Runtian what is hard about starting it? 

You know what just leave it as it is ... this is enough.

File: test/unit/org/apache/cassandra/service/DiskFailurePolicyTest.java [hunk 7]
```diff
@@ -0,0 +1,133 @@
...
+        StorageService.instance.registerDaemon(daemon);
+        killerForTests = new KillerForTests();
+        originalKiller = JVMStabilityInspector.replaceKiller(killerForTests);
+        originalDiskFailurePolicy = DatabaseDescriptor.getDiskFailurePolicy();
+        StorageService.instance.startGossiping();
+        Assert.assertTrue(Gossiper.instance.isEnabled());
+    }
+
+    @After
+    public void teardown() {
```
smiklosovic: braces should be on the new line, next test method should be fixed too.

test/unit/org/apache/cassandra/service/DiskFailurePolicyTest.java
```diff
@@ -0,0 +1,133 @@
...
+    public void testPolicies() {
+        DatabaseDescriptor.setDiskFailurePolicy(testPolicy);
+        JVMStabilityInspector.inspectThrowable(t);
+        Assert.assertEquals(expectJVMKilled, killerForTests.wasKilled());
+        Assert.assertEquals(expectJVMKilledQuiet, killerForTests.wasKilledQuietly());
+        if (!expectJVMKilled) {
+            // only verify gossip if JVM is not killed
+            Assert.assertEquals(expectGossipRunning, Gossiper.instance.isEnabled());
+        }
+
```
smiklosovic: nit: empty line

test/unit/org/apache/cassandra/utils/KillerForTests.java
```diff
@@ -29,6 +29,10 @@ public class KillerForTests extends JVMStabilityInspector.Killer
     @Override
     protected void killCurrentJVM(Throwable t, boolean quiet)
     {
+        if (killed) {
```
smiklosovic: braces (you could go without braces as well as there is just `return` here)

PR #2190, opened 2023-03-03
CASSANDRA-18299: Add support for prepared statements for accord transactions

dcapwell:


src/java/org/apache/cassandra/cql3/QueryProcessor.java
```diff
@@ -999,6 +999,16 @@ else if (statement instanceof BatchStatement)
                 }
                 return false;
             }
+            else if (statement instanceof TransactionStatement)
```
dcapwell: should batch/txn both share a interface to avoid this duplication of code?

src/java/org/apache/cassandra/cql3/ResultSet.java
```diff
@@ -316,6 +317,8 @@ public static ResultMetadata fromPrepared(CQLStatement statement)
         {
             if (statement instanceof SelectStatement)
                 return ((SelectStatement)statement).getResultMetadata();
+            if (statement instanceof TransactionStatement)
```
dcapwell: makes me wonder if this should be an interface to remove duplication

src/java/org/apache/cassandra/transport/messages/ExecuteMessage.java
```diff
@@ -137,8 +138,8 @@ protected Message.Response execute(QueryState state, long queryStartNanoTime, bo
 
             if (!prepared.fullyQualified
                 && !Objects.equals(state.getClientState().getRawKeyspace(), prepared.keyspace)
-                // We can not reliably detect inconsistencies for batches yet
-                && !(prepared.statement instanceof BatchStatement)
+                // We can not reliably detect inconsistencies for batches/txn yet
+                && !(prepared.statement instanceof BatchStatement || prepared.statement instanceof TransactionStatement)
```
dcapwell: makes me wonder if this should be an interface

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -1398,9 +1399,10 @@ protected void assertRowsNet(ProtocolVersion protocolVersion, ResultSet result,
             Object[] expected = rows[i];
             Row actual = iter.next();
 
-            Assert.assertEquals(String.format("Invalid number of (expected) values provided for row %d (using protocol version %s)",
-                                              i, protocolVersion),
-                                meta.size(), expected.length);
+            Assertions.assertThat(meta.size())
```
dcapwell, line 1402: functionally the same but was confused when I saw the error msg; it did not contain the row, and it was backwards so (expected = actual and actual = expected)

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -4397,6 +4397,11 @@ public static boolean getAccordTransactionsEnabled()
         return conf.accord_transactions_enabled;
     }
 
+    public static void setAccordTransactionsEnabled(boolean b)
+    {
+        conf.accord_transactions_enabled = b;
```
maedhroz, line 4402: nit: Does `accord_transactions_enabled` need to be volatile now? (I've seen comments on the mailing list around this not actually being required, but I'm not sure I understand. Is it that there are so many other locks held and released around touching these values that they're bound to be visible within a short time anyway?)

dcapwell: > nit: Does accord_transactions_enabled need to be volatile now?

since this is tests only, and that runs in `main` thread, I don't believe we need to.  This gets called before C* starts, so before any thread spins up

src/java/org/apache/cassandra/cql3/QueryProcessor.java
maedhroz: I think something like a `ComplexStatement` interface would be very nice. @jacek-lewandowski has actually hinted at wanting something like it for other reasons...

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
maedhroz: Ah didn't realize it was test-only

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -226,7 +226,7 @@ public TxnData coordinate(Txn txn, ConsistencyLevel consistencyLevel)
 
     private static RuntimeException throwTimeout(Txn txn, ConsistencyLevel consistencyLevel)
     {
-        throw txn.isWrite() ? new WriteTimeoutException(WriteType.TRANSACTION, consistencyLevel, 0, 0)
+        throw txn.isWrite() ? new WriteTimeoutException(WriteType.CAS, consistencyLevel, 0, 0)
```
ifesdjeen, line 229: I am not very familiar with Accord code, but I am assuming that you use CAS here because here is no support for transaction in the client yet, or? If so, maybe its worth to add a comment.

src/java/org/apache/cassandra/cql3/ResultSet.java
maedhroz: Yup, `ComplexStatement` to the rescue.

test/unit/org/apache/cassandra/Util.java
```diff
@@ -783,6 +784,13 @@ public static UnfilteredPartitionIterator executeLocally(PartitionRangeReadComma
         return command.queryStorage(cfs, controller);
     }
 
+    public static UnfilteredPartitionIterator executeLocally(SinglePartitionReadCommand command,
```
maedhroz, line 787: nit: Can we just make `ReadCommand#queryStorage()` public and merge these two methods?

test/unit/org/apache/cassandra/cql3/PreparedStatementsTest.java
```diff
@@ -581,4 +653,126 @@ private void testPrepareWithBatchLWT(ProtocolVersion version) throws Throwable
...
+        sb.append("APPLY BATCH");
+        return sb.toString();
+    }
+
+    private static List<String> columnNames(ResultSet rs)
+    {
+        return rs.getColumnDefinitions().asList().stream().map(d -> d.getName()).collect(Collectors.toList());
+    }
+
+    private static void updateTxnState()
```
maedhroz, line 773: nit: I guess we should throw in a TODO to remove this once transactional metadata is in place? (The cache disabling bit could just be in a `@Before` method...I think.)

test/unit/org/apache/cassandra/cql3/PreparedStatementsTest.java
```diff
@@ -581,4 +653,126 @@ private void testPrepareWithBatchLWT(ProtocolVersion version) throws Throwable
...
+    public void testPrepareWithAccordV5()
+    {
+        Assume.assumeTrue("Protocol v5 is CURRENT", ProtocolVersion.CURRENT != ProtocolVersion.V5);
+        testPrepareWithAccord(ProtocolVersion.V5);
+    }
+
+    @Test
+    public void testPrepareWithAccordCurrent()
+    {
+        testPrepareWithAccord(ProtocolVersion.CURRENT);
```
maedhroz, line 673: nit: Maybe just parameterize on `SUPPORTED_VERSIONS`?

maedhroz (approved):
Mostly LGTM, with some minor nits. Biggest thing is the `ComplexStatement` interface/base class idea, which seems worth doing.

jacek-lewandowski:
hey, yes, I have this PR https://github.com/apache/cassandra/pull/2195

src/java/org/apache/cassandra/cql3/QueryProcessor.java
dcapwell: What does `Complex` mean though?  I use `Group` in this ticket as it "groups" other statements together... having a type that works with multiple statements makes sense, so wonder what other goals/reasons exist.

src/java/org/apache/cassandra/service/accord/AccordService.java
dcapwell: > no support for transaction in the client yet,

correct, java driver takes our `String` and converts to an `Enum`, which then fails; java driver does not handle unknown types...  

> If so, maybe its worth to add a comment.

makes sense, would have avoided this if a comment existed

test/unit/org/apache/cassandra/Util.java
dcapwell: atm we only have 2 types, but wonder if we need to worry about adding methods to `ReadCommand` as other indexes would be aware?  I "feel" like adding is safe...

test/unit/org/apache/cassandra/cql3/PreparedStatementsTest.java
dcapwell: 100% agree but didn't want to rewrite all tests.

dcapwell: JUnit 5 allows different tests to have different parameters, but not in JUnit 4 as its done via the constructor.

dcapwell:
@jacek-lewandowski , @maedhroz took a look at the linked PR and `ComplexQualifiedStatement` just holds a `String`, so complex in that patch means "I take a string" but complex in this patch means "I am a bag of Statements"

src/java/org/apache/cassandra/cql3/QueryProcessor.java
dcapwell: testing out

```
public interface GroupCQLStatement extends CQLStatement
{
    Iterable<? extends CQLStatement> getStatements();
}
```

maedhroz: If we want to really find the best possible name, it might actually be `CompositeStatement` (in the sense that batches and txns are statements made up of other elements/statements).

src/java/org/apache/cassandra/cql3/ResultSet.java
dcapwell: testing out

```
public interface ReturningCQLStatement extends CQLStatement
{
    ResultSet.ResultMetadata getResultMetadata();
}
```

src/java/org/apache/cassandra/service/accord/AccordService.java
dcapwell: added

```
/**
 * Identifier for what type of operation timed out.  This type is driver facing as a String, but some drivers conver
 * this to an enum, meaning any changes to this type require protocol changes and drivers supporting
 */
```

jacek-lewandowski:
Yes @dcapwell, because I only needed that. Though it is related to the same types of statements. I think I'll wait for your PR before proceeding with my one, and add what I need to your complex statement.

src/java/org/apache/cassandra/cql3/QueryProcessor.java
```diff
@@ -989,10 +989,10 @@ else if (statement instanceof SelectStatement)
                 statementKsName = selectStatement.keyspace();
                 statementCfName = selectStatement.table();
             }
-            else if (statement instanceof BatchStatement)
+            else if (statement instanceof CompositeCQLStatement)
             {
-                BatchStatement batchStatement = ((BatchStatement) statement);
-                for (ModificationStatement stmt : batchStatement.getStatements())
+                CompositeCQLStatement transactionStatement = (CompositeCQLStatement) statement;
+                for (CQLStatement stmt : transactionStatement.getStatements())
```
jacek-lewandowski: nit: `transactionStatement` -> `batchStatement` ?

src/java/org/apache/cassandra/cql3/statements/QualifiedStatement.java
```diff
@@ -78,4 +78,50 @@ public String toString()
     {
         return ToStringBuilder.reflectionToString(this, ToStringStyle.SHORT_PREFIX_STYLE);
     }
+
+    public static abstract class CompositeQualifiedStatement extends QualifiedStatement
```
jacek-lewandowski: Shouldn't it implement `CompositeCQLStatement` ?

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
```diff
@@ -168,6 +168,33 @@ public void validate(ClientState state)
...
+                stream = Stream.concat(stream, Stream.of(returningSelect.select));
+            stream = Stream.concat(stream, updates.stream());
+            return stream.iterator();
+        };
+    }
+
+    @Override
+    public ResultSet.ResultMetadata getResultMetadata()
+    {
+        if (returningSelect != null)
```
jacek-lewandowski: Result metadata seems to be something final and perhaps it should be created in the constructor, wdyt?

src/java/org/apache/cassandra/service/accord/AccordService.java
jacek-lewandowski: Since one driver will support that value and the other will not - wouldn't it be worth making it somehow dynamically adjust to the driver version?

src/java/org/apache/cassandra/cql3/statements/QualifiedStatement.java
maedhroz: nit: Given most usage looks like `QualifiedStatement.CompositeQualifiedStatement`, I'd consider just naming this `Composite`, so it's more like `QualifiedStatement.Composite`...but not super important. Fine w/ whatever you go with.

src/java/org/apache/cassandra/cql3/statements/CompositeCQLStatement.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.cql3.statements;
+
+import org.apache.cassandra.cql3.CQLStatement;
+
+public interface CompositeCQLStatement extends CQLStatement
```
maedhroz: nit: What do you think about moving this and `ReturningCQLStatement` into `CQLStatement` itself, following the example of `SingleKeyspaceCqlStatement`? (I guess you could also name them `Composite` and `Returning` respectively, if you wanted to reference them like `CQLStatement.Composite`, but that definitely would make it inconsistent w/ `SingleKeyspaceCqlStatement`.)

src/java/org/apache/cassandra/db/WriteType.java
```diff
@@ -17,6 +17,10 @@
  */
 package org.apache.cassandra.db;
 
+/**
+ * Identifier for what type of operation timed out.  This type is driver facing as a String, but some drivers conver
```
maedhroz:
```suggestion
 * Identifier for what type of operation timed out.  This type is driver facing as a String, but some drivers convert
```

src/java/org/apache/cassandra/db/WriteType.java
```diff
@@ -17,6 +17,10 @@
...
+ * this to an enum, meaning any changes to this type require protocol changes and drivers supporting
```
maedhroz:
```suggestion
 * this to an enum, meaning any changes to this type require protocol changes and driver support.
```

src/java/org/apache/cassandra/transport/messages/ExecuteMessage.java
```diff
@@ -137,8 +137,8 @@ protected Message.Response execute(QueryState state, long queryStartNanoTime, bo
 
             if (!prepared.fullyQualified
                 && !Objects.equals(state.getClientState().getRawKeyspace(), prepared.keyspace)
-                // We can not reliably detect inconsistencies for batches yet
-                && !(prepared.statement instanceof BatchStatement)
+                // We can not reliably detect inconsistencies for groups yet
```
maedhroz:
```suggestion
                // We cannot reliably detect inconsistencies for composite statements yet
```

src/java/org/apache/cassandra/cql3/QueryProcessor.java
dcapwell: removed the variable, so no worries on names anymore =D

src/java/org/apache/cassandra/cql3/statements/QualifiedStatement.java
dcapwell: `QualifiedStatement extends CQLStatement.Raw`
`CompositeCQLStatement extends CQLStatement`

so, can't...

dcapwell: switched to `Composite`, debated this... if we pull out from this class should rename I guess

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
dcapwell: true we could, but I don't think we need to as the access is short lived here.  prepared statements holds ref in `org.apache.cassandra.cql3.QueryHandler.Prepared#Prepared` and txn statement calls this after we get the response back from the accord coordinator, so by keeping it something we "compute" we keep this object in short term memory rather than make it used 2 times in long term memory

src/java/org/apache/cassandra/cql3/statements/CompositeCQLStatement.java
dcapwell: moved, but kept the `CQLStatement` postfix as it matches `SingleKeyspaceCqlStatement`.

test/unit/org/apache/cassandra/Util.java
dcapwell: thought about this more and I rather not change virtual tables in this patch.  I asked this question in slack (see https://the-asf.slack.com/archives/CK23JSY2K/p1678145643633279) and it may make sense to make public and depend on it, but would need to flesh out virtual tables implementation as they don't use that method and will have failures if touched.

src/java/org/apache/cassandra/service/accord/AccordService.java
dcapwell: > wouldn't it be worth making it somehow dynamically adjust to the driver version?

I would hope we do this similar to how we downgrade to v4 or use smaller versions in internode... I would hope I provide `TRANSACTION` and there is some cql3 driver sending logic that knows that it needs to "downgrade" and hide that value...

Given that we wish to not change drivers for accord support, thought best to leave this to another ticket.

test/unit/org/apache/cassandra/cql3/PreparedStatementsTest.java
dcapwell: adding `//TODO Remove this method once CEP-21 and CEP-15 integrate`

> (The cache disabling bit could just be in a @Before method...I think.)

the current pattern is to do this w/e we do a schema change, so figured best to be consistent...

dcapwell:
@jacek-lewandowski @ifesdjeen  let me know what you think about the update; believe I addressed all feedback.

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
dcapwell: to clarify, in the prepared case we would compute this once and reuse in the `execute` response, but when not prepared you compute this on-demand and this is short lived...

thinking about this more maybe we should, so in the prepared case we only call `org.apache.cassandra.cql3.ResultSet#computeResultMetadataId` once and the non-prepared case we also only call once...  right now prepared case will call this the same amount as non-prepared and an additional one while prepare...  given that `org.apache.cassandra.cql3.ResultSet#computeResultMetadataId` does have a noticeable impact on perf, doing once would help the prepared case and be equal to the non-prepared case... ill make this change

dcapwell: made this change, we now compute this once per object, so prepared statements should not be hashing the names

jacek-lewandowski (approved):
+1

PR #2195, opened 2023-03-06
[CASSANDRA-18302] Statement source descriptions

jacek-lewandowski:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/cql3/StatementSource.java
+++ b/src/java/org/apache/cassandra/cql3/StatementSource.java
@@ -1,0 +1,76 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.cql3;
+
+import java.util.Objects;
+
+import org.antlr.runtime.Token;
+
+import static java.lang.Math.max;
+import static java.lang.Math.min;
+
+public class StatementSource
+{
+    public static final StatementSource INTERNAL = new StatementSource(0, 0);
+
+    public final int line;
+    public final int charPositionInLine;
+
+    public StatementSource(int line, int charPositionInLine)
+    {
+        this.line = line;
+        this.charPositionInLine = charPositionInLine;
+    }
+
+    @Override
+    public String toString()
+    {
+        if (this == INTERNAL)
+        {
+            return "<<<internal statement>>>";
+        }
+        else
+        {
+            if (!isEmpty())
+                return String.format("at [%d:%d]", line + 1, charPositionInLine + 1);
+            else
+                return "";
+        }
+    }
+
+    public boolean isEmpty()
+    {
+        return line > Character.MAX_VALUE || line == Character.MAX_VALUE && charPositionInLine > Character.MAX_VALUE;
+    }
+
+    // note - this can also reproduce the original statement raw text by getting TokenStream and calling toString(startToken, endToken)
+    public static StatementSource create(Token startToken)
+    {
+        Objects.requireNonNull(startToken);
+
+        if (startToken.getType() == Token.EOF)
+            return new StatementSource(Character.MAX_VALUE + 1, 0);
+
+        int startLine = min(max(startToken.getLine(), 1) - 1, Character.MAX_VALUE);
+        int startChar = min(max(startToken.getCharPositionInLine(), 0), Character.MAX_VALUE);
+
+        return new StatementSource(startLine, startChar);
+    }
+
+}
--- a/src/java/org/apache/cassandra/cql3/statements/BatchStatement.java
+++ b/src/java/org/apache/cassandra/cql3/statements/BatchStatement.java
@@ -18,7 +18,16 @@
 package org.apache.cassandra.cql3.statements;
 
 import java.nio.ByteBuffer;
-import java.util.*;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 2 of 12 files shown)


src/java/org/apache/cassandra/cql3/StatementSource.java
```diff
@@ -0,0 +1,162 @@
...
+            this.startLine = startLine;
+            this.startChar = startChar;
+            this.endLine = endLine;
+            this.endChar = endChar;
+        }
+
+        private boolean isEmpty()
+        {
+            return startLine == Character.MAX_VALUE ||
+                   startLine > endLine ||
```
dcapwell: what is the case that this happens?  this feels like a bug, should we not fail?

src/java/org/apache/cassandra/cql3/StatementSource.java
```diff
@@ -0,0 +1,162 @@
...
+                   startLine == endLine && (startChar == Character.MAX_VALUE || startChar > endChar);
```
dcapwell: `startChar > endChar` feels like a bug

src/java/org/apache/cassandra/cql3/StatementSource.java
```diff
@@ -0,0 +1,162 @@
...
+    public static Bounds createBounds(Token startToken, Token endToken)
+    {
+        Objects.requireNonNull(startToken);
+        Objects.requireNonNull(endToken);
+
+        int startLine, startChar, endLine, endChar;
+
+        if (startToken.getType() == Token.EOF)
+        {
+            startLine = Character.MAX_VALUE;
```
dcapwell: if start token is EOF do we not know that endToken is as well?

src/java/org/apache/cassandra/cql3/statements/DeleteStatement.java
```diff
@@ -174,7 +186,8 @@ protected ModificationStatement prepareInternal(ClientState state,
                                                        operations,
                                                        restrictions,
                                                        conditions,
-                                                       attrs);
+                                                       attrs,
+                                                       bounds.describeSource(enclosingStatement));
```
dcapwell: this feels backwards, we create the underline statements before we create the composite statement, so how do we pass it in?

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
```diff
@@ -207,12 +204,9 @@ public ResultSet.ResultMetadata getResultMetadata()
     TxnNamedRead createNamedRead(NamedSelect namedSelect, QueryOptions options, ClientState state)
     {
         SelectStatement select = namedSelect.select;
-        ReadQuery readQuery = select.getQuery(options, 0);
-        checkTrue(readQuery instanceof  SinglePartitionReadQuery.Group, ILLEGAL_RANGE_QUERY_MESSAGE, select.asCQL(options, state));
-
         // We reject reads from both LET and SELECT that do not specify a single row.
         @SuppressWarnings("unchecked")
-        SinglePartitionReadQuery.Group<SinglePartitionReadCommand> selectQuery = (SinglePartitionReadQuery.Group<SinglePartitionReadCommand>) readQuery;
+        SinglePartitionReadQuery.Group<SinglePartitionReadCommand> selectQuery = (SinglePartitionReadQuery.Group<SinglePartitionReadCommand>) select.getQuery(options, 0);
```
dcapwell, line 209: why prefer ClassCast rather than `ILLEGAL_RANGE_QUERY_MESSAGE`?

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
```diff
@@ -223,12 +217,9 @@ TxnNamedRead createNamedRead(NamedSelect namedSelect, QueryOptions options, Clie
     List<TxnNamedRead> createNamedReads(NamedSelect namedSelect, QueryOptions options, ClientState state)
     {
         SelectStatement select = namedSelect.select;
-        ReadQuery readQuery = select.getQuery(options, 0);
-        checkTrue(readQuery instanceof  SinglePartitionReadQuery.Group, ILLEGAL_RANGE_QUERY_MESSAGE, select.asCQL(options, state));
-
         // We reject reads from both LET and SELECT that do not specify a single row.
         @SuppressWarnings("unchecked")
-        SinglePartitionReadQuery.Group<SinglePartitionReadCommand> selectQuery = (SinglePartitionReadQuery.Group<SinglePartitionReadCommand>) readQuery;
+        SinglePartitionReadQuery.Group<SinglePartitionReadCommand> selectQuery = (SinglePartitionReadQuery.Group<SinglePartitionReadCommand>) select.getQuery(options, 0);
```
dcapwell, line 222: why prefer class cast over `ILLEGAL_RANGE_QUERY_MESSAGE`?

dcapwell (commented):
got to this late and there is parts that I feel I need to poke with an IDE to properly review, so I sadly have to delay real review to tomorrow...

src/java/org/apache/cassandra/cql3/StatementSource.java
jacek-lewandowski: In all those pointed out cases I tried to make this class super resistant to wrong input regardless of the circumstances. But it can fail as well.

src/java/org/apache/cassandra/cql3/StatementSource.java
jacek-lewandowski: Yes, we can assume that

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
jacek-lewandowski: In this and the other case, the statement at this point is guaranteed to not be partition range query. It has been verified during preparation.

src/java/org/apache/cassandra/cql3/statements/DeleteStatement.java
jacek-lewandowski: No, notice that the enclosing composite statement is a parsed statement and it already exists so as the parsed delete statement.

jacek-lewandowski:
thanks @dcapwell, I've applied some fixes basing on your feedback

src/java/org/apache/cassandra/cql3/statements/QualifiedStatement.java
```diff
@@ -81,13 +81,27 @@ public String toString()
...
         Composite()
         {
             super(null);
         }
 
         protected abstract Iterable<? extends QualifiedStatement> getStatements();
 
+        public void setRawText(String txt)
+        {
+            this.txt = txt != null ? txt.substring(0, Math.min(txt.length(), RAW_TEXT_MAX_LENGTH)) : null;
```
maedhroz: nit: if we actually do exceed the max length, it might be a nice UX tweak to include a "blah blah..." (ellipsis) at the end just to make it clear there's more content (but we're truncating).

src/java/org/apache/cassandra/cql3/statements/BatchStatement.java
```diff
@@ -235,7 +261,7 @@ public void validate() throws InvalidRequestException
             for (ModificationStatement stmt : statements)
             {
                 if (ksName != null && (!stmt.keyspace().equals(ksName) || !stmt.table().equals(cfName)))
-                    throw new InvalidRequestException("Batch with conditions cannot span multiple tables");
+                    throw new InvalidRequestException("Batch with conditions cannot span multiple tables: " + stmt.source.describe(true));
```
maedhroz: Is there any coverage for these two cases in `BatchTest`? In general, feels like it would be good to have a test for the `ModificationStatement` case...

src/java/org/apache/cassandra/cql3/statements/QualifiedStatement.java
```diff
@@ -81,13 +81,27 @@ public String toString()
 
     public static abstract class Composite extends QualifiedStatement
     {
+        private static final int RAW_TEXT_MAX_LENGTH = 4096;
```
maedhroz: If we set this to something lower (let's say 38), and then run `shouldRejectIllegalLimitInLet()`, you encounter a situation where the `Bounds` points to non-existent raw data. I suppose this won't be problematic unless we have very large statements.

In `CQLFragmentParser#parseAnyUnhandled()`, I think we already have the `Bounds` objects we would later use to filter the raw text, right? What if, instead of filtering later, and storing the raw text for the entire enclosing statement, we filter each statement in `CQLFragmentParser#parseAnyUnhandled()` avoiding the need for a limit in the first place? The individual statement would just get the simple bounds and the text from the raw statement text within those bounds...and we'd be able to present the error the same way (I think).

WDYT?

src/java/org/apache/cassandra/cql3/StatementSource.java
```diff
@@ -0,0 +1,162 @@
...
+    public String describe(boolean mayIncludeSourceText)
+    {
+        if (this == INTERNAL)
+        {
+            return "<<<internal statement>>>";
+        }
+        else
+        {
+            if (!bounds.isEmpty() && text != null && mayIncludeSourceText)
+                return String.format("at %d:%d - %s", bounds.startLine + 1, bounds.startChar + 1, text);
```
maedhroz: nit: Another way to do it that could avoid the dash...

`at [2:1] SELECT....`

Matter of style though, so feel free to ignore.

src/java/org/apache/cassandra/cql3/StatementSource.java
```diff
@@ -0,0 +1,87 @@
...
+        Objects.requireNonNull(endToken);
+        Objects.requireNonNull(input);
+
+        if (startToken.getType() == Token.EOF)
+            return new StatementSource(Character.MAX_VALUE + 1, 0, null);
+
+        int startLine = min(max(startToken.getLine(), 1) - 1, Character.MAX_VALUE);
+        int startChar = min(max(startToken.getCharPositionInLine(), 0), Character.MAX_VALUE);
+        String txt = input.toString(startToken, endToken);
+        txt = txt.replace('\n', ' ');
```
dcapwell: we should keep the newline,

```
SELECT a,
               b,
               c -- should fail at line 3 due to 3 not being in the schema
```

src/java/org/apache/cassandra/cql3/StatementSource.java
```diff
@@ -0,0 +1,87 @@
...
+        txt = TWO_OR_MORE_WHITESPACES.matcher(txt).replaceAll(" ").trim();
```
dcapwell: we should leave the input alone, if we want to show the string the users provided, we should show as-is

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
```diff
@@ -366,21 +357,19 @@ public ResultMessage execute(QueryState state, QueryOptions options, long queryS
...
         try
         {
+            // check again since now we have query options; note that statements are quaranted to be single partition reads at this point
             for (NamedSelect assignment : assignments)
-                checkAtMostOneRowSpecified(state.getClientState(), options, assignment.select, INCOMPLETE_PRIMARY_KEY_LET_MESSAGE);
+                checkFalse(isSelectingMultipleClusterings(assignment.select, options), INCOMPLETE_PRIMARY_KEY_SELECT_MESSAGE, "LET assignment", assignment.select.source);
 
             if (returningSelect != null)
-                checkAtMostOneRowSpecified(state.getClientState(), options, returningSelect.select, INCOMPLETE_PRIMARY_KEY_SELECT_MESSAGE);
+                checkFalse(isSelectingMultipleClusterings(returningSelect.select, options), INCOMPLETE_PRIMARY_KEY_SELECT_MESSAGE, "returning SELECT", returningSelect.select.source);
```
maedhroz, line 365: I'd even be fine if we just did what we do for INSERT/UPDATE...that is, just say `SELECT at [2:1]` or `LET at [1:1]`. Not sure if someone reading this would glean information from "returning" or it's just an implementation detail.

CC @dcapwell

test/unit/org/apache/cassandra/cql3/StatementSourceTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.antlr.runtime.Token;
+import org.mockito.Mockito;
+
+import static org.apache.cassandra.cql3.StatementSource.create;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
+
+public class StatementSourceTest
+{
+    private final static Logger logger = LoggerFactory.getLogger(StatementSourceTest.class);
```
maedhroz: logger unused now

maedhroz (approved):
LGTM

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
dcapwell: "returning select" I think helps as we can only have 1 of those

maedhroz: The only other SELECTs would be called LET here...anyway...I'm good with whatever the two of you like better.

PR #2196, opened 2023-03-06
[CASSANDRA-18304-4.1] Do not submit hints when hinted_handoff_enabled=false

pauloricardomg:


src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -2641,6 +2641,7 @@ public static Future<Void> submitHint(Mutation mutation,
                                           EndpointsForToken targets,
                                           AbstractWriteResponseHandler<IMutation> responseHandler)
     {
+        targets = targets.filter(StorageProxy::shouldHint);
```
Claudenw: If `targets` is empty then perhaps an `ImmediateFuture` future should be returned rather than continuing with the construction of a regular future that does nothing.

Claudenw (changes requested):

pmcfadin:
Closing as [CASSANDRA-18304](https://issues.apache.org/jira/browse/CASSANDRA-18304) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #2208, opened 2023-03-09
CASSANDRA-18317 Properly synchronize CQLSSTableWriter#build on the Schema.instance object

frankgh:
In this commit the `org.apache.cassandra.io.sstable.CQLSSTableWriter#build` method synchronizes on the
`Schema.instance` object (instead of the `CQLSSTableWriter.class`) to prevent concurrent schema operations
to fail when the offline tools also updates the schema.
    
For example, a table creation operation, which modifies the keyspace tables metadata, might end up
missing the update when a concurrent call to the `CQLSSTableWriter#build` method is accessing the
singleton Schema instance.

Cassandra 4.0 changes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,8078 +14,11 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-#
 
-version: 2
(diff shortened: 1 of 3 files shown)


frankgh:
Closed via https://github.com/apache/cassandra/commit/26c374da4f03e4a6b64e414805cd92f3eb0a36c6

PR #2215, opened 2023-03-14
[CASSANDRA-18153-3.11] Save host id to system.local and flush immediately after startup 

jacek-lewandowski:


src/java/org/apache/cassandra/db/SystemKeyspace.java
```diff
@@ -1150,6 +1150,7 @@ public static synchronized UUID setLocalHostId(UUID hostId)
     {
         String req = "INSERT INTO system.%s (key, host_id) VALUES ('%s', ?)";
         executeInternal(String.format(req, LOCAL, LOCAL), hostId);
+        forceBlockingFlush(LOCAL);
```
smiklosovic, line 1207: this is quite a good catch, actually.

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -112,6 +114,11 @@ CommitLog start()
         return this;
     }
 
+    public boolean hasFilesToReplay()
+    {
+        return new File(segmentManager.storageDirectory).listFiles(unmanagedFilesFilter).length > 0;
```
smiklosovic: IDEA complains that this:

new File(segmentManager.storageDirectory).listFiles(unmanagedFilesFilter)

can produce NPE.

You should rewrite this to use Cassandra's "File", not Java's one. 

I know this is branch 3.11 where that stuff is not in yet but IDE complaining about NPE should be fixed anyway somehow I guess.

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -177,7 +182,7 @@ public void recoverPath(String path) throws IOException
 
     private static UUID getLocalHostId()
     {
-        return Optional.ofNullable(StorageService.instance.getLocalHostUUID()).orElseGet(SystemKeyspace::getLocalHostId);
+        return StorageService.instance.getLocalHostUUID();
```
smiklosovic, line 202: if I understand that correctly, this will never be null anymore because we are doing that initialization in `CassandraDaemon.`

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -236,6 +236,13 @@ protected void setup()
         // This should be the first write to SystemKeyspace (CASSANDRA-11742)
         SystemKeyspace.persistLocalMetadata();
 
+        // We should store host ID as soon as possible in the system.local table and flush that table to disk so that
+        // we can be sure that those changes are stored in sstable and not in the commit log (see CASSANDRA-18153).
```
smiklosovic: sstable -> SSTable

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1980,12 +1980,14 @@ public Map<String, String> getTokenToEndpointMap()
...
     {
-        return getTokenMetadata().getHostId(FBUtilities.getBroadcastAddress()).toString();
+        UUID id = getLocalHostUUID();
+        return id != null ? id.toString() : null;
     }
 
     public UUID getLocalHostUUID()
     {
-        return getTokenMetadata().getHostId(FBUtilities.getBroadcastAddress());
+        UUID id = getTokenMetadata().getHostId(FBUtilities.getBroadcastAddress());
```
smiklosovic, line 2081: what is the probability that id returned by TokenMetadata and id returned by looking at disk would be different?

PR #2220, opened 2023-03-16
CASSANDRA-18337: Operations.migrateReadRequiredOperations fails due to concurrent access when TransactionStatement is prepared

dcapwell:


src/java/org/apache/cassandra/cql3/Operations.java
```diff
@@ -42,29 +44,46 @@ public final class Operations implements Iterable<Operation>
...
+        staticSubstitutions = new ArrayList<>(other.staticSubstitutions);
     }
 
-    public void migrateReadRequiredOperations()
+    @Nullable
+    public Operations migrateReadRequiredOperations()
     {
-        migrateReadRequiredOperations(staticOperations, staticSubstitutions);
-        migrateReadRequiredOperations(regularOperations, regularSubstitutions);
+        Operations other = new Operations(this);
```
dcapwell: one way to lower memory in the case of no migration is to do 2 passes, first to find if any need to migrate, and the second to migrate them.  This would lower the memory costs in the cases where `requiresRead` isn't found and only adds extra cycles to the cases where it is found....  It was hard for me to say which way to go as I don't really know how common both sides are, so not sure which to bias towards...  so I went with the less-work route and did a tiny change to existing `migrateReadRequiredOperations` at the cost of extra memory.

maedhroz: I guess you could do something like this...

```
@Nullable
public Operations migrateReadRequiredOperations()
{
    Operations other = null;

    if (regularOperations.stream().anyMatch(Operation::requiresRead))
    {
        other = new Operations(this);
        migrateReadRequiredOperations(other.regularOperations, other.regularSubstitutions);
        migrateReadRequiredOperations(other.staticOperations, other.staticSubstitutions);
    }
    else if (staticOperations.stream().anyMatch(Operation::requiresRead))
    {
        other = new Operations(this);
        migrateReadRequiredOperations(other.staticOperations, other.staticSubstitutions);
    }
    
    return other;
}
```
This is cheaper in the case where we don't need any migrations, and operations on static columns are probably the exceptional case...just not sure what the probabilities are, as you say.

maedhroz: One more idea...

We create `Operations` and add items to it in the `ModificationStatement.Parsed#prepareInternal()` implementations. Can we just add the `Operation` instances requiring reads to the correct list there and avoid later migrations altogether?

ex. In `ParsedUpdate#prepareInternal()`, we have `isForTxn` telling us whether or not we're processing a transaction. We could add operations that need a read via `Operations#add(ColumnMetadata, ReferenceOperation)` there.

maedhroz (approved):

src/java/org/apache/cassandra/cql3/Operations.java
dcapwell: had so many back and forth on this due to CAS.... I simplified it all by having `Operations` take a `boolean isForTxn` and it properly places operations where they belong, and added new constructor that does the migration in a single path

src/java/org/apache/cassandra/cql3/statements/UpdateStatement.java
```diff
@@ -237,12 +246,14 @@ public static class ParsedInsertJson extends ModificationStatement.Parsed
...
         private final boolean defaultUnset;
+        private final boolean isForTxn;
 
-        public ParsedInsertJson(QualifiedName name, Attributes.Raw attrs, Json.Raw jsonValue, boolean defaultUnset, boolean ifNotExists, StatementSource source)
+        public ParsedInsertJson(QualifiedName name, Attributes.Raw attrs, Json.Raw jsonValue, boolean defaultUnset, boolean ifNotExists, StatementSource source, boolean isForTxn)
         {
             super(name, StatementType.INSERT, attrs, null, ifNotExists, false, source);
             this.jsonValue = jsonValue;
             this.defaultUnset = defaultUnset;
+            this.isForTxn = isForTxn;
```
aweisberg, line 256: I think isForTxn is doesn't quite communicate that this is really for Accord so even CAS is included not just `TransactionStatement`. In some contexts we use lightweight transaction to describe CAS, but we don't always use this version for the modification statement for CAS just when it is on Accord.

aweisberg (changes requested):
This is great! Only suggestion is `isForTxn` could be probably be named better.

src/java/org/apache/cassandra/cql3/statements/UpdateStatement.java
dcapwell: > so even CAS is included
CAS isn't included, it's `false` there.  When CAS uses accord we "migrate" and create a Accord transaction version of these statements

aweisberg (approved):
Based on our discussion +1, seems like there isn't a great way to disambiguate with LWT since the generic term "Transaction" is what we use everywhere else to refer to the new bits.

PR #2199, opened 2023-03-07
Cassandra-12937 sstable compression

Claudenw:
Added code to set sstable compression option in the `conf.yaml` file.

Original static `CompressionParms.DEFAULT` configuration moved to `CompressionParams.defaultCompression()`
Added `DatabaseDescriptor.getSstableCompression()` and called it in place of the original `CompressionParams.DEFAULT`
Added test to verify that `CompressionParams` is properly constructed from the `Config.sstable_compressor`

Fixes: CASSANDRA-12937

Associated pull: https://github.com/apache/cassandra/pull/2254

PR #2222, opened 2023-03-16
CASSANDRA-12937 rebased against trunk

smiklosovic:
 
```

```


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -94,6 +94,7 @@
 import org.apache.cassandra.locator.InetAddressAndPort;
 import org.apache.cassandra.locator.Replica;
 import org.apache.cassandra.locator.SeedProvider;
+import org.apache.cassandra.schema.CompressionParams;
```
michaelsembwever, line 97: can we please not import schema class into this package.

this was already raised and addressed here: https://github.com/apache/cassandra/pull/2199#discussion_r1139429951

conf/cassandra.yaml
```diff
@@ -566,6 +566,14 @@ commitlog_segment_size: 32MiB
 #
 # flush_compression: fast
 
+# Defines the compression params used to read/write the sstables.  If undefined the default is LZ4 with 16kb chunk
+# length and min_compress_ratio of 0.0 (see CompressionParams.DEFAULT_MIN_COMPRESS_RATIO).
+# This is the default options for sstable compression, the (schema) table-level settings will override this option.
```
michaelsembwever, line 571:
```suggestion
# To disable compression set `type: none` or `type: none`NoopCompressor`
# This is the default options for sstable compression, the (schema) table-level settings will override this option.
```

conf/cassandra.yaml
```diff
@@ -566,6 +566,14 @@ commitlog_segment_size: 32MiB
...
+#sstable_compressor:
+#  class: "org.apache.cassandra.io.compress.LZ4Compressor"
+#  chunk_length_in_kb: "16"
+#  min_compress_ratio: "0"
```
michaelsembwever, line 575: can we please take these back to `parameters` like @Claudenw originally had them.

conf/cassandra.yaml
```diff
@@ -566,6 +566,14 @@ commitlog_segment_size: 32MiB
 #
 # flush_compression: fast
 
+# Defines the compression params used to read/write the sstables.  If undefined the default is LZ4 with 16kb chunk
+# length and min_compress_ratio of 0.0 (see CompressionParams.DEFAULT_MIN_COMPRESS_RATIO).
+# This is the default options for sstable compression, the (schema) table-level settings will override this option.
+#sstable_compressor:
+#  class: "org.apache.cassandra.io.compress.LZ4Compressor"
```
michaelsembwever, line 573:
```suggestion
#  class_name: "LZ4Compressor"
```

here, and in table options, it would be nice to have some aliases to known compressors: none, lz4, snappy, deflate, zstd.

src/java/org/apache/cassandra/config/ConfigurationMap.java
michaelsembwever, line 1: what's the purpose of this class?

conf/cassandra.yaml
smiklosovic: I disagree. We are using here the very same approach where we can plug-in things. It is well-known pattern that these things are referenced by FQCN (or just simple class name and package will be resolved automatically).

conf/cassandra.yaml
smiklosovic: so you want that to be same as for e.g.seed_provider? `class_name` + parameters so it translates to `ParametrizedClass`? That might work ... we would get rid of some helper classes.

src/java/org/apache/cassandra/config/ConfigurationMap.java
smiklosovic: Simple wrapper around hash map to have it dandy. We can replace it for `ParametrizedClass` as mentioned above.

conf/cassandra.yaml
michaelsembwever: what's the harm is providing aliases to the simple class names ? 

I see benefit and simplicity in permitting the user to define any of the following (to the same effect)
```
class_name: "lz4"
class_name: "LZ4Compressor"
class_name: "org.apache.cassandra.io.compress.LZ4Compressor"
```
(note, `class_name` is the standard in the yaml)

smiklosovic: ah right ... as long as we are not introducing some "type" field or similar ... that might work. Let's explore that. But anyway, "lz4" is not a _class_ to be exact. But if it simplifies things, why not ...

conf/cassandra.yaml
michaelsembwever: `class_name` is the standard in yaml and `class` in table options.

smiklosovic: Yeah, so `class_name`. ParameterizedClass has it everywhere like `class_name` if I am not mistaken. Properties for compressor would be "parameters"

    public class ParameterizedClass
    {
        public static final String CLASS_NAME = "class_name";
        public static final String PARAMETERS = "parameters";

        public String class_name;
        public Map<String, String> parameters;

        public ParameterizedClass()
        {
            // for snakeyaml
        }

I think it is better to follow yaml naming (class_name and parameters) and then parameters would have keys and values as in CQL to have it generic.

PR #2226, opened 2023-03-16
Mck/18049/trunk

michaelsembwever:


src/java/org/apache/cassandra/utils/binlog/BinLog.java
```diff
@@ -75,6 +77,12 @@ public class BinLog implements Runnable
     public static final String VERSION = "version";
     public static final String TYPE = "type";
 
+    static
+    {
+        // https://github.com/OpenHFT/Chronicle-Core/blob/chronicle-core-2.23.36/src/main/java/net/openhft/chronicle/core/announcer/Announcer.java#L32-L33
+        System.setProperty("chronicle.announcer.disable", "true");
```
ekaterinadimitrova2, line 84: Link to the discussion from the commit so we do not lose it - https://github.com/apache/cassandra/commit/30e8efb97475e3d53e9f5261ec9a488b1bef1980#r104781187

.build/cassandra-deps-template.xml
```diff
@@ -211,6 +213,40 @@
       <groupId>net.openhft</groupId>
       <artifactId>chronicle-threads</artifactId>
     </dependency>
+    <dependency>
+      <!-- transitive to chronicle-core -->
```
ekaterinadimitrova2, line 217: So, in order to get a higher version, we add them explicitly. Can we put a short explanation as saying "transitive", the reason for adding them explicitly doesn't come immediately to my mind? It reads a bit confusing

conf/jvm11-server.options
```diff
@@ -61,6 +61,7 @@
 -Djdk.attach.allowAttachSelf=true
 --add-exports java.base/jdk.internal.misc=ALL-UNNAMED
 --add-exports java.base/jdk.internal.ref=ALL-UNNAMED
+--add-exports java.base/jdk.internal.util=ALL-UNNAMED
```
ekaterinadimitrova2, line 64:
```
the 8x faster version of BytesInternal.contentEqual*(..) (and BytesStore.equals(..)) on jdk11+ (using SIMD comparison instruction) needs `--illegal-access=permit --add-exports java.base/jdk.internal.ref=ALL-UNNAMED --add-exports java.base/jdk.internal.util=ALL-UNNAMED`, ref [here](https://github.com/OpenHFT/Chronicle-Bytes/compare/chronicle-bytes-2.20.111...chronicle-bytes-2.23.33#diff-7be2a8746677c85a85b22da547a6132e51ba0a5b61538f9b19d3e8f6da43be20R115)
this opens for [CASSANDRA-13903](https://issues.apache.org/jira/browse/CASSANDRA-13903)
```
 Currently we do not use `BytesInternal.contentEqual*(..) (and BytesStore.equals(..)) ` but moving to a new version we need those exports for those in order to use them. Is that correct statement? Also, the reference points to a long list of commits not sure whether that was the right link you wanted to post

src/java/org/apache/cassandra/utils/binlog/BinLog.java
ekaterinadimitrova2: Considering your comment, I would say we can add a comment that enabling this will lead only to duplication of information and logs pollution

.build/parent-pom-template.xml
```diff
@@ -753,21 +804,13 @@
...
-            <groupId>net.java.dev.jna</groupId>
-          </exclusion>
-        </exclusions>
+        <version>2.23.25</version>
+      </dependency>
+      <dependency>
+        <!-- transitive to chronicle-queue -->
+        <groupId>net.openhft</groupId>
+        <artifactId>posix</artifactId>
+        <version>2.24ea4</version>
```
ekaterinadimitrova2, line 813: ea version? Shouldn't we use 2.23.2?

.build/parent-pom-template.xml
```diff
@@ -705,7 +756,7 @@
       <dependency>
         <groupId>net.openhft</groupId>
         <artifactId>chronicle-queue</artifactId>
-        <version>5.20.123</version>
```
ekaterinadimitrova2: Which bom are those? It would be nice to add it here in a comment

src/java/org/apache/cassandra/utils/binlog/BinLog.java
```diff
@@ -170,6 +178,7 @@ public synchronized void stop() throws InterruptedException
 
         shouldContinue = false;
         sampleQueue.put(NO_OP);
+        BackgroundResourceReleaser.stop();
```
ekaterinadimitrova2, line 182: To confirm my understanding - `BackgroundResourceReleaser.stop();` was missed to be done in the past? Bug?

ekaterinadimitrova2 (commented):
Left some small question and comments

.build/parent-pom-template.xml
michaelsembwever: won't work with an earlier version. despite the `ea` it is still an unchangeable artifact in maven2 repo (i.e. it is not a snapshot)

.build/parent-pom-template.xml
michaelsembwever: no bom, as requested we are upgrading to the latest versions (that are available in the maven1 repo)

src/java/org/apache/cassandra/utils/binlog/BinLog.java
michaelsembwever: the `stop()` method did not exist previously
https://github.com/OpenHFT/Chronicle-Core/blob/chronicle-core-2.20.126/src/main/java/net/openhft/chronicle/core/io/BackgroundResourceReleaser.java 

(and failing to stop it previously did not cause the failures it does now)

conf/jvm11-server.options
michaelsembwever: it is used internally by chronicle-*

the link is correct, but you have to manually go to `Files Changed`, scroll down to the  `BytesInternal.java` file, and `Load diff` to see the highlighted line.

src/java/org/apache/cassandra/utils/binlog/BinLog.java
ekaterinadimitrova2: ok

.build/parent-pom-template.xml
ekaterinadimitrova2: And those latest versions still correspond to specific BOM number, probably latest and I see in the past we were adding those looking at a BOM as my understanding is that the versions of the different dependencies which are listed in a BOM have been integrated and tested to work together.
I was looking into this one - https://mvnrepository.com/artifact/net.openhft/chronicle-bom/2.23.137
Then I also realized that we do not use the latest ea release (as we see we do with posix) but the previous one
https://mvnrepository.com/artifact/net.openhft/chronicle-queue

At first I thought you do not want to use ea version, but then you used for posix?

I think adding the BOM number in a comment reminds people not to update separate dependency when it is known that some should be upgraded together to tested compatible versions

michaelsembwever: I started off with the bom, then going through maven1 repo used the versions available (basically aligning to the bom, some patch versions differed).

Then fixing tests some versions needed bumping, posix the example here.

Then on the dev@ thread folk requested we use the latest versions available. The versions that then got bumped made sense to as they matched sibling libraries we already had (e.g. jna*), and/or were about newer patch versions supporting more arch/platforms.

While the initial goal may have been to re-use the tested QA of the bom's declaration, that's evolved to tailoring it best (with evidence) to our needs and relying on being in trunk to do our own testing and QA.  (IMO this approach would not be appropriate in one of our release branches.)

ekaterinadimitrova2: I am still not sure why posix is at ea latest version but the chronicle-* dependencies are at non-ea  latest versions.
Also, maybe a typo but chronicle-core has 2.23.37, you have 36 here.

michaelsembwever: > I am still not sure why posix is at ea latest version but the chronicle-* dependencies are at non-ea latest versions.

Only the bump of posix was needed to fix the related failures.

( Many, and newer, versions are simply not found in https://repo1.maven.org/maven2/net/openhft/ )

> Also, maybe a typo but chronicle-core has 2.23.37, you have 36 here.

While chronicle-core-2.23.37 itself is available publicly, it's parent pom is not, and uses other dependencies that are SNAPSHOTs. So that version is out for us.

.build/cassandra-deps-template.xml
michaelsembwever: i had that edit already :D 

`git stash apply` on its way…

src/java/org/apache/cassandra/utils/binlog/BinLog.java
michaelsembwever: pushed update: https://github.com/michaelsembwever/cassandra/commit/60732e4a4f1561371d60b1ce26285531e1fecbb1

.build/cassandra-deps-template.xml
michaelsembwever: pushed update: https://github.com/michaelsembwever/cassandra/commit/60732e4a4f1561371d60b1ce26285531e1fecbb1

michaelsembwever:
pushed update: https://github.com/michaelsembwever/cassandra/commit/60732e4a4f1561371d60b1ce26285531e1fecbb1

.build/parent-pom-template.xml
ekaterinadimitrova2: Thank you for explaining, I see what you mean now! 
Can we add a few sentences to the ticket about that? I know people who are rebasing their forks are doing another pass of review so it might be helpful for context as it is not immediately obvious.

ekaterinadimitrova2 (approved):
This seems ready for another CI round. +1 on clean CI

michaelsembwever:
> another CI round. +1 on clean CI 

what's changed that's need another CI run?

PR #2233, opened 2023-03-20
[CASSANDRA-18153-5.0] Save host id to system.local and flush immediately after startup

jacek-lewandowski:


src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -269,6 +269,27 @@ public boolean isCompacting(Iterable<ColumnFamilyStore> cfses, Predicate<SSTable
...
+                           cacheCleanupExecutor.getPendingTaskCount() +
+                           secondaryIndexExecutor.getPendingTaskCount();
+
+        int activeTasks = executor.getActiveTaskCount() +
+                          validationExecutor.getActiveTaskCount() +
+                          viewBuildExecutor.getActiveTaskCount() +
+                          cacheCleanupExecutor.getActiveTaskCount() +
+                          secondaryIndexExecutor.getActiveTaskCount();
+
+        return pendingTasks > 0 || activeTasks > 0;
```
smiklosovic: if `pendingTasks > 0`, we do not need to compute `activeTasks`, no? We can `return` after computing `pendingTasks` already.

        int pendingTasks = executor.getPendingTaskCount() +
                           validationExecutor.getPendingTaskCount() +
                           viewBuildExecutor.getPendingTaskCount() +
                           cacheCleanupExecutor.getPendingTaskCount() +
                           secondaryIndexExecutor.getPendingTaskCount();

        if (pendingTasks > 0)
            return true;

        int activeTasks = executor.getActiveTaskCount() +
                          validationExecutor.getActiveTaskCount() +
                          viewBuildExecutor.getActiveTaskCount() +
                          cacheCleanupExecutor.getActiveTaskCount() +
                          secondaryIndexExecutor.getActiveTaskCount();

        return activeTasks > 0;

test/unit/org/apache/cassandra/io/sstable/keycache/KeyCacheTest.java
```diff
@@ -396,7 +395,6 @@ public void testKeyCacheLoadTwoTablesTime() throws Exception
         // be zero after load.
         assertKeyCacheSize(numberOfRows, KEYSPACE1, columnFamily1);
         assertKeyCacheSize(numberOfRows, KEYSPACE2, columnFamily2);
-        assertEquals(sstableImplCachesKeys ? numberOfRows * tables.size() : 0, CacheService.instance.keyCache.size());
```
smiklosovic, line 399: why is this removed (above as well)? is that redundant? What is the relationship of this change to this ticket?

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
jacek-lewandowski: yes, we can, though I don't think there is any computation except summing

test/unit/org/apache/cassandra/io/sstable/keycache/KeyCacheTest.java
jacek-lewandowski: Because it asserts the global cache size which is out of control of this test. The test only controls the cache for the `KEYSPACE1` and `KEYSPACE2`, and when something gets written to the cache because of say system keyspace read/update, the test will fail.

PR #2236, opened 2023-03-21
CASSANDRA-18037 trunk: Use snake case for the names CQL native functions

adelapena:


src/java/org/apache/cassandra/cql3/functions/BytesConversionFcts.java
```diff
@@ -18,92 +18,104 @@
...
 import org.apache.cassandra.exceptions.InvalidRequestException;
 import org.apache.cassandra.serializers.MarshalException;
+import org.apache.cassandra.transport.ProtocolVersion;
+import org.apache.cassandra.utils.ByteBufferUtil;
 
 public abstract class BytesConversionFcts
 {
     public static void addFunctionsTo(NativeFunctions functions)
     {
-        // because text and varchar ends up being synonymous, our automatic makeToBlobFunction doesn't work
```
bereng, line 38: This has been removed in the new code?

adelapena: I have changed the function builders to receive the CQL type, instead of the internal one. That makes the special-casing for `varchar` and the `Set` unnecessary.

src/java/org/apache/cassandra/cql3/functions/CastFcts.java
```diff
@@ -126,25 +130,38 @@ private static <I extends Number> java.util.function.Function<I, BigDecimal> get
...
+     * Creates the legacy camel-cased name of the cast function used to cast to the specified type.
      *
      * @param outputType the output type
-     * @return the name of the cast function use to cast to the specified type
+     * @return the legacy camel-cased name of the cast function used to cast to the specified type
      */
-    public static String getFunctionName(CQL3Type outputType)
+    private static String getLegacyFunctionName(CQL3Type outputType)
+    {
+        return LEGACY_FUNCTION_NAME_PREFIX + WordUtils.capitalize(toLowerCaseString(outputType));
```
bereng: Will `WordUtils.capitalize()` ever kick in? It's going to be fed the lowercase class name, iiuc, which will never have blank spaces

src/java/org/apache/cassandra/cql3/functions/TimeFcts.java
```diff
@@ -101,94 +136,154 @@ public ByteBuffer execute(ProtocolVersion protocolVersion, List<ByteBuffer> para
...
-                return ByteBufferUtil.bytes(type.toTimeInMillis(bb));
-            }
-
-            @Override
-            public boolean isMonotonic()
-            {
-                return true;
-            }
-        };
+            super(useLegacyName ? "to_timestamp" : "totimestamp", TimestampType.instance, type);
```
bereng: Shouldn't this be reversed? There are a few like this one. I would suggest reviewing all just in case if my comment is correct and adding a quick test

src/java/org/apache/cassandra/cql3/functions/CastFcts.java
adelapena: It would, since the first word be capitalized independently of spaces or underscores. For example, `int` would become `Int`.

What we could do here, by the way, is replacing the deprecated `WordUtils.capitalize` by the newer `StringUtils.capitalize`.

src/java/org/apache/cassandra/cql3/functions/TimeFcts.java
adelapena: It should, good catch. I think I had a some point the inverse condition (`useNewName`) and forgot to update on refactor. Incidentally it wasn't breaking anything because in the end we add both names to the collection of native functions and we don't make any distinction between them. I have fixed it and added a new check to the existing test.

src/java/org/apache/cassandra/cql3/functions/CastFcts.java
bereng: Ah ok the first letter, yeah that's good.

test/unit/org/apache/cassandra/cql3/functions/NativeFunctionsTest.java
```diff
@@ -148,16 +148,18 @@ public void testDeprectedFunctionNames()
 
             Assertions.assertThat(newFunction).isNotNull();
             Assertions.assertThat(function).isNotEqualTo(newFunction);
+            Assertions.assertThat(function).isEqualTo(((NativeFunction) newFunction).withLegacyName());
             Assertions.assertThat(function.argTypes()).isEqualTo(newFunction.argTypes());
             Assertions.assertThat(function.returnType()).isEqualTo(newFunction.returnType());
             Assertions.assertThat(function.getClass()).isEqualTo(newFunction.getClass());
             Assertions.assertThat(function.name().name.toLowerCase())
                       .isEqualTo(StringUtils.remove(newFunction.name().name, '_'));
+
```
bereng: nit extra space

PR #2237, opened 2023-03-21
CASSANDRA-18129 Accord <-> Paxos live migration

aweisberg:


src/java/org/apache/cassandra/concurrent/SyncFutureTask.java
```diff
@@ -71,7 +71,11 @@ public void run()
         catch (Throwable t)
         {
             tryFailure(t);
-            ExecutionFailure.handle(t);
+            // A lot of exceptions are expected and will be handled by Cassandra
+            // by consuming the result of the future task so only treat Error
+            // as uncaught
+            if (t instanceof Error)
```
bdeggleston, line 77: What problem is this solving here? Not sure if only handling `Error` as an unhandled exception is a good idea. We also want to handle things like IOExceptions a certain way, and this relies on a registered callback doing the right thing.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2957,6 +2974,10 @@ public static boolean paxoTopologyRepairStrictEachQuorum()
         return conf.paxos_topology_repair_strict_each_quorum;
     }
 
+    // TODO imperative that at startup we check that the legacy paxos strategy is compatible
+    // with cluster metadata for consensus migration
+    // If another node updates the CM with consensus migration and we don't support it that is a serious issue as well
+    // Ideally this config could only ever exist in CM so that the cluster always agrees on what it is
```
bdeggleston, line 2986: `LegacyPaxosStrategy` this seems unintuitive as a config parameter since user facing docs typically refer to paxos operations  as light weight transactions. Something like `LightweightTransactionProtocol{paxos, accord}` would make it clearer I think.

Also, from a configuration perspective, it seems like this is more of a starting state, with the actual state being determined by cluster metadata and any migration history it contains.

src/java/org/apache/cassandra/cql3/Lists.java
```diff
@@ -52,6 +59,8 @@
  */
 public abstract class Lists
 {
+    private static final Logger logger = LoggerFactory.getLogger(Lists.class);
```
bdeggleston, line 64: unused

src/java/org/apache/cassandra/cql3/Lists.java
```diff
@@ -524,11 +533,18 @@ static void doAppend(Term.Terminal value, ColumnMetadata column, UpdateParameter
...
                 int dataSize = 0;
                 for (ByteBuffer buffer : elements)
                 {
-                    ByteBuffer uuid = ByteBuffer.wrap(params.nextTimeUUIDAsBytes());
-                    Cell<?> cell = params.addCell(column, CellPath.create(uuid), buffer);
+                    ByteBuffer cellPath;
+                    // Accord will need to replace this value later once it knows the executeAt timestamp
+                    // so just put a TimeUUID with MSB centinel for now
+                    if (params.constructingAccordBaseUpdate)
+                        cellPath = TimeUUID.atUnixMicrosWithLsb(0, cellIndex++).toBytes();
```
bdeggleston, line 580: we should consolidate accord cell path logic in the CellPath class and use something like `CellPath#cellPathForAccordPlaceholder` here. It could also go in this class. Either way, the logic for producing and reading/replacing special values should be in the same place

src/java/org/apache/cassandra/cql3/statements/CQL3CasRequest.java
```diff
@@ -65,20 +67,26 @@
...
+import static org.apache.cassandra.service.accord.txn.TxnDataName.Kind.CAS_READ;
+import static org.apache.cassandra.service.accord.txn.TxnResult.Kind.retry_new_protocol;
+
 
 /**
  * Processed CAS conditions and update on potentially multiple rows of the same partition.
  */
 public class CQL3CasRequest implements CASRequest
 {
+    private static final Logger logger = LoggerFactory.getLogger(CQL3CasRequest.class);
```
bdeggleston, line 93: unused

src/java/org/apache/cassandra/db/rows/Cell.java
```diff
@@ -41,6 +46,8 @@
  */
 public abstract class Cell<V> extends ColumnData
 {
+    private static final Logger logger = LoggerFactory.getLogger(Cell.class);
```
bdeggleston, line 50: unused

src/java/org/apache/cassandra/db/rows/Cell.java
```diff
@@ -282,7 +289,8 @@ public <V> Cell<V> deserialize(DataInputPlus in, LivenessInfo rowLiveness, Colum
                 }
             }
 
-            return accessor.factory().cell(column, timestamp, ttl, localDeletionTime, value, path);
+            Cell cell = accessor.factory().cell(column, timestamp, ttl, localDeletionTime, value, path);
```
bdeggleston: unnecessary assignment here, you can just return from the factory

src/java/org/apache/cassandra/db/rows/ColumnData.java
```diff
@@ -284,7 +287,19 @@ public static void digest(Digest digest, ColumnData cd)
      * This exists for the Paxos path, see {@link PartitionUpdate#updateAllTimestamp} for additional details.
      */
     public abstract ColumnData updateAllTimestamp(long newTimestamp);
-    public abstract ColumnData updateAllTimestampAndLocalDeletionTime(long newTimestamp, int newLocalDeletionTime);
+
+    /**
+     * @param cellToMaybeNewListPath If the cell is a list append cell a new cell path is returned generated based on the Accord executeAt timestamp
+     */
+    public abstract ColumnData updateAllTimesForAccord(@Nonnull Function<Cell, CellPath> cellToMaybeNewListPath, long newTimestamp, int newLocalDeletionTime);
```
bdeggleston: could this be renamed to `updateTimesAndPathsForAccord`?

src/java/org/apache/cassandra/dht/Range.java
```diff
@@ -40,6 +60,34 @@
 {
     public static final long serialVersionUID = 1L;
 
+    public static final boolean EXPENSIVE_CHECKS = Boolean.getBoolean("org.apache.cassandra.dht.Range.expensive_checks") | true;
```
bdeggleston: this is OR'd to true

src/java/org/apache/cassandra/metrics/TableMetrics.java
```diff
@@ -80,6 +80,11 @@
     public final static LatencyMetrics GLOBAL_WRITE_LATENCY = new LatencyMetrics(GLOBAL_FACTORY, GLOBAL_ALIAS_FACTORY, "Write");
     public final static LatencyMetrics GLOBAL_RANGE_LATENCY = new LatencyMetrics(GLOBAL_FACTORY, GLOBAL_ALIAS_FACTORY, "Range");
 
+    // TODO is the overhead of tracking latency of key and range migration too at the leel of detail of reads and writes too much?
```
bdeggleston: if the overhead of tracking reads and writes is acceptable, I'd imagine tracking migration latency is as well. It will also be useful when understanding what's happening in a cluster.

src/java/org/apache/cassandra/net/Verb.java
```diff
@@ -259,6 +288,13 @@
     ACCORD_GET_DEPS_RSP         (148, P2, writeTimeout, REQUEST_RESPONSE, () -> GetDepsSerializers.reply, RESPONSE_HANDLER),
     ACCORD_GET_DEPS_REQ         (147, P2, writeTimeout, ACCORD,               () -> GetDepsSerializers.request,       () -> AccordService.instance().verbHandler(), ACCORD_GET_DEPS_RSP),
 
+    CONSENSUS_KEY_MIGRATION_FINISHED(149, P1, writeTimeout, MUTATION, () -> ConsensusKeyMigrationFinished.serializer, () -> ConsensusKeyMigrationState.consensusKeyMigrationFinishedHandler),
```
bdeggleston: can you match the format of the other verb declarations?

src/java/org/apache/cassandra/net/Verb.java
```diff
@@ -259,6 +288,13 @@
...
+
+    UPDATE_CM_RSP(151, P1, writeTimeout, REQUEST_RESPONSE, () -> NoPayload.serializer, () -> ResponseVerbHandler.instance),
```
bdeggleston, line 293: Can the CM abbreviation be expanded?

src/java/org/apache/cassandra/repair/AccordRepairJob.java
```diff
@@ -0,0 +1,165 @@
...
+                else
+                {
+                    toRepair = splitter.subRange(range, offset, splitter.add(offset, length));
+                    checkState(iteration > 1 || toRepair.start().equals(range.start()));
+                }
+                checkState(!toRepair.equals(lastRepaired), "Shouldn't repair the same range twice");
+                checkState(lastRepaired == null || toRepair.start().equals(lastRepaired.end()), "Next range should directly follow previous range");
+                lastRepaired = toRepair;
+                // TODO Won't work until range transactions work
+//                AccordService.instance().barrier(toRepair, minEpoch.getEpoch(), nanoTime(), BarrierType.global_sync, false);
```
bdeggleston, line 132: Since regular repairs are incompatible with accord, accord repair should ensure that _all_ replicas are up to date to remain consistent with data replication guarantees, not just a quorum. The exception here would be for repairs run with --force, and possibly for Accord->Paxos migrations.

src/java/org/apache/cassandra/repair/CassandraRepairJob.java
```diff
@@ -86,13 +86,10 @@ public class RepairJob extends AsyncFuture<RepairResult> implements Runnable
      *  @param session RepairSession that this RepairJob belongs
      * @param columnFamily name of the ColumnFamily to repair
      */
-    public RepairJob(RepairSession session, String columnFamily)
+    public CassandraRepairJob(RepairSession session, String columnFamily)
     {
-        this.session = session;
-        this.taskExecutor = session.taskExecutor;
+        super(session, columnFamily);
```
bdeggleston, line 91: standard repair isn't going to be compatible with tables actively using accord. We should have a method, probably in AccordService, that will indicate if there is any accord history for the table and refuse to run a normal repair if there is.

src/java/org/apache/cassandra/repair/messages/RepairOption.java
```diff
@@ -164,6 +172,12 @@ public static Set<Range<Token>> parseRanges(String rangesStr, IPartitioner parti
      *             ranges to the same host multiple times</td>
      *             <td>false</td>
      *         </tr>
+     *         <tr>
+     *             <td>accordRepair</td>
+     *             <td>"true" if the repair should be of Accord in flight transactions. Will ensure
+     *             that once repair completes all Accord transactions are replicated at quorum</td>
+     *             <td>false</td>
```
bdeggleston, line 179: assuming we adjust accord repair guarantees, the description for the accord option will need adjusting

src/java/org/apache/cassandra/service/ActiveRepairService.java
```diff
@@ -664,6 +679,7 @@ public boolean invokeOnFailure()
             {
                 // we pre-filter the endpoints we want to repair for forced incremental repairs. So if any of the
                 // remaining ones go down, we still want to fail so we don't create repair sessions that can't complete
+                // TODO should there be a different default/behavior for Accord repairs?
```
bdeggleston, line 683: for AccordRepair, I think we should fail if we're not able to get a barrier accepted by a quorum if it's a force repair, and fail on any down nodes if not a forced repair.

src/java/org/apache/cassandra/service/ConsensusKeyMigrationState.java
```diff
@@ -0,0 +1,359 @@
...
+
+    private static final int EMPTY_KEY_SIZE = Ints.checkedCast(ObjectSizes.measureDeep(Pair.create(null, UUID.randomUUID())));
+    private static final int VALUE_SIZE = Ints.checkedCast(ObjectSizes.measureDeep(new ConsensusMigratedAt(Epoch.EMPTY, ConsensusMigrationTarget.accord)));
+    private static final Cache<Pair<ByteBuffer, UUID>, ConsensusMigratedAt> MIGRATION_STATE_CACHE =
+        Caffeine.newBuilder()
+                .maximumWeight(DatabaseDescriptor.getConsensusMigrationCacheSizeInMiB() << 20)
+                .<Pair<ByteBuffer, UUID>, ConsensusMigratedAt>weigher((k, v) -> EMPTY_KEY_SIZE + Ints.checkedCast(ByteBufferUtil.estimatedSizeOnHeap(k.left)) + VALUE_SIZE)
+                .executor(ImmediateExecutor.INSTANCE)
+                .build();
+    private static final Function<Pair<ByteBuffer, UUID>, ConsensusMigratedAt> LOADING_FUNCTION = k -> SystemKeyspace.loadConsensusKeyMigrationState(k.left, k.right);
```
bdeggleston: can this be moved into the cache `get` that uses it?

src/java/org/apache/cassandra/service/ConsensusTableMigrationState.java
```diff
@@ -0,0 +1,697 @@
...
+                    throw new IllegalArgumentException(value + " is not recognized");
+                case 0:
+                    return ConsensusMigrationRepairType.paxos;
+                case 1:
+                    return ConsensusMigrationRepairType.accord;
+            }
+        }
+    }
+
+    public enum ConsensusMigrationTarget {
```
bdeggleston: brace formatting

src/java/org/apache/cassandra/service/ConsensusTableMigrationState.java
```diff
@@ -0,0 +1,697 @@
...
+                    return paxos;
+                case 1:
+                    return accord;
+            }
+        }
+    }
+
+    public static class ConsensusMigrationRepairResult
+    {
+        private final ConsensusMigrationRepairType consensusMigrationRepairType;
```
bdeggleston: can we just name this `type`?

src/java/org/apache/cassandra/service/ConsensusTableMigrationState.java
```diff
@@ -0,0 +1,697 @@
...
+            this.minEpoch = minEpoch;
+        }
+
+        public static ConsensusMigrationRepairResult fromCassandraRepair(Epoch minEpoch, boolean didPaxosAndRegularRepair)
+        {
+            checkArgument(!didPaxosAndRegularRepair || minEpoch.isAfter(Epoch.EMPTY), "Epoch should not be empty if Paxos and regular repairs were performed");
+            if (didPaxosAndRegularRepair)
+                return new ConsensusMigrationRepairResult(ConsensusMigrationRepairType.paxos, minEpoch);
+            else
+                return new ConsensusMigrationRepairResult(ConsensusMigrationRepairType.ineligible, Epoch.EMPTY);
```
bdeggleston: a repair should also be ineligible if it was run with `--force` and repair then skipped one or more nodes

src/java/org/apache/cassandra/service/ConsensusTableMigrationState.java
```diff
@@ -0,0 +1,697 @@
...
+import static org.apache.cassandra.utils.CollectionSerializers.newHashMap;
+import static org.apache.cassandra.utils.CollectionSerializers.serializeCollection;
+import static org.apache.cassandra.utils.CollectionSerializers.serializeMap;
+import static org.apache.cassandra.utils.CollectionSerializers.serializedCollectionSize;
+import static org.apache.cassandra.utils.CollectionSerializers.serializedMapSize;
+
+/**
+ * Track and update the migration state of individual table and ranges within those tables
+ */
+public abstract class ConsensusTableMigrationState
```
bdeggleston: Let's make an `o.a.c.service.consensus` package, and possibly an `o.a.c.service.consensus.migration` package, and move some of this stuff in there

src/java/org/apache/cassandra/service/ConsensusTableMigrationState.java
```diff
@@ -0,0 +1,697 @@
...
+        ClusterMetadataService.instance.commit(new SetConsensusMigrationTargetProtocol(targetProtocol, tableIds));
+    }
+
+    public static void startMigrationToConsensusProtocol(@Nonnull String targetProtocolName,
+                                                         @Nonnull List<String> keyspaceNames,
+                                                         @Nonnull Optional<List<String>> maybeTables,
+                                                         @Nonnull Optional<String> maybeRangesStr)
+    {
+        checkArgument(!keyspaceNames.isEmpty(), "At least one keyspace must be specified");
+        checkArgument(keyspaceNames.size() == 1 || !maybeTables.isPresent(), "Can't specify tables with multiple keyspaces");
```
bdeggleston: could we interpret empty keyspaces as a command to start migration of all key spaces?

src/java/org/apache/cassandra/service/ConsensusTableMigrationState.java
```diff
@@ -0,0 +1,697 @@
...
+                                  .collect(Collectors.toList()));
+            this.migratingRanges = ImmutableList.copyOf(normalize(migratingRangesByEpoch.values().stream().flatMap(Collection::stream).collect(Collectors.toList())));
+            this.migratingAndMigratedRanges = ImmutableList.copyOf(normalize(ImmutableList.<Range<Token>>builder().addAll(migratedRanges).addAll(migratingRanges).build()));
+        }
+
+        public TableMigrationState withRangesMigrating(@Nonnull Collection<Range<Token>> ranges,
+                                                       @Nonnull ConsensusMigrationTarget target,
+                                                       @Nonnull Epoch epoch)
+        {
+            checkArgument(epoch.isAfter(Epoch.EMPTY), "Epoch shouldn't be empty");
```
bdeggleston: should validation be more tolerant of overlapping/duplicate begin migration commands? I don't think there's any real downside to logging that the range has already been migrated or is migrating so the command will be either be a noop or will affect a subset of the specified ranges. I'd imagine this will make life easier for operators at least.

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -2995,6 +3048,39 @@ public final boolean equals(Object o)
         }
     }
 
+    public static class ConsensusAttemptResult
```
bdeggleston, line 3092: wdyt about having a single private ctor, and having 3 static construction methods that named something like `txnResult`, `paxosResult`, and `switchProtocol`, or something like that? At least in the case of `retry_new_protocol`, it would make it a bit easier to understand what's happening without having to remember how the various ctor invocations affect it's state

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -83,6 +90,10 @@ private static long getThreadId(ExecutorService executor)
     private final ProgressLog progressLog;
     private final RangesForEpochHolder rangesForEpochHolder;
 
+    // TODO (expected): schedule regular pruning of this collection, and this is not persistent!
+    @Nullable
+    ReducingRangeMap<Timestamp> rejectBefore;
```
bdeggleston, line 95: should this be addressed? I think persistence could be mostly copied from paxos repair history as well, and scheduling a periodic command store maintenance task is pretty reasonable.

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -110,6 +121,38 @@ public int id()
...
+    @Override
+    public void setRejectBefore(ReducingRangeMap<Timestamp> newRejectBefore)
+    {
+        this.rejectBefore = newRejectBefore;
+    }
+
+    @Override
+    public Timestamp preaccept(TxnId txnId, Seekables<?, ?> keys, SafeCommandStore safeStore)
+    {
+        NodeTimeService time = safeStore.time();
```
bdeggleston, line 133: see my accord review comment about generalizing this on `InMemoryCommandStore`

src/java/org/apache/cassandra/service/accord/serializers/RecoverySerializers.java
```diff
@@ -90,11 +91,12 @@ void serializeOk(RecoverOk recoverOk, DataOutputPlus out, int version) throws IO
             CommandSerializers.ballot.serialize(recoverOk.accepted, out, version);
             serializeNullable(recoverOk.executeAt, out, version, CommandSerializers.timestamp);
             DepsSerializer.partialDeps.serialize(recoverOk.deps, out, version);
+            DepsSerializer.partialDeps.serialize(recoverOk.acceptedDeps, out, version);
```
bdeggleston, line 94: in many cases, I'd expect these to be identical to `deps`. Can we add a flags field, and use it to indicate `acceptDeps == deps`, and skip serialization/deserialization?

bdeggleston, line 94: in many cases, I'd expect these to be identical to `deps`. Can we add a flags field, and use it to indicate `acceptDeps == deps`, and skip serialization/deserialization?

src/java/org/apache/cassandra/service/accord/txn/TxnWrite.java
```diff
@@ -48,24 +52,28 @@
...
 
-import static org.apache.cassandra.utils.ArraySerializers.deserializeArray;
 import static org.apache.cassandra.service.accord.AccordSerializers.partitionUpdateSerializer;
+import static org.apache.cassandra.utils.ArraySerializers.deserializeArray;
 import static org.apache.cassandra.utils.ArraySerializers.serializeArray;
 import static org.apache.cassandra.utils.ArraySerializers.serializedArraySize;
 
 public class TxnWrite extends AbstractKeySorted<TxnWrite.Update> implements Write
 {
+    private static final Logger logger = LoggerFactory.getLogger(TxnWrite.class);
```
bdeggleston, line 77: unused

src/java/org/apache/cassandra/service/accord/txn/TxnNamedRead.java
```diff
@@ -122,24 +149,68 @@ public AsyncChain<Data> read(boolean isForWriteTxn, SafeCommandStore safeStore,
         // immediately after the transaction executed, and this simplifies things a great deal
         int nowInSeconds = (int) TimeUnit.MICROSECONDS.toSeconds(executeAt.hlc());
 
+        if (ConsensusRequestRouter.instance.isKeyInMigratingRangeFromPaxos(tms, key))
+            return performCoordinatedRead(consistencyLevel, command, nowInSeconds, metrics);
+        else
+            return performLocalRead(command, nowInSeconds);
+    }
+
+    private AsyncChain<Data> performCoordinatedRead(ConsistencyLevel consistencyLevel, SinglePartitionReadCommand command, int nowInSeconds, AccordClientRequestMetrics metrics)
```
bdeggleston: I think it would be possible to read data from 'the future'. If we read from a replica that has already applied this operation, we will break correctness. This is especially possible in recovery.

src/java/org/apache/cassandra/service/paxos/PaxosPrepare.java
```diff
@@ -877,24 +909,28 @@ public synchronized void onRefreshSuccess(Ballot isSupersededBy, InetAddressAndP
         final DecoratedKey partitionKey;
         final TableMetadata table;
 
-        AbstractRequest(Ballot ballot, Electorate electorate, SinglePartitionReadCommand read, boolean isForWrite)
+        final boolean isForRepair;
```
bdeggleston: this appears to be unused

src/java/org/apache/cassandra/service/paxos/PaxosPropose.java
```diff
@@ -126,6 +138,10 @@ private static class MaybeFailure extends Status
     final int participants;
     /** Number of accepts required */
     final int required;
+
+    /** Repairing an in flight txn not proposing a new one **/
+    final boolean isForRepair;
```
bdeggleston: could we call this `isForRecovery` to avoid confusion with actual repair?

src/java/org/apache/cassandra/service/paxos/PaxosState.java
```diff
@@ -79,6 +84,8 @@
  */
 public class PaxosState implements PaxosOperationLock
 {
+    private static final Logger logger = LoggerFactory.getLogger(PaxosState.class.getName());
```
bdeggleston, line 90: unused

src/java/org/apache/cassandra/service/paxos/PaxosState.java
```diff
@@ -633,28 +640,36 @@ else if (isAfter(ballot, latestWriteOrLowBound))
...
+                shouldRejectDueToConsensusMigration = false;
+            else
+                shouldRejectDueToConsensusMigration = ConsensusRequestRouter.instance
+                                                          .isKeyInMigratingOrMigratedRangeDuringPaxosAccept(proposal.update.metadata().id,
+                                                                                                            proposal.update.partitionKey());
             if (!proposal.isSameOrAfter(latest))
             {
                 Tracing.trace("Rejecting proposal {}; latest is now {}", proposal.ballot, latest);
-                return latest;
+                return new AcceptResult(latest, shouldRejectDueToConsensusMigration);
```
bdeggleston, line 670: we can use singletons for each `shouldRejectDueToConsensusMigration` value

src/java/org/apache/cassandra/service/paxos/PaxosState.java
```diff
@@ -671,7 +686,8 @@ public Ballot acceptIfLatest(Proposal proposal)
         // though this
         Tracing.trace("Accepting proposal {}", proposal);
         SystemKeyspace.savePaxosProposal(proposal);
-        return null;
+        checkState(!shouldRejectDueToConsensusMigration);
+        return new AcceptResult();
```
bdeggleston: we can use a singleton here

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosCleanupHistory.java
```diff
@@ -47,21 +51,24 @@ public void serialize(PaxosCleanupHistory message, DataOutputPlus out, int versi
             message.tableId.serialize(out);
             message.highBound.serialize(out);
             PaxosRepairHistory.serializer.serialize(message.history, out, version);
+            Epoch.messageSerializer.serialize(message.epoch, out, version);
```
bdeggleston: this will need a version check. There will already be history entries serialized without an epoch in the wild

src/java/org/apache/cassandra/tools/nodetool/ConsensusMigrationAdmin.java
```diff
@@ -0,0 +1,111 @@
...
+        protected void execute(NodeProbe probe)
+        {
+            checkNotNull(startToken, "Start token must be specified");
+            checkNotNull(endToken, "End token must be specified");
+            checkArgument(schemaArgs.size() >= 2, "Must specify a keyspace and at least one table");
+            checkArgument((endToken != null && startToken != null) || (endToken == null && startToken == null), "Must specify start and end token together");
+            String maybeRangesStr = startToken != null ? startToken + ":" + endToken : null;
+            List<String> keyspaceNames = schemaArgs.size() > 0 ? singletonList(schemaArgs.get(0)) : emptyList();
+            List<String> maybeTableNames = schemaArgs.size() > 1 ? schemaArgs.subList(1, schemaArgs.size()) : null;
+            probe.getStorageService().migrateConsensusProtocol(targetProtocol, keyspaceNames, maybeTableNames, maybeRangesStr);
```
bdeggleston, line 86: We should output a message indicating that the specified tables/ranges need to be repaired to complete migration

src/java/org/apache/cassandra/tools/nodetool/ConsensusMigrationAdmin.java
```diff
@@ -0,0 +1,111 @@
...
+    public static class SetTargetProtocol extends ConsensusMigrationAdmin
+    {
+        @Arguments(usage = "[<keyspace> <tables>...]", description = "The keyspace followed by one or many tables")
+        private List<String> schemaArgs = new ArrayList<>();
+
+        @Option(title = "target_protocol", name = {"-tp", "--target-protocol"}, description = "Use -tp to specify what consensus protocol should be migrated to", required=true)
+        private String targetProtocol = null;
+
+        @Option(title = "force_completion", name = {"-f", "--force-completion"}, description = "Nuclear option to force the migration state for all ranges in the table regardless")
```
bdeggleston: how about `Forces migration state for all ranges of the specified table regardless of whether migration completed successfully or not. Should only be used if table is empty or has had no writes since last repair.`. This does seem useful in some scenarios, so I don't think the docs should be too scary, but we should be explicit about when it's ok to use.

src/java/org/apache/cassandra/tools/nodetool/Repair.java
```diff
@@ -126,6 +126,9 @@ else if (preview)
         }
     }
 
+    @Option(title = "accord_repair", name = {"-accord", "--repair-accord"}, description = "Repair in flight Accord transactions so their results are guaranteed replicated to a quorum of nodes")
```
bdeggleston, line 129: assuming we adjust accord repair guarantees, the description for the accord option will need adjusting

test/distributed/org/apache/cassandra/distributed/test/OptimiseStreamsRepairTest.java
```diff
@@ -107,7 +106,7 @@ public void testBasic() throws Exception
     {
         public static void install(ClassLoader cl, int id)
         {
-            new ByteBuddy().rebase(RepairJob.class)
+            new ByteBuddy().rebase(AbstractRepairJob.class)
```
bdeggleston: should this target `CassandraRepairJob`?

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailCollectionSizeOnSSTableWriteTest.java
```diff
@@ -434,4 +436,16 @@ private String failMessage(String key)
     {
         return String.format("Detected collection v in row %s in table %s of size", key, qualifiedTableName);
     }
+
+    private void createTable(String cql)
```
bdeggleston, line 440: why replace schemaChange w/ createTable which disables compaction? The tests seem to pass ok without this change

bdeggleston: also applies to `SSTableIdGenerationTest.java` and `GuardrailItemsPerCollectionOnSSTableWriteTest.java`

test/simulator/test/org/apache/cassandra/simulator/test/ShortAccordSimulationTest.java
```diff
@@ -19,13 +19,23 @@
...
 public class ShortAccordSimulationTest
 {
+    @BeforeClass
+    public static void beforeClass()
+    {
+        for (Map.Entry<Object, Object> entry : System.getProperties().entrySet())
+        {
+            System.out.printf("%s -- %s%n", entry.getKey(), entry.getValue());
+        }
+    }
```
bdeggleston: can we remove this?

test/simulator/test/org/apache/cassandra/simulator/test/ShortPaxosSimulationTest.java
```diff
@@ -20,24 +20,47 @@
...
-    @Test
-    public void simulationTest() throws IOException
-    {
-        PaxosSimulationRunner.main(new String[] { "run", "-n", "3..6", "-t", "1000", "-c", "2", "--cluster-action-limit", "2", "-s", "30" });
-    }
+//    @Test
+//    public void simulationTest() throws IOException
+//    {
+//        PaxosSimulationRunner.main(new String[] { "run", "-n", "3..6", "-t", "1000", "-c", "2", "--cluster-action-limit", "2", "-s", "30" });
+//    }
```
bdeggleston: can we uncomment or remove the commented out code in this class?

test/unit/org/apache/cassandra/CassandraTestBase.java
```diff
@@ -0,0 +1,203 @@
...
+        if (hasClassAnnotation(DDDaemonInitialization.class))
+            DatabaseDescriptor.daemonInitialization();
+        else if (hasClassAnnotation(SchemaLoaderPrepareServer.class))
+            SchemaLoader.prepareServer();
+        else if (hasClassAnnotation(SchemaLoaderLoadSchema.class))
+            SchemaLoader.loadSchema();
+        else
+            DatabaseDescriptor.setPartitionerUnsafe(ByteOrderedPartitioner.instance);
+
+        if (hasClassAnnotation(UseMurmur3Partitioner.class))
```
bdeggleston, line 92: This class is really cool. WDYT about failing if multiple partitioners are declared on a method? overkill?

test/unit/org/apache/cassandra/dht/RangeTest.java
```diff
@@ -736,4 +754,118 @@ public void testGroupSubtract()
...
+            }
+            a = ImmutableList.copyOf(normalize(a));
+            List<Range<Token>> b = new ArrayList();
+            for (int ii = 0; ii < numRanges; ii++)
+            {
+                b.add(new Range<>(new LongToken(r.nextLong()), new LongToken(r.nextLong())));
+            }
+            b = ImmutableList.copyOf(normalize(b));
+
+//            List<Range<Token>> a = ImmutableList.of(fromString("(-9223372036854775808,-868105389193617348]"), fromString(("(647608745828733177,-9223372036854775808]")));
```
bdeggleston: this should be removed

bdeggleston (changes requested):
I think we need a way to segregate normal and accord data for paxos->accord migration to work correctly. Otherwise the repair and coordinated read operations can replicate data from the future onto the local node. Off the top of my head this sounds like an sstable metadata component that keeps accord data separate from normal data and would allow us to exclude accord data from data repairs and coordinated reads. The good news is that this would solve the problem of how to run data repairs in clusters where users are mixing normal and accord operations, and if this metadata was the accord execution epoch, it would be pretty useful in bootstrap/range movement/and extended failure recovery.

aweisberg:
> I think we need a way to segregate normal and accord data for paxos->accord migration to work correctly. Otherwise the repair and coordinated read operations can replicate data from the future onto the local node.

There is definitely a problem here with coordinators and recovery coordinators. @dcapwell  just fixed [CASSANDRA-18422](https://issues.apache.org/jira/browse/CASSANDRA-18422?filter=-3) which is similar where two coordinators are concurrently reading and writing and one coordinator accidentally reads the writes of the other. The fix there was to invalidate any in-flight reads when committing a write.

The invalidation approach won't help with repair. Seem like for repair the data starts segregated, and we could keep it segregated, and then commit it into Accord in a range transaction?

> The good news is that this would solve the problem of how to run data repairs in clusters where users are mixing normal and accord operations, and if this metadata was the accord execution epoch, it would be pretty useful in bootstrap/range movement/and extended failure recovery.
Can you elaborate more on this stable component? What would it contain?


For coordinated reads. Maybe it makes sense to run those in Accord and route them through their respective command stores so invalidation works? I am not clear if that is better or worse than segregating Accord writes since I don't understand the thinking there yet.

src/java/org/apache/cassandra/concurrent/SyncFutureTask.java
aweisberg: These exceptions aren't uncaught they are stored in the `SyncFutureTask` and can eventually be picked up by the consumer of the `SyncFutureTask`.

This was causing failures in simulator because some expected exceptions like timeouts were going to the uncaught exception handler even though it was just timeouts caused by simulator induced chaos.

It's fine for these to go to the uncaught exception handler I think, but simulator shouldn't fail on expected exceptions so maybe that is what should change?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
aweisberg: Agree re LWT, `paxos` doesn't make sense because it doesn't actually mean `paxos` it means `paxos_and_accord` which is why for now it is called `migration`. WDYT would be better than `migration`?

This config option is not intended to be a starting state. It's intended to allow you to prevent accidental Paxos usage once you have migrated to Accord by switching from `migration` to `accord`.

belliottsmith:
I replied on slack. The problem found by David is quite distinct and only related to the local state. This problem is resolved in the bootstrap patch by disabling reads for txnid less than some point determined after the repair snapshot was taken. It might be we want to hook into that.

src/java/org/apache/cassandra/cql3/Lists.java
aweisberg: I can remove the unused loggers but it's pretty handy to have the logger already declared rather than having to go and manually bring it back. I think it's worth the one line, but don't feel strongly about it.

src/java/org/apache/cassandra/net/Verb.java
aweisberg: These are definitely going away once we merge with TCM since they are just a fake version of it. I'll rename them, but the length of the verbs is an issue if we are going to try and also get the formatting to match. `CONSENSUS_KEY_MIGRATION_FINISHED` isn't fitting, and I can either make it shorter or I can push out all the other lines to match. WDYT?

src/java/org/apache/cassandra/repair/AccordRepairJob.java
aweisberg: Does that make sense in the context of how this is used? When is Accord repair going to be run that is not for migration?

src/java/org/apache/cassandra/repair/CassandraRepairJob.java
aweisberg: That's not 100% straightforward because a table can be running both `Accord` and `Paxos` and can migrate back and forth in either direction. The property your describing maybe maps better to whether the range being repaired is currently running on Accord or Paxos?

src/java/org/apache/cassandra/service/ConsensusKeyMigrationState.java
aweisberg: Looking at how it is currently used maybe it make sense to build a `LoadingCache` and move it into the `Cache` construction?

src/java/org/apache/cassandra/service/ConsensusTableMigrationState.java
aweisberg: I agree. I'll convert the errors to logging.

aweisberg:
> This problem is resolved in the bootstrap patch by disabling reads for txnid less than some point determined after the repair snapshot was taken. It might be we want to hook into that.

The pain point for coordinated reads is that they aren't a quorum of reads from Accord command stores they are a regular quorum read in Cassandra so there is no filtering there to prevent picking up data from other nodes that have moved on to later transactions.

There is no Accord transaction that starts the migration to Accord for a range that is after some snapshot. Migration uses repair and repair is unreliable so it starts after the migration to Accord begins and keys are migrating before the repair even starts.

I think both repair and coordinated reads have to change.

src/java/org/apache/cassandra/service/ConsensusTableMigrationState.java
aweisberg: How does this not fail a style check? I'll look into that.

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
aweisberg: This is from Benedict's exclusive sync code. He might have something newer for this?

It isn't used in live migration because we don't do exclusive sync for live migration.

src/java/org/apache/cassandra/service/accord/serializers/RecoverySerializers.java
aweisberg: More exclusive sync code. Will check with Benedict.

src/java/org/apache/cassandra/service/accord/txn/TxnNamedRead.java
aweisberg: I think you are right. There is no path forward that doesn't end with the coordinated reads being done from inside Accord through command stores.

src/java/org/apache/cassandra/service/paxos/PaxosState.java
aweisberg: I don't 100% follow. You mean something like an enum for `shouldRejectDueToConsensusMigration` or constants aliasing true/false?

The `supersededBy` changes in `AcceptResult` so that can't be a singleton?

test/distributed/org/apache/cassandra/distributed/test/OptimiseStreamsRepairTest.java
aweisberg: Oh yikes and the test still passes. I'll fix that as well and make sure the injected version is getting called.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailCollectionSizeOnSSTableWriteTest.java
aweisberg: If I recall these tests didn't actually disable auto-compaction like they thought they were because disabling auto-compaction on a table that hasn't been created is a no op! The auto-compaction config is on the table not the keyspace.

I don't recall what got me looking at these tests in the first place. They might have just failed randomly on me.

It's renamed `createTable` because that is what it was used for, although technically it runs any `cql` and then does the auto-compaction change. Definitely open to changing the name to something like `

src/java/org/apache/cassandra/cql3/Lists.java
aweisberg: I think it fits better in `Lists` because `CellPath` doesn't dictate the contents and the contents here are specific to `Lists`. I'll move it here.

src/java/org/apache/cassandra/cql3/Lists.java
aweisberg: Added @SuppressWarnings("unused")

src/java/org/apache/cassandra/cql3/statements/CQL3CasRequest.java
aweisberg: Added @SuppressWarnings("unused")

src/java/org/apache/cassandra/db/rows/Cell.java
aweisberg: Added @SuppressWarnings("unused")

src/java/org/apache/cassandra/db/rows/Cell.java
aweisberg: Returned directly now

src/java/org/apache/cassandra/db/rows/ColumnData.java
aweisberg: Renamed in several places since this is a somewhat deep call chain from `TxnWrite` and there are enclosing methods that had the same name.

src/java/org/apache/cassandra/dht/Range.java
aweisberg: Fixed

src/java/org/apache/cassandra/net/Verb.java
aweisberg: I aligned the first few columns with the ones that are already properly aligned earlier. I couldn't align the second to last parameter because it was already to far to the right.

src/java/org/apache/cassandra/net/Verb.java
aweisberg: Since UPDATE_CM* is going away once we have TrM I left it alone.

src/java/org/apache/cassandra/repair/CassandraRepairJob.java
aweisberg: I left this alone until we decide how standard repair is going to interoperate with Accord.

src/java/org/apache/cassandra/service/ConsensusTableMigrationState.java
aweisberg: Only 945 misplaced braces according to checkstyle. Will have to do that later.

src/java/org/apache/cassandra/service/ConsensusTableMigrationState.java
aweisberg: What would go into the `consensus` package besides the `migration` sub package? Multi-word package names aren't common, but nesting an empty directory also seems weird.

src/java/org/apache/cassandra/service/ConsensusTableMigrationState.java
aweisberg: Added this behavior and a test

src/java/org/apache/cassandra/service/ConsensusTableMigrationState.java
aweisberg: Sure, changed it to add all the non-local key spaces if key spaces is empty or null.

src/java/org/apache/cassandra/service/paxos/PaxosPrepare.java
aweisberg: Hmm I wonder if that was some kind of error because it was actually in use!

src/java/org/apache/cassandra/service/paxos/PaxosPropose.java
aweisberg: Renamed it

src/java/org/apache/cassandra/service/paxos/PaxosState.java
aweisberg: Added a singleton for success and retry on new protocol.

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosCleanupHistory.java
aweisberg: Added missing version check

src/java/org/apache/cassandra/tools/nodetool/ConsensusMigrationAdmin.java
aweisberg: Added a message although hopefully we automate so this can be removed.

src/java/org/apache/cassandra/tools/nodetool/ConsensusMigrationAdmin.java
aweisberg: Sure. That doesn't really communicate that this is correctness impacting.

src/java/org/apache/cassandra/concurrent/SyncFutureTask.java
bdeggleston: >  simulator shouldn't fail on expected exceptions so maybe that is what should change?

That's probably the right thing to do

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
bdeggleston: TransactionalStrategy would cover everything I guess?

aweisberg: This PR is out of date now that I have a new rebased branch elsewhere. Just FYI.

aweisberg:
Currently rebasing onto TCM and other things.

PR #2241, opened 2023-03-23
CASSANDRA-18355 Avoid serializing full TxnData instances to Accord state tables

maedhroz:
patch by Caleb Rackliffe; reviewed by ? for CASSANDRA-18355


maedhroz:
https://app.circleci.com/pipelines/github/maedhroz/cassandra?branch=CASSANDRA-18355

.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "modules/accord"]
 	path = modules/accord
-	url = https://github.com/apache/cassandra-accord.git
-	branch = trunk
+	url = ../cassandra-accord.git
+	branch = CASSANDRA-18355
```
maedhroz, line 4: This obv. reverts before committing.

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -652,8 +646,9 @@ public static Command loadCommand(AccordCommandStore commandStore, TxnId txnId)
                     return Command.SerializerSupport.committed(attributes, status, executeAt, promised, accepted, waitingOnCommit, waitingOnApply);
                 case PreApplied:
                 case Applied:
+                    return Command.SerializerSupport.executed(attributes, status, executeAt, promised, accepted, waitingOnCommit, waitingOnApply, writes, Result.APPLIED);
```
maedhroz, line 918: Nobody does anything w/ this right now anyway, but I don't know if we need to distinguish between pre-applied and applied...

dcapwell: `org.apache.cassandra.service.accord.AccordObjectSizes#command` needs to be updated right?

dcapwell: think serializing `accord.messages.CheckStatus.CheckStatusOkFull#result` also needs to be updated?  Which should be `org.apache.cassandra.io.IVersionedSerializer#serialize`

src/java/org/apache/cassandra/service/accord/serializers/ApplySerializers.java
```diff
@@ -52,7 +50,7 @@ public Apply deserializeBody(DataInputPlus in, int version, TxnId txnId, Partial
                                                      CommandSerializers.timestamp.deserialize(in, version),
                                                      DepsSerializer.partialDeps.deserialize(in, version),
                                                      CommandSerializers.writes.deserialize(in, version),
-                                                     TxnData.serializer.deserialize(in, version));
+                                                     null);
```
maedhroz: Wondering if we should just remove the `Result` argument from `create()` entirely...

dcapwell: so `Apply` only touches this by passing it to `accord.local.SafeCommand#preapplied`, which calls `accord.local.Command.Executed#executed(accord.local.CommonAttributes, accord.local.SaveStatus, accord.primitives.Timestamp, accord.primitives.Ballot, accord.primitives.Ballot, accord.local.Command.WaitingOn, accord.primitives.Writes, accord.api.Result)`... Given the fact that this case maps to `Result.APPLIED` later on I think its fine to drop, internally we can still pass `Result.APPLIED` in

@belliottsmith thoughts?

maedhroz: After making that change, I'm getting another simulator failure: https://app.circleci.com/pipelines/github/maedhroz/cassandra/709/workflows/221592d1-1ba2-4df3-a22a-c51ae76edd08/jobs/7699/tests#failed-test-0

Will see if I can reproduce locally here...

dcapwell:
```
Caused by: java.lang.IllegalArgumentException: Unexpected state FINISHED
	at accord.utils.Invariants.illegalArgument(Invariants.java:54)
	at accord.utils.Invariants.checkArgument(Invariants.java:202)
	at org.apache.cassandra.service.accord.async.AsyncOperation.fail(AsyncOperation.java:182)
	at org.apache.cassandra.service.accord.async.AsyncOperation.run(AsyncOperation.java:273)
	at org.apache.cassandra.service.accord.async.AsyncOperation.callback(AsyncOperation.java:157)
	at accord.utils.async.AsyncCallbacks.lambda$inExecutor$0(AsyncCallbacks.java:31)
```

that's a known issue

dcapwell: [CASSANDRA-18377](https://issues.apache.org/jira/browse/CASSANDRA-18377): CEP-15 (Accord) AsyncOperation can not fail as it has already reached FINISHED

maedhroz:
Note: Locally, `AsyncLoaderTest` and `AsyncOperationTest` seem to pass for me.

maedhroz:
Committed as https://github.com/apache/cassandra/commit/56e218638db6816f9a6fa66fa2b6fc35977a92a2

PR #2245, opened 2023-03-24
CASSANDRA-18364: CEP-15: (C*) Accord message processing should avoid being passed on to a Stage and run directly in the messageing handler

dcapwell:


src/java/org/apache/cassandra/concurrent/Stage.java
```diff
@@ -46,7 +46,7 @@
     MUTATION          (true,  "MutationStage",         "request",  DatabaseDescriptor::getConcurrentWriters,        DatabaseDescriptor::setConcurrentWriters,        Stage::multiThreadedLowSignalStage),
     COUNTER_MUTATION  (true,  "CounterMutationStage",  "request",  DatabaseDescriptor::getConcurrentCounterWriters, DatabaseDescriptor::setConcurrentCounterWriters, Stage::multiThreadedLowSignalStage),
     VIEW_MUTATION     (true,  "ViewMutationStage",     "request",  DatabaseDescriptor::getConcurrentViewWriters,    DatabaseDescriptor::setConcurrentViewWriters,    Stage::multiThreadedLowSignalStage),
-    ACCORD            (true,  "AccordStage",           "request",  DatabaseDescriptor::getConcurrentAccordOps,      DatabaseDescriptor::setConcurrentAccordOps,      Stage::multiThreadedLowSignalStage),
+    ACCORD            (true,  "AccordStage",           "request",  () -> 0,                                         null,                                            Stage::immediateExecutor),
```
belliottsmith: Isn't it confusing to call this the `AccordStage`? Shouldn't we just pass the IMMEDIATE stage?

src/java/org/apache/cassandra/service/accord/AccordCallback.java
```diff
@@ -32,18 +34,22 @@
...
     }
 
     @Override
     public void onResponse(Message<T> msg)
     {
         logger.debug("Received response {} from {}", msg.payload, msg.from());
-        callback.onSuccess(EndpointMapping.endpointToId(msg.from()), msg.payload);
+        commandStore.execute(PreLoadContext.empty(), ignore ->
+                                                     callback.onSuccess(EndpointMapping.endpointToId(msg.from()), msg.payload))
+        .begin(commandStore.agent());
```
belliottsmith: should this align with .execute()? If not, let's bring down the closing brace, so it's clear we're continuing the prior line.

belliottsmith (commented):
Looks broadly what we should be doing, but will hang fire until patch is complete before approving.

src/java/org/apache/cassandra/concurrent/Stage.java
dcapwell: pushed the following change; removed `ACCORD` and all usages now uses `IMMEDIATE`

test/simulator/main/org/apache/cassandra/simulator/ClusterSimulation.java
```diff
@@ -544,7 +543,7 @@ IInstanceConfig update(IInstanceConfig config)
                    .set("concurrent_counter_writes", take(1, 4))
                    .set("concurrent_materialized_view_writes", take(1, 4))
                    .set("concurrent_reads", take(1, 4))
-                   .forceSet("available_processors", take(3, 4));
+                   .set("available_processors", take(3, 4));
```
dcapwell, line 546: `forceSet` is the same as `set`... I don't remember why we added it... think this was before we refactored Yaml to support map

test/unit/org/apache/cassandra/config/YamlConfigurationLoaderTest.java
```diff
@@ -359,6 +362,36 @@ public void converters()
         assertThat(from("sstable_preemptive_open_interval_in_mb", -2).sstable_preemptive_open_interval).isNull();
     }
 
+    @Test
+    public void process()
```
dcapwell, line 366: I added this test as I was having issues with simulator and thought could be yaml related... kept as its always best to test our configs 😄

test/simulator/main/org/apache/cassandra/simulator/ClusterSimulation.java
belliottsmith: We previously verified that the property existed in the yaml I think? So if a property wasn't available in earlier versions it could be an issue, or something like that.

belliottsmith (approved):

test/simulator/main/org/apache/cassandra/simulator/ClusterSimulation.java
dcapwell: don't remember, it is identical logic to `set` now adays... we do not validate, we rely on `Config` and `YamlConfigLoader` to do that

src/java/org/apache/cassandra/service/accord/AccordCallback.java
```diff
@@ -32,18 +34,18 @@
 class AccordCallback<T extends Reply> implements RequestCallback<T>
 {
     private static final Logger logger = LoggerFactory.getLogger(AccordCallback.class);
-    private final Callback<T> callback;
+    private final SafeCallback<T> callback;
```
aweisberg: Extends safe callback instead of allocating an another object?

test/simulator/main/org/apache/cassandra/simulator/paxos/AbstractPairOfSequencesPaxosSimulation.java
```diff
@@ -209,6 +209,8 @@ public ActionPlan plan()
             public Action get()
             {
                 int[] primaryKeyIndex = consume(simulated.random, available);
+                if (primaryKeyIndex == null)
```
dcapwell, line 212: this should fix the issue where "consume" is empty, and by returning `Actions.empty` we don't halt the simulation

src/java/org/apache/cassandra/config/PositiveInt.java
```diff
@@ -0,0 +1,88 @@
...
+        return value == that.value;
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(value);
+    }
+
+    public static class UndefinedPositiveInt extends PositiveInt
```
aweisberg: Why have a base class `PositiveInt` and a separate derived `UndefinedPositiveInt`?

It doesn't encapsulate and hide the base class `value` (like say `Optional`) so maybe collapse it to just one class?

test/simulator/main/org/apache/cassandra/simulator/paxos/PaxosSimulation.java
```diff
@@ -286,19 +288,39 @@ RuntimeException failWith(Throwable t)
 
     private RuntimeException logAndThrow()
```
aweisberg, line 289: This doesn't return a RuntimeException, but I know it can also be handy for this kind of method to return something to throw so in the calling method the compiler knows execution stops there.

test/simulator/main/org/apache/cassandra/simulator/systems/SimulatedAction.java
```diff
@@ -382,8 +382,11 @@ List<Action> applyToMessage(IInvokableInstance from, IInvokableInstance to, IMes
                     notify = from;
                 }
                 boolean isTimeout = deliver != FAILURE;
+                Executor notifierExecutor = notify.executorFor(verb.id);
+                if (notifierExecutor instanceof ImmediateExecutor)
```
aweisberg, line 386: Why is this unwrapping necessary? Worth a comment?

aweisberg (changes requested):
Nothing major mostly just some questions

src/java/org/apache/cassandra/service/accord/AccordCallback.java
```diff
@@ -32,18 +34,18 @@
...
 
-    public AccordCallback(Callback<T> callback)
+    public AccordCallback(CommandStore commandStore, Callback<T> callback)
     {
-        this.callback = callback;
+        this.callback = new SafeCallback<>(commandStore, callback);
     }
 
     @Override
     public void onResponse(Message<T> msg)
```
aweisberg, line 44: Really unrelated to your changes, but I just noted that AccordCallback never invokes `slowResponse` or `timeout`

test/simulator/main/org/apache/cassandra/simulator/systems/SimulatedAction.java
dcapwell: its the same as the above logic.  If any stage is `ImmediateExecutor` then the executor isn't changed out during simulation... so the logic that requires a cast will cause a ClassCastException

src/java/org/apache/cassandra/config/PositiveInt.java
dcapwell: the idea is that `PositiveInt` makes sure the values are `>= 1` and for those that require `-1` (atm both users) they use this `UndefinedPositiveInt` that acts as the `null` case.

I tried `Optional<PositiveInt>` but this became `Optional<Integer>` at runtime (and would ClassCast if accessed); Jackson is smart enough in this case, but sadly SnakeYaml isn't (FYI @ekaterinadimitrova2, more points for Jackson)...

So we need a `null` case where `-1` means null... 

> Why have a base class PositiveInt and a separate derived UndefinedPositiveInt?

I think we need 2, but for this patch we only need `UndefinedPositiveInt`, so I can remove the `extends` and have it its own class...

src/java/org/apache/cassandra/service/accord/AccordCallback.java
dcapwell: `timeout` will be done, but will use the `onFailure` method, which sets `failure`....  `timeout` method is syntax sugar for some accord code to mark failed

```
public void timeout(Node.Id to)
    {
        failure(to, new Timeout(null, null));
    }
```

In C* we do

```
private static Throwable convertReason(RequestFailureReason reason)
    {
        return reason == RequestFailureReason.TIMEOUT ?
               new Timeout(null, null) :
               new RuntimeException(reason.toString());
    }
```

I do not know how to do `slowResponse`, but `timeout` is covered

test/unit/org/apache/cassandra/config/DatabaseDescriptorRefTest.java
```diff
@@ -137,8 +137,7 @@
     "org.apache.cassandra.config.GuardrailsOptions$ConsistencyLevels",
     "org.apache.cassandra.config.GuardrailsOptions$TableProperties",
     "org.apache.cassandra.config.ParameterizedClass",
-    "org.apache.cassandra.config.PositiveInt",
-    "org.apache.cassandra.config.PositiveInt$UndefinedPositiveInt",
+    "org.apache.cassandra.config.UndefinedPositiveInt",
```
aweisberg: This doesn't match the class name

dcapwell: refactor fail... I hate this test!

aweisberg (approved):
+1 TY

PR #2246, opened 2023-03-24
CASSANDRA-18352: Bound user provided timestamps via guardrail

jrwest:


src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -421,6 +424,25 @@ public final class Guardrails implements GuardrailsMBean
                      format("The keyspace %s has a replication factor of %s, above the %s threshold of %s.",
                             what, value, isWarning ? "warning" : "failure", threshold));
 
+    public static final MaxThreshold maximumAllowableTimestamp =
+    new MaxThreshold("timestamp_upper_bound",
```
adelapena: It seems this validation is still producing logging messages of the form `timestamp_upper_bound violated:`. Probably we want to use the same name that is used for the yaml property:
```suggestion
    new MaxThreshold("maximum_timestamp",
```
And same for the minimum timestamp.

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumTimestampTest.java
```diff
@@ -0,0 +1,84 @@
...
+    {
+        setGuardrail(TimeUnit.DAYS.toSeconds(1),
+                     TimeUnit.DAYS.toSeconds(2));
+        assertWarns("INSERT INTO %s (k, v) VALUES (1, 1) USING TIMESTAMP " +  (ClientState.getTimestamp() + (TimeUnit.DAYS.toMicros(1) + 4000)),
+                    "timestamp_upper_bound violated");
+    }
+
+
+
+
```
adelapena: Nit: unneeded blank lines

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinimumTimestampTest.java
```diff
@@ -0,0 +1,82 @@
...
+    @Test
+    public void testEnabledWarn() throws Throwable
+    {
+        setGuardrail(TimeUnit.DAYS.toSeconds(1),
+                     TimeUnit.DAYS.toSeconds(2));
+        assertWarns("INSERT INTO %s (k, v) VALUES (1, 1) USING TIMESTAMP " +  (ClientState.getTimestamp() - (TimeUnit.DAYS.toMicros(1) + 4000)),
+                    "timestamp_lower_bound violated");
+    }
+
+
```
adelapena: Nit: unneeded blank lines

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -760,6 +762,70 @@ public void setZeroTTLOnTWCSEnabled(boolean value)
...
 
+    @Override
+    public long getMaximumTimestampWarnThreshold(long currentTimestamp)
+    {
+        long configVal = config.maximum_timestamp_warn_threshold.toMicroseconds();
+        return configVal == 0 ? Long.MAX_VALUE : currentTimestamp + configVal;
+    }
+
+    @Override
+    public void setMaximumTimestampWarnThreshold(DurationSpec.LongMicrosecondsBound duration)
```
adelapena: All the new setters don't seem to ensure that the warn threshold is lower than the fail threshold, as we do with the other guardrails (see `validateWarnLowerThanFail`).

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -760,6 +762,70 @@ public void setZeroTTLOnTWCSEnabled(boolean value)
...
+    public long getMinimumTimestampFailThreshold(long currentTimestamp)
+    {
+        long configVal = config.minimum_timestamp_fail_threshold.toMicroseconds();
+        return configVal == 0 ? Long.MIN_VALUE : currentTimestamp - configVal;
+    }
+
+    @Override
+    public void setMinimumTimestampFailThreshold(DurationSpec.LongMicrosecondsBound duration)
+    {
+        updatePropertyWithLogging("timestamp_fail_lower_bound",
```
adelapena: This probably should be `minimum_timestamp_fail_threshold`, so we print in the logs the name of the yaml property.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -760,6 +762,70 @@ public void setZeroTTLOnTWCSEnabled(boolean value)
...
+    {
+        long configVal = config.minimum_timestamp_warn_threshold.toMicroseconds();
+        return configVal == 0 ? Long.MIN_VALUE : currentTimestamp - configVal;
+    }
+
+    @Override
+    public void setMinimumTimestampWarnThreshold(DurationSpec.LongMicrosecondsBound duration)
+    {
+        updatePropertyWithLogging("timestamp_warn_lower_bound",
+                                  duration,
```
adelapena: This probably should be `minimum_timestamp_warn_threshold`, so we print in the logs the name of the yaml property.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -835,6 +913,18 @@ private static void validateMaxRFThreshold(int warn, int fail)
...
+
+    public static void validateTimestampThreshold(DurationSpec.LongMicrosecondsBound warn,
+                                                  DurationSpec.LongMicrosecondsBound fail,
+                                                  String name)
+    {
+        // this function is used for both upper and lower thresholds because lower threshold is relative
+        // despite using MinThreshold we still want the warn threshold to be less than or equal to
+        // the fail threshold.
+        validateMaxLongThreshold(warn.toMicroseconds(), fail.toMicroseconds(),
+                                 "timestamp_" + name + "_bound", true);
```
adelapena: This is still producing the old upper/lower_bound names, instead of the yaml property names.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
jrwest: oversight will fix

test/unit/org/apache/cassandra/db/guardrails/GuardrailMaximumTimestampTest.java
jrwest: will remove

test/unit/org/apache/cassandra/db/guardrails/GuardrailMinimumTimestampTest.java
jrwest: will remove

src/java/org/apache/cassandra/config/GuardrailsOptions.java
jrwest: will fix

src/java/org/apache/cassandra/config/GuardrailsOptions.java
jrwest: will fix

src/java/org/apache/cassandra/config/GuardrailsOptions.java
jrwest: will fix

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -760,6 +762,82 @@ public void setZeroTTLOnTWCSEnabled(boolean value)
...
+        long configVal = config.maximum_timestamp_fail_threshold.toMicroseconds();
+        return configVal == 0 ? Long.MAX_VALUE : currentTimestamp + configVal;
+    }
+
+    @Override
+    public void setMaximumTimestampFailThreshold(DurationSpec.LongMicrosecondsBound duration)
+    {
+        if (duration.toMicroseconds() != 0)
+            validateTimestampThreshold(config.maximum_timestamp_warn_threshold,
+                                       config.maximum_timestamp_fail_threshold, "maximum_timestamp");
```
adelapena: This isn't validating the new duration argument, but the previous values. So the following test will wrongly pass:
```java
guardrails().setMaximumTimestampFailThreshold(1);
guardrails().setMaximumTimestampWarnThreshold(2);
```
Same for the other three setters. Also, there isn't any test verifying the behaviour of this validation.

conf/cassandra.yaml
```diff
@@ -1757,6 +1757,12 @@ drop_compact_storage_enabled: false
 # Guardrail to allow/disallow user-provided timestamps. Defaults to true.
 # user_timestamps_enabled: true
 #
+# Guardrail to bound user-provided timestamps within a given range. Default is infinite (denoted by 0)
```
adelapena: I think the convention for disables duration, sizes, etc. is that the disabled value is `null` instead of zero. 

Note also that, even if we were considering zero as the disabled value, the following config will wrongly fail validation:
```java
guardrails().setMaximumTimestampFailThreshold(0);
guardrails().setMaximumTimestampWarnThreshold(1);
```
That would fail because 0 is not consistently considered as the disabled value.

I think we should use `null` as the disabled value for the `DurationSpec` properties. The `int` values in seconds that we use on the JMX methods can instead use `-1` as their disabled value. More or less [this way](https://github.com/adelapena/cassandra/commit/9b41030a609e4b1ef0ec23a04f339ea4dc021fce).

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -651,4 +651,56 @@
...
+     * @return The highest acceptable timestamp before triggering a warning
+     */
+    long getMaximumTimestampWarnThreshold(long currentTimestamp);
+
+    /**
+     * Sets the warning upper bound for user supplied timestamps
+     *
+     * @param durationSeconds currentTimestamp + TimeUnit.SECONDS.toMicros(duration)
+     */
+    void setMaximumTimestampWarnThreshold(int durationSeconds);
```
adelapena: During the initial review for guardrails we decided to use a common setter method for both the warn and fail threshold. IIRC the reason was that the requisite of keeping the warn threshold lesser than the fail threshold imposes a certain order on how the separate setter would need to be called. For example, consider the following:
```java
guardrails().setMinimumTimestampWarnThreshold(1);
guardrails().setMinimumTimestampFailThreshold(2);
```
That might fail or not fail depending on the previous value of the fail threshold. However, the combined setter will always succeed:
```java
guardrails().setMinimumTimestampThreshold(1, 2);
```
I think we should keep the current style of combined setters in `GuardrailsMBean` for the sake of consistency.

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -651,4 +651,56 @@
...
+
+    /**
+     * @param currentTimestamp the current timestamp
+     * @return The highest acceptable timestamp before triggering a failure
+     */
+    long getMaximumTimestampFailThreshold(long currentTimestamp);
```
adelapena: Probably it would make sense to have JMX getters returning the value of the property as it appears in config, instead of the ones relative to the current timestamp. Those getters would be the counterpart of the setters, which seem to accept the yaml property as argument.

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -651,4 +651,56 @@
...
+
+    /**
+     * Sets the failure upper bound for user supplied timestamps
+     *
+     * @param durationSeconds currentTimestamp + TimeUnit.SECONDS.toMicros(duration)
```
adelapena: I think the setters accept the config duration alone, as defined in the yaml, without adding the current timestamp.

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
adelapena: Perhaps we can try to homogenize the JMX methods and config so they only deal with the config properties as they are expressed in yaml, and encapsulate in `Guardrails.java` the part about making those thresholds relative to the current time. More or less [this way](https://github.com/adelapena/cassandra/commit/514be38817e7e21e3da9244fcac338c1b512fbee). wdyt?

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
jrwest: Will change to use a single setter

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
jrwest: I think the challenge here is it will fail `JMXStandardsTest`

jrwest: Ah I think I see in your patch you may have addressed that. Will take a closer look.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -421,6 +424,24 @@ public final class Guardrails implements GuardrailsMBean
                      format("The keyspace %s has a replication factor of %s, above the %s threshold of %s.",
                             what, value, isWarning ? "warning" : "failure", threshold));
 
+    public static final MaxThreshold maximumAllowableTimestamp =
+    new MaxThreshold("maximum_timestamp",
+                     null,
```
adelapena: The optional "reason" argument was added quite recently, so most guardrails don't use it. In fact, the only guardrails that use it are `allowFilteringEnabled` and `zeroTTLOnTWCSEnabled`. That argument is meant to provide the reason for having such a guardrail, so it can be included in the warn/fail messages. 

Maybe we could add here something like `Timestamps too far in the future are not realistic`, or something like that, if that makes sense. wdyt?

pmcfadin:
Closing as [CASSANDRA-18352](https://issues.apache.org/jira/browse/CASSANDRA-18352) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #2251, opened 2023-03-28
CASSANDRA-18373 Node Draining Aborts All Current SSTables Imports

5:
Node draining aborts all current sstables imports

SS tables imports that will end up being ignored due to the node draining fail instead of succeeding.
Each active SS tables import periodically checks for the node status and sees whether or not it is `DRAINING`.
In case the node starts draining the import aborts immediately by throwing an `InterruptedException`.

patch by Yuriy Semchyshyn; reviewed by Yifan Cai for CASSANDRA-18373

### DIFF
--- a/src/java/org/apache/cassandra/db/SSTableImporter.java
+++ b/src/java/org/apache/cassandra/db/SSTableImporter.java
@@ -50,6 +50,7 @@ public class SSTableImporter
     private static final Logger logger = LoggerFactory.getLogger(ColumnFamilyStore.class);
 
     private final ColumnFamilyStore cfs;
+    private volatile boolean abort = false;
 
     public SSTableImporter(ColumnFamilyStore cfs)
     {
@@ -130,6 +131,8 @@ public class SSTableImporter
             {
                 try
                 {
+                    if (abort)
+                        throw new InterruptedException("SSTables import has been aborted");
                     Descriptor oldDescriptor = entry.getKey();
                     if (currentDescriptors.contains(oldDescriptor))
                         continue;
@@ -193,6 +196,15 @@ public class SSTableImporter
 
         logger.info("[{}] Done loading load new SSTables for {}/{}", importID, cfs.keyspace.getName(), cfs.getTableName());
         return failedDirectories;
+    }
+
+    /**
+     * Signals this {@link SSTableImporter} to abort its current activity as soon as possible, and
+     * then reject all future requests ({@link #signalAbort()} is called when the node is draining).
+     */
+    void signalAbort()
+    {
+        abort = true;
     }
 
     private void logLeveling(UUID importID, Set<SSTableReader> newSSTables)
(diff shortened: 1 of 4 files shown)


src/java/org/apache/cassandra/db/SSTableImporter.java
```diff
@@ -130,6 +131,8 @@ synchronized List<String> importNewSSTables(Options options)
             {
                 try
                 {
+                    if (abort)
+                        throw new InterruptedException("SSTables import has been aborted");
```
yifan-c: If the only reason for aborting is because of draining. Would using `operationMode` simplify the implementation? The changes in CFS and StorageService can be dropped. 

```
if (StorageService.instance.getOperationMode().equalsIgnoreCase(StorageService.Mode.DRAINING.toString()))
``` 

Another question is that is it the only place to abort the import? The sstables could be opening already before invoking `drain`. Those sstables won't be aborted. 
It is not possible/clean to add the check everywhere. It might make sense to add one extra check before line#183 to avoid loading and building SI of the newly opened sstables.

test/unit/org/apache/cassandra/service/StorageServiceDrainTest.java
```diff
@@ -0,0 +1,96 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.service;
+
+import java.nio.ByteBuffer;
+import java.util.Collections;
+import java.util.List;
```
yifan-c: unused import

src/java/org/apache/cassandra/db/SSTableImporter.java
5: Done.

test/unit/org/apache/cassandra/service/StorageServiceDrainTest.java
5: Removed.

src/java/org/apache/cassandra/db/SSTableImporter.java
```diff
@@ -182,19 +185,35 @@ synchronized List<String> importNewSSTables(Options options)
...
             for (SSTableReader reader : newSSTables)
             {
                 if (options.invalidateCaches && cfs.isRowCacheEnabled())
                     invalidateCachesForSSTable(reader);
             }
-
+        }
+        catch (Throwable t)
+        {
+            logger.error("[{}] FaiFailed adding sstables", importID, t);
```
dineshjoshi: Typo? `FaiFailed`?

src/java/org/apache/cassandra/db/SSTableImporter.java
```diff
@@ -182,19 +185,35 @@ synchronized List<String> importNewSSTables(Options options)
...
+            throw new RuntimeException("Failed adding sstables", t);
```
dineshjoshi: Nit: `sstables` -> `SSTables`

dineshjoshi (changes requested):
LGTM. Minor nits.

src/java/org/apache/cassandra/db/SSTableImporter.java
5: Fixed.

src/java/org/apache/cassandra/db/SSTableImporter.java
5: Fixed here and in six other places within this method.

PR #2253, opened 2023-03-29
CASSANDRA-18124 Make keystore_password nullable

maulin-vasavada:
The JIRA Ticket: [CASSANDRA-18124 JIRA](https://issues.apache.org/jira/browse/CASSANDRA-18124)

```
CASSANDRA-18124 Make keystore_password nullable.

Changes made:
1. Fixed the compilation issue for examples/ssl-factory module on trunk
2. Made keystore_password/outbound_keystore_password and truststore_password configurations nullable
3. Added keystore_password/outbound_keystore_password and truststore_password required for File based sslcontextfactories (except PEM)
4. Removed the warning upon keystore_password and key_password both are provided but match the values
5. Added a reference to the documentation for configuring the PEM based key material in the cassandra.yaml

```

```

```

### DIFF
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -150,7 +150,7 @@ Upgrading
      to TimeWindowCompactionStrategy before upgrading to this version.
    - The deprecated functions `dateOf` and `unixTimestampOf` have been removed. They were deprecated and replaced by
      `toTimestamp` and `toUnixTimestamp` in Cassandra 2.2.
-   - Keystore password configuration is nullable now and the code defaults of the keystore passwords to 'cassandra' are
+   - Keystore/truststore password configurations are nullable now and the code defaults of those passwords to 'cassandra' are
      removed. Any deployments that depend upon the code default to this password value without explicitly specifying
      it in cassandra.yaml will fail on upgrade. Please specify your keystore_password and truststore_password elements in cassandra.yaml with appropriate
      values to prevent this failure.
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -152,7 +152,7 @@ Upgrading
      `toTimestamp` and `toUnixTimestamp` in Cassandra 2.2.
    - Keystore password configuration is nullable now and the code defaults of the keystore passwords to 'cassandra' are
      removed. Any deployments that depend upon the code default to this password value without explicitly specifying
-     it in the cassandra.yaml would fail. Please specify your keystore passwords in the cassandra.yaml with appropriate
+     it in cassandra.yaml will fail on upgrade. Please specify your keystore_password and truststore_password elements in cassandra.yaml with appropriate
      configuration to prevent such failure.
 
 Deprecation
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -153,7 +153,7 @@ Upgrading
    - Keystore password configuration is nullable now and the code defaults of the keystore passwords to 'cassandra' are
      removed. Any deployments that depend upon the code default to this password value without explicitly specifying
      it in cassandra.yaml will fail on upgrade. Please specify your keystore_password and truststore_password elements in cassandra.yaml with appropriate
-     configuration to prevent such failure.
+     values to prevent this failure.
 
 Deprecation
 -----------
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -151,7 +151,7 @@ Upgrading
    - The deprecated functions `dateOf` and `unixTimestampOf` have been removed. They were deprecated and replaced by
      `toTimestamp` and `toUnixTimestamp` in Cassandra 2.2.
    - Keystore password configuration is nullable now and the code defaults of the keystore passwords to 'cassandra' are
-     removed now. The deployments that depend upon the code default to this password value without explicitly specifying
+     removed. Any deployments that depend upon the code default to this password value without explicitly specifying
      it in the cassandra.yaml would fail. Please specify your keystore passwords in the cassandra.yaml with appropriate
      configuration to prevent such failure.
 
--- a/NEWS.txt
+++ b/NEWS.txt
@@ -150,6 +150,10 @@ Upgrading
      to TimeWindowCompactionStrategy before upgrading to this version.
    - The deprecated functions `dateOf` and `unixTimestampOf` have been removed. They were deprecated and replaced by
      `toTimestamp` and `toUnixTimestamp` in Cassandra 2.2.
+   - Keystore password configuration is nullable now and the code defaults of the keystore passwords to 'cassandra' are
+     removed now. The deployments that depend upon the code default to this password value without explicitly specifying
+     it in the cassandra.yaml would fail. Please specify your keystore passwords in the cassandra.yaml with appropriate
+     configuration to prevent such failure.
 
 Deprecation
 -----------
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -1360,6 +1360,7 @@ server_encryption_options:
   # during upgrade to 4.0; otherwise, set to false.
   legacy_ssl_storage_port_enabled: false
   # Set to a valid keystore if internode_encryption is dc, rack or all
+  # For configurating PEM based key material refer, to https://cassandra.apache.org/doc/latest/cassandra/operating/security.html#using-pem-based-key-material
   keystore: conf/.keystore
   keystore_password: cassandra
   # During internode mTLS authentication, inbound connections (acting as servers) use keystore, keystore_password
@@ -1405,6 +1406,7 @@ client_encryption_options:
   # optional defaults to true when enabled is false, and false when enabled is true.
   # optional: true
   # Set keystore and keystore_password to valid keystores if enabled is true
+  # For configurating PEM based key material, refer to https://cassandra.apache.org/doc/latest/cassandra/operating/security.html#using-pem-based-key-material
   keystore: conf/.keystore
   keystore_password: cassandra
   # Verify client certificates
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -1360,7 +1360,7 @@ server_encryption_options:
   # during upgrade to 4.0; otherwise, set to false.
   legacy_ssl_storage_port_enabled: false
   # Set to a valid keystore if internode_encryption is dc, rack or all
-  # For configurating PEM based key material refer, to https://cassandra.apache.org/doc/latest/cassandra/operating/security.html#using-pem-based-key-material
+  # For configurating PEM based key material, refer to https://cassandra.apache.org/doc/latest/cassandra/operating/security.html#using-pem-based-key-material
   keystore: conf/.keystore
   keystore_password: cassandra
   # During internode mTLS authentication, inbound connections (acting as servers) use keystore, keystore_password
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -1360,9 +1360,13 @@ server_encryption_options:
   # during upgrade to 4.0; otherwise, set to false.
   legacy_ssl_storage_port_enabled: false
   # Set to a valid keystore if internode_encryption is dc, rack or all
-  # For configurating PEM based key material, refer to https://cassandra.apache.org/doc/latest/cassandra/operating/security.html#using-pem-based-key-material
   keystore: conf/.keystore
   keystore_password: cassandra
+  # Configure the way Cassandra creates SSL contexts.
+  # To use PEM-based key material, see org.apache.cassandra.security.PEMBasedSslContextFactory
+  # ssl_context_factory:
+  #     # Must be an instance of org.apache.cassandra.security.ISslContextFactory
+  #     class_name: org.apache.cassandra.security.DefaultSslContextFactory
   # During internode mTLS authentication, inbound connections (acting as servers) use keystore, keystore_password
   # containing server certificate to create SSLContext and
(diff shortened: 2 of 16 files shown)


src/java/org/apache/cassandra/security/FileBasedSslContextFactory.java
```diff
@@ -20,6 +20,7 @@
 
 import java.io.InputStream;
 import java.nio.file.Files;
+import java.nio.file.Paths;
```
smiklosovic: @maulin-vasavada you are using illegal imports, you should use `org.apache.cassandra.io.util.File.getPath` instead

````
checkstyle:
    [mkdir] Created dir: /home/fermat/dev/cassandra/cassandra-instaclustr/cassandra/build/checkstyle
[checkstyle] Running Checkstyle 8.45.1 on 2201 files
[checkstyle] [ERROR] /home/fermat/dev/cassandra/cassandra-instaclustr/cassandra/src/java/org/apache/cassandra/security/FileBasedSslContextFactory.java:23:1: Illegal import - java.nio.file.Paths. [IllegalImport]
````

Comment (maulin-vasavada) [reply src/java/org/apache/cassandra/security/FileBasedSslContextFactory.java]: Sure, made the change. Thanks. Will figure out why I am unable to see the same checkstyle error locally.

File: conf/cassandra.yaml [hunk 2]
```diff
@@ -1360,6 +1360,7 @@ server_encryption_options:
   # during upgrade to 4.0; otherwise, set to false.
   legacy_ssl_storage_port_enabled: false
   # Set to a valid keystore if internode_encryption is dc, rack or all
+  # For configurating PEM based key material, refer to https://cassandra.apache.org/doc/latest/cassandra/operating/security.html#using-pem-based-key-material
```
smiklosovic: I am not completely sure about this URL. There is no case of us adding links like this in whole cassandra.yaml.

maulin-vasavada: One of the ask on the ticket was to include an example in comments for the PEM formatted keys. However, since we have client and server both options, adding long repeated comments may not make sense. Hence I added a link like another http link I saw on the yaml document. Would love to hear from you and other members on this and get more direction.

maedhroz: There is something similar for `num_tokens`, but even that is sort of a "best practices" reference, not something that describes the configuration format itself. The problem here is that what is at the URL in question won't necessarily be versioned along w/ the code itself, right?

We already duplicate a small amount of information in the comments for server and client encryption. What about just adding a commented out section before `keystore` to both that looks something like this:

```
# Configure the way Cassandra creates SSL contexts.
# To use PEM-based key material, see org.apache.cassandra.security.PEMBasedSslContextFactory
# ssl_context_factory:
#     # Must be an instance of org.apache.cassandra.security.ISslContextFactory
#     class_name: org.apache.cassandra.security.DefaultSslContextFactory
```

`PEMBasedSslContextFactory` at least has comments that should be updated if necessary.

(As an aside, this is one of the reasons something like CASSANDRA-17292 would be useful. There's no documentation-friendly hierarchy in our configuration format.)

src/java/org/apache/cassandra/config/EncryptionOptions.java
```diff
@@ -147,9 +151,9 @@ public EncryptionOptions()
         ssl_context_factory = new ParameterizedClass("org.apache.cassandra.security.DefaultSslContextFactory",
                                                      new HashMap<>());
         keystore = "conf/.keystore";
-        keystore_password = "cassandra";
+        keystore_password = null;
         truststore = "conf/.truststore";
-        truststore_password = "cassandra";
+        truststore_password = null;
```
maedhroz, line 156: What's the worst thing that could happen as a result of this change? A `cassandra.yaml` out in the wild that uses "cassandra" as a password implicitly by virtue of not specifying it will suddenly break unless the password is explicitly set to the old default?

That's probably fine in a major release, in any case...

maedhroz (approved):
Approved, w/ some minor commentary.

src/java/org/apache/cassandra/config/EncryptionOptions.java
smiklosovic: @maedhroz 

I do not think that having passwords defaulting to some value is a good idea anyway. A password should be of some complex form, not "cassandra". What value is in that, anyway? It would be quite weird to have a commented password field and then having a certificate / keystore / truststore to have "cassandra" password. That is just strange and I dont think that is used in practice a lot ....

conf/cassandra.yaml
smiklosovic: I will leave this to @maulin-vasavada to decide.

src/java/org/apache/cassandra/config/EncryptionOptions.java
rtib: While I understand the concern about changing a default value belongs to a major release, I'm not sure that the original issue should wait for 5.0. Any solution shipped with the next 4.1 release is considered better than postpone to 5.0.

Would it possible to leave the default value `cassandra` in place while enabling it to be set null explicitly? IMO, that wouldn't conflict with SemVer policy and could be release with the next 4.1.

smiklosovic: @maulin-vasavada what do you think about @rtib comment above? I think that would be the most ideal solution which might be shipped in next 4.1 patch release (probably in 4.1.2).

maedhroz: Either way, I would probably add an entry in `NEWS.txt` for 5.0 in the "Upgrading" section to explain the default change.

> I do not think that having passwords defaulting to some value is a good idea anyway...That is just strange and I dont think that is used in practice a lot

100% agree, and I don't know why we ever did it that way in the first place :-|

conf/cassandra.yaml
maulin-vasavada: Thanks @maedhroz for the input. I like that idea. Changed the documentation like suggested. Please review.

src/java/org/apache/cassandra/config/EncryptionOptions.java
maulin-vasavada: If we keep the default value in the EncryptionOptions then it becomes difficult to have tests working for PEM use-cases for the unencrypted key. Hence I would like to keep the null assignment in the code with these changes. If that bumps this upto go in 5.x I am fine with it.

conf/cassandra.yaml
```diff
@@ -1406,9 +1410,13 @@ client_encryption_options:
...
   # optional: true
   # Set keystore and keystore_password to valid keystores if enabled is true
-  # For configurating PEM based key material, refer to https://cassandra.apache.org/doc/latest/cassandra/operating/security.html#using-pem-based-key-material
   keystore: conf/.keystore
   keystore_password: cassandra
+  # Configure the way Cassandra creates SSL contexts.
+  # To use PEM-based key material, see org.apache.cassandra.security.PEMBasedSslContextFactory
+  # ssl_context_factory:
+  #     # Must be an instance of org.apache.cassandra.security.ISslContextFactory
+  #     class_name: org.apache.cassandra.security.DefaultSslContextFactory
```
maedhroz, line 1419: Looks good...not too much duplication in the end.

src/java/org/apache/cassandra/config/EncryptionOptions.java
maedhroz: I'm fine either way as long as `NEWS.txt` mentions it.

maulin-vasavada: @maedhroz Do we need the change in NEWS.txt in this PR or that would be part of the 5.0 release? I can resolve this conversation after this for a committer to merge this.

maedhroz: Yes, even if it's in a major release, a change in behavior should be documented there.

smiklosovic: I am for documenting it and merge it to 4.1 as well as to trunk. I think this patch is for trunk only, currently. I can backport it to 4.1 quite easily, should be nicely backportable.

maulin-vasavada: @maedhroz and @smiklosovic I added a text in the News txt file under `Upgrades` for 5.0 for the removal of the default keystore passwords. Please take a look. After that we are good to merge this.

NEWS.txt
```diff
@@ -150,6 +150,10 @@ Upgrading
      to TimeWindowCompactionStrategy before upgrading to this version.
    - The deprecated functions `dateOf` and `unixTimestampOf` have been removed. They were deprecated and replaced by
      `toTimestamp` and `toUnixTimestamp` in Cassandra 2.2.
+   - Keystore password configuration is nullable now and the code defaults of the keystore passwords to 'cassandra' are
+     removed now. The deployments that depend upon the code default to this password value without explicitly specifying
```
maedhroz:
```suggestion
     removed. Any deployments that depend upon the code default to this password value without explicitly specifying
```

NEWS.txt
```diff
@@ -150,6 +150,10 @@ Upgrading
...
+     it in the cassandra.yaml would fail. Please specify your keystore passwords in the cassandra.yaml with appropriate
```
maedhroz:
```suggestion
     it in cassandra.yaml will fail on upgrade. Please specify your keystore_password and truststore_password elements in cassandra.yaml with appropriate
```

NEWS.txt
```diff
@@ -150,6 +150,10 @@ Upgrading
...
+     configuration to prevent such failure.
```
maedhroz:
```suggestion
     values to prevent this failure.
```

conf/cassandra.yaml
```diff
@@ -1362,6 +1362,11 @@ server_encryption_options:
   # Set to a valid keystore if internode_encryption is dc, rack or all
   keystore: conf/.keystore
   keystore_password: cassandra
```
maedhroz: Looking at this one last time, don't we want to comment out all 4 of the `keystore_password` and `truststore_password` field in the YAML? If we leave them there, and operators just start using the YAML we ship, we will essentially have preserved the "cassandra" default. The point of all this is that we want to force the operator to choose passwords, right?

CC @smiklosovic

maulin-vasavada: I agree with you @maedhroz. Will wait to hear Stefan's comment and would make appropriate changes.

maulin-vasavada: @smiklosovic any comments before we can close this?

smiklosovic: @maedhroz  @maulin-vasavada 

But if we leave it here uncommented, we also do not break existing configurations which are depending on it.

It is about "not breaking what is there". If we change this, all deployments which rely on the default will stop to work because we set it to null suddenly.

It is more about the _possibility_ to set it to null if one really wants that rather than setting it by default to null and break it everywhere it relies on not null (even it is "cassandra" as default).

People who are (and should) using stronger passwords have to set them explicitly anyway and they do not have this problem.

If this goes to 4.1 as well, people might be quite surprised that this will not work between stupid version bump in a patch release. Having it changed in 5.0 is probably OK but still ... 

So, I would leave it uncommented in 4.1 but I would comment it in 5.0.

That way we do not break patch release bump but we still educate users that there is no magic default for them in 5.0 and they should set it all explicitly.

NEWS.txt
```diff
@@ -150,6 +150,10 @@ Upgrading
      to TimeWindowCompactionStrategy before upgrading to this version.
    - The deprecated functions `dateOf` and `unixTimestampOf` have been removed. They were deprecated and replaced by
      `toTimestamp` and `toUnixTimestamp` in Cassandra 2.2.
+   - Keystore password configuration is nullable now and the code defaults of the keystore passwords to 'cassandra' are
```
smiklosovic: @maulin-vasavada we should mention keystore as well as truststore passwords here. It would be also nice to mention what section these configuration parameters are found in.

conf/cassandra.yaml
maedhroz: That works for me.

maulin-vasavada: Cool. So @smiklosovic and @maedhroz I have [a separate PR](https://github.com/apache/cassandra/pull/2273) for the 4.1 backport and in this PR I can comment out the password configuration from the cassandra yaml since this will be target for 5.0 release. Please let me know if that understanding is correct.

maulin-vasavada: Done commenting out the passwords in the cassandra yaml in this PR.

NEWS.txt
maulin-vasavada: done

conf/cassandra.yaml
smiklosovic: thank you, I will take a look shortly. Thanks for the patience.

smiklosovic: @maulin-vasavada looks good, please apply what was discussed here in 4.1 branch and we can finally ship it.

EDIT: Actually, we do not need to do anything for 4.1 in NEWS because we effectively have not changed anything. Upgrades will be working as before.

So I think we are done.

maulin-vasavada: Cool thanks @smiklosovic I've the backported changes for 4.1 [here in a different PR](https://github.com/apache/cassandra/pull/2273). Will remove news.txt update from there. Thanks for helping.

PR #2257, opened 2023-03-31
CASSANDRA-18375: Fix response verb for ACCORD_INFORM_OF_TXNID_REQ

dcapwell:


src/java/org/apache/cassandra/service/accord/async/AsyncOperation.java
```diff
@@ -179,7 +179,8 @@ private void finish(R result)
     private void fail(Throwable throwable)
     {
         Invariants.nonNull(throwable);
-        Invariants.checkArgument(state != State.FINISHED && state != State.FAILED, "Unexpected state %s", state);
+        if (!(state != State.FINISHED && state != State.FAILED))
```
maedhroz:
```suggestion
        if (state == State.FINISHED || state == State.FAILED)
```

dcapwell: I pushed the knowledge of complete to the `State` via a `isComplete()` method, and switched to `if (state.isComplete())`

maedhroz (approved):

PR #2258, opened 2023-04-02
CASSANDRA-18221: Add AccordConfig to track accord specific configs

kamalesh0406:
patch by Kamalesh Palanisamy;

```

```

### DIFF
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -596,6 +596,7 @@ public class Config
     public volatile boolean use_statements_enabled = true;
 
     public boolean accord_transactions_enabled = false;
+    public long accord_scheduler_delay_in_ms = 200L;
 
     /**
      * Optionally disable asynchronous UDF execution.
--- a/src/java/org/apache/cassandra/service/accord/AccordService.java
+++ b/src/java/org/apache/cassandra/service/accord/AccordService.java
@@ -39,6 +39,7 @@ import accord.messages.Request;
 import accord.primitives.Txn;
 import accord.primitives.TxnId;
 import accord.topology.TopologyManager;
+import accord.utils.AccordConfig;
 import accord.utils.DefaultRandom;
 import accord.utils.async.AsyncChains;
 import org.apache.cassandra.concurrent.Shutdownable;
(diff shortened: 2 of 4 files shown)


src/java/org/apache/cassandra/config/Config.java
```diff
@@ -596,6 +596,7 @@ public MemtableOptions()
     public volatile boolean use_statements_enabled = true;
 
     public boolean accord_transactions_enabled = false;
+    public long accord_scheduler_delay_in_ms = 200L;
```
kamalesh0406, line 599: This could also be named `accord_progress_log_scheduler_delay_in_ms` but I felt it to verbose compared to the current one. I am open to suggestions on it.

Claudenw, line 599: Shouldn't this be a DurationSpec rather than a long particularly if you expect very large values?

dcapwell: agree with this comment, `DurationSpec` (or one of the ones that limit the domain) is what we should do; `in_ms` should be avoided

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -137,7 +139,9 @@ private AccordService()
         this.messageSink = new AccordMessageSink();
         this.configService = new AccordConfigurationService(localId);
         this.scheduler = new AccordScheduler();
+        this.config = new AccordConfig(DatabaseDescriptor.getAccordSchedulerDelayInMS());
```
dcapwell, line 142: this pattern isn't something we can maintain; every new config comes with a massive burden to update this constructor; IMO we should push this logic into the YAML layer

dcapwell (changes requested):
I personally feel that we should have the following in `Config`

```
public final AccordConfig accord = new AccordConfig();
```

and that config class should be usable in a POJO manner (public fields count), this allows snakeyaml or any other tool access to support in a well understood manner.

Now, if you do that the configs look like

```
accord:
  scheduler_delay: 200ms
```

this brings up a valid question on *should* it be nested?  *should* we "embed" so all AccordConfigs are top level Config fields?

I know there push back from some when they see nested, so thinking the following would be desirable

```
@Embedded
public final AccordConfig accord = new AccordConfig();
```

If we create this new annotation, and implement it in the config loading layer (I can help with this) we could get the semantic that all fields in `AccordConfig` are top-level and prefixed with `accord_`...

src/java/org/apache/cassandra/config/Config.java
smiklosovic: @kamalesh0406 @dcapwell  is not this "wrong" way of doing it? There are no "...in_ms" properties anymore, everything is based on `DurationSpec` when it comes to time.

PR #2259, opened 2023-04-04
CASSANDRA-18422: C* side

dcapwell:


test/simulator/test/org/apache/cassandra/simulator/paxos/HistoryValidatorTest.java
```diff
@@ -356,79 +383,142 @@ private static Event writeOnly(int pk)
...
+                    {
+                        Read read = (Read) a;
+                        check.read(read.pk, read.id, read.count, read.seq);
+                    }
+                    else
+                    {
+                        Write write = (Write) a;
+                        check.write(write.pk, write.id, write.success);
+                    }
+                }
```
Claudenw: Since `actions` is a List<Operation> would it make sense to add a `validate(HistoryValidator.Checker)` method to `Operaion` and simply call it here passing the `check` variable as an argument?

test/simulator/test/org/apache/cassandra/simulator/paxos/HistoryValidatorTest.java
```diff
@@ -356,79 +383,142 @@ private static Event writeOnly(int pk)
...
+                else if (action instanceof Write)
+                {
+                    pks.add(((Write) action).pk);
+                }
+                else
+                {
+                    throw new IllegalStateException("Unknown type: " + action.getClass());
+                }
             }
+            return pks;
```
Claudenw: Like the previous comment, would it not make sense for Operation to have a `getPK()` method to so that the set of PK can be generated without `instanceof` checks?

test/simulator/test/org/apache/cassandra/simulator/paxos/HistoryValidatorTest.java
```diff
@@ -356,79 +383,142 @@ private static Event writeOnly(int pk)
...
+                else if (a instanceof Write)
+                {
+                    Write write = (Write) a;
+                    sb.append("\twrite(pk=").append(write.pk).append(", id=").append(write.id).append(", success=").append(write.success).append(")\n");
+                }
+                else
+                {
+                    throw new AssertionError("Unexpected type: " + a.getClass());
                 }
             }
```
Claudenw: As above, shouldn't the `Operation` implementations `Read` and `Write` override `toString()`?  Then the method could simply be called here on each action.

test/simulator/test/org/apache/cassandra/simulator/paxos/HistoryValidatorTest.java
dcapwell: thanks, did this and all other places that did `instanceOf` checks

dcapwell:
Thanks for the feedback @Claudenw , pushed changes to address your feedback

test/simulator/test/org/apache/cassandra/simulator/paxos/HistoryValidatorTest.java
```diff
@@ -356,79 +383,146 @@ private static Event writeOnly(int pk)
...
     }
 
-    private void fromLog(String log)
+    private interface Operation
     {
-        IntSet pks = new IntHashSet();
-        class Read
+        int pk();
+        void check(HistoryValidator.Checker check);
+        void toString(StringBuilder sb);
```
Claudenw: This name conflicts semantically with the `Object.toString()`  semantically this is probably better named `addString()`  or `appendString()`

test/simulator/test/org/apache/cassandra/simulator/paxos/HistoryValidatorTest.java
```diff
@@ -356,79 +383,146 @@ private static Event writeOnly(int pk)
...
+    }
+
+    private static class Read implements Operation
+    {
+        final int pk, id, count;
+        final int[] seq;
```
Claudenw, line 396: Is there any reason not to make the variable private?

test/simulator/test/org/apache/cassandra/simulator/paxos/HistoryValidatorTest.java
```diff
@@ -356,79 +383,146 @@ private static Event writeOnly(int pk)
         return new Event(EnumSet.of(Event.Type.WRITE), pk, null);
     }
 
-    private void fromLog(String log)
+    private interface Operation
```
Claudenw, line 386: If `Operation` was an abstract class rather than an interface you could implement pk() directly in the class and not repeat it in the subclasses.  I note that ID is also used in every instance of operation. 

Do you expect other implementations of Operation that do not have the `pk` or `id` specified in the constructor?

Is there any reason not to make the variable private?

test/simulator/test/org/apache/cassandra/simulator/paxos/HistoryValidatorTest.java
```diff
@@ -356,79 +383,146 @@ private static Event writeOnly(int pk)
...
+        public void toString(StringBuilder sb)
+        {
+            sb.append("write(pk=").append(pk).append(", id=").append(id).append(", success=").append(success).append(")\n");
+        }
+    }
+
+    private static class Witness
+    {
+        final int start, end;
+        final List<Operation> actions = new ArrayList<>();
```
Claudenw, line 459: Is there any reason not to make the variables private?

test/simulator/test/org/apache/cassandra/simulator/paxos/HistoryValidatorTest.java
dcapwell: more verbose and have no impact.  Since all access is in the same file java will allow all classes to reach in and touch...

now, when it comes to our style, when its not directly needed we should remove; so in this case since `private` doesn't matter our style says it shouldn't be there.

test/simulator/test/org/apache/cassandra/simulator/paxos/HistoryValidatorTest.java
dcapwell: honestly I prefer interfaces over abstract class, and having an abstract class just too hold `pk` is not something I care too much for.

so this points down to style, and I prefer this style =)

> Is there any reason not to make the variable private?

redundant so not required

test/simulator/test/org/apache/cassandra/simulator/paxos/HistoryValidatorTest.java
dcapwell: the class is private, so its redundant...  if you add `private` you don't block access to anything that can see this class or related classes; so as it only adds verbosity our style says to drop

dcapwell:
@Claudenw pushed feedback or commented addressing why I didn't make changes, let me know if this makes sense to you

aweisberg (approved):
🚢 it

Claudenw (approved):

PR #2261, opened 2023-04-06
CASSANDRA-18430: When decommissioning should set Severity to limit traffic

dcapwell:


src/java/org/apache/cassandra/gms/Gossiper.java
```diff
@@ -2141,6 +2141,11 @@ public void addLocalApplicationState(ApplicationState applicationState, Versione
         addLocalApplicationStates(Arrays.asList(Pair.create(applicationState, value)));
     }
 
+    public void addLocalApplicationState(ApplicationState state1, VersionedValue value1, ApplicationState state2, VersionedValue value2)
```
dcapwell, line 2144: added this method as the locking is done at the call level, so when we have STATUS and STATUS_AND_PORT as 2 different calls we can get into a state where we gossip one changed and the other unchanged; by using this new method both status are updated in the same lock

src/java/org/apache/cassandra/locator/DynamicEndpointSnitch.java
```diff
@@ -224,6 +226,11 @@ private <C extends ReplicaCollection<? extends C>> C sortedByProximityWithBadnes
         return replicas;
     }
 
+    private static double defaultStore(InetAddressAndPort target)
+    {
+        return USE_SEVERITY ? getSeverity(target) : 0.0;
```
dcapwell, line 231: not sure if this change should also be behind a config flag... this is a change in behavior if you set severity... this is done to make this patch work in a stable manner, but outside of decom severity is used to limit traffic, yet we had different code paths to ignore it (such as reset every 10m).  By defaulting to severity we always respect the operators intent

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1617,10 +1618,19 @@ private static void sendMessagesToNonlocalDC(Message<? extends IMutation> messag
             target = targets.get(0);
         }
 
+        Tracing.trace("Sending mutation to remote replica {}", target);
         MessagingService.instance().sendWriteWithCallback(message, target, handler);
         logger.trace("Sending message to {}@{}", message.id(), target);
     }
 
+    private static Replica pickReplica(EndpointsForToken targets)
```
dcapwell, line 1626: I originally created a method so I could use BB to rewrite the result... now that I had stable tests I no longer have that BB logic... so I could move this back...

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1617,10 +1618,19 @@ private static void sendMessagesToNonlocalDC(Message<? extends IMutation> messag
...
+    {
+        EndpointsForToken healthy = targets.filter(r -> DynamicEndpointSnitch.getSeverity(r.endpoint()) == 0);
```
dcapwell, line 1628: one negative with this is what we keep parsing severity... it may make sense for EndpointState to cache...

test/distributed/org/apache/cassandra/distributed/test/topology/DecommissionAvoidTimeouts.java
```diff
@@ -0,0 +1,240 @@
...
+            {
+                ByteBuffer key = Murmur3Partitioner.LongToken.keyForToken(token);
+
+                toDecom.coordinator().execute("INSERT INTO " + table + "(pk) VALUES (?)", ConsistencyLevel.EACH_QUORUM, key);
+            }
+
+            CompletableFuture<Void> nodetool = CompletableFuture.runAsync(() -> toDecom.nodetoolResult("decommission").asserts().success());
+
+            StatusChangeListener.hooks(DECOM_NODE).leave.await();
+            cluster.filters().verbs(Verb.GOSSIP_DIGEST_SYN.id).drop();
```
dcapwell, line 115: this makes sure that the LEFT gossip state never makes it to nodes.  I should also block the gossip shutdown message, that way the only thing coordinator knows is severity (and LEAVING)

src/java/org/apache/cassandra/concurrent/SEPExecutor.java
```diff
@@ -70,7 +70,7 @@ public class SEPExecutor implements LocalAwareExecutorPlus, SEPExecutorMBean
     SEPExecutor(SharedExecutorPool pool, int maximumPoolSize, MaximumPoolSizeListener maximumPoolSizeListener, String jmxPath, String name)
     {
         this.pool = pool;
-        this.name = name;
+        this.name = NamedThreadFactory.globalPrefix() + name;
```
dcapwell, line 73: this only impacts jvm-dtest.  We added this prefix to make the logs easier to read, the shared stages were missing this so was hard to see what `ReadStage` was the coming from which node

src/java/org/apache/cassandra/gms/Gossiper.java
driftx: was this an actual problem? STATUS is deprecated in 4.0 and not supposed to be used in 5.0 (that probably needs its own ticket though)

src/java/org/apache/cassandra/service/StorageProxy.java
driftx: I think that should be fast enough without caching.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5083,8 +5083,11 @@ private List<DecoratedKey> keySamples(Iterable<ColumnFamilyStore> cfses, Range<T
      */
     private void startLeaving()
     {
-        Gossiper.instance.addLocalApplicationState(ApplicationState.STATUS_WITH_PORT, valueFactory.leaving(getLocalTokens()));
-        Gossiper.instance.addLocalApplicationState(ApplicationState.STATUS, valueFactory.leaving(getLocalTokens()));
+        DatabaseDescriptor.getSeverityDuringDecommission()
+                          .ifPresent(severity ->
+                                     Gossiper.instance.addLocalApplicationState(ApplicationState.SEVERITY, valueFactory.severity(severity)));
```
driftx: Why not call DES.setSeverity()?

src/java/org/apache/cassandra/gms/Gossiper.java
dcapwell: > was this an actual problem? 

its 100% unrelated... I tried solving this a different way but moved away from it (add a status_hash app state, so we can know if the cluster is in-sync with a change made locally).  I can force simulate an issue, but I have not heard a real problem in the wild yet.

src/java/org/apache/cassandra/service/StorageProxy.java
dcapwell: cache would require volatile, where as parsing the string to an int is fixed cost/memory... so likely better than the volatile.

src/java/org/apache/cassandra/service/StorageService.java
dcapwell: at first I didn't do it as `org.apache.cassandra.locator.DynamicEndpointSnitch#setSeverity` is a method and not function (tied to the current snitch).  I could always make it static.    To avoid copy/paste ill make it static and delegate to it

dcapwell: pushed a commit fixing this.  see https://github.com/apache/cassandra/pull/2261/commits/f99bd8b1ef940ed2cdafc7b8af91e6ab1173937e

`setSeverity` is part of the `MBean` so I can not touch.  I also can not have a static method with the same name, so I created a new `addSeverity` static function, and have `setSeverity` and this logic call it

src/java/org/apache/cassandra/gms/Gossiper.java
driftx: with STATUS deprecated, I don't think there should be a problem as long as STATUS_AND_PORT is set.

dcapwell: I can remove from this patch

PR #2262, opened 2023-04-06
Update documentation to fix typos & add missing info

lukechatfield:
In virtualtables.adoc:
- Fixed incorrect USE statement examples (incorrect keyspace + typo)
- Fixed order of results for DESCRIBE tables command to match the order in cqlsh
- Updated the descriptions of metrics provided in each virtual table to add missing info / add more info
- Fixed typos / punctuation / formatting

In cqlsh.adoc:
- Fixed typos / punctuation / formatting

In faq/index.adoc
- Fixed typos / punctuation

### DIFF
--- a/doc/modules/cassandra/pages/faq/index.adoc
+++ b/doc/modules/cassandra/pages/faq/index.adoc
@@ -62,7 +62,7 @@ making sure that each node will generate a random token on the next
 restart.
 
 [[change-replication-factor]]
-== Can I change the replication factor (a a keyspace) on a live cluster?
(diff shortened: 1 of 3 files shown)


smiklosovic:
@lukechatfield thanks for the effort! btw could we spend a little bit more time on these virtual tables? There is also doc/modules/pages/new/virtualtables.adoc and I do not have a clue why it is there, should not that be just deleted? We might probably also add more virtual tables like system_views.system_logs (I think it is called like that).

pmcfadin:
No response from OP. Closing

PR #2276, opened 2023-04-17
CASSANDRA-18336

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java
+++ b/src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java
@@ -19,6 +19,10 @@
 package org.apache.cassandra.service;
 
 
+import java.util.Set;
+
+import com.google.common.collect.ImmutableSet;
+
 import org.apache.cassandra.io.util.File;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -33,6 +37,8 @@ import org.apache.cassandra.utils.JVMStabilityInspector;
 public class DefaultFSErrorHandler implements FSErrorHandler
 {
     private static final Logger logger = LoggerFactory.getLogger(DefaultFSErrorHandler.class);
+
+    private static final Set<Class<?>> exceptionsSkippingDataRemoval = ImmutableSet.of(OutOfMemoryError.class);
 
     @Override
     public void handleCorruptSSTable(CorruptSSTableException e)
@@ -81,7 +87,7 @@ public class DefaultFSErrorHandler implements FSErrorHandler
 
                 // for both read and write errors mark the path as unwritable.
                 DisallowedDirectories.maybeMarkUnwritable(new File(e.path));
-                if (e instanceof FSReadError)
+                if (e instanceof FSReadError && shouldMaybeRemoveData(e))
                 {
                     File directory = DisallowedDirectories.maybeMarkUnreadable(new File(e.path));
                     if (directory != null)
@@ -94,6 +100,22 @@ public class DefaultFSErrorHandler implements FSErrorHandler
             default:
                 throw new IllegalStateException();
         }
+    }
+
+    private boolean shouldMaybeRemoveData(Throwable error)
+    {
+        for (Throwable t = error; t != null; t = t.getCause())
+        {
+            for (Class<?> c : exceptionsSkippingDataRemoval)
+                if (c.isAssignableFrom(t.getClass()))
+                    return false;
+            for (Throwable s : t.getSuppressed())
+                for (Class<?> c : exceptionsSkippingDataRemoval)
+                    if (c.isAssignableFrom(s.getClass()))
+                        return false;
+        }
+
+        return true;
     }
 
     @Override
--- a/test/unit/org/apache/cassandra/service/DiskFailurePolicyTest.java
+++ b/test/unit/org/apache/cassandra/service/DiskFailurePolicyTest.java
@@ -23,7 +23,6 @@ import java.util.Arrays;
 import java.util.Collection;
 
 import org.junit.After;
-import org.junit.Assert;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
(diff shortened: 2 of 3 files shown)


test/unit/org/apache/cassandra/service/DiskFailurePolicyTest.java
```diff
@@ -124,12 +135,29 @@ public void teardown()
...
         DatabaseDescriptor.setDiskFailurePolicy(testPolicy);
-        JVMStabilityInspector.inspectThrowable(t);
-        Assert.assertEquals(expectJVMKilled, killerForTests.wasKilled());
-        Assert.assertEquals(expectJVMKilledQuiet, killerForTests.wasKilledQuietly());
-        if (!expectJVMKilled) {
+        try
+        {
+            JVMStabilityInspector.inspectThrowable(t);
+        }
+        catch (OutOfMemoryError e)
```
smiklosovic, line 142: this is eventually thrown by `JVMStabilityInspector.inspectThrowable(t);` if you follow it deeply. We just need to make the difference between oom thrown as part of the test and oom thrown _while running a test_.

src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java
```diff
@@ -96,6 +96,20 @@ public void handleFSError(FSError e)
...
 
+    private boolean isCausedByOutOfMemoryException(Throwable error)
+    {
+        for (Throwable t = error; t != null; t = t.getCause())
+        {
+            if (t instanceof OutOfMemoryError)
+                return true;
+            for (Throwable s : t.getSuppressed())
+                if (s instanceof OutOfMemoryError)
+                return true;
```
Maxwell-Guo: code format : the "return true;" should have a "tab"  before ?
`if (s instanceof OutOfMemoryError)
                return true;`

src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java
```diff
@@ -81,7 +81,7 @@ public void handleFSError(FSError e)
 
                 // for both read and write errors mark the path as unwritable.
                 DisallowedDirectories.maybeMarkUnwritable(new File(e.path));
-                if (e instanceof FSReadError)
+                if (e instanceof FSReadError && !isCausedByOutOfMemoryException(e))
```
Maxwell-Guo: what about change the function to "isNotCausedByOutOfMemeoryException", so no "!" is needed before the function ,also the function logic should be rewrite ~~~~

src/java/org/apache/cassandra/utils/JVMStabilityInspector.java
```diff
@@ -168,7 +168,7 @@ else if (t instanceof UnrecoverableIllegalStateException)
             inspectThrowable(t.getCause(), fn);
     }
 
-    private static final Set<String> FORCE_HEAP_OOM_IGNORE_SET = ImmutableSet.of("Java heap space", "GC Overhead limit exceeded");
+    private static final Set<String> FORCE_HEAP_OOM_IGNORE_SET = ImmutableSet.of("Java heap space", "GC Overhead limit exceeded", "Java heap space test");
```
Maxwell-Guo: I think "Java heap space test" may be not needed here, as FORCE_HEAP_OOM_IGNORE_SET is not used in the ut.

test/unit/org/apache/cassandra/service/DiskFailurePolicyTest.java
```diff
@@ -79,22 +88,24 @@ public DiskFailurePolicyTest(DiskFailurePolicy testPolicy, boolean isStartUpInPr
...
+                             { die, false, new FSReadError(new IOException(), "blah"), false, true, false},
+                             { ignore, false, new FSReadError(new IOException(), "blah"), true, false, false},
+                             { stop, false, new FSReadError(new IOException(), "blah"), false, false, false},
+                             { stop_paranoid, false, new FSReadError(new IOException(), "blah"), false, false, false},
+                             { die, false, new CorruptSSTableException(new IOException(), "blah"), false, true, false},
+                             { ignore, false, new CorruptSSTableException(new IOException(), "blah"), true, false, false},
+                             { stop, false, new CorruptSSTableException(new IOException(), "blah"), true, false, false},
+                             { stop_paranoid, false, new CorruptSSTableException(new IOException(), "blah"), false, false, false},
+                             { best_effort, false, new FSReadError(new IOException(new OutOfMemoryError("Java heap space test")), "best_effort_oom"), true, false, false},
```
Maxwell-Guo: as "Java heap space test" removed in FORCE_HEAP_OOM_IGNORE_SET , so I think some other like "Java heap space" can be used here. I don't know if this is the right suggestion.

Maxwell-Guo (changes requested):

src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java
smiklosovic: thanks

src/java/org/apache/cassandra/utils/JVMStabilityInspector.java
smiklosovic: No, this has to be there in order to skip the logic in the method where it is referenced.

test/unit/org/apache/cassandra/service/DiskFailurePolicyTest.java
smiklosovic: see above

src/java/org/apache/cassandra/service/DefaultFSErrorHandler.java
smiklosovic: I rewrote the logic to be more generic.

src/java/org/apache/cassandra/utils/JVMStabilityInspector.java
Maxwell-Guo: what I mean is  such information("Java heap space test") will certainly not exist in the actual production environment, and the purpose of existence （"Java heap space test"） seems to be used only for ut testing. If it is purely used for testing, then why the information in ut does not directly use the existing exception message

Maxwell-Guo (changes requested):

src/java/org/apache/cassandra/utils/JVMStabilityInspector.java
smiklosovic: @Maxwell-Guo  What if _test itself_ throws OOM exception with a message "Java heap space"? How would you make the difference between legit OOM thrown during tests and OOM thrown as part of the test itself?

smiklosovic: @Maxwell-Guo If you have better solution, I am all ears.

smiklosovic: @Maxwell-Guo  I rewrote it so it is not there anymore.

Maxwell-Guo: what about using System.setProperty?
at https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/utils/JVMStabilityInspector.java#L182

`        if (FORCE_HEAP_OOM_IGNORE_SET.contains(oom.getMessage()))
            return;
        if (System.getProperty("oom.ut.test.message") != null && System.getProperty("oom.ut.test.message").contains(oom.getMessage()))
            return;
        logger.error("Force heap space OutOfMemoryError in the presence of", oom);
        // Start to produce heap space OOM forcibly.
        List<long[]> ignored = new ArrayList<>();`

and at https://github.com/apache/cassandra/blob/trunk/test/unit/org/apache/cassandra/service/DiskFailurePolicyTest.java#L57

we can setting 
`  static
    {
        System.setProperty("oom.ut.test.message", "Java heap space test");
    }
`
others can be left as it is

smiklosovic: I think the current solution I just pushed is good enough. You probably posted the comment before checking it?

Maxwell-Guo: sorry @smiklosovic , the version when I commented was not the latest, so I have such a question. 
+1 on this, and thanks for this work.

PR #2282, opened 2023-04-18
CASSANDRA-12937 - Default SSTable compression in cassandra.yaml

smiklosovic:
 
```

```


michaelsembwever:
why are we changing `sstable_compression` to `class` in the cql table properties? (this is a compatibility break) 

we need more tests:  
- tests around yaml-to-schema overrides, 
- tests around schema compression profiles over upgrade paths with new different yaml profiles,

[EDIT] thinking about it a bit more, the latter point doesn't need to be about upgrades. Just that these settings are applied only at table creation time, and that changing the yaml settings never interferes (even if you recreate the table from its DDL).

smiklosovic:
@michaelsembwever `sstable_compression` was marked as deprecated since 3.0 at minimum, maybe even earlier. That one is eligible for removal in 5.0, no?

michaelsembwever:
> sstable_compression was marked as deprecated since 3.0 at minimum, maybe even earlier. That one is eligible for removal in 5.0, no?

ah, so `class` as a replacement for `sstable_compression` is not something new.
Yes, definitely it is time for removal. Thanks, for the work and the explanation.

Claudenw:
Where are we with the change?

michaelsembwever:
the `class` change is good. 

it would be great to see a test around these settings being applied on a table at creation time, and then that changing the yaml settings never interferes (even if you recreate the table from its DDL).  there's a few variations here, and the consequences can be bad.

Claudenw:
@michaelsembwever I have been looking for examples similar to the testing your requested in your comments but I haven't found anything.  Can you point me to a test that is similar in that it is testing the change in Yaml doesn't interfere with some operation?  Or some other test upon which I can model the tests you ask for?

michaelsembwever:
> Can you point me to a test that is similar in that it is testing the change in Yaml doesn't interfere with some operation? Or some other test upon which I can model the tests you ask for?

maybe ConfigCompatabilityTest and CreateTableValidationTest are starting points…

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -43,189 +44,239 @@
...
                                                                        // 4 KiB is often the underlying disk block size
                                                                        1024 * 4,
                                                                        Integer.MAX_VALUE,
                                                                        DEFAULT_MIN_COMPRESS_RATIO,
-                                                                       Collections.emptyMap());
+                                                                       emptyMap());
 
-    private static final String CRC_CHECK_CHANCE_WARNING = "The option crc_check_chance was deprecated as a compression option. " +
-                                                           "You should specify it as a top-level table option instead";
+    public static final CompressionParams DEFAULT = new CompressionParams(LZ4Compressor.create(Collections.emptyMap()),
```
Claudenw: Do not change this to public.  Make it private and create a method `defaultFastCompression()` that returns DEFAULT, then call that in `DataComponent`.  We could also consider renaming DEFAULT to DEFAULT_FAST_COMPRESSOR.

test/distributed/org/apache/cassandra/distributed/test/SSTableCompressionTest.java
```diff
@@ -0,0 +1,454 @@
...
+        c.with(NATIVE_PROTOCOL, NETWORK, GOSSIP); // need gossip to get hostid for Java driver
+        c.set("flush_compression", "fast"); // this is default
+        c.set("sstable_compression", ImmutableMap.builder().put("class_name", "snappy").build());
+    };
+
+    private static final Consumer<IInstanceConfig> CHANGED_CONFIG = c -> {
+        c.with(NATIVE_PROTOCOL, NETWORK, GOSSIP); // need gossip to get hostid for Java driver
+        c.set("flush_compression", "fast"); // this is default
+        c.set("sstable_compression", ImmutableMap.builder().put("class_name", "deflate").build());
+    };
```
Claudenw, line 94: Let's change the names of the Consumer<IInstanceConfig> to DEFAULT_CONFIG, FAST_CONFIG, and SLOW_CONFIG.  
Also add string constants 
```
FAST_PARAM = "WITH compression = {'class': 'SnappyCompressor'}";
SLOW_PARAM = "WITH compression = {'class': 'DeflateCompressor'}"
DEFAULT_PARAM = "";
```

This should make the code easier to read.

test/distributed/org/apache/cassandra/distributed/test/SSTableCompressionTest.java
```diff
@@ -0,0 +1,454 @@
...
+            parameters = getCompressionParameters(CLUSTER.get(1), tableName);
+            assertEquals(ZstdCompressor.class.getName(), parameters.get("class"));
+        }
+        finally
+        {
+            tearDownCluster();
+        }
+    }
+
+    @Test
```
Claudenw, line 234: I think this test should have a comment explaining what it is testing and that it should be changed to cycle through the 3 configs (default, fast, slow) and create a table for each of the 3 table parameters (default, fast, slow).  For each table add some data, flush, create snapshot, verify snapshot compression type.  Reload the data, verify table compression is as expected.  This should give us a reasonable expectation that all the possible combinations of defaults and tables should work.

Claudenw (changes requested):

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -230,21 +230,27 @@ private static CompressionParams fromClassAndOptions(String sstableCompressionCl
...
+                return FBUtilities.newCompressor(parseCompressorClass(sstableCompressionClass), opt);
             }
-            catch (IllegalArgumentException expected)
+            else
             {
-                compressorType = CompressorType.forClass(sstableCompressionClass);
+                return FBUtilities.newCompressor(parseCompressorClass(defaultParams().klass().getName()), opt);
             }
-        }
+        };
```
Claudenw: THis block can be replaced with:

```
       Function<Map<String,String>, ICompressor> creator = null;
     	// try to set compressor type
        CompressorType compressorType = null;
        if (!StringUtils.isEmpty(sstableCompressionClass))
        {
            try
            {
                compressorType = CompressorType.valueOf(sstableCompressionClass);
            }
            catch (IllegalArgumentException expected)
            {
                compressorType = CompressorType.forClass(sstableCompressionClass);
            }
            creator = compressorType != null ? compressorType.creator : (opt) -> FBUtilities.newCompressor(parseCompressorClass(sstableCompressionClass), opt);
        } else {
            creator =  (opt) -> FBUtilities.newCompressor(parseCompressorClass(defaultParams().klass().getName()), opt);
        }
```

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -43,189 +44,246 @@
...
+        Function<Map<String,String>, ICompressor> creator = compressorType != null ? compressorType.creator : (opt) -> {
+            if (sstableCompressionClass != null)
+            {
+                return FBUtilities.newCompressor(parseCompressorClass(sstableCompressionClass), opt);
+            }
+            else
+            {
+                return FBUtilities.newCompressor(parseCompressorClass(defaultParams().klass().getName()), opt);
+            }
+        };
```
Claudenw: THis block can be replaced with:

       Function<Map<String,String>, ICompressor> creator = null;
     	// try to set compressor type
        CompressorType compressorType = null;
        if (!StringUtils.isEmpty(sstableCompressionClass))
        {
            try
            {
                compressorType = CompressorType.valueOf(sstableCompressionClass);
            }
            catch (IllegalArgumentException expected)
            {
                compressorType = CompressorType.forClass(sstableCompressionClass);
            }
            creator = compressorType != null ? compressorType.creator : (opt) -> FBUtilities.newCompressor(parseCompressorClass(sstableCompressionClass), opt);
        } else {
            creator =  (opt) -> FBUtilities.newCompressor(parseCompressorClass(defaultParams().klass().getName()), opt);
        }


This will reduce the reflection overhead on every lookup.

michaelsembwever:
your PRs need more informative titles @smiklosovic :)

conf/cassandra.yaml
```diff
@@ -570,6 +570,50 @@ commitlog_segment_size: 32MiB
...
+# Defines the default compression used on tables when none is specified
+# in the CQL command.
+#
+# The class_name is the compressor class name. It may be one of the aliases,
+# the class name of a system ICompressor implementation, or fully qualified
+# name of a class that implements ICompressor and has a public static 'create' method that accepts
+# a Map<String, String> argument and returns an instance of the class.
+#
+# class aliases are:
+#  Alias    System compressor impl.
```
michaelsembwever: this line isn't needed.

conf/cassandra.yaml
```diff
@@ -570,6 +570,50 @@ commitlog_segment_size: 32MiB
 #
 # flush_compression: fast
 
+# Defines the default compression used on tables when none is specified
+# in the CQL command.
+#
+# The class_name is the compressor class name. It may be one of the aliases,
+# the class name of a system ICompressor implementation, or fully qualified
+# name of a class that implements ICompressor and has a public static 'create' method that accepts
+# a Map<String, String> argument and returns an instance of the class.
```
michaelsembwever:
```suggestion
# Defines the default compression used on tables when none is specified
# in the CQL table schema.
#
# The class_name is the compressor class name or alias. External ICompressor implementations need to be fully qualified names.
```

conf/cassandra.yaml
```diff
@@ -570,6 +570,50 @@ commitlog_segment_size: 32MiB
...
+#
+# Only one of the min_compress_ratio and max_compressed_length options can be specified.
+# They are mathematically related in that 'chunk_length / max_compressed_length = min_compress_ratio'.
+# If neither option is specified a min_compress_ratio of 0.0 and a max_compressed_length of
+# Integer.MAX_VALUE KB is the default.
+#
+# Only one of chunk_length or chunk_length_in_kb may be specified.
+#
+# Additional class specific parameters may be added to the parameters section. The value of the class specific
+# parameter must be a string.
```
michaelsembwever:
```suggestion
# Parameter             Usage
# enabled               Disables compression if set to false. Defaults to true.
# chunk_length          The length of the compresson chunks, must include KiB, MiB or GiB suffix, defaults to 16KiB.
# min_compress_ratio    The minimal acceptable compression, must greater than or equal to 1.0.
# max_compressed_length The maximum size for a compressed block. Must be less than or equal to chunk_length. Cannot be specified with min_compress_ratio. 
#                       Must include KiB, MiB or GiB suffix. Defaults to Integer.MAX_VALUE
#
# The relation between parameters is 'chunk_length / max_compressed_length = min_compress_ratio'.
# If neither min_compress_ratio or max_compressed_length is specified these default to min_compress_ratio=0.0 and max_compressed_length=Integer.MAX_VALUE KB .
#
```

src/java/org/apache/cassandra/schema/TableParams.java
```diff
@@ -360,8 +360,14 @@ public Builder()
         {
         }
 
-        public TableParams build()
+        public TableParams build(String keyspace, String name)
         {
+            if (compression == null)
```
michaelsembwever: braces please (multiple ident exist)

src/java/org/apache/cassandra/schema/TableParams.java
```diff
@@ -360,8 +360,14 @@ public Builder()
...
+                if (SchemaConstants.getLocalAndReplicatedSystemKeyspaceNames().contains(keyspace))
+                    compression = CompressionParams.DEFAULT;
+                else
+                    compression = CompressionParams.defaultParams();
+
```
michaelsembwever: i'm uneasy about this change, hardcoding it to `.getLocalAndReplcateSystem KeyspaceNames()`, and introducing keyspace and table_name (the latter which is not even used anywhere) into the `.build()` method (which impacts a lot of calling code).


I would like more reviewers on this particular aspect, e.g. @blerer 
(and because of this I think we don't try to squeeze/rush this into 5.0)

smiklosovic: @michaelsembwever I am all ears how you would do that differently. This seems to be the least invasive change. I am aware of the fact that table parameter is not used _yet_. What if we want to have more granular resolution of this logic in the future? So table param would be already there ...

smiklosovic: BTW why is hardcoding wrong? We just want to exclude all of them, no?  We might have a dedicated structure where table -> compression is stored and we read it from there but that would mean somebody needs to take care of that every single time a new system table is going to be introduced.

Claudenw: We can remove the SchemaConstants.getLocalAndReplicatedSystemKeyspaceNames() and move the logic into CompressionParams, however, we will still need to know the keyspace name whenever a TableParams is used to construct a table so that we can provide the proper default CompressionParams when needed.

This solution could lead us to a point where we can specify different CompressionParams for different keyspaces, since we will already be doing that for the System tables as well as a few other keyspaces.

michaelsembwever: i'll have time to answer these questions in September, just too much of a backlog on everything else…

michaelsembwever: I believe this is resolved now …? @Claudenw 
the keyspace name is always passed into the builders.

Claudenw: This change has been merged into pull #3168 where the issues have been cleaned up and/or reworked.

PR #2285, opened 2023-04-19
CASSANDRA-18260 Add details to Error: Not enough space for compaction

henrikingo:
Adds separate WARNing level messages to the check itself in Directories.java.
    
patch by Henrik Ingo; reviewed by Mick Semb Wever for CASSANDRA-18260

### DIFF
--- a/src/java/org/apache/cassandra/db/Directories.java
+++ b/src/java/org/apache/cassandra/db/Directories.java
@@ -492,7 +492,6 @@ public class Directories
     {
         long writeSize = expectedTotalWriteSize / estimatedSSTables;
         long totalAvailable = 0L;
-        boolean hasSpace = true;
 
         for (DataDirectory dataDir : paths)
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -492,18 +492,39 @@ public boolean hasAvailableDiskSpace(long estimatedSSTables, long expectedTotalW
     {
         long writeSize = expectedTotalWriteSize / estimatedSSTables;
         long totalAvailable = 0L;
+        boolean hasSpace = true;
```
michaelsembwever: where is this used?

henrikingo: Nowhere, apparently. It is used in the trunk version, where this check has a different structure.

I'll remove it and thanks.

PR #2295, opened 2023-04-25
CASSANDRA-18441/trunk: Fix sstable formats configuration

jacek-lewandowski:
- refactored sstable format configuration
- sstable formats are discovered via ServiceLoader
- options configuration for sstable formats can be included in yaml
- yaml may include selected sstable format and version (version is not yet supported)
- auto saved caches refactored - they include additional metadata component which contains necessary mappings

The [Cassandra Jira](https://issues.apache.org/jira/projects/CASSANDRA/issues/CASSANDRA-18441)


src/java/org/apache/cassandra/config/Config.java
```diff
@@ -359,9 +352,14 @@ public MemtableOptions()
...
-                                                                                              ImmutableMap.of(SSTABLE_FORMAT_ID, "0",
-                                                                                                              SSTABLE_FORMAT_NAME, "big")));
+    public static class SSTableFormats
+    {
+        public String selected_format;
+        public String selected_version;
+        public Map<String, Map<String, String>> options = new HashMap<>();
+    }
+
+    public SSTableFormats sstable_formats = new SSTableFormats();
```
dcapwell: nit: can use `final` here

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -359,9 +352,14 @@ public MemtableOptions()
 
     public String[] data_file_directories = new String[0];
 
-    public List<ParameterizedClass> sstable_formats = ImmutableList.of(new ParameterizedClass(BigFormat.class.getName(),// "org.apache.cassandra.io.sstable.format.big.BigFormat",
-                                                                                              ImmutableMap.of(SSTABLE_FORMAT_ID, "0",
-                                                                                                              SSTABLE_FORMAT_NAME, "big")));
+    public static class SSTableFormats
+    {
+        public String selected_format;
+        public String selected_version;
```
dcapwell: I honestly keep debating this personally... should this be in `options`?

src/java/org/apache/cassandra/config/Config.java
dcapwell: This doesn't seem to match the yaml format talked about in JIRA?

```
  sstable:
    format:
      default: bti
      bti:
        row_index_granularity: 4KiB
```

src/java/org/apache/cassandra/config/YamlConfigurationLoader.java
```diff
@@ -203,6 +202,8 @@ private static void verifyReplacements(Map<Class<?>, Map<String, Replacement>> r
         Yaml rawYaml = new Yaml(loaderOptions);
 
         Map<String, Object> rawConfig = rawYaml.load(new ByteArrayInputStream(configBytes));
+        if (rawConfig == null)
```
dcapwell, line 203: What case caused this?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1349,68 +1359,97 @@ public static void applyPartitioner(Config conf)
...
     private static void applySSTableFormats()
     {
-        if (sstableFormatFactories != null)
-            logger.warn("Reinitializing SSTableFactories - this should happen only in tests");
+        ServiceLoader<SSTableFormat.Factory> loader = ServiceLoader.load(SSTableFormat.Factory.class, DatabaseDescriptor.class.getClassLoader());
+        List<SSTableFormat.Factory> factories = Iterables.toList(loader);
+        if (factories.isEmpty())
+            applySSTableFormats(ImmutableList.of(new BigFormat.BigFormatFactory()), conf.sstable_formats);
+        else
+            applySSTableFormats(factories, conf.sstable_formats);
```
dcapwell:
```suggestion
        if (factories.isEmpty())
            factories = ImmutableList.of(new BigFormat.BigFormatFactory());
        applySSTableFormats(factories, conf.sstable_formats);
```

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1349,68 +1359,97 @@ public static void applyPartitioner(Config conf)
...
+    private static void validateSSTableFormatFactories(Iterable<SSTableFormat.Factory> factories)
     {
-        ImmutableMap.Builder<String, Supplier<SSTableFormat<?, ?>>> sstableFormatFactories = ImmutableMap.builderWithExpectedSize(configuredFormats.size());
-        Set<String> names = new HashSet<>(configuredFormats.size());
-        Set<Integer> ids = new HashSet<>(configuredFormats.size());
+        Map<String, SSTableFormat.Factory> factoryByName = new HashMap<>();
+        for (SSTableFormat.Factory factory : factories)
+        {
+            if (factory.name() == null)
+                throw new ConfigurationException(String.format("SSTable format name in %s cannot be null", factory.getClass().getName()));
```
dcapwell:
```suggestion
                throw new ConfigurationException(String.format("SSTable format name in %s cannot be null", factory.getClass().getCanonicalName()));
```

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1349,68 +1359,97 @@ public static void applyPartitioner(Config conf)
...
+
+            if (!factory.name().matches("^[a-z]+$"))
+                throw new ConfigurationException(String.format("SSTable format name for %s must be non-empty, lower-case letters only string", factory.getClass().getName()));
```
dcapwell:
```suggestion
                throw new ConfigurationException(String.format("SSTable format name for %s must be non-empty, lower-case letters only string", factory.getClass().getCanonicalName()));
```

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1349,68 +1359,97 @@ public static void applyPartitioner(Config conf)
...
 
-        for (ParameterizedClass formatConfig : configuredFormats)
+            SSTableFormat.Factory prev = factoryByName.put(factory.name(), factory);
+            if (prev != null)
+                throw new ConfigurationException(String.format("Multiple sstable format implementations with the same name %s: %s and %s", factory.name(), factory.getClass().getName(), prev.getClass().getName()));
```
dcapwell:
```suggestion
                throw new ConfigurationException(String.format("Multiple sstable format implementations with the same name %s: %s and %s", factory.name(), factory.getClass().getCanonicalName(), prev.getClass().getCanonicalName()));
```

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1349,68 +1359,97 @@ public static void applyPartitioner(Config conf)
...
-                    throw new ConfigurationException(String.format("Class %s for sstable format %s does not implement %s", formatConfig.class_name, name, SSTableFormat.class.getName()));
-
-                SSTableFormat<?, ?> sstableFormat = FBUtilities.instanceOrConstruct(cls.getName(), "sstable format");
-                sstableFormat.setup(id, name, params);
-                return sstableFormat;
-            };
-            sstableFormatFactories.put(name, factory);
-            names.add(name);
-            ids.add(id);
+            Map<String, String> formatOptions = options != null ? ImmutableMap.copyOf(options.getOrDefault(factory.name(), ImmutableMap.of())) : ImmutableMap.of();
```
dcapwell: rather than checking `options` in the loop you can do the following outside the loop

```
if (options == null)
  options = ImmutableMap.of();
```

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1349,68 +1359,97 @@ public static void applyPartitioner(Config conf)
...
+        }
+        return providers;
+    }
 
-        return sstableFormatFactories.build();
+    private static Pair<SSTableFormat<?, ?>, Version> getAndValidateFormatAndVersion(Map<String, SSTableFormat<?, ?>> sstableFormats, String selectedFormatName, String selectedVersionStr)
+    {
+        SSTableFormat<?, ?> selectedFormat;
+        if (StringUtils.isBlank(selectedFormatName))
+            selectedFormatName = BigFormat.NAME;
```
dcapwell, line 1420: I prefer this to live in `Config` so defaults are more clear and not scattered all over the place

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
```diff
@@ -113,99 +114,85 @@ IScrubber getScrubber(ColumnFamilyStore cfs,
...
-
     /**
      * Deletes the existing components of the sstables represented by the provided descriptor.
      * The method is also responsible for cleaning up the in-memory resources occupied by the stuff related to that
      * sstables, such as row key cache entries.
      */
     void delete(Descriptor descriptor);
 
+    /**
+     * This class is not completely redundant
```
dcapwell: I do feel that this is redundant, its just another reference to the values in `org.apache.cassandra.config.DatabaseDescriptor#sstableFormats` and 100% depends on that...

`current` -> `DD.getSelectedSSTableFormat`
`getByName` -> `DD.sstableFormats.get(name)`
`getByClass` -> 

```
DatabaseDescriptor.getSSTableFormats().values().stream()
                              .filter(f -> f.getClass().equals(formatClass))
                              .findFirst()
                              .orElseGet(() -> {
                                  throw new NoSuchElementException("Unknown sstable format class: " + formatClass);
                              });
```

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
```diff
@@ -54,9 +50,14 @@
  */
 public interface SSTableFormat<R extends SSTableReader, W extends SSTableWriter>
 {
-    int ordinal();
     String name();
 
+    @Deprecated
+    default Type getType()
```
dcapwell: only used for tests, and is redundant as `name` and `this.getClass()` is known by holding reference to this

test/unit/org/apache/cassandra/config/DatabaseDescriptorRefTest.java
```diff
@@ -198,10 +200,11 @@
     "org.apache.cassandra.io.sstable.MetricsProviders",
     "org.apache.cassandra.io.sstable.SSTable",
     "org.apache.cassandra.io.sstable.SSTable$Builder",
-    "org.apache.cassandra.io.sstable.format.AbstractSSTableFormat",
     "org.apache.cassandra.io.sstable.format.SSTableFormat",
     "org.apache.cassandra.io.sstable.format.SSTableFormat$Components",
     "org.apache.cassandra.io.sstable.format.SSTableFormat$Components$Types",
+    "org.apache.cassandra.io.sstable.format.SSTableFormat$Factory",
+    "org.apache.cassandra.io.sstable.format.SSTableFormat$Factory",
```
dcapwell: duplicate

test/unit/org/apache/cassandra/config/DatabaseDescriptorRefTest.java
```diff
@@ -217,6 +220,8 @@
     "org.apache.cassandra.io.sstable.format.SortedTableWriter$Builder",
     "org.apache.cassandra.io.sstable.format.Version",
     "org.apache.cassandra.io.sstable.format.big.BigFormat",
+    "org.apache.cassandra.io.sstable.format.big.BigFormat$BigFormatFactory",
+    "org.apache.cassandra.io.sstable.format.big.BigFormat$BigFormatFactory",
```
dcapwell: duplicate

test/unit/org/apache/cassandra/schema/MockSchema.java
```diff
@@ -170,7 +170,7 @@ public static SSTableReader sstable(int generation, int size, boolean keepRef, l
                                                sstableId(generation),
                                                format.getType());
 
-        if (format == BigFormat.getInstance())
+        if (format instanceof BigFormat)
```
dcapwell: I am ok with this, but the rest of the patch uses `org.apache.cassandra.io.sstable.format.big.BigFormat#is`, which does a `name` check

dcapwell (changes requested):
Did first pass, didn't look at the cache changes but overall the idea and direction LGTM, most of the feedback are small things and not real structural things.

Ill take a look at the cache layer next

src/java/org/apache/cassandra/config/Config.java
dcapwell: my thinking is, if we do want to later (not this patch) allow overriding via JMX or Settings table, how can you do thing safely?

Lets say we have the following cases

1) keep same format, just change version (use case; major/minor upgrade wants to delay using latest until "stable")

```
sstable_formats.selected_version = next // this is "safe"
```

2) switch from big to bti or some 3rd party format

```
sstable_formats.selected_version = bti_version // unsafe, breaks BIG
sstable_formats.selected_format = bti
```

or

```
sstable_formats.selected_format = bti // unsafe as "selected_version" still points to BIG version
sstable_formats.selected_version = bti_version
```

src/java/org/apache/cassandra/cache/AutoSavingCache.java
```diff
@@ -196,12 +205,18 @@ public int loadSaved()
         // modern format, allows both key and value (so key cache load can be purely sequential)
         File dataPath = getCacheDataPath(CURRENT_VERSION);
         File crcPath = getCacheCrcPath(CURRENT_VERSION);
-        if (dataPath.exists() && crcPath.exists())
+        File metadataPath = getCacheMetadataPath(CURRENT_VERSION);
+        if (dataPath.exists() && crcPath.exists() && metadataPath.exists())
```
dcapwell, line 209: unrelated to this patch, sad that upgrading from 4.1 -> 5.0 ignores the cache as we don't support reading older versions.

dcapwell (commented):
cache logic LGTM, like the new metadata file optmization

src/java/org/apache/cassandra/config/YamlConfigurationLoader.java
jacek-lewandowski: Just for testing - wanted to make it possible to pass an empty yaml to assert the defaults. I think this is a valid situation - whole yaml should be optional.

test/unit/org/apache/cassandra/config/DatabaseDescriptorRefTest.java
jacek-lewandowski: 🤦🏻‍♂️

jacek-lewandowski:
Thank you @dcapwell for quick feedback, really appreciate that!

test/unit/org/apache/cassandra/config/DatabaseDescriptorRefTest.java
jacek-lewandowski: fixed

test/unit/org/apache/cassandra/config/DatabaseDescriptorRefTest.java
jacek-lewandowski: fixed

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
jacek-lewandowski: fixed

src/java/org/apache/cassandra/config/Config.java
jacek-lewandowski: fixed

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
jacek-lewandowski: Did I really write "not"??? 🤦🏻‍♂️ I wanted to write it is redundant and that's why I marked it deprecated. That "not" is a typing mistake.

The only reason I haven't removed it is that the additional changeset would be quite large. So if we agree on that, we can do that.

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
jacek-lewandowski: yes, that's why it is marked as deprecated and left only to not pollute the pull request.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
jacek-lewandowski: fixed

src/java/org/apache/cassandra/cache/AutoSavingCache.java
jacek-lewandowski: This is doable, but I don't know if complicating this code is worth restoring the cache from previous version 🤷🏻‍♂️

test/unit/org/apache/cassandra/schema/MockSchema.java
jacek-lewandowski: fixed

src/java/org/apache/cassandra/config/Config.java
jacek-lewandowski: Where would we express the expected target format and version then? If we ever go for changing that at runtime, this should always be changed together.

dcapwell: > this should always be changed together

agree, but we would need locking or copy-on-write configs to support that, which we don't have in this patch (and not arguing we add).  This means a config change happens each time a field is modified, which could put us in a inconsistence state waiting for the operator to update the other fields.

So, by having the version in `options` this at least scopes things to where they make sense... if you are using `big` and want to test out `bti v42` you can update the configs for `bti.version=42`, this would have no runtime impact as its untouched.  But let's say you are using `bti v41` and want to switch to `v42`?  you then do `bit.version=42` and the next sstable uses `v42`.  By having the version in the options it makes the concurrency easier to reason about

src/java/org/apache/cassandra/config/YamlConfigurationLoader.java
dcapwell: I am ok with it, just making sure there wasn't an existing bug lurking

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
dcapwell: I am cool with you leaving this until the review is complete, then fixing at the end.

src/java/org/apache/cassandra/cache/AutoSavingCache.java
dcapwell: Given that this change already landed, I don't mind this being punted from this PR, but would be good to ask in dev@?  This would be a performance issue during major upgrades but once warm you stop seeing it...

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1349,68 +1359,98 @@ public static void applyPartitioner(Config conf)
...
-                return sstableFormat;
-            };
-            sstableFormatFactories.put(name, factory);
-            names.add(name);
-            ids.add(id);
+            Map<String, String> formatOptions = ImmutableMap.copyOf(options.getOrDefault(factory.name(), ImmutableMap.of()));
+            providersBuilder.put(factory.name(), () -> factory.getInstance(formatOptions));
+        }
+        ImmutableMap<String, Supplier<SSTableFormat<?, ?>>> providers = providersBuilder.build();
+        if (options != null)
```
dcapwell, line 1407: not possible after latest commit

dcapwell (commented):
Thanks for all the changes! Here is my list of changes needed before I +1

1) SSTableFormat.Type goes away
2) Config matches YAML defined in JIRA
3) version move to options (this is in-line with 2, this seems to be the agreed format in JIRA)

src/java/org/apache/cassandra/config/Config.java
jacek-lewandowski: agreed; though I was also thinking about specifying the write format like:

```
write_format: big
```
or
```
write_format: big-nc
```

Though, as discussed in Jira, we will define a version per format.

conf/cassandra.yaml
```diff
@@ -1928,14 +1928,5 @@ drop_compact_storage_enabled: false
...
-# implementing org.apache.cassandra.io.sstable.format.SSTableFormat. Parameters must include unique 'id' integer
-# which is used in some serialization to denote the format type in a compact way (such as local key cache); and 'name'
-# which will be used to recognize the format type - in particular that name will be used in sstable file names and in
-# stream headers so the name has to be the same for the same format across all the nodes in the cluster.
-sstable_formats:
-  - class_name: org.apache.cassandra.io.sstable.format.big.BigFormat
-    parameters:
-      id: 0
-      name: big
+#sstable_formats:
```
dcapwell: can you update the yaml to match the `Config` change?

src/java/org/apache/cassandra/db/lifecycle/LogFile.java
```diff
@@ -504,7 +507,7 @@ List<String> getFilePaths()
 
     private String getFileName()
     {
-        return StringUtils.join(SSTableFormat.Type.current().info.getLatestVersion(), LogFile.SEP, // remove version and separator when downgrading to 4.x is becomes unsupported
+        return StringUtils.join(version, LogFile.SEP, // remove version and separator when downgrading to 4.x is becomes unsupported TODO should this include format type as well?
```
dcapwell: > TODO should this include format type as well?

if the version does not include the format, I think it would be good... `nc` isn't useful as I don't know this is `big` or `bti`

Another option is to use `version.toFormatAndVersionString`?

dcapwell (changes requested):
very small changes, but overall LGTM; once cassandra.yml is updated I am +1

src/java/org/apache/cassandra/db/lifecycle/LogFile.java
jacek-lewandowski: Yes, I think that I oversaw that in CEP-17

dcapwell (approved):

jacek-lewandowski:
Committed https://github.com/apache/cassandra/commit/fe0e04c2319afab958b3da83e7b54c84bced9dc2

PR #2299, opened 2023-04-27
CASSANDRA-18485: CEP-15: (C*) Enhance in-memory FileSystem to work with mmap and support tests to add custom logic

dcapwell:


src/java/org/apache/cassandra/utils/ByteBufferUtil.java
```diff
@@ -866,4 +867,15 @@ private static boolean startsWith(ByteBuffer src, ByteBuffer prefix, int offset)
 
         return true;
     }
+
+    public static void readFully(FileChannel channel, ByteBuffer dst, long position) throws IOException
```
dcapwell, line 871: I feel that this has to exist somewhere....

src/java/org/apache/cassandra/utils/SyncUtil.java
```diff
@@ -99,6 +101,12 @@
     public static MappedByteBuffer force(MappedByteBuffer buf)
     {
         Preconditions.checkNotNull(buf);
+        Object attachment = MemoryUtil.getAttachment(buf);
+        if (attachment instanceof Runnable)
```
dcapwell, line 105: the `FileSystem` runs in app `ClassLoader` but this logic runs in `InstanceClassLoader`, so I can't use a non standard type here...

This is a field access, but if there are concerns about this I can always make this a test flag, so we only enter this if a static field is set to true

test/distributed/org/apache/cassandra/distributed/impl/AbstractCluster.java
```diff
@@ -189,12 +189,14 @@
...
             CassandraRelevantProperties.FLUSH_LOCAL_SCHEMA_CHANGES.reset();
             CassandraRelevantProperties.NON_GRACEFUL_SHUTDOWN.reset();
+            CassandraRelevantProperties.IGNORE_MISSING_NATIVE_FILE_HINTS.setBoolean(true);
         }
 
         public AbstractBuilder(Factory<I, C, B> factory)
         {
             super(factory);
             withSharedClasses(SHARED_PREDICATE);
+            withRoot(org.apache.cassandra.io.util.Files.newInMemoryFileSystem().getPath("/cassandra"));
```
dcapwell: for now I enable by default just to see how CI works, feel it would be best to make this a flag for tests to opt-in/out of

test/simulator/main/org/apache/cassandra/simulator/ClusterSimulation.java
```diff
@@ -618,9 +615,17 @@ public ClusterSimulation(RandomSource random, long seed, int uniqueNum,
...
+                                                                                                                          .setMaxSize(4L << 30).setBlockSize(512)
+                                                                                                                          .build()));
+        jimfs.listen((ListenableFileSystem.OnRead) (path, channel, position, dst, read) -> {
+            if (read > 0 && path.getFileName().toString().endsWith("Index.db") && random.decide(.05f))
+            {
+                int offset = random.uniform(0, read);
+                int index = dst.position() - offset;
+                dst.put(index, (byte) (dst.get(index) + 1));
+            }
+        });
```
dcapwell: this will go away, I was using it to test this specific case... in reviewing Index.db code I saw we didn't validate it, so figured it would be easy to test corruption real quick... this listener does what I wanted and corrupts 5% of `FileChannel.read` on `Index.db`... (map is backed by `.read`, so will also trigger)

test/simulator/main/org/apache/cassandra/simulator/ClusterSimulation.java
```diff
@@ -683,9 +688,10 @@ public ClusterSimulation(RandomSource random, long seed, int uniqueNum,
                              .set("memtable_allocation_type", builder.memoryListener != null ? "unslabbed_heap_buffers_logged" : "heap_buffers")
                              .set("file_cache_size", "16MiB")
                              .set("use_deterministic_table_id", true)
-                             .set("disk_access_mode", "standard")
+//                             .set("disk_access_mode", "standard")
```
dcapwell: wondering if it would be good to have these configs based off the `RandomSource`?  that way we test all code paths rather than specific ones?

test/simulator/test/org/apache/cassandra/simulator/test/ShortPaxosSimulationTest.java
```diff
@@ -30,7 +30,7 @@
     @Test
     public void simulationTest() throws IOException
     {
-        PaxosSimulationRunner.main(new String[] { "run", "-n", "3..6", "-t", "1000", "-c", "2", "--cluster-action-limit", "2", "-s", "30" });
+        PaxosSimulationRunner.main(new String[] { "run", "-n", "3..6", "-t", "1000", "-c", "2", "--cluster-action-limit", "2", "-s", "30", "--seed", "0xa03aba138c555b0d"});
```
dcapwell: will revert

test/unit/org/apache/cassandra/io/filesystem/ListenableFileSystem.java
```diff
@@ -0,0 +1,510 @@
...
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.apache.cassandra.utils.ByteBufferUtil;
+import org.apache.cassandra.utils.memory.MemoryUtil;
+
+public class ListenableFileSystem extends ForwardingFileSystem
+{
+    public interface Listener
```
dcapwell, line 64: not a fan of this API, it was mostly just to quickly instrument the different paths so I could write tests that would corrupt specific cases... I was thinking of having a Interceptor Chain pattern, but this was quick and dirty to write...

test/unit/org/apache/cassandra/io/filesystem/ListenableFileSystem.java
```diff
@@ -0,0 +1,510 @@
...
+        public long transferFrom(ReadableByteChannel src, long position, long count) throws IOException
+        {
+            notifyListeners(onTransferFrom, l -> l.preTransferFrom(path, this, src, position, count));
+            long transfered = super.transferFrom(src, position, count);
+            notifyListeners(onTransferFrom, l -> l.postTransferFrom(path, this, src, position, count, transfered));
+            return transfered;
+        }
+
+        @Override
+        public MappedByteBuffer map(MapMode mode, long position, long size) throws IOException
```
dcapwell, line 762: since this is based off the buffer cache... we "might" be able to get this to be "correct" by having the read/write methods change behavior when a buffer exists for the provided offset?  I didn't go down this path as nothing in C* looked to need that complexity

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -318,6 +319,8 @@ public static JMXServiceURL getJMXServiceURL() throws MalformedURLException
     @BeforeClass
     public static void setUpClass()
     {
+        ServerTestUtils.daemonInitialization();
```
dcapwell, line 323: this is to allow `InMemory` class, as I need to setup FileSystem before this function call

test/unit/org/apache/cassandra/db/CorruptPrimaryIndexTest.java
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.cassandra.cql3.CQLTester;
+import org.apache.cassandra.cql3.UntypedResultSet;
+import org.apache.cassandra.io.filesystem.ListenableFileSystem;
+import org.apache.cassandra.io.sstable.Descriptor;
+import org.apache.cassandra.io.util.File;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+public class CorruptPrimaryIndexTest extends CQLTester.InMemory
```
dcapwell, line 29: this test is what motived me to take up this work...

test/unit/org/apache/cassandra/io/filesystem/ListenableFileSystem.java
```diff
@@ -0,0 +1,541 @@
...
+    private final List<List<? extends Listener>> lists = Arrays.asList(onOpen, onRead, onTransferTo, onWrite, onTransferFrom, onChannelMeta);
+    private final ListenableFileSystemProvider provider;
+
+    public ListenableFileSystem(FileSystem delegate)
+    {
+        super(delegate);
+        this.provider = new ListenableFileSystemProvider(super.provider());
+    }
+
+    public Unsubscribable listen(Listener listener)
```
dcapwell, line 187: feedback *strongly* welcomed on this API as I actually hate it... but not sure what would be better...  want to expose every call to file system so tests can do w/e they want, but current one is rather ugly to work with...

test/unit/org/apache/cassandra/io/util/Files.java
```diff
@@ -0,0 +1,94 @@
...
+import com.google.common.base.StandardSystemProperty;
+import com.google.common.jimfs.Configuration;
+import com.google.common.jimfs.Jimfs;
+
+import org.apache.cassandra.io.filesystem.ForwardingFileSystem;
+import org.apache.cassandra.io.filesystem.ListenableFileSystem;
+
+public class Files
+{
+    public static ListenableFileSystem newInMemoryFileSystem()
```
maedhroz: nit: unused

test/unit/org/apache/cassandra/io/util/Files.java
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    private static Configuration jimfsConfig()
+    {
+        return Configuration.unix().toBuilder()
+                            .setMaxSize(4L << 30).setBlockSize(512)
+                            .build();
+    }
+
+    public static File maybeCreateTmp()
```
maedhroz: nit: What if we just move this to `File` itself? Also, alternate API could be something like `createTmpDir(boolean overwrite)`, which would just be called w/ `createTmpDir(false)`...dunno

test/unit/org/apache/cassandra/io/util/Files.java
```diff
@@ -0,0 +1,94 @@
...
+        return global(File.unsafeGetFilesystem());
+    }
+
+    public static ListenableFileSystem global(FileSystem real)
+    {
+        FileSystem current = File.unsafeGetFilesystem();
+        ListenableFileSystem fs = new ListenableFileSystem(new ForwardingFileSystem(real)
+        {
+            @Override
+            public void close() throws IOException
```
maedhroz: nit: Does anything ever actually close this?

test/unit/org/apache/cassandra/io/util/Files.java
```diff
@@ -0,0 +1,94 @@
...
+import java.nio.file.FileSystem;
+
+import com.google.common.base.StandardSystemProperty;
+import com.google.common.jimfs.Configuration;
+import com.google.common.jimfs.Jimfs;
+
+import org.apache.cassandra.io.filesystem.ForwardingFileSystem;
+import org.apache.cassandra.io.filesystem.ListenableFileSystem;
+
+public class Files
```
maedhroz: Just thinking about the general usage of this class...

1.) I'd consider naming it `Filesystem` or `Filesystems`

2.) Would it be better to separate filesystem creation and replacement? For example...

```
FileSystem jim = Filesystems.newJimFs();
Filesystems.replace(jim, true); // where in replace(FileSystem fs, boolean restoreOnClose), restoreOnClose specifies whether to restore the original FS
```

test/unit/org/apache/cassandra/io/filesystem/ForwardingFileStore.java
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+package org.apache.cassandra.io.filesystem;
+
+import java.io.IOException;
+import java.nio.file.FileStore;
+import java.nio.file.attribute.FileAttributeView;
+import java.nio.file.attribute.FileStoreAttributeView;
+
+public class ForwardingFileStore extends FileStore
```
maedhroz: nit: unused

test/unit/org/apache/cassandra/io/filesystem/ForwardingFileSystem.java
```diff
@@ -0,0 +1,129 @@
...
+
+    protected Path unwrap(Path p)
+    {
+        return p;
+    }
+
+    @Override
+    public FileSystemProvider provider()
+    {
+        return delegate().provider();
```
maedhroz, line 60: heh, IDEA's static analysis complains all over the place about these `delegate()` calls not being in try-with-resources blocks, but it isn't smart enough to realize it's equivalent to `delegate.provider()` and isn't actually creating anything :D

Could make the delegate private and use it directly internally w/ only the accessor available at protected, but whatever...

test/unit/org/apache/cassandra/io/filesystem/ListenableFileSystem.java
```diff
@@ -0,0 +1,541 @@
...
+                for (OpenOption opt : options)
+                {
+                    // All OpenOption values except for APPEND and WRITE are allowed
+                    if (opt == StandardOpenOption.APPEND ||
+                        opt == StandardOpenOption.WRITE)
+                        throw new UnsupportedOperationException("'" + opt + "' not allowed");
+                }
+            }
+            Set<OpenOption> opts = new HashSet<>();
+            opts.addAll(Arrays.asList(options));
```
maedhroz: Could just be `Set<OpenOption> opts = new HashSet<>(Arrays.asList(options));`

test/unit/org/apache/cassandra/io/filesystem/ListenableFileSystem.java
```diff
@@ -0,0 +1,541 @@
...
+                }
+            }
+            opts.add(StandardOpenOption.WRITE);
+            return Channels.newOutputStream(newFileChannel(path, opts));
+        }
+
+        @Override
+        public InputStream newInputStream(Path path, OpenOption... options) throws IOException
+        {
+            if (options.length > 0)
```
maedhroz: nit: if check not strictly necessary

test/unit/org/apache/cassandra/io/util/Files.java
dcapwell: I don't like moving to `File` as that has a path already, so `new File("/home").createTmpDir(false)` is not clear what happens...

dcapwell: we could have a static function... but then why not leave here with other functions that work with `Files`?

test/unit/org/apache/cassandra/io/util/Files.java
dcapwell: if you try to create a new FS per test you need to, else the old fs can't free memory... jimfs registers itself with the global `FileSystems`, so a reference will be held if you do not *close*

test/unit/org/apache/cassandra/io/filesystem/ListenableFileSystem.java
maedhroz: YAGNI

test/unit/org/apache/cassandra/io/util/Files.java
dcapwell: in the single example in this PR, its localized to the class... so don't actually need to... but I wanted to make sure you can create a fs for your test even if another exists, or if each test wants to create in isolation...

test/unit/org/apache/cassandra/io/filesystem/ListenableFileSystem.java
maedhroz: Had a quick chat about this in Slack, will return to review later when changes that came out of that are in place...

src/java/org/apache/cassandra/utils/SyncUtil.java
maedhroz: What touches this? The commitlog? I suppose erring on the side of caution would be okay...

test/unit/org/apache/cassandra/io/util/Files.java
dcapwell: `Filesystems.replace` is a duplicate of `org.apache.cassandra.io.util.File#unsafeSetFilesystem`... and `global` method kinda does what you are asking, but when `true`...

so in short, if you want `false` just call `org.apache.cassandra.io.util.File#unsafeSetFilesystem`.  If you want `true` call `global`?

test/unit/org/apache/cassandra/io/filesystem/ForwardingFileStore.java
dcapwell: I had it for completnexx... but yeah I didn't end up using... can remove

test/unit/org/apache/cassandra/io/filesystem/ForwardingFileSystem.java
dcapwell: I prefer this method as it allows subclasses to override... I actually used that during testing to make sure I knew what was touched and what wasn't

test/unit/org/apache/cassandra/io/filesystem/ListenableFileSystem.java
dcapwell: was copy/paste from java... prob to avoid iterators?

test/unit/org/apache/cassandra/io/util/Files.java
maedhroz: Right, it would be static. My reasoning was that this class works w/ filesystems, not files. (Hence the renaming suggestion.)

test/unit/org/apache/cassandra/io/util/Files.java
maedhroz: Right, you can get all this behavior w/ the existing methods...just trying to put an alternate API on top of it.

maedhroz: Either way, the class name `Files` seems suboptimal.

test/unit/org/apache/cassandra/io/filesystem/ListenableFileSystem.java
maedhroz: Depends on how smart the runtime is I guess?

test/unit/org/apache/cassandra/io/filesystem/ForwardingFileSystem.java
maedhroz: k, I guess if you want to override the delegate on the fly, not just at construction time, makes sense

maedhroz (approved):

src/java/org/apache/cassandra/utils/SyncUtil.java
dcapwell: listenable fs... we need a way to fake sync, and C* detects that we don't have a mmap buffer and rejects

test/unit/org/apache/cassandra/io/util/Files.java
dcapwell: fair, but this is only used for new file systems, and not a core `File` usage

test/unit/org/apache/cassandra/io/util/Files.java
dcapwell: > Right, you can get all this behavior w/ the existing methods...just trying to put an alternate API on top of it.

in this case I feel 2 methods are better than a boolean; easier to document and I feel easier to read while reviewing code

test/unit/org/apache/cassandra/db/CorruptPrimaryIndexTest.java
```diff
@@ -0,0 +1,56 @@
...
+        // Set listener early, before the file is opened; mmap access can not be listened to, so need to observe the open, which happens on flush
+        fs.onPostRead(isCurrentTableIndexFile(keyspace()), (path, channel, position, dst, read) -> {
+            // Reading the Primary index for the test!
+            // format
+            // 2 bytes: length of bytes for PK
+            // 4 bytes: pk as an int32
+            // variable bytes (see org.apache.cassandra.io.sstable.format.big.RowIndexEntry.IndexSerializer.deserialize(org.apache.cassandra.io.util.FileDataInput))
+            assertThat(position).describedAs("Unexpected access, should start read from start of file").isEqualTo(0);
+
+            // simulate bit rot by having 1 byte change... but make sure its the pk!
```
maedhroz:
```suggestion
            // simulate bit rot by having 1 byte change... but make sure it's the pk!
```

test/unit/org/apache/cassandra/io/filesystem/ListenableFileSystem.java
```diff
@@ -0,0 +1,850 @@
...
+            }
+            else if (mode == MapMode.READ_WRITE)
+            {
+                if (delegate().size() - position > 0)
+                {
+                    ByteBufferUtil.readFully(this, mapped.bb, position);
+                    mapped.bb.flip();
+                }
+                // with real files the FD gets copied so the close of the channel does not block the BB mutation
+                // from flushing...  its possible to support this use case, but kept things simplier for now by
```
maedhroz:
```suggestion
                // from flushing...  it's possible to support this use case, but kept things simplier for now by
```

test/unit/org/apache/cassandra/io/filesystem/ListenableFileSystem.java
```diff
@@ -0,0 +1,850 @@
...
+        {
+            onPrePosition.add((OnPrePosition) listener);
+            match = true;
+        }
+        if (listener instanceof OnPostPosition)
+        {
+            onPostPosition.add((OnPostPosition) listener);
+            match = true;
+        }
+        if (!match)
```
maedhroz: nit: What if, instead of tracking a boolean here, we just track which list (ex. `List<? extends Listener> match`) we add to, and then have the closure remove specifically from that list, instead of trying to remove it from everything? (i.e. `match` is null when there's no match above, an we throw and do nothing)

PR #2300, opened 2023-04-28
CASSANDRA-15254 Support updates over SettingsTable (prorperty validation with DatabaseDescriptor)

Mmuzaf:
 
```

```


src/java/org/apache/cassandra/config/registry/DatabaseConfigurationSource.java
```diff
@@ -0,0 +1,308 @@
...
+    private final Map<String, Property> properties;
+
+    public DatabaseConfigurationSource(Config source)
+    {
+        this.source = source;
+        properties = ImmutableMap.copyOf(defaultLoader()
+                                         .flatten(Config.class)
+                                         .entrySet()
+                                         .stream()
+                                         .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));
```
dcapwell, line 73: why are you converting to a stream if you are taking the map and creating an identical map?

src/java/org/apache/cassandra/config/registry/DatabaseConfigurationSource.java
```diff
@@ -0,0 +1,308 @@
...
+            // Do conversion if the value is not null and the type is not the same as the property type.
+            Object convertedValue = ofNullable(value)
+                                    .map(Object::getClass)
+                                    .map(from -> converters.get(from, originalType)
+                                                           .convert(value))
+                                    .orElse(null);
+            // Use given converted value if it is not null, otherwise use the default value.
+            ConfigurationSource validationSource = PropertyValidationSource.create(this, name, convertedValue);
+            for (ConfigurationSourceValidator validator : validators)
+                validator.validate(validationSource);
```
dcapwell, line 103: this should be scoped to the field and not a global set of validators...

src/java/org/apache/cassandra/config/registry/DatabaseConfigurationSource.java
```diff
@@ -0,0 +1,308 @@
...
+
+
+/**
+ * This is a simple configuration property registry that stores all the {@link Config} settings, it doesn't
+ * take into account any configuration changes that might happen during properties replacement between releases.
+ */
+public class DatabaseConfigurationSource implements ConfigurationSource
+{
+    private static final Logger logger = LoggerFactory.getLogger(DatabaseConfigurationSource.class);
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
```
dcapwell, line 59: the locking in this class isn't thread safe, but may also cause performance regressions... the current model does not need locking, so not sure why we need to add a new set of locks (and global at that)

src/java/org/apache/cassandra/config/registry/TypeConverterRegistry.java
```diff
@@ -0,0 +1,131 @@
...
+    {
+        Class<?> fromShaded = primitiveToWrapper(from);
+        if (fromShaded.equals(to))
+            return to::cast;
+        if (converters.get(key(fromShaded, to)) == null)
+            return defaultConverter;
+        return (TypeConverter<V>) converters.get(key(fromShaded, to));
+    }
+
+    private static void registerConverters(Map<ConverterKey, TypeConverter<?>> converters)
```
dcapwell, line 66: we have `org.apache.cassandra.config.YamlConfigurationLoader#updateFromMap` which makes sure we always reuse the same input set that was supported in YAML... is there a reason to reinvent this?

dcapwell:
I wonder why go down this route when we could offer a simpler solution.

yaml currently handles everything but validation/listeners, so could decorate the `Properties` with a new validation and listener properties

```
public interface Listener<A, B>
    {
        default void onPreSet(A object, B value) {}
        default void onPostSet(A object, B value) {}
    }

    public static class CompositeListener<A, B> implements Listener<A, B>
    {
        private final List<Listener<A, B>> listeners;

        public CompositeListener(List<Listener<A, B>> listeners)
        {
            this.listeners = listeners;
        }

        @Override
        public void onPreSet(A object, B value)
        {
            for (Listener<A, B> l : listeners)
                l.onPreSet(object, value);
        }

        @Override
        public void onPostSet(A object, B value)
        {
            for (Listener<A, B> l : listeners)
                l.onPostSet(object, value);
        }
    }

    public interface Validator<A, B> extends Listener<A, B>
    {
        void validate(B value);
        default void validate(A object, B value)
        {
            validate(value);
        }

        @Override
        default void onPreSet(A object, B value)
        {
            validate(object, value);
        }
    }

    public static class ListenableProperty<A, B> extends ForwardingProperty
    {
        private final Listener<A, B> listener;
        public ListenableProperty(String name, Property property, Listener<A, B> listener)
        {
            super(name, property);
            this.listener = listener;
        }

        @Override
        public void set(Object o, Object o1) throws Exception
        {
            listener.onPreSet((A) o, (B) o1);
            super.set(o, o1);
            listener.onPostSet((A) o, (B) o1);
        }
    }
```

We need something to hold the `Map<String, Property>, so flexible where this is, but feel like `DatabaseDescriptor` is likely the best place.

With this you only worry about `Property` and don't need to worry about validation/notification as the property was decorated to add such capabilities (and don't pay the cost to check this if nothing validates or listens)...  We also can set this up in `DatabaseDescriptor` so its immutable (we don't need mutability with this), which mostly solves the mapping problem (as DD knows everything, so can build this).

With this, we keep the logic small, 100% align with YAML (can't possibly drift as YAML is what does the string mapping), and solve the "how does the framework" know problem as we push this to `DatabaseDescriptor`, the place that actually knows about this.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -521,20 +549,20 @@ else if (conf.disk_access_mode == Config.DiskAccessMode.mmap_index_only)
             logger.warn("concurrent_replicates has been deprecated and should be removed from cassandra.yaml");
 
         if (conf.networking_cache_size == null)
-            conf.networking_cache_size = new DataStorageSpec.IntMebibytesBound(Math.min(128, (int) (Runtime.getRuntime().maxMemory() / (16 * 1048576))));
+            conf.networking_cache_size = new DataStorageSpec.IntMebibytesBound(Math.min(128, SPACE_UPPER_BOUND_MB));
 
         if (conf.file_cache_size == null)
-            conf.file_cache_size = new DataStorageSpec.IntMebibytesBound(Math.min(512, (int) (Runtime.getRuntime().maxMemory() / (4 * 1048576))));
+            conf.file_cache_size = new DataStorageSpec.IntMebibytesBound(Math.min(512, SPACE_UPPER_BOUND_MB));
```
blerer, line 555: The replaced value do not match the original value.

PR #2303, opened 2023-05-02
CASSANDRA-8720 trunk: Provide offline tool for finding large partitions

adelapena:


doc/modules/cassandra/pages/tools/sstable/sstablepartitions.adoc
```diff
@@ -0,0 +1,174 @@
+= sstabledump
```
driftx: "typo"

doc/modules/cassandra/pages/tools/sstable/sstablepartitions.adoc
```diff
@@ -0,0 +1,174 @@
...
+
+Identifies large partitions of SSTables and outputs the partition size in bytes, row count, cell count, and tombstone count.
+
+You can supply any number of sstables file paths, or directories containing sstables. Each sstable will be analyzed separately.
+
+Cassandra must be stopped before this tool is executed, or unexpected
```
driftx: Is this true?

smiklosovic:
@adelapena please add this tool into `debian/cassandra-tools.install` file as well as into `redhat/cassandra.spec`

doc/modules/cassandra/pages/tools/sstable/sstablepartitions.adoc
adelapena: Fixed, thanks

doc/modules/cassandra/pages/tools/sstable/sstablepartitions.adoc
adelapena: I think it's not. That warning is present on the documentation for all the existing sstable tools. However, I think it's not necessary for read-only tools like the proposed `sstablepartitions`. If that's right, we probably should open a separate ticket to remove the message from those tools (it affects multiple branches).

adelapena: Talking about warnings, I have added a disclaimer saying that the regular output is not meant to be machine-readable, and that scripts should use the `--csv` flag. Maybe that could avoid the problems that we have recently had when changing the output of other tools.

driftx (approved):

PR #2312, opened 2023-05-08
NPE when deserializing malformed collections from client

jonmeredith:
See CASSANDRA-18505

### DIFF
--- a/src/java/org/apache/cassandra/serializers/ListSerializer.java
+++ b/src/java/org/apache/cassandra/serializers/ListSerializer.java
@@ -72,6 +72,8 @@ public class ListSerializer<T> extends CollectionSerializer<List<T>>
             for (int i = 0; i < n; i++)
             {
                 V value = readValue(input, accessor, offset, version);
+                if (value == null)
+                    throw new MarshalException("Not enough bytes to read value in list");
                 offset += sizeOfValue(value, accessor, version);
                 elements.validate(value, accessor);
             }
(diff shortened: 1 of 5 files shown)


src/java/org/apache/cassandra/serializers/ListSerializer.java
```diff
@@ -104,6 +106,8 @@ public <V> List<T> deserializeForNativeProtocol(V input, ValueAccessor<V> access
             {
                 // We can have nulls in lists that are used for IN values
                 V databb = readValue(input, accessor, offset, version);
+                if (databb == null)
+                    throw new MarshalException("Not enough bytes to read value in list");
                 offset += sizeOfValue(databb, accessor, version);
                 if (databb != null)
```
maedhroz, line 113: nit: This check has become unnecessary?

maedhroz: The comment above seems to think we should be able to have nulls?

smiklosovic: Why not to throw that exception from that method? We do not need to throw it explicitly everywhere then.

maedhroz: I think we need to leave the possibility of a null open for lists on a -1 size.

maedhroz: For sets and maps, I would almost have a separate `readNonNullValue()` that does throw internally.

jonmeredith: Went with your suggestion - null in the input seems more likely than corruption so a better, cleaner exception.

maedhroz (approved):
We could always make a small change to `SerDeserTest` to codify our expectations, catch thew new `MarshalException` etc., but I'll leave that up to you.

PR #2318, opened 2023-05-10
CASSANDRA-18511 Add support for JMX in the in-jvm dtest framework

JeetKunDoug:
Additionally, while testing these changes this commit found and fixed some
JDK17 build/run issues.


src/java/org/apache/cassandra/utils/JMXServerUtils.java
```diff
@@ -362,5 +365,24 @@ public String[] list() throws RemoteException {
...
+         * builds, and the superclass never expects to be shut down and therefore doesn't
+         * handle this edge case at all.
+         */
+        @VisibleForTesting
+        public void close() {
+            try
+            {
+                UnicastRemoteObject.unexportObject(this, true);
+            }
+            catch (NoSuchObjectException e)
```
frankgh: maybe rename the variable to indicate it's ignored
```suggestion
            catch (NoSuchObjectException ignored)
```

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -219,6 +219,8 @@
 
     /** This property indicates whether disable_mbean_registration is true */
     IS_DISABLED_MBEAN_REGISTRATION("org.apache.cassandra.disable_mbean_registration"),
+    /** This property indicates if the code is running under the in-jvm dtest framework */
```
frankgh: does the order of the enum matter here? Some enums might be sensitive to the order, so usually new enums are added at the end. It might not be the case here

build.xml
```diff
@@ -263,6 +265,8 @@
         <string>--add-opens java.base/java.math=ALL-UNNAMED</string>
         <string>--add-opens java.base/java.lang.reflect=ALL-UNNAMED</string>
         <string>--add-opens java.base/java.net=ALL-UNNAMED</string>
+
+        <string>--add-opens java.rmi/sun.rmi.transport.tcp=ALL-UNNAMED</string>
```
frankgh, line 269: is this `--add-opens` scoped to the in-jvm dtests only?

build.xml
```diff
@@ -207,6 +207,7 @@
         <string>--add-exports java.management.rmi/com.sun.jmx.remote.internal.rmi=ALL-UNNAMED</string>
         <string>--add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED</string>
         <string>--add-exports java.rmi/sun.rmi.server=ALL-UNNAMED</string>
+        <string>--add-exports java.rmi/sun.rmi.transport.tcp=ALL-UNNAMED</string>
```
frankgh, line 210: is this `--add-opens` scoped to the in-jvm dtests only?

src/java/org/apache/cassandra/utils/MBeanWrapper.java
```diff
@@ -170,6 +186,125 @@ public void unregisterMBean(ObjectName mbeanName, OnException onException)
...
+    {
+        private final MBeanServer mbs;
+        public final UUID id = UUID.randomUUID();
+
+        public InstanceMBeanWrapper(String hostname)
+        {
+            mbs = MBeanServerFactory.createMBeanServer(hostname + "-" + id);
+        }
+
+        public void registerMBean(Object obj, ObjectName mbeanName, OnException onException)
```
frankgh, line 234: can we add the `@Override` annotation to distinguish between internal methods and overridden methods?

src/java/org/apache/cassandra/utils/MBeanWrapper.java
```diff
@@ -170,6 +186,125 @@ public void unregisterMBean(ObjectName mbeanName, OnException onException)
...
+            return mbs;
+        }
+
+        public void close() {
+            mbs.queryNames(null, null).forEach(name -> {
+                try {
+                    if (!name.getCanonicalName().contains("MBeanServerDelegate"))
+                        mbs.unregisterMBean(name);
+                } catch (Throwable e) {
+                    logger.debug("Could not unregister mbean {}", name.getCanonicalName());
```
frankgh, line 289: log at warn maybe?
```suggestion
                    logger.warn("Could not unregister mbean {}", name.getCanonicalName(), e);
```

src/java/org/apache/cassandra/utils/MBeanWrapper.java
```diff
@@ -170,6 +186,125 @@ public void unregisterMBean(ObjectName mbeanName, OnException onException)
...
+        }
+
+        public void setDelegate(MBeanWrapper wrapper) {
+            delegate = wrapper;
+        }
+
+        public MBeanWrapper getDelegate()
+        {
+            return delegate;
+        }
```
frankgh, line 313: new line after brace?
```suggestion
        }

```

src/java/org/apache/cassandra/utils/RMIClientSocketFactoryImpl.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.net.InetAddress;
+import java.net.Socket;
+import java.rmi.server.RMIClientSocketFactory;
+
+/**
+ * This class is used to override the local address the JMX client calculates when trying to connect,
+ * which can otherwise be influences by the system property "java.rmi.server.hostname" in strange and
```
frankgh:
```suggestion
 * which can otherwise be influenced by the system property "java.rmi.server.hostname" in strange and
```

test/distributed/org/apache/cassandra/distributed/impl/INodeProvisionStrategy.java
```diff
@@ -56,6 +56,11 @@ public int nativeTransportPort(int nodeNum)
                     {
                         return 9041 + nodeNum;
                     }
+
+                    public int jmxPort(int nodeNum)
```
frankgh, line 60: `@Override` annotation?

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
```diff
@@ -754,6 +786,146 @@ else if (cluster instanceof Cluster)
...
+
+    private void stopJmx() throws ClassNotFoundException, IllegalAccessException, InvocationTargetException, NoSuchMethodException, IOException, NoSuchFieldException, InterruptedException
+    {
+        if (!config.has(JMX))
+            return;
+        // First, swap the mbean wrapper back to a NoOp wrapper
+        // This prevents later attempts to unregister mbeans from failing in Cassandra code, as we're going to
+        // unregister all of them here
+        ((MBeanWrapper.DelegatingMbeanWrapper)MBeanWrapper.instance).setDelegate(new MBeanWrapper.NoOpMBeanWrapper());
+        try {
```
frankgh: seems like formatting is off for this method, for example 
```suggestion
        try
        {
```

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
```diff
@@ -754,6 +786,146 @@ else if (cluster instanceof Cluster)
...
+    }
+
+    /**
+     * Checks to make sure the endpoint in question was created by the instance classloader
+     * for this instance to prevent us from over-removing items.
+     * @param entry The map entry to be checked
+     * @return true, if the TCPEndpoint in the map entry was created by the InstanceClassloader instance
+     *         created by this Instance. Otherwise, false.
+     */
+    private boolean endpointCreatedByThisClassloader(Map.Entry<TCPEndpoint, LinkedList> entry) {
```
frankgh: formatting?
```suggestion
    private boolean endpointCreatedByThisClassloader(Map.Entry<TCPEndpoint, LinkedList> entry)
    {
```

frankgh (commented):
Great work! I've added some minor comments, mostly around the formatting

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
JeetKunDoug: Thankfully no - this enum is really just a convenience wrapper around system properties, so the ordering doesn't matter (and other folks seem to add things where they make sense vs. at the end as well).

build.xml
JeetKunDoug: Neither is scoped specifically to in-jvm dtests, but they aren't in any production startup scripts either... it's only for IDE support and running with `ant test-run` which is documented as `Not for production use!` so I think it's appropriate to have them here.

src/java/org/apache/cassandra/utils/MBeanWrapper.java
JeetKunDoug: Yeah - none of the other existing ones had `@Override` annotations before so I didn't add them here... I can just add them to everything so it's at least consistent.

JeetKunDoug: So added the `@Override` annotations and looking at the result, it just adds a ton of noise to the file (and the diff) - will discuss in person and we can figure out which way we want to go here.

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
JeetKunDoug: On rebase, it ends up it now needs to be maintained in alphabetical order, so I had to move things around anyway.

src/java/org/apache/cassandra/utils/MBeanWrapper.java
JeetKunDoug: Decided to leave off the `@Override`s for now as it would spam the review with noise. If desired, we could do a follow-up to enforce the code style on files not currently following it.

src/java/org/apache/cassandra/utils/MBeanWrapper.java
JeetKunDoug: Done

src/java/org/apache/cassandra/utils/MBeanWrapper.java
JeetKunDoug: The in-jvm dtest enables debug logging everywhere anyway, and I don't actually think this is a `warn`-level thing as the whole JMX infrastructure gets shut down eventually anyway.

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
```diff
@@ -144,22 +156,30 @@
...
+import org.apache.cassandra.utils.JMXServerUtils;
 import org.apache.cassandra.utils.JVMStabilityInspector;
+import org.apache.cassandra.utils.MBeanWrapper;
+import org.apache.cassandra.utils.RMIClientSocketFactoryImpl;
 import org.apache.cassandra.utils.Throwables;
 import org.apache.cassandra.utils.concurrent.Ref;
 import org.apache.cassandra.utils.logging.LoggingSupportFactory;
 import org.apache.cassandra.utils.memory.BufferPools;
 import org.apache.cassandra.utils.progress.jmx.JMXBroadcastExecutor;
+import sun.rmi.transport.tcp.TCPEndpoint;
```
jonmeredith: should be earlier

jonmeredith (approved):

.build/build-resolver.xml
```diff
@@ -57,7 +57,7 @@
             <remoterepo id="resolver-central" url="${artifact.remoteRepository.central}"/>
             <remoterepo id="resolver-apache" url="${artifact.remoteRepository.apache}"/>
             <!-- Snapshots are not allowed, but for feature branches they may be needed, so uncomment the below to allow snapshots to work -->
-            <!-- <remoterepo id="resolver-apache-snapshot" url="https://repository.apache.org/content/repositories/snapshots" releases="false" snapshots="true" updates="always" checksums="fail" /> -->
+            <remoterepo id="resolver-apache-snapshot" url="https://repository.apache.org/content/repositories/snapshots" releases="false" snapshots="true" updates="always" checksums="fail" />
```
ifesdjeen: Just to make sure we do not forget, this should be reverted before commit.

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -247,6 +247,8 @@
     INTERNODE_EVENT_THREADS("cassandra.internode-event-threads"),
     IO_NETTY_EVENTLOOP_THREADS("io.netty.eventLoopThreads"),
     IO_NETTY_TRANSPORT_ESTIMATE_SIZE_ON_SUBMIT("io.netty.transport.estimateSizeOnSubmit"),
+    /** This property indicates if the code is running under the in-jvm dtest framework */
+    IS_IN_JVM_DTEST("org.apache.cassandra.is_in_jvm_dtest"),
```
ifesdjeen: similar to howe have have `test` prefix, should we have a `dtest` prefix?

src/java/org/apache/cassandra/utils/MBeanWrapper.java
```diff
@@ -36,23 +42,36 @@
...
+
+    static MBeanWrapper getmBeanWrapper()
     {
         if (ORG_APACHE_CASSANDRA_DISABLE_MBEAN_REGISTRATION.getBoolean())
             return new NoOpMBeanWrapper();
 
         String klass = MBEAN_REGISTRATION_CLASS.getString();
         if (klass == null)
-            return new PlatformMBeanWrapper();
+            if (IS_IN_JVM_DTEST.getBoolean())
```
ifesdjeen: nit: brackets for the outer if

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
```diff
@@ -759,6 +789,170 @@ else if (cluster instanceof Cluster)
         initialized = true;
     }
 
+    private void startJmx() throws InterruptedException
```
ifesdjeen: this is more of a nit than an actual comment, but do you think it would be possible to move JMX-related initializers to their own class? I know we already got the jmx util class in .15, but we could also have one cassandra-locally. Most other initializations live in messaging, gossip, commit log, and other subsystems. I understand this is not necessarily matching a pattern here, but maybe we we could have a class under /impl/

ifesdjeen (approved):

src/java/org/apache/cassandra/utils/MBeanWrapper.java
```diff
@@ -119,17 +138,23 @@ static ObjectName create(String mbeanName, OnException onException)
...
+    class NoOpMBeanWrapper implements MBeanWrapper
     {
         public void registerMBean(Object obj, ObjectName mbeanName, OnException onException) {}
         public void registerMBean(Object obj, String mbeanName, OnException onException) {}
         public boolean isRegistered(ObjectName mbeanName, OnException onException) { return false; }
         public boolean isRegistered(String mbeanName, OnException onException) { return false; }
         public void unregisterMBean(ObjectName mbeanName, OnException onException) {}
         public void unregisterMBean(String mbeanName, OnException onException) {}
+        public Set<ObjectName> queryNames(ObjectName name, QueryExp query) {return Collections.emptySet(); }
+        public MBeanServer getMbs() {return null; }
```
frankgh:
```suggestion
        public MBeanServer getMbs() { return null; }
```

src/java/org/apache/cassandra/utils/MBeanWrapper.java
frankgh: should we still log `e` here? or will it create too much output

JeetKunDoug: I don't think it's necessary - it generally happens because it's already been unregistered elsewhere (so it's actually not a huge problem, was more there for, well, debugging), and it would be super-verbose.

src/java/org/apache/cassandra/utils/MBeanWrapper.java
JeetKunDoug: After discussion, we've decided to just add all of the missing single-statement curly braces to this class across all branches as it's really confusing without them at this point.

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
JeetKunDoug: Done - it really helps clean things up as well, so thanks for the suggestion.

src/java/org/apache/cassandra/utils/MBeanWrapper.java
```diff
@@ -36,23 +42,46 @@
...
+        // mbean wrapper even if we start off with no-op, so it can be switched later
+        if (DTEST_IS_IN_JVM_DTEST.getBoolean())
+        {
+            return new DelegatingMbeanWrapper(getmBeanWrapper());
+        }
+
+        return getmBeanWrapper();
+    }
+
+    static MBeanWrapper getmBeanWrapper()
```
ifesdjeen: one more nit: `mBean` vs `MBean` in all other cases in this class.

src/java/org/apache/cassandra/utils/MBeanWrapper.java
```diff
@@ -119,17 +154,23 @@ static ObjectName create(String mbeanName, OnException onException)
         }
     }
 
-    static class NoOpMBeanWrapper implements MBeanWrapper
+    Set<ObjectName> queryNames(ObjectName name, QueryExp query);
+
+    MBeanServer getMbs();
```
ifesdjeen: nit: maybe call it `getMBeanServer`?

PR #2320, opened 2023-05-11
CASSANDRA-18467: ant generate-idea-files support for JDK 17

jakubzytka:
ant generate-idea-files now support JDK 8, JDK 11 and JDK 17. To add support of another JDK the java-jvmargs property must be set for the JDK in question (see how it's done in build.xml for Java 11 and 17)

Other minor, but notable changes are:
- test jvmargs are now added to idea run configurations
- .idea dir and project iml file are first removed and then recreated during ant generate-idea-files

patch by Jakub Żytka ; reviewed by Ekaterina Dimitrova for CASSANDRA-18467

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -188,6 +188,14 @@
       <isset property="arch_x86" />
     </condition>
 
+    <resources id="_jvm8_arg_items">
+        <!-- add necessary args in <string> tags; see _jvm11_arg_items as an example -->
+    </resources>
+    <pathconvert property="_jvm8_args_concat" refid="_jvm8_arg_items" pathsep=" "/>
+    <condition property="java-jvmargs" value="${_jvm8_args_concat}">
+        <equals arg1="${ant.java.version}" arg2="1.8"/>
+    </condition>
+
     <resources id="_jvm11_arg_items">
         <string>-Djdk.attach.allowAttachSelf=true</string>
         <string>-XX:+UseConcMarkSweepGC</string>
@@ -219,8 +227,8 @@
         <string>--add-opens java.base/jdk.internal.util.jar=ALL-UNNAMED</string>
         <string>--add-opens jdk.management/com.sun.management.internal=ALL-UNNAMED</string>
     </resources>
-    <pathconvert property="_jvm_args_concat" refid="_jvm11_arg_items" pathsep=" "/>
-    <condition property="java11-jvmargs" value="${_jvm_args_concat}" else="">
+    <pathconvert property="_jvm11_args_concat" refid="_jvm11_arg_items" pathsep=" "/>
+    <condition property="java-jvmargs" value="${_jvm11_args_concat}">
         <equals arg1="${ant.java.version}" arg2="11"/>
     </condition>
 
@@ -264,8 +272,8 @@
         <string>--add-opens java.base/java.lang.reflect=ALL-UNNAMED</string>
         <string>--add-opens java.base/java.net=ALL-UNNAMED</string>
     </resources>
-    <pathconvert property="_jvm_args_concat2" refid="_jvm17_arg_items" pathsep=" "/>
-    <condition property="java17-jvmargs" value="${_jvm_args_concat2}" else="">
+    <pathconvert property="_jvm17_args_concat" refid="_jvm17_arg_items" pathsep=" "/>
+    <condition property="java-jvmargs" value="${_jvm17_args_concat}">
         <equals arg1="${ant.java.version}" arg2="17"/>
     </condition>
 
@@ -299,10 +307,13 @@
         <string>-Dio.netty.tryReflectionSetAccessible=true</string>
     </resources>
     <pathconvert property="_jvm17_test_arg_items_concat" refid="_jvm17_test_arg_items" pathsep=" "/>
-    <condition property="_std-test-jvmargs11" value="${_jvm11_test_arg_items_concat}" else=" ">
+    <condition property="_std-test-jvmargs" value="${_jvm8_test_arg_items_concat}">
+        <equals arg1="${ant.java.version}" arg2="1.8"/>
+    </condition>
+    <condition property="_std-test-jvmargs" value="${_jvm11_test_arg_items_concat}">
             <equals arg1="${ant.java.version}" arg2="11"/>
     </condition>
-    <condition property="_std-test-jvmargs17" value="${_jvm17_test_arg_items_concat}" else=" ">
+    <condition property="_std-test-jvmargs" value="${_jvm17_test_arg_items_concat}">
         <equals arg1="${ant.java.version}" arg2="17"/>
     </condition>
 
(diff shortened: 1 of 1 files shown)


build.xml
```diff
@@ -1808,11 +1816,21 @@
     <option name="ADDITIONAL_OPTIONS_STRING" value="--add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED" />
   </component>
 </project>]]></echo>
-      <echo>"IDE configuration updated for use with JDK11"</echo>
+      <echo>"IDE configuration updated for use with JDK${ant.java.version}"</echo>
+      <echo>Please note that `ant realclean` does *NOT* remove IntelliJ-specific files</echo>
+      <echo/>
+      <echo>Warning! If you observer strange behavior in IntelliJ please</echo>
```
smiklosovic, line 1822: observer -> observe

build.xml
```diff
@@ -1808,11 +1816,21 @@
     <option name="ADDITIONAL_OPTIONS_STRING" value="--add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED" />
   </component>
 </project>]]></echo>
-      <echo>"IDE configuration updated for use with JDK11"</echo>
+      <echo>"IDE configuration updated for use with JDK${ant.java.version}"</echo>
+      <echo>Please note that `ant realclean` does *NOT* remove IntelliJ-specific files</echo>
```
smiklosovic, line 1820: would it be possible to have some system property which would clean IDEA-specific files too?

Does this mean that `ant realclean` in trunk would be different from `ant realclean` in older branches? Is this something we have changed here or you are just explaining how it always was?

I think if we change the behavior so it is not removed on realclean, we should backport this to older branches as well. Stuff like this should work across all branches just same.

build.xml
```diff
@@ -1808,11 +1816,21 @@
...
+      <echo>verify that the SDK that is being used is indeed version ${ant.java.version}.</echo>
+      <echo>The used JDK version can be verified in Project Structure/Project Setting/Project</echo>
+      <echo>and Project Structure/Platform Setting/SDKs</echo>
+      <echo>In particular it is worth checking if the used JDK, typically named "${ant.java.version}",</echo>
+      <echo>indeed points to a java ${ant.java.version} installation.</echo>
   </target>
 
   <!-- Generate IDEA project description files -->
   <target name="generate-idea-files" depends="init,resolver-dist-lib,gen-cql3-grammar,generate-jflex-java,createVersionPropFile" description="Generate IDEA files">
+    <delete dir=".idea"/>
```
smiklosovic, line 1832: Yeah, so this would be cool to add to older branches as well if we choose that path. It should be consistent everyhere.

smiklosovic: BTW, what about an ant target for cleanup of these files? Like "ant clean-idea-files"? Now, if I generate them, there is no way to clean it except manually removing it. I would make generate-idea-files dependent on clean-idea-files. So we do have both target for cleanup and it will be reused on (re)generation.

michaelsembwever: I disagree with the need for older branches to implement new behaviour, or for a separate clean target.

realclean does not change its behaviour here. The only change is that `generate-idea-files` in trunk now ensures only newly generated files are in the .idea/ folder.  Older branches work as they always did, regardless if you're now switching from trunk.

build.xml
michaelsembwever: see comment in https://github.com/apache/cassandra/pull/2320#discussion_r1246637341

michaelsembwever:
updated patch here: https://github.com/apache/cassandra/compare/trunk...thelastpickle:cassandra:mck/18467/trunk

build.xml
ekaterinadimitrova2: Agreed here with @michaelsembwever

smiklosovic: @michaelsembwever technically, we still do not have a way how to clean idea files. realclean is not removing what generate-idea-files generated. (at least not in this branch, right?) 

That being said, why would one ever need such target anyway? It is not like we are switching from Netbeans to IDEA and back frequently so we need to have _a dedicated target for that_. At least there is less clutter. People just call that target once and that's it. I do not see any use case for having a separate target to clean it. People can just remove the files if they switch to Netbeans or whatever else once in 5 years.

jakubzytka: > People just call that target once and that's it. 

This assumes that you work with only one JDK in one repo.
I admit this is how *I* do work, but I know that others do not - and that they switch the JDK version. Thus they need to re-generate idea files.

ekaterinadimitrova2: Personal experience, working on upgrading to JDK17 I very often have to switch between JDKs
And I am sure there will be more examples and more future upgrades

michaelsembwever: When is the need to remove the .idea folder when you are not going to call generate-idea-files ? 

(delete is kept inside generate-idea-files)

ekaterinadimitrova2: `When is the need to remove the .idea folder when you are not going to call generate-idea-files ?`
I am not aware of such a need.

michaelsembwever: then we are good?

smiklosovic: I find worktree very convenient here. There is no need to switch jdks at all.

I have `.sdkmanrc` is each worktree

    20:53 $ git worktree list
    /home/fermat/dev/cassandra/cassandra-instaclustr/cassandra            e9d2802953 [CASSANDRA-18467]
    /home/fermat/dev/cassandra/cassandra-instaclustr/cassandra-3.0        4ea7bb25b4 [cassandra-3.0]
    /home/fermat/dev/cassandra/cassandra-instaclustr/cassandra-3.11       d3f29c2602 [CASSANDRA-14204]
    /home/fermat/dev/cassandra/cassandra-instaclustr/cassandra-4.0        b0f1ef9832 [CASSANDRA-14204-4.0]
    /home/fermat/dev/cassandra/cassandra-instaclustr/cassandra-4.1        c3327855e4 [cassandra-4.1]
    /home/fermat/dev/cassandra/cassandra-instaclustr/cassandra-trunk-j11  bbe17c23f6 [CASSANDRA-18467-j11]
    /home/fermat/dev/cassandra/cassandra-instaclustr/cassandra-trunk-j17  8563757422 [trunk-j17]

So I have 3 separate IDEA windows for each trunk, trunk on j11 and trunk on j17.

Then in each dir I have .sdkmanrc like

    ✔ ~/dev/cassandra/cassandra-instaclustr/cassandra-trunk-j17 [trunk-j17 L|✔] 
    20:54 $ cat .sdkmanrc 
    java=17.0.7-tem


or 

    ✔ ~/dev/cassandra/cassandra-instaclustr/cassandra-trunk-j11 [CASSANDRA-18467-j11 L|✔] 
    20:55 $ cat .sdkmanrc 
    java=11.0.19-tem


and I have .sdkmanrc in `~/$HOME/.gitignore` as well as this in ~/$HOME/.sdkrcman/etc/config

sdkman_auto_env=true

That means that as I am switching the worktree directories, Java invoked in that Cassandra project root dir is changed dynamically as well upon every "cd".

This was a long explanation why I do not need to have a target which cleans generated files for IDEA - I never regenerate them. And currently that directory is clean upon each generation. So really, why we need a target? In my case never.

I was thinking that I actually bring some benefit to a user by clean target but is there really any if the dir it is going to generate it into is wiped out anyway?

PR #2321, opened 2023-05-11
CASSANDRA-18500: Add guardrail for partition size

adelapena:
The patch adds a new guardrail for partition size that is checked when writing a sstable (flush, compaction, etc.)

Triggering the failure threshold will emit an ERROR log message and a diagnostic event, but it won't abort the sstable write operation.

This guardrail replaces the `compaction_large_partition_warning_threshold` property, which is now marked as deprecated. Both checks will exist until we remove `compaction_large_partition_warning_threshold` on the next major. The reason for keeping both properties and not just renaming `compaction_large_partition_warning_threshold` to `partition_size_warn_threshold` is that the behaviour, log messages, etc. are slightly different.

It's worth mentioning that the existing `compaction_large_partition_warning_threshold`:
* Isn't actually triggered only on compaction, but on any sstable write.
* Cannot be dynamically updated.
* Doesn't emit any kind of diagnostic events.

### DIFF
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -1541,7 +1541,8 @@ batch_size_fail_threshold: 50KiB
 # Log WARN on any batches not of type LOGGED than span across more partitions than this limit
 unlogged_batch_across_partitions_warn_threshold: 10
 
-# Log a warning when compacting partitions larger than this value
+# Log a warning when compacting partitions larger than this value.
+# As of Cassandra 5.0, this property is deprecated in favour of partition_size_warn_threshold.
 compaction_large_partition_warning_threshold: 100MiB
 
 # Log a warning when writing more tombstones than this value to a partition
@@ -1812,6 +1813,15 @@ drop_compact_storage_enabled: false
 # Guardrail to warn about or reject write consistency levels. By default, all consistency levels are allowed.
 # write_consistency_levels_warned: []
 # write_consistency_levels_disallowed: []
+#
+# Guardrail to warn or fail when writing partitions larger than threshold, expressed as 100MiB, 1GiB, etc.
+# The guardrail is only checked when writing sstables (flush and compaction), and exceeding the fail threshold on that
+# moment will only log an error message, without interrupting the operation.
+# This operates on a per-sstable basis, so it won't detect a large partition if it is spread across multiple sstables.
+# The warning threshold replaces the deprecated config property compaction_large_partition_warning_threshold.
+# The two thresholds default to null to disable.
+# partition_size_warn_threshold:
+# partition_size_fail_threshold:
 #
 # Guardrail to warn or fail when writing column values larger than threshold.
 # This guardrail is only applied to the values of regular columns because both the serialized partitions keys and the
--- a/src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
+++ b/src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
@@ -469,6 +469,33 @@ public interface GuardrailsMBean
     void setWriteConsistencyLevelsDisallowedCSV(String consistencyLevels);
 
     /**
+     * @return The threshold to warn when encountering partitions larger than threshold, as a string formatted as in,
+     * for example, {@code 10GiB}, {@code 20MiB}, {@code 30KiB} or {@code 40B}. A {@code null} value means disabled.
+     */
+    @Nullable
+    String getPartitionSizeWarnThreshold();
+
+    /**
+     * @return The threshold to fail when encountering partitions larger than threshold, as a string formatted as in,
+     * for example, {@code 10GiB}, {@code 20MiB}, {@code 30KiB} or {@code 40B}. A {@code null} value means disabled.
+     * Triggering a failure emits a log message and a diagnostic  event, but it doesn't throw an exception interrupting
+     * the offending sstable write.
+     */
+    @Nullable
+    String getPartitionSizeFailThreshold();
+
+    /**
+     * @param warnSize The threshold to warn when encountering partitions larger than threshold, as a string formatted
+     *                 as in, for example, {@code 10GiB}, {@code 20MiB}, {@code 30KiB} or {@code 40B}.
+     *                 A {@code null} value means disabled.
+     * @param failSize The threshold to fail when encountering partitions larger than threshold, as a string formatted
+     *                 as in, for example, {@code 10GiB}, {@code 20MiB}, {@code 30KiB} or {@code 40B}.
+     *                 A {@code null} value means disabled. Triggering a failure emits a log message and a diagnostic
+     *                 event, but it desn't throw an exception interrupting the offending sstable write.
+     */
+    void setPartitionSizeThreshold(@Nullable String warnSize, @Nullable String failSize);
+
+    /**
      * @return The threshold to warn when encountering column values larger than threshold, as a string  formatted as
      * in, for example, {@code 10GiB}, {@code 20MiB}, {@code 30KiB} or {@code 40B}. A {@code null} value means disabled.
      */
--- a/src/java/org/apache/cassandra/io/sstable/format/SortedTableWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/format/SortedTableWriter.java
@@ -213,6 +213,7 @@ public abstract class SortedTableWriter<P extends SortedTablePartitionWriter> ex
 
         long endPosition = dataWriter.position();
         long rowSize = endPosition - partitionWriter.getInitialPosition();
+        guardPartitionSize(key, rowSize);
         maybeLogLargePartitionWarning(key, rowSize);
         maybeLogManyTombstonesWarning(key, metadataCollector.totalTombstones);
         metadataCollector.addPartitionSizeInBytes(rowSize);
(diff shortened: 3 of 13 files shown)


conf/cassandra.yaml
```diff
@@ -1804,6 +1805,15 @@ drop_compact_storage_enabled: false
 # write_consistency_levels_warned: []
 # write_consistency_levels_disallowed: []
 #
+# Guardrail to warn or fail when writing partitions larger than threshold.
+# The guardrail is only checked when writing sstables (flush, compaction, streaming, etc.), and exceeding the fail
+# threshold on that moment will only log an error message, without interrupting the operation.
+# This operates on a per-sstable basis, so it won't detect a large partition if it is spread across multiple sstables.
+# The warning threshold replaces the deprecated config property compaction_large_partition_warning_threshold.
+# The two thresholds default to null to disable.
```
bereng, line 1822: Can you add a few valid examples to see syntax/units? 100MiB, 100MB, 10KB,...

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -468,6 +468,30 @@
...
 
+    /**
+     * @return The threshold to warn when encountering partitions larger than threshold, as a string  formatted as in,
+     * for example, {@code 10GiB}, {@code 20MiB}, {@code 30KiB} or {@code 40B}. A {@code null} value means disabled.
+     */
+    @Nullable
+    String getPartitionSizeWarnThreshold();
+
+    /**
+     * @return The threshold to fail when encountering partitions larger than threshold, as a string  formatted as in,
```
bereng: I would reword all the 'fail' variant comments to highlight the operation won't fail and it will only get logged at ERROR level. Otherwise it's confusing imo.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailPartitionSizeTest.java
```diff
@@ -0,0 +1,140 @@
...
+import static java.nio.ByteBuffer.allocate;
+
+/**
+ * Tests the guardrail for the size of partition size, {@link Guardrails#partitionSize}.
+ * <p>
+ * This test only includes the activation of the guardrail during sstable writes, focusing on the emmitted log messages.
+ * The tests for config, client warnings and diagnostic events are in
+ * {@link org.apache.cassandra.db.guardrails.GuardrailPartitionSizeTest}.
+ */
+public class GuardrailPartitionSizeTest extends GuardrailTester
```
bereng, line 40: Do you test anywhere the thresholds can be dynamically changed?

conf/cassandra.yaml
```diff
@@ -1541,7 +1541,8 @@ batch_size_fail_threshold: 50KiB
 # Log WARN on any batches not of type LOGGED than span across more partitions than this limit
 unlogged_batch_across_partitions_warn_threshold: 10
 
-# Log a warning when compacting partitions larger than this value
+# Log a warning when compacting partitions larger than this value.
+# As of Cassandra 5.0, this property is deprecated in favour of partition_size_warn_threshold.
```
Maxwell-Guo, line 1545: can change to "5.0 and later versions ", I think this will be more  rigorous

conf/cassandra.yaml
Maxwell-Guo: My question is : should we regardless of the stream situation? Because the stream data is generated by flush and compaction or logging is enough ?

src/java/org/apache/cassandra/io/sstable/format/SortedTableWriter.java
```diff
@@ -324,6 +325,19 @@ protected FileHandle openDataFile(long lengthOverride, StatsMetadata statsMetada
         return dataFile;
     }
 
+    private void guardPartitionSize(DecoratedKey key, long rowSize)
+    {
+        if (Guardrails.partitionSize.triggersOn(rowSize, null))
+        {
+            String what = String.format("%s.%s:%s on sstable %s",
```
Maxwell-Guo, line 332: our experience is we prefer metadata.keyspace +    metadata.name +                              metadata().partitionKeyType.getString(key.getKey()) +   getFilename() to contact these strings than String.format. 
as String.format  will cost more time ,But it seems cassandra just use String.format  all time.So this comment can be ignored.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailPartitionSizeTest.java
adelapena: That is tested on the utest, whereas this dtest is just focused on verifying the not-redacted messages. But I have nevertheless added a new `testPartitionSizeWithDynamicUpdate`.

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
adelapena: I guess the meaning of "fail" on all background guardrails like this one is logging an error and emit a failure event. Not very sure how to rename, but I have added a line telling that it will not throw an exception to avoid stopping the write.

conf/cassandra.yaml
adelapena: I think that "as of" means "starting from", or versions >= 5.0, so the "and later versions" part seems redundant. Also, since it's a deprecation I think it wouldn't make sense for it to not be deprecated in later versions.

src/java/org/apache/cassandra/io/sstable/format/SortedTableWriter.java
adelapena: Indeed `String.format` is everywhere. In this case, the message is only built when the guardrail is actually triggered, which shouldn't be very often.

conf/cassandra.yaml
adelapena: > Can you add a few valid examples to see syntax/units? 100MiB, 100MB, 10KB,...

Yes, I have added a couple of examples about the expected values, like the ones on the JMX methods.

> My question is : should we regardless of the stream situation? Because the stream data is generated by flush and compaction or logging is enough ?

I'm not sure I understand what we should do and what you mean by the stream situation. Tables streamed from other nodes would have triggered the guardrail in their node of origin. We probably don't want to scan them again in the receiver node as they arrive to avoid the performance cost of scanning them. Also, large partitions in a streamed table can be later be detected by the guardrail when those partitions are compacted.

conf/cassandra.yaml
Maxwell-Guo: ok

src/java/org/apache/cassandra/io/sstable/format/SortedTableWriter.java
Maxwell-Guo: yes ,I found C* use String.format almost everywhere. So I said this comment can be ignored.

conf/cassandra.yaml
Maxwell-Guo: These sstables that are going to stream to other nodes are generated by flush or compaction , and we may just need to use the guardrail to detecte at flush or compaction stage. 

The reveiver  or the sending node do not need to use the guardrail to detecte at stream stage. 

That is what I mean.

bereng: Thx on the examples.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailPartitionSizeTest.java
bereng: +1

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
bereng: +1

bereng:
I don't have further comments so far besides that we're missing a final CI run

conf/cassandra.yaml
adelapena: > These sstables that are going to stream to other nodes are generated by flush or compaction , and we may just need to use the guardrail to detecte at flush or compaction stage.

I see, I have updated the comment to exclude streaming.

adelapena:
CI is running here:
* https://app.circleci.com/pipelines/github/adelapena/cassandra/2906/workflows/e322a36b-b38a-4387-9464-4eeede1edbd9
* https://app.circleci.com/pipelines/github/adelapena/cassandra/2906/workflows/c38b9241-4dd0-4180-b7a8-b006ba52c8cc

Maxwell-Guo:
+1 on this patch

bereng:
It seems circle is having issues and the CI runs don't look too nice. Could we re-trigger them please?

adelapena:
Yes, I've already done it:
* https://app.circleci.com/pipelines/github/adelapena/cassandra/2906/workflows/9ce9fb2a-fe2a-4619-ba1c-ce4d29a1016c
* https://app.circleci.com/pipelines/github/adelapena/cassandra/2906/workflows/d28b205c-1b51-49aa-959a-948be3fa06eb

All green except for a consistent failure on `GetFullQueryLogTest` that has just been introduced in trunk by [CASSANDRA-18550](https://issues.apache.org/jira/browse/CASSANDRA-18550)

bereng:
+1

PR #2322, opened 2023-05-11
CASSANDRA-18511 Add support for JMX in the in-jvm dtest framework (3.11)

JeetKunDoug:


src/java/org/apache/cassandra/utils/MBeanWrapper.java
```diff
@@ -33,62 +38,144 @@
...
+    {
+        if (Boolean.getBoolean(IS_DISABLED_MBEAN_REGISTRATION))
+            return new NoOpMBeanWrapper();
+
+        String klass = System.getProperty(MBEAN_REGISTRATION_CLASS);
+        if (klass == null)
+            if (Boolean.getBoolean(IS_IN_JVM_DTEST))
+                return new NoOpMBeanWrapper();
+            else
+                return new PlatformMBeanWrapper();
```
jonmeredith: nit: my personal preference would be to include the outer `if` braces `{}` but don't think this goes against the style guide.

test/distributed/org/apache/cassandra/distributed/test/jmx/JMXFeatureTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    private void testInstance(Set<String> instancesContacted, IInvokableInstance instance) throws IOException
+    {
+        // NOTE: At some point, the hostname of the broadcastAddress can be resolved
+        // and then the `getHostString`, which would otherwise return the IP address,
+        // starts returning `localhost` - use `.getAddress().getHostAddress()` to work around this.
+        String jmxHost = instance.config().broadcastAddress().getAddress().getHostAddress();
+        int jmxPort = instance.config().jmxPort();
+        String url = String.format(JMX_SERVICE_URL_FMT, jmxHost, jmxPort);
+        try (JMXConnector jmxc = JMXConnectorFactory.connect(new JMXServiceURL(url), null))
```
jonmeredith: is it worth making a convenience method that takes an IInstanceConfig and returns the `JMXConnector`. Is there any test where you wouldn't do those same four statements to make the connector?

test/distributed/org/apache/cassandra/distributed/test/jmx/JMXGetterCheckTest.java
```diff
@@ -0,0 +1,138 @@
...
+            {
+                AssertionError root = new AssertionError();
+                errors.forEach(root::addSuppressed);
+                throw root;
+            }
+        }
+    }
+
+    /**
+     * This class is meant to make new errors easier to read, by adding the JMX endpoint, and cleaning up the unneded JMX/Reflection logic cluttering the stacktrace
```
jonmeredith:
```suggestion
     * This class is meant to make new errors easier to read, by adding the JMX endpoint, and cleaning up the unneeded JMX/Reflection logic cluttering the stacktrace
```

src/java/org/apache/cassandra/utils/MBeanWrapper.java
JeetKunDoug: Yeah - I get dinged for _including_ them so I try not to do it if it's only single-lined.

test/distributed/org/apache/cassandra/distributed/test/jmx/JMXFeatureTest.java
JeetKunDoug: So would this be worth putting into the dtest API itself in that case? Unfortunately would require another round of review/release of the dtest API but everything you need to get a connection is available in the API.

Cleanup is probably easier if we just do it in the Cassandra code, as you could, in theory, stash the JMXConnector on the instance itself, and then close it when the instance is closed, which would be easier to enforce.

src/java/org/apache/cassandra/utils/MBeanWrapper.java
JeetKunDoug: Rereading the style guide, it does specifically say they _may_ be elided so I'm actually going to put them in, as I generally agree with you on this.

Actually, the entire rest of the file has elided the braces for single-lines, so I'm going to leave it alone.

test/distributed/org/apache/cassandra/distributed/test/jmx/JMXFeatureTest.java
JeetKunDoug: We pulled this into the dtest framework.

test/distributed/org/apache/cassandra/distributed/test/jmx/JMXGetterCheckTest.java
JeetKunDoug: Was actually existing code (in the other branches), but I'll fix the typo anyway while I'm at it.

jonmeredith (approved):

PR #2331, opened 2023-05-12
CASSANDRA-18525 Adding keyspace column to clients virtual table

nvharikrishna:
Added new column `keyspace_name` (used by the client) to clients virtual table.

[CASSANDRA-18525](https://issues.apache.org/jira/browse/CASSANDRA-18525)

### DIFF
--- a/test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
+++ b/test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
@@ -1,0 +1,124 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.db.virtual;
+
+import java.io.IOException;
+import java.net.InetAddress;
+import java.util.List;
+
+import com.google.common.collect.ImmutableList;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.Row;
+import com.datastax.driver.core.Session;
+import org.apache.cassandra.ServerTestUtils;
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.cql3.CQLTester;
+import org.apache.cassandra.service.EmbeddedCassandraService;
+
+public class ClientsTableKeyspaceColTest extends CQLTester
+{
+
+    private static Cluster cluster;
+    private static EmbeddedCassandraService cassandra;
+    private static final String KS_NAME = "vts";
+
+    @BeforeClass
+    public static void setupClass() throws IOException
+    {
+        cassandra = ServerTestUtils.startEmbeddedCassandraService();
+
+        cluster = Cluster.builder().addContactPoint("127.0.0.1")
+                         .withPort(DatabaseDescriptor.getNativeTransportPort())
+                         .build();
+
+        ClientsTable table = new ClientsTable(KS_NAME);
+        VirtualKeyspaceRegistry.instance.register(new VirtualKeyspace(KS_NAME, ImmutableList.of(table)));
+    }
+
+    @AfterClass
+    public static void tearDown()
+    {
+        if (cluster != null)
+            cluster.close();
+        if (cassandra != null)
+            cassandra.stop();
+    }
+
+    @Test
+    public void testWithoutConnectingToKeyspace()
+    {
+        try (Session session = cluster.connect())
+        {
+            List<Row> rows = session.execute("SELECT * from " + KS_NAME + ".clients").all();
+            Assert.assertTrue("At least one client should be returned.", rows.size() > 0);
+            for (Row r : rows)
+            {
+                // No keyspace is specifed while connecting. 'keyspace' column should be null.
+                Assert.assertNull(r.getString("keyspace_name"));
+            }
+        }
+    }
+
+    @Test
+    public void testChangingKeyspace()
+    {
+        String keyspace1 = "system_distributed";
+        String keyspace2 = "system_auth";
+        try (Session session = cluster.connect(keyspace1))
+        {
+
+            InetAddress poolConnection = null;
+            int port = -1;
+            List<Row> rows = session.execute("SELECT * from " + KS_NAME + ".clients").all();
+            for (Row r : rows)
+            {
+                // Keyspace is used for pool connection only (control connection is not using keyspace).
+                // Using keyspace != null as a hint to identify a pool connection as we can't identify
+                // control connection based on information in this table.
+                String keyspace = r.getString("keyspace_name");
+                if (keyspace1.equals(keyspace))
+                {
+                    poolConnection = r.getInet("address");
+                    port = r.getInt("port");
+                }
+            }
+
+            Assert.assertNotEquals(-1, port);
+            Assert.assertNotNull(poolConnection);
+
+            session.execute("USE " + keyspace2);
+
+            String actualKeyspace = null;
+            for (Row r : session.execute("SELECT * from " + KS_NAME + ".clients").all())
+            {
+                if (poolConnection.equals(r.getInet("address")) && port == r.getInt("port"))
+                {
+                    actualKeyspace = r.getString("keyspace_name");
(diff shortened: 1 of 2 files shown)


test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
```diff
@@ -0,0 +1,124 @@
...
+
+import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.Row;
+import com.datastax.driver.core.Session;
+import org.apache.cassandra.ServerTestUtils;
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.cql3.CQLTester;
+import org.apache.cassandra.service.EmbeddedCassandraService;
+
+public class ClientsTableKeyspaceColTest extends CQLTester
```
smiklosovic: @nvharikrishna could you please add this to `ClientsTableTest` class in this package? Do we really need a new file for it? That one extends `CQLTester` too.

nvharikrishna: I wanted to control connection/session establishment so that keyspace_name can be verified just for that user. CQLTester doesn't have a public/protected method to get cluster and session instances. I did not find CQLTester using `cluster.connect(keyspace)`. So I created a separate test class with cluster and session instances explicitly constructed. It cannot be done with `ClientsTableTest` as everything done through CQLTester.setUpClass(). I realised that this test doesn't have to extend CQLTester. Thanks for pointing it out. Cleared it.

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
```diff
@@ -0,0 +1,123 @@
...
+        if (cassandra != null)
+            cassandra.stop();
+    }
+
+    @Test
+    public void testWithoutConnectingToKeyspace()
+    {
+        try (Session session = cluster.connect())
+        {
+            List<Row> rows = session.execute("SELECT * from " + KS_NAME + ".clients").all();
```
smiklosovic, line 82: @nvharikrishna could you please put this into the `for` loop instead of `rows`? Your very last `for` loop in this test class in the other test method is putting it directly into the loop while this does not (neither the first loop in `testChangingKeyspace` test). I would like to see it consistent everywhere.

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
```diff
@@ -0,0 +1,123 @@
...
+            Assert.assertTrue("At least one client should be returned.", rows.size() > 0);
```
smiklosovic: @nvharikrishna could you please import all these assertion methods statically?

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
```diff
@@ -0,0 +1,123 @@
...
+            for (Row r : rows)
+            {
```
smiklosovic: If you have just one statement in the body of the for loop, we tend to just omit the braces.

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
```diff
@@ -0,0 +1,123 @@
...
+        try (Session session = cluster.connect(keyspace1))
+        {
+
+            InetAddress poolConnection = null;
+            int port = -1;
+            List<Row> rows = session.execute("SELECT * from " + KS_NAME + ".clients").all();
+            for (Row r : rows)
+            {
+                // Keyspace is used for pool connection only (control connection is not using keyspace).
+                // Using keyspace != null as a hint to identify a pool connection as we can't identify
```
smiklosovic: I do not see actually using `keyspace != null` anywhere.

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
```diff
@@ -0,0 +1,123 @@
...
+                // control connection based on information in this table.
+                String keyspace = r.getString("keyspace_name");
+                if (keyspace1.equals(keyspace))
+                {
+                    poolConnection = r.getInet("address");
+                    port = r.getInt("port");
```
smiklosovic, line 108: Is not it true that you can `break` from this loop once we enter this `if` and `poolConnection` and `port` are set?

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
```diff
@@ -0,0 +1,123 @@
...
+                }
+            }
+
+            Assert.assertNotEquals(-1, port);
+            Assert.assertNotNull(poolConnection);
+
+            session.execute("USE " + keyspace2);
+
+            String actualKeyspace = null;
+            for (Row r : session.execute("SELECT * from " + KS_NAME + ".clients").all())
```
smiklosovic, line 119: for (Row r ...
        if (pooConnection ....
        {
            actualKeyspace = ....
            break;
        }

Please correct me if I am wrong.

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
```diff
@@ -0,0 +1,123 @@
...
+                         .build();
+
+        ClientsTable table = new ClientsTable(KS_NAME);
+        VirtualKeyspaceRegistry.instance.register(new VirtualKeyspace(KS_NAME, ImmutableList.of(table)));
+    }
+
+    @AfterClass
+    public static void tearDown()
+    {
+        if (cluster != null)
```
smiklosovic: what if `cluster.close()` throws? I think it can throw some runtime exception. That means that it will never get to actually stop `cassandra`. I would rewrite it to 

        try
        {
            if (cluster != null)
                cluster.close();
        }
        finally
        {
            if (cassandra != null)
                cassandra.stop();
        }

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
```diff
@@ -0,0 +1,123 @@
...
+                    port = r.getInt("port");
+                }
+            }
+
+            Assert.assertNotEquals(-1, port);
+            Assert.assertNotNull(poolConnection);
+
+            session.execute("USE " + keyspace2);
+
+            String actualKeyspace = null;
```
smiklosovic: I would rename this to `usedKeyspace`.

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
nvharikrishna: done

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
nvharikrishna: done

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
nvharikrishna: I am asserting that at least one row is returned in the next statement so assigned the results to rows. Changed it at other places.

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
nvharikrishna: done

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
nvharikrishna: done

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
nvharikrishna: done

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
nvharikrishna: done

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
nvharikrishna: Updated the comment

nvharikrishna (commented):
@smiklosovic addressed your reviews comments

test/unit/org/apache/cassandra/db/virtual/ClientsTableKeyspaceColTest.java
```diff
@@ -0,0 +1,131 @@
...
+                String keyspace = r.getString("keyspace_name");
+                if (keyspace1.equals(keyspace))
+                {
+                    poolConnection = r.getInet("address");
+                    port = r.getInt("port");
+                    break;
+                }
+            }
+
+            Assert.assertNotEquals(-1, port);
```
smiklosovic, line 113: @nvharikrishna could you please import also these two? Otherwise I think we are done.

smiklosovic: Will be done on commit.

michaelsembwever (approved):

PR #2333, opened 2023-05-14
CASSANDRA-18305 trunk Enhance nodetool compactionstats with existing MBean metrics

mghildiy:
```
Enhance nodetool compactionstats with existing MBean metrics

Print some more metrics related to compaction:

- Compactions completed, and different meter rates
- Concurrent compactors
- compactions completed
- data compacted
- compactions aborted
- compactions reduced

patch by @mghildiy

```

The [Cassandra Jira](https://issues.apache.org/jira/browse/CASSANDRA-18305)


src/java/org/apache/cassandra/tools/nodetool/CompactionStats.java
```diff
@@ -53,29 +60,81 @@ public class CompactionStats extends NodeToolCmd
...
-            }
+                numTotalPendingTasks += tableEntry.getValue();
         }
-        out.println();
-        reportCompactionTable(cm.getCompactions(), probe.getCompactionThroughputBytes(), humanReadable, vtableOutput, out);
+
+        return numTotalPendingTasks;
+    }
+
+    private static int numConcurrentCompactorsConfigured(NodeProbe probe)
```
driftx: We don't really need this method, we can call probe directly in the single usage there is of it.

src/java/org/apache/cassandra/tools/nodetool/CompactionStats.java
```diff
@@ -53,29 +60,81 @@ public class CompactionStats extends NodeToolCmd
...
+        toPrint.append(LINE_SEPARATOR);
+        toPrint.append(String.format("\tMean rate:    %s/second", formatter.format(totalCompactionsCompletedMetrics.getMeanRate())));
+        toPrint.append(LINE_SEPARATOR);
+
+        return toPrint.toString();
+    }
+
+    private static String compactionThroughPutStats(NodeProbe probe)
+    {
+        Config conf = DatabaseDescriptor.loadConfig();
```
driftx: Nodetool should work against any node; we shouldn't import DD at all and instead get this information from the probe.

src/java/org/apache/cassandra/tools/nodetool/CompactionStats.java
```diff
@@ -53,29 +60,81 @@ public class CompactionStats extends NodeToolCmd
...
+        double compactionTPConfigured = conf.compaction_throughput.toMebibytesPerSecond();
+        double compactionTPActual = probe.getCompactionThroughputMebibytesAsDouble();
+        double percentage = (compactionTPActual / compactionTPConfigured) * 100;
```
driftx: Need the handle the case when the configured throughput is zero.

driftx (changes requested):

src/java/org/apache/cassandra/tools/nodetool/CompactionStats.java
mghildiy: NodeProbe.getCompactionThroughputMebibytesAsDouble too reads from DatabaseDescriptor(via StorageService bean). Not sure what should be the correct source of data here.

driftx: Yes, but over JMX from the node connected, not the local value, so it will be correct.

mghildiy: So issue here is that I have to get hold of both configured value and actual value. I was under improession that NodeProbe.getCompactionThroughputMebibytesAsDouble would give me actual value.

driftx: There isn't a running actual throughtput value available, you'll need to expose compactionRateLimiter.getRate in CompactionManager to JMX.

Mmuzaf:
@driftx @mghildiy 

I noticed that `CompactionStatsTest#testCompactionStats` checks the output of the command. Should we also improve the test coverage to be sure that new metrics are present in the command output?

mghildiy:
> @driftx @mghildiy
> 
> I noticed that `CompactionStatsTest#testCompactionStats` checks the output of the command. Should we also improve the test coverage to be sure that new metrics are present in the command output?

@Mmuzaf, I would look into it.

mghildiy:
Updated test.

src/java/org/apache/cassandra/tools/nodetool/CompactionStats.java
```diff
@@ -53,29 +58,82 @@ public class CompactionStats extends NodeToolCmd
...
+    }
+
+    private static String pendingTasksAndConcurrentCompactorsStats(NodeProbe probe)
+    {
         Map<String, Map<String, Integer>> pendingTaskNumberByTable =
             (Map<String, Map<String, Integer>>) probe.getCompactionMetric("PendingTasksByTableName");
-        int numTotalPendingTask = 0;
+        StringBuffer toPrint = new StringBuffer();
+        toPrint.append(String.format("%s concurrent compactors, %s pending tasks", probe.getConcurrentCompactors()
+        , numPendingTasks(pendingTaskNumberByTable)));
```
Maxwell-Guo: I think there is a problem of code format
line 75 should have some space before ',' , so that In this way “，”  can be aligned with "%$"

src/java/org/apache/cassandra/tools/nodetool/CompactionStats.java
```diff
@@ -53,29 +58,82 @@ public class CompactionStats extends NodeToolCmd
...
+        toPrint.append(LINE_SEPARATOR);
         for (Entry<String, Map<String, Integer>> ksEntry : pendingTaskNumberByTable.entrySet())
         {
+            String ksName = ksEntry.getKey();
             for (Entry<String, Integer> tableEntry : ksEntry.getValue().entrySet())
-                numTotalPendingTask += tableEntry.getValue();
+            {
+                toPrint.append("- " + ksName + '.' + tableEntry.getKey() + ": " + tableEntry.getValue());
```
Maxwell-Guo: Can we use String.format uniformly.
Although its performance is not very good, and actually I prefer to use string concatenation.
But Cassandra seems to be all String.formt,So it seems that unity looks better

src/java/org/apache/cassandra/tools/nodetool/CompactionStats.java
```diff
@@ -53,29 +58,82 @@ public class CompactionStats extends NodeToolCmd
...
+        double actual = probe.getCompactionRate() / (1024 * 1024);
+        if(configured == 0)
+        {
+            return String.format("compaction throughput absolute: %s MBps", actual);
         }
-        out.println();
-        reportCompactionTable(cm.getCompactions(), probe.getCompactionThroughputBytes(), humanReadable, vtableOutput, out);
+        else
+        {
+            double percentage = (actual / configured) * 100;
```
Maxwell-Guo, line 133: I think I would change the code to :`        double configured = probe.getCompactionThroughputMebibytesAsDouble();
      
        if(configured == 0)
        {
            return String.format("compaction throughput absolute: %s MBps", actual);
        }
        else
        {
           double actual = probe.getCompactionRate() / (1024 * 1024);
            double percentage = (actual / configured) * 100;
            return String.format("compaction throughput ratio: %s MBps / %s MBps (%s%s)", actual, configured, percentage, "%");
        }`

if configured is 0, there is no need to calculate the value of actual.

test/unit/org/apache/cassandra/tools/nodetool/CompactionStatsTest.java
```diff
@@ -130,6 +130,14 @@ compactionId, OperationType.COMPACTION, CQLTester.KEYSPACE, currentTable(), byte
                                                     CompactionInfo.Unit.BYTES, (double) bytesCompacted / bytesTotal * 100);
         Assertions.assertThat(stdout).containsPattern(expectedStatsPattern);
 
+        Assertions.assertThat(stdout).containsPattern("[0-9]* concurrent compactors, [0-9]* pending tasks");
```
Maxwell-Guo: I think to be on the safe side, you can add a test case that uses the invokeNodetool method for full-link testing.
Though when using invokeNodetool from a client side , It is impossible to specifically determine the value of the test, but we can test and output some expected results, such as whether the output information has the print out field we need, and when there is no data, and when we write a certain amount of data then do a major, These print out data should met our expectations。
@driftx WDYT?

smiklosovic: I think it is just fine as is.

src/java/org/apache/cassandra/tools/nodetool/CompactionStats.java
smiklosovic: yeah just align it like this

    toPrint.append(String.format("%s concurrent compactors, %s pending tasks",
                                 probe.getConcurrentCompactors(), 
                                 numPendingTasks(pendingTaskNumberByTable)));

src/java/org/apache/cassandra/tools/nodetool/CompactionStats.java
smiklosovic: yeah just be consistent, use String.format

PR #2336, opened 2023-05-15
CASSANDRA-18506: Upgrade tests cleanup

jakubzytka:
This PR cleans up and simplifies the in-jvm upgrade tests machinery.

The assumptions used in this PR:
- only upgrades touching the `CURRENT` version should be run during testing
- the API for defining which upgrades to run should not be imperative ("do this and this upgrade"), but instead should be a declaration of test case applicability

Specific changes are:
- introduce new API in `UpgradeTestBase` that clearly expresses test case applicability
- deprecate the old, confusing `UpgradeTestBase` API, and express it in terms of the new one
- remote old, confusing, unused `UpgradeTestBase` API 
- improve javadoc for `UpgradeTestBase` API
- remove `List` usage for target version in `TestVersions`, as these lists always had precisely one element
- fix `CompactionHistorySystemTableUpgradeTest` executing the same cases multiple tiems
- remove obsolete test cases in `MixedModeFrom3UnloggedBatchTest` and `MixedModeFrom3LoggedBatchTest`
- removal of duplicate test cases with `V3X` in test name


test/distributed/org/apache/cassandra/distributed/upgrade/CompactionHistorySystemTableUpgradeTest.java
```diff
@@ -19,44 +19,26 @@
...
-    }
-
     @Test
     public void compactionHistorySystemTableTest() throws Throwable
     {
         new TestCase()
         .nodes(1)
         .nodesToUpgrade(1)
-        .upgradesToCurrentFrom(version)
+        .minimalApplicableVersion(v30)
```
jakubzytka: prior to the change the same upgrades would have been run (e.g. for parameter v30 and v3x the same set of v4->v5 and v4.1 -> v5 upgrades would have been run)

test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeAvailabilityOneAllTest.java
```diff
@@ -23,10 +23,10 @@
...
  * {@link MixedModeAvailabilityTestBase} for upgrades from v30 with ONE-ALL write-read consistency.
  */
-public class MixedModeAvailabilityV30OneAllTest extends MixedModeAvailabilityTestBase
+public class MixedModeAvailabilityOneAllTest extends MixedModeAvailabilityTestBase
 {
-    public MixedModeAvailabilityV30OneAllTest()
+    public MixedModeAvailabilityOneAllTest()
     {
-        super(v30, ConsistencyLevel.ONE, ConsistencyLevel.ALL);
+        super(ConsistencyLevel.ONE, ConsistencyLevel.ALL);
```
jakubzytka, line 30: the test cases are in `MixedModeAvailabilityTestBase` so it makes sense to define applicability there.

test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeAvailabilityTestBase.java
```diff
@@ -46,51 +46,46 @@
...
+        testAvailability(false, MINIMAL_APPLICABLE_VERSION, writeConsistencyLevel, readConsistencyLevel);
     }
 
     @Test
     public void testAvailabilityCoordinatorUpgraded() throws Throwable
     {
-        testAvailability(true, initial, writeConsistencyLevel, readConsistencyLevel);
-    }
-
-    protected static void testAvailability(Semver initial,
```
jakubzytka: this was unused

test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeConsistencyTest.java
```diff
@@ -38,9 +38,10 @@
 import static org.apache.cassandra.distributed.shared.AssertUtils.assertRows;
 import static org.apache.cassandra.distributed.shared.AssertUtils.row;
 
-public class MixedModeConsistencyTestBase extends UpgradeTestBase
+public class MixedModeConsistencyTest extends UpgradeTestBase
 {
-    protected static void testConsistency(Semver initial) throws Throwable
+    @Test
+    public void testConsistency() throws Throwable
```
jakubzytka, line 44: there used to be a single subclass for `v30`. I removed it and just inlined `v30`.
I don't see value in a separate class (unless we care for butler history)

test/distributed/org/apache/cassandra/distributed/upgrade/MixedModeFrom3UnloggedBatchTest.java
```diff
@@ -22,12 +22,6 @@
 
 public class MixedModeFrom3UnloggedBatchTest extends MixedModeBatchTestBase
 {
-    @Test
```
jakubzytka: this test case doesn't make sense on v4.0+

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -111,14 +110,14 @@ public UpgradeableCluster.Builder builder()
     public static class TestVersions
     {
         final Version initial;
```
jakubzytka, line 112: rename to `source`?

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -111,14 +110,14 @@ public UpgradeableCluster.Builder builder()
...
-        final List<Version> upgrade;
-        final List<Semver> upgradeVersions;
+        final Version targetVersion;
+        final Semver targetSemver;
```
jakubzytka, line 114: these lists always had length 1.
Is there a value in retaining the lists?

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -383,7 +483,7 @@ private static void triggerGC()
     protected TestCase allUpgrades(int nodes, int... toUpgrade)
     {
         return new TestCase().nodes(nodes)
-                             .upgradesToCurrentFrom(v30)
+                             .applicableVersionsRange(OLDEST, CURRENT)
```
jakubzytka, line 486: changed from `v30` to `OLDEST` if this is supposed to be `allUpgrades`

It is of no consequence on v5, but it's clearer I think.

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -176,83 +179,177 @@ public TestCase nodes(int nodeCount)
...
+         * Beware, there may be no supported direct upgrade path from -> CURRENT, and this is OK.
+         * e.g. when CURRENT == v4
+         * {@code upgradesToCurrentStartingNoEarlierThan(3.0); // produces: 3.0 -> CURRENT, 3.11 -> CURRENT, …}
+         * and when CURRENT == v5
+         * {@code upgradesToCurrentStartingNoEarlierThan(3.0); // produces: 4.0 -> CURRENT, 4.1 -> CURRENT, …}
+         * and when CURRENT == v3.11
+         * {@code upgradesToCurrentStartingNoEarlierThan(3.0); // produces: 3.0 -> CURRENT, …}
+         **/
+        @Deprecated
+        public TestCase upgradesToCurrentFrom(Semver from)
```
jakubzytka, line 259: I retained the old API in case some test is written in an older version and then propagated to `trunk`.
I wanted such test to work out-of-the-box (with no compilation errors etc)
I don't know if this is C* way of doing things.

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -176,83 +179,177 @@ public TestCase nodes(int nodeCount)
...
         {
-            return upgradesTo(from, CURRENT);
+            this.minimalApplicableVersion = minimalSupportedVersion;
+            addUpgrades();
+            return this;
         }
 
         /**
-         * performs all supported upgrade paths to the "to" target; example
-         * {@code upgradesTo(3.0, 4.0); // produces: 3.0 -> 4.0, 3.11 -> 4.0}
```
jakubzytka, line 209: the above javadoc is not true

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -176,83 +179,177 @@ public TestCase nodes(int nodeCount)
...
+         * - v40 -> v41
          */
-        public TestCase upgradesTo(Semver from, Semver to)
+        public TestCase applicableVersionsRange(Semver minimalSupportedVersion, Semver maximalSupportedVersion)
         {
-            List<TestVersions> upgrade = new ArrayList<>();
-            NavigableSet<Semver> vertices = sortedVertices(SUPPORTED_UPGRADE_PATHS);
-            for (Semver start : vertices.subSet(from, true, to, false))
-            {
-                // only include pairs that are allowed, and start or end on CURRENT
```
jakubzytka, line 217: this comment is not true. `contains` doesn't meant that `start == CURRENT || end == CURRENT`

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -176,83 +179,177 @@ public TestCase nodes(int nodeCount)
             return this;
         }
 
-        /** performs all supported upgrade paths that exist in between from and end on CURRENT (inclusive)
-         * {@code upgradesToCurrentFrom(3.0); // produces: 3.0 -> CURRENT, 3.11 -> CURRENT, …}
-         **/
-        public TestCase upgradesToCurrentFrom(Semver from)
```
jakubzytka: all javadoc say about "all supported upgrade paths" whereas only paths with length 1 are supported (see `SUPPORTED_UPGRADE_PATHS` graph)

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -176,83 +179,177 @@ public TestCase nodes(int nodeCount)
...
+            this.maximalApplicableVersion = maximalSupportedVersion;
+            addUpgrades();
             return this;
         }
 
         /**
-         * performs all supported upgrade paths from the "from" target; example
-         * {@code upgradesFrom(4.0, 4.2); // produces: 4.0 -> 4.1, 4.0 -> 4.2}
-         */
-        public TestCase upgradesFrom(Semver from, Semver to)
```
jakubzytka: likewise wrt javadoc

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -176,83 +179,177 @@ public TestCase nodes(int nodeCount)
...
+         * {@code upgradesToCurrentStartingNoEarlierThan(3.0); // produces: 3.0 -> CURRENT, …}
+         **/
+        @Deprecated
+        public TestCase upgradesToCurrentFrom(Semver from)
         {
-            List<TestVersions> upgrade = new ArrayList<>();
-            NavigableSet<Semver> vertices = sortedVertices(SUPPORTED_UPGRADE_PATHS);
-            for (Semver end : vertices.subSet(from, false, to, true))
-            {
-                // only include pairs that are allowed, and start or end on CURRENT
```
jakubzytka, line 236: likewise

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
```diff
@@ -176,83 +179,177 @@ public TestCase nodes(int nodeCount)
...
-            }
-            logger.info("Adding upgrades of\n{}", upgrade.stream().map(TestVersions::toString).collect(Collectors.joining("\n")));
-            this.upgrade.addAll(upgrade);
-            return this;
+            return applicableVersionsRange(from, CURRENT);
         }
 
         /**
-         * performs all supported upgrade paths that exist in between from and to that include the current version.
-         * This call is equivalent to calling {@code upgradesTo(from, CURRENT).upgradesFrom(CURRENT, to)}.
```
jakubzytka, line 247: it is very confusing that his javadoc says that this method is equivalent to sth, and then the method's code is different. There are no performance considerations etc.

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
michaelsembwever: in this class we don't need to go through the deprecation cycle.

it would be better to apply the patch to all branches to avoid this problem.

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
michaelsembwever: "all supported upgrade paths" refers to multiple edges in the supported upgrade matrix. it is not implying multi-step (or multi-edge) paths.

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
michaelsembwever: the apidoc is correct. it is the `contains(..)` method that is wrong.

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
michaelsembwever: the comment is correct. it is the `contains(..)` method that is wrong.

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
michaelsembwever: the apidoc is correct. it is the `contains(..)` method that is wrong.

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
michaelsembwever: they need to be lists.

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
jakubzytka: path in graph is a path. not an edge

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
jakubzytka: no it isn't. the function also filters using `CURRENT`

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
jakubzytka: The context of `contains` is not set, but range, because all this API works on ranges (`updateFrom`, `updateTo` etc).
The natural way to express `contains` in set context is to use type system: pass a set.

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
jakubzytka: why?

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
jakubzytka: The context of contains is not set, but range, because all this API works on ranges (`updateFrom`, `updateTo` etc).
The natural way to express contains in set context is to use type system: pass a set.

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
michaelsembwever: we're referring to versions being vertices, and an edge between two vertices. this helps because edge then implies only two endpoints.  a path then would be ambiguous (can contain multiple edges, and more than two vertices).  (yes there's terminologies being mixes here 🤷🏻‍♀️)

michaelsembwever:
can you please start with this commit: https://github.com/apache/cassandra/compare/trunk...thelastpickle:cassandra:mck/18506/trunk

this fixes the `.contains(..)` bug, and should make things a lot clearer to understand, without having to touch method signatures (which touches all the other classes).  starting off like this will help the review process progress (we can re-evaluate the need to improve method names later on in the review process…)

test/distributed/org/apache/cassandra/distributed/upgrade/UpgradeTestBase.java
jakubzytka: The term `path` is not ambiguous when used in the context of graph traversal, just like in this case.
I agree that using the term "edge" would be better than e.g. "a path with length 1" which I used, but I wanted to retain a connection to the previous formulation so that a reader that is already familiar with the old API and javadocs has the proper intuition.

jakubzytka:
> can you please start with this commit: trunk...thelastpickle:cassandra:mck/18506/trunk
> this fixes the .contains(..) bug,

Are you sure this commit fixes the `.contains()` bug?
First of all, the `edgeTouchesTarget(Semver from, Semver to, Semver target)` has special handling for patch etc. versions. This behaviour is never explained.
Furthermore, this function may be buggy.
Consider a test which defines its applicability from version `X.Y.Z` to some version `X+1.0.0`. If `CURRENT` is `X.Y.Z-1` then this function returns `true` even though the test author specifically prohibited running it. 
Possibly, this function builds on some assumption that is never explicitly stated, like: "the only valid lower applicability bounds are non-patch etc. versions"
I'd recommend using an assertion to express assumptions under which the function is valid.

> should make things a lot clearer to understand,

the commit in question adds more complexity to already complex code and doesn't fix existing inconsistencies between javadocs or the code. I don't see it as a step in the right direction, considering that a simpler alternative exists in this PR.

> without having to touch method signatures (which touches all the other classes)

would you kindly elaborate? My PR doesn't change the signatures of the existing public API.

> starting off like this will help the review process progress 

I don't understand what you mean. As already mentioned, the commit 9e32b906fb2fe822be39005ba40de882407e7848 adds additional code, whereas my PR mostly removes code. I don't see how could these two approaches meet. Unless, obviously, my change is based on some invalid assumptions, or doesn't take into account additional assumptions. 
I admit I struggled with that as I was unable to find documentation of these assumptions.

Is there any particular concern why reviewing this PR would be in any way difficult? If so, that's perhaps a reason to improve it. I tried to simplify the code as much as possible. Specifically, I'm removing complex code and replacing it with a simpler alternative.

michaelsembwever:
I'm keen to address the `.contains(…)` bug, using the commit provided, before moving forward.  This will ensure all assumptions about the correct code and naming used is understood.

dcapwell (changes requested):
tried taking a look at this patch, but the style changes make it hard to review, and they go against our style guide (see https://cassandra.apache.org/_/development/code_style.html).

Can we simplify the patch and only change the min set required for the ticket?  Style changes should be avoided.

dcapwell:
> would you kindly elaborate? My PR doesn't change the signatures of the existing public API.

this changes public APIs, which increases the scope of this patch.

* `upgrades` is now `applicableVersionsRange`
* `upgradesToCurrentFrom` is now `minimalApplicableVersion`

this adds a lot more noise to the patch, making it hard to see what is a bug fix, and what is a personal style change.

michaelsembwever:
@jakubzytka, as David says, patches require a minimal approach in the project, along with meeting the style.

I know I led you down the garden path by encouraging improvements to UpgradeTestBase to make it easier to understand. apologies for that.

The `.contains(..)` bug is now fixed, it is in the `.edgeTouchesTarget(..)` method.  And the existing `.contains(..)` method is renamed to `.rangeCoversTarget(..)`. 

With that in place, you can see that I've only then needed to change the parameter names to the method: `upgradesToCurrentFrom(..)`, `upgradesTo(..)`, `upgradesFrom(..)` and `upgrades(..)`.

This should help demonstrate my earlier points that 
- no existing apidoc or code comments were wrong, and
- upgrade and upgradeVersion needs to be lists.

To go into more detail…

> First of all, the edgeTouchesTarget(Semver from, Semver to, Semver target) has special handling for patch etc. versions.

The supported upgrade paths do not go to the granularity of patch versions.

> Furthermore, this function may be buggy. Consider a test which defines its applicability from version X.Y.Z

This is a valid point, but is not a problem today and out-of-scope.  AFAIK no one is passing patch-number explicit versions to these methods.   I agree that an assertion that we don't (yet) support patch-specific bounds would be appropriate.

--
I've completed the patch for the ticket with an additional commit on my branch.

jakubzytka:
> > would you kindly elaborate? My PR doesn't change the signatures of the existing public API.
> 
> this changes public APIs, which increases the scope of this patch.
> 
> * `upgrades` is now `applicableVersionsRange`
> * `upgradesToCurrentFrom` is now `minimalApplicableVersion`
> 
> this adds a lot more noise to the patch, making it hard to see what is a bug fix, and what is a personal style change.

The above statements are not true.
Both `upgrades` and `upgradesToCurrentFrom` are present and offer the same semantics as previously.

michaelsembwever:
> This is a valid point, but is not a problem today and out-of-scope. AFAIK no one is passing patch-number explicit versions to these methods. I agree that an assertion that we don't (yet) support patch-specific bounds would be appropriate.


Instead of adding an assertion, I've added the apidoc. 
It is ok if patch versions are passed to the the `edgeTouchesTarget(..)` method, it is intentionally only compares majors and minors.

The potential bug you describe won't happen, because it is not this method that applies the bounds.  That happens with the `vertices.subSet(..)` call.

pmcfadin:
Closing as [CASSANDRA-18506](https://issues.apache.org/jira/browse/CASSANDRA-18506) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #2338, opened 2023-05-16
CASSANDRA-18508: Allow javax.net.ssl properties to be specified in a file

ossarga:
This PR fixes the issue where sensitive JMX SSL configuration options are easily exposed when viewing the Cassandra process. It fixes the issue by adding a Cassandra system property called `cassandra.jmx.remote.ssl.config.file`. This property specifies a path to a file containing the `javax.net.ssl.* properties`. It is an alternative to specifying the properties directly in the _cassandra-env.sh_ file. It can be used as a method to hide security sensitive properties from the process output.

# Reproduce the issue using these steps

This issue exists in all versions of Cassandra.

**1. Generate JKS fomat keystore and truststore files**

This can be done using the following [instructions](https://stackoverflow.com/questions/47434877/how-to-generate-keystore-and-truststore)

Name the generated keystore and truststore keystore.jks and truststore.jks respectively, and place them in _/etc/ssl/_. Ensure their permissions are set to be readable only by the user running the Cassandra process.

**2. Configure Cassandra to allow encrypted remote JMX connections**

Modify the _cassandra-env.sh_ file as per the following snippet.

```
...
JMX_PORT="7199"
JMXREMOTE_PORT="7198"
LOCAL_JMX="no"

if [ "$LOCAL_JMX" = "yes" ]; then
  JVM_OPTS="$JVM_OPTS -Dcassandra.jmx.local.port=$JMX_PORT"
  JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.authenticate=false"
else
  JVM_OPTS="$JVM_OPTS -Dcassandra.jmx.remote.port=$JMX_PORT"
  # if ssl is enabled the same port cannot be used for both jmx and rmi so either
  # pick another value for this property or comment out to use a random port (though see CASSANDRA-7087 for origins)
   JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.rmi.port=$JMXREMOTE_PORT"

  # turn on JMX authentication. See below for further options
  JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.authenticate=true"

  # jmx ssl options
  JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.ssl=true"
  JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.ssl.need.client.auth=true"
  #JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.ssl.enabled.protocols=<enabled-protocols>"
  #JVM_OPTS="$JVM_OPTS -Dcom.sun.management.jmxremote.ssl.enabled.cipher.suites=<enabled-cipher-suites>"
  JVM_OPTS="$JVM_OPTS -Djavax.net.ssl.keyStore=/etc/ssl/keystore.jks"
  JVM_OPTS="$JVM_OPTS -Djavax.net.ssl.keyStorePassword=keystorepassword"
  JVM_OPTS="$JVM_OPTS -Djavax.net.ssl.trustStore=/etc/ssl/truststore.jks"
  JVM_OPTS="$JVM_OPTS -Djavax.net.ssl.trustStorePassword=truststorepassword"
fi
...
```

**3. Start Cassandra and inspect the system process**

This assumes installation via `tar.gz` binary distribution. The output generated by the `ps` command has been modified to highlight the issue this PR fixes. 

```
$ cd /opt/cassandra/bin
$ ./cassandra

$ ps aux | grep "cassandra"

cassand+      ... /opt/java/openjdk/bin/java -ea -da:net.openhft... -XX:+UseThreadPriorities ... -Dcassandra.jmx.remote.port=7199 -Dcom.sun.management.jmxremote.rmi.port=7198 -Dcom.sun.management.jmxremote.authenticate=true -Dcom.sun.management.jmxremote.ssl=true -Dcom.sun.management.jmxremote.ssl.need.client.auth=true -Djavax.net.ssl.keyStore=/etc/ssl/cassandra_keystore.jks -Djavax.net.ssl.keyStorePassword=cassandraprivkeypassword -Djavax.net.ssl.trustStore=/etc/ssl/common_truststore.jks -Djavax.net.ssl.trustStorePassword=truststorepassword -Dcom.sun.management.jmxremote.password.file=/etc/cassandra/jmxremote.password...: org.apache.cassandra.service.CassandraDaemon
```

patch by Anthony Grasso for CASSANDRA-18508

### DIFF
--- a/src/java/org/apache/cassandra/utils/JMXServerUtils.java
+++ b/src/java/org/apache/cassandra/utils/JMXServerUtils.java
@@ -50,11 +50,14 @@ import javax.security.auth.Subject;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableList;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.cassandra.auth.jmx.AuthenticationProxy;
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.config.EncryptionOptions;
 
 import static org.apache.cassandra.config.CassandraRelevantProperties.COM_SUN_MANAGEMENT_JMXREMOTE_ACCESS_FILE;
 import static org.apache.cassandra.config.CassandraRelevantProperties.COM_SUN_MANAGEMENT_JMXREMOTE_AUTHENTICATE;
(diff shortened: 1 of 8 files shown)


src/java/org/apache/cassandra/utils/JMXServerUtils.java
```diff
@@ -149,6 +155,55 @@ public static JMXConnectorServer createJMXServer(int port, boolean local) throws
...
+                    continue;
+                }
+
+                if (!parts[0].startsWith("javax.net.ssl"))
+                {
+                    logger.warn(
+                        "JMX remote SSL configuration file {} must only contain javax.net.ssl properties; " +
+                        "skipping property: {}", remoteSslConfigFilePath, parts[0]);
+                    continue;
+                }
```
thobe: I don't think we should implement our own parsing for properties files, when `java.util.Properties.load(java.io.Reader)` can do that according to the known specification.

test/unit/org/apache/cassandra/utils/JMXServerUtilsTest.java
```diff
@@ -0,0 +1,75 @@
...
+    {
+        System.setProperty("cassandra.jmx.remote.ssl.config.file", "/bad/file/path");
+        try
+        {
+            jmxServer = JMXServerUtils.createJMXServer(9999, false);
+            fail("Test failed; should have thrown an exception");
+        }
+        catch (Exception e)
+        {
+            assertTrue(e.getClass().toString().contains("FileNotFoundException"));
```
thobe: You get much better failure messages if you use hamcrest for this:
```suggestion
            assertThat(e, isA(FileNotFoundException.class));
```

test/unit/org/apache/cassandra/utils/JMXServerUtilsTest.java
```diff
@@ -0,0 +1,75 @@
...
+            assertTrue(e.getMessage().contains("JMX remote SSL configuration file does not exist: /bad/file/path"));
```
thobe: You get much better failure messages if you use hamcrest for this:
```suggestion
            assertThat(e.getMessage(), containsString("JMX remote SSL configuration file does not exist: /bad/file/path"));
```

thobe (commented):
Just some small comments from a quick look.

This is a useful trick to get sensitive information into system properties without having to expose them via the command that launches the JVM. However, I don't think there is anything about this that is specific to JMX, and I would be sad to see the same feature re-implemented for another purpose. So instead I would suggest that we have one single way to load system properties into the JVM from a file into Cassandra in general.

PR #2339, opened 2023-05-16
CASSANDRA-18519: (CEP-15: (C*) Add notion of CommandsForRanges and make this durable in C*)

dcapwell:


src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -40,19 +54,105 @@
...
+import org.apache.cassandra.utils.concurrent.AsyncPromise;
 import org.apache.cassandra.utils.concurrent.UncheckedInterruptedException;
 
 import static org.apache.cassandra.concurrent.ExecutorFactory.Global.executorFactory;
 
 public class AccordCommandStore extends CommandStore
 {
+    private static final Logger logger = LoggerFactory.getLogger(AccordCommandStore.class);
+
+    private static final class RangeCommandSummary
```
dcapwell: I can always serialize this to a `ByteBuffer` like we do for `CommandsForKeys`, could even use the same logic I think... I just went this way to be quicker and easier to debug while I am testing

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -248,10 +407,113 @@ public void completeOperation(AccordSafeCommandStore store,
...
+            put(ranges, summary);
+        }
+    }
+
+    private void put(Ranges ranges, RangeCommandSummary summary)
+    {
+        IntervalTree.Builder<RoutableKey, RangeCommandSummary, Interval<RoutableKey, RangeCommandSummary>> builder = rangesToCommands.unbuild();
+        //TODO double check this tree has same inclusive/exclusive semantics as this range...
+        for (Range range : ranges)
+            builder.add(new Interval<>(range.start(), range.end(), summary));
```
dcapwell: I believe this is not correct as this method gets called on state change, and `IntervalTree` will just add the update (so we see new and old); will look how to filter this

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -248,10 +407,113 @@ public void completeOperation(AccordSafeCommandStore store,
...
+            RangeCommandSummary summary = new RangeCommandSummary(txnId, current.saveStatus(), current.executeAt(), dependsOn);
+            Ranges ranges = (Ranges) keys;
+            put(ranges, summary);
+        }
+    }
+
+    private void put(Ranges ranges, RangeCommandSummary summary)
+    {
+        IntervalTree.Builder<RoutableKey, RangeCommandSummary, Interval<RoutableKey, RangeCommandSummary>> builder = rangesToCommands.unbuild();
+        //TODO double check this tree has same inclusive/exclusive semantics as this range...
```
dcapwell: this looks inclusive on both sides, and our range tend to be `(, ]`, so not a perfect match... will look to improve

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -156,12 +174,33 @@ ImmutableSortedMap<Timestamp, ByteBuffer> getValues(CommandsForKey cfk)
         }
     }
 
+    private enum TokenType
+    {
+        Murmur3((byte) 1),
```
dcapwell, line 179: TODO add more partitioner support... our tests only use murmur3 so did this just for place holder

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -156,12 +174,33 @@ ImmutableSortedMap<Timestamp, ByteBuffer> getValues(CommandsForKey cfk)
...
+        ;
+
+        private final byte value;
```
dcapwell, line 183: my thinking is that a `murmur3` token can't compare with a `bytes order` token, so by pulling out the `type` as the bytes prefix, we know that only `murmur3` tokens will be seen during the query if the range is for a `murmur3` token range

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -537,6 +581,20 @@ public static Mutation getCommandMutation(AccordCommandStore commandStore, Accor
         }
     }
 
+    public static ByteBuffer serializeToken(Token token)
+    {
+        return serializeToken(token, ByteBufferAccessor.instance);
+    }
+
+    private static <V> V serializeToken(Token token, ValueAccessor<V> accessor)
```
dcapwell, line 592: I could have written this off `ByteBuffer`, but had to do a lot of `ValueAccessor` work recently so was more habit than anything

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -595,13 +653,257 @@ public static UntypedResultSet loadCommandRow(CommandStore commandStore, TxnId t
...
+        private WalkCommandsForDomain(int commandStore, Routable.Domain domain, Set<String> requiredColumns, Executor executor, Observable<UntypedResultSet.Row> callback)
+        {
+            super(executor, callback);
+            this.storeId = commandStore;
+            this.domain = domain.ordinal();
+            cql = String.format("SELECT %s " +
+                                "FROM %s " +
+                                "WHERE store_id = ? " +
+                                "      AND domain = ? " +
+                                "      AND token(store_id, domain, txn_id) > token(?, ?, (?, ?, ?)) " +
```
dcapwell, line 788: since we know this table has a token layout that matches the data (local partitioner) this helps make sure that the exec engine does what we want.

src/java/org/apache/cassandra/dht/Token.java
```diff
@@ -41,6 +42,16 @@ public abstract class Token implements RingPosition<Token>, Serializable
         public abstract ByteBuffer toByteArray(Token token);
         public abstract Token fromByteArray(ByteBuffer bytes);
 
+        public byte[] toOrderedByteArray(Token token, ByteComparable.Version version)
+        {
+            return ByteSourceInverse.readBytes(asComparableBytes(token, version));
+        }
+
+        public Token fromOrderedByteArray(byte[] bytes, ByteComparable.Version version)
```
aweisberg, line 50: Happens to be unused, but fine.

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -252,6 +341,47 @@ public void completeOperation(AccordSafeCommandStore store,
...
+                }
+            }
+            break;
+            default:
+                throw new AssertionError("Unknown domain: " + keysOrRanges.domain());
+        }
+        return accumulate;
+    }
+
+    CommandsForRanges.Builder unbuild()
```
aweisberg: Pedantic, but this doesn't `unbuild` the `CommandStore` it unbuilds a `CommandsForRanges`.

src/java/org/apache/cassandra/gms/ApplicationState.java
```diff
@@ -57,6 +57,9 @@
      **/
     SSTABLE_VERSIONS,
     DISK_USAGE,
+
+    // TODO (replace with TCM) : This is a temp solution to know what epoch is current for each instance in the cluster, used for topology ACK callbacks
```
dcapwell: should be deleted in https://issues.apache.org/jira/browse/CASSANDRA-18444

src/java/org/apache/cassandra/service/accord/AccordConfigurationService.java
```diff
@@ -20,30 +20,53 @@
...
+    @GuardedBy("this")
     private final List<Topology> epochs = new ArrayList<>();
 
     public AccordConfigurationService(Node.Id localId)
     {
         this.localId = localId;
         epochs.add(Topology.EMPTY);
+
+        // TODO (replace with TCM): Need to know when an epoch is known on the peers, so use gossip until TCM is ready
+        Gossiper.instance.register(new IEndpointStateChangeSubscriber()
```
dcapwell: should be deleted in https://issues.apache.org/jira/browse/CASSANDRA-18444

test/distributed/org/apache/cassandra/distributed/test/accord/NewSchemaTest.java
```diff
@@ -0,0 +1,85 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(NewSchemaTest.class);
+
+    @Override
+    protected Logger logger()
+    {
+        return logger;
+    }
+
+    @Test
+    public void test()
```
dcapwell, line 42: there were bugs when we would add new key spaces, so added this test to make sure this is better covered

src/java/org/apache/cassandra/service/accord/CommandsForRanges.java
```diff
@@ -0,0 +1,337 @@
...
+                    }
+                    rangeIt = null;
+                    if (!ksIt.hasNext())
+                    {
+                        ksIt = null;
+                        return endOfData();
+                    }
+                    String ks = ksIt.next();
+                    List<Key> keys = groupByKeyspace.get(ks);
+                    Map<Range, Set<RangeCommandSummary>> groupByRange = new TreeMap<>(Range::compare);
```
aweisberg, line 387: Seems like this could just be a list since keys should already be sorted?

src/java/org/apache/cassandra/service/accord/async/AsyncOperation.java
```diff
@@ -57,7 +57,7 @@
     static class Context
     {
         final HashMap<TxnId, AccordSafeCommand> commands = new HashMap<>();
-        final HashMap<RoutableKey, AccordSafeCommandsForKey> commandsForKeys = new HashMap<>();
+        final TreeMap<RoutableKey, AccordSafeCommandsForKey> commandsForKeys = new TreeMap<>();
```
aweisberg, line 60: What does this need to be sorted for now?

test/unit/org/apache/cassandra/cql3/PreparedStatementsTest.java
```diff
@@ -772,8 +773,11 @@ private static List<String> columnNames(ResultSet rs)
         return rs.getColumnDefinitions().asList().stream().map(d -> d.getName()).collect(Collectors.toList());
     }
 
-    private static void updateTxnState()
+    private void updateTxnState()
     {
+        // this class keeps dropping tables, so the commands_for_keys cache points to stale data
+        AccordKeyspace.truncateTables();
```
aweisberg: This should be removed at some point since Accord should be able to handle that? Do we need a TODO to remember to remove this?

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -105,6 +191,9 @@ public boolean inStore()
     @Override
     protected void registerHistoricalTransactions(Deps deps)
     {
+        // TODO (impl) : Its not clear why this exists
+        // accord.coordinate.CoordinateSyncPoint.coordinate(accord.local.Node, accord.primitives.TxnId, accord.primitives.Seekables<?,?>)
```
aweisberg: As mentioned later it seems like for bootstrap we want to load in some historical information about what transactions existed? @belliottsmith  can you explain?

I don't understand the cases where it is skipping keys either.

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -231,6 +276,7 @@ private static ColumnMetadata getColumn(TableMetadata metadata, String name)
               "accord commands per key",
               "CREATE TABLE %s ("
               + "store_id int, "
+              + "key_hash blob, " // can't use "token" as this is restricted word in CQL
```
aweisberg: Nit, since we generally call them `Token` it might be clearer to call it `key_token`?

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -595,13 +664,230 @@ public static UntypedResultSet loadCommandRow(CommandStore commandStore, TxnId t
...
                                txnId.msb, txnId.lsb, txnId.node.id);
     }
 
+    public static void findAllCommandsByDomain(int commandStore, Routable.Domain domain, Set<String> columns, Observable<UntypedResultSet.Row> callback)
+    {
+        WalkCommandsForDomain work = new WalkCommandsForDomain(commandStore, domain, columns, Stage.READ.executor(), callback);
+        work.schedule();
+    }
+
+    private static abstract class TableWalk implements Runnable, DebuggableTask
```
aweisberg, line 674: I deleted most of my comments on this CQL stuff since it seems temporary anyways.

src/java/org/apache/cassandra/service/accord/AccordSafeCommandStore.java
```diff
@@ -174,19 +187,34 @@ private <O> O mapReduceForKey(Routables<?, ?> keysOrRanges, Ranges slice, BiFunc
...
                         return accumulate;
                 }
-                break;
+            }
+            break;
             case Range:
-                // TODO (required): implement
-                throw new UnsupportedOperationException();
+            {
+                // TODO (correctness): if a range is used, then the range must exist in the PreLoadContext, else the commandsForKeys won't be in-sync... can this be detected?
```
aweisberg: It can, it's a safe command store so it still has the preload context?

src/java/org/apache/cassandra/service/accord/AccordVerbHandler.java
```diff
@@ -43,6 +43,19 @@ public AccordVerbHandler(Node node)
     public void doVerb(Message<T> message) throws IOException
     {
         logger.debug("Receiving {} from {}", message.payload, message.from());
-        message.payload.process(node, EndpointMapping.getId(message.from()), message);
+        T request = message.payload;
+        Node.Id from = EndpointMapping.getId(message.from());
```
aweisberg, line 47: Move this into the innermost `if` since it is unused outside.

aweisberg (changes requested):
Not very substantial because most of what I would normally comment on is not long for this world.

src/java/org/apache/cassandra/dht/Token.java
dcapwell: unused, but I like to make sure the to/from are both there... I know our style guide says not to add dead code... but I felt its still best to have to/from APIs go together

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
dcapwell: switched to `update` and did a refactor that was in the back of my head; split `Builder` into 2: `Builder` and `Updater`...  one with a `build` method (for init) and one with `apply` which is for mutations

src/java/org/apache/cassandra/service/accord/CommandsForRanges.java
dcapwell: don't think so, so the keys are ordered, but some ranges maybe larger than others

```
[10, 20]
[0, 10000]
```

so if you have the following keys

```
15 -> [10, 20], [0, 10000]
35 -> [0, 10000]
```

here we see `[0, 10000]` twice, so we would need search for it, hence why map made sense to me (lower search cost)

src/java/org/apache/cassandra/service/accord/async/AsyncOperation.java
dcapwell: its for `org.apache.cassandra.service.accord.AccordSafeCommandStore#mapReduceForKey`

```
for (RoutableKey key : commandsForKeys.keySet())
                {
                    //TODO (duplicate code): this is a repeat of Key... only change is checking contains in range
                    if (!keysOrRanges.contains(key)) continue;
                    SafeCommandsForKey forKey = commandsForKey(key);
...
```

I thought it looked like callers of `mapReduce` expect ordering, but `map` logically shouldn't... so I might be able to remove; would just need to double check this ordering expectations.

test/unit/org/apache/cassandra/cql3/PreparedStatementsTest.java
dcapwell: I think it was fixed by the change to ignore unknown ranges... if I remove the tests still pass

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
dcapwell: I used `token` before, which is a reserved word... I could def use `key_token`

src/java/org/apache/cassandra/service/accord/AccordSafeCommandStore.java
dcapwell: added defensive check

src/java/org/apache/cassandra/service/accord/AccordVerbHandler.java
dcapwell: `from` is needed on line 59 as well, so if the outer if is false, then we do `request.process(node, from, message);`

src/java/org/apache/cassandra/service/accord/CommandsForRanges.java
```diff
@@ -0,0 +1,368 @@
...
+                }
+            }
+        };
+    }
+
+    private static Range toRange(Interval<RoutableKey, RangeCommandSummary> interval)
+    {
+        TokenKey start = (TokenKey) interval.min;
+        TokenKey end = (TokenKey) interval.max;
+        // TODO (correctness) : accord doesn't support wrap around, so decreaseSlightly may fail in some cases
```
aweisberg, line 406: Just leaving a comment here so I remember to check the outcome of this

src/java/org/apache/cassandra/service/accord/async/AsyncOperation.java
dcapwell: I actually think it should stay, so we have deterministic ordering (something needed by simulator)

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
dcapwell: I ended up implementing this... `CommandsForRanges` now has a `local` and `remote` tracking case for this... for `remote` we "learn" the ranges by merging, but for `local` we trust the local txn register path.

I think I have the logic mostly matching InMemmoryCommandStore, but did refactor to keep things simple... for one the logical `mapReduceForRange` will include the history in all cases, but since I register as `NotWitnessed` this causes `minStatus == null && test_deps == ANY_DEP` to have basically the same semantics...  this allows a cleaner setup where I don't need to have 2 different code paths

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -94,17 +120,67 @@ public AccordCommandStore(int id,
...
     {
-        return Thread.currentThread().getId() == threadId;
+        AsyncPromise<CommandsForRanges> future = new AsyncPromise<>();
+        AccordKeyspace.findAllCommandsByDomain(id, Routable.Domain.Range, ImmutableSet.of("txn_id", "status", "txn", "execute_at", "dependencies"), new Observable<UntypedResultSet.Row>()
+        {
+            private CommandsForRanges.Builder builder = new CommandsForRanges.Builder();
+            @Override
+            public void onNext(UntypedResultSet.Row row) throws Exception
+            {
+                TxnId txnId = AccordKeyspace.deserializeTxnId(row);
```
dcapwell, line 135: just thought about this... the listener isn't durable, so I might need to load them to re-create the listeners... :sigh:

src/java/org/apache/cassandra/service/accord/CommandsForRanges.java
```diff
@@ -0,0 +1,505 @@
...
+            // accord.impl.CommandTimeseries.mapReduce does the check on status and deps type, and NotWitnessed should match the semantics hard coded in InMemorySafeStore...
+            // in that store, the remote history is only ever included when minStauts == null and deps == ANY... but mapReduce sees accord.local.Status.KnownDeps.hasProposedOrDecidedDeps == false
+            // as a mis-match, so will be excluded... since NotWitnessed will return false it will only be included IFF deps = ANY.
+            // When it comes to the minStatus check, the current usage is "null", "Committed", "Accepted"... so NotWitnessed will only be included in the null case;
+            // the only subtle difference is if minStatus = NotWitnessed, this API will include these but InMemoryStore won't
+            RangeCommandSummary oldValue = txnToRange.get(txnId);
+            RangeCommandSummary newValue = oldValue == null ?
+                                           new RangeCommandSummary(txnId, ranges, SaveStatus.NotWitnessed, null, Collections.emptyList())
+                                           : oldValue.withRanges(ranges, remappingFunction);
+            if (newValue == null)
```
aweisberg, line 237: Can `newValue` actually be null here? `oldValue.withRanges` doesn't appear to return null.

src/java/org/apache/cassandra/service/accord/CommandsForRanges.java
```diff
@@ -0,0 +1,505 @@
...
+            {
+                remove(txnId);
+            }
+            else if (!oldValue.equalsDeep(newValue))
```
aweisberg, line 241: When do we hit this case where oldValue is not null, but `newValue` has more information? Is it when we are a minority in some transaction that is missing information? Seems like if we have a local value we should never be loading remote info?

My only concern here is that this reverts our knowledge of the range transaction to a less knowledgable state in some situation.

aweisberg (approved):
+1, `registerHistoricalTransactions` will need to persist everything it loads at some point.

src/java/org/apache/cassandra/service/accord/CommandsForRanges.java
dcapwell: This is to keep the semantics of `Map.merge` which is used in in-memory store.  As far as I can tell, this happens when you cross epochs, as we collect `Deps` for a single epoch, so a different epoch may have more ranges; so we collect all

aweisberg: Ah cool, forgot it merges them at the top of the function.

src/java/org/apache/cassandra/service/accord/CommandsForRanges.java
dcapwell: in the current caller, no, but from a semantic point of view yes (as we match `Map.merge`).

PR #2351, opened 2023-05-19
Move build-scripts in-tree

jmckenzie-dev:
|     Move build scripts and jenkins artifact stage in-tree
|     
|     Bring the artifact/deb/rpm build scripts (and associated docker images) from cassandra-builds repo to the .build directory. The declarative Jenkinsfile now directly declares the artifacts jobs in its pipeline. And the packaging
|     jobs can be separated and run in parallel.
|     
|     In the Jenkinsfile the initial declarative section is simplified to present a 10k ft overview of the pipeline and the mapping of each pipeline stage to local script. The local scripts are dockerised and CI-agnostic: they can be  run locally without a CI system. The Jenkinsfile can be run on any jenkins, not just ASF's ci-cassandra.a.o : any ci-cassandra.a.o specific functionality is required to quietly fail on other CI platforms. The migration of tests in-tree will happen in a follow up ticket.
|     
|      patch by Mick Semb Wever; reviewed by XXX for CASSANDRA-17869, CASSANDRA-18133

### DIFF
--- a/.build/README.md
+++ b/.build/README.md
@@ -1,0 +1,130 @@
+Building and Testing with the helper sripts
+-------------------------------------------
+
+Information on building and testing beyond the use of ant.
+All scripts also print help if the first argument is `-h`.
+
+Code Checks and Lints
+---------------------
+
+Run in docker:
+
+    .build/docker/check-code.sh
+
+
+Run without docker:
+
+    .build/check-code.sh
+
+
+Run in docker with a specific jdk.
+The following applies to all build scripts.
+
+    .build/docker/check-code.sh 11
+
+
+Run in docker with a specific build path.
+This permits parallel builds off the same source path.
+The following applies to all build scripts.
+
+    build_dir=/tmp/cass_Mtu462n .build/docker/check-code.sh
+
+
+Building Artifacts (tarball and maven)
+-------------------------------------
+
+Build with docker:
+
+    .build/docker/build-artifacts.sh
+
+
+Build without docker:
+
+    .build/build-artifacts.sh
+
+
+Build in docker with a specific jdk:
+
+    .build/docker/build-artifacts.sh 11
+
+
+Building Debian and RedHat packages
+-----------------------------------
+
+The packaging scripts are only intended to be used with docker.
+
+Build:
+
+    .build/docker/build-debian.sh
+    .build/docker/build-redhat.sh
+
+
+Build with a specific jdk:
+
+    .build/docker/build-debian.sh 11
+    .build/docker/build-redhat.sh rpm 11
+
+
+Build with centos7 and a specific jdk:
+
+    .build/docker/build-redhat.sh noboolean 11
+
+
+Running Tests
+-------------
+
+Running unit tests with docker:
+
+    .build/docker/run-tests.sh test
+
+
+Running unittests without docker:
+
+    .build/run-tests.sh test
+
+
+Running only a split of unittests, with docker:
+
+    .build/docker/run-tests.sh test 1/64
+
+
+Running unittests with a specific jdk with docker:
+
+    .build/docker/run-tests.sh test 1/64 11
+
+
+Running only unit tests matching a regexp, with docker:
+
+    .build/docker/run-tests.sh test VerifyTest 11
+
+
+Running other types of tests with docker:
+
+    .build/docker/run-tests.sh test
+    .build/docker/run-tests.sh stress-test
+    .build/docker/run-tests.sh fqltool-test
+    .build/docker/run-tests.sh microbench
+    .build/docker/run-tests.sh test-cdc
+    .build/docker/run-tests.sh test-compression
+    .build/docker/run-tests.sh test-burn
+    .build/docker/run-tests.sh long-test
+    .build/docker/run-tests.sh cqlsh-test
+    .build/docker/run-tests.sh jvm-dtest
+    .build/docker/run-tests.sh jvm-dtest-upgrade
+    .build/docker/run-tests.sh dtest
+    .build/docker/run-tests.sh dtest-novnode
+    .build/docker/run-tests.sh dtest-offheap
+    .build/docker/run-tests.sh dtest-large
+    .build/docker/run-tests.sh dtest-large-novnode
+    .build/docker/run-tests.sh dtest-upgrade
+
+
+Running python dtests without docker:
+
+    .build/run-python-dtests.sh dtest
+
+
+Other test types without docker:
+
+    .build/run-tests.sh test
+
--- a/.build/build-artifacts.sh
+++ b/.build/build-artifacts.sh
@@ -1,0 +1,28 @@
+#!/bin/sh -e
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# variables, with defaults
+[ "x${CASSANDRA_DIR}" != "x" ] || CASSANDRA_DIR="$(readlink -f $(dirname "$0")/..)"
+
+# pre-conditions
+command -v ant >/dev/null 2>&1 || { echo >&2 "ant needs to be installed"; exit 1; }
+[ -d "${CASSANDRA_DIR}" ] || { echo >&2 "Directory ${CASSANDRA_DIR} must exist"; exit 1; }
+[ -f "${CASSANDRA_DIR}/build.xml" ] || { echo >&2 "${CASSANDRA_DIR}/build.xml must exist"; exit 1; }
+
+# execute
+ant -f "${CASSANDRA_DIR}/build.xml" artifacts -Dno-checkstyle=true -Drat.skip=true -Dant.gen-doc.skip=true
+exit $?
--- a/.build/build-git.xml
+++ b/.build/build-git.xml
@@ -18,7 +18,8 @@
 -->
 <project basedir="." name="apache-cassandra-git-tasks"
          xmlns:if="ant:if">
-    <target name="get-git-sha">
+
+    <target name="_get-git-sha">
         <exec executable="git" osfamily="unix" dir="${basedir}" logError="false" failonerror="false" failifexecutionfails="false" resultproperty="git.is-available.exit-code">
             <arg value="rev-parse"/>
             <arg value="--is-inside-work-tree"/>
--- a/.build/build-jars.sh
+++ b/.build/build-jars.sh
@@ -1,0 +1,30 @@
+#!/bin/sh -e
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# temporary between CASSANDRA-18133 and CASSANDRA-18594
+
+# variables, with defaults
+[ "x${CASSANDRA_DIR}" != "x" ] || CASSANDRA_DIR="$(readlink -f $(dirname "$0")/..)"
+
+# pre-conditions
+command -v ant >/dev/null 2>&1 || { echo >&2 "ant needs to be installed"; exit 1; }
+[ -d "${CASSANDRA_DIR}" ] || { echo >&2 "Directory ${CASSANDRA_DIR} must exist"; exit 1; }
+[ -f "${CASSANDRA_DIR}/build.xml" ] || { echo >&2 "${CASSANDRA_DIR}/build.xml must exist"; exit 1; }
+
+# execute
+ant -f "${CASSANDRA_DIR}/build.xml" jar -Dno-checkstyle=true -Drat.skip=true -Dant.gen-doc.skip=true -Dno-javadoc=true
+exit $?
--- a/.build/build-owasp.xml
+++ b/.build/build-owasp.xml
@@ -18,14 +18,14 @@
   -->
 <project basedir="." name="apache-cassandra-owasp-tasks">
     <property name="dependency-check.version" value="8.3.1"/>
-    <property name="dependency-check.home" value="${build.dir}/dependency-check-ant-${dependency-check.version}"/>
+    <property name="dependency-check.home.base" value="${build.dir}"/>
+    <property name="dependency-check.home" value="${dependency-check.home.base}/dependency-check-ant-${dependency-check.version}"/>
 
     <condition property="is.dependency.check.jar">
         <available file="${dependency-check.home}/dependency-check-ant/dependency-check-ant.jar" type="file" />
     </condition>
 
     <target name="dependency-check-download"
-            depends="build"
             description="Fetch OWASP Dependency checker"
             unless="is.dependency.check.jar">
 
@@ -33,8 +33,10 @@
 
         <mkdir dir="${dependency-check.home}"/>
 
-        <get src="https://github.com/jeremylong/DependencyCheck/releases/download/v${dependency-check.version}/dependency-check-ant-${dependency-check.version}-release.zip"
-             dest="${dependency-check.home}/dependency-check-ant-${dependency-check.version}-release.zip"/>
+        <retry retrycount="3" retrydelay="10" >
+            <get src="https://github.com/jeremylong/DependencyCheck/releases/download/v${dependency-check.version}/dependency-check-ant-${dependency-check.version}-release.zip"
+                 dest="${dependency-check.home}/dependency-check-ant-${dependency-check.version}-release.zip"/>
+        </retry>
 
         <unzip src="${dependency-check.home}/dependency-check-ant-${dependency-check.version}-release.zip" dest="${dependency-check.home}"/>
     </target>
--- a/.build/build-rat.xml
+++ b/.build/build-rat.xml
@@ -86,10 +86,10 @@
                  <!-- Documentation files -->
                  <exclude name="**/doc/modules/**/*"/>
                  <exclude name="**/src/java/**/*.md"/>
-                 <!-- NOTICE files -->
+                 <!-- NOTICE, LICENSE, README files -->
                  <exclude name="**/NOTICE.md"/>
-                 <!-- LICENSE files -->
                  <exclude name="**/LICENSE.md"/>
+                 <exclude name="**/README.md"/>
             </fileset>
         </rat:report>
         <exec executable="grep" outputproperty="rat.failed.files" failifexecutionfails="false">
@@ -105,15 +105,4 @@
             </condition>
         </fail>
     </target>
-
-    <target name="_assert_rat_output">
-        <fail message="The rat report at build/rat.txt was not generated. Please ensure that the rat-check task is able to run successfully. For dev builds only, touch build/rat.txt to skip this check">
-            <condition>
-                <not>
-                    <available file="${build.dir}/rat.txt" />
-                </not>
-            </condition>
-        </fail>
-    </target>
-
 </project>
--- a/.build/build-resolver.xml
+++ b/.build/build-resolver.xml
@@ -46,8 +46,10 @@
         <echo>Downloading Resolver ANT Tasks...</echo>
         <mkdir dir="${local.repository}/org/apache/maven/resolver/maven-resolver-ant-tasks/${resolver-ant-tasks.version}" />
 
-        <get src="${resolver-ant-tasks.url}/${resolver-ant-tasks.version}/maven-resolver-ant-tasks-${resolver-ant-tasks.version}-uber.jar"
-             dest="${resolver-ant-tasks.local}" usetimestamp="true" quiet="true"/>
+        <retry retrycount="3" retrydelay="10" >
+            <get src="${resolver-ant-tasks.url}/${resolver-ant-tasks.version}/maven-resolver-ant-tasks-${resolver-ant-tasks.version}-uber.jar"
+                 dest="${resolver-ant-tasks.local}" usetimestamp="true" quiet="true"/>
+        </retry>
     </target>
 
     <target name="resolver-init" depends="init,_resolver_download" unless="resolver-ant-tasks.initialized" description="Initialize Resolver ANT Tasks">
@@ -175,15 +177,19 @@
         <resolvepom file="${build.dir}/${final.name}.pom" id="all-pom" />
         <resolvepom file="${build.dir}/tmp-${final.name}-deps.pom" id="pom-deps" />
 
-        <resolve>
-            <dependencies pomRef="all-pom"/>
-            <files dir="${build.dir.lib}/jars" layout="{artifactId}-{version}-{classifier}.{extension}" scopes="compile,provided,!system"/>
-        </resolve>
-        <resolve>
-            <dependencies pomRef="pom-deps"/>
-            <!-- Needed to include compile here, so ant _build-test would not fail on missing jimfs dependency -->
-            <files dir="${test.lib}/jars" layout="{artifactId}-{version}-{classifier}.{extension}" scopes="compile,test,!provide,!system"/>
-        </resolve>
+        <retry retrycount="3" retrydelay="10" >
+            <resolve>
+                <dependencies pomRef="all-pom"/>
+                <files dir="${build.dir.lib}/jars" layout="{artifactId}-{version}-{classifier}.{extension}" scopes="compile,provided,!system"/>
+            </resolve>
+        </retry>
+        <retry retrycount="3" retrydelay="10" >
+            <resolve>
+                <dependencies pomRef="pom-deps"/>
+                <!-- Needed to include compile here, so ant _build-test would not fail on missing jimfs dependency -->
+                <files dir="${test.lib}/jars" layout="{artifactId}-{version}-{classifier}.{extension}" scopes="compile,test,!provide,!system"/>
+            </resolve>
+        </retry>
 
 
         <!-- jacoco agent jar comes wrapped in a jar -->
@@ -198,44 +204,19 @@
     <target name="resolver-dist-lib" depends="resolver-retrieve-build">
         <resolvepom file="${build.dir}/${final.name}.pom" id="all-pom" />
 
-        <resolve failOnMissingAttachments="true">
-            <dependencies pomRef="all-pom"/>
-            <files dir="${build.lib}" layout="{artifactId}-{version}-{classifier}.{extension}" scopes="compile,!provide,!system"/>
-        </resolve>
+        <retry retrycount="3" retrydelay="10" >
+            <resolve failOnMissingAttachments="true">
+                <dependencies pomRef="all-pom"/>
+                <files dir="${build.lib}" layout="{artifactId}-{version}-{classifier}.{extension}" scopes="compile,!provide,!system"/>
+            </resolve>
+        </retry>
         <mkdir dir="${local.repository}/org/apache/cassandra/deps/sigar-bin"/>
         <mkdir dir="${build.lib}/sigar-bin"/>
 
-        <!-- files.pythonhosted.org/packages -->
-        <get src="${artifact.python.pypi}/59/a0/cf4cd997e1750f0c2d91c6ea5abea218251c43c3581bcc2f118b00baf5cf/futures-2.1.6-py2.py3-none-any.whl" dest="${local.repository}/org/apache/cassandra/deps/futures-2.1.6-py2.py3-none-any.zip" usetimestamp="true" quiet="true" skipexisting="true"/>
-        <get src="${artifact.python.pypi}/73/fb/00a976f728d0d1fecfe898238ce23f502a721c0ac0ecfedb80e0d88c64e9/six-1.12.0-py2.py3-none-any.whl" dest="${local.repository}/org/apache/cassandra/deps/six-1.12.0-py2.py3-none-any.zip" usetimestamp="true" quiet="true" skipexisting="true"/>
-        <get src="${artifact.python.pypi}/37/b2/ef1124540ee2c0b417be8d0f74667957e6aa084a3f26621aa67e2e77f3fb/pure_sasl-0.6.2-py2-none-any.whl" dest="${local.repository}/org/apache/cassandra/deps/pure_sasl-0.6.2-py2-none-any.zip" usetimestamp="true" quiet="true" skipexisting="true"/>
-        <get src="${artifact.python.pypi}/59/7c/e39aca596badaf1b78e8f547c807b04dae603a433d3e7a7e04d67f2ef3e5/wcwidth-0.2.5-py2.py3-none-any.whl" dest="${local.repository}/org/apache/cassandra/deps/wcwidth-0.2.5-py2.py3-none-any.zip" usetimestamp="true" quiet="true" skipexisting="true"/>
+        <retry retrycount="3" retrydelay="10" >
+            <antcall target="_resolver-dist-lib_get_files"/>
+        </retry>
 
-        <!-- apache/cassandra/lib -->
-        <get src="${lib.download.base.url}/lib/geomet-0.1.0.zip" dest="${local.repository}/org/apache/cassandra/deps/geomet-0.1.0.zip" usetimestamp="true" quiet="true" skipexisting="true"/>
-        <get dest="${local.repository}/org/apache/cassandra/deps/sigar-bin/" quiet="true" usetimestamp="true" skipexisting="true">
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-amd64-freebsd-6.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-amd64-linux.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-amd64-solaris.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ia64-hpux-11.sl"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ia64-linux.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-pa-hpux-11.sl"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ppc-aix-5.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ppc-linux.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ppc64-aix-5.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ppc64-linux.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ppc64le-linux.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-s390x-linux.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-sparc-solaris.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-sparc64-solaris.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-universal-macosx.dylib"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-universal64-macosx.dylib"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-x86-freebsd-5.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-x86-freebsd-6.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-x86-linux.so"/>
-            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-x86-solaris.so"/>
-        </get>
-        
         <copy todir="${build.lib}" quiet="true">
             <file file="${local.repository}/org/apache/cassandra/deps/futures-2.1.6-py2.py3-none-any.zip"/>
             <file file="${local.repository}/org/apache/cassandra/deps/six-1.12.0-py2.py3-none-any.zip"/>
@@ -266,4 +247,38 @@
             <file file="${local.repository}/org/apache/cassandra/deps/sigar-bin/libsigar-x86-solaris.so"/>
         </copy>
     </target>
+
+    <target name="_resolver-dist-lib_get_files">
+        <!-- files.pythonhosted.org/packages -->
+        <get src="${artifact.python.pypi}/59/a0/cf4cd997e1750f0c2d91c6ea5abea218251c43c3581bcc2f118b00baf5cf/futures-2.1.6-py2.py3-none-any.whl" dest="${local.repository}/org/apache/cassandra/deps/futures-2.1.6-py2.py3-none-any.zip" usetimestamp="true" quiet="true" skipexisting="true"/>
+        <get src="${artifact.python.pypi}/73/fb/00a976f728d0d1fecfe898238ce23f502a721c0ac0ecfedb80e0d88c64e9/six-1.12.0-py2.py3-none-any.whl" dest="${local.repository}/org/apache/cassandra/deps/six-1.12.0-py2.py3-none-any.zip" usetimestamp="true" quiet="true" skipexisting="true"/>
+        <get src="${artifact.python.pypi}/37/b2/ef1124540ee2c0b417be8d0f74667957e6aa084a3f26621aa67e2e77f3fb/pure_sasl-0.6.2-py2-none-any.whl" dest="${local.repository}/org/apache/cassandra/deps/pure_sasl-0.6.2-py2-none-any.zip" usetimestamp="true" quiet="true" skipexisting="true"/>
+        <get src="${artifact.python.pypi}/59/7c/e39aca596badaf1b78e8f547c807b04dae603a433d3e7a7e04d67f2ef3e5/wcwidth-0.2.5-py2.py3-none-any.whl" dest="${local.repository}/org/apache/cassandra/deps/wcwidth-0.2.5-py2.py3-none-any.zip" usetimestamp="true" quiet="true" skipexisting="true"/>
+
+        <!-- apache/cassandra/lib -->
+        <get src="${lib.download.base.url}/lib/geomet-0.1.0.zip" dest="${local.repository}/org/apache/cassandra/deps/geomet-0.1.0.zip" usetimestamp="true" quiet="true" skipexisting="true"/>
+        <get dest="${local.repository}/org/apache/cassandra/deps/sigar-bin/" quiet="true" usetimestamp="true" skipexisting="true">
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-amd64-freebsd-6.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-amd64-linux.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-amd64-solaris.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ia64-hpux-11.sl"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ia64-linux.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-pa-hpux-11.sl"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ppc-aix-5.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ppc-linux.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ppc64-aix-5.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ppc64-linux.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-ppc64le-linux.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-s390x-linux.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-sparc-solaris.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-sparc64-solaris.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-universal-macosx.dylib"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-universal64-macosx.dylib"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-x86-freebsd-5.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-x86-freebsd-6.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-x86-linux.so"/>
+            <url url="${lib.download.base.url}/lib/sigar-bin/libsigar-x86-solaris.so"/>
+        </get>
+    </target>
+
 </project>
--- a/.build/docker/_build-debian.sh
+++ b/.build/docker/_build-debian.sh
@@ -1,0 +1,129 @@
+#!/bin/bash
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+################################
+#
+# Prep
+#
+################################
+
+# variables, with defaults
+[ "x${CASSANDRA_DIR}" != "x" ] || CASSANDRA_DIR="$(readlink -f $(dirname "$0")/..)"
+[ "x${DIST_DIR}" != "x" ] || DIST_DIR="${CASSANDRA_DIR}/build"
+
+# pre-conditions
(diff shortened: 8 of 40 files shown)


.build/build-rat.xml
```diff
@@ -106,14 +106,4 @@
         </fail>
     </target>
 
-    <target name="_assert_rat_output">
```
jmckenzie-dev, line 109: Why the removal here?

.build/docker/_create_user.sh
```diff
@@ -0,0 +1,54 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+################################
+#
+# Prep
+#
+################################
+
+if [ "$1" == "-h" ]; then
```
jmckenzie-dev, line 24: We should add a little context here in the help output to indicate what the method does / is used for. Also might be worth validating correct argument # are input.

.build/docker/_docker_init_tests.sh
```diff
@@ -0,0 +1,32 @@
...
+# limitations under the License.
+
+# inside the docker container, setup env before calling cassandra-test.sh
+export LANG=en_US.UTF-8
+export LC_CTYPE=en_US.UTF-8
+export PYTHONIOENCODING=utf-8
+export PYTHONUNBUFFERED=true
+export ANT_OPTS="${ANT_OPTS} -Dbuild.dir=${DIST_DIR}"
+
+pushd ${CASSANDRA_DIR} >/dev/null
```
jmckenzie-dev: We don't check to ensure the various env var are set correctly here. Also - there's a lot of overlap in env var used between scripts; I think we should strongly consider moving those env var declaration to a central script that the others then source.

.build/docker/_docker_run.sh
```diff
@@ -0,0 +1,121 @@
...
+
+# variables, with defaults
+[ "x${cassandra_dir}" != "x" ] || cassandra_dir="$(readlink -f $(dirname "$0")/../..)"
+[ "x${build_dir}" != "x" ] || build_dir="${cassandra_dir}/build"
+[ -d "${build_dir}" ] || { mkdir -p "${build_dir}" ; }
+
+java_version_default=`grep 'property\s*name="java.default"' ${cassandra_dir}/build.xml |sed -ne 's/.*value="\([^"]*\)".*/\1/p'`
+java_version_supported=`grep 'property\s*name="java.supported"' ${cassandra_dir}/build.xml |sed -ne 's/.*value="\([^"]*\)".*/\1/p'`
+
+# arguments
```
jmckenzie-dev, line 43: Same as in another(s) script: we should check arg count and print out usage if incorrect # are passed in. See line 21 w/usage information

.build/docker/almalinux-build.docker
```diff
@@ -0,0 +1,60 @@
...
+   java-11-openjdk-devel \
+   java-17-openjdk-devel \
+   make \
+   rpm-build \
+   sudo \
+   python3-pip \
+   procps \
+   rsync
+
+RUN until curl -f -S -s --retry 9 --retry-connrefused --retry-delay 1 http://mirror.centos.org/centos/7/os/x86_64/Packages/ant-junit-1.9.4-2.el7.noarch.rpm -o ant-junit-1.9.4-2.el7.noarch.rpm ; do echo "curl failed… trying again in 10s… " ; sleep 10 ; done
```
jmckenzie-dev, line 55: Why 9 retries on curl here? Could we document why retries at all and why the choice of number?

.build/docker/almalinux-build.docker
```diff
@@ -0,0 +1,60 @@
...
+    && echo " - CASSANDRA_DIR=${CASSANDRA_DIR}" \
+    && echo " - UID_ARG=${UID_ARG}" \
+    && echo " - GID_ARG=${GID_ARG}"
+
+# install deps
+RUN yum -y install \
+   ant \
+   git \
+   java-1.8.0-openjdk-devel \
+   java-11-openjdk-devel \
```
jmckenzie-dev, line 46: Will these packages drift in their version? i.e. as new JDK 8, 11, and 17 are released, will this potentially change underneath us in our build scripts?

Might be worth having these versions also declared in a central .sh file we source here. One-stop-shop for all env var for the build we can then update as needed.

There may be an issue where different environments have different JDK releases; in that case we could declare the desired JDK version locally in their respective docker files and reference those variables.

.build/docker/build-redhat.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# arguments
+rpm_dist=$1
```
jmckenzie-dev, line 26: Same as in other scripts - usage and help text?

.build/docker/bullseye-build.docker
```diff
@@ -0,0 +1,53 @@
...
+VOLUME ${DIST_DIR}
+VOLUME ${CASSANDRA_DIR}
+
+RUN echo "Building with arguments:" \
+    && echo " - DIST_DIR=${DIST_DIR}" \
+    && echo " - BUILD_HOME=${BUILD_HOME}" \
+    && echo " - CASSANDRA_DIR=${CASSANDRA_DIR}"
+
+# configure apt to retry downloads
+RUN echo 'APT::Acquire::Retries "99";' > /etc/apt/apt.conf.d/80-retries
```
jmckenzie-dev, line 37: We have 9 retries in some places, 3 in others, and now 99 here. Could we document why this number of retries is necessary?

.build/docker/bullseye-build.docker
```diff
@@ -0,0 +1,53 @@
...
+RUN echo 'Acquire::http::Timeout "60";' > /etc/apt/apt.conf.d/80proxy.conf
+RUN echo 'Acquire::ftp::Timeout "60";' >> /etc/apt/apt.conf.d/80proxy.conf
+
+# install deps
+RUN until apt-get update && apt-get -y install ant build-essential curl devscripts git sudo python3-pip rsync procps dh-python quilt bash-completion ; \
+    do echo "apt failed… trying again in 10s… " ; sleep 10 ; done
+
+RUN echo 'deb http://deb.debian.org/debian sid main' >> /etc/apt/sources.list
+
+RUN until apt-get update && apt-get install -y --no-install-recommends openjdk-11-jdk openjdk-17-jdk ; \
```
jmckenzie-dev: We have JDK8, 11, and 17 on some platforms and just 11 and 17 here. Was this an oversight?

.build/docker/bullseye-build.docker
```diff
@@ -0,0 +1,53 @@
...
+    && echo " - BUILD_HOME=${BUILD_HOME}" \
+    && echo " - CASSANDRA_DIR=${CASSANDRA_DIR}"
+
+# configure apt to retry downloads
+RUN echo 'APT::Acquire::Retries "99";' > /etc/apt/apt.conf.d/80-retries
+RUN echo 'Acquire::http::Timeout "60";' > /etc/apt/apt.conf.d/80proxy.conf
+RUN echo 'Acquire::ftp::Timeout "60";' >> /etc/apt/apt.conf.d/80proxy.conf
+
+# install deps
+RUN until apt-get update && apt-get -y install ant build-essential curl devscripts git sudo python3-pip rsync procps dh-python quilt bash-completion ; \
```
jmckenzie-dev: We break up long lines with \ in some other files to make them a bit more readable; could be nice to have that here as well.

.build/docker/centos7-build.docker
```diff
@@ -0,0 +1,85 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM centos:7
+MAINTAINER Apache Cassandra <dev@cassandra.apache.org>
+
+# CONTEXT is expected to be cassandra/.build
+
+ENV BUILD_HOME=/home/build
```
jmckenzie-dev, line 22: Is there anything we can do to reduce redundancy between our various .docker files? Perhaps look into using docker compose to generate things w/some shared base env vars so we only have one place we need to go to update shared settings?

.build/docker/run-tests.sh
```diff
@@ -0,0 +1,259 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#
+#
+# A wrapper script to run-tests.sh (or dtest-python.sh)
+#  that splits the test list into multiple docker runs, collecting results.
+#
+# Each split chunk may be further parallelised over docker containers based on the host's available cpu and memory.
```
jmckenzie-dev: Is this required because we have a heterogenous execution environment? And if so, will that remain true going forward?

Even if not, the ability to disable / bypass this step might be useful in the case where something is executing in a heterogenous environment.

.build/docker/run-tests.sh
```diff
@@ -0,0 +1,259 @@
...
+    echo "unrecognized \"${target}\""
+    exit 1
+    ;;
+esac
+
+docker_runs=1 # tmp FIXME
+
+# Break up the requested split chunk into a number of concurrent docker runs, as calculated above
+# This will typically be between one to four splits. Five splits would require >25 cores and >25GB ram
+inner_splits=$(( $(echo $split_chunk | cut -d"/" -f2 ) * ${docker_runs} ))
```
jmckenzie-dev: This was a bit hard for me to parse and follow. Could refactor it into something like the following as you see fit:

```
# Extract the numerator and denominator from the split_chunk variable
numerator=$(echo $split_chunk | cut -d"/" -f1)
denominator=$(echo $split_chunk | cut -d"/" -f2)

# Calculate inner_splits and inner_split_first using the extracted values
inner_splits=$((denominator * docker_runs))
inner_split_first=$(((numerator * docker_runs) - (docker_runs - 1)))

# Calculate docker_cpus with a scale of 2 decimal places
docker_cpus=$(echo "scale=2; ${cores} / (${jenkins_executors} * ${docker_runs})" | bc)
```

.build/docker/run-tests.sh
```diff
@@ -0,0 +1,259 @@
...
+    docker_flags="--cpus=${docker_cpus} -m 5g --memory-swap 5g"
+fi
+docker_flags="${docker_flags} --env-file build/env.list -d --rm"
+
+# make sure build_dir is good
+mkdir -p ${cassandra_dir}/build/tmp || true
+mkdir -p ${cassandra_dir}/build/test/logs || true
+chmod -R ag+rwx ${cassandra_dir}/build
+
+# FIXME python
```
jmckenzie-dev: What's the word here? Why FIXME?

.build/docker/run-tests.sh
```diff
@@ -0,0 +1,259 @@
...
+        docker_runs=$(( ${max_docker_runs_by_cores} < ${max_docker_runs_by_mem} ? ${max_docker_runs_by_cores} : ${max_docker_runs_by_mem} ))
+        docker_runs=$(( ${docker_runs} < 1 ? 1 : ${docker_runs} ))
+    ;;
+    *)
+    echo "unrecognized \"${target}\""
+    exit 1
+    ;;
+esac
+
+docker_runs=1 # tmp FIXME
```
jmckenzie-dev: ?

.build/docker/run-tests.sh
```diff
@@ -0,0 +1,259 @@
...
+EOF
+
+declare -a docker_ids
+declare -a process_ids
+declare -a statuses
+
+for i in `seq 1 ${docker_runs}` ; do
+    inner_split=$(( ${inner_split_first} + ( $i - 1 ) ))
+
+    container_name="cassandra_${dockerfile/.docker/}_${target}_jdk${java_version/./-}_arch-$(arch)_${python_version/./-}_cython-${cython}_${inner_split}_${inner_splits}__$(LC_ALL=C tr -dc A-Za-z0-9 </dev/urandom | head -c 6 ; echo '')"
```
jmckenzie-dev: This was pretty hard to parse. Another refactor option (I haven't tested it; needs review):
```
dockerfile_name="${dockerfile/.docker/}"
formatted_java_version="${java_version/./-}"
architecture="$(arch)"
formatted_python_version="${python_version/./-}"
random_string="$(LC_ALL=C tr -dc A-Za-z0-9 </dev/urandom | head -c 6 ; echo '')"

container_name="cassandra_${dockerfile_name}_${target}_jdk${formatted_java_version}_arch-${architecture}_${formatted_python_version}_cython-${cython}_${inner_split}_${inner_splits}__${random_string}"
```

.build/docker/run-tests.sh
```diff
@@ -0,0 +1,259 @@
...
+    # start the container
+    docker_id=$(docker run --name ${container_name} ${docker_flags} ${docker_mounts} ${image_name} sleep inf)
+
+    echo "Running container ${container_name} ${docker_id}"
+
+    docker exec --user root ${container_name} bash -c "\${CASSANDRA_DIR}/.build/docker/_create_user.sh cassandra $(id -u) $(id -g)"
+
+    # capture logs and pid for container
+    docker exec --user cassandra ${container_name} dumb-init bash -ilc -c "${docker_command}" > build/test/logs/docker_attach_${i}.log &
+    #docker attach --no-stdin ${docker_id} > build/test/logs/docker_attach_${i}.log &
```
jmckenzie-dev: Commented out code

.build/docker/run-tests.sh
```diff
@@ -0,0 +1,259 @@
...
+done
+
+exit_result=0
+i=0
+for process_id in "${process_ids[@]}" ; do
+    # wait for each container to complete
+    docker_id=${docker_ids[$i]}
+    inner_split=$(( $inner_split_first + $i ))
+    cat build/test/logs/docker_attach_$(( $i + 1 )).log
+    tail -F build/test/logs/docker_attach_$(( $i + 1 )).log &
```
jmckenzie-dev: Wasn't clear to me why we're tailing the process here; is the hope to catch things that hang and have visibility into them? If so, a comment would help clarify.

.build/docker/run-tests.sh
```diff
@@ -0,0 +1,259 @@
...
+        echo "${docker_id} failed (${status}), debug…"
+        docker inspect ${docker_id}
+        echo "–––"
+        docker logs ${docker_id}
+        echo "–––"
+        docker ps -a
+        echo "–––"
+        docker info
+        echo "–––"
+        exit_result=1
```
jmckenzie-dev: Why exit with 1 instead of the value in $status?

.build/docker/run-tests.sh
```diff
@@ -0,0 +1,259 @@
...
+    else
+        echo "${docker_id} done (status=${status}), copying files…"
+        # jvm tests
+        docker cp ${docker_id}:/home/cassandra/cassandra/build/test/output/. build/test/output 2>/dev/null
```
jmckenzie-dev: Can the `docker cp` calls fail? If so, might be worth looking for that and handling their return codes here.

.build/build-rat.xml
```diff
@@ -106,14 +106,4 @@
         </fail>
     </target>
```
jmckenzie-dev: nit: after removal have whitespace here

.build/build-owasp.xml
```diff
@@ -17,24 +17,26 @@
...
             description="Fetch OWASP Dependency checker"
             unless="is.dependency.check.jar">
 
         <echo>Downloading OWASP Dependency checks ...</echo>
 
         <mkdir dir="${dependency-check.home}"/>
 
-        <get src="https://github.com/jeremylong/DependencyCheck/releases/download/v${dependency-check.version}/dependency-check-ant-${dependency-check.version}-release.zip"
-             dest="${dependency-check.home}/dependency-check-ant-${dependency-check.version}-release.zip"/>
+        <retry retrycount="3" retrydelay="10" >
```
jmckenzie-dev, line 36: Why the retrycount of 3? Why not more or less? Could we document this?

.build/build-resolver.xml
```diff
@@ -46,8 +46,10 @@
         <echo>Downloading Resolver ANT Tasks...</echo>
         <mkdir dir="${local.repository}/org/apache/maven/resolver/maven-resolver-ant-tasks/${resolver-ant-tasks.version}" />
 
-        <get src="${resolver-ant-tasks.url}/${resolver-ant-tasks.version}/maven-resolver-ant-tasks-${resolver-ant-tasks.version}-uber.jar"
-             dest="${resolver-ant-tasks.local}" usetimestamp="true" quiet="true"/>
+        <retry retrycount="3" retrydelay="10" >
```
jmckenzie-dev, line 49: Similar to owasp - why 3 retries? Could we document?

.build/build-git.xml
```diff
@@ -51,4 +52,5 @@
         <echo level="warning" if:true="${is-dirty}">Repository state is dirty</echo>
         <echo level="warning" if:true="${is-dirty}">${git.diffstat}</echo>
     </target>
+
```
jmckenzie-dev: nit: Whitespace

.build/docker/ubuntu2004_test.docker
```diff
@@ -0,0 +1,160 @@
...
+RUN virtualenv --python=python3.8 env3.8
+RUN chmod +x env3.8/bin/activate
+RUN /bin/bash -c "export CASS_DRIVER_NO_CYTHON=1 CASS_DRIVER_NO_EXTENSIONS=1 && source ~/env3.8/bin/activate && pip3 install --upgrade pip && pip3 install -r /opt/requirements.txt && pip3 freeze --user"
+
+# FIXME 3.11 pip broken with html5lib
+#RUN virtualenv --python=python3.11 env3.11
+#RUN chmod +x env3.11/bin/activate
+#RUN /bin/bash -c "export CASS_DRIVER_NO_CYTHON=1 CASS_DRIVER_NO_EXTENSIONS=1 && source ~/env3.11/bin/activate && pip install --upgrade pip && pip install -r /opt/requirements.txt && pip freeze --user"
+
+# we need to make SSH less strict to prevent various dtests from failing when they attempt to
```
jmckenzie-dev: Is there no way for us to trust the keys for gh so we don't have to open this up? Can't immediately think of an attack vector but worth keeping in mind.

.build/dtest-python.sh
```diff
@@ -0,0 +1,131 @@
...
+    exit 1
+fi
+
+SPLIT_TESTS=""
+if [ "x${DTEST_SPLIT_CHUNK}" != "x" ] ; then
+    ./run_dtests.py --cassandra-dir=$CASSANDRA_DIR ${DTEST_ARGS} --dtest-print-tests-only --dtest-print-tests-output=${WORKSPACE}/test_list.txt 2>&1 > ${WORKSPACE}/test_stdout.txt
+    SPLIT_TESTS=$(split -n r/${DTEST_SPLIT_CHUNK} ${WORKSPACE}/test_list.txt)
+fi
+
+PYTEST_OPTS="-vv --log-cli-level=DEBUG --junit-xml=nosetests.xml --junit-prefix=${DTEST_TARGET} -s"
```
jmckenzie-dev: So -vv and DEBUG everywhere is going to be pretty noisy. How important is it that we have all the gory details for everything vs. just having the test failure logged we can then optionally re-run with verbose logging on?

.build/run-tests.sh
jmckenzie-dev, line 1: Having a docker/run-tests.sh and .build/run-tests.sh is a little confusing. Recommend renaming one of them.

.build/run-tests.sh
```diff
@@ -0,0 +1,198 @@
...
+[ -f "${CASSANDRA_DIR}/build.xml" ] || { echo >&2 "${CASSANDRA_DIR}/build.xml must exist"; exit 1; }
+
+# print debug information on versions
+ant -version
+git --version
+pip --version
+java -version
+javac -version
+
+# lists all tests for the specific test type
```
jmckenzie-dev, line 54: Is there a way for us to have more intelligence around how we bucket tests? For instance, by their runtime, so we don't overload specific docker containers with a heavy split?

Not sure where we'd persist that and pull from; might be something we could manually run periodically and then update the runtime with.

.jenkins/Jenkinsfile
```diff
@@ -16,524 +16,486 @@
...
+/**
+ * Saves JUnit results, fails if no reports are found
+ * Copies (single-file unified) junit results and logfiles to nightlies
+ * Cleans the agent (docker prune) afterwards
+ * Stops the pipeline on any failure
+ **/
+def testJob(test_script, test_type, splits=1, toCopy='') {
+  //def archs = ['cassandra', 'cassandra-arm64'] // uncomment when ready for arm64 testing
+  def archs = ['cassandra']
+  //def jdks = "${javaVersionsSupported}".split(/,/, -1)
```
jmckenzie-dev: Why is this commented out / still here?

.jenkins/Jenkinsfile
```diff
@@ -16,524 +16,486 @@
...
+         }
+       }
+       stage('cqlsh') {
+         steps {
+           testJob(".build/docker/run-tests.sh", "cqlsh-test", 1)
+         }
+       }
+     }
+    }
+    // stage('Distributed Test') {
```
jmckenzie-dev: Why all commented out? Just interim local testing or something, or is this not yet working?

.jenkins/Jenkinsfile
```diff
@@ -16,524 +16,486 @@
...
+  }
+}
+
+/**
+ * Saves JUnit results, fails if no reports are found
+ * Copies (single-file unified) junit results and logfiles to nightlies
+ * Cleans the agent (docker prune) afterwards
+ * Stops the pipeline on any failure
+ **/
+def testJob(test_script, test_type, splits=1, toCopy='') {
```
jmckenzie-dev: With the rather strong caveat that I haven't tested this and it needs review, this was definitely hard to parse and reason about. An alternative refactor might be something like:
```
/**
 * Saves JUnit results, fails if no reports are found
 * Copies (single-file unified) junit results and logfiles to nightlies
 * Cleans the agent (docker prune) afterwards
 * Stops the pipeline on any failure
 **/
def testJob(test_script, test_type, splits = 1, toCopy = '') {
  def axis = [
    ['cassandra'],
    ['11'],   // "${javaVersionsSupported}".split(/,/, -1)
    ['python3.7'],
    ['no']
  ]

  def generateTests(axis, split, toCopy, test_script, test_type, index = 0, tests_per_axis = [:]) {
    if (index < axis.size()) {
      def tests_per_current = [:]
      for (a in axis[index]) {
        tests_per_current[a] = {
          generateTests(axis, split, toCopy, test_script, test_type, index + 1, tests_per_current)
        }
      }
      parallel(tests_per_current)
    } else if (index == axis.size()) {
      def tests_per_split = [:]
      for (s in 1..split) {
        tests_per_split[s] = {
          _testAxis(test_script, test_type, toCopy, *axis.collect { it }, "${s}/${split}")
        }
      }
      parallel(tests_per_split)
    }
  }

  // Call the function
  generateTests(axis, splits, toCopy, test_script, test_type)
}
```
And of course the JDK axis would need to be updated (or pulled in from a central env file!)

.jenkins/Jenkinsfile
```diff
@@ -16,524 +16,486 @@
...
-                    script {
-                        copyTestResults('test', test.getNumber())
-                    }
-                }
-            }
+            buildJob(".build/docker/build-debian.sh", "cassandra_*,cassandra-tools_*")
           }
         }
-        stage('long units') {
+        stage('redhat') {
```
jmckenzie-dev: Why do we have buildJobs listed under the Test stage? Will that cause issues on other platforms, or is there something I'm missing here?

.jenkins/Jenkinsfile
```diff
@@ -16,524 +16,486 @@
...
+    //  }
+    // }
+    stage('Summary') {
+      steps {
+       // FIXME is this needed if each job is doing it ??
+       //junit testResults: '**/build/test/**/TEST*.xml,**/cqlshlib.xml,**/nosetests.xml', testDataPublishers: [[$class: 'StabilityTestDataPublisher']]
+
+       //sendNotifications()
+       
+       // FIXME how to collect and build one test report file ??
```
jmckenzie-dev: So we have this method called below but it looks like it's doing this for a given specific test axis. Is the outstanding question "how do we do this for multiple axes generated by the testJob"?

.jenkins/Jenkinsfile
```diff
@@ -16,524 +16,486 @@
...
+ * Return the supported JDKs defined by build.xml
+ **/
+def javaVersionsSupported() {
+  sh (returnStdout: true, script: 'grep \'property\\s*name=\"java.supported\"\' build.xml | sed -ne \'s/.*value=\"\\([^\"]*\\)\".*/\\1/p\'').trim()
+}
+
+/**
+ * Is this a post-commit build (or a pre-commit build)
+ **/
+def isPostCommit() {
```
jmckenzie-dev: Could use a bit more description for why this works. Because I don't get it. :)

.jenkins/Jenkinsfile
```diff
@@ -16,524 +16,486 @@
...
+  echo "Spawning `${test_script} ${test_type} ${split} ${jdk}` with arch=${arch} jdk=${jdk} python=${python} cython=${cython}"
+  node(arch) {
+    echo "Starting `${test_script} ${test_type} ${split} ${jdk}` with arch=${arch} jdk=${jdk} python=${python} cython=${cython} on ${NODE_NAME}"
+    cleanWs()
+    checkout scm
+    tool type: 'jdk', name: "jdk_${jdk}_latest"
+    // FIXME python
+    // FIXME cython
+    def statusCode = 0, attempt = 1
+    retry(2) {
```
jmckenzie-dev: Why retry 2?

.jenkins/Jenkinsfile
```diff
@@ -16,524 +16,486 @@
...
+      if (attempt > 1) { sleep(60 * attempt) }
+      def logfile = "${JOB_NAME}_${BUILD_NUMBER}_${test_type}_jdk${jdk}_python_${python}_${cython}_${arch}_attempt_${attempt}.log"
+      try {
+        //("${JOB_NAME}-${test_type}-${BUILD_NUMBER}")
```
jmckenzie-dev: Commented out code

.jenkins/Jenkinsfile
```diff
@@ -16,524 +16,486 @@
...
+  node(arch) {
+    echo "Starting `${test_script} ${test_type} ${split} ${jdk}` with arch=${arch} jdk=${jdk} python=${python} cython=${cython} on ${NODE_NAME}"
+    cleanWs()
+    checkout scm
+    tool type: 'jdk', name: "jdk_${jdk}_latest"
+    // FIXME python
+    // FIXME cython
+    def statusCode = 0, attempt = 1
+    retry(2) {
+      if (attempt > 1) { sleep(60 * attempt) }
```
jmckenzie-dev: Would it make sense to cleanWs() after a failed attempt before the retry?

.jenkins/Jenkinsfile
```diff
@@ -16,524 +16,486 @@
...
+        sh "xz -f *_attempt_*.log"
+        copyToNightlies("*_attempt_*.log.xz,console.log.xz,TESTS-TestSuites.xml.xz,build/test/logs/**,${toCopy}", "${test_type}/${arch}/jdk${jdk}/${python}/cython_${cython}/" + "split_${split}".replace("/", "_"))
+        cleanAgent(test_script)
+        if (0 != statusCode) { 
+          currentBuild.result = 'FAILURE'
+          error("Failed `${test_script} ${test_type} ${split} ${jdk}` with arch=${arch} jdk=${jdk} python=${python} cython=${cython} on ${NODE_NAME}") 
+        }
       }
     }
+    //if (build_result.result != 'SUCCESS') unstable("${build_name} failures")
```
jmckenzie-dev: Commented out code

.jenkins/Jenkinsfile
```diff
@@ -16,524 +16,486 @@
...
   }
+  echo "archived to https://nightlies.apache.org/${remotePath}"
+}
+
+def cleanAgent(job_name) {
+  def maxJobHours = 12
+  echo "Cleaning project, processes, docker for '${job_name}' on ${NODE_NAME}…" ;
+  sh """
+      git clean -qxdff -e build/test/jmh-result.json || true;
+      if pgrep -xa docker || pgrep -af "build/docker" || pgrep -af "cassandra-builds/build-scripts" ; then docker system prune --all --force --filter "until=${maxJobHours}h" || true ; else  docker system prune --force --volumes || true ;  fi;
```
jmckenzie-dev: Do we run the risk of a lot of back to back jobs filling up hosts since we only prune once > 12 hours? Or is the idea that other build_script / jdk combos could still be legitimately running when we go to clean it off?

README.asc
```diff
@@ -11,10 +11,7 @@ For more information, see http://cassandra.apache.org/[the Apache Cassandra web
 
 Issues should be reported on https://issues.apache.org/jira/projects/CASSANDRA/issues/[The Cassandra Jira].
 
-Requirements
-------------
-. Java >= 1.8 (OpenJDK and Oracle JVMS have been tested)
-. Python 3.6+ (for cqlsh)
+
```
jmckenzie-dev, line 18: Remove Requirements entirely or should they be replaced w/new requirements?

jmckenzie-dev (commented):
One last general thought:
- The amount of duplication between build-*.sh files is worrying. Could we have a shared lib .sh files with some of the basic flow put into methods that each of the files then calls?
	- For instance: duplication of regex_tag and _tentative is concerning. Be much easier if we had one place to modify those instead of multiple of our naming evolves.
	- And consider combining them into something such as: `cassandra-(([0-9.]+)(-(alpha|beta|rc)[0-9]+)?)(-tentative)?$`

So - structurally sound, the bones are there and it looks really good. There's a lot of stuff that's just hygiene and tidying (as well as explaining the thinking behind some redundancy, retries, etc) and some cleaning up for parse-ability and maintainability.

All told it's looking really solid. Appreciate all the hard work on this!

.build/build-rat.xml
michaelsembwever: not used anywhere. (and underscore prefix mean it was only ever meant to be internal)

.build/docker/bullseye-build.docker
michaelsembwever: No, just expecting jdk8 will be dropped before this gets merged. (easy to fix later if not)

.build/docker/run-tests.sh
michaelsembwever: because… exactly your comment https://github.com/apache/cassandra/pull/2351/files#r1199137328

.build/docker/run-tests.sh
michaelsembwever: the test scripts are still wip.

here, testing different python versions isn't implemented.

.build/docker/run-tests.sh
michaelsembwever: these docker loops (inner splits) may go altogether…

.build/docker/run-tests.sh
michaelsembwever: no, they are optimistic (not all logs exists in all test types…)

.build/run-tests.sh
michaelsembwever: all these scripts are intended to have the same name, the docker folder being the "docker" context. see the readme

.jenkins/Jenkinsfile
michaelsembwever: better throughput. we want as much parallel as possible

.jenkins/Jenkinsfile
michaelsembwever: still testing. testing scripts still wip.

.jenkins/Jenkinsfile
michaelsembwever: dev branches (i.e. pre-commit) won't be found in the origin "apache" repo.

.jenkins/Jenkinsfile
michaelsembwever: just testing

README.asc
michaelsembwever: this is just a readme file, which will always fall out of date.

this information can be found in other places (where it is more likely to be correct)

.jenkins/Jenkinsfile
michaelsembwever: totally agree.

michaelsembwever:
comments are good. i'm still wip on the testing scripts, so will come back to address many of the comments once that stuff it ironed out

.build/docker/run-tests.sh
jmckenzie-dev: Gotcha. Any chance we could augment the comment here in the interim with something about what FIXME looks like in the future?

.build/docker/run-tests.sh
jmckenzie-dev: Same as in one of the other cases - you up for a quick jot down of a comment on context for what comes next? That way someone could collab w/you on it. :)

.build/docker/run-tests.sh
jmckenzie-dev: Ah, gotcha. A comment around the shape of that would be good for future maintainers.

README.asc
jmckenzie-dev: Could we just link from the README to the related files where things will be more likely to be correct rather than just having inconsistent freshness on things here?

michaelsembwever:
> The amount of duplication between build-*.sh files is worrying.

can this be follow-up? i'll be pushed for time to get this tested and working… once in-tree, improving is much easier.

.build/docker/almalinux-build.docker
michaelsembwever: magic number. what we were already using in cassandra-builds

.jenkins/Jenkinsfile
michaelsembwever: not sure. not how it's done today. keen to defer.

.jenkins/Jenkinsfile
michaelsembwever: 12 hours because docker cache++ can speed things up.
this is how it is today. i expect this to evolve a lot once we're running in k8s and can benchmark and debug better.

.build/docker/almalinux-build.docker
michaelsembwever: Yes, that's why they're in-tree, and why the image tags and container names include these variables.

.build/docker/bullseye-build.docker
michaelsembwever: magic numbers :) 
see my comment above

.build/docker/centos7-build.docker
michaelsembwever: defer (please).  will be easier once in-tree

.build/docker/run-tests.sh
michaelsembwever: yes. and quite likely yes (currently is disabled for this reason).

.build/docker/ubuntu2004_test.docker
michaelsembwever: 🤷🏻‍♀️

jmckenzie-dev:
> > The amount of duplication between build-*.sh files is worrying.
> 
> can this be follow-up? i'll be pushed for time to get this tested and working… once in-tree, improving is much easier.

Sure. Let's just make sure we actually capture that debt in a JIRA that follows on from this work so it doesn't fall by the wayside.

.build/docker/almalinux-build.docker
jmckenzie-dev: So to make sure I understand, the image tags / container names include variables that'll constrain `java-1.8.0-openjdk-devel` to a singular reproducible version on that env?

.build/docker/almalinux-build.docker
jmckenzie-dev: If a magic number can we document something around it as to why? Even just calling out that we're inheriting it from the old build script without context?

.build/run-tests.sh
jmckenzie-dev: Could you name them:
```
docker/docker-run-tests.sh
./build/build-run-tests.sh
```
or some other permutation? Do they _have_ to have the same name?

.jenkins/Jenkinsfile
jmckenzie-dev: > dev branches (i.e. pre-commit) won't be found in the origin "apache" repo.

Comment that in the script?

.jenkins/Jenkinsfile
jmckenzie-dev: Capture in a follow up debt JIRA on this?

.build/docker/almalinux-build.docker
michaelsembwever: image names are like: `apache/cassandra-ubuntu2004_test:b09545573ea248c487eba7547094821d` 
 where the tag is a md5sum of the dockerfile.

since each each supports the different axis values.

container names are like contain the values, plus a random hex, so they are easy to identifer.
ref:
 - https://github.com/apache/cassandra/pull/2351/files#diff-e8f0cb5844d900a3d89a43743814930e66575abc2b6731c07d6a691de51a4d65R97
 -  https://github.com/apache/cassandra/pull/2351/files#diff-d6af0f7ab7ddf83429e201fad8288dbdfb90fbd89fd9cf1bbb74fd06a65eb0d8R189

(i will make those lines easier to read…)

The idea here is that the tag (md5sum) allows us to
- deploy the image to dockerhub, allowing it be be downloaded instead of built,
- be easily cached between builds,
- provide reproducible builds (answers your concern most IIUC)
- be able to patch the dockerfile and have it used in the CI run without fuss

.build/docker/run-tests.sh
michaelsembwever: for now just going with `random_string` as that's the biggest win here (the others only make the line longer).

.build/build-artifacts.sh
```diff
@@ -0,0 +1,29 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# variables, with defaults
+[ "x${CASSANDRA_DIR}" != "x" ] || CASSANDRA_DIR="$(readlink -f $(dirname "$0")/..)"
+
+# pre-conditions
+command -v ant >/dev/null 2>&1 || { echo >&2 "ant needs to be installed"; exit 1; }
+command -v pip >/dev/null 2>&1 || { echo >&2 "pip needs to be installed"; exit 1; }
```
driftx: I don't think pip is required here, I don't use it to build C* and I commented it out just fine.

michaelsembwever:
TODOs:
- cython parameterisation
- decide whether to keep inner splitting or not… (it's a significant throughput increase for unit tests on over-capacity hosts)

.build/build-owasp.xml
driftx: I think 3 is reasonable before giving up.  Owasp fails on downloading stuff fairly often so retrying makes sense.

.build/build-debian.sh
```diff
@@ -0,0 +1,127 @@
...
+#
+################################
+
+# variables, with defaults
+[ "x${CASSANDRA_DIR}" != "x" ] || CASSANDRA_DIR="$(readlink -f $(dirname "$0")/..)"
+[ "x${DIST_DIR}" != "x" ] || DIST_DIR="${CASSANDRA_DIR}/build"
+
+# pre-conditions
+command -v ant >/dev/null 2>&1 || { echo >&2 "ant needs to be installed"; exit 1; }
+command -v pip >/dev/null 2>&1 || { echo >&2 "pip needs to be installed"; exit 1; }
```
driftx: You don't need pip here... but are we going to build packages outside of docker?  I think this should be in docker/build-debian.sh only so nobody tries to run this outside of docker.

.build/build-jars.sh
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+
+# temporary between CASSANDRA-18133 and CASSANDRA-18594
+
+# variables, with defaults
+[ "x${CASSANDRA_DIR}" != "x" ] || CASSANDRA_DIR="$(readlink -f $(dirname "$0")/..)"
+
+# pre-conditions
+command -v ant >/dev/null 2>&1 || { echo >&2 "ant needs to be installed"; exit 1; }
+command -v pip >/dev/null 2>&1 || { echo >&2 "pip needs to be installed"; exit 1; }
```
driftx: No pip needed again.

.build/build-redhat.sh
```diff
@@ -0,0 +1,123 @@
...
+################################
+
+# variables, w/ defaults, w/ checks
+[ "x${CASSANDRA_DIR}" != "x" ] || CASSANDRA_DIR="$(readlink -f $(dirname "$0")/..)"
+[ "x${DIST_DIR}" != "x" ] || DIST_DIR="${CASSANDRA_DIR}/build"
+[ "x${RPM_BUILD_DIR}" != "x" ] || RPM_BUILD_DIR="$(mktemp -d /tmp/rpmbuild.XXXXXX)"
+
+# pre-conditions
+command -v ant >/dev/null 2>&1 || { echo >&2 "ant needs to be installed"; exit 1; }
+command -v pip >/dev/null 2>&1 || { echo >&2 "pip needs to be installed"; exit 1; }
```
driftx: Pip again, but like debs I question when we would run this outside of docker.

.build/docker/centos7-build.docker
driftx: That's not what docker-compose does and I don't think it's worthy DRYing this up, it's not that much repetition and each docker file stands alone well; different distros may diverge.

.build/docker/build-debian.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if [ "$1" == "-h" ]; then
+   echo "$0 [-h] [<java_version>]"
+   echo " build debian packages"
+   exit 1
+fi
+
+echo
+echo "==="
+echo "WARNING: this script modifies local versioned files"
```
driftx, line 27: What is this referring to?  This warning exists for running in docker but not outside of it?

.build/run-python-dtests.sh
```diff
@@ -0,0 +1,160 @@
...
+export CCM_HEAP_NEWSIZE="512M"
+export CCM_CONFIG_DIR=${CASSANDRA_DIR}/build/.ccm
+export NUM_TOKENS="16"
+#Have Cassandra skip all fsyncs to improve test performance and reliability
+export CASSANDRA_SKIP_SYNC=true
+export TMPDIR="${DIST_DIR}/tmp"
+
+# pre-conditions
+command -v ant >/dev/null 2>&1 || { echo >&2 "ant needs to be installed"; exit 1; }
+command -v pip3 >/dev/null 2>&1 || { echo >&2 "pip3 needs to be installed"; exit 1; }
```
driftx: I don't think pip3 is needed since we're going to activate a virtualenv that has it.

.build/run-tests.sh
```diff
@@ -0,0 +1,220 @@
...
+command -v ant >/dev/null 2>&1 || { echo >&2 "ant needs to be installed"; exit 1; }
+command -v git >/dev/null 2>&1 || { echo >&2 "git needs to be installed"; exit 1; }
+[ -d "${CASSANDRA_DIR}" ] || { echo >&2 "Directory ${CASSANDRA_DIR} must exist"; exit 1; }
+[ -f "${CASSANDRA_DIR}/build.xml" ] || { echo >&2 "${CASSANDRA_DIR}/build.xml must exist"; exit 1; }
+[ -d "${DIST_DIR}" ] || { mkdir -p "${DIST_DIR}" ; }
+
+# print debug information on versions
+ant -version
+git --version
+pip --version
```
driftx: There is no check for pip before this, but I question if pip is required.

redhat/noboolean/cassandra.spec
```diff
@@ -69,7 +69,7 @@ Cassandra is a distributed (peer-to-peer) system for the management and storage
 %build
 export LANG=en_US.UTF-8
 export JAVA_TOOL_OPTIONS="-Dfile.encoding=UTF-8"
-ant clean jar -Dversion=%{upstream_version}
+ant jar -Dversion=%{upstream_version} -Dno-checkstyle=true -Drat.skip=true -Dant.gen-doc.skip=true
```
driftx, line 75: why not clean? shouldn't be needed but won't hurt?

driftx (changes requested):

.build/run-python-dtests.sh
michaelsembwever: pip3 is required inside a new virtualenv, see line 89-92
 doesn't that mean pip3 is needed generally…?

redhat/noboolean/cassandra.spec
michaelsembwever: these scripts now intentionally don't clean. so that they can be used (sequentially) together.

for example the jenkins impl on top uses a stash from the first build, and unstashes that onto every subsequent build/test.

i haven't thought too much about the deb/rpm building, but figured constraints of "only building it from clean" should generally be implied and imposed by the caller. 🤷

.build/run-python-dtests.sh
driftx: The virtualenv provides it.  If you tell me you need pip, that's my signal to setup a venv so you don't mess with my local python...but you are going to setup a venv anyway.

michaelsembwever:
> TODOs:
> - cython parameterisation
> - decide whether to keep inner splitting or not… (it's a significant throughput increase for unit tests on over-capacity hosts)

both done (inner splitting is removed. removes a lot of complexity.  testing showed that a split would only improve throughput ~20% at most. there are other ways we can improve unit testing throughput)

.build/docker/almalinux-build.docker
```diff
@@ -0,0 +1,59 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM almalinux:8
+MAINTAINER Apache Cassandra <dev@cassandra.apache.org>
+
+# CONTEXT is expected to be cassandra/.build
+
+ENV BUILD_HOME=/home/build
```
jmckenzie-dev, line 22: It looks like some of the "docker free" build scripts are relying on these environment variables being set. Could we separate out the expected env vars to be used in environments where someone's not expecting to use docker to decouple this implicit coupling?

Actually, scratch that; I see that they default to reading local dir pathing so they'll work if they're relatively pathed.

.build/docker/_build-debian.sh
```diff
@@ -0,0 +1,129 @@
...
+      else
+         echo "Warning: could not recognize version from branch. dpkg version is $git_version"
+      fi
+   fi
+    # if CASSANDRA_VERSION is -alphaN, -betaN, -rcN, then rpmbuild fails on the '-' char; replace with '~'
+    CASSANDRA_VERSION=${buildxml_version/-/\~}
+    dt=`date +"%Y%m%d"`
+    ref=`git rev-parse --short HEAD`
+    CASSANDRA_REVISION="${dt}git${ref}"
+    dch -D unstable -v "${CASSANDRA_VERSION}-${CASSANDRA_REVISION}" --package "cassandra" "TODO msg"
```
Mmuzaf: Did you explicitly add `TODO`? I guess the idea was to fetch it from the commit message. It also touches the `changelog` file during the build making them dirty, I don't know was it what was intended? (_edited_: found the message in log).

.build/run-tests.sh
```diff
@@ -0,0 +1,222 @@
...
+  grep "name=\"${1}\"" build.xml | awk -F'"' '{print $4}'
+}
+
+_build_all_dtest_jars() {
+    pushd $TMP_DIR >/dev/null
+    rm -fR ${TMP_DIR}/cassandra-dtest-jars
+    until git clone --quiet --depth 1 --no-single-branch https://github.com/apache/cassandra.git cassandra-dtest-jars ; do echo "git clone failed… trying again… " ; done
+
+    # cassandra-4 branches need CASSANDRA_USE_JDK11 to allow jdk11
+    [ "${java_version}" -eq 11 ] && export CASSANDRA_USE_JDK11=true
```
Mmuzaf, line 84: Do we still need this line if the [CASSANDRA-18133](https://issues.apache.org/jira/browse/CASSANDRA-18133) is targeted only for 5.x ?

.build/run-tests.sh
```diff
@@ -0,0 +1,222 @@
...
+
+    pushd ${TMP_DIR}/cassandra-dtest-jars >/dev/null
+    for branch in cassandra-4.0 cassandra-4.1 trunk ; do
```
Mmuzaf, line 87: This is the same as the comment below, do we need it in 5.x? https://github.com/apache/cassandra/pull/2351/files/54040fb3b9d988b50401dd1cf1cb6d0791accf93#r1258606866

.build/README.md
```diff
@@ -0,0 +1,134 @@
...
+    .build/docker/run-tests.sh dtest-novnode
+    .build/docker/run-tests.sh dtest-offheap
+    .build/docker/run-tests.sh dtest-large
+    .build/docker/run-tests.sh dtest-large-novnode
+    .build/docker/run-tests.sh dtest-upgrade
+
+
+Running python dtests without docker:
+
+    .build/docker/run-python-dtests.sh dtest
```
Mmuzaf:
```suggestion
    .build/run-python-dtests.sh dtest
```

.build/README.md
```diff
@@ -0,0 +1,134 @@
...
+Build without docker:
+
+    .build/build-debian.sh
+    .build/build-redhat.sh
+
+
+Build in docker with a specific jdk:
+
+    .build/docker/build-debian.sh 11
+    .build/docker/build-redhat.sh 11
```
Mmuzaf:
```suggestion
    .build/docker/build-redhat.sh rpm 11
```

.build/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+
+Running python dtests without docker:
+
+    .build/docker/run-python-dtests.sh dtest
+
+
+Other test types without docker:
+
+    .build/docker/run-tests.sh test
```
Mmuzaf:
```suggestion
    .build/run-tests.sh test
```

.build/docker/centos7-build.docker
```diff
@@ -0,0 +1,88 @@
...
+ARG UID_ARG=1000
+ARG GID_ARG=1000
+
+LABEL org.cassandra.buildenv=centos
+
+VOLUME ${DIST_DIR}
+VOLUME ${CASSANDRA_DIR}
+
+RUN echo "Building with arguments:" \
+    && echo " - DEB_DIST_DIR=${DEB_DIST_DIR}" \
```
jmckenzie-dev: Should this read `DIST_DIR` instead of `DEB_DIST_DIR`?

.build/docker/almalinux-build.docker
```diff
@@ -0,0 +1,59 @@
...
+ARG UID_ARG=1000
+ARG GID_ARG=1000
+
+LABEL org.cassandra.buildenv=almalinux
+
+VOLUME ${DIST_DIR}
+VOLUME ${CASSANDRA_DIR}
+
+RUN echo "Building with arguments:" \
+    && echo " - DEB_DIST_DIR=${DEB_DIST_DIR}" \
```
jmckenzie-dev: Same as in the centos file; looks like we're using `DIST_DIR` and not using `DEB_DIST_DIR`, so the printed output won't be representative of anything used.

Mmuzaf (approved):

.build/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+Build with docker:
+
+    .build/docker/build-debian.sh
+    .build/docker/build-redhat.sh
+
+
+Build without docker:
+
+    .build/build-debian.sh
```
smiklosovic: @michaelsembwever these scripts are not in `.build` directory.

michaelsembwever: fixed

.build/README.md
michaelsembwever: fixed

.build/build-resolver.xml
michaelsembwever: magic number. just carried over from what was previously happening in the sc

.build/README.md
michaelsembwever: fixed

.build/README.md
michaelsembwever: fixed

.build/docker/almalinux-build.docker
michaelsembwever: fixed

.build/docker/centos7-build.docker
michaelsembwever: fixed

.build/run-tests.sh
michaelsembwever: yes, because this is building older branches

.build/run-tests.sh
michaelsembwever: which comment? (link is lost :( )

README.asc
michaelsembwever: fixed

PR #2360, opened 2023-05-23
Fixing some doc typos and fixing inconsistent headers.

jeromatron:


polandll (approved):
LGTM - looks like files were included before at a different level, but reorganized.

PR #2368, opened 2023-05-26
Track the amount of read data per row

jacek-lewandowski:
If an sstable is corrupted in a nasty way, we may read invalid cell sizes and try to read much more data for a row than we should. In rare scenarios this can lead even to OOMs.

This simple fix adds tracking and limiting the amount of data that is read per row. Row has its size stored in preamble which can be used as a limit. If the deserialization code tries to read more than that, it will simply fail with EOF which will prevent more serious problems later.

```

```

### DIFF
--- a/src/java/org/apache/cassandra/io/util/TrackedDataInputPlus.java
+++ b/src/java/org/apache/cassandra/io/util/TrackedDataInputPlus.java
@@ -19,7 +19,11 @@ package org.apache.cassandra.io.util;
 
 import java.io.DataInput;
 import java.io.DataInputStream;
+import java.io.EOFException;
 import java.io.IOException;
+
+import net.nicoulaj.compilecommand.annotations.Inline;
+import org.apache.cassandra.db.TypeSizes;
 
 /**
  * This class is to track bytes read from given DataInput
@@ -27,11 +31,25 @@ import java.io.IOException;
 public class TrackedDataInputPlus implements DataInputPlus, BytesReadTracker
 {
     private long bytesRead;
+    private final long limit;
     final DataInput source;
 
+    /**
+     * Create a TrackedDataInputPlus from given DataInput with no limit of bytes to read
+     */
     public TrackedDataInputPlus(DataInput source)
     {
+        this(source, -1);
+    }
+
+    /**
+     * Create a TrackedDataInputPlus from given DataInput with limit of bytes to read. If limit is reached
+     * {@link IOException} will be thrown when trying to read more bytes.
+     */
+    public TrackedDataInputPlus(DataInput source, long limit)
+    {
         this.source = source;
+        this.limit = limit;
     }
 
     public long getBytesRead()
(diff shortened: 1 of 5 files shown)


src/java/org/apache/cassandra/io/util/TrackedDataInputPlus.java
```diff
@@ -49,53 +66,61 @@ public void reset(long count)
 
     public boolean readBoolean() throws IOException
     {
+        checkCanRead(1);
```
bereng: Do you want to use `TypeSizes` instead here?

test/unit/org/apache/cassandra/db/rows/UnfilteredSerializerTest.java
```diff
@@ -0,0 +1,135 @@
...
+            out.write(replacement);
+            out.close();
+            return out.asNewBuffer();
+        }
+        catch (IOException ex)
+        {
+            throw new RuntimeException(ex);
+        }
+    }
+}
```
bereng: Missing new line?

bereng:
We need CI runs and will you port this to 4.0 and 4.1 also?

src/java/org/apache/cassandra/io/util/TrackedDataInputPlus.java
jacek-lewandowski: well, I don't think so, that would require also refactoring that in `bytesRead` updates

jacek-lewandowski:
@bereng thanks for super quick response. I'll provide patches for older branches but probably it will be next week

src/java/org/apache/cassandra/io/util/TrackedDataInputPlus.java
Maxwell-Guo: I think it is more clear to use the size number 1/2/4/8 ,which make the code more readable as the code below just add these numbers

src/java/org/apache/cassandra/io/util/TrackedDataInputPlus.java
```diff
@@ -129,22 +156,34 @@ public String readUTF() throws IOException
...
-        int skipped = source.skipBytes(n);
+        int skipped = source.skipBytes(limit < 0 ? n : (int) Math.min(limit - bytesRead, n));
         bytesRead += skipped;
         return skipped;
     }
+
+    @Inline
+    private void checkCanRead(int size) throws IOException
+    {
+        if (limit >= 0 && bytesRead + size > limit)
```
Maxwell-Guo: will limit just be zero？

test/unit/org/apache/cassandra/db/rows/UnfilteredSerializerTest.java
```diff
@@ -0,0 +1,135 @@
...
+             DataInputBuffer replIn = new DataInputBuffer(replacement, false))
+        {
+            out.writeByte(origIn.readByte()); // flag
+            out.writeUnsignedVInt(origIn.readUnsignedVInt()); // row size
+            replIn.readByte(); // skip flag
+            replIn.readUnsignedVInt(); // skip row size
+            out.write(replacement);
+            out.close();
+            return out.asNewBuffer();
+        }
```
Maxwell-Guo: I think a test for some corner case is needed such as what I said when limit is 0？

src/java/org/apache/cassandra/io/util/TrackedDataInputPlus.java
Maxwell-Guo: I just test some times,it seems limit can not be 0 ~~~

jacek-lewandowski: Yes, we always encode at least a flag

src/java/org/apache/cassandra/io/util/TrackedDataInputPlus.java
bereng: Not following. Why would you need to refactor anything? You guys think `1` is more readable than `TypeSizes.BOOL_SIZE` and on top of that you're hardcoding a value? It's like magic numbers where we have a proper abstraction ready unless I am missing sthg.

`checkCanRead(TypeSizes.BOOL_SIZE)` reads good in my eyes :shrug:

jacek-lewandowski:
@bereng is it ok now?

src/java/org/apache/cassandra/io/util/TrackedDataInputPlus.java
Maxwell-Guo: I agree with your point of view. From the perspective of code readability, TypeSizes.BOOL_SIZE is definitely much better than hard code 1.
What I said above means that if the existing code is already written as 1/4/8, our newly added code also fills in 1/4/8, then it just echoes back and forth, so the readability is not bad.
Of course, if Jacek is willing to change the existing numbers to TypeSizes, I definitely support your point of view.

Maxwell-Guo: (1)`checkCanRead(TypeSizes.BOOL_SIZE);
        byte b = source.readByte();
        bytesRead += TypeSizes.BOOL_SIZE;
        return b;`
 (2) `checkCanRead(1);
        byte b = source.readByte();
        bytesRead += 1;
        return b;`
(3)`checkCanRead(TypeSizes.BOOL_SIZE);
        byte b = source.readByte();
        bytesRead += 1;
        return b;`
I think the code itself should explain its behavior. 
just from my point of view, this order is (1) > (2) > (3) 
and it seems Jacek has already resolved this issue.

test/unit/org/apache/cassandra/db/rows/UnfilteredSerializerTest.java
Maxwell-Guo: No need for this

Maxwell-Guo (approved):

bereng:
Approach looks good to me. CI and PRs for the rest of the versions are missing iiuc.

jacek-lewandowski:
@bereng there was no point in running CI before code review, same for other PRs

bereng:
Ha you're brave. I always run CI before...

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/720/workflows/7495bccf-c7f8-47db-a145-e0e5f6e508c2

bereng:
^That is j11 only, j8 seems to not have been triggered?

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/720/workflows/f6fd1707-5cd1-4680-b234-0a79939fb4d2

PR #2369, opened 2023-05-26
CASSANDRA-17194 trunk: Add guardrail for partition tombstones and deprecate compaction_tombstone_warning_threshold

adelapena:


src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -567,6 +568,31 @@ public void setPartitionSizeThreshold(@Nullable DataStorageSpec.LongBytesBound w
...
+    @Override
+    public long getPartitionTombstonesFailThreshold()
+    {
+        return config.partition_tombstones_fail_threshold;
+    }
+
+    public void setPartitionTombstonesThreshold(long warn, long fail)
+    {
+        validateMaxLongThreshold(warn, fail, "partition_tombstones", false);
+        updatePropertyWithLogging("partition_tombstones_warn_threshold",
```
Maxwell-Guo, line 586: There is one thing I want to discuss, why do we use the name of this variable attribute here， see "partition_tombstones_warn_threshold" ? Because I feel that it is easy for developers to spell mistakes, and if the test does not cover it, there may be some problems。

Or we can use just some map which key is property name value is the property value or some enum class ？

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -323,6 +323,18 @@ public final class Guardrails implements GuardrailsMBean
                              format("Partition %s has size %s, this exceeds the %s threshold of %s.",
                                     what, value, isWarning ? "warning" : "failure", threshold));
 
+    /**
+     * Guardrail on the number of rows of a partition.
+     */
+    public static final MaxThreshold partitionTombstones =
+    new MaxThreshold("partition_tombstones",
+                     "Partitions with too many tombstones can cause performance problems. ",
```
Maxwell-Guo: There is an extra space at the end of the sentence：
`"Partitions with too many tombstones can cause performance problems. ",` 
can be changed to 
`"Partitions with too many tombstones can cause performance problems.",`

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailPartitionSizeTest.java
```diff
@@ -75,12 +75,11 @@ protected Cluster getCluster()
     @Test
     public void testPartitionSize()
     {
+        // test yaml-loaded config
         testPartitionSize(WARN_THRESHOLD, FAIL_THRESHOLD);
-    }
+        schemaChange("DROP TABLE %s");
```
Maxwell-Guo, line 81: one more question that may have no relation with this issue : why merge the two test case into one ?"test yaml-loaded config" and "test dynamic config"
as https://github.com/apache/cassandra/commit/b0aa44b27da97b37345ee6fafbee16d66f3b384f described 
there are two cases

NEWS.txt
```diff
@@ -187,6 +188,9 @@ Deprecation
     - The config property `compaction_large_partition_warning_threshold` has been deprecated in favour of the new
       guardrail for partition size. That guardrail is based on the properties `partition_size_warn_threshold` and
       `partition_size_fail_threshold`. The warn threshold has a very similar behaviour to the old config property.
+    - The config property `compaction_tombstone_warning_threshold` has been deprecated in favour of the new guardrail
```
bereng, line 197: Can you amend the description of the JIRA ticket to reflect the correct names. Otherwise I'm getting lost in names :-DD

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailPartitionSizeTest.java
adelapena: The tests extending `o.a.c.distributed.test.guardrails.GuardrailTester` use a single cluster for the entire suite. The regular test case uses the yaml configuration set at startup, the one that is set at `setupCluster`. In contrast, the test for dynamic update changes that config on the running instance by simulating the JMX methods.

That works if the test for static config runs before the test for dynamic update, but not the other way around. If the test for dynamic config runs first, the test for static config will find the dynamically updated config, breaking the test. We could simply restart the node before each test, but I think the test will save resources if we simply combine both tests. I think the tests are brief enough to remain readable.

So `GuardrailPartitionSizeTest` only worked incidentally because the tests were run in the fortunate order. The new `GuardrailPartitionTombstonesTest` however was not so fortunate, exposing the bug. So I have changed both to make them independent of whatever order the test runner decides for us.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
adelapena: Right, good catch. Fixed it here and in `partitionSize`.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
adelapena: Indeed it's not ideal that we don't know the name of the yaml property that will be written to the log output. The original prototype for guardrails had a much more robust system of proper config objects for every guardrail, instead of the current flat system. However, new nested configs were vetoed by a committer until we moved the entire config file to a nested structure, even if we already have some of them in the yaml. We haven't been still able to settle on that new nested format, but I guess that when/if that happens this will be less cumbersome.

I guess that for not repeating the name of the property we could try to read its name from the `Config` var, or something like that. In any case, that looks like a general refactor of the framework to be done in a separate ticket.

NEWS.txt
adelapena: Those examples on the JIRA ticket are quite old and they use the initial nested syntax that was vetoed. I have updated them to the new flat format with the same names that we use on the patch.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
Maxwell-Guo: got it

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailPartitionSizeTest.java
Maxwell-Guo: Thank you for your answer

PR #2370, opened 2023-05-26
CASSANDRA-18239: Incorporate CheckerFramework for static analysis

jacek-lewandowski:
This introduced CheckerFramework for static code analysis, intended to replace Eclipse-Warnings. 

Although there is no perfect tool that we can use, CF provides better accuracy and fewer false positives. Additionally, it lets the developer add special annotations to inform about a resource ownership transfer from one object to another, which helps determine who is responsible for closing the resource.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -547,11 +547,37 @@
         <copy todir="${basedir}/conf" file="${build.classes.main}/META-INF/hotspot_compiler"/>
     </target>
 
-    <target name="check" depends="_main-jar,build-test" description="Verifies the source code and dependencies. This task is intended to run on pre-commit and locally. It should verify mostly modified files compared to the upstream base branch." unless="check.skip">
+    <target name="cf-all" depends="build,fqltool-build,stress-build" description="Run CheckerFramework against all production source files. This goal is intended to be run on CI.">
+        <run-cf includes="**" listfiles="false" src="${build.src.java}" outputDir="${checkerframework.output.dir}/main"/>
+        <run-cf includes="**" listfiles="false" src="${fqltool.build.src}" outputDir="${checkerframework.output.dir}/fqltool"/>
+        <run-cf includes="**" listfiles="false" src="${stress.build.src}" outputDir="${checkerframework.output.dir}/stress"/>
+    </target>
+
+    <target name="cf" depends="-git-init,build,fqltool-build,stress-build" description="Run CheckerFramework against all changed production source files compared to the latest common ancestor between HEAD and trunk. This goal is intended to be run locally before starting CI.">
+        <run-cf-changed src="${build.src.java}" outputDir="${checkerframework.output.dir}/main"/>
+        <run-cf-changed src="${fqltool.build.src}" outputDir="${checkerframework.output.dir}/fqltool"/>
+        <run-cf-changed src="${stress.build.src}" outputDir="${checkerframework.output.dir}/stress"/>
+    </target>
+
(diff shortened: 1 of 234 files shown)


jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/743/workflows/56c4dc00-7c8e-48c9-8227-e14a00dd772f

pkolaczk (commented):
I haven't looked at all changes in detail, as the number of changes is really huge. 
But as a fan of static analysis in general I've got a few observations:

1. I really like the fact that this framework makes resource ownership and cleanup responsibilities visible in the code by means of `@Owning` / `@MustCall` annotations. This makes code more readable and the internal APIs harder to misuse.
2. The framework is much more powerful than just tracking resource ownership. I feel this PR is just a start, and we'll be adding more annotations gradually in the future to make better use of its full potential. 
3. I'm worried about the huge number of `@SuppressWarnings` though. I guess they are very unlikely to be fixed in the future once they are committed, because they basically hide problems. Suppressing warnings stands in a direct opposition to the static analysis and a good  framework wouldn't need to add many of those. Does the high number of suppressions simply stem from the lack of time to analyze each case in a 350 kLOC codebase, or is it an inherent limitation of the framework that's not capable of grasping the invariants in our code? I hope it is the former case. Maybe we should try to clean up those suppressions in a few components to test the robustness of the framework? If it turns out too hard to satisfy and people start adding suppressions everywhere, the net value of it would be very low if not even negative. 
4. Even with large number of suppressions that must be added now, I can still see it usable if we can make all the new code not require suppressions. There is probably not enough time now to fix all existing suppressions, so I'm fine committing as-is and using this checker framework for new code. 
5. Does it detect mutable sharing / thread safety issues? Unfortunately I can't see much about it in the docs, besides  locking (which we mostly don't use). So I'm kinda slightly disappointed, but anyways, the other checkers look nice.
6. `ant check` doesn't work for me:
```
BUILD FAILED
/home/pkolaczk/Projekty/DataStax/cassandra-oss/build.xml:552: The following error occurred while executing this line:
/home/pkolaczk/Projekty/DataStax/cassandra-oss/.build/build-resolver.xml:185: The following error occurred while executing this line:
/home/pkolaczk/Projekty/DataStax/cassandra-oss/.build/build-resolver.xml:87: Task [resolver:pom] failed after [3] attempts; giving up.
Error messages:
Reference all not found.
Reference all not found.
Reference all not found.
Reference all not found.
```

build.xml
```diff
@@ -2102,4 +2102,5 @@
   <import file="${build.helpers.dir}/build-rat.xml"/>
   <import file="${build.helpers.dir}/build-owasp.xml"/>
   <import file="${build.helpers.dir}/build-git.xml"/>
+  <import file="${basedir}/.build/build-checkerframework.xml"/>
```
smiklosovic: @jacek-lewandowski this can be "build.helpers.dir", no? as on the previous lines. build.helpers.dir is alias for basedir/.build

src/java/org/apache/cassandra/index/sasi/plan/Operation.java
```diff
@@ -71,7 +75,7 @@ public boolean apply(boolean a, boolean b)
 
     protected Operation left, right;
 
-    private Operation(OperationType operation,
+    private @MustCallAlias Operation(OperationType operation,
```
smiklosovic: other arguments should be aligned

PR #2372, opened 2023-05-26
CASSANDRA-18554 Adding Mutual Tls authenticators for client & internode connections

jyothsnakonisa:


src/java/org/apache/cassandra/auth/AuthKeyspace.java
```diff
@@ -69,6 +70,15 @@ private AuthKeyspace()
               + "member_of set<text>,"
               + "PRIMARY KEY(role))");
 
+    private static final TableMetadata IdentityToRoles =
+        parse(IDENTITY_TO_ROLES,
+              "mtls authorized identities lookup table",
+            "CREATE TABLE %s ("
```
yifan-c: indentation

conf/cassandra.yaml
```diff
@@ -150,7 +150,14 @@ batchlog_replay_throttle: 1024KiB
 #   users. It keeps usernames and hashed passwords in system_auth.roles table.
 #   Please increase system_auth keyspace replication factor if you use this authenticator.
 #   If using PasswordAuthenticator, CassandraRoleManager must also be used (see below)
-authenticator: AllowAllAuthenticator
+authenticator:
+  class_name : org.apache.cassandra.auth.AllowAllAuthenticator
```
yifan-c, line 154: format change. The original format should be accepted, otherwise it could prevent the instance to start when using the existing cassandra.yaml.

src/antlr/Parser.g
```diff
@@ -1872,6 +1896,12 @@ username
     | QUOTED_NAME { addRecognitionError("Quoted strings are are not supported for user names and USER is deprecated, please use ROLE");}
     ;
 
+identity
```
yifan-c, line 1905: What is the different between `IDENTITY` and `USERNAME`? Their definitions are identical.

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
```diff
@@ -0,0 +1,208 @@
...
+            }
+
+            final String identity = certificateValidator.identity(clientCertificateChain);
+            if (StringUtils.isEmpty(identity))
+            {
+                final String msg = "Unable to extract client identity from certificate for authentication";
+                nospamLogger.error(msg);
+                throw new AuthenticationException(msg);
+            }
+            final String role = identityCache.get(identity);
```
yifan-c: drop final

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
```diff
@@ -0,0 +1,208 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(MutualTlsAuthenticator.class);
+    private static final NoSpamLogger nospamLogger = NoSpamLogger.getLogger(logger, 1L, TimeUnit.MINUTES);
+    private static final String VALIDATOR_CLASS_NAME = "validator_class_name";
+    private static final String CACHE_NAME = "IdentitiesCache";
+    private final IdentityCache identityCache = new IdentityCache();
+    private final MutualTlsCertificateValidator certificateValidator;
+
+    public MutualTlsAuthenticator(Map<String, String> parameters)
+    {
+        final String certificateValidatorClassName = parameters.get(VALIDATOR_CLASS_NAME);
```
yifan-c, line 73: drop final

src/java/org/apache/cassandra/auth/MutualTlsCertificateValidator.java
```diff
@@ -0,0 +1,63 @@
...
+     */
+    boolean isValidCertificate(Certificate[] clientCertificateChain);
+
+    /**
+     * This method should provide logic to extract identity out of a certificate to perform mTLS authentication.
+     *
+     * An example of identity could be the following
+     *  - an identifier in SAN of the certificate like SPIFFE
+     *  - CN of the certificate
+     *  - any other fields in the certificate can be combined and be used as indentifier of the certificate
```
yifan-c:
```suggestion
     *  - any other fields in the certificate can be combined and be used as identifier of the certificate
```

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
```diff
@@ -0,0 +1,208 @@
...
+ * validator and configure it in cassandra.yaml.Below is an example on how to configure validator.
+ * note that this example uses SPIFFE based validator, It could be any other validator with any defined identifier format.
+ *
+ * Example:
+ * authenticator:
+ *   class_name : org.apache.cassandra.auth.MutualTlsAuthenticator
+ *   parameters :
+ *     validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
+ */
+
```
yifan-c: nit: remove the empty line

src/java/org/apache/cassandra/auth/IRoleManager.java
```diff
@@ -226,4 +227,59 @@ default Set<Role> getRoleDetails(RoleResource grantee)
      * For example, use this method to create any required keyspaces/column families.
      */
     void setup();
+
+    /**
+     * Each valid identity is associated with a role in the identity_to_role table, this method returns role
+     * of a given identity
+     *
+     * @param identity identity whose role to be retreived
```
yifan-c:
```suggestion
     * @param identity identity whose role to be retrieved
```

src/java/org/apache/cassandra/auth/MutualTlsInternodeAuthenticator.java
```diff
@@ -0,0 +1,186 @@
...
+import org.apache.cassandra.config.ParameterizedClass;
+import org.apache.cassandra.exceptions.AuthenticationException;
+import org.apache.cassandra.exceptions.ConfigurationException;
+import org.apache.cassandra.utils.NoSpamLogger;
+
+/*
+ * Performs mTLS authentication for internode connections by extracting identities from outbound certificates
+ * configured for the cassandra instance.
+ *
+ * If the certificate identitiy is same as the identity present in the outbound keystore, the user is authorized
```
yifan-c:
```suggestion
 * If the certificate identity is same as the identity present in the outbound keystore, the user is authorized
```

src/java/org/apache/cassandra/auth/SpiffeCertificateValidator.java
```diff
@@ -0,0 +1,99 @@
...
+ * parameters :
+ * validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
+ * authenticator:
+ * class_name : org.apache.cassandra.auth.MutualTlsInternodeAuthenticator
+ * parameters :
+ * validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
+ */
+public class SpiffeCertificateValidator implements MutualTlsCertificateValidator
+{
+    private static final Logger LOGGER = LoggerFactory.getLogger(SpiffeCertificateValidator.class);
```
yifan-c: unused

src/java/org/apache/cassandra/cql3/statements/AddIdentityStatement.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Override
+    public void validate(ClientState state)
+    {
+        state.ensureNotAnonymous();
+
+        if (!DatabaseDescriptor.getRoleManager().isExistingRole(RoleResource.role(role)))
+        {
+            throw new InvalidRequestException(String.format("Can not add identity for non existent role '%s'", role));
```
yifan-c: nit
```suggestion
            throw new InvalidRequestException(String.format("Can not add identity for non-existent role '%s'", role));
```

src/java/org/apache/cassandra/cql3/statements/AddIdentityStatement.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public ResultMessage execute(ClientState state) throws RequestExecutionException, RequestValidationException
+    {
+        if (ifNotExists && DatabaseDescriptor.getRoleManager().isExistingIdentity(identity))
+        {
+            return null;
+        }
+        DatabaseDescriptor.getRoleManager().addIdentity(identity, role);
+        return null;
```
yifan-c: Returning `null` for both cases looks confusing. I think it can be simplified as
```suggestion
        if (!DatabaseDescriptor.getRoleManager().isExistingIdentity(identity))
        {
            DatabaseDescriptor.getRoleManager().addIdentity(identity, role);
        }
        return null;
```
It is a *nit* though

src/java/org/apache/cassandra/cql3/statements/DropIdentityStatement.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Override
+    public ResultMessage execute(ClientState state) throws RequestExecutionException, RequestValidationException
+    {
+        // not rejected in validate()
+        if (ifExists && !DatabaseDescriptor.getRoleManager().isExistingIdentity(identity))
+            return null;
+
+        DatabaseDescriptor.getRoleManager().dropIdentity(identity);
+        return null;
```
yifan-c: The same suggestion as the prior one for this method body

test/conf/cassandra-mtls-backward-compatibility.yaml
```diff
@@ -0,0 +1,76 @@
...
+  internode_encryption: none
+  keystore: test/conf/cassandra_ssl_test.keystore
+  keystore_password: cassandra
+  outbound_keystore: test/conf/cassandra_ssl_test_outbound.keystore
+  outbound_keystore_password: cassandra
+  truststore: test/conf/cassandra_ssl_test.truststore
+  truststore_password: cassandra
+  require_client_auth: true
+internode_authenticator: org.apache.cassandra.auth.AllowAllInternodeAuthenticator
+authenticator: org.apache.cassandra.auth.AllowAllAuthenticator
```
yifan-c, line 76: Looks like backward compatibility is covered. Ignore my first comment

test/unit/org/apache/cassandra/auth/AuthConfigTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+public class AuthConfigTest
+{
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+
+    @Test
+    public void testNewInstanceForMutualTlsInternodeAuthenticator() throws IOException, CertificateException
+    {
+        final Config config = load("cassandra-mtls.yaml");
```
yifan-c: drop `final`

test/unit/org/apache/cassandra/cql3/statements/AddIdentityStatementTest.java
```diff
@@ -0,0 +1,198 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class AddIdentityStatementTest
+{
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+    private static String SUPER_USER_ROLE = "cassandra";
+    private static String  IDENTITY = "urn:certmanager:idmsGroup/1234";
```
yifan-c:
```suggestion
    private static final String SUPER_USER_ROLE = "cassandra";
    private static final String IDENTITY = "urn:certmanager:idmsGroup/1234";
```

test/unit/org/apache/cassandra/cql3/statements/AddIdentityStatementTest.java
```diff
@@ -0,0 +1,198 @@
...
+        expectedException.expectMessage("You have not logged in");
+        QueryProcessor.executeInternal(ADD_QUERY);
+    }
+
+    @Test
+    public void testAddingNonExistentRole()
+    {
+        final String query = "ADD IDENTITY 'urn:certmanager:idmsGroup/1234' TO ROLE 'non-existing-role';";
+        expectedException.expect(InvalidRequestException.class);
+        expectedException.expectMessage("Can not add identity for non existent role 'non-existing-role'");
```
yifan-c: nit
```suggestion
        expectedException.expectMessage("Can not add identity for non-existent role 'non-existing-role'");
```

test/unit/org/apache/cassandra/cql3/statements/AddIdentityStatementTest.java
```diff
@@ -0,0 +1,198 @@
...
+        DatabaseDescriptor.getRoleManager().createRole(authenticatedUser, RoleResource.role("readwrite_user"), AuthTestUtils.getLoginRoleOptions());
+    }
+
+    @Test
+    public void ifNotExistsTest()
+    {
+        // Assert that identity is not present in the table
+        assertNull(DatabaseDescriptor.getRoleManager().roleForIdentity(IDENTITY));
+
+        final String addQueryWithIfNotExists = String.format("ADD IDENTITY IF NOT EXISTS '%s' TO ROLE '%s';", IDENTITY, SUPER_USER_ROLE);
```
yifan-c: no finals

test/unit/org/apache/cassandra/cql3/statements/DropIdentityStatementTest.java
```diff
@@ -0,0 +1,156 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class DropIdentityStatementTest
+{
+    @Rule
+    public ExpectedException expectedException = ExpectedException.none();
+    private static String  IDENTITY = "urn:certmanager:idmsGroup/1234";
```
yifan-c: static final

test/unit/org/apache/cassandra/cql3/statements/DropIdentityStatementTest.java
```diff
@@ -0,0 +1,156 @@
...
+        DropIdentityStatement dropIdentityStatement = (DropIdentityStatement) statement;
+        assertEquals(IDENTITY, dropIdentityStatement.identity);
+        assertFalse(dropIdentityStatement.ifExists);
+    }
+
+    @Test
+    public void testDroppingValidIdentity()
+    {
+        DatabaseDescriptor.getRoleManager().addIdentity(IDENTITY, "cassandra");
+        QueryProcessor.process(DROP_QUERY, ConsistencyLevel.QUORUM, getClientState(), 10L);
```
yifan-c, line 86: Add an assertion?

yifan-c (commented):
minor comments on code style mainly.

conf/cassandra.yaml
jyothsnakonisa: There is backward compatibility. Based on whether the authenticator has parameters or not we create the authenticator.

src/antlr/Parser.g
jyothsnakonisa: Even though they both of them are strings we wanted to define different types for both of them.

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
```diff
@@ -0,0 +1,207 @@
...
+                String msg = "Unable to extract client identity from certificate for authentication";
+                nospamLogger.error(msg);
+                throw new AuthenticationException(msg);
+            }
+            String role = identityCache.get(identity);
+            if (role == null)
+            {
+                String msg = "None of the users " + identity + " are authorized";
+                nospamLogger.error(msg);
+                throw new AuthenticationException(msg);
```
jonmeredith: nit: this is a possibility for log injection - the identity provided in the client certificate could include text to confuse a log scanner,  e.g. an identity of "%property{my_sensitive_property}" could cause it to be logged and be exploited if any more log4shell type vulnerabilities are discovered.

You could use `org.slf4j.helpers.MessageFormatter` to create the `msg` instead, that would also reduce the size of the map nospamLogger might create for previous timestamps.
```suggestion
                String msg = "None of the users {} are authorized";
                nospamLogger.error(msg, identity);
                throw new AuthenticationException(MessageFormatter.format(msg, identity).getMessage());
```

jonmeredith: `Certificate identity '{}' not authorized` might also make a clearer message in logs and exceptions - as I think there's only one identity.

src/java/org/apache/cassandra/auth/MutualTlsInternodeAuthenticator.java
```diff
@@ -0,0 +1,186 @@
...
+
+    protected boolean authenticateInternodeWithMtls(InetAddress remoteAddress, int remotePort, Certificate[] certificates,
+                                                    IInternodeAuthenticator.InternodeConnectionDirection connectionType)
+    {
+        if (connectionType == IInternodeAuthenticator.InternodeConnectionDirection.INBOUND)
+        {
+            String identity = certificateValidator.identity(certificates);
+            if (!certificateValidator.isValidCertificate(certificates))
+            {
+                noSpamLogger.error("Not a valid outbound certificate {}", identity);
```
jonmeredith: should this be `"Not a valid inbound certificate {}"`?

src/java/org/apache/cassandra/auth/SpiffeCertificateValidator.java
```diff
@@ -0,0 +1,94 @@
...
+        X509Certificate[] castedCerts = castCertsToX509(clientCertificates);
+        Collection<List<?>> subjectAltNames = castedCerts[0].getSubjectAlternativeNames();
+
+        if (subjectAltNames != null)
+        {
+            for (List<?> item : subjectAltNames)
+            {
+                Integer type = (Integer) item.get(0);
+                String spiffe = (String) item.get(1);
+                if (type == URI_TYPE && spiffe.contains("spiffe"))
```
jonmeredith: could we strengthen this to `spiffe.startsWith("spiffe://")` or at least include the `://` if you're concerned about certificates with whitespace before the spiffe.  I would go for the first option given this is security related.

test/unit/org/apache/cassandra/auth/MutualTlsInternodeAuthenticatorTest.java
```diff
@@ -0,0 +1,163 @@
...
+        DatabaseDescriptor.daemonInitialization();
+        StorageService.instance.initServer(0);
+    }
+
+    @Before
+    public void before()
+    {
+        Config config = DatabaseDescriptor.getRawConfig();
+        config.server_encryption_options = config.server_encryption_options.withOutboundKeystore("test/conf/cassandra_ssl_test_outbound.keystore")
+                                                                           .withOutboundKeystorePassword("cassandra");
```
jonmeredith, line 87: Do these test stores have any expiry times? If so how would somebody regenerate them to fix tests.

test/unit/org/apache/cassandra/config/ConfigCompatabilityTest.java
```diff
@@ -88,6 +88,14 @@
                                                      .add("streaming_socket_timeout_in_ms") // CASSANDRA-12229
                                                      .build();
 
+    private static final Set<String> EXPECTED_ERRORS_403 = ImmutableSet.<String>builder()
```
jonmeredith: should probably rename this `EXPECTED_FOR_50` or something like that?

test/unit/org/apache/cassandra/auth/MutualTlsInternodeAuthenticatorTest.java
jyothsnakonisa: @jonmeredith  Yes the certificate does have an expiry. However these certificates are not used for SSL/TLS handshake.  They are used only for testing extracting identities and in authenticators, so even if they expire they do not make the tests to fail.

src/java/org/apache/cassandra/auth/MutualTlsInternodeAuthenticator.java
jyothsnakonisa: We are authenticating for inbound connection and validating an outbound certificate. So I think the log message should have outbound certificate.

jonmeredith: The problem is outbound is ambiguous as to the type of certificate or type of connection.  This is an inbound connection by the `if` statement above - but is handling another nodes outbound certificate.

Perhaps it would be more valuable to say something like `noSpamLogger.error("Not a valid certificate from {}:{} with identity '{}'", remoteAddress, remotePort, identity);"`  then it's clear where the cert is from and who the identity is.

jyothsnakonisa: That's a more clearer message. Updated the log!

yukim:
Hi @jyothsnakonisa 

Shouldn't we change the condition here so Cassandra does not log confusing message to the user of mTLS auth with cache config changed?

```java
        if (!(authenticator instanceof PasswordAuthenticator)
            && (conf.credentials_update_interval != null
                || conf.credentials_validity.toMilliseconds() != 2000
                || conf.credentials_cache_max_entries != 1000))
        {
            logger.info("Configuration options credentials_update_interval, credentials_validity and " +
                        "credentials_cache_max_entries may not be applicable for the configured authenticator ({})",
                        authenticator.getClass().getName());
        }

```
https://github.com/apache/cassandra/pull/2372/files#diff-6cb647a8107efc82a7abdd3639ea1da0d2e58feb37350770bd29fe09ac46f687R64-R72

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
```diff
@@ -0,0 +1,208 @@
...
+    @Override
+    public SaslNegotiator newSaslNegotiator(InetAddress clientAddress, Certificate[] certificates)
+    {
+        return new CertificateNegotiator(certificates);
+    }
+
+    @Override
+    public AuthenticatedUser legacyAuthenticate(Map<String, String> credentials) throws AuthenticationException
+    {
+        return null;
```
yukim: This shouldn't be `null`, otherwise, when the node is using `CassandraLoginModule`  for JMX auth, it throws NullPointerException. Should at least return `AuthenticatedUser.ANONYMOUS_USER` like `AllowAllAuthenticator`.

yukim: Or maybe throw `AuthenticationException` with the message indicating mTLS auth is not supported for CassandraLoginModule.

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
```diff
@@ -0,0 +1,125 @@
...
+
+== What is an Identity
+
+One can define their own identity for certificates by extracting some fields or
+information from the certificates. one can implement the interface `MutualTlsCertificateValidator`
+to provide logic for validating & extracting identity from the certificates.
+
+There is a default implementation of `MutualTlsCertificateValidator` with
+https://spiffe.io/docs/latest/spiffe-about/spiffe-concepts/[SPIFFE] as the identity
+of the certifactes.This requires spiffe to be present in the SAN of the certificate.
```
yifan-c:
```suggestion
of the certificates.This requires spiffe to be present in the SAN of the certificate.
```

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
```diff
@@ -0,0 +1,125 @@
...
+the operators can run cassandra in optional mTLS mode and use
+`MutualTlsWithPasswordFallbackAuthenticator` which will accept both mTLS & password
+based connections, based on the type of connection client is making. These settings
+can be configured in `cassandra.yaml`. Once all the clients migrate to using mTLS,
+turn off optional mode and set the authenticator to be `MutualTlsAuthenticator`. From
+that point only mTLS client connections will be accepted.
+
+* For Internode connections, while doing rolling upgrades from non-mTLS based configuration
+to mTLS based configuration, use optional mode for the new nodes to be able to connect to
+old nodes which are still using non-mTLS based configuation during upgrade. After this, change
```
yifan-c:
```suggestion
old nodes which are still using non-mTLS based configuration during upgrade. After this, change
```

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
```diff
@@ -67,16 +67,23 @@ identity is present in the `identity_to_roles` table, access will be granted.
 
 == Configuring mTLS authenticator for Internode connections
 
-Internode authenticator trusts certificates which have the same identity as the node.
+Internode authenticator trusts certificates whose identities are present in
+`internode_authenticator.parameters.trusted_peer_identities` if configured.
+
+Otherwise, it trusts conections which have the same identity as the node.
```
yifan-c:
```suggestion
Otherwise, it trusts connections which have the same identity as the node.
```

src/java/org/apache/cassandra/auth/AuthConfig.java
```diff
@@ -61,7 +61,7 @@ public static void applyAuth()
         // the configuration options regarding credentials caching are only guaranteed to
         // work with PasswordAuthenticator, so log a message if some other authenticator
         // is in use and non-default values are detected
-        if (!(authenticator instanceof PasswordAuthenticator)
+        if (!(authenticator instanceof PasswordAuthenticator || authenticator instanceof MutualTlsAuthenticator)
```
yifan-c, line 64: Is the condition equivalent to `!authenticator.requireAuthentication()`?

doc/modules/cassandra/pages/developing/cql/cql_singlefile.adoc
```diff
@@ -2235,6 +2235,57 @@ LIST ROLES;
...
+
+bc(syntax).. +
+::= ADD IDENTITY ( IF NOT EXISTS )? TO ROLE ?
+
+_Sample:_
+
+bc(sample). +
+ADD IDENTITY 'id1' TO ROLE 'role1';
+
+Only a `SUPERUSER` can add an identity
```
yifan-c: Is it still true? Asking because `org.apache.cassandra.cql3.statements.CreateRoleStatement#authorize` checks for CREATE permission not SUPERUSER.

yifan-c (commented):
The new changes mostly LGTM. There are some doc changes needed.

src/java/org/apache/cassandra/auth/AuthConfig.java
jyothsnakonisa: No, I think this is related to change logging when some cache configuration properties change in password authenticator and mTLS authenticator. We are using common cache configuration properties for both password authenticator and mTLS authenticator.

doc/modules/cassandra/pages/developing/cql/cql_singlefile.adoc
```diff
@@ -2235,6 +2235,57 @@ LIST ROLES;
...
+
+bc(syntax).. +
+::= ADD IDENTITY ( IF NOT EXISTS )? TO ROLE ?
+
+_Sample:_
+
+bc(sample). +
+ADD IDENTITY 'id1' TO ROLE 'role1';
+
+Only a user with previleges to add roles can add identities
```
jonmeredith:
```suggestion
Only a user with privileges to add roles can add identities
```

doc/modules/cassandra/pages/developing/cql/cql_singlefile.adoc
```diff
@@ -2235,6 +2235,57 @@ LIST ROLES;
...
+::= DROP IDENTITY ( IF EXISTS )?  +
+p.
+
+_Sample:_
+
+bc(sample). +
+DROP IDENTITY 'testIdentity'; +
+DROP IDENTITY IF EXISTS 'testIdentity';
+
+Only a user with previleges to drop roles can remove identities
```
jonmeredith:
```suggestion
Only a user with privileges to drop roles can remove identities
```

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
```diff
@@ -0,0 +1,133 @@
...
+Internode connections, `MutualTlsAuthenticator` & `MutualTlsInternodeAuthenticator`
+can be used for the authentication by leveraging the client certificates from the
+SSL handshake.
+
+After SSL handshake, identity from the client certificates is extracted and only
+authorized users will be granted access.
+
+== What is an Identity
+
+One can define their own identity for certificates by extracting some fields or
```
jonmeredith:
```suggestion
Operators can define their own identity for certificates by extracting some fields or
```

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
```diff
@@ -0,0 +1,133 @@
...
+information from the certificates. one can implement the interface `MutualTlsCertificateValidator`
+to provide logic for validating & extracting identity from the certificates.
```
jonmeredith:
```suggestion
information from the certificates. Implementing the interface `MutualTlsCertificateValidator`
supports validating & extracting identity from the certificates that can be used by `MutualTlsAuthenticator` and `MutualTlsInternodeAuthenticator` to customize for the certificate conventions used in the deployment environment.
```

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+There is a default implementation of `MutualTlsCertificateValidator` with
+https://spiffe.io/docs/latest/spiffe-about/spiffe-concepts/[SPIFFE] as the identity
+of the certificates.This requires spiffe to be present in the SAN of the certificate.
+One can implement their own validator that suits their needs.
```
jonmeredith: I'd just drop the `One can implement`...sentence as you already covered above.

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
```diff
@@ -0,0 +1,133 @@
...
+
+For example, instead of using `SPIFFE` based validator a `CN` based validator can be
+implemented.
```
jonmeredith:
```suggestion
Instead of using `SPIFFE` based validator a `CN` based validator could by an operator-supplied custom class that implements `MutualTlsCertificateValidator`.
```

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
```diff
@@ -0,0 +1,133 @@
...
+[source, plaintext]
+----
+client_encryption_options:
+  enabled: true
+  optional: false
+  keystore: conf/.keystore
+  keystore_password: cassandra
+  truststore: conf/.truststore
+  truststore_password: cassandra
+  require_client_auth: false // to enable mTLS
```
jonmeredith:
```suggestion
  require_client_auth: false // while migrating to mTLS, keep permitting non-validated certs
```

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
```diff
@@ -0,0 +1,133 @@
...
+[source, plaintext]
+----
+server_encryption_options:
+  internode_encryption: all
+  optional: true
+  keystore: conf/.keystore
+  keystore_password: cassandra
+  outbound_keystore: conf/.outbound_keystore
+  outbound_keystore_password: cassandra
+  require_client_auth: true  // for enabling mTLS
```
jonmeredith, line 99: I think we force this true for internode connections, so not sure if this setting really matters

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
```diff
@@ -0,0 +1,133 @@
...
+----
+internode_authenticator:
+  class_name : org.apache.cassandra.auth.MutualTlsInternodeAuthenticator
+  parameters :
+    validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
+    trusted_peer_identities : "spiffe1,spiffe2"
+----
+
+*STEP 3: Bounce the cluster*
+Once the cluster is bounced all internode communications will be authenticated by mTLS
```
jonmeredith:
```suggestion
Once all nodes in the cluster are restarted, all internode communications will be authenticated by mTLS
```

might be nice to say it a different way - bounce is a bit of an idiom so may be helpful to spell it out.

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
```diff
@@ -0,0 +1,133 @@
...
+* For client connections, since the migration will not happen overnight,
+the operators can run cassandra in optional mTLS mode and use
+`MutualTlsWithPasswordFallbackAuthenticator` which will accept both mTLS & password
+based connections, based on the type of connection client is making. These settings
+can be configured in `cassandra.yaml`. Once all the clients migrate to using mTLS,
+turn off optional mode and set the authenticator to be `MutualTlsAuthenticator`. From
+that point only mTLS client connections will be accepted.
+
+* For Internode connections, while doing rolling upgrades from non-mTLS based configuration
+to mTLS based configuration, use optional mode for the new nodes to be able to connect to
```
jonmeredith: use optional mode - it would be good to be more specific with how to set optional mode. An operator could think  you mean set `optional: true` in encryption settings which could weaken settings during a migration, rather than set the authenticator to `MutualTlsWithPasswordFallbackAuthenticator`

src/java/org/apache/cassandra/auth/MutualTlsInternodeAuthenticator.java
```diff
@@ -0,0 +1,226 @@
...
+import org.apache.cassandra.exceptions.ConfigurationException;
+import org.apache.cassandra.utils.NoSpamLogger;
+
+/*
+ * Performs mTLS authentication for internode connections by extracting identities from the certificates of incoming
+ * connection and verifying them against a list of authorized peers. Authorized peers can be configured in
+ * trusted_peer_identities in cassandra yaml, otherwise authenticator trusts connections from peers which has the same
+ * identity as the one that the node uses for making outbound connections.
+ *
+ * Optionally one can validate the identity extracted from outbound keystore with node_identity that is configured in
```
jonmeredith:
```suggestion
 * Optionally Cassandra can validate the identity extracted from outbound keystore with node_identity that is configured in
```

src/java/org/apache/cassandra/auth/MutualTlsInternodeAuthenticator.java
```diff
@@ -0,0 +1,226 @@
...
+ * cassandra.yaml to avoid any configuration errors.
+ *
+ * Authenticator & Certificate validator can be configured using cassandra.yaml, one can write their own mTLS certificate
```
jonmeredith:
```suggestion
 * Authenticator & Certificate validator can be configured using cassandra.yaml, operators can write their own mTLS certificate
```

src/java/org/apache/cassandra/auth/MutualTlsInternodeAuthenticator.java
```diff
@@ -0,0 +1,226 @@
...
+ * validator and configure it in cassandra.yaml.Below is an example on how to configure validator.
+ * Note that this example uses SPIFFE based validator,It could be any other validator with any defined identifier format.
```
jonmeredith:
```suggestion
 * Note that this example uses SPIFFE based validator, it could be any other validator with any defined identifier format.
```

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
jyothsnakonisa: This setting is not for client certificate validation. It is for allowing only TLS connections. Do you mean to say that we have to use optional mode while migrating to mTLS?

doc/modules/cassandra/pages/getting-started/mtlsauthenticators.adoc
jyothsnakonisa: If the authenticator is not mTLS authenticator, then `require_client_auth: false` should be the configuration. In case of internode mtls authentication, we need to explicitly set this to be true.

jonmeredith: I was just thinking of this code... https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/security/SSLFactory.java#L190

PR #2374, opened 2023-05-27
A new nodetool/JMX command that tells whether node's decommission fai…

jaydeepkumar1984:
```
A new nodetool/JMX command that tells whether node's decommission failed or not

Currently, when a node is being decommissioned and if any failure happens, then an exception is thrown back to the caller.

But Cassandra's decommission takes considerable time ranging from minutes to hours to days. There are various scenarios in that the caller may need to probe the status again:

The caller times out
It is not possible to keep the caller hanging for such a long time
And If the caller does not know what happened internally, then it cannot retry, etc., leading to other issues.

So, in this ticket, I am going to add a new nodetool/JMX command that can be invoked by the caller anytime, and it will return the correct status.

It might look like a smaller change, but when we need to operate Cassandra at scale in a large-scale fleet, then this becomes a bottleneck and require constant operator intervention.


patch by Jaydeepkumar Chovatia <chovatia.jaydeep@gmail.com>; reviewed by <Reviewers> for CASSANDRA-18555

Co-authored-by: Jaydeepkumar Chovatia <chovatia.jaydeep@gmail.com>

```

The [Cassandra Jira](https://issues.apache.org/jira/browse/CASSANDRA-18555)

### DIFF
--- a/src/java/org/apache/cassandra/metrics/StorageMetrics.java
+++ b/src/java/org/apache/cassandra/metrics/StorageMetrics.java
@@ -45,6 +45,7 @@ public class StorageMetrics
     public static final Counter totalHintsInProgress  = Metrics.counter(factory.createMetricName("TotalHintsInProgress"));
     public static final Counter totalHints = Metrics.counter(factory.createMetricName("TotalHints"));
     public static final Counter repairExceptions = Metrics.counter(factory.createMetricName("RepairExceptions"));
+    public static final Counter errorDecommissiong = Metrics.counter(factory.createMetricName("ErrorDecommissioning"));
 
     private static Gauge<Long> createSummingGauge(String name, ToLongFunction<KeyspaceMetrics> extractor)
     {
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -433,6 +433,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
     private static final boolean allowSimultaneousMoves = CONSISTENT_SIMULTANEOUS_MOVES_ALLOW.getBoolean();
     private static final boolean joinRing = JOIN_RING.getBoolean();
     private boolean replacing;
+    private volatile boolean hasDecommissionFailed = false;
 
     private final StreamStateStore streamStateStore = new StreamStateStore();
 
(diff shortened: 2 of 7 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5127,100 +5128,113 @@ private void startLeaving()
...
+            catch (ExecutionException e)
+            {
+                logger.error("Error while decommissioning node ", e.getCause());
+                throw new RuntimeException("Error while decommissioning node: " + e.getCause().getMessage());
+            }
+            finally
+            {
+                isDecommissioning.set(false);
+            }
+        } catch (Exception e)
```
Maxwell-Guo, line 5227: I think there is a problem for code format with cath '{'

Maxwell-Guo: Besides , can we use Throwable instand of Exception ？

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5236,6 +5250,21 @@ private void leaveRing()
         Uninterruptibles.sleepUninterruptibly(delay, MILLISECONDS);
     }
 
+    public boolean isDecommissionFailed()
+    {
+        if (operationMode == Mode.LEAVING && hasDecommissionFailed)
```
Maxwell-Guo, line 5255: I think we can change this to `if (hasDecommissionFailed && operationMode == Mode.LEAVING)`
if hasDecommissionFailed is false we will not need to do the comparison with operationMode
Edit :
or why we just return 
hasDecommissionFailed && operationMode == Mode.LEAVING
? and not need for if () else

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -433,6 +433,7 @@ public enum Mode { STARTING, NORMAL, JOINING, LEAVING, DECOMMISSIONED, MOVING, D
     private static final boolean allowSimultaneousMoves = CONSISTENT_SIMULTANEOUS_MOVES_ALLOW.getBoolean();
     private static final boolean joinRing = JOIN_RING.getBoolean();
     private boolean replacing;
+    private volatile boolean hasDecommissionFailed = false;
```
Maxwell-Guo, line 436: I think we can change the name to isDecommissionFailed as src/java/org/apache/cassandra/service/StorageServiceMBean.java  method isDecommissionFailed();
Edit :
can we just AtomicBoolean

test/unit/org/apache/cassandra/service/StorageServiceServerTest.java
```diff
@@ -658,4 +660,26 @@ public void isReplacingSameHostAddressAndHostIdTest() throws UnknownHostExceptio
...
\ No newline at end of file
+
+    @Test
+    public void testIsDecommissionNotFailed()
+    {
+        assertFalse(StorageService.instance.isDecommissionFailed());
+        assertEquals(0, StorageMetrics.errorDecommissiong.getCount());
+    }
+
+    @Test
```
Maxwell-Guo, line 671: I think we can also add a new test case use ToolRunner.invokeNodetool see GossipInfoTest
we can use at frist set the node's mode to Un-normal and then use ToolRunner.invokeNodetool  to get the status of descommission.

Maxwell-Guo (changes requested):

src/java/org/apache/cassandra/tools/NodeTool.java
```diff
@@ -163,6 +163,7 @@ public int execute(String... args)
                 InvalidatePermissionsCache.class,
                 InvalidateRolesCache.class,
                 InvalidateRowCache.class,
+                IsDecommissionFailed.class,
```
smiklosovic, line 166: @jaydeepkumar1984 where is this class? I do not see it implemented as part of this PR?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5127,100 +5128,113 @@ private void startLeaving()
...
-            }
 
-            startLeaving();
-            long timeout = Math.max(RING_DELAY_MILLIS, BatchlogManager.instance.getBatchlogTimeout());
-            setMode(Mode.LEAVING, "sleeping " + timeout + " ms for batch processing and pending range setup", true);
-            Thread.sleep(timeout);
+                startLeaving();
+                long timeout = Math.max(RING_DELAY_MILLIS, BatchlogManager.instance.getBatchlogTimeout());
+                setMode(Mode.LEAVING, "sleeping " + timeout + " ms for batch processing and pending range setup", true);
+                Thread.sleep(timeout);
```
smiklosovic, line 5189: maybe it is good time to sleep uninterruptibly? Look into Uninterruptibles in google guava

src/java/org/apache/cassandra/metrics/StorageMetrics.java
```diff
@@ -45,6 +45,7 @@
     public static final Counter totalHintsInProgress  = Metrics.counter(factory.createMetricName("TotalHintsInProgress"));
     public static final Counter totalHints = Metrics.counter(factory.createMetricName("TotalHints"));
     public static final Counter repairExceptions = Metrics.counter(factory.createMetricName("RepairExceptions"));
+    public static final Counter errorDecommissiong = Metrics.counter(factory.createMetricName("ErrorDecommissioning"));
```
smiklosovic, line 48: @jaydeepkumar1984 Why do we need metrics for knowing if decommission failed? What do you plan to do if you determine that the decommission has failed?

src/java/org/apache/cassandra/tools/NodeTool.java
Maxwell-Guo: right, I think this patch may fail to compile. It is need for us to executed  test locally before post the pr.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5127,100 +5128,113 @@ private void startLeaving()
...
                     throw new UnsupportedOperationException("no other normal nodes in the ring; decommission would be pointless");
-            if (operationMode != Mode.NORMAL)
-                throw new UnsupportedOperationException("Node in " + operationMode + " state; wait for status to become normal or restart");
-        }
-        if (!isDecommissioning.compareAndSet(false, true))
-            throw new IllegalStateException("Node is still decommissioning. Check nodetool netstats.");
+                if (operationMode != Mode.NORMAL)
+                    throw new UnsupportedOperationException("Node in " + operationMode + " state; wait for status to become normal or restart");
+            }
+            if (!isDecommissioning.compareAndSet(false, true))
```
Maxwell-Guo, line 5143: I think we can not include some code into the try catch , like line 5143 
` if (!isDecommissioning.compareAndSet(false, true))`
first time you have execute a decommission, and the node is being doing decommission which may take a long time ,
and after a while you may execute another decommission , you will got a false. But actually the node's decommission is be doing now and have not failed. 
you will get a wrong status.

Maxwell-Guo (changes requested):

PR #2378, opened 2023-05-29
CASSANDRA-17433 Remove dependency on pytz library for setting CQLSH t…

lmtrombone:
PEP 615 added support for the IANA Time Zone Database in the standard python library.  This removed the dependency on the pytz library to be able to set the timezone in CQLSH on versions >= 3.9.  This MR contains a version check where we will use the standard library on versions >= 3.9.


bschoening:
Just for general hygiene, since you are updating cqlsh.adoc, could you remove the obsolete line in Compatibility: 'cqlsh is compatible with Python 2.7.'

lmtrombone:
> Just for general hygiene, since you are updating cqlsh.adoc, could you remove the obsolete line in Compatibility: 'cqlsh is compatible with Python 2.7.'

Done

PR #2381, opened 2023-06-01
CEP-7 SAI rebase on trunk 6cea0cd448d11dc633237ceb1f73f36c2eafb368

maedhroz:
@mike-tr-adamson @blambov Here are my notes from the rebase:

- Should `TrieNode.ordinal` initialize to `-1` or `0`?
- It looks like in `DataOutputPlus`, SAI added an actual implementation for `padToPageBoundary()`. Is that going to be problematic?
- `SequentialWriter` now uses the `bytesLeftInPage()` implementation in `PageAware` rather than a custom one.
- The tests I've spot-checked have looked pretty good, but `SortedTermsTest#testFileValidation()` is failing to verify the checksum on the `PRIMARY_KEY_TRIE`. I've tried to revert `DataOutputPlus` and `SequentialWriter` to their current `cep-7-sai` versions, but that doesn't seem to make a difference. Not sure what's going on there...


maedhroz:
CircleCI: https://app.circleci.com/pipelines/github/maedhroz/cassandra?branch=cep-7-sai-5-30-rebase

PR #2386, opened 2023-06-05
CASSANDRA-18566: Avoid unnecessary deserialization of terminal arguments when executing CQL functions

adelapena:


src/java/org/apache/cassandra/cql3/functions/ScalarFunction.java
```diff
@@ -66,19 +65,19 @@ public default boolean isMonotonic()
...
      * @return a function corresponding to the partial application of this function to the parameters of
-     * {@code partialParameters} that are not {@link #UNRESOLVED}.
+     * {@code partialArguments} that are not {@link #UNRESOLVED}.
      */
-    public default ScalarFunction partialApplication(ProtocolVersion protocolVersion, List<ByteBuffer> partialParameters)
+    default ScalarFunction partialApplication(ProtocolVersion protocolVersion, List<ByteBuffer> partialArguments)
     {
         int unresolvedCount = 0;
-        for (ByteBuffer parameter : partialParameters)
+        for (ByteBuffer parameter : partialArguments)
```
blerer: It is a minor thing but we should probably use argument instead of parameter to be consistent.

src/java/org/apache/cassandra/cql3/selection/AbstractFunctionSelector.java
```diff
@@ -74,19 +78,27 @@ protected Selector deserialize(DataInputPlus in, int version, TableMetadata meta
                                                             .collect(joining(", "))));
 
             boolean isPartial = in.readBoolean();
+            // if the function is partial we need to retrieve the resolved parameters.
```
blerer: We should may be replace `parameters` by `arguments`.

src/java/org/apache/cassandra/cql3/functions/ToJsonFct.java
```diff
@@ -54,14 +54,28 @@ private ToJsonFct(String name, AbstractType<?> argType)
...
+                return null;
+
+            return argTypes.get(0).toJSONString(buffer, protocolVersion);
+        });
+    }
+
+    @Override
+    public ByteBuffer execute(Arguments arguments) throws InvalidRequestException
+    {
+        assert arguments.size() == 1 : "Expected 1 argument for toJson(), but got " + arguments.size();
```
blerer: Is it possible?

blerer (commented):
It seems that the `getArgumentDeserializer` method is missing for: `DoubleType`,  `DecimalType`, `Int32Type` and `IntegerType`.

src/java/org/apache/cassandra/cql3/functions/ToJsonFct.java
adelapena: It isn't, the factory should have caught it before. Removed.

src/java/org/apache/cassandra/cql3/functions/ScalarFunction.java
adelapena: Right, changed.

src/java/org/apache/cassandra/cql3/selection/AbstractFunctionSelector.java
adelapena: Indeed we should, done.

adelapena:
> It seems that the getArgumentDeserializer method is missing for: DoubleType, DecimalType, Int32Type and IntegerType.

Good catch; added.

blerer (approved):

PR #2390, opened 2023-06-06
CASSANDRA-18555 expose failed decommission via nodetool info

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -263,6 +263,8 @@ import static org.apache.cassandra.net.NoPayload.noPayload;
 import static org.apache.cassandra.net.Verb.REPLICATION_DONE_REQ;
 import static org.apache.cassandra.service.ActiveRepairService.ParentRepairStatus;
 import static org.apache.cassandra.service.ActiveRepairService.repairCommandExecutor;
+import static org.apache.cassandra.service.StorageService.Mode.DECOMMISSIONED;
+import static org.apache.cassandra.service.StorageService.Mode.DECOMMISSION_FAILED;
 import static org.apache.cassandra.utils.Clock.Global.currentTimeMillis;
 import static org.apache.cassandra.utils.Clock.Global.nanoTime;
 import static org.apache.cassandra.utils.FBUtilities.getBroadcastAddressAndPort;
@@ -414,7 +416,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
     /* the probability for tracing any particular request, 0 disables tracing and 1 enables for all */
     private double traceProbability = 0.0;
 
-    public enum Mode { STARTING, NORMAL, JOINING, LEAVING, DECOMMISSIONED, MOVING, DRAINING, DRAINED }
+    public enum Mode { STARTING, NORMAL, JOINING, LEAVING, DECOMMISSIONED, DECOMMISSION_FAILED, MOVING, DRAINING, DRAINED }
     private volatile Mode operationMode = Mode.STARTING;
 
     /* Used for tracking drain progress */
@@ -626,7 +628,7 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
      * they get the Gossip shutdown message, so even if
      * we don't get time to broadcast this, it is not a problem.
      *
-     * See {@link Gossiper#markAsShutdown(InetAddressAndPort)}
+     * See Gossiper.markAsShutdown(InetAddressAndPort)
      */
     private void shutdownClientServers()
     {
@@ -2157,7 +2159,8 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
     /**
      * All MVs have been created during bootstrap, so mark them as built
      */
-    private void markViewsAsBuilt() {
+    private void markViewsAsBuilt()
+    {
         for (String keyspace : Schema.instance.getUserKeyspaces().names())
         {
             for (ViewMetadata view: Schema.instance.getKeyspaceMetadata(keyspace).views)
@@ -2168,9 +2171,16 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
     /**
      * Called when bootstrap did finish successfully
      */
-    private void bootstrapFinished() {
+    private void bootstrapFinished()
+    {
         markViewsAsBuilt();
         isBootstrapMode = false;
+    }
+
+    @Override
+    public String getBootstrapState()
+    {
+        return SystemKeyspace.getBootstrapState().name();
     }
 
     public boolean resumeBootstrap()
@@ -5128,18 +5138,32 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
 
     public void decommission(boolean force) throws InterruptedException
     {
+        if (operationMode == DECOMMISSIONED)
+        {
+            logger.info("This node was already decommissioned. There is no point in decommissioning it again.");
+            return;
+        }
+
+        if (isDecommissioning())
+        {
+            logger.info("This node is still decommissioning.");
+            return;
+        }
+
         TokenMetadata metadata = tokenMetadata.cloneAfterAllLeft();
+        // there is no point to do this logic again once node was decommissioning but failed to do so
         if (operationMode != Mode.LEAVING)
         {
             if (!tokenMetadata.isMember(FBUtilities.getBroadcastAddressAndPort()))
                 throw new UnsupportedOperationException("local node is not a member of the token ring yet");
-            if (metadata.getAllEndpoints().size() < 2)
+            if (metadata.getAllEndpoints().size() < 2 && metadata.getAllEndpoints().contains(FBUtilities.getBroadcastAddressAndPort()))
                     throw new UnsupportedOperationException("no other normal nodes in the ring; decommission would be pointless");
-            if (operationMode != Mode.NORMAL)
+            if (operationMode != Mode.NORMAL && operationMode != DECOMMISSION_FAILED)
                 throw new UnsupportedOperationException("Node in " + operationMode + " state; wait for status to become normal or restart");
         }
+
         if (!isDecommissioning.compareAndSet(false, true))
-            throw new IllegalStateException("Node is still decommissioning. Check nodetool netstats.");
+            throw new IllegalStateException("Node is still decommissioning. Check nodetool netstats or nodetool info.");
 
         if (logger.isDebugEnabled())
             logger.debug("DECOMMISSIONING");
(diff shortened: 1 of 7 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5182,41 +5190,40 @@ public void decommission(boolean force) throws InterruptedException
...
-                    }
-                    catch (IOError ioe)
-                    {
-                        logger.info("failed to shutdown message service: {}", ioe);
-                    }
+                MessagingService.instance().shutdown();
+            }
+            catch (IOError ioe)
+            {
+                logger.info("failed to shutdown message service", ioe);
```
Maxwell-Guo, line 5233: what about change the logging level to warn ?

test/distributed/org/apache/cassandra/distributed/test/DecommissionTest.java
```diff
@@ -0,0 +1,153 @@
...
+                {
+                    StorageService.instance.decommission(true);
+
+                    assertFalse("system.local column 'bootstrapped' should not contain value " + DECOMMISSION_FAILED,
+                                SystemKeyspace.hasDecommissionFailed());
+
+                    assertEquals("system.local column 'bootstrapped' should contain value " + DECOMMISSIONED,
+                                 DECOMMISSIONED.name(),
+                                 StorageService.instance.getBootstrapState());
+                }
```
Maxwell-Guo, line 116: should we add a fail("some messages "); after ` assertEquals("system.local column 'bootstrapped' should contain value " + DECOMMISSIONED,
                                 DECOMMISSIONED.name(),
                                 StorageService.instance.getBootstrapState());`
as if the content in this try does not throw an exception?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -626,7 +626,7 @@ public void stopTransports()
      * they get the Gossip shutdown message, so even if
      * we don't get time to broadcast this, it is not a problem.
      *
-     * See {@link Gossiper#markAsShutdown(InetAddressAndPort)}
+     * See Gossiper.markAsShutdown(InetAddressAndPort)
```
bereng, line 631: Is this syntax correct?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1070,16 +1070,16 @@ private void prepareToJoin() throws ConfigurationException
         {
             Map<ApplicationState, VersionedValue> appStates = new EnumMap<>(ApplicationState.class);
 
-            if (SystemKeyspace.wasDecommissioned())
+            if (SystemKeyspace.wasDecommissioned() || SystemKeyspace.hasDecommissionFailed())
             {
                 if (OVERRIDE_DECOMMISSION.getBoolean())
                 {
-                    logger.warn("This node was decommissioned, but overriding by operator request.");
+                    logger.warn("This node was decommissioned or its decommission has failed, but overriding by operator request.");
```
bereng: Here you could easily make the warn message text actually reflect if it was a failed decommission or the other

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5182,41 +5190,40 @@ public void decommission(boolean force) throws InterruptedException
             }
 
             startLeaving();
-            long timeout = Math.max(RING_DELAY_MILLIS, BatchlogManager.instance.getBatchlogTimeout());
+            long timeout = Math.max(RING_DELAY_MILLIS, BatchlogManager.getBatchlogTimeout());
             setMode(Mode.LEAVING, "sleeping " + timeout + " ms for batch processing and pending range setup", true);
-            Thread.sleep(timeout);
+            Uninterruptibles.sleepUninterruptibly(timeout, MILLISECONDS);
```
bereng: Any particular reason to change this to `sleepUninterruptibly()`?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5182,41 +5190,40 @@ public void decommission(boolean force) throws InterruptedException
...
+            // let op be responsible for killing the process
         }
-        catch (ExecutionException e)
+        catch (Throwable t)
         {
-            logger.error("Error while decommissioning node ", e.getCause());
-            throw new RuntimeException("Error while decommissioning node: " + e.getCause().getMessage());
+            Throwable cause = t.getCause() == null ? t : t.getCause();
+            logger.error("Error while decommissioning node ", cause);
+            SystemKeyspace.setBootstrapState(SystemKeyspace.BootstrapState.DECOMMISSION_FAILED);
```
Maxwell-Guo: Is it necessary to add Mode.DECOMMISSIONED_FAILED and setMode( Mode.DECOMMISSIONED_FAILED, true) after line 5221?

As talking about the Mode state , Do we need to think again about the state of DECOMMISSIONED FAILED, because every step in the try catch may fail. If there is a failure, are there some corner problems that have not been considered?
Such as when we faild at streaming data , the node statue may show the node is leaving ,and should show users the statues of DECOMMISSIONED FAILED in nodetool status or something else.

test/distributed/org/apache/cassandra/distributed/test/DecommissionTest.java
smiklosovic: I dont understand. When assertEquals fails, it will throw `ComparisionFailure` / AssertionError. That is caught by `catch` and rethrown?

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: @bereng probably isnt but if you take a closer look, that method is not accessible from this class and it marks it as errorneous in IDEA (whole file) and it drives me crazy. 

As that method is not reachable from here, I would just completely remove that link, wdyt?

I just tried to keep it there and not mark it as error.

smiklosovic: I changed it to `     * See Gossiper#markAsShutdown(InetAddressAndPort)`

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: @bereng there are 9 Uninterruptibles and 2 Thread.sleep in this class. It is little bit messy. I changed it  back to Thread.sleep ...

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: Answer to your first question: I do not think so.

We have:

    public enum BootstrapState
    {
        NEEDS_BOOTSTRAP,
        COMPLETED,
        IN_PROGRESS,
        DECOMMISSIONED,
        DECOMMISSION_FAILED
    }

But "modes" are:

    public enum Mode { STARTING, NORMAL, JOINING, LEAVING, DECOMMISSIONED, MOVING, DRAINING, DRAINED }

So the question is what consequence it has to introduce "DECOMMISSION_FAILED" mode? 

Second question:

I was thinking about this and if you look at it as it is done currently, there is exactly same behavior. It may error out as of now while decommissioning and the problems you described are already there. This patch at least tells us that it failed to decommission what was not the case before.

We may look at this later in separate ticket. I think testing all corner cases is not in the scope of this ticket.

smiklosovic: EDIT: I have added there "mode" as well.

Maxwell-Guo: > We may look at this later in separate ticket. I think testing all corner cases is not in the scope of this ticket.
Ok

Maxwell-Guo (approved):

src/java/org/apache/cassandra/service/StorageService.java
Maxwell-Guo: Sorry, I have a new question that may need your advice, that is, the initial node joins the ring successfully, and the system table is set with the COMPLETED state. In the old process, if the decommission process is in progress, and the system table has not been updated to need_bootstrap or the final decommissioned ,then the process  decommission failed, the system table is still COMPLETE, but with the new process system table will be updated to decommissioned_failed. 

Although you have updated the judgment of the prepareToJoin function about SystemKeyspace.wasDecommissioned(), there is an isReplacing judgment before prepareToJoin, and the old and new processes may be different in this place. What about decommission failed, and start doing replace node ?
Not sure if my understanding is correct @smiklosovic  @bereng @driftx, can you help to confirm ?

`

    public boolean isReplacing()
    {
        if (replacing)
            return true;

        if (REPLACE_ADDRESS_FIRST_BOOT.getString() != null && SystemKeyspace.bootstrapComplete())
        {
            logger.info("Replace address on the first boot requested; this node is already bootstrapped");
            return false;
        }

        return DatabaseDescriptor.getReplaceAddress() != null;
    }
`

smiklosovic: What is the current behavior? In trunk, if we decommission and it fails, it stays in `COMPLETED` bootstrap state. 

Then you restart a node and you want to do what? To "replace" a node? Why would you want to replace a node which failed to decommission itself?

I think that is undefined behavior which is currently not covered. I think that it should be like this:

    if (REPLACE_ADDRESS_FIRST_BOOT.getString() != null 
        && (SystemKeyspace.wasDecommissioned()
               || SystemKeyspace.hasDecommissionFailed()
               || SystemKeyspace.bootstrapComplete()))
    {
        return false;


Basically, we should not proceed with replacing procedure if we see that the node was already bootstrapped or it failed to decommission or it was decommissioned.

Maxwell-Guo: > Then you restart a node and you want to do what? To "replace" a node? Why would you want to replace a node which failed to decommission itself?

I am not going to replace a node after failed to decommission, I am just saying , the newly added STATUS of  decommission_failed can change the old behavior.  

Because , for us, we actually don’t know how users will use it, but what we can guarantee is that if we don’t know how they use it, it is better to be consistent with the original behavior, or we clearly stated in the document that we will What changes.
But for a managed cassandra service that users do not need to care about the Cassandra's operation and maintenance, of course it is ok .

src/java/org/apache/cassandra/service/StorageService.java
Maxwell-Guo: the logging level have restored to info from warn ~~~

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5182,42 +5215,48 @@ public void decommission(boolean force) throws InterruptedException
             }
 
             startLeaving();
-            long timeout = Math.max(RING_DELAY_MILLIS, BatchlogManager.instance.getBatchlogTimeout());
+            long timeout = Math.max(RING_DELAY_MILLIS, BatchlogManager.getBatchlogTimeout());
```
driftx, line 5217: Is there some nuance to this that I don't see, or is it unnecessary?

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -118,6 +118,7 @@
 import org.apache.cassandra.db.SizeEstimatesRecorder;
 import org.apache.cassandra.db.SnapshotDetailsTabularData;
 import org.apache.cassandra.db.SystemKeyspace;
+import org.apache.cassandra.db.SystemKeyspace.BootstrapState;
```
driftx: Do we really need this convenience?  We still have to import SystemKeyspace so it doesn't seem like much savings.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -5150,27 +5175,35 @@ public void decommission(boolean force) throws InterruptedException
...
-            if (operationMode != Mode.LEAVING) // If we're already decommissioning there is no point checking RF/pending ranges
+            // If we're already decommissioning there is no point checking RF/pending ranges
+            if (operationMode != Mode.LEAVING)
             {
                 int rf, numNodes;
                 for (String keyspaceName : Schema.instance.getNonLocalStrategyKeyspaces().names())
                 {
                     if (!force)
                     {
+                        boolean throwException = false;
```
driftx: can we name this more descriptively, like 'sufficientNodes' or something?

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: @driftx IDEA was telling me that we are accessing static variable `BATCHLOG_REPLAY_TIMEOUT` via instance reference which just does not make sense, does it? It is unnecessary to do it like that.

driftx: I see. I'm fine with keeping it, was just curious if I was missing something, thanks!

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: sure, I can revert that.

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: renamed to `notEnoughLiveNodes`

PR #2391, opened 2023-06-06
Cassandra 18569 trunk

bereng:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -1424,6 +1424,8 @@ public class DatabaseDescriptor
         if (selectedFormat == null)
             throw new ConfigurationException(String.format("Selected sstable format '%s' is not available.", selectedFormatName));
 
+        getStorageCompatibilityMode().validateSstableFormat(selectedFormat);
+
         return selectedFormat;
     }
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,10 +21,10 @@ jobs:
   j8_jvm_upgrade_dtests:
     docker:
     - image: apache/cassandra-testing-ubuntu2004-java11-w-dependencies:latest
-    resource_class: medium
+    resource_class: xlarge
     working_directory: ~/
     shell: /bin/bash -eo pipefail -l
-    parallelism: 1
+    parallelism: 4
     steps:
     - attach_workspace:
         at: /home/cassandra
@@ -141,7 +141,7 @@ jobs:
     resource_class: medium
     working_directory: ~/
     shell: /bin/bash -eo pipefail -l
-    parallelism: 4
+    parallelism: 25
     steps:
     - attach_workspace:
         at: /home/cassandra
@@ -292,10 +292,10 @@ jobs:
   j8_cqlsh_dtests_py311_vnode:
     docker:
     - image: apache/cassandra-testing-ubuntu2004-java11-w-dependencies:latest
-    resource_class: medium
+    resource_class: large
     working_directory: ~/
     shell: /bin/bash -eo pipefail -l
-    parallelism: 4
+    parallelism: 50
     steps:
     - attach_workspace:
         at: /home/cassandra
@@ -400,10 +400,10 @@ jobs:
   j11_dtests_vnode_repeat:
     docker:
     - image: apache/cassandra-testing-ubuntu2004-java11:latest
-    resource_class: medium
+    resource_class: large
     working_directory: ~/
     shell: /bin/bash -eo pipefail -l
-    parallelism: 4
+    parallelism: 25
     steps:
     - attach_workspace:
         at: /home/cassandra
@@ -555,7 +555,7 @@ jobs:
   j8_dtests_large_vnode:
     docker:
     - image: apache/cassandra-testing-ubuntu2004-java11-w-dependencies:latest
-    resource_class: medium
+    resource_class: xlarge
     working_directory: ~/
     shell: /bin/bash -eo pipefail -l
     parallelism: 4
@@ -642,7 +642,7 @@ jobs:
     resource_class: medium
     working_directory: ~/
     shell: /bin/bash -eo pipefail -l
-    parallelism: 4
+    parallelism: 25
     steps:
     - attach_workspace:
         at: /home/cassandra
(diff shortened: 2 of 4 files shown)


bereng:
Ci is [green](https://app.circleci.com/pipelines/github/bereng/cassandra/974/workflows/0d4cecaf-462a-4051-998b-d8852095d3e6)

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1424,9 +1425,20 @@ private static void validateSSTableFormatFactories(Iterable<SSTableFormat.Factor
...
 
+        validateWriteFormatVsStorageCompatibilityMode(selectedFormat, getStorageCompatibilityMode());
+
         return selectedFormat;
     }
 
+    @VisibleForTesting
+    public static void validateWriteFormatVsStorageCompatibilityMode(SSTableFormat<?, ?> selectedFormat, StorageCompatibilityMode mode)
+    {
+        if (selectedFormat.name().equals(BtiFormat.NAME) && mode != StorageCompatibilityMode.NONE)
```
adelapena: Why is BTI rejected with `StorageCompatibilityMode.UPGRADING`?

test/unit/org/apache/cassandra/config/DatabaseDescriptorTest.java
```diff
@@ -803,6 +805,15 @@ public void testDefaultSslContextFactoryConfiguration()
         Assert.assertTrue(config.server_encryption_options.ssl_context_factory.parameters.isEmpty());
     }
 
+    @Test
+    public void testBtiFormatAndStorageCompatibilityMode()
```
adelapena: Shouldn't we test the three possible values of `StorageCompatibilityMode`?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1424,9 +1425,20 @@ private static void validateSSTableFormatFactories(Iterable<SSTableFormat.Factor
...
+            throw new ConfigurationException(String.format("Selected sstable format '%s' is not available when in storage compatibility mode '%s'.",
```
adelapena: Perhaps we could take a guardrail-ish approach and delegate the formatting of the message to `StorageCompatibilityMode`. Something like: https://github.com/adelapena/cassandra/commit/30119694b64beb09d354984738ebc74fb447f7a9

That way we can make sure that all compatibility validation exceptions have a similar formatting. wdyt?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
bereng: To be on the safe side and not muddy the waters further while on mixed compatibility mode clusters. That would just complicate things further for no/little benefit imo.

test/unit/org/apache/cassandra/config/DatabaseDescriptorTest.java
bereng: Yep can do that

adelapena: Actually, we can test all compatibility modes against bti and big, and make sure that the test will be updated on new compatibility modes: https://github.com/adelapena/cassandra/blob/30565119702200d7dac215abef33ebc123d8e62c/test/unit/org/apache/cassandra/config/DatabaseDescriptorTest.java#L810-L835

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
adelapena: `UPGRADING` allows testing a subset of the nodes with bti while keeping the other nodes on big. And that composes with other features, like enabling other version-dependant properties and tries on a subset of the nodes. 

The original patch for bti was intended to work on the scenario described by `UPGRADING`. What cases do you think might not work on a mixed cluster so that bti should be disallowed until all nodes are upgraded? 

Wouldn't that force users to go first to `big-oa` on a first restart with `UPGRADING`, and then to `bti` on a second restart with `NONE`? If bti is supported on `UPGRADING` upgrades could skip the `nc->oa` step.

bereng: I really dislike having moving parts while `UPGRADING`. This means we'd have to test the whole codebase in `UPGRADING` which is yet another CI duplication of work. But tries might be a special case being behind a pretty solid feature flag. Also avoiding the extra oa hop while upgrading seems worth it. Lets do it for tries.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
bereng: I have addressed the latest 2 comments together. Mixed bag between you and me:
- Yes lets loop all values
- I have moved the validation into SCM. We'll most probably have more validations so having a centralized point, instead of adding and spreading noise to the callers, seems easier to handle. Everything seems tidier and easier to reason about imo.
- I have gone for a straight exception. The added indirection of `validateFeature` seemed unnecessary in this case and I favor the direct call here.

wdyt?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
adelapena: That situation, being able to support new features during rolling upgrades, is not new. I don't think that offering the possibility of running in compatibility mode changes it. IMO it doesn't remove the need of having the rolling upgrade dtests that we usually add to prevent multiplexing entire CI. For the scope of this ticket, I think we can assume that tries work in rolling upgrade mode, since that was the assumption when they were merged.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
adelapena: I like the idea of moving the method to `StorageCompatibilityMode`. That creates coupling between `StorageCompatibilityMode` and the sstable format, but that coupling allows to clearly see and centralize all dependencies. It also nicely hides the fact that `DatabaseDescriptor. validateWriteFormatVsStorageCompatibilityMode` was a separate method just to allow unit testing. The point of using a generic method on `StorageCompatibilityMode` was guaranteeing that all similar messages have the same format, like ending in `is not available when in storage compatibility mode '%s'`. We can add back that generic message formatting when we have more than one caller, adding it now is probably overkill.

adelapena (approved):

PR #2405, opened 2023-06-09
CASSANDRA-14351 Feat: notify systemd when startup is complete

tcataldo:
Notifying systemd will permit writing proper service files with cleaner starting behavior. 

https://issues.apache.org/jira/browse/CASSANDRA-14351

### DIFF
--- a/src/java/org/apache/cassandra/service/SystemD.java
+++ b/src/java/org/apache/cassandra/service/SystemD.java
@@ -1,0 +1,94 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/service/SystemD.java
```diff
@@ -0,0 +1,102 @@
...
+		if (!isAvailable()) {
+			throw new RuntimeException("SystemD is not available");
+		}
+		return INSTANCE;
+	}
+
+	private static Api init() {
+		SystemDLocation loc = figureOutLocation();
+		if (loc != null) {
+			RawApi rawApi = (RawApi) Native.loadLibrary(loc.systemdLibLocation, RawApi.class);
```
tcataldo: Maybe Native.loadLibrary("systemd") would work & would avoid trying multiple locations

tcataldo:
Hi, I updated the PR to let the linker figure the systemd library location instead of hardcoding some path. Hope this is ok.

pmcfadin:

PR #2416, opened 2023-06-14
fix Segment#intersects to compare bound instead of token

jasonstack:


src/java/org/apache/cassandra/index/sai/disk/v1/segment/Segment.java
```diff
@@ -87,14 +87,15 @@ public boolean intersects(AbstractBounds<PartitionPosition> keyRange)
         if (keyRange instanceof Range && ((Range<?>)keyRange).isWrapAround())
             return keyRange.contains(minKeyBound) || keyRange.contains(maxKeyBound);
 
-        int cmp = keyRange.right.getToken().compareTo(minKey);
+        int cmp = keyRange.right.compareTo(minKeyBound);
         // if right is minimum, it means right is the max token and bigger than maxKey.
         // if right bound is less than minKey, no intersection
```
maedhroz:
```suggestion
        // if right bound is less than minKeyBound, no intersection
```

src/java/org/apache/cassandra/index/sai/disk/v1/segment/Segment.java
```diff
@@ -87,14 +87,15 @@ public boolean intersects(AbstractBounds<PartitionPosition> keyRange)
...
         if (!keyRange.right.isMinimum() && (!keyRange.inclusiveRight() && cmp == 0 || cmp < 0))
             return false;
 
+        cmp = keyRange.left.compareTo(maxKeyBound);
         // if left bound is bigger than maxKey, no intersection
-        return keyRange.isStartInclusive() || keyRange.left.getToken().compareTo(maxKey) < 0;
+        return (keyRange.isStartInclusive() || cmp != 0) && cmp <= 0;
```
maedhroz, line 92: With the changes in this method, we no longer need the `minKey` or `maxKey` fields in this class, right?

CC @mike-tr-adamson

src/java/org/apache/cassandra/index/sai/disk/v1/segment/Segment.java
```diff
@@ -87,14 +87,15 @@ public boolean intersects(AbstractBounds<PartitionPosition> keyRange)
...
 
-        int cmp = keyRange.right.getToken().compareTo(minKey);
+        int cmp = keyRange.right.compareTo(minKeyBound);
         // if right is minimum, it means right is the max token and bigger than maxKey.
         // if right bound is less than minKey, no intersection
         if (!keyRange.right.isMinimum() && (!keyRange.inclusiveRight() && cmp == 0 || cmp < 0))
             return false;
 
+        cmp = keyRange.left.compareTo(maxKeyBound);
         // if left bound is bigger than maxKey, no intersection
```
maedhroz:
```suggestion
        // if left bound is bigger than maxKeyBound, no intersection
```

maedhroz (changes requested):
@jasonstack Mostly looks good. Just a few little things to clean up as a result of the changes...

src/java/org/apache/cassandra/index/sai/disk/v1/segment/Segment.java
jasonstack: removed `minKey` and `maxKey`

maedhroz (approved):

PR #2420, opened 2023-06-15
CASSANDRA-18345 Stream all components registered by an SSTable

pkolaczk:
Custom indexes may register custom components.
We need to include them in streaming as well.

### DIFF
--- a/src/java/org/apache/cassandra/db/streaming/ComponentManifest.java
+++ b/src/java/org/apache/cassandra/db/streaming/ComponentManifest.java
@@ -24,6 +24,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
@@ -33,7 +34,7 @@ import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.db.TypeSizes;
 import org.apache.cassandra.io.IVersionedSerializer;
 import org.apache.cassandra.io.sstable.Component;
-import org.apache.cassandra.io.sstable.Descriptor;
+import org.apache.cassandra.io.sstable.SSTable;
 import org.apache.cassandra.io.sstable.format.SSTableFormat;
 import org.apache.cassandra.io.util.DataInputPlus;
 import org.apache.cassandra.io.util.DataOutputPlus;
@@ -52,13 +53,17 @@ public final class ComponentManifest implements Iterable<Component>
     }
 
     @VisibleForTesting
-    public static ComponentManifest create(Descriptor descriptor)
+    public static ComponentManifest create(SSTable sstable)
     {
-        LinkedHashMap<Component, Long> components = new LinkedHashMap<>(descriptor.getFormat().streamingComponents().size());
+        Set<Component> streamingComponents = sstable.getComponents();
+        LinkedHashMap<Component, Long> components = new LinkedHashMap<>(streamingComponents.size());
 
-        for (Component component : descriptor.getFormat().streamingComponents())
+        for (Component component : streamingComponents)
         {
-            File file = descriptor.fileFor(component);
+            if (component == SSTableFormat.Components.TOC)
+                continue;
+
+            File file = sstable.descriptor.fileFor(component);
             if (!file.exists())
                 continue;
 
--- a/src/java/org/apache/cassandra/db/streaming/ComponentManifest.java
+++ b/src/java/org/apache/cassandra/db/streaming/ComponentManifest.java
@@ -55,14 +55,11 @@ public final class ComponentManifest implements Iterable<Component>
     @VisibleForTesting
     public static ComponentManifest create(SSTable sstable)
     {
-        Set<Component> streamingComponents = sstable.getComponents();
+        Set<Component> streamingComponents = sstable.getStreamingComponents();
         LinkedHashMap<Component, Long> components = new LinkedHashMap<>(streamingComponents.size());
 
         for (Component component : streamingComponents)
         {
-            if (component == SSTableFormat.Components.TOC)
-                continue;
-
             File file = sstable.descriptor.fileFor(component);
             if (!file.exists())
                 continue;
--- a/src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
+++ b/src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
@@ -20,6 +20,11 @@ package org.apache.cassandra.index.sai.disk.format;
 
 import org.apache.cassandra.index.sai.disk.v1.postings.PostingsWriter;
 import org.apache.cassandra.index.sai.disk.v1.trie.TrieTermsDictionaryWriter;
+import org.apache.cassandra.io.sstable.Component;
+
+import static org.apache.cassandra.index.sai.disk.format.Version.SAI_DESCRIPTOR;
+import static org.apache.cassandra.index.sai.disk.format.Version.SAI_SEPARATOR;
+
 
 /**
  * This is a definitive list of all the on-disk components for all versions
@@ -79,10 +84,20 @@ public enum IndexComponent
      */
     GROUP_COMPLETION_MARKER("GroupComplete");
 
+    public final Component.Type type;
     public final String name;
 
     IndexComponent(String name)
     {
         this.name = name;
+        this.type = componentType(name);
+    }
+
+    private static Component.Type componentType(String name)
+    {
+        return Component.Type.create(SAI_DESCRIPTOR + SAI_SEPARATOR + name.toUpperCase(),
+                                     SAI_DESCRIPTOR + ".*." + name + ".db",
+                                     true,
+                                     null);
     }
 }
--- a/src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
+++ b/src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
@@ -18,6 +18,8 @@
 
 package org.apache.cassandra.index.sai.disk.format;
 
+import java.util.regex.Pattern;
+
 import org.apache.cassandra.index.sai.disk.v1.postings.PostingsWriter;
 import org.apache.cassandra.index.sai.disk.v1.trie.TrieTermsDictionaryWriter;
 import org.apache.cassandra.io.sstable.Component;
@@ -84,8 +86,8 @@ public enum IndexComponent
      */
     GROUP_COMPLETION_MARKER("GroupComplete");
 
-    public final Component.Type type;
     public final String name;
+    public final Component.Type type;
 
     IndexComponent(String name)
     {
@@ -95,9 +97,10 @@ public enum IndexComponent
 
     private static Component.Type componentType(String name)
     {
-        return Component.Type.create(SAI_DESCRIPTOR + SAI_SEPARATOR + name.toUpperCase(),
-                                     SAI_DESCRIPTOR + ".*." + name + ".db",
-                                     true,
-                                     null);
+        String componentName = SAI_DESCRIPTOR + SAI_SEPARATOR + name;
+        String repr = Pattern.quote(SAI_DESCRIPTOR + SAI_SEPARATOR)
+                      + ".*"
+                      + Pattern.quote(SAI_SEPARATOR + name + ".db");
+        return Component.Type.create(componentName, repr, true, null);
     }
 }
--- a/src/java/org/apache/cassandra/io/sstable/Component.java
+++ b/src/java/org/apache/cassandra/io/sstable/Component.java
@@ -63,12 +63,12 @@ public class Component
          *
          * @param name         type name, must be unique for this and all parent formats
          * @param repr         the regular expression to be used to recognize a name represents this type
-         * @param isStreamable whether components of this type should be streamed to other nodes
+         * @param streamable   whether components of this type should be streamed to other nodes
          * @param formatClass  format class for which this type is defined for
          */
-        public static Type create(String name, String repr, boolean isStreamable, Class<? extends SSTableFormat<?, ?>> formatClass)
+        public static Type create(String name, String repr, boolean streamable, Class<? extends SSTableFormat<?, ?>> formatClass)
         {
-            return new Type(name, repr, false, isStreamable, formatClass);
+            return new Type(name, repr, false, streamable, formatClass);
         }
 
         /**
@@ -76,12 +76,12 @@ public class Component
          *
          * @param name         type name, must be unique for this and all parent formats
          * @param repr         the regular expression to be used to recognize a name represents this type
-         * @param isStreamable whether components of this type should be streamed to other nodes
+         * @param streamable   whether components of this type should be streamed to other nodes
          * @param formatClass  format class for which this type is defined for
          */
-        public static Type createSingleton(String name, String repr, boolean isStreamable, Class<? extends SSTableFormat<?, ?>> formatClass)
+        public static Type createSingleton(String name, String repr, boolean streamable, Class<? extends SSTableFormat<?, ?>> formatClass)
         {
-            return new Type(name, repr, true, isStreamable, formatClass);
+            return new Type(name, repr, true, streamable, formatClass);
         }
 
         private Type(String name, String repr, boolean isSingleton, boolean streamable, Class<? extends SSTableFormat<?, ?>> formatClass)
--- a/src/java/org/apache/cassandra/io/sstable/Component.java
+++ b/src/java/org/apache/cassandra/io/sstable/Component.java
@@ -52,6 +52,7 @@ public class Component
         public final int id;
         public final String name;
         public final String repr;
+        public final boolean streamable;
         private final Component singleton;
 
         @SuppressWarnings("rawtypes")
@@ -60,31 +61,34 @@ public class Component
         /**
          * Creates a new non-singleton type and registers it a global type registry - see {@link #registerType(Type)}.
          *
-         * @param name        type name, must be unique for this and all parent formats
-         * @param repr        the regular expression to be used to recognize a name represents this type
-         * @param formatClass format class for which this type is defined for
+         * @param name         type name, must be unique for this and all parent formats
+         * @param repr         the regular expression to be used to recognize a name represents this type
+         * @param isStreamable whether components of this type should be streamed to other nodes
+         * @param formatClass  format class for which this type is defined for
          */
-        public static Type create(String name, String repr, Class<? extends SSTableFormat<?, ?>> formatClass)
+        public static Type create(String name, String repr, boolean isStreamable, Class<? extends SSTableFormat<?, ?>> formatClass)
         {
-            return new Type(name, repr, false, formatClass);
+            return new Type(name, repr, false, isStreamable, formatClass);
         }
 
         /**
          * Creates a new singleton type and registers it in a global type registry - see {@link #registerType(Type)}.
          *
-         * @param name        type name, must be unique for this and all parent formats
-         * @param repr        the regular expression to be used to recognize a name represents this type
-         * @param formatClass format class for which this type is defined for
+         * @param name         type name, must be unique for this and all parent formats
+         * @param repr         the regular expression to be used to recognize a name represents this type
+         * @param isStreamable whether components of this type should be streamed to other nodes
+         * @param formatClass  format class for which this type is defined for
          */
-        public static Type createSingleton(String name, String repr, Class<? extends SSTableFormat<?, ?>> formatClass)
+        public static Type createSingleton(String name, String repr, boolean isStreamable, Class<? extends SSTableFormat<?, ?>> formatClass)
         {
-            return new Type(name, repr, true, formatClass);
+            return new Type(name, repr, true, isStreamable, formatClass);
         }
 
-        private Type(String name, String repr, boolean isSingleton, Class<? extends SSTableFormat<?, ?>> formatClass)
+        private Type(String name, String repr, boolean isSingleton, boolean streamable, Class<? extends SSTableFormat<?, ?>> formatClass)
         {
             this.name = Objects.requireNonNull(name);
             this.repr = repr;
+            this.streamable = streamable;
             this.id = typesCollector.size();
             this.formatClass = formatClass == null ? SSTableFormat.class : formatClass;
             this.singleton = isSingleton ? new Component(this) : null;
--- a/src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
@@ -65,7 +65,7 @@ public class SSTableZeroCopyWriter extends SSTable implements SSTableMultiWriter
                                                .filter(c -> !c.type.streamable)
                                                .collect(Collectors.toSet());
         if (!unsupported.isEmpty())
-            throw new AssertionError(format("Unsupported streaming component detected %s", unsupported));
+            throw new AssertionError(format("Unsupported streaming components detected: %s", unsupported));
 
         for (Component c : components)
             componentWriters.put(c.name, makeWriter(descriptor, c));
--- a/src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
@@ -199,6 +199,7 @@ public class SSTableZeroCopyWriter extends SSTable implements SSTableMultiWriter
 
     public void writeComponent(Component component, DataInputPlus in, long size) throws ClosedChannelException
     {
+        @SuppressWarnings({"resource", "RedundantSuppression"})  // all writers are closed in close()
         SequentialWriter writer = componentWriters.get(component.name);
         logger.info("Writing component {} to {} length {}", component, writer.getPath(), prettyPrintMemory(size));
 
--- a/src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
@@ -36,7 +36,6 @@ import org.apache.cassandra.db.rows.UnfilteredRowIterator;
 import org.apache.cassandra.dht.AbstractBounds;
 import org.apache.cassandra.dht.Token;
 import org.apache.cassandra.io.FSWriteError;
-import org.apache.cassandra.io.sstable.format.SSTableFormat;
 import org.apache.cassandra.io.sstable.format.SSTableReader;
 import org.apache.cassandra.io.util.DataInputPlus;
 import org.apache.cassandra.io.util.SequentialWriter;
@@ -63,8 +62,7 @@ public class SSTableZeroCopyWriter extends SSTable implements SSTableMultiWriter
         this.componentWriters = new HashMap<>();
 
         Set<Component> unsupported = components.stream()
-                                               .filter(c -> !descriptor.getFormat().streamingComponents().contains(c))
-                                               .filter(c-> c.type != SSTableFormat.Components.Types.CUSTOM)
+                                               .filter(c -> !c.type.streamable)
                                                .collect(Collectors.toSet());
         if (!unsupported.isEmpty())
             throw new AssertionError(format("Unsupported streaming component detected %s", unsupported));
--- a/src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
@@ -25,8 +25,6 @@ import java.util.HashMap;
 import java.util.Map;
 
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
-import com.google.common.collect.Sets;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -42,7 +40,6 @@ import org.apache.cassandra.io.util.SequentialWriter;
 import org.apache.cassandra.net.AsyncStreamingInputPlus;
 import org.apache.cassandra.schema.TableId;
 
-import static java.lang.String.format;
 import static org.apache.cassandra.utils.FBUtilities.prettyPrintMemory;
 
 public class SSTableZeroCopyWriter extends SSTable implements SSTableMultiWriter
@@ -50,7 +47,7 @@ public class SSTableZeroCopyWriter extends SSTable implements SSTableMultiWriter
     private static final Logger logger = LoggerFactory.getLogger(SSTableZeroCopyWriter.class);
 
     private volatile SSTableReader finalReader;
-    private final Map<Component.Type, SequentialWriter> componentWriters;
+    private final Map<String, SequentialWriter> componentWriters; // indexed by component name
 
     public SSTableZeroCopyWriter(Builder<?, ?> builder,
                                  LifecycleNewTracker lifecycleNewTracker,
@@ -61,12 +58,8 @@ public class SSTableZeroCopyWriter extends SSTable implements SSTableMultiWriter
         lifecycleNewTracker.trackNew(this);
         this.componentWriters = new HashMap<>();
 
-        if (!descriptor.getFormat().streamingComponents().containsAll(components))
-            throw new AssertionError(format("Unsupported streaming component detected %s",
-                                            Sets.difference(ImmutableSet.copyOf(components), descriptor.getFormat().streamingComponents())));
-
         for (Component c : components)
-            componentWriters.put(c.type, makeWriter(descriptor, c));
+            componentWriters.put(c.name, makeWriter(descriptor, c));
     }
 
     @Override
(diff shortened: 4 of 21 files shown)


maedhroz:
nit: There are some unused import errors on J11 checkstyle

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.cassandra.distributed.util.QueryResultUtil;
+import org.assertj.core.api.Assertions;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class IndexStreamingTest extends TestBaseImpl
+{
+    private static final ByteBuffer BLOB = ByteBuffer.wrap(new byte[1 << 16]);
+    // zero copy streaming sends all components, so the events will include non-Data files as well
+    private static final int NUM_COMPONENTS = 17;
```
maedhroz: nit: We can probably make this a little more self-documenting if we sum up `V1OnDiskFormat.PER_SSTABLE_COMPONENTS`, `V1OnDiskFormat.LITERAL_COMPONENTS`, and the expected core SSTable components (perhaps from `SSTableFormat#streamingComponents()` sans compression info, which I guess we don't have here?)

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
```diff
@@ -0,0 +1,134 @@
...
+                                                             .set("stream_entire_sstables", zeroCopyStreaming).set("streaming_slow_events_log_timeout", "0s"))
+                                           .start()))
+        {
+            // streaming sends events every 65k, so need to make sure that the files are larger than this to hit
+            // all cases of the vtable
+            cluster.schemaChange(withKeyspace(
+                "CREATE TABLE %s.test (pk int PRIMARY KEY, v text, b blob) WITH compression = { 'enabled' : false };"
+            ));
+            cluster.schemaChange(withKeyspace(
+                "CREATE CUSTOM INDEX ON %s.test(v) USING 'StorageAttachedIndex';"
```
maedhroz, line 87: Once numerics support merges, I guess we'll want to parameterize this test further to use those. The number of expected index components might not actually change though. CC @mike-tr-adamson 

I'm not sure if that means we should merge this first and adjust inside a rebased CASSANDRA-18067 or the other way around.

src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
```diff
@@ -61,12 +58,8 @@ public SSTableZeroCopyWriter(Builder<?, ?> builder,
         lifecycleNewTracker.trackNew(this);
         this.componentWriters = new HashMap<>();
 
-        if (!descriptor.getFormat().streamingComponents().containsAll(components))
-            throw new AssertionError(format("Unsupported streaming component detected %s",
-                                            Sets.difference(ImmutableSet.copyOf(components), descriptor.getFormat().streamingComponents())));
-
```
maedhroz, line 67: @pkolaczk @mike-tr-adamson I'm not sure we want to just remove this validation, but checking SAI components raises some other questions. The simplest way around this for now might be...

```
Set<Component> unsupported = components.stream()
                                       .filter(c -> !descriptor.getFormat().streamingComponents().contains(c))
                                       .filter(c-> c.type != SSTableFormat.Components.Types.CUSTOM)
                                       .collect(Collectors.toSet());

if (!unsupported.isEmpty())
    throw new AssertionError(format("Unsupported streaming component detected %s", unsupported));
```

However, it might also be worth looking at why we use `CUSTOM` for SAI components in the first place. Is now the time to simply create a new first-class type in `Types`?

```
public static final Component.Type SAI = Component.Type.create("SAI", "SAI\\+.*.db", null);
```

(The regex could be stricter, but details...)

If we do this, we the `CUSTOM` in the check above becomes `SAI`, we replace `CUSTOM` w/ `SAI` in a few places in the SAI code, like `Version`, and things should just work, right?

maedhroz: btw, I just did exactly this, and it works fine...at least `IndexStreamingTest` has no issues

src/java/org/apache/cassandra/db/streaming/ComponentManifest.java
```diff
@@ -52,13 +53,17 @@ public ComponentManifest(Map<Component, Long> components)
...
     {
-        LinkedHashMap<Component, Long> components = new LinkedHashMap<>(descriptor.getFormat().streamingComponents().size());
+        Set<Component> streamingComponents = sstable.getComponents();
+        LinkedHashMap<Component, Long> components = new LinkedHashMap<>(streamingComponents.size());
 
-        for (Component component : descriptor.getFormat().streamingComponents())
+        for (Component component : streamingComponents)
         {
-            File file = descriptor.fileFor(component);
+            if (component == SSTableFormat.Components.TOC)
```
maedhroz: nit: I guess the spirit of the original code here was to delegate this as much as possible to the SSTable format. What if we tried to preserve that (and coupled it with the `SAI` component I suggested below) and did something like...

```
if (!sstable.descriptor.getFormat().streamingComponents().contains(component) && component.type != BigFormat.Components.Types.SAI)
```
I like the idea of using `sstable.getComponents()` to make sure we don't forget anything, but it seems like picking out the `TOC` alone is a little brittle. WDYT?

maedhroz: We could also hide this in `SSTable` with a `getStreamingComponents()` and leave this logic in the manifest almost unchanged? It already makes a copy of `components`, and this isn't really a hot path anyway, so filtering there w/ something like...

```
public Set<Component> getStreamingComponents()
{
    return components.stream()
                     .filter(c -> descriptor.getFormat().streamingComponents().contains(c) || c.type == Components.Types.SAI)
                     .collect(Collectors.toSet());
}
```

maedhroz (changes requested):
The approach looks fine here. There are just a few things we should at least double check in terms of how we generate and validate the components we want to stream (...and we have to decide whether to merge this now and make the numerics patch add some testing or wait for numerics and test in this patch).

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
pkolaczk: Good idea. Fixed

src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
pkolaczk: Why do we want to artificially restrict the set of components streamed in the place that can stream any component?
It's not like streaming supports only some kind of components and doesn't support some other types. It will copy anything we'll throw at it.

src/java/org/apache/cassandra/db/streaming/ComponentManifest.java
pkolaczk: First, I like the idea of having `getStreamingComponents` method on the sstable. This looks like a better place to tell which components are eligible for streaming. 

However, even with such method, we still have a "responsibility leak" from SAI here. 

```
c.type == Components.Types.SAI
```

Why should SSTable code know anything about SAI? Why should it assume all SAI components should be streamed?  IMHO this is something that should be decided by the SAI code (or whoever owns the component), not by sstable code.

SAI already controls the type of the component, so I guess the ideal-world solution would be to have this information associated directly with the type so instead of special-casing SAI vs core components, we could write the condition like this:

```
 return components.stream()
                     .filter(c -> c.type.shouldBeStreamed)
                     .collect(Collectors.toSet());
```

WDYT?

src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
pkolaczk: As for this:
```
public static final Component.Type SAI = Component.Type.create("SAI", "SAI\\+.*.db", null);
```

This wouldn't be very consistent with what we use component types for now in the core - because now each core sstable component  has its own type. So why should we throw all different SAI component types into a single type? 

I think we need to take another look at how custom components are registered and simply allow modules to provide their own types. So SAI component types should be defined in SAI. We could also allow the component type to decide whether it should be streamed or not.

pkolaczk:
I addressed all review comments, except the part about introducing the SAI type for SAI components, because IMHO that needs a wider discussion.

src/java/org/apache/cassandra/db/streaming/ComponentManifest.java
maedhroz: Yeah, at the end of the day, some things are just not part of the SSTable format, so if/when that is the case, having the component decide whether it's "streamable" doesn't seem like a bad idea. (Or we could do that for all components.)

src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
maedhroz: > I think we need to take another look at how custom components are registered and simply allow modules to provide their own types. So SAI component types should be defined in SAI. We could also allow the component type to decide whether it should be streamed or not.

I think this would work. +1

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
```diff
@@ -59,7 +60,18 @@ public interface SSTableFormat<R extends SSTableReader, W extends SSTableWriter>
...
     Set<Component> allComponents();
 
-    Set<Component> streamingComponents();
+    /**
+     * Returns the components that should be streamed to other nodes on repair / rebuild.
+     * This includes only the core SSTable components produced by this format.
+     * Custom components registered by e.g. secondary indexes are not included.
+     * Use {@link SSTableReader#getStreamingComponents()} for the list of all components including the custom ones.
+     */
+    default Set<Component> streamingComponents()
```
maedhroz: Once all `Component` instances have their "streamability" embedded, we don't need this method at all, right? i.e. `SSTableZeroCopyWriter` will no longer need special casing?

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
```diff
@@ -156,23 +168,23 @@ public static class Types
...
+            public static final Component.Type DIGEST = Component.Type.createSingleton("DIGEST", "Digest.crc32", true, null);
             // holds the CRC32 for chunks in an uncompressed file.
-            public static final Component.Type CRC = Component.Type.createSingleton("CRC", "CRC.db", null);
+            public static final Component.Type CRC = Component.Type.createSingleton("CRC", "CRC.db", true, null);
             // table of contents, stores the list of all components for the sstable
-            public static final Component.Type TOC = Component.Type.createSingleton("TOC", "TOC.txt", null);
+            public static final Component.Type TOC = Component.Type.createSingleton("TOC", "TOC.txt", false, null);
             // built-in secondary index (may exist multiple per sstable)
-            public static final Component.Type SECONDARY_INDEX = Component.Type.create("SECONDARY_INDEX", "SI_.*.db", null);
+            public static final Component.Type SECONDARY_INDEX = Component.Type.create("SECONDARY_INDEX", "SI_.*.db", true, null);
```
maedhroz: nit: Are `SECONDARY_INDEX` components streamable? I thought we always rebuilt them at the receiver...

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
```diff
@@ -156,23 +168,23 @@ public static class Types
...
             // custom component, used by e.g. custom compaction strategy
-            public static final Component.Type CUSTOM = Component.Type.create("CUSTOM", null, null);
+            public static final Component.Type CUSTOM = Component.Type.create("CUSTOM", null, true, null);
```
maedhroz, line 174: I guess not all CUSTOM components should be streamable?

Is the plan here to just create something like the other `Types` inner classes, but in SAI's `V1OnDiskFormat`? If we did that, we could specify "streamability" there in those types. For example:

`Component.Type TERMS_DATA = Component.Type.create("TermsData", "SAI\\+.*\\+TermsData.db", true, null);`

This is really where all the stuff in `IndexComponent` probably should have gone in the first place.

CC @mike-tr-adamson

maedhroz: ...or more completely, something like...

```
public static class Components
{
    public static class Types extends AbstractSSTableFormat.Components.Types
    {
        public static final Component.Type TERMS_DATA = Component.Type.create("TermsData", "SAI\\+.*\\+TermsData.db", true, null);
...
    }

    public final static Component TERMS_DATA = Types.TERMS_DATA.getSingleton();
...
}
```

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
pkolaczk: Good catch. I forgot to change it.

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
pkolaczk: Looks like we need it for testing. I'll think of a way to get rid of it completely.

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
pkolaczk: Yes, that is the plan. I haven't finished that yet.

src/java/org/apache/cassandra/io/sstable/format/big/BigFormat.java
```diff
@@ -82,9 +82,9 @@ public static class Components extends SSTableFormat.Components
         public static class Types extends SSTableFormat.Components.Types
         {
             // index of the row keys with pointers to their positions in the data file
-            public static final Component.Type PRIMARY_INDEX = Component.Type.createSingleton("PRIMARY_INDEX", "Index.db", BigFormat.class);
+            public static final Component.Type PRIMARY_INDEX = Component.Type.createSingleton("PRIMARY_INDEX", "Index.db", true, BigFormat.class);
```
adelapena, line 85: We'll need to update [the doc on `SSTable_API.md`](https://github.com/apache/cassandra/blob/cep-7-sai/src/java/org/apache/cassandra/io/sstable/SSTable_API.md#components) to include the new `streamable` parameter.

src/java/org/apache/cassandra/io/sstable/Component.java
```diff
@@ -60,31 +61,34 @@ public final static class Type
...
+         * @param name         type name, must be unique for this and all parent formats
+         * @param repr         the regular expression to be used to recognize a name represents this type
+         * @param isStreamable whether components of this type should be streamed to other nodes
+         * @param formatClass  format class for which this type is defined for
          */
-        public static Type create(String name, String repr, Class<? extends SSTableFormat<?, ?>> formatClass)
+        public static Type create(String name, String repr, boolean isStreamable, Class<? extends SSTableFormat<?, ?>> formatClass)
         {
-            return new Type(name, repr, false, formatClass);
+            return new Type(name, repr, false, isStreamable, formatClass);
```
adelapena: Nit: the parameter could be named `streamable` for consistency with the attribute and constructor names.

src/java/org/apache/cassandra/io/sstable/Component.java
```diff
@@ -60,31 +61,34 @@ public final static class Type
...
-         * @param name        type name, must be unique for this and all parent formats
-         * @param repr        the regular expression to be used to recognize a name represents this type
-         * @param formatClass format class for which this type is defined for
+         * @param name         type name, must be unique for this and all parent formats
+         * @param repr         the regular expression to be used to recognize a name represents this type
+         * @param isStreamable whether components of this type should be streamed to other nodes
+         * @param formatClass  format class for which this type is defined for
          */
-        public static Type createSingleton(String name, String repr, Class<? extends SSTableFormat<?, ?>> formatClass)
+        public static Type createSingleton(String name, String repr, boolean isStreamable, Class<? extends SSTableFormat<?, ?>> formatClass)
```
adelapena: Nit: the parameter could be named `streamable` for consistency with the attribute and constructor names.

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
```diff
@@ -0,0 +1,151 @@
...
+    public void notZeroCopy() throws IOException
+    {
+        test(false);
+    }
+
+    private void test(boolean zeroCopyStreaming) throws IOException
+    {
+        try (Cluster cluster = init(Cluster.build(2)
+                                           .withConfig(c -> c.with(Feature.values())
+                                                             .set("stream_entire_sstables", zeroCopyStreaming).set("streaming_slow_events_log_timeout", "0s"))
```
adelapena: Nit: break the long line
```suggestion
                                                             .set("stream_entire_sstables", zeroCopyStreaming)
                                                             .set("streaming_slow_events_log_timeout", "0s"))
```

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
```diff
@@ -0,0 +1,151 @@
...
+                                           .start()))
+        {
+            // streaming sends events every 65k, so need to make sure that the files are larger than this to hit
+            // all cases of the vtable
```
adelapena: Maybe this comment would be better placed either right before the `INSERT` or when declaring the `BLOB` size.

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
```diff
@@ -0,0 +1,151 @@
...
+            second.nodetoolResult("rebuild", "--keyspace", KEYSPACE).asserts().success();
+
+            SimpleQueryResult qr = first.executeInternalWithResult("SELECT * FROM system_views.streaming");
+            String txt = QueryResultUtil.expand(qr);
+            qr.reset();
+            assertThat(qr.toObjectArrays().length).describedAs("Found rows\n%s", txt).isEqualTo(1);
+            assertThat(qr.hasNext()).isTrue();
+            Row row = qr.next();
+            QueryResultUtil.assertThat(row)
+                           .isEqualTo("peers", Collections.singletonList("/127.0.0.2:7012"))
```
adelapena:
```suggestion
                           .isEqualTo("peers", Collections.singletonList(second.broadcastAddress().toString()))
```

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
```diff
@@ -0,0 +1,151 @@
...
+            assertThat(totalBytes).isGreaterThan(0);
+
+            qr = second.executeInternalWithResult("SELECT * FROM system_views.streaming");
+            txt = QueryResultUtil.expand(qr);
+            qr.reset();
+            assertThat(qr.toObjectArrays().length).describedAs("Found rows\n%s", txt).isEqualTo(1);
+            assertThat(qr.hasNext()).isTrue();
+
+            QueryResultUtil.assertThat(qr.next())
+                           .isEqualTo("peers", Collections.singletonList("/127.0.0.1:7012"))
```
adelapena:
```suggestion
                           .isEqualTo("peers", Collections.singletonList(first.broadcastAddress().toString()))
```

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
```diff
@@ -0,0 +1,151 @@
...
+                           .isEqualTo("operation", "Rebuild")
+                           .isEqualTo("status", "success")
+                           .isEqualTo("progress_percentage", 100.0F)
+                           .isEqualTo("success_message", null).isEqualTo("failure_cause", null)
+                           .columnsEqualTo("files_to_receive", "files_received").isEqualTo("files_received", expectedFiles)
+                           .columnsEqualTo("bytes_to_receive", "bytes_received").isEqualTo("bytes_received", totalBytes)
+                           .columnsEqualTo("files_sent", "files_to_send", "bytes_sent", "bytes_to_send").isEqualTo("files_sent", 0L);
+
+            // did we trigger slow event log?
+            cluster.forEach(i -> Assertions.assertThat(i.logs().grep("Handling streaming events took longer than").getResult()).describedAs("Unable to find slow log for node%d", i.config().num()).isNotEmpty());
```
adelapena: Nit: break the long line
```suggestion
            cluster.forEach(i -> Assertions.assertThat(i.logs().grep("Handling streaming events took longer than").getResult())
                                           .describedAs("Unable to find slow log for node%d", i.config().num())
                                           .isNotEmpty());
```

src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
```diff
@@ -195,14 +197,15 @@ public void close()
             writer.close();
     }
 
-    public void writeComponent(Component.Type type, DataInputPlus in, long size) throws ClosedChannelException
+    public void writeComponent(Component component, DataInputPlus in, long size) throws ClosedChannelException
     {
-        logger.info("Writing component {} to {} length {}", type, componentWriters.get(type).getPath(), prettyPrintMemory(size));
+        SequentialWriter writer = componentWriters.get(component.name);
```
adelapena, line 203: Build fails here with `eclipse-warnings` saying `Potential resource leak: 'writer' may not be closed`

src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
```diff
@@ -61,12 +61,14 @@ public SSTableZeroCopyWriter(Builder<?, ?> builder,
...
         this.componentWriters = new HashMap<>();
 
-        if (!descriptor.getFormat().streamingComponents().containsAll(components))
-            throw new AssertionError(format("Unsupported streaming component detected %s",
-                                            Sets.difference(ImmutableSet.copyOf(components), descriptor.getFormat().streamingComponents())));
+        Set<Component> unsupported = components.stream()
+                                               .filter(c -> !c.type.streamable)
+                                               .collect(Collectors.toSet());
+        if (!unsupported.isEmpty())
+            throw new AssertionError(format("Unsupported streaming component detected %s", unsupported));
```
maedhroz:
```suggestion
            throw new AssertionError(format("Unsupported streaming components detected: %s", unsupported));
```

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+    static
+    {
+        DatabaseDescriptor.clientInitialization();
+        NUM_COMPONENTS = sstableStreamingComponentsCount()
+                         + V1OnDiskFormat.PER_SSTABLE_COMPONENTS.size()
+                         + V1OnDiskFormat.LITERAL_COMPONENTS.size();
+    }
+
+    private static int sstableStreamingComponentsCount() {
```
maedhroz: nit: newline for `{`

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
```diff
@@ -0,0 +1,151 @@
...
+{
+    private static final ByteBuffer BLOB = ByteBuffer.wrap(new byte[1 << 16]);
+    private static final int NUM_COMPONENTS;
+
+    static
+    {
+        DatabaseDescriptor.clientInitialization();
+        NUM_COMPONENTS = sstableStreamingComponentsCount()
+                         + V1OnDiskFormat.PER_SSTABLE_COMPONENTS.size()
+                         + V1OnDiskFormat.LITERAL_COMPONENTS.size();
```
maedhroz, line 51: Nice. We'll expand on this whenever numerics merge.

CC @mike-tr-adamson

src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
```diff
@@ -79,10 +84,20 @@ public enum IndexComponent
...
     IndexComponent(String name)
     {
         this.name = name;
+        this.type = componentType(name);
+    }
+
+    private static Component.Type componentType(String name)
+    {
+        return Component.Type.create(SAI_DESCRIPTOR + SAI_SEPARATOR + name.toUpperCase(),
+                                     SAI_DESCRIPTOR + ".*." + name + ".db",
```
maedhroz: Looking at `Version.defaultFileNameFormat()`, I think you'd want something like:

`SAI_DESCRIPTOR + "\\+.*\\+" + name + ".db"`

`".*."` doesn't need the second period?

src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
```diff
@@ -79,10 +84,20 @@ public enum IndexComponent
...
 
     IndexComponent(String name)
     {
         this.name = name;
+        this.type = componentType(name);
+    }
+
+    private static Component.Type componentType(String name)
+    {
+        return Component.Type.create(SAI_DESCRIPTOR + SAI_SEPARATOR + name.toUpperCase(),
```
maedhroz: nit: `toUpperCase()` here seems unnecessary, as it will give us weird things like "GROUPCOMPLETE". Camel is ok?

src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
```diff
@@ -79,10 +84,20 @@ public enum IndexComponent
      */
     GROUP_COMPLETION_MARKER("GroupComplete");
 
+    public final Component.Type type;
```
maedhroz: super OCD nit: maybe `type` could go below `name` so they're in the same order in the ctor :D

maedhroz (approved):
Made a few more nits, but this looks pretty nice now.

Again, we'll have to modify/add some things here to deal w/ the numerics patch when it's ready (or in that patch if this merges), but that shouldn't be hard.

src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
pkolaczk: Fixed

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
pkolaczk: I simply moved to testing code.

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
pkolaczk: Fixed

src/java/org/apache/cassandra/io/sstable/format/SSTableFormat.java
pkolaczk: Fixed

src/java/org/apache/cassandra/io/sstable/format/big/BigFormat.java
pkolaczk: Fixed

src/java/org/apache/cassandra/io/sstable/Component.java
pkolaczk: Fixed

src/java/org/apache/cassandra/io/sstable/Component.java
pkolaczk: Fixed

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
pkolaczk: Indeed. This test has been mostly copied from RebuildStreamingTest. Those long lines are there as well.
Do you want me to cleanup long lines in there as well?

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
pkolaczk: Yup, moving it makes sense. Moved to above BLOB.

src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
pkolaczk: That looks like a false positive to me. It isn't supposed to be closed here. All writers are closed in the `close()` method. How do I suppress this?

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
pkolaczk: Fixed

src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
pkolaczk: Indeed. Fixed

src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
pkolaczk: +1

src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
adelapena: You can add `@SuppressWarnings("resource")` with a brief comment about where it will be closed. We have a bunch of those, for example [this one](https://github.com/apache/cassandra/blob/eda0bcbefaf38e4069ac6db3a00937d778090083/src/java/org/apache/cassandra/service/reads/range/RangeCommands.java#L55).

pkolaczk: I'm not getting this error when building. And the project has built fine in CircleCI. 
I tried adding `@SuppressWarnings("resource")` but intellij immediately marks it as redundant :shrug:

src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
pkolaczk: Fixed, and also added pattern quoting.

src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
adelapena: I get this error locally when running `eclipse-warnings` with JDK 8:
```
eclipse-warnings:
    [mkdir] Created dir: /Users/adelapena/src/cassandra/trunk/build/ecj
     [echo] Running Eclipse Code Analysis.  Output logged to /Users/adelapena/src/cassandra/trunk/build/ecj/eclipse_compiler_checks.txt
     [java] ----------
     [java] 1. ERROR in /Users/adelapena/src/cassandra/trunk/src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java (at line 202)
     [java] 	SequentialWriter writer = componentWriters.get(component.name);
     [java] 	                 ^^^^^^
     [java] Potential resource leak: 'writer' may not be closed
     [java] ----------
     [java] 1 problem (1 error)

BUILD FAILED
```
CircleCI also hits it for j8: https://app.circleci.com/pipelines/github/adelapena/cassandra/2986/workflows/82e3ccc4-83ce-49c8-ab0d-036d100cae8e/jobs/54244

However, it doesn't fail for j11 builds: https://app.circleci.com/pipelines/github/adelapena/cassandra/2986/workflows/554d79db-c157-4e53-944a-ced1dffc242f

Do you have a link to the CircleCI run? Maybe it was a j11-only run? j8 support will we dropped very soon ([CASSANDRA-18255](https://issues.apache.org/jira/browse/CASSANDRA-18255)), but we still need it.

pkolaczk: Ah! Gotcha. So that's the difference: JDK 8. I was using 11. Checking with JDK 8....

maedhroz: If you want IDEA to stop complaining, use `@SuppressWarnings({ "resource", "RedundantSuppression" })`

pkolaczk: fixed

pkolaczk:
CI run: https://app.circleci.com/pipelines/github/pkolaczk/cassandra/16/workflows/41cc8da8-cf2b-4985-be0c-cf5ffd9598d8

src/java/org/apache/cassandra/io/sstable/SSTableZeroCopyWriter.java
```diff
@@ -195,14 +197,16 @@ public void close()
             writer.close();
     }
 
-    public void writeComponent(Component.Type type, DataInputPlus in, long size) throws ClosedChannelException
+    public void writeComponent(Component component, DataInputPlus in, long size) throws ClosedChannelException
     {
-        logger.info("Writing component {} to {} length {}", type, componentWriters.get(type).getPath(), prettyPrintMemory(size));
+        @SuppressWarnings({"resource", "RedundantSuppression"})  // all writers are closed in close()
```
maedhroz, line 202: Curious...can individual writers fail and cause some writers not to be closed? (above, on 197)

adelapena: I guess if that's the case, we could use `FBUtitilites#closeAll` at `SSTableZeroCopyWriter#close`. But we are not modifying that here, so I think it can be done on a separate ticket.

adelapena (approved):

maedhroz:
Committed as https://github.com/apache/cassandra/commit/5d3f257477cba2d7f33f842dba4582d0660f5738

PR #2421, opened 2023-06-16
Cassandra 18180 trunk

ekaterinadimitrova2:
 
```

```

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1060,7 +1060,7 @@
   </target>
 
   <target name="_check-test-names">
-    <java taskname="check-test-names" fork="true" classname="org.apache.cassandra.anttasks.TestNameCheckTask" classpath="${test.classes}">
+    <java taskname="check-test-names" fork="true" failonerror="yes" classname="org.apache.cassandra.anttasks.TestNameCheckTask" classpath="${test.classes}">
       <classpath>
         <path refid="cassandra.classpath" />
         <pathelement location="${test.classes}"/>
--- a/build.xml
+++ b/build.xml
@@ -1887,7 +1887,10 @@
         <echo file=".idea/compiler.xml"><![CDATA[<?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="JavacSettings">
-    <option name="ADDITIONAL_OPTIONS_STRING" value="--add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED --add-exports java.rmi/sun.rmi.transport.tcp=ALL-UNNAMED" />
+    <option name="ADDITIONAL_OPTIONS_STRING" value="--add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED
+                                                    --add-exports java.rmi/sun.rmi.transport.tcp=ALL-UNNAMED
+                                                    --add-exports java.base/jdk.internal.ref=ALL-UNNAMED
+                                                    --add-exports java.base/sun.nio.ch=ALL-UNNAMED" />
   </component>
 </project>]]></echo>
         <echo>"IDE configuration updated for use with JDK17"</echo>
--- a/build.xml
+++ b/build.xml
@@ -310,9 +310,17 @@
         <equals arg1="${ant.java.version}" arg2="17"/>
     </condition>
 
-    <!-- needed to compile org.apache.cassandra.utils.JMXServerUtils -->
-    <!-- needed to compile org.apache.cassandra.distributed.impl.Instance-->
-    <condition property="jdk11plus-javac-exports" value="--add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED --add-exports java.rmi/sun.rmi.transport.tcp=ALL-UNNAMED" else="">
+    <resources id="_jdk11plus_javac_export_args">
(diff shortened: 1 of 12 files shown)


src/java/org/apache/cassandra/utils/concurrent/Ref.java
```diff
@@ -91,7 +93,7 @@
  * Once the Ref.GlobalState has been completely released, the Tidy method is called and it removes the global reference
  * to itself so it may also be collected.
  */
-public final class Ref<T> implements RefCounted<T>
+public class Ref<T> implements RefCounted<T>
```
ekaterinadimitrova2: I am worried about having this public class not final; I think I mentioned it already in one of the commits yesterday, but adding here for completeness.

test/anttasks/org/apache/cassandra/anttasks/TestNameCheckTask.java
```diff
@@ -45,6 +45,11 @@ public TestNameCheckTask()
     {
     }
 
+    public static void main(String[] args)
+    {
+        new TestNameCheckTask().execute();
```
ekaterinadimitrova2, line 50: This is incorrect, as we no longer fail the ant task. 
It completes successfully if I test with an invalid test name; there is only one exception in the logs, which people will miss

ekaterinadimitrova2: I did some testing, and the issue with the ant task is only with 11. So as a workaround, we can make the task to run at the moment only with JDK11.
While it is good to run it with both JDKs, the output is the same with both JDKs.
So I suggest we open a follow-up ticket and make it run only with JDK11 here to unblock ourselves.

ekaterinadimitrova2:
I still have to dig into the GaloisCaounterMode but I wanted to give you first review comments:
1) tested checkstyle and it seems it does not work as expected anymore. I suggested a workaround in comments
2) I am not sure it is a good idea to make Ref not final anymore
3) Did some archeology and talked to David Capwell about:
`However, then I found that SyncUtil.force uses MemoryUtil.getAttachment and expects that the attachment may be Runnable. I'm not sure yet how that gets set though - I didn't any other setAttachment calls with Runnable except in a unit test. `
That check for instance of Runnable was added in CASSANDRA-18485. There is a PR attached to that ticket that led me to this comment https://github.com/apache/cassandra/pull/2299#discussion_r1179737237
So I confirmed with David that this is used only for tests or in his own words:
"this is only done for the ListenableFileSystem which needs to override sync
so anything using CQLTester.InMemory should be hitting this code path
so if those tests are passing... then prob fine"
It seems he added comment in the PR and forgot to move it to the code itself. We agreed we can add something as part of this ticket so people do not have to do archeology again :-)
4) I also ran CI and I did not find any new failures https://app.circleci.com/pipelines/github/ekaterinadimitrova2/cassandra/2378/workflows/cfb2f852-1a68-4a7f-8b61-c49efc5403f6

More next week

src/java/org/apache/cassandra/utils/concurrent/Ref.java
djatnieks: Okay, I could change `DirectBufferRef` to wrap a `Ref` instead of extending `Ref` - then `Ref` can remain `final`.

test/anttasks/org/apache/cassandra/anttasks/TestNameCheckTask.java
djatnieks: Thank you! I made a test class with an invalid name and see that while there is an error logged, the build continues.

I think I found another option - adding `failonerror="yes"` to the java task, like this:
```
    <java taskname="check-test-names" fork="true" failonerror="yes" classname="org.apache.cassandra.anttasks.TestNameCheckTask" classpath="${test.classes}">
```

With this change, running `ant build-test` with an invalid test name both logs an error and stops the build with an error.

build.xml
```diff
@@ -1853,7 +1866,10 @@
       <echo file=".idea/compiler.xml"><![CDATA[<?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
   <component name="JavacSettings">
-    <option name="ADDITIONAL_OPTIONS_STRING" value="--add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED --add-exports java.rmi/sun.rmi.transport.tcp=ALL-UNNAMED" />
+    <option name="ADDITIONAL_OPTIONS_STRING" value="--add-exports java.rmi/sun.rmi.registry=ALL-UNNAMED
+                                                    --add-exports java.rmi/sun.rmi.transport.tcp=ALL-UNNAMED
+                                                    --add-exports java.base/jdk.internal.ref=ALL-UNNAMED
+                                                    --add-exports java.base/sun.nio.ch=ALL-UNNAMED" />
```
ekaterinadimitrova2, line 1872: It seems John Meredith added two weeks ago JDK17 target for IntelliJ :D 
We need to add these exports there too.

ekaterinadimitrova2: @djatnieks I just pushed commit to fix this. Tested locally

ekaterinadimitrova2:
Committed

PR #2427, opened 2023-06-20
Extend grammar to accept the improved syntax.

jacek-lewandowski:


src/java/org/apache/cassandra/service/accord/txn/TxnMultiUpdate.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public Write apply(Timestamp executeAt, @Nullable Data data)
+    {
+        for (int i = 0; i < updates.length; i++)
+        {
+            TxnUpdate update = updates[i];
+            if (update.checkCondition(data))
+            {
+                TxnData txnData = data != null ? ((TxnData) data) : new TxnData();
+                txnData.setSelectedBranch(i);
```
dcapwell, line 67: this shouldn't be around outside of this local processing, so doesn't look to help?

src/java/org/apache/cassandra/service/accord/txn/TxnUpdate.java
```diff
@@ -57,7 +56,7 @@
 import static org.apache.cassandra.utils.ByteBufferUtil.serializedSizeWithVIntLength;
 import static org.apache.cassandra.utils.ByteBufferUtil.writeWithVIntLength;
 
-public class TxnUpdate implements Update
```
dcapwell: why?  when we don't have multiple conditions, we can safely just use this class

test/unit/org/apache/cassandra/cql3/statements/TransactionStatementTest.java
```diff
@@ -359,12 +374,218 @@ public void shouldRejectLetSelectWithIncompletePartitionKey()
...
+                       "  END IF\n" +
+                       "COMMIT TRANSACTION";
+
+        Assertions.assertThatThrownBy(() -> prepare(query))
+                  .isInstanceOf(SyntaxException.class)
+                  .hasMessageContaining("mismatched input 'SELECT' expecting K_END");
+    }
+
+    @Test
+    public void shouldRejectMissingSelectInBranch()
```
dcapwell: why is this an error? feels weird for me, would expect empty result set in this case rather than an error

jacek-lewandowski: hmm... you are right, there can be an empty result; this will simplify things a bit

src/java/org/apache/cassandra/service/accord/txn/TxnUpdate.java
jacek-lewandowski: I made this change temporarily to make sure `TxnUpdate` is not used anywhere. Anyway, why would we want to use it? I suppose it could complicate the logic as we would have to deal with single and multiple conditions separately

src/java/org/apache/cassandra/service/accord/txn/TxnMultiUpdate.java
jacek-lewandowski: Could you explain a bit more?

PR #2433, opened 2023-06-21
CASSANDRA-18615 Add support for USING and index creation by name with CREATE INDEX

maedhroz:
patch by Caleb Rackliffe; reviewed by ? for CASSANDRA-18615

### DIFF
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -1956,6 +1956,15 @@ drop_compact_storage_enabled: false
 # if zero_ttl_on_twcs_enabled is set to false, this property is irrelevant as such statements will fail.
 #zero_ttl_on_twcs_warned: true
 
+# The default secondary index implementation when CREATE INDEX does not specify one via USING.
+# ex. "legacy_local_table" - (default) legacy secondary index, implemented as a hidden table
+# ex. "sai" - "storage-attched" index, implemented via optimized SSTable/Memtable-attached indexes
+#default_secondary_index: legacy_local_table
+
+# Whether a default secondary index implementation is allowed. If this is "false", CREATE INDEX must
+# specify an index implementation via USING.
+#default_secondary_index_enabled: true
+
 # Startup Checks are executed as part of Cassandra startup process, not all of them
 # are configurable (so you can disable them) but these which are enumerated bellow.
 # Uncomment the startup checks and configure them appropriately to cover your needs.
--- a/doc/cql3/CQL.textile
+++ b/doc/cql3/CQL.textile
@@ -513,13 +513,23 @@ __Sample:__
 bc(sample). 
 CREATE INDEX userIndex ON NerdMovies (user);
 CREATE INDEX ON Mutants (abilityId);
-CREATE INDEX ON users (keys(favs));
+CREATE INDEX ON users (KEYS(favs));
+CREATE INDEX ON users (age) USING 'sai';
 CREATE CUSTOM INDEX ON users (email) USING 'path.to.the.IndexClass';
 CREATE CUSTOM INDEX ON users (email) USING 'path.to.the.IndexClass' WITH OPTIONS = {'storage': '/mnt/ssd/indexes/'};
 
 The @CREATE INDEX@ statement is used to create a new (automatic) secondary index for a given (existing) column in a given table. A name for the index itself can be specified before the @ON@ keyword, if desired. If data already exists for the column, it will be indexed asynchronously. After the index is created, new data for the column is indexed automatically at insertion time.
 
 Attempting to create an already existing index will return an error unless the @IF NOT EXISTS@ option is used. If it is used, the statement will be a no-op if the index already exists.
+
+h4(#usingIndex). Index Types
+
+The @USING@ keyword optionally specifies an index type. There are two built-in types:
+
+* legacy_local_table - (default) legacy secondary index, implemented as a hidden local table
+* sai - "storage-attched" index, implemented via optimized SSTable/Memtable-attached indexes
+
+To create a custom index, a fully qualified class name must be specified.
 
 h4(#keysIndex). Indexes on Map Keys
 
--- a/doc/modules/cassandra/examples/CQL/create_index.cql
+++ b/doc/modules/cassandra/examples/CQL/create_index.cql
@@ -1,6 +1,7 @@
 CREATE INDEX userIndex ON NerdMovies (user);
 CREATE INDEX ON Mutants (abilityId);
-CREATE INDEX ON users (keys(favs));
+CREATE INDEX ON users (KEYS(favs));
+CREATE INDEX ON users (age) USING 'sai';
 CREATE CUSTOM INDEX ON users (email) 
    USING 'path.to.the.IndexClass';
 CREATE CUSTOM INDEX ON users (email) 
--- a/src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
+++ b/src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
@@ -35,6 +35,7 @@ import org.apache.cassandra.db.Keyspace;
 import org.apache.cassandra.db.guardrails.Guardrails;
 import org.apache.cassandra.db.marshal.MapType;
 import org.apache.cassandra.exceptions.InvalidRequestException;
+import org.apache.cassandra.index.internal.CassandraIndex;
 import org.apache.cassandra.index.sasi.SASIIndex;
 import org.apache.cassandra.schema.*;
 import org.apache.cassandra.schema.Keyspaces.KeyspacesDiff;
@@ -77,6 +78,7 @@ public final class CreateIndexStatement extends AlterSchemaStatement
     public static final String INDEX_DUPLICATE_OF_EXISTING = "Index %s is a duplicate of existing index %s";
     public static final String KEYSPACE_DOES_NOT_MATCH_TABLE = "Keyspace name '%s' doesn't match table name '%s'";
     public static final String KEYSPACE_DOES_NOT_MATCH_INDEX = "Keyspace name '%s' doesn't match index name '%s'";
+    public static final String MUST_SPECIFY_INDEX_IMPLEMENTATION = "Must specify index implementation via USING";
 
     private final String indexName;
     private final String tableName;
@@ -316,6 +318,27 @@ public final class CreateIndexStatement extends AlterSchemaStatement
 
             if (indexName.hasKeyspace() && !keyspaceName.equals(indexName.getKeyspace()))
                 throw ire(KEYSPACE_DOES_NOT_MATCH_INDEX, keyspaceName, tableName);
+            
+            // Set the configured default 2i implementation if one isn't specified with USING:
+            if (attrs.customClass == null)
+            {
+                if (DatabaseDescriptor.getDefaultSecondaryIndexEnabled())
+                    attrs.customClass = DatabaseDescriptor.getDefaultSecondaryIndex();
+                else
+                    // However, operators may require an implementation be specified
+                    throw ire(MUST_SPECIFY_INDEX_IMPLEMENTATION);
+            }
+            
+            // If we explicitly specify the index type "cassandra", we can just clear the custom class, and the
+            // non-custom 2i creation process will begin. Otherwise, if an index type has been specified with 
+            // USING, make sure the appropriate custom index is created.
+            if (attrs.customClass != null)
+            {
+                if (!attrs.isCustom && attrs.customClass.equalsIgnoreCase(CassandraIndex.NAME))
+                    attrs.customClass = null;
+                else
+                    attrs.isCustom = true;
+            }
 
             return new CreateIndexStatement(keyspaceName, tableName.getName(), indexName.getName(), rawIndexTargets, attrs, ifNotExists);
         }
--- a/src/java/org/apache/cassandra/index/internal/CassandraIndex.java
+++ b/src/java/org/apache/cassandra/index/internal/CassandraIndex.java
@@ -69,6 +69,8 @@ import static org.apache.cassandra.cql3.statements.RequestValidations.checkFalse
  */
 public abstract class CassandraIndex implements Index
 {
+    public static final String NAME = "legacy_local_table";
+    
     private static final Logger logger = LoggerFactory.getLogger(CassandraIndex.class);
 
     public final ColumnFamilyStore baseCfs;
--- a/src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
+++ b/src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
@@ -95,6 +95,8 @@ import org.apache.cassandra.utils.concurrent.OpOrder;
 
 public class StorageAttachedIndex implements Index
 {
+    public static final String NAME = "sai";
+    
     private static final Logger logger = LoggerFactory.getLogger(StorageAttachedIndex.class);
 
     private static class StorageAttachedIndexBuildingSupport implements IndexBuildingSupport
--- a/src/java/org/apache/cassandra/schema/IndexMetadata.java
+++ b/src/java/org/apache/cassandra/schema/IndexMetadata.java
@@ -67,7 +67,8 @@ public final class IndexMetadata
 
     static
     {
-        indexNameAliases.put(StorageAttachedIndex.class.getSimpleName(), StorageAttachedIndex.class.getCanonicalName());
+        indexNameAliases.put(StorageAttachedIndex.NAME, StorageAttachedIndex.class.getCanonicalName());
+        indexNameAliases.put(StorageAttachedIndex.class.getSimpleName().toLowerCase(), StorageAttachedIndex.class.getCanonicalName());
         indexNameAliases.put(SASIIndex.class.getSimpleName(), SASIIndex.class.getCanonicalName());
     }
 
@@ -154,7 +155,7 @@ public final class IndexMetadata
         if (isCustom())
         {
             String className = options.get(IndexTarget.CUSTOM_INDEX_OPTION_NAME);
-            return indexNameAliases.getOrDefault(className, className);
+            return indexNameAliases.getOrDefault(className.toLowerCase(), className);
         }
         return CassandraIndex.class.getName();
     }
--- a/test/unit/org/apache/cassandra/cql3/validation/entities/SecondaryIndexTest.java
+++ b/test/unit/org/apache/cassandra/cql3/validation/entities/SecondaryIndexTest.java
@@ -25,9 +25,12 @@ import java.util.concurrent.CountDownLatch;
 import com.google.common.collect.ImmutableSet;
 
 import org.apache.commons.lang3.StringUtils;
+import org.junit.BeforeClass;
 import org.junit.Test;
 
 import org.apache.cassandra.exceptions.ConfigurationException;
+import org.apache.cassandra.index.internal.CassandraIndex;
+import org.apache.cassandra.index.sai.StorageAttachedIndex;
 import org.apache.cassandra.schema.ColumnMetadata;
 import org.apache.cassandra.schema.TableMetadata;
 import org.apache.cassandra.config.DatabaseDescriptor;
@@ -70,6 +73,12 @@ import static org.junit.Assert.fail;
 public class SecondaryIndexTest extends CQLTester
 {
     public static final int TOO_BIG = 1024 * 65;
+    
+    @BeforeClass
+    public static void setDefaultSecondaryIndex()
+    {
+        DatabaseDescriptor.setDefaultSecondaryIndex(CassandraIndex.NAME);
+    }
 
     @Test
     public void testCreateAndDropIndex() throws Throwable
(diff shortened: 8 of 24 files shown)


maedhroz:
https://app.circleci.com/pipelines/github/maedhroz/cassandra?branch=CASSANDRA-18615

src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
```diff
@@ -316,6 +318,23 @@ public CreateIndexStatement prepare(ClientState state)
 
             if (indexName.hasKeyspace() && !keyspaceName.equals(indexName.getKeyspace()))
                 throw ire(KEYSPACE_DOES_NOT_MATCH_INDEX, keyspaceName, tableName);
+            
+            // Set the configured default 2i implementation if one isn't specified with USING:
+            if (attrs.customClass == null)
+                if (DatabaseDescriptor.getDefaultSecondaryIndexEnabled())
```
maedhroz, line 325: The usage here is why I didn't end up making this a guardrail.

src/java/org/apache/cassandra/index/internal/CassandraIndex.java
```diff
@@ -68,6 +68,8 @@
  */
 public abstract class CassandraIndex implements Index
 {
+    public static final String NAME = "cassandra";
```
maedhroz: The name "cassandra" here isn't perfect I suppose, but it's trying to a.) align w/ the class name at least somewhat and b.) connote that this is the "traditional cassandra hidden table" index.

Maxwell-Guo: This is the point I most want to comment on this patch~~~ Because  cassandra vs SAI cannot be aligned conceptually in my mind.  Even native2i is better than cassandra

maedhroz: I’d really like to call it “legacy” to reflect my belief that we will eventually treat SAI as the default 2i implementation.

Maxwell-Guo: legacy is also one of the potential options I think

conf/cassandra.yaml
```diff
@@ -1956,6 +1956,15 @@ drop_compact_storage_enabled: false
 # if zero_ttl_on_twcs_enabled is set to false, this property is irrelevant as such statements will fail.
 #zero_ttl_on_twcs_warned: true
 
+# The default secondary index implementation when CREATE INDEX does not specify one via USING.
+# ex. "cassandra" - (default) traditional secondary index, implemented as a hidden table attached to the primary table
+# ex. "sai" - storage-attched index, implemented via optimized SSTable/Memtable-attached indexes
```
Maxwell-Guo: as for native secondary index ,we may use "cassandra" or some other keyword to represent the default index, then What about indexes other than sai? like sasi (if sasi is not fully removed when sai is released) and some plugin index using CUSTOM keyword like what I have said [lucene-index](https://github.com/Stratio/cassandra-lucene-index) (though we know that this plugin is not update now, but what  If someone what to add a plugin index by himself ? as for version before this patch ,he may use the interface defined by CUSTOM , but it seems this patch do not support this feature if someone want to develop his own index-plugin through CUSTOM or there is no description for this kind of index)

src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
```diff
@@ -316,6 +318,23 @@ public CreateIndexStatement prepare(ClientState state)
...
+                    attrs.customClass = DatabaseDescriptor.getDefaultSecondaryIndex();
+                else
+                    // However, operators may require an implementation be specified
+                    throw ire(MUST_SPECIFY_INDEX_IMPLEMENTATION);
+            
+            // If we explicitly specify the index type "cassandra", we can just clear the custom class, and the
+            // non-custom 2i creation process will begin. Otherwise, if an index type has been specified with 
+            // USING, make sure the appropriate custom index is created.
+            if (attrs.customClass != null)
+                if (!attrs.isCustom && attrs.customClass.equalsIgnoreCase(CassandraIndex.NAME))
```
Maxwell-Guo, line 337: Wouldn't it be more appropriate for this equation to become  
`!attrs.isCustom && attrs.customClass.equalsIgnoreCase(DatabaseDescriptor.getDefaultSecondaryIndex()) `
？

src/java/org/apache/cassandra/schema/IndexMetadata.java
```diff
@@ -67,7 +67,8 @@ public final class IndexMetadata
 
     static
     {
-        indexNameAliases.put(StorageAttachedIndex.class.getSimpleName(), StorageAttachedIndex.class.getCanonicalName());
+        indexNameAliases.put(StorageAttachedIndex.NAME, StorageAttachedIndex.class.getCanonicalName());
```
Maxwell-Guo, line 70: should we change keyword from "sai" to "StorageAttachedIndex" ? as I think StorageAttachedIndex is more expressive than sai to express the situation of indexing clearly. 

Sorry, I may be missing some final discussion process for the content of DISCUSS, so I will finally express the idea of ​​​​replacing the sai keyword.

test/unit/org/apache/cassandra/index/sai/cql/CollectionIndexingTest.java
```diff
@@ -220,4 +222,9 @@ private void assertUnsupportedIndexOperator(int expectedSize, String query, Obje
         assertInvalidMessage(StatementRestrictions.REQUIRES_ALLOW_FILTERING_MESSAGE, query, values);
         assertEquals(expectedSize, execute(query + " ALLOW FILTERING").size());
     }
+
+    private static String createIndexDDL(String target)
```
Maxwell-Guo, line 226: I think rename createIndexDDL to createSAIIndexDDL is more expressive

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
```diff
@@ -170,7 +174,7 @@ public void shouldFailCreationUsingMode()
     {
         createTable("CREATE TABLE %s (id text PRIMARY KEY, val text)");
 
-        assertThatThrownBy(() -> executeNet("CREATE CUSTOM INDEX ON %s(val) USING " +
+        assertThatThrownBy(() -> executeNet("CREATE INDEX ON %s(val) USING " +
```
Maxwell-Guo: so StorageAttachedIndex and sai are both  valid, so the document should be update, I think.

Maxwell-Guo: Besides I think the purpose of your test should be to test "USING 'sai' " ?

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
```diff
@@ -285,11 +335,10 @@ public void shouldCreateIndexOnReversedType() throws Throwable
     }
 
     @Test
-    public void shouldCreateIndexWithAlias()
+    public void shouldCreateIndexWithFullClassName()
     {
         createTable("CREATE TABLE %s (id text PRIMARY KEY, val text)");
-
-        createIndex("CREATE CUSTOM INDEX ON %s(val) USING 'StorageAttachedIndex'");
+        createIndex("CREATE CUSTOM INDEX ON %s(val) USING 'org.apache.cassandra.index.sai.StorageAttachedIndex'");
```
Maxwell-Guo, line 445: USING org.apache.cassandra.index.sai.StorageAttachedIndex also passed the test ,So the user document should be updated.

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
```diff
@@ -303,7 +352,7 @@ public void shouldCreateSASI() throws Throwable
     {
         createTable(CREATE_TABLE_TEMPLATE);
 
-        createIndex("CREATE CUSTOM INDEX ON %s(v1) USING 'org.apache.cassandra.index.sasi.SASIIndex' WITH OPTIONS = {'mode': 'CONTAINS',\n" +
+        createIndex("CREATE INDEX ON %s(v1) USING 'org.apache.cassandra.index.sasi.SASIIndex' WITH OPTIONS = {'mode': 'CONTAINS',\n" +
```
Maxwell-Guo: as I have comment [does sasi will fully removed after sai release](https://issues.apache.org/jira/browse/CASSANDRA-18615) , so it seems SASI is supported too, but as a new user , 
When I saw the description in this [yaml file](https://github.com/apache/cassandra/pull/2433/files#diff-77707d0908c31940828b6425dcb09a7409827db99b48c371f71c63294dfe1562) , I actually didn't know what the usage of native 2i, sai, sasi, and custom index plugin that could be used before was changed after the update of this patch.
So we may need more detailed documentation and add some CHANGE.txt that can clearly tell users about the changes, and how to configure the old functions on the new version。

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
```diff
@@ -94,6 +94,8 @@
 
 public class StorageAttachedIndex implements Index
 {
+    public static final String NAME = "sai";
```
Maxwell-Guo, line 98: what about add a new method named getIndexName() in the interface java Index.java , and every implementation index implement the method and return the real name ?

Maxwell-Guo: But IndexMetada seems to use the name to build indexNameAliases ....

test/unit/org/apache/cassandra/cql3/validation/entities/SecondaryIndexTest.java
```diff
@@ -70,6 +72,12 @@
 public class SecondaryIndexTest extends CQLTester
 {
     public static final int TOO_BIG = 1024 * 65;
+    
+    @BeforeClass
+    public static void setDefaultSecondaryIndex()
+    {
+        DatabaseDescriptor.setDefaultSecondaryIndex(CassandraIndex.NAME);
```
Maxwell-Guo, line 80: this test do not cover that when we do not set defaultSecondaryIndex.

conf/cassandra.yaml
maedhroz: There's nothing stopping anyone from developing custom 2i implementations as they always have and creating them using the `CREATE CUSTOM INDEX` DDL. We could always build a service provider framework or something here to register the index name aliases for custom 2i implementations and support them in CREATE INDEX...but not in this patch.

src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
maedhroz: No. if the default is "sai", we need the value of `customClass` (which is "sai") to be retained, not discarded. This is to specifically reflect the fact that the "cassandra" or "legacy" type doesn't need a `customClass`.

src/java/org/apache/cassandra/schema/IndexMetadata.java
maedhroz: You can also use "StorageAttachedIndex" (see the line below).

test/unit/org/apache/cassandra/index/sai/cql/CollectionIndexingTest.java
maedhroz: It's an SAI test, but perhaps.

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
maedhroz: "StorageAttachedIndex" support could be removed entirely, but it doesn't hurt to keep it. I'm kind of fine either way. CC @adelapena @mike-tr-adamson

maedhroz: The purpose of the test is to make sure we reject invalid `WITH OPTIONS` content, but I'll probably change it to `USING 'sai'`. I just missed it.

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
maedhroz: We don't want users to use this for SAI. It's only left here for legacy support. I wouldn't object to a client warning of some kind though.

test/unit/org/apache/cassandra/cql3/validation/entities/SecondaryIndexTest.java
maedhroz: There were two options to make sure this test works as intended independent of what the YAML says. One is this, just making sure the legacy 2i is the default, and none of the tests themselves need to change their index creation DDL. The other would be changing all the index creation DDL. I chose the former to keep the diff smaller.

`shouldCreateCassandraIndexExplicitly` though...that could change, or a new test be added to check explicit legacy 2i creation when it isn't the default.

maedhroz: Added a new case... `shouldCreateCassandraIndexWhenNotDefault`

conf/cassandra.yaml
maedhroz: The scope of this patch isn't really meant to extend to the `CREATE CUSTOM INDEX` DDL or its documentation. Whether we want to deprecate it is another question...

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
maedhroz: See https://github.com/apache/cassandra/pull/2433#discussion_r1242403756

> We could always build a service provider framework or something here to register the index name aliases for custom 2i implementations and support them in CREATE INDEX...but not in this patch.

The other complication here is that `CassandraIndex` is an abstract class. It's subclasses were never meant to be exposed to the user, which makes the SPI thing a little more sticky.

src/java/org/apache/cassandra/index/internal/CassandraIndex.java
maedhroz: Posted an informal poll in ASF Slack's `#cassandra-sai` chat

conf/cassandra.yaml
Maxwell-Guo: Thank you for explaining @maedhroz , so what about add a corresponding description to this document? This way, users may understand that the old usage(CREATE CUSTOM INDEX) is still possible.

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
Maxwell-Guo: Yes, if we suggest users to use a certain method, at least a document description or some warning is required. In my experience, if not, users either cannot use it or use it blindly.

In the end, this experience was terrible

Maxwell-Guo: As far as I am concerned, I have participated  in a little discussion  with the previous DISSCUSS and had a look at this patch  once before, so I have a rough idea of which usage is feasible. For first-time users, the process may not be as smooth.

conf/cassandra.yaml
maedhroz: My only concern w/ this patch was not to break the `CREATE CUSTOM INDEX` DDL that *happens* to still work as a result of how support for custom indexes was implemented. Even if it's possible, we don't want to document it or encourage people to use it. To be honest, I think the `CREATE CUSTOM INDEX` syntax was probably a mistake.

Maxwell-Guo: At least, "this usage will not affect previous usage" is needed though you said you don't want to document it ?
Update the document do not means we encourage people to use it,  just let them know more clear about the change. Then they can know how to use it clearly.

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
maedhroz: Once we settle on a name for the legacy 2i and update that in the YAML and `Config`, it would be useful to update `CQL.textile` and `indexes.adoc` (which is the source for the OSS website docs) to reflect the recommended usage. I'll have another commit that includes that shortly...

conf/cassandra.yaml
maedhroz: There is no change. SAI hasn't been released yet. There is no example of an SAI index being created w/ that syntax in the wild.

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
Maxwell-Guo: ok~~thanks

maedhroz: Pushed. Even this is just to illustrate basic supported usage. The SAI `README.md` in `o.a.c.index.sai` is probably a better in-tree overview of what SAI actually is.

test/unit/org/apache/cassandra/index/sai/cql/CollectionIndexingTest.java
Maxwell-Guo: 1. `CREATE CUSTOM INDEX ON %s(VALUES(value)) USING 'StorageAttachedIndex'" `
2. `createIndexDDL(String target)` 


1 is more intuitive than 2, as I know I am creating SAI when I see StorageAttachedIndex, but I have no idea about what kind of index I am going to create when see "createIndexDDL".

Of course, you don’t need to change it if you do not want to, because this is just a  very small point, which is insignificant

doc/cql3/CQL.textile
```diff
@@ -511,14 +511,24 @@ __Sample:__
...
 Attempting to create an already existing index will return an error unless the @IF NOT EXISTS@ option is used. If it is used, the statement will be a no-op if the index already exists.
 
+h4(#usingIndex). Index Types
+
+The @USING@ keyword optionally specifies an index type. There are two built-in types:
+
+* legacy - (default) legacy secondary index, implemented as a hidden table
+* sai - "storage-attched" index, implemented via optimized SSTable/Memtable-attached indexes
+
+To create a custom index, a fully qualified class name must be specified.
```
Maxwell-Guo, line 532: thanks, I think this is enough.

doc/modules/cassandra/examples/CQL/create_index.cql
```diff
@@ -1,6 +1,7 @@
 CREATE INDEX userIndex ON NerdMovies (user);
 CREATE INDEX ON Mutants (abilityId);
-CREATE INDEX ON users (keys(favs));
+CREATE INDEX ON users (KEYS(favs));
+CREATE INDEX ON users (age) USING 'sai';
```
adelapena, line 4: I think we need to add the doc changes in `cql_singlefile.adoc` too. That file seems to be an aggregation of the individual files in the same directory.

src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
```diff
@@ -316,6 +318,23 @@ public CreateIndexStatement prepare(ClientState state)
...
+                if (DatabaseDescriptor.getDefaultSecondaryIndexEnabled())
+                    attrs.customClass = DatabaseDescriptor.getDefaultSecondaryIndex();
+                else
+                    // However, operators may require an implementation be specified
+                    throw ire(MUST_SPECIFY_INDEX_IMPLEMENTATION);
+            
+            // If we explicitly specify the index type "cassandra", we can just clear the custom class, and the
+            // non-custom 2i creation process will begin. Otherwise, if an index type has been specified with 
+            // USING, make sure the appropriate custom index is created.
+            if (attrs.customClass != null)
```
adelapena, line 335: Nit: I'd add braces to the outer if

src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
```diff
@@ -316,6 +318,23 @@ public CreateIndexStatement prepare(ClientState state)
 
             if (indexName.hasKeyspace() && !keyspaceName.equals(indexName.getKeyspace()))
                 throw ire(KEYSPACE_DOES_NOT_MATCH_INDEX, keyspaceName, tableName);
+            
+            // Set the configured default 2i implementation if one isn't specified with USING:
+            if (attrs.customClass == null)
```
adelapena, line 323: Nit: I'd add braces to the outer if

test/unit/org/apache/cassandra/cql3/validation/entities/SecondaryIndexTest.java
```diff
@@ -161,6 +170,37 @@ private static String removeQuotes(String indexName)
         return StringUtils.remove(indexName, '\"');
     }
 
+    @Test
+    public void shouldCreateCassandraIndexExplicitly() throws Throwable
+    {
+        createTable("CREATE TABLE %s (userid uuid PRIMARY KEY, firstname text, lastname text, age int)");
+        createIndex("CREATE INDEX byAge ON %s(age) USING 'legacy'");
```
adelapena: The test doesn't verify that the proper index implementation is selected. I mean, the test would pass if a SAI index were created. I think we can do that check with something like:
```java
SecondaryIndexManager indexManager = getCurrentColumnFamilyStore().indexManager;
Index index = indexManager.getIndexByName("byage");
assertTrue(index instanceof CassandraIndex);
```

test/unit/org/apache/cassandra/cql3/validation/entities/SecondaryIndexTest.java
```diff
@@ -161,6 +170,37 @@ private static String removeQuotes(String indexName)
...
+
+    @Test
+    public void shouldCreateCassandraIndexWhenNotDefault() throws Throwable
+    {
+        DatabaseDescriptor.setDefaultSecondaryIndex(StorageAttachedIndex.NAME);
+        
+        try
+        {
+            createTable("CREATE TABLE %s (userid uuid PRIMARY KEY, firstname text, lastname text, age int)");
+            createIndex("CREATE INDEX byAge ON %s(age) USING 'legacy'");
```
adelapena: As before, we probably should verify that a legacy index is created.

test/unit/org/apache/cassandra/index/sai/cql/CollectionIndexingTest.java
adelapena: I think I prefer `createIdexDDL` because we are inside the SAI package. We have lots of mentions to "index" in that package and IMO it won't be practical to prefix all of them with "SAI" given the context.

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
```diff
@@ -255,20 +257,66 @@ public void shouldFailCreateWithInvalidCharactersInColumnName()
...
 
+    @Test
+    public void shouldFailWithDefaultIndexDisabled()
+    {
+        DatabaseDescriptor.setDefaultSecondaryIndex(StorageAttachedIndex.NAME);
+        boolean original = DatabaseDescriptor.getDefaultSecondaryIndexEnabled();
+        
+        try
+        {
+
```
adelapena: Nit: unneeded blank line

src/java/org/apache/cassandra/index/internal/CassandraIndex.java
maedhroz: @adelapena chimed in with "legacy_local_table", which I like a lot. I'm thinking about proceeding w/ that.

Maxwell-Guo (approved):
+1 with green CI results

src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
```diff
@@ -316,6 +318,27 @@ public CreateIndexStatement prepare(ClientState state)
...
+            {
+                if (DatabaseDescriptor.getDefaultSecondaryIndexEnabled())
+                    attrs.customClass = DatabaseDescriptor.getDefaultSecondaryIndex();
+                else
+                    // However, operators may require an implementation be specified
+                    throw ire(MUST_SPECIFY_INDEX_IMPLEMENTATION);
+            }
+            
+            // If we explicitly specify the index type "cassandra", we can just clear the custom class, and the
```
adelapena, line 332:
```suggestion
            // If we explicitly specify the index type "legacy_local_table", we can just clear the custom class, and the
```

maedhroz: Good catch! I'll fix on commit...

maedhroz:
Committed as https://github.com/apache/cassandra/commit/4b43c75b00703c5863228331326ecc95d3fe1605

PR #2434, opened 2023-06-22
CASSANDRA-18618: Separate task for running static analysis

jacek-lewandowski:


jacek-lewandowski:
https://ci-cassandra.apache.org/view/patches/job/Cassandra-devbranch/2517/

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/744/workflows/45d20631-96a2-4116-9a92-14fadbaaddb8/jobs/15328 (see the "Run static code analysis step")

michaelsembwever:
can we please move the different check/list ant targets into a new .build/build-checks.xml files? (or separate .build/build-xxx.xml file)

it really helps readability of build.xml (encapsulating concerns).

.build/build-rat.xml
```diff
@@ -106,7 +106,7 @@
         </fail>
     </target>
 
-    <target name="_assert_rat_output">
+    <target name="_assert_rat_output" unless="${rat.skip}">
```
michaelsembwever:
```suggestion
    <target name="_assert_rat_output" unless="rat.skip">
```

?

michaelsembwever (approved):

.build/build-rat.xml
jacek-lewandowski: what do you mean? it is defined in the same way as in https://github.com/apache/cassandra/pull/2434/files#diff-d274784b2fecdbee879b6175c05a9860374c65c0c0b7e87efe8ec66c0fc3d838R50

michaelsembwever: that is a mistake.  it should just be the string literal.

.circleci/config.yml
```diff
@@ -4131,7 +4131,7 @@ jobs:
             git clean -fd
             # Loop to prevent failure due to maven-ant-tasks not downloading a jar..
             for x in $(seq 1 3); do
-                ${ANT_HOME}/bin/ant realclean; ${ANT_HOME}/bin/ant jar dtest-jar
+                ${ANT_HOME}/bin/ant realclean; ${ANT_HOME}/bin/ant jar dtest-jar -Dno-checkstyle=true -Dno-check=true
```
michaelsembwever: it shouldn't be necessary to supply both `-Dno-checkstyle=true -Dno-check=true`

i.e. `-Dno-check=true` should imply `-Dno-checkstyle=true -Drat.skip=true -Declipse-warnings.skip=true -Ddependency-check.skip=true`

here's a patch i was playing with before… 
https://github.com/apache/cassandra/compare/trunk...thelastpickle:cassandra:mck/buildfiles-linter-docs/4.1

( that patch also allows you to specify there shorter form of `-Dno-linter=` )

jacek-lewandowski: This is for older branches - are we going to apply those changes to those older branches as well?

michaelsembwever: erh, no. go nuts then.

jacek-lewandowski:
@michaelsembwever I've rebased and applied the changes according to what was concluded on the ML

.circleci/config.yml
```diff
@@ -4131,7 +4131,7 @@ jobs:
             git clean -fd
             # Loop to prevent failure due to maven-ant-tasks not downloading a jar..
             for x in $(seq 1 3); do
-                ${ANT_HOME}/bin/ant realclean; ${ANT_HOME}/bin/ant jar dtest-jar
+                ${ANT_HOME}/bin/ant realclean; ${ANT_HOME}/bin/ant jar dtest-jar -Dno-checkstyle=true -Drat.skip=true
```
michaelsembwever: are these still needed if `check` is now its own target …?

build.xml
```diff
@@ -540,6 +539,14 @@
         <copy todir="${basedir}/conf" file="${build.classes.main}/META-INF/hotspot_compiler"/>
     </target>
 
+    <target name="check" depends="_main-jar,build-test" description="Verifies the source code and dependencies">
+        <antcall target="rat-check" inheritrefs="true"/>
+        <antcall target="_assert_rat_output" inheritrefs="true"/>
+        <antcall target="checkstyle" inheritrefs="true"/>
+        <antcall target="checkstyle-test" inheritrefs="true"/>
+        <antcall target="eclipse-warnings" inheritrefs="true"/>
```
michaelsembwever, line 554: what about dependency-check ?

.circleci/config.yml
jacek-lewandowski: yes, because dtest jars are built for older Cassandra versions and we decided to not port this change to older branches

build.xml
jacek-lewandowski: We should have a dedicated ticket for that. Unless the project dependencies are changed, the developer cannot be blamed for CVEs. 

Also, if the dependency set is not changed, it does not make sense to download and analyze the project against the whole database because we can be affected only by CVEs which are discovered recently.

There are a couple of valuable things we can do about dependency checks, including license analysis, and I'd like to avoid mixing this PR with that.

.build/build-git.xml
```diff
@@ -51,7 +51,34 @@
...
+        </condition>
+
+        <exec if:true="${git.is-available}" executable="git" osfamily="unix" dir="${basedir}" logError="true" failonerror="false" failifexecutionfails="false">
+            <arg value="diff"/>
+            <arg value="--name-only"/>
+            <arg value="--no-renames"/>
+            <arg value="--ignore-all-space"/>
+            <arg value="--merge-base"/>
+            <arg value="${_ref_branch}"/>
+            <arg value="${build.helpers.dir}/cassandra-build-deps-template.xml"/>
```
jacek-lewandowski: actually we should not care about cassandra-build-deps

.build/build-owasp.xml
```diff
@@ -16,9 +16,11 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   -->
-<project basedir="." name="apache-cassandra-owasp-tasks">
+<project basedir="." name="apache-cassandra-owasp-tasks"
+         xmlns:if="ant:if"
+         xmlns:unless="ant:unless">
     <property name="dependency-check.version" value="8.3.1"/>
-    <property name="dependency-check.home" value="${build.dir}/dependency-check-ant-${dependency-check.version}"/>
+    <property name="dependency-check.home" value="${user.home}/.ant/dependency-check-ant-${dependency-check.version}"/>
```
jacek-lewandowski: I set it to user.home because downloading the database is painfully slow and it can be shared between projects / worktrees. Redownloading it after ant realclean was frustrating

.build/build-owasp.xml
```diff
@@ -83,4 +85,67 @@
...
+    <target depends="init" name="license-report">
+        <exec executable="jq" osfamily="unix" dir="${basedir}" logError="true" failonerror="false" failifexecutionfails="false">
+            <arg value="-M"/>
+            <arg value=".dependencies[] | .fileName + &quot;=&quot; +  .license"/>
+            <arg value="${build.dir}/owasp/dependency-check-report.json"/>
+            <redirector output="${build.dir}/owasp/license-report.txt"/>
+        </exec>
+
+        <loadfile property="wrongly-licensed" srcFile="${build.dir}/owasp/license-report.txt">
+            <filterchain>
```
jacek-lewandowski: Here we have defined patterns for allowed licenses

.build/build-owasp.xml
```diff
@@ -83,4 +85,67 @@
...
+                <linecontainsregexp negate="true">
+                    <regexp pattern="^.*=.*(ISC)"/>
+                </linecontainsregexp>
+                <replaceregex pattern="&quot;" replace="" byline="true" flags="g"/>
+                <trim/>
+                <sortfilter/>
+            </filterchain>
+        </loadfile>
+
+        <echo message="${wrongly-licensed}" output="${build.dir}/owasp/license-report.txt"/>
```
jacek-lewandowski: `build/owasp/license-report.txt` contains list of those jars whose licenses are unknown or are not approved (there have not been filtered out by the above regex filters). So this is the file the developer should look at when analysis fails.

.build/build-owasp.xml
```diff
@@ -83,4 +85,67 @@
...
+
+        <loadfile property="wrongly-licensed-comparable" srcFile="${build.dir}/owasp/license-report.txt">
+            <filterchain>
+                <replaceregex pattern="=.*$" replace="" byline="true" flags="g"/>
+                <striplinebreaks/>
+            </filterchain>
+        </loadfile>
+
+        <loadfile property="license-exceptions-comparable" srcfile="${build.helpers.dir}/license-exceptions.txt" failonerror="false">
```
jacek-lewandowski: In `.build/license-exceptions.txt` there is a list of jars we allow despite their license is unknown or not approved. In the end we compare the list of reported jars (see above) and the list of exceptions defined here. If those lists do not match, we fail the build.

.build/build-owasp.xml
michaelsembwever: is it specific to `ant` ?  

could it be under `${user.home}/.owasp/dependency-check-ant-${dependency-check.version}` instead?

how big does it get? should it go under `${user.home}/.cache/owasp/…` instead?  (mine was ~500M , but it will grow over time…)

@smiklosovic thoughts?

.build/build-git.xml
```diff
@@ -51,7 +51,34 @@
...
+            <arg value="${build.helpers.dir}/cassandra-deps-template.xml"/>
+            <arg value="${build.helpers.dir}/parent-pom-template.xml"/>
+            <redirector outputproperty="git.diff.deps-xmls"/>
+        </exec>
+        <filelist id="git.diff.deps-xmls.list" files="${git.diff.deps-xmls}" description="List of changed production sources between HEAD and the latest common ancestor of HEAD and trunk"/>
+        <pathconvert property="git.diff.deps-xmls.path" pathsep="," refid="git.diff.deps-xmls.list">
+        </pathconvert>
```
michaelsembwever: why? can this go into a private ant target with a descriptive name pls.

.build/build-owasp.xml
michaelsembwever: nice.

we should add a comment pointing to https://apache.org/legal/resolved

.build/build-owasp.xml
jacek-lewandowski: I'm fine with any location under user home. It takes ~250M currently

jacek-lewandowski: @michaelsembwever maybe there is some public server which hosts such database and we could connect to it instead (or at least as a first choice)?

smiklosovic: @michaelsembwever `~/.cache` seems to be a good dir to put it under. We are really caching it and there are other caches in place already on desktops. Maybe `~/.cache/cassandra-owasp` would be the best to say that such owasp cache is from Cassandra project. Otherwise who would know?

smiklosovic: @jacek-lewandowski what happens if `~/.cache` dir does not exist? Highly improbable but still ... does it create parent dirs too?

.build/build-owasp.xml
```diff
@@ -83,4 +85,67 @@
             </fileset>
         </dependency-check>
     </target>
+
+    <target depends="init" name="license-report">
+        <exec executable="jq" osfamily="unix" dir="${basedir}" logError="true" failonerror="false" failifexecutionfails="false">
```
smiklosovic: @jacek-lewandowski  @michaelsembwever  This is suspicious. So we need to have `jq` locally, correct? I think we need be very picky and explicit about this. Until now, all you need is git, ant, python and shell.

smiklosovic: we might have a check for this in `build.xml` as we have for Ant.

    <fail message="You need to use Ant of version at least 1.10 to continue.">
      <condition>
        <not>
          <antversion atleast="1.10"/>
        </not>
      </condition>
    </fail>

Even "antversion" is in-built target (I think so), something similar should be done for all other needed tooling and error out as soon as possible if it is not available so it will not fail unnecessarily somewhere down the road.

smiklosovic: I see there is `failifexecutionfails="false", could you explain the consequences? So we are OK when this fails? So if that command is not present, we will just silently skip it?

I think if it just prints some report from owasp, that might be acceptable and we do not need to do anything after all.

.build/build-git.xml
jacek-lewandowski: what exactly are you referring to?

.build/build-owasp.xml
jacek-lewandowski: yes, it creates the parent directories as well, I've set it to `~/.cache/dependency-check-ant/owasp-db`

.build/build-owasp.xml
jacek-lewandowski: ok, I realized the same and refactored it to run with JavaScript (no worries about j17). Now it seems to be better / cleaner from the user pov.

jacek-lewandowski:
@michaelsembwever @smiklosovic - I've refactored it, I agree `jq` may not be a common thing people have in their systems. I don't think it is a big problem - there are some prerequisites to work with the projects and this is not very problematic or uncommon tool.

However, I found implementing it with JavaScript is better. Now we have:
- `allowed-licenses.txt` which includes patterns to match the licenses we approve, it can contain line comments as well
- `license-overrides.txt` which explicitly sets the licenses for jars for which they couldn't be recognized automatically

`check` runs `dependency-check` only if dependencies were touched

I needed to explicitly include Nashorn because JDK-17 does not include it anymore. I know it is GPL but we do not ship it - it is included only in build dependencies, not the `lib` directory.

.build/build-owasp.xml
michaelsembwever: i agree with @smiklosovic , `~/.cache/cassandra-owasp-dependency-check` is a better name.

Users will want to know where a half gigabyte folder came from.

.build/build-git.xml
michaelsembwever: what does this addition of code/xml do? 

a private ant target with a descriptive name beats a comment.

.build/parent-pom-template.xml
```diff
@@ -1049,6 +1049,11 @@
         <artifactId>big-math</artifactId>
         <version>2.3.0</version>
       </dependency>
+      <dependency>
+        <groupId>org.openjdk.nashorn</groupId>
```
smiklosovic: @jacek-lewandowski  can we really just add it like that?

jacek-lewandowski: this is only the dependency management - we declare dependency versions here. The dependency in fact is defined in [build-resolver.xml](https://github.com/apache/cassandra/pull/2434/files#diff-b7d8cc9a23d28a1713d5242f83da30838851d0685528abdb937afe7d9b899de6R137-R146)

.build/build-resolver.xml
```diff
@@ -134,6 +134,16 @@
             </dependencies>
             <path refid="checkstyle.classpath" classpath="runtime"/>
         </resolve>
+        <resolve>
+<!-- Required for running JavaScript regardless of JDK version, this is only a build dependency, not inlcuded in the distribution -->
```
smiklosovic: @jacek-lewandowski could you aligned this? just above `<dependencies>`

jacek-lewandowski: will do

michaelsembwever (approved):

PR #2436, opened 2023-06-22
CASSANDRA-18613 trunk: Add support for vectors on UDFs

adelapena:


src/java/org/apache/cassandra/cql3/functions/types/CodecRegistry.java
```diff
@@ -847,6 +864,12 @@ private <T> TypeCodec<T> maybeCreateCodec(DataType cqlType, T value)
             }
         }
 
+        if ((cqlType == null || cqlType.getName() == VECTOR) && value instanceof List)
```
Maxwell-Guo, line 867: I know we have imported the DataType type class, so it is ok to set "cqlType.getName() == VECTOR"
I just have a little bit of cleanliness, I think it is better to align with the similar logic below, either remove their ”DataType.Name.“, or align with them. change to :

`
    
        if ((cqlType == null || cqlType.getName() == TUPLE)
            && value instanceof TupleValue)
        {
            return (TypeCodec<T>)
                   TypeCodec.tuple(cqlType == null ? ((TupleValue) value).getType() : (TupleType) cqlType);
        }

        if ((cqlType == null || cqlType.getName() == UDT) && value instanceof UDTValue)
        {
            return (TypeCodec<T>)
                   TypeCodec.userType(cqlType == null ? ((UDTValue) value).getType() : (UserType) cqlType);
        }
`

or `if ((cqlType == null || cqlType.getName() == DataType.Name.VECTOR) && value instanceof List)`

src/java/org/apache/cassandra/cql3/functions/types/VectorType.java
```diff
@@ -0,0 +1,84 @@
...
+    private final int dimensions;
+
+    VectorType(DataType subtype, int dimensions)
+    {
+        super(Name.VECTOR);
+        this.subtype = subtype;
+        this.dimensions = dimensions;
+    }
+
+    public DataType getSubtype() {
```
Maxwell-Guo: I think “{”  should start from a new line

src/java/org/apache/cassandra/cql3/functions/types/VectorType.java
```diff
@@ -0,0 +1,84 @@
...
+        return this.subtype;
+    }
+
+    public int getDimensions() {
```
Maxwell-Guo: I think “{”  should start from a new line

src/java/org/apache/cassandra/cql3/functions/types/VectorType.java
```diff
@@ -0,0 +1,84 @@
...
+    public int hashCode()
+    {
+        int result = name.hashCode();
+        result = 31 * result + dimensions;
+        result = 31 * result + subtype.hashCode();
+        return result;
+    }
+
+    @Override
+    public String toString() {
```
Maxwell-Guo: I think “{”  should start from a new line

src/java/org/apache/cassandra/cql3/functions/types/VectorType.java
```diff
@@ -0,0 +1,84 @@
...
+public class VectorType extends DataType
+{
+    private final DataType subtype;
+    private final int dimensions;
+
+    VectorType(DataType subtype, int dimensions)
+    {
+        super(Name.VECTOR);
+        this.subtype = subtype;
+        this.dimensions = dimensions;
```
Maxwell-Guo, line 36: Do we need to do some validation for dimensions here ? if some one set a dimensions that less than 0 ? Though I know org.apache.cassandra.db.marshal.VectorType 's construct function will do that .

src/java/org/apache/cassandra/cql3/functions/types/VectorCodec.java
```diff
@@ -0,0 +1,211 @@
...
+import com.google.common.base.Splitter;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterables;
+
+import org.apache.cassandra.cql3.functions.types.exceptions.InvalidTypeException;
+import org.apache.cassandra.transport.ProtocolVersion;
+import org.apache.cassandra.utils.vint.VIntCoding;
+
+/**
+ * {@link TypeCodec} for vectors. Vectors are represented as {@link List}s for convenience, since it's probably easier
```
Maxwell-Guo: "{@link TypeCodec} for vectors. Vectors"  , should be vector instead of vectors ?

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
```diff
@@ -248,25 +250,103 @@ public void token()
...
+                            "RETURNS vector<int, 4> " +
+                            "LANGUAGE java " +
+                            "AS '" +
+                            "if (x == null || y == null) return null;" +
+                            "List<Integer> l = new ArrayList<Integer>(x); " +
+                            "l.addAll(y); " +
+                            "return l;'");
+        assertRows(execute(format("SELECT %s(value, [3, 4]) FROM %%s", f)), row(vector(1, 2, 3, 4)));
+        assertRows(execute(format("SELECT %s([2, 3], value) FROM %%s", f)), row(vector(2, 3, 1, 2)));
+        assertRows(execute(format("SELECT %s(null, null) FROM %%s", f)), row((Vector<Integer>) null));
```
Maxwell-Guo, line 440: Should some exception cases be added?
like: vector<int, element1> (element1 is not number), vector<int, element2> (element is equal or less than 0).
I know this usage is not right.But what we want to see is if the wrong usage  in line with our expectations

src/java/org/apache/cassandra/cql3/functions/types/VectorCodec.java
```diff
@@ -0,0 +1,211 @@
...
+               : new VariableLength<>(type, subtypeCodec);
+    }
+
+    @Override
+    public List<E> parse(String value) throws InvalidTypeException
+    {
+        if (value == null || value.isEmpty() || value.equalsIgnoreCase("NULL")) return null;
+
+        ImmutableList.Builder<E> values = ImmutableList.builder();
+        for (String element : Splitter.on(", ").split(value.substring(1, value.length() - 1)))
```
Maxwell-Guo, line 66: Sure  to use  hardcode like this ", " ?  and  it seems that there must be a space after the comma?

src/java/org/apache/cassandra/cql3/functions/types/CodecRegistry.java
adelapena: Here `VECTOR` follows the style of the previous checks for `LIST`, `SET` and `MAP`. The differences start on the checks for `TUPLE` and `UDT`. I have changed the previous code to use the static import in all cases.

src/java/org/apache/cassandra/cql3/functions/types/VectorType.java
adelapena: +1, I missed that while copy-pasting the driver's code, which uses a different code style.

src/java/org/apache/cassandra/cql3/functions/types/VectorType.java
adelapena: Done

src/java/org/apache/cassandra/cql3/functions/types/VectorType.java
adelapena: Done

src/java/org/apache/cassandra/cql3/functions/types/VectorType.java
adelapena: Indeed this is checked before getting here. I have added a defensive assertion.

src/java/org/apache/cassandra/cql3/functions/types/VectorCodec.java
adelapena: I'd say that plural is correct here since this is a codec for vector instances (vectors). But I have rephrased it to refer to the (singular) vector type instead. So it's `{@link TypeCodec} for the CQL type {@code vector}`, which is the format used by other codec builders.

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
adelapena: I'd say those checks are more related to CQL parsing than to UDF execution in particular, but I have added a few tests for completeness.

src/java/org/apache/cassandra/cql3/functions/types/VectorCodec.java
adelapena: The driver's `parse` method complements the `format` method. The latter uses Guava's `Iterables#toString`, which also uses `", "`: https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Iterators.java#L292

Maxwell-Guo:
+1

src/java/org/apache/cassandra/cql3/functions/types/VectorCodec.java
Maxwell-Guo: ok

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
Maxwell-Guo: thanks

Maxwell-Guo (approved):

src/java/org/apache/cassandra/cql3/functions/types/GettableByIndexData.java
```diff
@@ -465,6 +465,21 @@ public interface GettableByIndexData
      */
     public <K, V> Map<K, V> getMap(int i, TypeToken<K> keysType, TypeToken<V> valuesType);
 
+    /**
+     * Returns the {@code i}th value as a vector.
+     *
+     * <p>This method uses the {@link CodecRegistry} to find a codec to convert the underlying CQL
+     * type to a {@link List} (if the CQL type is a UDT, the registry will generate a codec
```
bereng: You mention UDT here?

adelapena: I've wrongly copied that comment from the getter for UDTs. Revisiting this has made me realise that the getters/setters are incorrect for usage with nested types. I have just fixed them and added a new test case covering them.

bereng: Nice catch! I hadn't gone that deep yet. Glad you caught it :-)

src/java/org/apache/cassandra/cql3/functions/types/TypeCodec.java
```diff
@@ -488,6 +501,16 @@ public DataType getCqlType()
         return cqlType;
     }
 
+    /**
+     * Checks if all values are of fixed length.
+     *
+     * @return {@code true} if all values are of fixed length, {@code false} otherwise.
+     */
+    public boolean isValueLengthFixed()
```
bereng: I wonder if you could create Fixed/Variable classes to avoid all those isValueLengthFixed() methods

src/java/org/apache/cassandra/cql3/functions/types/VectorCodec.java
```diff
@@ -0,0 +1,211 @@
...
+    {
+        return subtypeCodec.isValueLengthFixed()
+               ? new FixedLength<>(type, subtypeCodec)
+               : new VariableLength<>(type, subtypeCodec);
+    }
+
+    @Override
+    public List<E> parse(String value) throws InvalidTypeException
+    {
+        if (value == null || value.isEmpty() || value.equalsIgnoreCase("NULL")) return null;
```
bereng: formatting?

src/java/org/apache/cassandra/cql3/functions/types/VectorCodec.java
```diff
@@ -0,0 +1,211 @@
...
+        @Override
+        public boolean isValueLengthFixed()
+        {
+            return true;
+        }
+
+        @Override
+        public ByteBuffer serialize(List<E> value, ProtocolVersion protocolVersion) throws InvalidTypeException
+        {
+            if (value == null || type.getDimensions() <= 0)
```
bereng, line 102: Given you know in advance the number of elements and their sizes can't you collapse both loops into a single one? Similar to what you do in deserialize?

src/java/org/apache/cassandra/cql3/functions/types/VectorCodec.java
```diff
@@ -0,0 +1,211 @@
...
+     */
+    private static class VariableLength<E> extends VectorCodec<E>
+    {
+        public VariableLength(VectorType type, TypeCodec<E> subtypeCodec)
+        {
+            super(type, subtypeCodec);
+        }
+
+        @Override
+        public ByteBuffer serialize(List<E> values, ProtocolVersion version) throws InvalidTypeException
```
bereng, line 157: I was trying to think of a way to encode all sizes upfront, but only the deltas so save some bytes. But we don't support negative VInts...

test/unit/org/apache/cassandra/cql3/validation/entities/UFSecurityTest.java
```diff
@@ -118,7 +118,13 @@ public void testSecurityPermissions() throws Throwable
                                     "     java.lang.Runtime.getRuntime().exec(\"/tmp/foo\"); return 0d;" +
                                     "} catch (Exception t) {" +
                                     "     throw new RuntimeException(t);" +
-                                    '}'}
+                                    '}'},
```
bereng, line 121: Formatting?

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
```diff
@@ -248,25 +250,155 @@ public void token()
...
+        assertInvalidThrowMessage("vectors may only have positive dimentions; given 0",
+                                  InvalidRequestException.class,
+                                  "CREATE FUNCTION %s (x vector<int, 2>) " +
+                                  "CALLED ON NULL INPUT " +
+                                  "RETURNS vector<int, 0> " +
+                                  "LANGUAGE java " +
+                                  "AS 'return x;'");
+
+        // function reading and writing a udt vector field
+        String udt = createType("CREATE TYPE %s (v vector<int,2>)");
```
bereng, line 476: There's some cleanup error on the test if you run it locally, presumably related to this. It'd be good to cleanup in case we ever need to run the test repeatedly.

test/unit/org/apache/cassandra/cql3/validation/entities/UFSecurityTest.java
adelapena: What do you want to do with formatting here? I have tried to follow the style of the test. Specifically, what is done for long class names like `org.apache.cassandra.service.StorageService` above.

bereng: It's all misaligned. The try-catch isn't indented to align with the ones above yes?

bereng: Oh sorry, ICWYM with the long class name now. Ok I guess that's it then, it can't be improved...

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
adelapena: The repeated runs of the test pass: https://app.circleci.com/pipelines/github/adelapena/cassandra/3010/workflows/cfcbe3c8-ec11-46db-96cd-a0b5ef9f6493/jobs/57489/steps

It seems there is some defect in how `CQLTester` does the cleaning up of functions. `CQLTester#afterTest` correctly drops first the function and then the type. But somehow the dropping of the type ends up happening after the dropping of the function. Indeed, if we add a short sleep after dropping the function the error doesn't happen. That's something worth investigating in a separate ticket. For this one, I have just made the test drop the function.

bereng: I think it's good you fixed it for the time being :-)

bereng:
@adelapena you're looking into the other things I mentioned? Asking just in case we're not both waiting on each other lol

src/java/org/apache/cassandra/cql3/functions/types/TypeCodec.java
adelapena: I think that would create a multiple inheritance problem. For example, `PrimitiveLongCodec` extends `TypeCodec` and has multiple subclasses with both fixed and variable encoding. Those subclasses can't inherit from both `PrimitiveLongCodec` and the proposed Fixed/Variable classes. 

Alternatively, we could make a separate interface just for the `isValueLengthFixed ` method and make `TypeCodec` implement it. That interface would be extended by two Fixed/Variable interfaces with default implementations, and the terminal subclasses of `TypeCodec` would implement them. 

However, I find that adds too much complexity for a single boolean method. I prefer the current approach, which is identical to the one used in `AbstractType#valueLengthIfFixed` et al.

src/java/org/apache/cassandra/cql3/functions/types/VectorCodec.java
adelapena: I guess we can do that by adding a new `TypeCodec#valueLengthIfFixed` method.

src/java/org/apache/cassandra/cql3/functions/types/VectorCodec.java
adelapena: Any improvement on how vectors are encoded should be done in `VectorType`, the `VectorCodec` proposed here is just the driver codec mimicking it. That kind of improvement, if it's worth it, would also apply to collections.

src/java/org/apache/cassandra/cql3/functions/types/VectorCodec.java
adelapena: I've put the `return` in a new line, if it's what you meant.

adelapena:
@bereng I've tried to address the rest of the comments. CI is running for j8: https://app.circleci.com/pipelines/github/adelapena/cassandra/3015/workflows/00c4b26c-2cf3-48c3-a46a-5553be9d45c7

I'll run j11 if j8 is ok and you agree with the changes.

src/java/org/apache/cassandra/cql3/functions/types/VectorCodec.java
bereng: The last commit can be greatly simplified imo, just serialize the first element and that tells you the size you need. No need for the `TypeCodec#valueLengthIfFixed` method.

If you prefer this approach though I would rename it to `serializedSize` which seems to be the name used in other serializers.

Wdyt?

adelapena: `valueLengthIfFixed` is the name used in `AbstractType`, but I guess the names can diverge based on preferences. I would however keep the fixed size method rather than trying to infer it from the first value.

bereng:
Ok +1 on green CI

adelapena:
Thanks, I've rebased and squashed. CI results in a bit.

PR #2447, opened 2023-06-25
Fix spelling error

josh-wong:
This PR fixes a spelling error ("DESRIBE" to "DESCRIBE") across multiple files.

I found this spelling mistake in the following error message when attempting to run a `DESC` command in `cqlsh`:

```cqlsh
WARN: DESCRIBE|DESC was moved to server side in Cassandra 4.0. As a consequence DESRIBE|DESC will not work in cqlsh '6.0.0' connected to Cassandra '3.0.8', the version that you are connected to. DESCRIBE does not exist server side prior Cassandra 4.0.
```


smiklosovic:
@josh-wong could you please create a PR against cassandra-4.0 branch instead of trunk? I think we need to fix this in 4.0 as well.

smiklosovic (changes requested):
please create a PR against cassandra-4.0 branch.

PR #2451, opened 2023-06-27
CASSANDRA-18494 Upgrade lucene-core library to the latest stable version

jbellis:
This PR upgrades to Lucene 9.7.

Notes on the upgrade path:
- RamIndexOutput is replaced with ResettableByteBuffersIndexOutput, an extension of ByteBuffersIndexOutput, which was the closest thing to a replacement of RamIndexOutput.
- ByteArrayIndexInput is gone with no real replacement.  ByteBufferIndexInput *looks* like it could be a replacement but it is very cumbersome to use for anything but mmapped buffers.
- Lucene exposes the code we needed from DirectReaders more or less directly in DirectReader now, so we can delete the old copied code.
- Lucene changed its data files to be little endian, but to keep its compatibility story simple it retained BE for the header and footer ints.  That's the cause of the changes in SAICodecUtils.
- We could gain a bit of performance making the rest of our code natively little endian but that is too big of a change for this patch.


jbellis:
CI results look good -- the only failures are ones that already existed

https://app.circleci.com/pipelines/github/adelapena/cassandra/2984/workflows/5d173e63-b61e-4480-ab56-513adcc643a4
https://app.circleci.com/pipelines/github/adelapena/cassandra/2984/workflows/62f01841-6e91-491c-8331-ce4e7aff3107

jbellis:
Note that Lucene has required JDK 11 since Lucene 8.  This creates a sequencing problem for us wrt [CASSANDRA-16895](https://issues.apache.org/jira/browse/CASSANDRA-16895) / [CASSANDRA-18255](https://issues.apache.org/jira/browse/CASSANDRA-18255) but it can't be helped, and upgrading is more important that most people probably realize since Lucene switched from big-endian byte order to little-endian in 9.0.  This means we will have to undertake *significantly* more complexity to maintain backwards compatibility if we GA with 7.5 and upgrade later.

My recommendation is to upgrade to 9.x now and accept the dependency on CASSANDRA-18255 to remove JDK 8.

src/java/org/apache/cassandra/index/sai/disk/oldlucene/ByteArrayIndexInput.java
```diff
@@ -0,0 +1,162 @@
...
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.store.RandomAccessInput;
+
+/**
+ * A {@link IndexInput} backed by a byte array.  Copied from Lucene 7.5.
+ *
+ * (ByteBufferIndexInput looks at first glance like a modern replacement for this, but
+ * it's really only intended to be used for reading mmapped buffers.)
+ */
+public final class ByteArrayIndexInput extends IndexInput implements RandomAccessInput
```
mike-tr-adamson: This isn't needed because the only use of the class in `MetadataSource` can be replaced with `ByteArrayDataInput` which looks to still exist in 9.x.

src/java/org/apache/cassandra/index/sai/disk/ResettableByteBuffersIndexOutput.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public ByteBuffersDataInput toDataInput()
+    {
+        return delegate.toDataInput();
+    }
+
+    public IndexInput toIndexInput()
+    {
+        return new ByteBuffersIndexInput(toDataInput(), "");
+    }
```
adelapena: Unused method

src/java/org/apache/cassandra/index/sai/disk/ResettableByteBuffersIndexOutput.java
```diff
@@ -0,0 +1,147 @@
...
+    {
+        delegate.reset();
+    }
+
+    public String toString()
+    {
+        return "Resettable" + bbio.toString();
+    }
+
+    public void close() throws IOException
```
adelapena, line 77: Nit: Starting on this one, all methods should use `@Override`

src/java/org/apache/cassandra/index/sai/disk/oldlucene/ByteArrayIndexInput.java
```diff
@@ -0,0 +1,162 @@
...
+import java.io.EOFException;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.Locale;
+
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.store.RandomAccessInput;
+
+/**
+ * A {@link IndexInput} backed by a byte array.  Copied from Lucene 7.5.
```
adelapena: I think we could add a line on the class JavaDoc saying that this comes from a previous version of Lucene. For example, `This is a copy of Lucene's ByteArrayIndexInput as it was in 7.5.0, which isn't present in the current version.`, or something like that.

Also, I think even if it's a copy of Lucene's code we should use our own brace placement style (BSD vs K&R)

src/java/org/apache/cassandra/index/sai/disk/ResettableByteBuffersIndexOutput.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.lucene.store.ByteBuffersIndexOutput;
+import org.apache.lucene.store.DataInput;
+import org.apache.lucene.store.IndexInput;
+import org.apache.lucene.store.IndexOutput;
+
+/***
+ * A wrapper around {@link ByteBuffersIndexOutput} that adds several methods that interact
+ * with the underlying delegate.
+ */
+public class ResettableByteBuffersIndexOutput extends IndexOutput
```
mike-tr-adamson, line 34: Again, this can be replaced by just using a `ByteBuffersDataOutput`.

jbellis: I think this is technically correct (there are several calls to getFilePointer() but I think they are equivalent to size()) but do we want to pin all usage to that specific class?

mike-tr-adamson: Actually, I'm looking at its usage in CASSANDRA-18067 (numerics) and it does need to be an `IndexOutput`. As such, this is a good replacement for `RAMIndexOutput`.

src/java/org/apache/cassandra/index/sai/disk/io/IndexInputReader.java
```diff
@@ -72,36 +75,6 @@ public void readBytes(byte[] bytes, int off, int len) throws IOException
...
-    /**
-     * Using {@link RandomAccessReader#readLong()} directly is faster than {@link DataInput#readLong()} which
-     * calls {@link DataInput#readByte()} one by one
-     */
-    @Override
-    public long readLong() throws IOException
-    {
-        return input.readLong();
-    }
-
```
maedhroz, line 104: Are these optimizations no longer necessary? it seems like `DataInput` still reads byte-by-byte...

Do we just never use these methods?

CC @mike-tr-adamson

src/java/org/apache/cassandra/index/sai/disk/io/IndexInputReader.java
```diff
@@ -33,6 +33,9 @@
  */
 public class IndexInputReader extends IndexInput
 {
+    // the byte order of `input`'s native readX operations doesn't matter,
+    // because we only use `readFully` and `readByte` methods. IndexInput calls these
+    // (via DataInput) with methods that enforce LittleEndian-ness.
```
maedhroz: Another fun way to do this that adds some navigability would be...

```
/**
 * The byte order of {@link RandomAccessReader}'s native read operations doesn't matter,
 * because we only use {@link RandomAccessReader#readFully} and {@link RandomAccessReader#readByte} methods.
 * IndexInput calls these (via {@link DataInput}) with methods that enforce LittleEndian-ness.
 */
```

src/java/org/apache/cassandra/index/sai/disk/io/IndexOutputWriter.java
```diff
@@ -37,6 +37,9 @@ public class IndexOutputWriter extends IndexOutput
 {
     private static final Logger logger = LoggerFactory.getLogger(IndexOutputWriter.class);
 
+    // the byte order of `out`'s native writeX operations doesn't matter,
+    // because we only use `write(byte[])` and `writeByte` methods. IndexOutput calls these
+    // (via DataOutput) with methods that enforce LittleEndian-ness.
```
maedhroz: Similar more navigable form of this:

```
/**
 * The byte order of {@link SequentialWriter}'s native write operations doesn't matter,
 * because we only use {@link SequentialWriter#write(byte[])} and {@link SequentialWriter#writeByte} methods.
 * {@link IndexOutput} calls these (via {@link DataOutput}) with methods that enforce LittleEndian-ness.
 */
```

src/java/org/apache/cassandra/index/sai/disk/io/IndexInputReader.java
jbellis: those changes are necessary because of the comment that you wanted to turn into a javadoc

simpler and probably more performant to remove them and let the parent class do its thing, than to override and add in a reverseBytes

src/java/org/apache/cassandra/index/sai/disk/io/IndexInputReader.java
jbellis: I could go either way on this, in general I prefer to keep implementation details as a comment instead of javadoc

src/java/org/apache/cassandra/index/sai/disk/ResettableByteBuffersIndexOutput.java
```diff
@@ -0,0 +1,146 @@
...
+    public byte[] toArrayCopy() {
+        return delegate.toArrayCopy();
+    }
+
+    public void reset()
+    {
+        delegate.reset();
+    }
+
+    public String toString()
```
maedhroz, line 71: super nit: throw an `@Override` on `toString()`

src/java/org/apache/cassandra/index/sai/memory/ByteSliceReader.java
```diff
@@ -116,6 +115,12 @@ public void nextSlice()
         }
     }
 
+    @Override
+    public void skipBytes(long l) throws IOException
```
maedhroz: nit: Don't strictly need the `throws IOException`

src/java/org/apache/cassandra/index/sai/disk/io/IndexInputReader.java
maedhroz: Yeah, most people will never read offline JavaDoc for this stuff, so I like to be able to click through to things in-IDE :)

src/java/org/apache/cassandra/index/sai/disk/v1/postings/PostingsReader.java
```diff
@@ -318,7 +313,7 @@ private int nextFoRValue()
         }
```
maedhroz: So above on 310...

Given the change below, replacing `null` w/ `LongValues.ZEROES`, do we even need `if/else` here, or does the logic just work without it?

src/java/org/apache/cassandra/index/sai/disk/v1/postings/PostingsReader.java
```diff
@@ -334,7 +329,10 @@ private void advanceOnePosition(long nextPosting)
     private void reBuffer() throws IOException
     {
         long pointer = summary.offsets.get(blockIndex);
-
+        if (pointer < 4) {
+            // the first 4 bytes must be CODEC_MAGIC
+            throw new CorruptIndexException(String.format("Invalid block offset %d for postings block idx %d", pointer, blockIndex), input);
+        }
```
maedhroz, line 328: nit: could just blow the curly braces away on this one...it'll never be more than a single line I guess

src/java/org/apache/cassandra/index/sai/disk/v1/MetadataWriter.java
```diff
@@ -43,17 +43,17 @@ public Builder builder(String name)
         return new Builder(name);
     }
 
-    public class Builder extends RAMIndexOutput implements Closeable
+    public class Builder extends ResettableByteBuffersIndexOutput implements Closeable
     {
         private Builder(String name)
         {
-            super(name);
+            super(1024, name);
```
maedhroz: Curious about why 1024 made the most sense here...

maedhroz (approved):
Made a couple little nits, but overall, LGTM

src/java/org/apache/cassandra/index/sai/disk/v1/postings/PostingsReader.java
jbellis: good catch, fixed

src/java/org/apache/cassandra/index/sai/disk/v1/postings/PostingsReader.java
jbellis: true, but i prefer the braces just to make it obvious that it's correct, even though the second line is just a comment

src/java/org/apache/cassandra/index/sai/disk/v1/MetadataWriter.java
jbellis: I ... don't remember.  I put it back to 128 like it was before and updated CASSANDRA-18280 to point to the new code.

src/java/org/apache/cassandra/index/sai/disk/v1/postings/PostingsReader.java
maedhroz: Quick, put the first `{` on a new line!

PR #2458, opened 2023-06-29
CASSANDRA-18438

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
+++ b/src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
@@ -22,30 +22,72 @@ import java.io.DataInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.Map;
 
 import com.google.common.collect.ImmutableMap;
 
+import org.apache.cassandra.config.CassandraRelevantProperties;
+import org.apache.cassandra.config.DurationSpec;
+import org.apache.cassandra.exceptions.ConfigurationException;
+
+import static java.lang.String.format;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 abstract class AbstractCloudMetadataServiceConnector
 {
+    static final String METADATA_URL_PROPERTY = "metadata_url";
+    static final String METADATA_REQUEST_TIMEOUT_PROPERTY = "metadata_request_timeout";
+    static final String DEFAULT_METADATA_REQUEST_TIMEOUT = "30s";
+
     protected final String metadataServiceUrl;
+    protected final int requestTimeoutMs;
 
-    protected AbstractCloudMetadataServiceConnector(String metadataServiceUrl)
+    public AbstractCloudMetadataServiceConnector(SnitchProperties properties)
     {
-        this.metadataServiceUrl = metadataServiceUrl;
+        String parsedMetadataServiceUrl = properties.get(METADATA_URL_PROPERTY, null);
+
+        try
+        {
+            URL url = new URL(parsedMetadataServiceUrl);
+            url.toURI();
+
+            this.metadataServiceUrl = parsedMetadataServiceUrl;
+        }
+        catch (MalformedURLException | IllegalArgumentException | URISyntaxException ex)
+        {
+            throw new ConfigurationException(format("Snitch metadata service URL '%s' is invalid. Please review snitch properties " +
+                                                    "defined in the configured '%s' configuration file.",
+                                                    parsedMetadataServiceUrl,
+                                                    CassandraRelevantProperties.CASSANDRA_RACKDC_PROPERTIES.getKey()),
+                                             ex);
+        }
+
+
+        String metadataRequestTimeout = properties.get(METADATA_REQUEST_TIMEOUT_PROPERTY, DEFAULT_METADATA_REQUEST_TIMEOUT);
+
+        try
+        {
+            this.requestTimeoutMs = new DurationSpec.IntMillisecondsBound(metadataRequestTimeout).toMilliseconds();
+        }
+        catch (IllegalArgumentException ex)
+        {
+            throw new ConfigurationException(format("%s as value of %s is invalid duration! " + ex.getMessage(),
+                                                    metadataRequestTimeout,
+                                                    METADATA_REQUEST_TIMEOUT_PROPERTY));
+        }
     }
 
-    public String apiCall(String query) throws IOException
+    public final String apiCall(String query) throws IOException
     {
-        return apiCall(metadataServiceUrl, query, 200);
+        return apiCall(metadataServiceUrl, query, "GET", ImmutableMap.of(), 200);
     }
 
-    public String apiCall(String url, String query, int expectedResponseCode) throws IOException
+    public final String apiCall(String query, Map<String, String> extraHeaders) throws IOException
     {
-        return apiCall(url, query, "GET", ImmutableMap.of(), expectedResponseCode);
+        return apiCall(metadataServiceUrl, query, "GET", extraHeaders, 200);
     }
 
     public String apiCall(String url,
@@ -61,6 +103,7 @@ abstract class AbstractCloudMetadataServiceConnector
             conn = (HttpURLConnection) new URL(url + query).openConnection();
             extraHeaders.forEach(conn::setRequestProperty);
             conn.setRequestMethod(method);
+            conn.setConnectTimeout(requestTimeoutMs);
             if (conn.getResponseCode() != expectedResponseCode)
                 throw new HttpException(conn.getResponseCode(), conn.getResponseMessage());
 
@@ -84,6 +127,14 @@ abstract class AbstractCloudMetadataServiceConnector
         }
     }
 
+    @Override
+    public String toString()
+    {
+        return format("%s{%s=%s,%s=%s}", getClass().getName(),
+                      METADATA_URL_PROPERTY, metadataServiceUrl,
+                      METADATA_REQUEST_TIMEOUT_PROPERTY, requestTimeoutMs);
+    }
+
     public static final class HttpException extends IOException
     {
         public final int responseCode;
@@ -94,6 +145,14 @@ abstract class AbstractCloudMetadataServiceConnector
             super("HTTP response code: " + responseCode + " (" + responseMessage + ')');
             this.responseCode = responseCode;
             this.responseMessage = responseMessage;
+        }
+    }
+
+    public static class DefaultCloudMetadataServiceConnector extends AbstractCloudMetadataServiceConnector
+    {
+        public DefaultCloudMetadataServiceConnector(SnitchProperties properties)
+        {
+            super(properties);
         }
     }
 }
--- a/src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceSnitch.java
+++ b/src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceSnitch.java
@@ -1,0 +1,108 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.locator;
+
+import java.util.Map;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.cassandra.db.SystemKeyspace;
+import org.apache.cassandra.gms.ApplicationState;
+import org.apache.cassandra.gms.EndpointState;
+import org.apache.cassandra.gms.Gossiper;
+import org.apache.cassandra.utils.FBUtilities;
+import org.apache.cassandra.utils.Pair;
+
+import static java.lang.String.format;
+
+abstract class AbstractCloudMetadataServiceSnitch extends AbstractNetworkTopologySnitch
+{
+    static final Logger logger = LoggerFactory.getLogger(AbstractCloudMetadataServiceSnitch.class);
+
+    static final String DEFAULT_DC = "UNKNOWN-DC";
+    static final String DEFAULT_RACK = "UNKNOWN-RACK";
+
+    protected final AbstractCloudMetadataServiceConnector connector;
+    protected final SnitchProperties snitchProperties;
+
+    private final String localRack;
+    private final String localDc;
+
+    private Map<InetAddressAndPort, Map<String, String>> savedEndpoints;
+
+    public AbstractCloudMetadataServiceSnitch(AbstractCloudMetadataServiceConnector connector,
+                                              SnitchProperties snitchProperties,
+                                              Pair<String, String> dcAndRack)
+    {
+        this.connector = connector;
+        this.snitchProperties = snitchProperties;
+        this.localDc = dcAndRack.left;
+        this.localRack = dcAndRack.right;
+
+        logger.info(format("%s using datacenter: %s, rack: %s, connector: %s, properties: %s",
+                           getClass().getName(), getLocalDatacenter(), getLocalRack(), connector, snitchProperties));
+    }
+
+    @Override
+    public final String getLocalRack()
+    {
+        return localRack;
+    }
+
+    @Override
+    public final String getLocalDatacenter()
+    {
+        return localDc;
+    }
+
+    @Override
+    public final String getRack(InetAddressAndPort endpoint)
+    {
+        if (endpoint.equals(FBUtilities.getBroadcastAddressAndPort()))
+            return getLocalRack();
+        EndpointState state = Gossiper.instance.getEndpointStateForEndpoint(endpoint);
+        if (state == null || state.getApplicationState(ApplicationState.RACK) == null)
+        {
+            if (savedEndpoints == null)
+                savedEndpoints = SystemKeyspace.loadDcRackInfo();
+            if (savedEndpoints.containsKey(endpoint))
+                return savedEndpoints.get(endpoint).get("rack");
+            return DEFAULT_RACK;
+        }
+        return state.getApplicationState(ApplicationState.RACK).value;
+    }
+
+    @Override
+    public final String getDatacenter(InetAddressAndPort endpoint)
+    {
+        if (endpoint.equals(FBUtilities.getBroadcastAddressAndPort()))
+            return getLocalDatacenter();
+        EndpointState state = Gossiper.instance.getEndpointStateForEndpoint(endpoint);
+        if (state == null || state.getApplicationState(ApplicationState.DC) == null)
+        {
+            if (savedEndpoints == null)
+                savedEndpoints = SystemKeyspace.loadDcRackInfo();
+            if (savedEndpoints.containsKey(endpoint))
+                return savedEndpoints.get(endpoint).get("data_center");
+            return DEFAULT_DC;
+        }
+        return state.getApplicationState(ApplicationState.DC).value;
+    }
+}
--- a/src/java/org/apache/cassandra/locator/AlibabaCloudSnitch.java
+++ b/src/java/org/apache/cassandra/locator/AlibabaCloudSnitch.java
@@ -17,130 +17,41 @@
  */
 package org.apache.cassandra.locator;
 
-import java.io.DataInputStream;
-import java.io.FilterInputStream;
 import java.io.IOException;
-import java.net.HttpURLConnection;
-import java.net.MalformedURLException;
-import java.net.SocketTimeoutException;
-import java.net.URL;
-import java.nio.charset.StandardCharsets;
-import java.util.Map;
-import org.apache.cassandra.db.SystemKeyspace;
-import org.apache.cassandra.exceptions.ConfigurationException;
-import org.apache.cassandra.gms.ApplicationState;
-import org.apache.cassandra.gms.EndpointState;
-import org.apache.cassandra.gms.Gossiper;
-import org.apache.cassandra.io.util.FileUtils;
-import org.apache.cassandra.utils.FBUtilities;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+
+import com.google.common.collect.ImmutableMap;
+
+import org.apache.cassandra.locator.AbstractCloudMetadataServiceConnector.DefaultCloudMetadataServiceConnector;
+
+import static org.apache.cassandra.locator.AbstractCloudMetadataServiceConnector.METADATA_URL_PROPERTY;
 
 /**
- *  A snitch that assumes an ECS region is a DC and an ECS availability_zone
- *  is a rack. This information is available in the config for the node. the 
- *  format of the zone-id is like :cn-hangzhou-a where cn means china, hangzhou
- *  means the hangzhou region, a means the az id. We use cn-hangzhou as the dc,
- *  and f as the zone-id.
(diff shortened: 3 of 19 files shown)


src/java/org/apache/cassandra/locator/AlibabaCloudSnitch.java
```diff
@@ -17,130 +17,47 @@
...
- *  A snitch that assumes an ECS region is a DC and an ECS availability_zone
- *  is a rack. This information is available in the config for the node. the 
- *  format of the zone-id is like :cn-hangzhou-a where cn means china, hangzhou
- *  means the hangzhou region, a means the az id. We use cn-hangzhou as the dc,
- *  and f as the zone-id.
+ * A snitch that assumes an ECS region is a DC and an ECS availability_zone
+ * is a rack. This information is available in the config for the node. the
+ * format of the zone-id is like :cn-hangzhou-a where cn means china, hangzhou
+ * means the hangzhou region, a means the az id. We use cn-hangzhou as the dc,
+ * and f as the zone-id.
```
Maxwell-Guo: sorry I make a misstake at the first time push this pr, this should be "a as the zone-id" not f.

src/java/org/apache/cassandra/locator/AlibabaCloudSnitch.java
```diff
@@ -17,130 +17,47 @@
...
-        {
-            if (savedEndpoints == null)
-                savedEndpoints = SystemKeyspace.loadDcRackInfo();
-            if (savedEndpoints.containsKey(endpoint))
-                return savedEndpoints.get(endpoint).get("rack");
-            return DEFAULT_RACK;
-        }
-        return state.getApplicationState(ApplicationState.RACK).value;
-    
+        this(properties, new DefaultCloudMetadataServiceConnector(properties.get(METADATA_URL_PROPERTY,
```
Maxwell-Guo: what about merge METADATA_URL_PROPERTY and ZONE_NAME_QUERY_URL  together as before not just split them into two Strings and make them configurable. 
or make ZONE_NAME_QUERY_URL configurable through Snitchpropetry ? (In this way I may suggest changing the name to ZONE_NAME_QUERY_URL_PATH)

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
```diff
@@ -25,22 +25,39 @@
...
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 abstract class AbstractCloudMetadataServiceConnector
 {
+    static final String METADATA_URL_PROPERTY = "metadata_url";
+
+    public static class DefaultCloudMetadataServiceConnector extends AbstractCloudMetadataServiceConnector
+    {
+        protected DefaultCloudMetadataServiceConnector(String metadataServiceUrl)
```
Maxwell-Guo: as said before , what about make this metaserviceUrl with the full url path ,not only the url address.
Or make both zone path configurable which may be more flexible.

Maxwell-Guo (commented):
Sorry, I should have added some configuration  [here](https://github.com/apache/cassandra/blob/cassandra-4.0.8/conf/cassandra.yaml#L1063) at the beginning when I start alibabaSnitch (also I saw google snitch is missed too), that is why I want to make cloud platform snitch configurable, I do not want to add a comment in yaml, every time I after add a new snitch. next time Azure snitch is added we should a for that
The missing words are 
"
 GoogleCloudSnitch:
    Appropriate for Google Compute Engine deployments in a single Region. Loads Region
    and Availability Zone information from the Compute Engine API. The Region is
    treated as the datacenter, and the Availability Zone as the rack.
    Only private IPs are used, so this will not work across multiple
    Regions.

 AlibabaCloudSnitch:
    Appropriate for ECS deployments in a single Region. Loads Region
    and Availability Zone information from the ECS API. The Region is
    treated as the datacenter, and the Availability Zone as the rack.
    Only private IPs are used, so this will not work across multiple
    Regions.

"
@smiklosovic  can you help to add when this pr merged.

Besides, as this pr may suggest users to set configuration through snitch properties ,we may add some guidance docs for this;

we should add a example at [here](https://github.com/apache/cassandra/blob/trunk/conf/cassandra-rackdc.properties) ,as  you just add dc_suffix for meta_url ,there should be doc for dc_suffix, if you have accept my suggestion with add zone_query_url or merge them together ,  maybe the documentation needs another change.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
smiklosovic: It is metadata service address only because that is variable we configure in mocking tests. See `Ec2V2ConnectorMockingTest` in trunk.

So the idea is that we change just the service url but not the endpoints.

src/java/org/apache/cassandra/locator/AlibabaCloudSnitch.java
smiklosovic: See above comment about why it is split into two properties.

Once we implement "DefaultCloudServiceSnitch", we can do it like 1 address with whole endpoint.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceSnitch.java
```diff
@@ -0,0 +1,104 @@
...
+{
+    static final Logger logger = LoggerFactory.getLogger(AbstractCloudMetadataServiceSnitch.class);
+
+    static final String DEFAULT_DC = "UNKNOWN-DC";
+    static final String DEFAULT_RACK = "UNKNOWN-RACK";
+
+    protected final AbstractCloudMetadataServiceConnector connector;
+    protected final SnitchProperties snitchProperties;
+
+    protected String localRack;
```
jacek-lewandowski: So my preference is that those two fields are private final in the subclasses and in this class we only have abstract methods like `getLocalRack()` and `getLocalDC()` which are called from `getRack(endpoint)` and `getDatacenter(endpoint)`

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceSnitch.java
```diff
@@ -0,0 +1,104 @@
...
+            if (savedEndpoints == null)
+                savedEndpoints = SystemKeyspace.loadDcRackInfo();
+            if (savedEndpoints.containsKey(endpoint))
+                return savedEndpoints.get(endpoint).get("data_center");
+            return DEFAULT_DC;
+        }
+        return state.getApplicationState(ApplicationState.DC).value;
+    }
+
+    protected void parseResponse(String response)
```
jacek-lewandowski: I think that this method does not belong to this abstract implementation. There should be no details about the response interpretation. If this method is used in many places, perhaps it could go as static to something like `SnitchUtils`.

src/java/org/apache/cassandra/locator/Ec2MetadataServiceConnector.java
```diff
@@ -45,7 +45,7 @@ abstract class Ec2MetadataServiceConnector extends AbstractCloudMetadataServiceC
 
     Ec2MetadataServiceConnector(SnitchProperties properties)
     {
-        super(properties.get(EC2_METADATA_URL_PROPERTY, DEFAULT_EC2_METADATA_URL));
+        super(properties.get(EC2_METADATA_URL_PROPERTY, properties.get(METADATA_URL_PROPERTY, DEFAULT_EC2_METADATA_URL)));
```
jacek-lewandowski: why do we need this?

src/java/org/apache/cassandra/locator/DefaultCloudMetadataServiceSnitch.java
```diff
@@ -0,0 +1,63 @@
...
+ * Default cloud-based metadata service snitch which parses datacenter and rack of a node by
+ * executing GET query for URL configured in cassandra-rackdc.properties under
+ * key {@link AbstractCloudMetadataServiceConnector#METADATA_URL_PROPERTY}. It expects HTTP response code 200.
+ * It is not passing any headers to GET request.
+ * <p>
+ * The response will be parsed like following - if the response body is, for example "us-central1-a", then
+ * the datacenter will be "us-central1" and rack will be "a". There is value of "dc_suffix" key in
+ * cassandra-rackdc.properties appended to datacenter.
+ */
+public class DefaultCloudMetadataServiceSnitch extends AbstractCloudMetadataServiceSnitch
```
jacek-lewandowski: does this class need to be public?

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
```diff
@@ -25,22 +25,41 @@
...
 
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 abstract class AbstractCloudMetadataServiceConnector
 {
+    static final String METADATA_URL_PROPERTY = "metadata_url";
```
jacek-lewandowski, line 41: CassandraRelevantProperties?

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
```diff
@@ -25,22 +25,41 @@
...
 
+    @VisibleForTesting // for mockito
+    protected String getMetadataServiceUrl()
+    {
+        return metadataServiceUrl;
+    }
+
     protected AbstractCloudMetadataServiceConnector(String metadataServiceUrl)
     {
+        Preconditions.checkNotNull(metadataServiceUrl, METADATA_URL_PROPERTY + " for a connector can not be null");
```
jacek-lewandowski: Since you do validation, maybe also validate if it is a valid URL?

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
```diff
@@ -25,22 +25,41 @@
...
         this.metadataServiceUrl = metadataServiceUrl;
     }
 
     public String apiCall(String query) throws IOException
     {
-        return apiCall(metadataServiceUrl, query, 200);
+        return apiCall(getMetadataServiceUrl(), query, 200);
```
jacek-lewandowski: why?

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
```diff
@@ -61,6 +80,7 @@ public String apiCall(String url,
             conn = (HttpURLConnection) new URL(url + query).openConnection();
             extraHeaders.forEach(conn::setRequestProperty);
             conn.setRequestMethod(method);
+            conn.setConnectTimeout(30000);
```
jacek-lewandowski: Should we take the timeout from the properties with 30s as default?

src/java/org/apache/cassandra/locator/DefaultCloudMetadataServiceSnitch.java
```diff
@@ -0,0 +1,63 @@
...
+{
+    public DefaultCloudMetadataServiceSnitch() throws IOException
+    {
+        this(new SnitchProperties());
+    }
+
+    public DefaultCloudMetadataServiceSnitch(SnitchProperties properties) throws IOException
```
jacek-lewandowski: maybe let it accept the default url as argument so that the inheritors can set it directly?

src/java/org/apache/cassandra/locator/DefaultCloudMetadataServiceSnitch.java
```diff
@@ -0,0 +1,63 @@
...
+ * executing GET query for URL configured in cassandra-rackdc.properties under
+ * key {@link AbstractCloudMetadataServiceConnector#METADATA_URL_PROPERTY}. It expects HTTP response code 200.
+ * It is not passing any headers to GET request.
+ * <p>
+ * The response will be parsed like following - if the response body is, for example "us-central1-a", then
+ * the datacenter will be "us-central1" and rack will be "a". There is value of "dc_suffix" key in
+ * cassandra-rackdc.properties appended to datacenter.
+ */
+public class DefaultCloudMetadataServiceSnitch extends AbstractCloudMetadataServiceSnitch
+{
```
jacek-lewandowski: would you rename it into something like `GenericCloudMetadataServiceSnitch`?

jacek-lewandowski (changes requested):

src/java/org/apache/cassandra/locator/DefaultCloudMetadataServiceSnitch.java
smiklosovic: Yes, I believe it needs to be public. Keep in mind that we might reference this snitch in `cassandra.yaml` so we would be able to use this snitch for all cloud deployments when the response is parsed in a very opinionated way, for free, out of the box. 

If this was not public, then it is not able to be constructed as making it package protected would throw:

    org.apache.cassandra.exceptions.ConfigurationException: Default constructor for snitch class 
    'org.apache.cassandra.locator.MySnitch' is inaccessible.
	at org.apache.cassandra.utils.FBUtilities.construct(FBUtilities.java:762)
	at org.apache.cassandra.utils.FBUtilities.construct(FBUtilities.java:751)
	at org.apache.cassandra.config.DatabaseDescriptor.createEndpointSnitch(DatabaseDescriptor.java:1491)
	at org.apache.cassandra.config.DatabaseDescriptor.applySnitch(DatabaseDescriptor.java:1338)
	at org.apache.cassandra.config.DatabaseDescriptor.applyAll(DatabaseDescriptor.java:435)
	at org.apache.cassandra.config.DatabaseDescriptor.daemonInitialization(DatabaseDescriptor.java:244)
	at org.apache.cassandra.config.DatabaseDescriptor.daemonInitialization(DatabaseDescriptor.java:228)
	at org.apache.cassandra.service.CassandraDaemon.applyConfig(CassandraDaemon.java:819)
	at org.apache.cassandra.service.CassandraDaemon.activate(CassandraDaemon.java:762)
	at org.apache.cassandra.service.CassandraDaemon.main(CassandraDaemon.java:890)

With snitch like this (same holds for the DefaultCloudMetadataServiceSnitch)

    class MySnitch extends AbstractNetworkTopologySnitch
    {
        @Override
        public String getRack(InetAddressAndPort endpoint)
        {
            return "r1";
        }

        @Override
        public String getDatacenter(InetAddressAndPort endpoint)
        {
            return "dc1";
        }
    }

This does not help either

    class MySnitch extends AbstractNetworkTopologySnitch
    {
         public MySnitch() {}
         ....
    }

We can not make it final either because `AlibabaCloudSnitch` is extending it.

src/java/org/apache/cassandra/locator/Ec2MetadataServiceConnector.java
smiklosovic: Good question.

`METADATA_URL_PROPERTY = "metadata_url"`

and 

`EC2_METADATA_URL_PROPERTY = "ec2_metadata_url"`.

The latter was introduced when we were doing IDMSv2 for AWS but with this PR in place, I wanted to be backward compatible.

As `ec2_metadata_url` key is in 3.0 -> 4.1 but it would not be in trunk, then upgrading to trunk would fail because that key would not be there anymore. 

On the other hand, if one deploys it to trunk for the first time, it can set the url like "metadata_url" and I like the idea that metadata_url property would be same for all cloud snitches.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
smiklosovic: I dont think that is actually necessary. It is not like this property is a _system property_. This is just a key of a property in cassandra-rackdc.properties.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
smiklosovic: This is here due to tests. I had to mock what that method would return, this can not be done on a field.

src/java/org/apache/cassandra/locator/DefaultCloudMetadataServiceSnitch.java
smiklosovic: I don't think that is really necessary. 

I can do this though.

    public DefaultCloudMetadataServiceSnitch(String url) throws IOException
    {
        this(new SnitchProperties(Pair.create(METADATA_URL_PROPERTY, url)));
    }

If you look how the response is parsed etc, we also need dc suffix propagated and in case like I did it it would not be parsed from property file.

I can also do this

    public DefaultCloudMetadataServiceSnitch(SnitchProperties properties, String url) throws IOException
    {
        this(properties, new DefaultCloudMetadataServiceConnector(url));
    }

But why would you do that if it is going to be taken from `properties` anyway as we need them?

src/java/org/apache/cassandra/locator/DefaultCloudMetadataServiceSnitch.java
smiklosovic: works for me.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
smiklosovic: well it is tricky to pass anything to these connectors from the properties as currently it is accepting metadata service url only. We might either refactor bunch of stuff for such a triviality or we can just have a system property in `CassandraRelevantProperties`? It is hard to believe this would be something which would be tweaked by users regularly, maybe for power users as the ultimate way to weak it but otherwise ....

src/java/org/apache/cassandra/locator/Ec2MetadataServiceConnector.java
jacek-lewandowski: perhaps we should not bother that much - who would set it to something else than a default value?

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
jacek-lewandowski: ah, yes, right

src/java/org/apache/cassandra/locator/Ec2MetadataServiceConnector.java
smiklosovic: yeah ... true, hm.

smiklosovic: anyway it does not hurt if it is here.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceSnitch.java
```diff
@@ -0,0 +1,107 @@
...
+                           getClass().getName(), getLocalDc(), getLocalRack(), connector, snitchProperties));
+    }
+
+    @Override
+    public String getLocalRack()
+    {
+        return localRack;
+    }
+
+    public String getLocalDc()
```
jacek-lewandowski: I don't know if it is intentional, but there is a `getLocalDatacenter()` method in the `IEndpointSnitch` interface with some default implementation

smiklosovic: great catch.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
jacek-lewandowski: Nevertheless, I think this is not needed. I saw your mock and the reason why it fails with NPE is passing `anyString()` as url arg for `apiCall` - simply change it to `any()` and it will accept nulls.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
jacek-lewandowski: since we have a generic metadata_url - why not just have a generic metadata_request_timeout in snitch properties? Though, I don't insist on this one 😃

Maxwell-Guo: the original codes are all hard code, I think we can make this property in snitch with default value of 30s , also  I don't insist on this one ,same with jacek.

src/java/org/apache/cassandra/locator/DefaultCloudMetadataServiceSnitch.java
jacek-lewandowski: ok, let's leave it as it is

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
```diff
@@ -22,25 +22,55 @@
...
         this.metadataServiceUrl = metadataServiceUrl;
     }
 
     public String apiCall(String query) throws IOException
     {
-        return apiCall(metadataServiceUrl, query, 200);
+        return apiCall(getMetadataServiceUrl(), query, 200);
     }
 
     public String apiCall(String url, String query, int expectedResponseCode) throws IOException
```
jacek-lewandowski: I think this method should be converted into:

```
apiCall(String query, Map<String, String> extraHeaders)
```

and then used, for example in Google snitch. 

In general, we should avoid using `apiCall` which accepts url - it should be reserved only for extensions. That is, the other `apiCall` overloads should be made final.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
```diff
@@ -22,25 +22,55 @@
...
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 
+import org.apache.cassandra.exceptions.ConfigurationException;
+
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 abstract class AbstractCloudMetadataServiceConnector
 {
+    static final String METADATA_URL_PROPERTY = "metadata_url";
```
jacek-lewandowski, line 41: This property is owned by `GenericMDSSnitch` so maybe move it there?

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
```diff
@@ -22,25 +22,55 @@
...
     protected AbstractCloudMetadataServiceConnector(String metadataServiceUrl)
     {
+        try
+        {
+            URL url = new URL(metadataServiceUrl);
+            url.toURI();
+        }
+        catch (MalformedURLException | IllegalArgumentException | URISyntaxException ex)
+        {
+            throw new ConfigurationException(String.format("URL you provided (%s) is not valid!", metadataServiceUrl), ex);
```
jacek-lewandowski: Maybe something like: 

Snitch metadata service URL '%s' is invalid. Please review snitch properties defined in the 'cassandra-rackdc.properties' configuration file.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
smiklosovic: nobody is ever going to configure it, we dont need this

smiklosovic: Anyway, I think I will do it in such a way that we will pass SnitchProperties to connector. Because my gut feeling tells me that there might be connectors which might be somehow further configured. Now it is pretty simple, just metadata url and possibly timeout, but what if connectors need something fancy? I dont know ... certificates? You feel me. So there needs to be a way how to get the access to all snitch properties.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
smiklosovic: I dont think it is owned by it. It is good where it is imho.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
```diff
@@ -22,30 +22,77 @@
...
+import org.apache.cassandra.config.CassandraRelevantProperties;
+import org.apache.cassandra.exceptions.ConfigurationException;
+
+import static java.lang.String.format;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 abstract class AbstractCloudMetadataServiceConnector
 {
+    static final String METADATA_URL_PROPERTY = "metadata_url";
+    static final String METADATA_REQUEST_TIMEOUT_PROPRETY = "metadata_request_timeout";
```
jacek-lewandowski: typo PROPRETY -> PROPERTY

jacek-lewandowski (approved):

Maxwell-Guo (approved):

PR #2460, opened 2023-06-30
Cassandra 18490 Fix checksum validation and checksum validate indexes streaming and SSTable import

pkolaczk:
* Fix checksum writer used by SAI
* Introduce `IndexValidation` enum for greater internal API flexibility
* Do full checksum validation on opening sstables, except node reboot / sstable flush / compaction.

### DIFF
--- a/src/java/org/apache/cassandra/index/sai/IndexContext.java
+++ b/src/java/org/apache/cassandra/index/sai/IndexContext.java
@@ -162,9 +162,9 @@ public class IndexContext
     /**
      * @return A set of SSTables which have attached to them invalid index components.
      */
-    public Collection<SSTableContext> onSSTableChanged(Collection<SSTableReader> oldSSTables, Collection<SSTableContext> newSSTables, boolean validate)
+    public Collection<SSTableContext> onSSTableChanged(Collection<SSTableReader> oldSSTables, Collection<SSTableContext> newSSTables, IndexValidation validation)
     {
-        return viewManager.update(oldSSTables, newSSTables, validate);
+        return viewManager.update(oldSSTables, newSSTables, validation);
     }
 
     public ColumnMetadata getDefinition()
@@ -408,7 +408,7 @@ public class IndexContext
      * @return the indexes that are built on the given SSTables on the left and corrupted indexes'
      * corresponding contexts on the right
      */
-    public Pair<Collection<SSTableIndex>, Collection<SSTableContext>> getBuiltIndexes(Collection<SSTableContext> sstableContexts, boolean validate)
+    public Pair<Collection<SSTableIndex>, Collection<SSTableContext>> getBuiltIndexes(Collection<SSTableContext> sstableContexts, IndexValidation validation)
     {
         Set<SSTableIndex> valid = new HashSet<>(sstableContexts.size());
         Set<SSTableContext> invalid = new HashSet<>();
@@ -433,9 +433,9 @@ public class IndexContext
 
             try
             {
-                if (validate)
+                if (validation != IndexValidation.NONE)
                 {
-                    if (!sstableContext.indexDescriptor.validatePerIndexComponents(this))
+                    if (!sstableContext.indexDescriptor.validatePerIndexComponents(this, validation))
                     {
                         logger.warn(logMessage("Invalid per-column component for SSTable {}"), sstableContext.descriptor());
                         invalid.add(sstableContext);
--- a/src/java/org/apache/cassandra/index/sai/IndexValidation.java
+++ b/src/java/org/apache/cassandra/index/sai/IndexValidation.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.index.sai;
+
+/**
+ * Whether to validate indexes and how
+ */
+public enum IndexValidation
+{
+    /**
+     * No validation to be performed
+     */
+    NONE,
+
+    /**
+     * Basic header validation, no data validation (fast)
+     */
+    BASIC,
+
+    /**
+     * Full validation with checksumming data (slow)
+     */
+    CHECKSUM;
+
+}
--- a/src/java/org/apache/cassandra/index/sai/IndexValidation.java
+++ b/src/java/org/apache/cassandra/index/sai/IndexValidation.java
@@ -31,7 +31,7 @@ public enum IndexValidation
     /**
      * Basic header/footer validation, but no data validation (fast)
      */
-    BASIC,
+    HEADER_FOOTER,
 
     /**
      * Full validation with checksumming data (slow)
--- a/src/java/org/apache/cassandra/index/sai/IndexValidation.java
+++ b/src/java/org/apache/cassandra/index/sai/IndexValidation.java
@@ -29,7 +29,7 @@ public enum IndexValidation
     NONE,
 
     /**
-     * Basic header validation, no data validation (fast)
+     * Basic header/footer validation, but no data validation (fast)
      */
     BASIC,
 
--- a/src/java/org/apache/cassandra/index/sai/IndexValidation.java
+++ b/src/java/org/apache/cassandra/index/sai/IndexValidation.java
@@ -36,6 +36,6 @@ public enum IndexValidation
     /**
      * Full validation with checksumming data (slow)
      */
-    CHECKSUM;
+    CHECKSUM
 
 }
--- a/src/java/org/apache/cassandra/index/sai/IndexValidation.java
+++ b/src/java/org/apache/cassandra/index/sai/IndexValidation.java
@@ -18,9 +18,6 @@
 
 package org.apache.cassandra.index.sai;
 
-/**
- * Whether to validate indexes and how
- */
 public enum IndexValidation
 {
     /**
--- a/src/java/org/apache/cassandra/index/sai/SSTableContextManager.java
+++ b/src/java/org/apache/cassandra/index/sai/SSTableContextManager.java
@@ -47,12 +47,12 @@ public class SSTableContextManager
      *
      * @param removed SSTables being removed
      * @param added SSTables being added
-     * @param validate if true, header and footer will be validated.
+     * @param validation Controls how indexes should be validated
      *
      * @return a set of contexts for SSTables with valid per-SSTable components, and a set of
      * SSTables with invalid or missing components
      */
-    public Pair<Set<SSTableContext>, Set<SSTableReader>> update(Collection<SSTableReader> removed, Iterable<SSTableReader> added, boolean validate)
+    public Pair<Set<SSTableContext>, Set<SSTableReader>> update(Collection<SSTableReader> removed, Iterable<SSTableReader> added, IndexValidation validation)
     {
         release(removed);
 
@@ -77,7 +77,7 @@ public class SSTableContextManager
             try
             {
                 // Only validate on restart or newly refreshed SSTable. Newly built files are unlikely to be corrupted.
-                if (validate && !sstableContexts.containsKey(sstable) && !indexDescriptor.validatePerSSTableComponents())
+                if (!sstableContexts.containsKey(sstable) && !indexDescriptor.validatePerSSTableComponents(validation))
                 {
                     logger.warn(indexDescriptor.logMessage("Invalid per-SSTable component for SSTable {}"), sstable.descriptor);
                     invalid.add(sstable);
--- a/src/java/org/apache/cassandra/index/sai/StorageAttachedIndexBuilder.java
+++ b/src/java/org/apache/cassandra/index/sai/StorageAttachedIndexBuilder.java
@@ -262,7 +262,7 @@ public class StorageAttachedIndexBuilder extends SecondaryIndexBuilder
         // if per-table files are incomplete or checksum failed during full rebuild.
         IndexDescriptor indexDescriptor = IndexDescriptor.create(sstable);
         if (!indexDescriptor.isPerSSTableIndexBuildComplete() ||
-            (isFullRebuild && !indexDescriptor.validatePerSSTableComponentsChecksum()))
+            (isFullRebuild && !indexDescriptor.validatePerSSTableComponents(IndexValidation.CHECKSUM)))
         {
             CountDownLatch latch = CountDownLatch.newCountDownLatch(1);
             if (inProgress.putIfAbsent(sstable, latch) == null)
@@ -307,7 +307,7 @@ public class StorageAttachedIndexBuilder extends SecondaryIndexBuilder
 
         // register custom index components into existing sstables
         sstable.registerComponents(StorageAttachedIndexGroup.getLiveComponents(sstable, existing), tracker);
-        Set<StorageAttachedIndex> incomplete = group.onSSTableChanged(Collections.emptyList(), Collections.singleton(sstable), existing, false);
+        Set<StorageAttachedIndex> incomplete = group.onSSTableChanged(Collections.emptyList(), Collections.singleton(sstable), existing, IndexValidation.NONE);
 
         if (!incomplete.isEmpty())
         {
--- a/src/java/org/apache/cassandra/index/sai/StorageAttachedIndexBuilder.java
+++ b/src/java/org/apache/cassandra/index/sai/StorageAttachedIndexBuilder.java
@@ -261,8 +261,9 @@ public class StorageAttachedIndexBuilder extends SecondaryIndexBuilder
     {
         // if per-table files are incomplete or checksum failed during full rebuild.
         IndexDescriptor indexDescriptor = IndexDescriptor.create(sstable);
-        if (!indexDescriptor.isPerSSTableIndexBuildComplete() ||
-            (isFullRebuild && !indexDescriptor.validatePerSSTableComponents(IndexValidation.CHECKSUM)))
+        if (!indexDescriptor.isPerSSTableIndexBuildComplete()
+            || isFullRebuild
+            || !indexDescriptor.validatePerSSTableComponents(IndexValidation.CHECKSUM))
         {
             CountDownLatch latch = CountDownLatch.newCountDownLatch(1);
             if (inProgress.putIfAbsent(sstable, latch) == null)
--- a/src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
+++ b/src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
@@ -411,7 +411,7 @@ public class StorageAttachedIndexGroup implements Index.Group, INotificationCons
     {
         contextManager.clear();
         onSSTableChanged(baseCfs.getLiveSSTables(), Collections.emptySet(), indexes, IndexValidation.NONE);
-        onSSTableChanged(Collections.emptySet(), baseCfs.getLiveSSTables(), indexes, IndexValidation.BASIC);
+        onSSTableChanged(Collections.emptySet(), baseCfs.getLiveSSTables(), indexes, IndexValidation.HEADER_FOOTER);
     }
 
     /**
--- a/src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
+++ b/src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
@@ -250,7 +250,7 @@ public class StorageAttachedIndexGroup implements Index.Group, INotificationCons
             SSTableAddedNotification notice = (SSTableAddedNotification) notification;
 
             // Avoid validation for index files just written following Memtable flush.
-            boolean validate = !notice.memtable().isPresent();
+            IndexValidation validate = notice.memtable().isPresent() ? IndexValidation.NONE : IndexValidation.CHECKSUM;
             onSSTableChanged(Collections.emptySet(), notice.added, indexes, validate);
         }
         else if (notification instanceof SSTableListChangedNotification)
@@ -258,7 +258,7 @@ public class StorageAttachedIndexGroup implements Index.Group, INotificationCons
             SSTableListChangedNotification notice = (SSTableListChangedNotification) notification;
 
             // Avoid validation for index files just written during compaction.
-            onSSTableChanged(notice.removed, notice.added, indexes, false);
+            onSSTableChanged(notice.removed, notice.added, indexes, IndexValidation.NONE);
         }
         else if (notification instanceof MemtableRenewedNotification)
         {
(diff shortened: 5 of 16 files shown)


src/java/org/apache/cassandra/index/sai/IndexContext.java
```diff
@@ -160,7 +160,7 @@ public IndexMetadata getIndexMetadata()
     /**
      * @return A set of SSTables which have attached to them invalid index components.
      */
-    public Collection<SSTableContext> onSSTableChanged(Collection<SSTableReader> oldSSTables, Collection<SSTableContext> newSSTables, boolean validate)
+    public Collection<SSTableContext> onSSTableChanged(Collection<SSTableReader> oldSSTables, Collection<SSTableContext> newSSTables, IndexValidation validate)
```
adelapena: Nit: I'd call the parameter `validation`, instead of the current boolean-style `validate`

src/java/org/apache/cassandra/index/sai/IndexContext.java
```diff
@@ -406,7 +406,7 @@ public String logMessage(String message)
      * @return the indexes that are built on the given SSTables on the left and corrupted indexes'
      * corresponding contexts on the right
      */
-    public Pair<Collection<SSTableIndex>, Collection<SSTableContext>> getBuiltIndexes(Collection<SSTableContext> sstableContexts, boolean validate)
+    public Pair<Collection<SSTableIndex>, Collection<SSTableContext>> getBuiltIndexes(Collection<SSTableContext> sstableContexts, IndexValidation validate)
```
adelapena: Nit: I'd call the parameter `validation`, instead of the current boolean-style `validate`

src/java/org/apache/cassandra/index/sai/SSTableContextManager.java
```diff
@@ -52,7 +52,7 @@ public class SSTableContextManager
      * @return a set of contexts for SSTables with valid per-SSTable components, and a set of
      * SSTables with invalid or missing components
      */
-    public Pair<Set<SSTableContext>, Set<SSTableReader>> update(Collection<SSTableReader> removed, Iterable<SSTableReader> added, boolean validate)
+    public Pair<Set<SSTableContext>, Set<SSTableReader>> update(Collection<SSTableReader> removed, Iterable<SSTableReader> added, IndexValidation validate)
```
adelapena: I'd call the parameter `validation`. Also, the JavaDoc `@param` above is outdated.

src/java/org/apache/cassandra/index/sai/disk/v1/V1OnDiskFormat.java
```diff
@@ -213,6 +211,7 @@ public boolean validatePerColumnIndexComponents(IndexDescriptor indexDescriptor,
                 }
             }
         }
+        long endTime = Clock.Global.currentTimeMillis();
```
adelapena: This seems unused

test/unit/org/apache/cassandra/index/sai/disk/v1/trie/TrieValidationTest.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.cassandra.io.tries.IncrementalDeepTrieWriterPageAware;
+import org.apache.cassandra.utils.bytecomparable.ByteComparable;
+import org.apache.cassandra.utils.bytecomparable.ByteSource;
+import org.apache.lucene.store.IndexInput;
+
+import static org.apache.cassandra.index.sai.disk.v1.trie.TrieTermsDictionaryReader.trieSerializer;
+
+public class TrieValidationTest extends SAIRandomizedTester
+{
+    IndexDescriptor indexDescriptor;
```
adelapena: Nit: can be `private`

test/unit/org/apache/cassandra/index/sai/disk/v1/trie/TrieValidationTest.java
```diff
@@ -0,0 +1,93 @@
...
+    public void testChecksumValidation() throws Throwable
+    {
+        createSimpleTrie(indexDescriptor);
+        try (IndexInput input = indexDescriptor.openPerSSTableInput(IndexComponent.PRIMARY_KEY_TRIE))
+        {
+            SAICodecUtils.validateChecksum(input);
+        }
+    }
+
+    private void createSimpleTrie(IndexDescriptor indexDescriptor) throws Throwable
```
adelapena: Nit: can be `static`

test/unit/org/apache/cassandra/index/sai/disk/v1/trie/TrieValidationTest.java
```diff
@@ -0,0 +1,93 @@
...
+            trieWriter.add(v -> createMultiPart(v, "def", "ghi", "jkl"), 4L);
+            trieWriter.add(v -> UTF8Type.instance.asComparableBytes(UTF8Type.instance.fromString("abcdef"), v), 5L);
+            trieWriter.add(v -> UTF8Type.instance.asComparableBytes(UTF8Type.instance.fromString("abdefg"), v), 6L);
+            trieWriter.add(v -> UTF8Type.instance.asComparableBytes(UTF8Type.instance.fromString("abdfgh"), v), 7L);
+            trieWriter.complete();
+            SAICodecUtils.writeFooter(trieOutput);
+        }
+    }
+
+    private ByteSource createMultiPart(ByteComparable.Version version, String... parts)
```
adelapena: Nit: can be `static`

src/java/org/apache/cassandra/index/sai/disk/io/IndexFileUtils.java
```diff
@@ -89,60 +89,20 @@ static class IncrementalChecksumSequentialWriter extends SequentialWriter implem
...
-
-        @Override
-        public void writeLong(long v) throws IOException
-        {
-            super.writeLong(v);
-            addTochecksum(v, 8);
-        }
-
-        public long getChecksum()
+        public long getChecksum() throws IOException
```
adelapena, line 87: Nit: add `@Override`

src/java/org/apache/cassandra/index/sai/IndexContext.java
pkolaczk: fixed

src/java/org/apache/cassandra/index/sai/SSTableContextManager.java
pkolaczk: fixed

src/java/org/apache/cassandra/index/sai/disk/v1/V1OnDiskFormat.java
pkolaczk: good catch, I forgot to remove it after testing

pkolaczk: fixed

test/unit/org/apache/cassandra/index/sai/disk/v1/trie/TrieValidationTest.java
pkolaczk: fixed

test/unit/org/apache/cassandra/index/sai/disk/v1/trie/TrieValidationTest.java
pkolaczk: fixed

test/unit/org/apache/cassandra/index/sai/disk/v1/trie/TrieValidationTest.java
pkolaczk: fixed

src/java/org/apache/cassandra/index/sai/IndexContext.java
pkolaczk: fixed

adelapena (approved):
Changes look good to me, besides [this missing nit](https://github.com/apache/cassandra/pull/2460#discussion_r1252248315). 

I think now we should rebase and run CI with the config generated by `.circleci/generate_11_and_17.sh -p`, setting `BASE_BRANCH=cep-7-sai` in that script so it picks the right modified tests.

src/java/org/apache/cassandra/index/sai/disk/io/IndexFileUtils.java
pkolaczk: fixed

pkolaczk:
Ok, rebased and squashed review fixups.

adelapena:
It looks great, have you run CI on it?

pkolaczk:
CI workflow: https://app.circleci.com/pipelines/github/pkolaczk/cassandra/31

adelapena:
It seems CI is hitting some new test failures around `NodeRestartTest` and `SnapshotTest`, I haven't looked into them yet.

src/java/org/apache/cassandra/index/sai/disk/io/IndexFileUtils.java
```diff
@@ -77,7 +77,7 @@ public IndexInput openBlockingInput(File file)
 
     public interface ChecksumWriter
     {
-        long getChecksum();
+        long getChecksum() throws IOException;
     }
 
     static class IncrementalChecksumSequentialWriter extends SequentialWriter implements ChecksumWriter
```
maedhroz: I'm actually not sure why we have the `ChecksumWriter` interface at all here. (CC @mike-tr-adamson) Also not sure what's incremental about this now, given we just write it on flush.

There is the class `ChecksummedSequentialWriter` in `o.a.c.io.util`, and that even shares essentially the same `flushData()` method (but it writes to a concrete class called `ChecksumWriter` which has an incremental component as well as its own `DataOutput` to another file).

Perhaps we could address the `flushData()` duplication by having a `SequentialWriter` subclass w/ an abstract method that updates the checksum in a subclass-appropriate way. Then you'd have two subclasses, one strictly in memory for this case, and the other essentially doing what `o.a.c.io.util.ChecksummedSequentialWriter` does now. This might not be the only way to address this. (Naming would be interesting, as usual. `InMemoryChecksumSequentialWriter` is awful, but sort of describes what the one in this class does...)

maedhroz: The idea of bypassing this whole problem by checksumming on flush is great though :D

maedhroz:
> It seems CI is hitting some new test failures around NodeRestartTest and SnapshotTest, I haven't looked into them yet.

@pkolaczk @adelapena I think the failures are due to us calling `validatePerIndexComponents()` and `validatePerSSTableComponents()` in `IndexDescriptor` w/ `IndexValidation.NONE` rather than not at all. The counter injections in `SAITester` depend on this. You could fix by changing them to...

```
protected static final Injections.Counter perSSTableValidationCounter = Injections.newCounter("PerSSTableValidationCounter")
                                                                                  .add(newInvokePoint().onClass(OnDiskFormat.class)
                                                                                                       .onMethod("validatePerSSTableIndexComponents"))
                                                                                  .build();

protected static final Injections.Counter perColumnValidationCounter = Injections.newCounter("PerColumnValidationCounter")
                                                                                 .add(newInvokePoint().onClass(OnDiskFormat.class)
                                                                                                      .onMethod("validatePerColumnIndexComponents"))
                                                                                 .build();
```

After that, `shouldTakeAndRestoreSnapshots()` in `SnapshotTest` will still fail, and that's because the test was always incorrect. Changing line 107 to `assertValidationCount(4, 2);` should properly verify that full rebuild checksums the SSTable-level index components.

Alright, on to the rest of the review. Hopefully that helps :)

src/java/org/apache/cassandra/index/sai/IndexValidation.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Basic header validation, no data validation (fast)
+     */
+    BASIC,
+
+    /**
+     * Full validation with checksumming data (slow)
+     */
+    CHECKSUM;
```
maedhroz:
```suggestion
    CHECKSUM
```

src/java/org/apache/cassandra/index/sai/IndexValidation.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public enum IndexValidation
+{
+    /**
+     * No validation to be performed
+     */
+    NONE,
+
+    /**
+     * Basic header validation, no data validation (fast)
```
maedhroz:
```suggestion
     * Basic header/footer validation, but no data validation (fast)
```

src/java/org/apache/cassandra/index/sai/IndexValidation.java
```diff
@@ -0,0 +1,41 @@
...
+     */
+    BASIC,
```
maedhroz:
```suggestion
    HEADER_FOOTER,
```
nit: Closer to what it actually does?

src/java/org/apache/cassandra/index/sai/IndexValidation.java
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.index.sai;
+
+/**
+ * Whether to validate indexes and how
+ */
```
maedhroz: nit: I would be fine if we just removed the class-level JavaDoc here. The field docs seem to cover what we care about...

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
```diff
@@ -327,28 +328,33 @@ public long sizeOnDiskOfPerIndexComponent(IndexComponent indexComponent, IndexCo
...
+        if (validation != IndexValidation.NONE)
+        {
+            logger.info(logMessage("Validating per-sstable index components using mode " + validation));
+            boolean checksum = validation == IndexValidation.CHECKSUM;
+            return version.onDiskFormat().validatePerSSTableIndexComponents(this, checksum);
+        }
+        else
+        {
+            return true;
+        }
```
maedhroz: nit: It's more or less a matter of preference, but these methods might be clearer w/ just a...

```
if (validation == IndexValidation.NONE)
    return true;
```

...at the start and then the rest w/o a full `if/else`.

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
```diff
@@ -250,15 +250,15 @@ public void handleNotification(INotification notification, Object sender)
             SSTableAddedNotification notice = (SSTableAddedNotification) notification;
 
             // Avoid validation for index files just written following Memtable flush.
-            boolean validate = !notice.memtable().isPresent();
+            IndexValidation validate = notice.memtable().isPresent() ? IndexValidation.NONE : IndexValidation.CHECKSUM;
```
maedhroz, line 253: So this is the crux of the change, correct? (i.e. SSTables added via streaming or import will be checksummed, unlike before, where they would undergo only header/footer validation?)

maedhroz (approved):

maedhroz:
Pushed another branch w/ the test fixes here: https://github.com/maedhroz/cassandra/commits/CASSANDRA-18490-test-fixes

...and a Circle run w/ more hardware here: https://app.circleci.com/pipelines/github/maedhroz/cassandra/728/workflows/3668460f-3857-48b5-9143-28efac884baa

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
maedhroz: One of the general things I looked at while reviewing here was whether the notification order of the different `INotificationConsumer` things matters. `StorageAttachedIndexGroup` and `SecondaryIndexManager` in particular are the interesting ones. Basically...do we checksum the things we should and not the things we shouldn't, given we happen to have SIM registered before the index group? Here's what I've come up with:

1.) When we stream entire SSTables w/ their attached index components (or locally import new SSTables w/ index components already built), SIM handles the notification, but the index build it triggers should be a no-op. Even if it built something it wouldn't register as a full rebuild and no checksumming would be done. Then when `StorageAttachedIndexGroup` handles the notification here, it does the checksum, effectively protecting us from corruption introduced by a streaming source. This all seems reasonable.

2.) When we don't stream an entire SSTable during repair or partial SSTable bootstrapping (or locally import new SSTables w/ index components not already built), SIM handles the notification and blocks to build SSTable-attached index bits. Then (again, this is an accident of our ordering) `StorageAttachedIndexGroup` handles the notification. It sees that there isn't a Memtable and validates the checksum. _This isn't the end of the world, but it would seem unnecessary, given we just build the indexes from the streamed SSTables._ If we knew that all of the added SSTables didn't come with SAI components, we'd bypass checksumming, right?

I'm not sure how exactly we want to address this, or whether we would want to address it in this patch (although that might be nice, since we have it loaded into our working memories), but perhaps we could agree that it's an issue (or not, and that I'm missing something) and go from there?

Two things that come to mind to "fix" this are a.) having more information in the notification or b.) having the index group handle notifications *before* SIM does. (SIM only handles `SSTableAddedNotification` FWIW).

CC @mike-tr-adamson @adelapena

src/java/org/apache/cassandra/index/sai/disk/io/IndexFileUtils.java
pkolaczk: > Perhaps we could address the flushData() duplication by having a SequentialWriter subclass w/ an abstract method that updates the checksum in a subclass-appropriate way.

I don't agree with that. Subclassing, especially subclassing with implementation inheritance / overrides or from abstract base classes, introduces more complexity and often causes way more pain than a few duplicated lines of code. In this particular case, there were several issues caused by `writeXXX` overrides which is a textbook example of inheritance gone bad. If I'm going to do a refactoring, it would be reducing the amount of subclassing, not adding more of it ;)

`ChecksummedSequentialWriter` looks like a class that has too many responsibilities. It computes the checksum *and* writes checksum to a separate component. It looks what we really want is to split those responsibilities, and really have a separate `ChecksummedSequentialWriter` that does only the checksummng part, and a separate optional component/function responsible for writing the checksum to a file. If I refactor code like that, then I could get rid of duplication, because we'd have only one checksumming writer (and no subclasses).

maedhroz: Yeah, it is doing a lot of things. In any case, I like that approach better than what I’ve suggested…

pkolaczk: Just looked at it again and I'm afraid it is not so easy - this `ChecksummedSequentialWriter` actually writes a series of checksums to separate file. So it is kinda doing a different thing than what we want (at the moment) in SAI. SAI is just interested in computing one checksum. Note that each writer class uses even a different structure to compute the checksum. I'm afraid that trying to extract the common logic between those two or trying to blend those two usecases into a single class is going to actually make the code more complex and less readable, even if slghtly less duplicated.

And we'll have to redo this anyways, when we have proper incremental (per block) checksums for SAI - and we really would love to use the *same* checksumming components as C* is using. So I'm for not touching this now, except maybe finding a better name for this SAI checksummed writer.

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
pkolaczk: ok

src/java/org/apache/cassandra/index/sai/IndexValidation.java
pkolaczk: fixed

src/java/org/apache/cassandra/index/sai/disk/io/IndexFileUtils.java
pkolaczk: I simplified `flushData` and removed the redundant one-impl interface and removed `Incremental` from the name.

pkolaczk:
> After that, shouldTakeAndRestoreSnapshots() in SnapshotTest will still fail, and that's because the test was always incorrect. Changing line 107 to assertValidationCount(4, 2); should properly verify that full rebuild checksums the SSTable-level index components.

I'm not convinced the test is incorrect. The comment says that the index shouldn't validate after rebuild / compaction and that makes IMHO sense. Why validate an index we just wrote? 

@mike-tr-adamson ^

pkolaczk:
So the test looks correct. I think the problem was that the the test requests rebuilding the index and we validated the old index in this situation, when we shouldn't be. The fact we changed the validation counters to count both checksum and non-checksum validations simply exposed that bug.

pkolaczk:
Launched another CI run: https://app.circleci.com/pipelines/github/pkolaczk/cassandra/39/workflows/d3df886e-4ac1-440d-8c47-d23bc08b4e5a

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
mike-tr-adamson: We have been looking at this in some detail today to try and establish the best way to correct it.

From our perspective, the `SSTableAddedNotification` is handled in 2 places. It is handled by the `SecondaryIndexManager` that with invoke a `fullRebuild = false` index build on the added SSTables, and it is handled by the `StorageAttachedIndexGroup` that will validate and register the SSTables. 

The problem with the current implementation is:
- The invoked index build will rebuild the per-column index components for `fullRebuild = false` even if the components are present and valid. We do, though, only build the per-SSTable components if they are invalid, in this case.
- The SAIG handler will validate the components even though they have been already validated by the build.

What should be noted here is that the SIM will only invoke an index build if a memtable is not present in the notification, and the SAIG currently only validates the index components if the memtable isn't present but still registers the components in either case.

We feel that the following changes could be made to improve this:
1. The index build should only build the per-column indexes if `fullRebuild=true`, the index doesn't exist, or the index is invalid. In either case, the SSTables must be registered (on build completion or immediately).
2. The SAIG should only register the SSTables if the memtable is present and shouldn't validate the components because this is the result of a memtable flush.

This will require some rework of the index builder, to make sure that the SSTables are always registered correctly and, as such, we would prefer that this was done on a separate ticket from this one. Which phase that ticket goes into is up for discussion. The downsides of the current implementation are that we can rebuild indexes that already have valid components and we can end up validating components twice. 

WDYT?

maedhroz: @mike-tr-adamson @pkolaczk I'm okay w/ doing this in another Jira. There's enough risk to make sure we focus on it specifically, and after this patch we'll only be doing something sub-optimally, not incorrectly. Now...in solution space...

> The index build should only build the per-column indexes if fullRebuild=true, the index doesn't exist, or the index is invalid. In either case, the SSTables must be registered (on build completion or immediately).

I don't see any problem w/ this, although I'm not sure I understand the problem in the streaming context.

> The invoked index build will rebuild the per-column index components for fullRebuild = false even if the components are present and valid. We do, though, only build the per-SSTable components if they are invalid, in this case.

Right after streaming, no index components for the new SSTables are present, right?

> The SAIG should only register the SSTables if the memtable is present and shouldn't validate the components because this is the result of a memtable flush.

`StorageAttachedIndexGroup` registering and validating newly streamed entire index components is what makes this patch work, right? If we stopped doing that, who would register/checksum when the `SIM`-triggered build is a no-op?

maedhroz: See [CASSANDRA-18653](https://issues.apache.org/jira/browse/CASSANDRA-18653)

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexBuilder.java
```diff
@@ -261,8 +261,9 @@ private CountDownLatch shouldWritePerSSTableFiles(SSTableReader sstable)
     {
         // if per-table files are incomplete or checksum failed during full rebuild.
```
maedhroz, line 262: nit: Comment needs to change slightly now, something like...

```
// if per-SSTable files are incomplete, this is a full rebuild, or they are complete but checksum failed
```

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexBuilder.java
```diff
@@ -261,8 +261,9 @@ private CountDownLatch shouldWritePerSSTableFiles(SSTableReader sstable)
...
         IndexDescriptor indexDescriptor = IndexDescriptor.create(sstable);
-        if (!indexDescriptor.isPerSSTableIndexBuildComplete() ||
-            (isFullRebuild && !indexDescriptor.validatePerSSTableComponentsChecksum()))
+        if (!indexDescriptor.isPerSSTableIndexBuildComplete()
+            || isFullRebuild
```
maedhroz, line 265: When we initially wrote this, we couldn't think of any cases where the SSTable-level indexes would need to be rebuilt if existing and valid. The column-level bits are typically what need to change, say if you slightly changed filtering/analysis on a text field. However, I'm not opposed to this if you think it's clearer that "full rebuild" really means "full" rebuild :)

maedhroz:
Committed as https://github.com/apache/cassandra/commit/9057ff3e6dad89b0197d83f404a9b0fca544e0f7

PR #2462, opened 2023-07-03
CASSANDRA-18646 Azure snitch

smiklosovic:
 
```

```

### DIFF
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -1316,6 +1316,11 @@ slow_query_log_timeout: 500ms
 #    Snitch for getting dc and rack of a node from metadata service of Alibaba cloud.
 #    This snitch that assumes an ECS region is a DC and an ECS availability_zone is a rack.
 #
+# AzureSnitch:
+#    Gets datacenter from 'location' and rack from 'zone' fields of 'compute' object
+#    from instance metadata service. If the availability zone is not enabled, it will use the fault
+#    domain and get its respective value.
+#
 # CloudstackSnitch:
 #    A snitch that assumes a Cloudstack Zone follows the typical convention 
 #    country-location-az and uses a country/location tuple as a datacenter
--- a/src/java/org/apache/cassandra/locator/AzureSnitch.java
+++ b/src/java/org/apache/cassandra/locator/AzureSnitch.java
@@ -1,0 +1,102 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.locator;
+
+import java.io.IOException;
+
+import com.google.common.collect.ImmutableMap;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import org.apache.cassandra.locator.AbstractCloudMetadataServiceConnector.DefaultCloudMetadataServiceConnector;
+import org.apache.cassandra.utils.JsonUtils;
+import org.apache.cassandra.utils.Pair;
+
+import static java.lang.String.format;
+import static org.apache.cassandra.locator.AbstractCloudMetadataServiceConnector.METADATA_URL_PROPERTY;
+
+/**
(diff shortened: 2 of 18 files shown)


src/java/org/apache/cassandra/locator/AzureSnitch.java
```diff
@@ -0,0 +1,190 @@
...
+            return format("%s{%s=%s,%s=%s}",
+                          AzureCloudConnector.class.getName(),
+                          METADATA_SERVICE_PROPERTY_KEY,
+                          metadataServiceUrl,
+                          API_VERSION_PROPERTY_KEY,
+                          apiVersion);
+        }
+    }
+
+    public String getRack(InetAddressAndPort endpoint)
```
jacek-lewandowski: This method, and `getDatacenter` method, are identical in all snitch implementations. The only difference is that it returns a value of a field `rack` and, say in EC2, it returns value of a field `ec2zone`. Could you extract simple method like `getRackInternal() { return rack }`, define it abstract in `AbstractNetworkTopologySnitch` and move the implementation of `getRack(endpoint)` to `AbstractNetworkTopologySnitch`? You can do the same for `getDatacenter(endpoint)` method, and do that for all the snitch implementations.

src/java/org/apache/cassandra/locator/AzureSnitch.java
```diff
@@ -0,0 +1,109 @@
...
+        super(connector, properties, resolveDcAndRack(connector));
+    }
+
+    private static Pair<String, String> resolveDcAndRack(AzureCloudConnector connector) throws IOException
+    {
+        String response = connector.getResponse();
+        JsonNode jsonNode = JsonUtils.JSON_OBJECT_MAPPER.readTree(response);
+
+        JsonNode location = jsonNode.get("location");
+        JsonNode zone = jsonNode.get("zone");
```
xgerman, line 73: Azure being azure there is another way to distribute nodes called fault domains (see https://learn.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-manage-fault-domains) so if zone is empty you need to check `platformFaultDomain` https://learn.microsoft.com/en-us/azure/virtual-machines/instance-metadata-service?tabs=linux

src/java/org/apache/cassandra/locator/AzureSnitch.java
```diff
@@ -0,0 +1,109 @@
...
+
+        String datacenter;
+        String rack;
+
+        if (location == null || location.isNull() || location.asText().isEmpty())
+            datacenter = DEFAULT_DC;
+        else
+            datacenter = location.asText();
```
xgerman, line 82: not sure if having multiple data centers in the same location is a goal. If so, this would not work for that scenario.

smiklosovic: @xgerman I am not sure what you mean by this. Can you elaborate?

xgerman: Assume you have two data centers DC1 and DC2 both in the same region, say useast2. Then `datacenter` would be for both useast2...

smiklosovic: resolved privately, in practice one is supposed to use dc_suffix for this kind of scenario

conf/cassandra.yaml
```diff
@@ -1302,6 +1302,10 @@ slow_query_log_timeout: 500ms
 #    Snitch for getting dc and rack of a node from metadata service of Alibaba cloud.
 #    This snitch that assumes an ECS region is a DC and an ECS availability_zone is a rack.
 #
+# AzureSnitch:
+#    Gets datacenter from 'location' and rack from 'zone' fields of 'compute' object
+#    from instance metadata service.
```
xgerman: If availability zone is not enabled it will use fault domains and get the respective value from

xgerman (approved):

src/java/org/apache/cassandra/locator/AzureSnitch.java
```diff
@@ -0,0 +1,128 @@
...
+        JsonNode zone = jsonNode.get("zone");
+        JsonNode platformFaultDomain = jsonNode.get("platformFaultDomain");
+
+        String datacenter;
+        String rack;
+
+        if (location == null || location.isNull() || location.asText().isEmpty())
+            datacenter = DEFAULT_DC;
+        else
+        {
```
jacek-lewandowski: nit: braces seems redundant

smiklosovic: done

conf/cassandra.yaml
```diff
@@ -1302,6 +1302,11 @@ slow_query_log_timeout: 500ms
 #    Snitch for getting dc and rack of a node from metadata service of Alibaba cloud.
 #    This snitch that assumes an ECS region is a DC and an ECS availability_zone is a rack.
 #
+# AzureSnitch:
```
jacek-lewandowski, line 1319: nit: 
```
# AzureSnitch:
#    Gets datacenter from 'location' and rack from 'zone' fields of 'compute' object
#    from instance metadata service. If the availability zone is not enabled, it will use the fault
#    domain and get its respective value.
#
```

jacek-lewandowski (approved):
+1, please make sure CI is passing before merge

PR #2464, opened 2023-07-05
Ldtdeser trunk CASSANDRA-18648

bereng:
 
```

```


bereng:
CI [j8](https://app.circleci.com/pipelines/github/bereng/cassandra/1006/workflows/acc08201-3ee5-49a3-9689-d1f0e79bf7ab): All green except for 2 known flakies
CI [j11](https://app.circleci.com/pipelines/github/bereng/cassandra/1006/workflows/61923e73-5aa6-4882-b622-33562176060d): All green except for a known flaky. The upgrade failures are aligned to a trunk reference run I triggered as [baseline](https://app.circleci.com/pipelines/github/bereng/cassandra/1007/workflows/46c54b5c-bc00-4db8-863b-14ceaa331d95)

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,100 @@ public static Serializer getSerializer(Version version)
...
+     * Since mfda is micros since the Unix Epoch with 7 bytes we can encode 2^56 ~= 1142 years. That leaves 
+     * 1 free byte we can use to store flags for sentinel values with some space saving.
+     *
+     * Currently only a flag for the LIVE ldt is being used.
+     */
     public static class Serializer implements ISerializer<DeletionTime>
     {
+        private final static int IS_LIVE_DELETION = 0x01;
+        // Long.MAX_VALUE sentinel value equivalent in 7bytes
+        public final static long MAX_MFDA = 0xffffffffffffffL;
```
jakubzytka: wouldn't `(1L << 56) - 1` be simpler and easier to read? wdyt?
or
(1L << (7 * 8)) - 1

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,100 @@ public static Serializer getSerializer(Version version)
...
         public void serialize(DeletionTime delTime, DataOutputPlus out) throws IOException
         {
-            out.writeInt(delTime.localDeletionTimeUnsignedInteger);
-            out.writeLong(delTime.markedForDeleteAt());
+            if (delTime.equals(LIVE))
+                out.writeByte(IS_LIVE_DELETION);
+            else
+            {
+                if (delTime.markedForDeleteAt() > MAX_MFDA && delTime.markedForDeleteAt() != Long.MAX_VALUE)
+                    throw new IOException("Value too high for markForDeleteAt to encode in 7 bytes: " + delTime.markedForDeleteAt());
```
jakubzytka: out of curiosity: why do we throw instead of changing the value to max?

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,100 @@ public static Serializer getSerializer(Version version)
...
+                return 1;
+            else
+                // 1 for the flags, 7 for mfda and 4 for ldt
+                return TypeSizes.INT_SIZE
+                       + TypeSizes.LONG_SIZE;
+        }
+
+        private long sevenBytesToMFDA(int bytes4, int bytes2, int bytes1)
+        {
+                long mfda = bytes4 & 0xFFFFFFFFL;
```
jakubzytka: it's only 4 bytes here, but still `(1L << 32) - 1` might be preferable

test/microbench/org/apache/cassandra/test/microbench/DeletionTimeDeSerBench.java
```diff
@@ -0,0 +1,363 @@
...
+                {
+                    throw new RuntimeException(e);
+                }
+
+            default:
+                throw new IllegalStateException("Unknown backing media for ser.");
+        }
+    }
+
+    private static List<DeletionTime> generateDTs(int livePC)
```
jakubzytka: could we make it `livePercentage` and use it like percentage (i.e. `70` instead of `7` etc)

test/microbench/org/apache/cassandra/test/microbench/DeletionTimeDeSerBench.java
```diff
@@ -0,0 +1,363 @@
...
+    @Benchmark
+    public void testE2ESerializeDT(final Blackhole bh) throws IOException
+    {
+        ByteBuffer buffer = serMMap;
+        Serializer serializer = getSerializer(sstableParam);
+        List<DeletionTime> dts = getDTs(liveDTPcParam);
+        
+        try(DataOutputStreamPlus out = getSerOut(diskRAMParam, serMMap))
+        {
+            for (int i = 0, m = dts.size(); i < m; i++)
```
jakubzytka, line 109: is there a jmh-related reason to use this kind of iteration instead of plain
```
for (DeletionTime dt: dts)
   serializer.serialize(dt, out);
```
?

If there is then let's make it explicit that `get(..)` is O(1) - i.e. use `ArrayList` explicitly instead of `List`.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Yep yep yep. I like `(1L << (7 * 8)) - 1` better

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: `markedForDeleteAt` is the timestamp of a deletion. I see it as _data_. So I want to reject invalid _data_ and I don't want to silently modify _data_. Also sbdy would have to explain me why we are getting timestamps 2K years in the future and past year 2106, which is a hard limit atm bc of localDeletionTime max value. Makes sense?

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Do you mind I leave it as it is to be the same format as the as the next 2 lines? being only 4 bytes it's no eye offensive as the `MAX_MFDA` one

test/microbench/org/apache/cassandra/test/microbench/DeletionTimeDeSerBench.java
bereng: Sure thing!

test/microbench/org/apache/cassandra/test/microbench/DeletionTimeDeSerBench.java
bereng: The for loop is the cheapest I know. Others will exercise the iterators machinery and opens the door to possible hidden costs within the iterators methods.

Yep we can.

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,100 @@ public static Serializer getSerializer(Version version)
...
-            out.writeInt(delTime.localDeletionTimeUnsignedInteger);
-            out.writeLong(delTime.markedForDeleteAt());
+            if (delTime.equals(LIVE))
+                out.writeByte(IS_LIVE_DELETION);
+            else
+            {
+                if (delTime.markedForDeleteAt() > MAX_MFDA && delTime.markedForDeleteAt() != Long.MAX_VALUE)
+                    throw new IOException("Value too high for markForDeleteAt to encode in 7 bytes: " + delTime.markedForDeleteAt());
+                // The flags byte is all zeros atm here, so we can write a long directly
+                out.writeLong(delTime.markedForDeleteAt() == Long.MAX_VALUE ? MAX_MFDA : delTime.markedForDeleteAt());
```
jacek-lewandowski: why don't you store this information as a separate flag? say, if flag 0x2 is set, don't have to store mfda at all

src/java/org/apache/cassandra/db/DeletionTime.java
jacek-lewandowski: nit: either 0xff.. or 0xFF..

src/java/org/apache/cassandra/db/DeletionTime.java
jacek-lewandowski: I'm wondering if it should be verified earlier - that is, when the user provides a mutation timestamp. There is also a guardrail for that: `maximum_timestamp_fail_threshold` - maybe it deserves a default value?

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Right. I plan to javadoc the whole sstable format and hope to find more opportunities to use the flags so I was saving them. Some ideas are the Long.MAX_VALUE, signaling some Vint to some epoch, etc So I was trying to save bits to that purpose.

Given any future change will imply another sstable format change anyway we can use that bit now for sure.

bereng: On the other hand though you still have to store `localDeletionTime`. That starts muddying the waters a bit with the straightforward simple algorithm we have right now and Long.MAX_VALUE, during my tests, didn't come up except for junits. Maybe we want to leave things as they are. I am all ears for opinions.

bereng:
Note to self. Should add CHANGES.txt entry and notes on upgrades

src/java/org/apache/cassandra/db/DeletionTime.java
jacek-lewandowski: I thought that reducing the number of written bytes in certain cases is the reason for introducing a flag?

bereng: Indeed. But if that is going to be a rare corner case vs added complexity... I guess we'll have to jmh a bit if you find it worth it.

src/java/org/apache/cassandra/db/DeletionTime.java
jakubzytka: Even if we verify it earlier it doesn't solve all our problems.
`MAX_MFDA` is not an implementation detail of serialization but, whether we like it or not, an externally visible limitation imposed on client-provided data.

If we start throwing, I'm worried that after the upgrade suddenly e.g. compactions start failing (because there are writes with too big MFDA already).

Thus I was wondering if stating the limitation as "all MFDA greater than MAX_MFDA will be converted to MAX_MFDA" wouldn't be a better choice. (IIUC currently it is "all MFDA greater than MAX_MFDA will break your at times, and you'd better overwrite it with smaller values")

bereng: ^Icwym, But a MFDA 2000 years into the future shouldn't happen and we should never hit that throws. We should never have an MFDA for the year 4200. The current codebase is dead in year 2106 already.

Let me ping @driftx just to double check. Brandon apologies for the ping, not trying to loop you into the review, but do you think it's reasonable to never expect a MFDA that high. Or can you think of any scenarios where this could be a problem?

bereng: I guess a customer issuing a DELETE using a timestamp that high is the only possibility :thinking: We'd have to add a note and either fail or change the value as Kuba suggests...

driftx: I think 2000 years is plenty reasonable.  Anyone issuing a delete that far into the future is almost certainly making a mistake...giving maximum_timestamp_fail_threshold a default value makes sense too I think, but we should float that on the ML.

bereng: Yeah you're right. That's my exact doubt. Whether to consider that a bug/mistake and correct it rejecting it or to be more accommodating with a cap. ML is the place +1

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Fixed in the latest commit already. Resolving but feel free to reopen if needed.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Email sent

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Right you are

test/microbench/org/apache/cassandra/test/microbench/DeletionTimeDeSerBench.java
bereng: Fixed in the latest commit already. Resolving but feel free to reopen if needed.

test/microbench/org/apache/cassandra/test/microbench/DeletionTimeDeSerBench.java
bereng: We talked offline and some jmh confirms the plain for seems to perform better. Fixed in the latest commit already. Resolving but feel free to reopen if needed.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: jmh looked good so you'll find the changes in the latest commit :-)

```
     [java] Benchmark                                    (diskRAMParam)  (liveDTPcParam)  (sstableParam)  Mode  Cnt        Score        Error  Units
     [java] DeletionTimeDeSerBench.testE2EDeSerializeDT             RAM         70PcLive              NC  avgt   15   667672.766 ±   9373.015  ns/op
     [java] DeletionTimeDeSerBench.testE2EDeSerializeDT             RAM         70PcLive              OA  avgt   15   664025.516 ±   9940.654  ns/op
     [java] DeletionTimeDeSerBench.testE2EDeSerializeDT             RAM         30PcLive              NC  avgt   15   962307.068 ±  15941.703  ns/op
     [java] DeletionTimeDeSerBench.testE2EDeSerializeDT             RAM         30PcLive              OA  avgt   15   971137.196 ±  15428.975  ns/op
     [java] DeletionTimeDeSerBench.testE2EDeSerializeDT            Disk         70PcLive              NC  avgt   15  1392247.306 ±  82162.975  ns/op
     [java] DeletionTimeDeSerBench.testE2EDeSerializeDT            Disk         70PcLive              OA  avgt   15  1044635.877 ±  20614.003  ns/op
     [java] DeletionTimeDeSerBench.testE2EDeSerializeDT            Disk         30PcLive              NC  avgt   15  1719827.358 ± 117661.536  ns/op
     [java] DeletionTimeDeSerBench.testE2EDeSerializeDT            Disk         30PcLive              OA  avgt   15  1507395.582 ±  36998.279  ns/op
     [java] DeletionTimeDeSerBench.testE2ESerializeDT               RAM         70PcLive              NC  avgt   15   332001.653 ±   2247.707  ns/op
     [java] DeletionTimeDeSerBench.testE2ESerializeDT               RAM         70PcLive              OA  avgt   15   337418.362 ±   4235.931  ns/op
     [java] DeletionTimeDeSerBench.testE2ESerializeDT               RAM         30PcLive              NC  avgt   15   460708.379 ±   3550.575  ns/op
     [java] DeletionTimeDeSerBench.testE2ESerializeDT               RAM         30PcLive              OA  avgt   15   498494.239 ±   5669.185  ns/op
     [java] DeletionTimeDeSerBench.testE2ESerializeDT              Disk         70PcLive              NC  avgt   15  1688473.271 ± 255780.071  ns/op
     [java] DeletionTimeDeSerBench.testE2ESerializeDT              Disk         70PcLive              OA  avgt   15   832362.512 ±  91935.451  ns/op
     [java] DeletionTimeDeSerBench.testE2ESerializeDT              Disk         30PcLive              NC  avgt   15  1765952.897 ± 172248.680  ns/op
     [java] DeletionTimeDeSerBench.testE2ESerializeDT              Disk         30PcLive              OA  avgt   15  1326467.617 ± 139599.811  ns/op
```

And with a [green](https://app.circleci.com/pipelines/github/bereng/cassandra/1015/workflows/cf54c599-6cfe-4225-a310-c280fe8cebfc) CI

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,126 @@ public static Serializer getSerializer(Version version)
...
+        private final static int IS_LIVE_DELETION = 0x01;
+        private final static int IS_MAX_MFDA = 0x02;
+        // Long.MAX_VALUE sentinel value equivalent in 7bytes
+        public final static long MAX_MFDA = (1L << (7 * 8)) - 1;
+
         public void serialize(DeletionTime delTime, DataOutputPlus out) throws IOException
         {
-            out.writeInt(delTime.localDeletionTimeUnsignedInteger);
-            out.writeLong(delTime.markedForDeleteAt());
+            if (delTime.equals(LIVE))
```
jacek-lewandowski: this is very nit, don't bother if you don't want - should it be enough to just check `delTime == LIVE`?

pkolaczk:
What are the situations when we need both MFDA and LDT at the same time?
I feel that we need both concepts, but not at the same time.

When the cell is live, we don't know when it will be deleted, so why store LDT?
When the cell is dead, we don't need to store MFDA anymore because it was logically deleted already.
So if they are a proper union, then we can save up even more bytes.


What am I missing?

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,126 @@ public static Serializer getSerializer(Version version)
...
+                out.writeByte(IS_LIVE_DELETION);
+            else if (delTime.markedForDeleteAt() == Long.MAX_VALUE)
+            {
+                out.writeByte(IS_MAX_MFDA);
+                out.writeInt(delTime.localDeletionTimeUnsignedInteger);
+            }
+            else
+            {
+                if (delTime.markedForDeleteAt() > MAX_MFDA && delTime.markedForDeleteAt() != Long.MAX_VALUE)
+                    throw new IOException("Value too high for markForDeleteAt to encode in 7 bytes: " + delTime.markedForDeleteAt());
```
jacek-lewandowski: nit - perhaps a message explicitly saying the limit, maybe even translated to a real date would be more useful for admins?

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,126 @@ public static Serializer getSerializer(Version version)
...
+                // The flags byte is all zeros atm here, so we can write a long directly
+                out.writeLong(delTime.markedForDeleteAt() == Long.MAX_VALUE ? MAX_MFDA : delTime.markedForDeleteAt());
```
jacek-lewandowski: this case is already handled, so condition is not needed right?

jacek-lewandowski:
@pkolaczk 

> When the cell is live, we don't know when it will be deleted, so why store LDT?

~~we don't - we store only a flag that the cell is live~~

> When the cell is dead, we don't need to store MFDA anymore because it was logically deleted already

~~I need to check, but if deletion clears MFDA, the second flag comes into play; otherwise maybe we should clear MFDA along with deletion~~

bereng:
I don't think @pkolaczk is asking that but:
- If a cell has been deleted why store ldt? mfda should be enough
- if a cell has become dead why store the mdfa? ldt should be enough

I didn't know yesterday but then I remembered gc_seconds is a configurable parameter. It can change per table and through time so it needs to be recorded somehow. If that were the case we could store ldt as a delta in seconds from mfda and lift the 2106 limit... :thinking:

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Fixed in review commits. Resolving to clear the PR a bit but feel free to reopen if you disagree.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Fixed in review commits and I see you already commented on it. Resolving to clear the PR a bit but feel free to reopen if you disagree.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Yeah this is my OCD kicking in. If we ever changed DeletionTime in a way we want _content_ comparison, overload `equals`.... I prefer to leave it as it is.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Right

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Agreed. But let's wait see how the mail to the ML resolves first and then we can shoot for the right solution.

src/java/org/apache/cassandra/db/DeletionTime.java
jacek-lewandowski: then, maybe add `if (o == this) return true` to `DeletionTime.equals`. Checking `instanceof` is much less efficient

jacek-lewandowski:
right, my comment was off

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Ok I'll do `==` then

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: If I have gathered all opinions correctly despite the ML email getting little response:
- I added the guardrail default + a test
- I've left it to reject

I still have torn thoughts about either rejecting or capping... :thinking:

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Added the real date.

bereng:
@pkolaczk I think the question you raised is legit. But it touches on one of the corner stones of the codebase. I am not comfortable touching that so close to a release. I think the change is big enough it deserves it's own ticket. Are you ok we address that separately?

pkolaczk:
> Are you ok we address that separately?

@bereng sure, no problem. For the most common case - live cells the timestamps aren't stored at all so this is already a very good improvement.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Fixed in latest commit. Feel free to reopen if needed.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Fixed in latest commit. Feel free to reopen if needed.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Fixed in latest commit. Feel free to reopen if needed.

src/java/org/apache/cassandra/db/DeletionTime.java
blambov: Is `Long.MAX_VALUE / MAX_MFDA` something that is supposed to appear at all as a deletion time? I do not think so, and I would personally prefer to keep it short and simple and remove the `MAX_MFDA` special case.

src/java/org/apache/cassandra/db/DeletionTime.java
blambov: We have another option here. Negative timestamps are not valid, so we could encode `LIVE` as a `0xFF` byte (leaving any other value above `0x7F` for other tasks later) and still correctly store any previously-valid timestamp up to `Long.MAX_VALUE`.

Because the only thing we really want to do here is save 11 bytes for live data, this would still solve the problem but will avoid having to make these decisions.

Alternatively, we could save one byte always by using the sign of the 7th byte for the `LIVE` marker, reducing the permitted values to 1000 years.

blambov:
AFAIK both local and markedFor times are needed and serve different purposes. They should, however, be close to each other normally (i.e. local should be ~`markedFor/1000`). How about, as an additional space saving measure, storing `local - markedFor / 1000` as a signed vint instead of the four-byte int for local? (Note: the vint may become up to 6 bytes in odd cases, but that's something I'd happily trade for having 1 most of the time.)

I do not insist on this, saving the bytes for the `LIVE` encoding is a sufficiently good improvement.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: It is. I found it through some test failures. I can revisit though...

bereng:
@blambov thx for the suggestion. Yes a Vint to mfda, to some synthetic sstable epoch, storing gc_grace_sedonds instead etc. Are all alternatives I want to look at. But first I want to javadoc the whole OA sstable format for everyone's benefit and to see if there are more opportunities around as I don't know this code.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Right lets not use the full byte until we have a justification for that and avoid that user restriction. We'll need an sstable format change when we get there anyway and we might have better ideas when we revisit the ldt serialization given the recent suggestions.

I have rebase and pushed, let me know if iiuc. Thx.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: The latest push supports the full Long range.

bereng:
I have rebased and pushed a version whose only change is that we only use the sign bit to encode LIVE DTs. That allows us to support the full Long range and imposes no restrictions on the users. We can revisit this and use the full byte when we have a justification in the future, have a better opinion on how to improve the ldt serialization, etc.

Ci is [green](https://app.circleci.com/pipelines/github/bereng/cassandra/1025/workflows/90d32ab3-bdd3-43c7-a439-5a34e87bd7f2) with only known flakies.
Jmh numbers still look very good.

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,95 @@ public static Serializer getSerializer(Version version)
...
-    // Serializer for Usigned Integer ldt
+    /* Serializer for Usigned Integer ldt
+     *
+     * ldt is encoded as a uint in seconds since unix epoch, it can go up o 2106-02-07T06:28:13+00:00 only. 
+     * mfda is a positive timestamp. We use the sign bit to enconde LIVE DeletionTimes
+     */
     public static class Serializer implements ISerializer<DeletionTime>
     {
+        // We use the sign bit to signal LIVE DeletionTimes
+        private final static int IS_LIVE_DELETION = 0b1000_0000;
```
blambov, line 207: Nit: I would prefer 0xFF as the leading byte for `LIVE`. This leaves more space if (for whatever reason) we need unsigned values or encode something else with the top bit/bits.

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,95 @@ public static Serializer getSerializer(Version version)
...
-            return mfda == Long.MIN_VALUE && localDeletionTimeUnsignedInteger == Cell.NO_DELETION_TIME_UNSIGNED_INTEGER
-                   ? LIVE
-                   : new DeletionTime(mfda, localDeletionTimeUnsignedInteger);
+            int flags = buf.get(offset);
+            if ((flags & IS_LIVE_DELETION) != 0)
+                return LIVE;
+            else
+            {
+                // Read the remaining 7 bytes
+                int bytes4 = buf.getInt(offset + 1);
```
blambov: You can safely do `buf.getLong(offset)` here, without the `readBytesToMFDA` call.

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,95 @@ public static Serializer getSerializer(Version version)
...
-            return mfda == Long.MIN_VALUE && localDeletionTimeUnsignedInteger == Cell.NO_DELETION_TIME_UNSIGNED_INTEGER
-                 ? LIVE
-                 : new DeletionTime(mfda, localDeletionTimeUnsignedInteger);
+            int flags = in.readByte();
+            if ((flags & IS_LIVE_DELETION) != 0)
+                return LIVE;
+            else
+            {
+                // Read the remaining 7 bytes
+                int bytes4 = in.readInt();
```
blambov: Nit: I would do the reading in the opposite order: second byte, second word, second dword. This way if the long was aligned, so are the partial reads -- though it does not matter significantly.

test/microbench/org/apache/cassandra/test/microbench/DeletionTimeDeSerBench.java
```diff
@@ -0,0 +1,366 @@
...
+        if (sstableParam == oaSstableParam)
+        {
+            for (int i = 0, m = dts.size(); i < m; i++)
+            {
+                buffer.clear();
+                if (dts.get(i).equals(DeletionTime.LIVE))
+                    buffer.put((byte) 0b1000_0000);
+                else
+                {
+                    buffer.putLong(Math.abs(random.nextLong()));
```
blambov, line 166: This can in theory be MIN_VALUE, because `abs(MIN_VALUE) == MIN_VALUE`. I would do `random.nextLong() & Long.MAX_VALUE`.
Similarly below.

blambov (approved):
LGTM.

You will probably need to regenerate the legacy files after rebasing because of a new flag from CASSANDRA-18397.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Atm this is a byte we're not using at all but for one bit. Any future change to that would imply a new sstable format change where we'd free to redo things from scratch. But if this is sounding better to you I am happy to change it :-)

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Good catch I missed.

test/microbench/org/apache/cassandra/test/microbench/DeletionTimeDeSerBench.java
bereng: Done in latest commit

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Done in latest commit

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: I ended up reverting to `0b1000_0000`. With 0xFF the bitwise `&` operator was obviously catching all values of mfda with anything above the 56th bit. That means adding another `&` operation or affecting readability. Given you marked it as a nit and that we'd have to do a new sstable format if we wanted to change that I thought it would be ok.

bereng:
@jacek-lewandowski @pkolaczk this is ready for (hopefully) final review. Then I'll rebase and re-run CI.

src/java/org/apache/cassandra/db/DeletionTime.java
jacek-lewandowski: just a question - if `0b1000_0000` is used as a flag indicator in the future, shouldn't `LIVE` flag be encoded as `0b1000_0001`? If LIVE's presence is encoded as all zeros, it can be weird to encode the lack of it

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,90 @@ public static Serializer getSerializer(Version version)
...
-            return mfda == Long.MIN_VALUE && localDeletionTimeUnsignedInteger == Cell.NO_DELETION_TIME_UNSIGNED_INTEGER
-                   ? LIVE
-                   : new DeletionTime(mfda, localDeletionTimeUnsignedInteger);
+            int flags = buf.get(offset);
+            if ((flags & IS_LIVE_DELETION) != 0)
+                return LIVE;
+            else
+            {
+                long mfda = buf.getLong(offset);
+                int localDeletionTimeUnsignedInteger = buf.getInt(offset + 8);
```
jacek-lewandowski: super nit: be consistent - `getInt(offset + TypeSizes.LONG_SIZE)` :)

jacek-lewandowski (approved):
+1

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: The `1` in `0b1000_0000` is what marks the presence of LIVE. The rest of the zeros are not used at all and have no meaning. Makes sense? I am very bad at explaining things sometimes. Also in the future if we change this it will come with an sstable format change for compatibility reasons afaik.

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: ^Yes yes and yes. I hate magic numbers.

src/java/org/apache/cassandra/db/DeletionTime.java
jacek-lewandowski: I was under the impression that the `1` in `0b1000_0000` should be a flags byte presence marker - so it denotes that the remaining 7 bits should be interpreted as flags. Then, LIVE is one of those flags. Perhaps it does not makes much sense because it is hardly to imagine what could be stored there 🤷🏻‍♂️ Anyway, I've already approved so you do not have to bother

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,90 @@ public static Serializer getSerializer(Version version)
...
 
         public DeletionTime deserialize(DataInputPlus in) throws IOException
         {
-            int localDeletionTimeUnsignedInteger = in.readInt();
-            long mfda = in.readLong();
-            return mfda == Long.MIN_VALUE && localDeletionTimeUnsignedInteger == Cell.NO_DELETION_TIME_UNSIGNED_INTEGER
-                 ? LIVE
-                 : new DeletionTime(mfda, localDeletionTimeUnsignedInteger);
+            int flags = in.readByte();
+            if ((flags & IS_LIVE_DELETION) != 0)
```
blambov, line 224: In the context of possible corruption (I'm looking at CASSANDRA-18676), this version of the check may be opening us up to problems.

This should be something like:
```
if ((flags & SPECIAL) != 0)
{
    assert flags == LIVE_DELETION : "Invalid deletionTime lead byte " + flags;
    return LIVE;
}
```
possibly replacing the assertion with some other error indicating corruption.

bereng: Right icwym as I have also been looking into [CASSANDRA-18676](https://issues.apache.org/jira/browse/CASSANDRA-18676). Should throwing a `CorruptSSTableException` be enough here or should we merge and solve this issue in 18676 better?

bereng:
Rebased to trunk. CI with only the [expected failures](https://app.circleci.com/pipelines/github/bereng/cassandra/1028/workflows/347d7668-7aa9-41d4-a2ee-819226d5a013) after the jdk8 drop.

The resolution of 18676 will inform how to solve the corruption detection.

bereng:
Added sstable corruption detection and CI is [green](https://app.circleci.com/pipelines/github/bereng/cassandra/1031/workflows/aba91c38-98b6-44eb-b5d4-e27626c7cda9)

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -221,7 +222,11 @@ public DeletionTime deserialize(DataInputPlus in) throws IOException
         {
             int flags = in.readByte();
             if ((flags & IS_LIVE_DELETION) != 0)
+            {
+                if ((flags & 0xFF) != IS_LIVE_DELETION)
+                    throw new CorruptSSTableException(new Exception("Corrupted sstable. Invalid flags found deserializing DeletionTime"), "");
```
blambov: Code as low level as this should not throw `CorruptSSTableException`. Make it an `AssertionError` or `IOException`, so that the calling code can catch it, add the sstable name to the exception, and, most importantly, mark it as corrupt.

bereng:
Switched to IOException and all is [green](https://app.circleci.com/pipelines/github/bereng/cassandra/1034/workflows/323bd7d4-2b40-4b61-98fc-67c5ae51c9b3). Failures are the j17 known ones.

blambov:
LGTM

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Switched to IOException

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Fixed in latest commit

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,103 @@ public static Serializer getSerializer(Version version)
...
-            int localDeletionTimeUnsignedInteger = buf.getInt(offset);
-            long mfda = buf.getLong(offset + 4);
-            return mfda == Long.MIN_VALUE && localDeletionTimeUnsignedInteger == Cell.NO_DELETION_TIME_UNSIGNED_INTEGER
-                   ? LIVE
-                   : new DeletionTime(mfda, localDeletionTimeUnsignedInteger);
+            int flags = buf.get(offset);
+            if ((flags & IS_LIVE_DELETION) != 0)
+            {
+                if ((flags & 0xFF) != IS_LIVE_DELETION)
+                    throw new IOException("Corrupted sstable. Invalid flags found deserializing DeletionTime");
```
pkolaczk: Log the actual flags we got here in the message.

src/java/org/apache/cassandra/db/DeletionTime.java
```diff
@@ -195,42 +195,103 @@ public static Serializer getSerializer(Version version)
...
         {
-            return TypeSizes.sizeof(Integer.MAX_VALUE)
-                   + TypeSizes.sizeof(delTime.markedForDeleteAt());
+            if (delTime == LIVE)
+                // Just the flags
+                return 1;
+            else
+                // 1 for the flags, 7 for mfda and 4 for ldt
+                return TypeSizes.INT_SIZE
+                       + TypeSizes.LONG_SIZE;
```
pkolaczk: OCD-style nit: The order is different in the comment than in the actual return. The comment first mentions the long and int and the return is reversed. My initial reaction was "how do I fit 1 and 7 bytes in an int" ;)

jacek-lewandowski (approved):

pkolaczk (approved):

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Will add on commit

src/java/org/apache/cassandra/db/DeletionTime.java
bereng: Will add on commit

bereng:
Squashed before commit.

bereng:
Note on reviewers: @jakubzytka is OOO currently, knowing his only concern has been cleared, knowing him personally and after having checked with PMC I have been allowed to merge.

PR #2465, opened 2023-07-05
Add basic text analysis to SAI:

mike-tr-adamson:
- Adds the case_sensitive, normalize and ascii options to the index.


test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
```diff
@@ -263,6 +263,120 @@ public void shouldCreateIndexIfExists()
         assertEquals(1, saiCreationCounter.get());
     }
 
+    @Test
+    public void shouldBeCaseSensitiveByDefault() throws Throwable
+    {
+        createTable("CREATE TABLE %s (id text PRIMARY KEY, val text)");
+
+        createIndex("CREATE CUSTOM INDEX ON %s(val) USING 'StorageAttachedIndex'");
+        waitForIndexQueryable();
```
adelapena: Since [CASSANDRA-18521](https://issues.apache.org/jira/browse/CASSANDRA-18521) `CQLTester#waitForIndexQueryable` requires an argument specifying the index name. Its absence here produces a build error. Maybe it has been missed during rebase?

adelapena: Note that also `CQLTester#waitForTableIndexesQueryable` can be used to just wait for all the indexes used by the table.

src/java/org/apache/cassandra/index/sai/analyzer/AbstractAnalyzer.java
```diff
@@ -22,12 +22,22 @@
...
 
 import org.apache.cassandra.db.marshal.AbstractType;
+import org.apache.cassandra.db.marshal.AsciiType;
+import org.apache.cassandra.db.marshal.UTF8Type;
+import org.apache.cassandra.exceptions.InvalidRequestException;
+import org.apache.cassandra.index.sai.utils.TypeUtil;
 
 public abstract class AbstractAnalyzer implements Iterator<ByteBuffer>
 {
+    public static final Set<AbstractType<?>> ANALYZABLE_TYPES = ImmutableSet.of(UTF8Type.instance, AsciiType.instance);
```
adelapena: Maybe it's simpler to get rid of `ANALYZABLE_TYPES` and `TypeUtil#isIn` and just use `(type instanceof StringType)` to check if a type is analyzable?

src/java/org/apache/cassandra/index/sai/analyzer/AbstractAnalyzer.java
```diff
@@ -22,12 +22,22 @@
...
+
     protected ByteBuffer next = null;
+    String nextLiteral = null;
```
adelapena: Nit: can be `protected`

src/java/org/apache/cassandra/index/sai/analyzer/filter/BasicResultFilters.java
```diff
@@ -0,0 +1,81 @@
...
+    public static class LowerCase extends FilterPipelineTask
+    {
+        private final Locale locale;
+
+        public LowerCase()
+        {
+            this.locale = DEFAULT_LOCALE;
+        }
+
+        public String process(String input)
```
adelapena, line 43: Nit: add `@Override`

src/java/org/apache/cassandra/index/sai/analyzer/filter/BasicResultFilters.java
```diff
@@ -0,0 +1,81 @@
...
+        {
+            return input.toLowerCase(locale);
+        }
+    }
+
+    public static class Normalize extends FilterPipelineTask
+    {
+        public Normalize() { }
+
+        public String process(String input)
```
adelapena, line 54: Nit: add `@Override`

src/java/org/apache/cassandra/index/sai/analyzer/filter/BasicResultFilters.java
```diff
@@ -0,0 +1,81 @@
...
+            if (input == null) return null;
+            return Normalizer.isNormalized(input, Normalizer.Form.NFC) ? input : Normalizer.normalize(input, Normalizer.Form.NFC);
+        }
+    }
+
+    public static class Ascii extends FilterPipelineTask
+    {
+        public Ascii() { }
+
+        public String process(String input)
```
adelapena, line 66: Nit: add `@Override`

src/java/org/apache/cassandra/index/sai/analyzer/filter/BasicResultFilters.java
```diff
@@ -0,0 +1,81 @@
...
+            // The output can (potentially) be 4 times the size of the input
+            char[] outputChars = new char[inputChars.length * 4];
+            int outputSize = ASCIIFoldingFilter.foldToASCII(inputChars, 0, outputChars, 0, inputChars.length);
+            return new String(outputChars, 0, outputSize);
+        }
+    }
+
+    public static class NoOperation extends FilterPipelineTask
+    {
+        public String process(String input)
```
adelapena, line 80: Nit: add `@Override`

src/java/org/apache/cassandra/index/sai/analyzer/filter/FilterPipelineExecutor.java
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.index.sai.analyzer.filter;
+
+/**
+ * Executes all linked Pipeline Tasks serially and returns
+ * output (if exists) from the executed logic
```
adelapena: Nit: no need to break the line

test/unit/org/apache/cassandra/index/sai/analyzer/NonTokenizingAnalyzerTest.java
```diff
@@ -0,0 +1,124 @@
...
+import org.apache.cassandra.db.marshal.UTF8Type;
+import org.apache.cassandra.utils.ByteBufferUtil;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+
+/**
+ * Tests for the non-tokenizing analyzer
+ */
+public class NonTokenizingAnalyzerTest
```
adelapena, line 33: It seems there is some code duplication across the tests in this class. Maybe we could use a utility method such as, for example:
```java
private void test(String input, String expected, AbstractAnalyzer analyzer) throws Exception
{
    ByteBuffer toAnalyze = ByteBuffer.wrap(input.getBytes());
    analyzer.reset(toAnalyze);
    ByteBuffer analyzed = null;
    while (analyzer.hasNext())
    {
        analyzed = analyzer.next();
    }
    String result = ByteBufferUtil.string(analyzed);
    assertEquals(expected, result);
}
```

mike-tr-adamson:
@adelapena @bereng Please note that I have just made a number of changes for the Lucene 9.7 upgrade.

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
mike-tr-adamson: Since `waitForTableIndexesQueryable` was added to `createIndex`, I have removed these calls.

test/unit/org/apache/cassandra/index/sai/analyzer/NonTokenizingAnalyzerTest.java
mike-tr-adamson: Agreed, I've done this and generally simplified / tidied this test.

adelapena: It looks nice now :)

test/unit/org/apache/cassandra/index/sai/analyzer/NonTokenizingAnalyzerTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+        assertEquals("nip it in the bud", getAnalyzedString("Nip it in the bud", options));
+    }
+
+    @Test
+    public void caseSensitiveAnalyzer() throws Exception
+    {
+        NonTokenizingOptions options = NonTokenizingOptions.getDefaultOptions();
+
+        assertNotEquals("nip it in the bud", getAnalyzedString("Nip it in the bud", options));
```
adelapena: Maybe `assertEquals("Nip it in the bud", getAnalyzedString("Nip it in the bud", options));` achieves the same and is more strict by verifying what is exactly returned?

src/java/org/apache/cassandra/index/sai/analyzer/filter/BasicResultFilters.java
adelapena: I think this `@Override` is missed in the last commit

mike-tr-adamson: Sorry, not sure how I missed that one.

test/unit/org/apache/cassandra/index/sai/analyzer/NonTokenizingAnalyzerTest.java
mike-tr-adamson: Agreed, I'm not entirely sure why the original test did this because it would have failed on any string.

adelapena:
@mike-tr-adamson I think it would be good to add a dtest quickly verifying that tokenisation and analysis don't break RFP (classic CASSANDRA-8272). I think something like this would work:
```java
public class ReplicaFilteringProtectionTest extends TestBaseImpl
{
    private static final int REPLICAS = 2;

    @Test
    public void testRFPWithIndexTransformations() throws IOException
    {
        try (Cluster cluster = init(Cluster.build()
                                           .withNodes(REPLICAS)
                                           .withConfig(config -> config.set("hinted_handoff_enabled", false)
                                                                       .set("commitlog_sync", "batch")).start()))
        {
            String tableName = "sai_rfp";
            String fullTableName = KEYSPACE + '.' + tableName;

            cluster.schemaChange("CREATE TABLE " + fullTableName + " (k int PRIMARY KEY, v text)");
            cluster.schemaChange("CREATE CUSTOM INDEX ON " + fullTableName + "(v) USING 'StorageAttachedIndex' " +
                                 "WITH OPTIONS = { 'case_sensitive' : false}");

            // both nodes have the old value
            cluster.coordinator(1).execute("INSERT INTO " + fullTableName + "(k, v) VALUES (0, 'OLD')", ALL);

            String select = "SELECT * FROM " + fullTableName + " WHERE v = 'old'";
            Object[][] initialRows = cluster.coordinator(1).execute(select, ALL);
            assertRows(initialRows, row(0, "OLD"));

            // only one node gets the new value
            cluster.get(1).executeInternal("UPDATE " + fullTableName + " SET v = 'new' WHERE k = 0");

            // querying by the old value shouldn't return the old surviving row
            SimpleQueryResult oldResult = cluster.coordinator(1).executeWithResult(select, ALL);
            assertRows(oldResult.toObjectArrays());
        }
    }
}
```

mike-tr-adamson:
@adelapena I've added the `ReplicaFilteringProtectionTest` but, I have to admit, I'm not entirely sure how/why it's working. My main concern is that the test is passing for reasons that aren't the ones that we are testing for. I will have a bit of a dig to confirm this.

adelapena:
@mike-tr-adamson I think RFP is working because `StorageAttachedIndexSearcher #filterReplicaFilteringProtection` takes care of applying the filters to the expressions that are used in the coordinator: https://github.com/apache/cassandra/blob/723cc160921cff6afa1fa7219b3a8a1b4c41c961/src/java/org/apache/cassandra/index/sai/plan/StorageAttachedIndexSearcher.java#L85-L103

You can artificially see the test failing if you modify that method or if you make `DataResolver#needsReplicaFilteringProtection` return false.

mike-tr-adamson:
Thank you, that's what I was trying to find.

test/unit/org/apache/cassandra/index/sai/analyzer/NonTokenizingAnalyzerTest.java
```diff
@@ -0,0 +1,79 @@
...
+
+import java.nio.ByteBuffer;
+
+import org.junit.Test;
+
+import org.apache.cassandra.db.marshal.UTF8Type;
+import org.apache.cassandra.utils.ByteBufferUtil;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
```
adelapena: This unused import is breaking the build: https://app.circleci.com/pipelines/github/adelapena/cassandra/3004/workflows/cff4d9ec-aa49-466d-9f32-8535c3826a64/jobs/56610

mike-tr-adamson: My bad, sorry. I'll get my commit process correct one of these days.

adelapena (approved):

maedhroz:
Committed as https://github.com/apache/cassandra/commit/05dd58783a83c3bd2073f8f96924a99a73cc4f8a

PR #2467, opened 2023-07-06
CASSANDRA-18617 Disable the deprecated keyspace/table thresholds and convert them to Guardrails

djatnieks:
 
```

```


src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -1124,16 +1124,6 @@ public void enableAuditLog(String loggerName, String includedKeyspaces, String e
     public boolean autoOptimisePreviewRepairStreams();
     public void setAutoOptimisePreviewRepairStreams(boolean enabled);
 
-    // warning thresholds will be replaced by equivalent guardrails
```
adelapena: I think we should keep the JMX methods, since it's a public API. We could change their implementation in `StorageService` to use the getters and setters in `Guardrails.instance` instead of using `DatabaseDescriptor`.

src/java/org/apache/cassandra/schema/SchemaConstants.java
```diff
@@ -126,4 +129,23 @@ public static Set<String> getSystemKeyspaces()
...
+        return Sets.union(LOCAL_SYSTEM_KEYSPACE_NAMES, REPLICATED_SYSTEM_KEYSPACE_NAMES);
+    }
+    
+    /**
+     * Returns the set of all local and replicated system table names
+     * @return all local and replicated system table names
+     */
+    public static Set<String> getLocalAndReplicatedSystemTableNames()
+    {
+        return Sets.union(Sets.union(SystemKeyspace.TABLE_NAMES, ImmutableSet.copyOf(SchemaKeyspaceTables.ALL)),
```
adelapena: Nit: The four nested usages of `Sets.union` aren't probably the easier thing to read. Perhaps we could use instead something like:
```java
return ImmutableSet.<String>builder()
                   .addAll(SystemKeyspace.TABLE_NAMES)
                   .addAll(SchemaKeyspaceTables.ALL)
                   .addAll(TraceKeyspace.TABLE_NAMES)
                   .addAll(AuthKeyspace.TABLE_NAMES)
                   .addAll(SystemDistributedKeyspace.TABLE_NAMES)
                   .build();
```

test/unit/org/apache/cassandra/cql3/SystemKeyspaceTablesNamesTest.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.cassandra.db.SystemKeyspace;
+import org.apache.cassandra.schema.Schema;
+import org.apache.cassandra.schema.SchemaConstants;
+import org.apache.cassandra.schema.SchemaKeyspaceTables;
+import org.apache.cassandra.schema.SystemDistributedKeyspace;
+import org.apache.cassandra.tracing.TraceKeyspace;
+
+import static org.junit.Assert.assertTrue;
+
+public class SystemKeyspaceTablesNamesTest extends CQLTester
```
adelapena, line 42: It's nice to have thins test :)

test/unit/org/apache/cassandra/cql3/SystemKeyspaceTablesNamesTest.java
```diff
@@ -0,0 +1,112 @@
...
+    {
+        // Needed for distributed keyspaces
+        requireNetwork();
+    }
+
+    @Test
+    public void testSystemKeyspaceTableNames()
+    {
+        Set<String> tables = Schema.instance.getKeyspaceMetadata(SchemaConstants.SYSTEM_KEYSPACE_NAME).tables
+                             .stream().map(t -> t.name).collect(Collectors.toSet());
```
adelapena: This block is more or less repeated in all the tests, producing a warning about the chance of the `KeyspaceMetadata` being null. We could put it in a utility method:
```java
private static Set<String> tableNames(String keyspaceName)
{
    KeyspaceMetadata keyspace = Schema.instance.getKeyspaceMetadata(keyspaceName);
    assertNotNull(keyspace);
    return keyspace.tables.stream().map(t -> t.name).collect(Collectors.toSet());
}
```

djatnieks: Good idea ... and I think some additional refactoring will improve it some more.

src/java/org/apache/cassandra/service/StorageServiceMBean.java
djatnieks: Good point - I'll try doing that. 
I guess they should be left `@Deprecated` since this maintains an existing api and is not meant to be an official guardrails api.

adelapena: I also think they should be left `@Deprecated`, since they are only for backward compatibility. If we directly translate the arguments of those methods to the guardrail thresholds we are changing their meaning, since they used to count system keyspace/tables. 

I think the setter methods on `StorageService` should subtract the number of system keyspaces/tables. Conversely,  the getters should add the number of system keyspaces/tables to the guardrail number. That way the MBean methods would behave the same way they used to, being a JMX version of the old config properties.

test/unit/org/apache/cassandra/cql3/SystemKeyspaceTablesNamesTest.java
adelapena: The refactored test looks nice. Just a nit, `tablesFromSchema` could be inlined into `assertExpectedTablesInKeyspace`, and that one could be `static`:
```java
private static void assertExpectedTablesInKeyspace(String keyspaceName, String expectedTableSource, Set<String> expectedTables)
{
    KeyspaceMetadata keyspace = Schema.instance.getKeyspaceMetadata(keyspaceName);
    assertNotNull(keyspace);
    Set<String> actualKeyspaceTables = keyspace.tables.stream().map(t -> t.name).collect(Collectors.toSet());

    Sets.SetView<String> diff = Sets.difference(actualKeyspaceTables, expectedTables);
    assertTrue(format("The following tables are missing from %s: %s", expectedTableSource, diff), diff.isEmpty());
    
    diff = Sets.difference(expectedTables, actualKeyspaceTables);
    assertTrue(format("The following tables are in %s but should not be: %s", expectedTableSource,  diff), diff.isEmpty());
}
```

src/java/org/apache/cassandra/service/StorageServiceMBean.java
djatnieks: Hmm, yes, I think you are right about that. I was thinking that legacy values from the config would have been converted already, and users would want to work with current guardrail values. But, as you point out, that changes the semantics of the api, and since we are leaving the api methods deprecated they should work the same.

test/unit/org/apache/cassandra/cql3/SystemKeyspaceTablesNamesTest.java
djatnieks: Just wanted to mention I kept individual test methods for each keyspace - I thought about putting them all together, but I somewhat like that each gets checked independently.

adelapena: It looks perfect to me 👍

adelapena (approved):

PR #2469, opened 2023-07-06
CASSANDRA-18652 trunk: Improve vector value validation errors

adelapena:


src/java/org/apache/cassandra/db/marshal/VectorType.java
```diff
@@ -352,10 +352,10 @@ private void check(List<?> values)
         }
     }
 
-    private static  <V> void checkConsumedFully(V buffer, ValueAccessor<V> accessor, int offset)
+    private <V> void checkConsumedFully(V buffer, ValueAccessor<V> accessor, int offset)
     {
         if (!accessor.isEmptyFromOffset(buffer, offset))
-            throw new MarshalException("Unexpected extraneous bytes after vector value");
+            throw new MarshalException("Unexpected extraneous bytes after " + asCQL3Type() + " value");
```
Maxwell-Guo: Can we give more details about the  left size from offset using : `default int sizeFromOffset(V value, int offset)
    {
        return size(value) - offset;
    }`

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
```diff
@@ -139,6 +140,95 @@ public void insertNonPK()
...
+                                  "INSERT INTO %s (pk, value) VALUES (0, ['a'])");
+        assertInvalidThrowMessage("Not enough bytes to read a vector<text, 2>",
+                                  InvalidRequestException.class,
+                                  "INSERT INTO %s (pk, value) VALUES (0, ?)", vector("a"));
+
+        // more values than expected, with literals and bind markers
+        assertInvalidThrowMessage("Invalid vector literal for value of type vector<text, 2>; expected 2 elements, but given 3",
+                                  InvalidRequestException.class,
+                                  "INSERT INTO %s (pk, value) VALUES (0, ['a', 'b', 'c'])");
+        assertInvalidThrowMessage("Unexpected extraneous bytes after vector<text, 2> value",
```
Maxwell-Guo: what about show more detail messages about the left size as I have describe before ?
May be can be : "Unexpected extraneous bytes after vector<text, 2> value,  xxx size of text is given"

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
```diff
@@ -139,6 +140,95 @@ public void insertNonPK()
         test.run();
     }
 
+    @Test
+    public void cannotInsertWrongNumberOfDimensionsFixedWidth() throws Throwable
```
Maxwell-Guo: what about change the function name to : wrongNumberOfDimensionsFixedWidth, and the following function names are also the same to remove "cannotInsert", although the function name explains what to do in detail, but I feel a bit long; 
In my mind these tests are all validations for throw exception messages(assertInvalidThrowMessage function everywhere), that is to say the code in the function is saying that wrongNumberxxx will throw an exception.
And wrongNumber and wrongElement is saying this is abnormal error test test(the name is wrongXXX).

adelapena: Yes, the names feel a bit redundant. Changed.

src/java/org/apache/cassandra/db/marshal/VectorType.java
adelapena: I'm not sure I see a lot of value in printing the specific size of the remaining bytes, and we don't do it in similar methods, but I have added it nevertheless.

Maxwell-Guo (approved):

ekaterinadimitrova2 (approved):
Only CHANGES.txt entry needs to be added on commit
+1 on green cqlshlib tests (failed in the last run, probably something transient)

PR #2474, opened 2023-07-10
Cassandra-17914: Argparse migration as the Python Optparse library is deprecated on Cassandra-4.1

vgali7:


bin/cqlsh.py
```diff
@@ -70,6 +70,7 @@
     CASSANDRA_CQL_HTML = 'file://' + CASSANDRA_PATH + '/doc/cql3/CQL.html'
 elif os.path.exists('/usr/share/doc/cassandra/CQL.html'):
     # fallback to package file
+    print(112)
```
bschoening: Is this a leftover debug statement?

vgali7: Yes, it is removed now

vgali7: Yes, it is removed now

PR #2488, opened 2023-07-13
CASSANDRA-18329: Upgrade Jamm

blerer:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/audit/BinAuditLogger.java
+++ b/src/java/org/apache/cassandra/audit/BinAuditLogger.java
@@ -26,6 +26,7 @@ import org.slf4j.LoggerFactory;
 
 import net.openhft.chronicle.wire.WireOut;
 import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.fql.FullQueryLogger.Query;
 import org.apache.cassandra.io.util.File;
 import org.apache.cassandra.utils.ObjectSizes;
 import org.apache.cassandra.utils.binlog.BinLog;
@@ -98,6 +99,11 @@ public class BinAuditLogger implements IAuditLogger
     @VisibleForTesting
     public static class Message extends BinLog.ReleaseableWriteMarshallable implements WeightedQueue.Weighable
     {
+        /**
+         * The shallow size of a {@code Query} object.
+         */
+        private static final long EMPTY_SIZE = ObjectSizes.measure(new Message(""));
+
         private final String message;
 
         public Message(String message)
@@ -130,7 +136,7 @@ public class BinAuditLogger implements IAuditLogger
         @Override
         public int weight()
         {
-            return Ints.checkedCast(ObjectSizes.sizeOf(message));
+            return Ints.checkedCast(EMPTY_SIZE + ObjectSizes.sizeOf(message));
         }
     }
 }
--- a/src/java/org/apache/cassandra/fql/FullQueryLogger.java
+++ b/src/java/org/apache/cassandra/fql/FullQueryLogger.java
@@ -50,7 +50,7 @@ import org.apache.cassandra.utils.binlog.BinLog;
 import org.apache.cassandra.utils.binlog.BinLogOptions;
 import org.apache.cassandra.utils.concurrent.UncheckedInterruptedException;
 import org.apache.cassandra.utils.concurrent.WeightedQueue;
-import org.github.jamm.MemoryLayoutSpecification;
+import org.github.jamm.Unmetered;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 
@@ -84,9 +84,6 @@ public class FullQueryLogger implements QueryEvents.Listener
 
     private static final int EMPTY_LIST_SIZE = Ints.checkedCast(ObjectSizes.measureDeep(new ArrayList<>(0)));
     private static final int EMPTY_BYTEBUF_SIZE;
-
-    private static final int OBJECT_HEADER_SIZE = MemoryLayoutSpecification.SPEC.getObjectHeaderSize();
-    private static final int OBJECT_REFERENCE_SIZE = MemoryLayoutSpecification.SPEC.getReferenceSize();
 
     public static final FullQueryLogger instance = new FullQueryLogger();
 
@@ -332,12 +329,25 @@ public class FullQueryLogger implements QueryEvents.Listener
 
     public static class Query extends AbstractLogEntry
     {
+        /**
+         * The shallow size of a {@code Query} object.
+         */
+        private static final long EMPTY_SIZE = ObjectSizes.measure(new Query());
+
         private final String query;
 
         public Query(String query, QueryOptions queryOptions, QueryState queryState, long queryStartTime)
         {
             super(queryOptions, queryState, queryStartTime);
             this.query = query;
+        }
+
+        /**
+         * Constructor only use to compute this class shallow size.
+         */
+        private Query()
+        {
+            this.query = null;
         }
 
         @Override
@@ -356,12 +366,17 @@ public class FullQueryLogger implements QueryEvents.Listener
         @Override
         public int weight()
         {
-            return Ints.checkedCast(ObjectSizes.sizeOf(query)) + super.weight();
+            return Ints.checkedCast(EMPTY_SIZE + ObjectSizes.sizeOf(query) + super.fielsdSize());
         }
     }
 
     public static class Batch extends AbstractLogEntry
     {
+        /**
+         * The shallow size of a {@code Batch} object.
+         */
+        private static final long EMPTY_SIZE = ObjectSizes.measure(new Batch());
+
         private final int weight;
         private final BatchStatement.Type batchType;
         private final List<String> queries;
@@ -380,25 +395,33 @@ public class FullQueryLogger implements QueryEvents.Listener
             this.values = values;
             this.batchType = batchType;
 
-            int weight = super.weight();
-
-            // weight, queries, values, batch type
-            weight += Integer.BYTES +            // cached weight
-                      2 * EMPTY_LIST_SIZE +      // queries + values lists
-                      3 * OBJECT_REFERENCE_SIZE; // batchType and two lists references
-
+            // We assume that all the lists are ArrayLists and that the size of each underlying array is the one of the list 
+            // (which is obviously wrong but not worst than the previous computation that was ignoring part of the arrays size in the computation).
+            long queriesSize = EMPTY_LIST_SIZE + ObjectSizes.sizeOfReferenceArray(queries.size());
             for (String query : queries)
-                weight += ObjectSizes.sizeOf(checkNotNull(query)) + OBJECT_REFERENCE_SIZE;
+                queriesSize += ObjectSizes.sizeOf(checkNotNull(query));
 
+            long valuesSize = EMPTY_LIST_SIZE + ObjectSizes.sizeOfReferenceArray(values.size());
             for (List<ByteBuffer> subValues : values)
             {
-                weight += EMPTY_LIST_SIZE + OBJECT_REFERENCE_SIZE;
-
-                for (ByteBuffer value : subValues)
-                    weight += ObjectSizes.sizeOnHeapOf(value) + OBJECT_REFERENCE_SIZE;
+                valuesSize += EMPTY_LIST_SIZE + ObjectSizes.sizeOfReferenceArray(subValues.size());
+                for (ByteBuffer subValue : subValues)
+                    valuesSize += ObjectSizes.sizeOnHeapOf(subValue);
             }
 
-            this.weight = weight;
+            // No need to add the batch type which is an enum. 
+            this.weight = Ints.checkedCast(EMPTY_SIZE + super.fielsdSize() + queriesSize + valuesSize);
+        }
+
+        /**
+         * Constructor only use to compute this class shallow size.
+         */
+        private Batch()
+        {
+            this.weight = 0;
+            this.batchType = null;
+            this.queries = null;
+            this.values = null;
         }
 
         @Override
@@ -441,6 +464,7 @@ public class FullQueryLogger implements QueryEvents.Listener
     {
         private final long queryStartTime;
         private final int protocolVersion;
+        @Unmetered
         private final ByteBuf queryOptionsBuffer;
 
         private final long generatedTimestamp;
@@ -483,6 +507,19 @@ public class FullQueryLogger implements QueryEvents.Listener
             }
         }
 
+        /**
+         * Constructor only use to compute sub-classes shallow size.
+         */
+        private AbstractLogEntry()
+        {
+            this.queryStartTime = 0;
+            this.protocolVersion = 0;
+            this.queryOptionsBuffer = null;
+            this.generatedTimestamp = 0;
+            this.generatedNowInSeconds = 0;
+            this.keyspace = null;
+        }
+
         @Override
         protected long version()
         {
@@ -508,16 +545,14 @@ public class FullQueryLogger implements QueryEvents.Listener
             queryOptionsBuffer.release();
         }
 
-        @Override
-        public int weight()
+        /**
+         * Returns the sum of the non primitive fields sizes.
+         * @return the sum of the non primitive fields sizes.
+         */
+        protected long fielsdSize()
         {
-            return OBJECT_HEADER_SIZE
-                 + Long.BYTES                                                                 // queryStartTime
-                 + Integer.BYTES                                                              // protocolVersion
-                 + OBJECT_REFERENCE_SIZE + EMPTY_BYTEBUF_SIZE + queryOptionsBuffer.capacity() // queryOptionsBuffer
-                 + Long.BYTES                                                                 // generatedTimestamp
-                 + Long.BYTES                                                                 // generatedNowInSeconds
-                 + OBJECT_REFERENCE_SIZE + Ints.checkedCast(ObjectSizes.sizeOf(keyspace));    // keyspace
+            return EMPTY_BYTEBUF_SIZE + queryOptionsBuffer.capacity() // queryOptionsBuffer
+                   + ObjectSizes.sizeOf(keyspace);                    // keyspace
         }
     }
 
--- a/src/java/org/apache/cassandra/fql/FullQueryLogger.java
+++ b/src/java/org/apache/cassandra/fql/FullQueryLogger.java
@@ -366,17 +366,21 @@ public class FullQueryLogger implements QueryEvents.Listener
         @Override
         public int weight()
         {
-            return Ints.checkedCast(EMPTY_SIZE + ObjectSizes.sizeOf(query) + super.fielsdSize());
+            // Object deep size = Object' shallow size + query field deep size + deep size of the parent fields
+            return Ints.checkedCast(EMPTY_SIZE + ObjectSizes.sizeOf(query) + super.fieldsSize());
         }
     }
 
     public static class Batch extends AbstractLogEntry
     {
         /**
-         * The shallow size of a {@code Batch} object.
+         * The shallow size of a {@code Batch} object (which includes primitive fields).
          */
         private static final long EMPTY_SIZE = ObjectSizes.measure(new Batch());
 
+        /**
+         * The way is pre-computed in the constructor and represent the object deep size.
+         */
         private final int weight;
         private final BatchStatement.Type batchType;
         private final List<String> queries;
@@ -398,6 +402,7 @@ public class FullQueryLogger implements QueryEvents.Listener
             // We assume that all the lists are ArrayLists and that the size of each underlying array is the one of the list 
             // (which is obviously wrong but not worst than the previous computation that was ignoring part of the arrays size in the computation).
             long queriesSize = EMPTY_LIST_SIZE + ObjectSizes.sizeOfReferenceArray(queries.size());
+
             for (String query : queries)
                 queriesSize += ObjectSizes.sizeOf(checkNotNull(query));
 
@@ -409,8 +414,11 @@ public class FullQueryLogger implements QueryEvents.Listener
                     valuesSize += ObjectSizes.sizeOnHeapOf(subValue);
             }
 
-            // No need to add the batch type which is an enum. 
-            this.weight = Ints.checkedCast(EMPTY_SIZE + super.fielsdSize() + queriesSize + valuesSize);
+            // No need to add the batch type which is an enum.
+            this.weight = Ints.checkedCast(EMPTY_SIZE            // Shallow size object
+                                            + super.fieldsSize() // deep size of the parent fields (non-primitives as they are included in the shallow size) 
+                                            + queriesSize        // deep size queries field
+                                            + valuesSize);       // deep size values field
         }
 
         /**
@@ -464,7 +472,6 @@ public class FullQueryLogger implements QueryEvents.Listener
     {
         private final long queryStartTime;
         private final int protocolVersion;
-        @Unmetered
         private final ByteBuf queryOptionsBuffer;
 
         private final long generatedTimestamp;
@@ -546,10 +553,10 @@ public class FullQueryLogger implements QueryEvents.Listener
         }
 
         /**
-         * Returns the sum of the non primitive fields sizes.
-         * @return the sum of the non primitive fields sizes.
+         * Returns the sum of the non-primitive fields' deep sizes.
+         * @return the sum of the non-primitive fields' deep sizes.
          */
-        protected long fielsdSize()
+        protected long fieldsSize()
         {
             return EMPTY_BYTEBUF_SIZE + queryOptionsBuffer.capacity() // queryOptionsBuffer
                    + ObjectSizes.sizeOf(keyspace);                    // keyspace
--- a/src/java/org/apache/cassandra/utils/ObjectSizes.java
+++ b/src/java/org/apache/cassandra/utils/ObjectSizes.java
@@ -24,23 +24,25 @@ import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
 
-import org.github.jamm.MemoryLayoutSpecification;
 import org.github.jamm.MemoryMeter;
+import org.github.jamm.MemoryMeter.ByteBufferMode;
+import org.github.jamm.MemoryMeter.Guess;
+
+import static org.github.jamm.MemoryMeterStrategy.MEMORY_LAYOUT;
+import static org.github.jamm.utils.ArrayMeasurementUtils.computeArraySize;
 
 /**
  * A convenience class for wrapping access to MemoryMeter
  */
 public class ObjectSizes
 {
-    private static final MemoryMeter meter = new MemoryMeter().withGuessing(MemoryMeter.Guess.FALLBACK_UNSAFE)
-                                                              .ignoreKnownSingletons();
-    private static final MemoryMeter omitSharedMeter = meter.omitSharedBufferOverhead();
+    private static final MemoryMeter meter = MemoryMeter.builder().withGuessing(Guess.INSTRUMENTATION_AND_SPECIFICATION,
+                                                                                Guess.UNSAFE)
+                                                                  .build();
 
-    private static final long EMPTY_HEAP_BUFFER_SIZE = measure(ByteBufferUtil.EMPTY_BYTE_BUFFER);
-    private static final long EMPTY_BYTE_ARRAY_SIZE = measure(new byte[0]);
-    private static final long EMPTY_STRING_SIZE = measure("");
-
-    private static final long DIRECT_BUFFER_HEAP_SIZE = measure(ByteBuffer.allocateDirect(0));
+    private static final long HEAP_BUFFER_SHALLOW_SIZE = measure(ByteBufferUtil.EMPTY_BYTE_BUFFER);
+    private static final long DIRECT_BUFFER_SHALLOW_SIZE = measure(ByteBuffer.allocateDirect(0));
+    private static final long DIRECT_BUFFER_DEEP_SIZE = measureDeep(ByteBuffer.allocateDirect(0));
 
     public static final long IPV6_SOCKET_ADDRESS_SIZE = ObjectSizes.measureDeep(new InetSocketAddress(getIpvAddress(16), 42));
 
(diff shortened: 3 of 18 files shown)


src/java/org/apache/cassandra/fql/FullQueryLogger.java
```diff
@@ -356,12 +366,17 @@ public void writeMarshallablePayload(WireOut wire)
         @Override
         public int weight()
         {
-            return Ints.checkedCast(ObjectSizes.sizeOf(query)) + super.weight();
+            return Ints.checkedCast(EMPTY_SIZE + ObjectSizes.sizeOf(query) + super.fielsdSize());
```
ekaterinadimitrova2:
```suggestion
            return Ints.checkedCast(EMPTY_SIZE + ObjectSizes.sizeOf(query) + super.fieldsSize());
```

src/java/org/apache/cassandra/fql/FullQueryLogger.java
```diff
@@ -508,16 +545,14 @@ public void release()
...
         {
-            return OBJECT_HEADER_SIZE
-                 + Long.BYTES                                                                 // queryStartTime
-                 + Integer.BYTES                                                              // protocolVersion
-                 + OBJECT_REFERENCE_SIZE + EMPTY_BYTEBUF_SIZE + queryOptionsBuffer.capacity() // queryOptionsBuffer
-                 + Long.BYTES                                                                 // generatedTimestamp
-                 + Long.BYTES                                                                 // generatedNowInSeconds
-                 + OBJECT_REFERENCE_SIZE + Ints.checkedCast(ObjectSizes.sizeOf(keyspace));    // keyspace
+            return EMPTY_BYTEBUF_SIZE + queryOptionsBuffer.capacity() // queryOptionsBuffer
+                   + ObjectSizes.sizeOf(keyspace);                    // keyspace
```
ekaterinadimitrova2, line 562: Wondering why was there the `Ints.checkedCast`? Precaution? Do we need it back?

src/java/org/apache/cassandra/fql/FullQueryLogger.java
```diff
@@ -508,16 +545,14 @@ public void release()
             queryOptionsBuffer.release();
         }
 
-        @Override
-        public int weight()
+        /**
+         * Returns the sum of the non primitive fields sizes.
+         * @return the sum of the non primitive fields sizes.
```
ekaterinadimitrova2:
```suggestion
         * Returns the sum of the non-primitive fields' sizes.
         * @return the sum of the non-primitive fields' sizes.
```

src/java/org/apache/cassandra/fql/FullQueryLogger.java
ekaterinadimitrova2: It seems the weight was a deep size, and to differentiate it from what you decided to weigh here, you changed the name? Is this correct? So only I need to know why the primitives were excluded if that was the case.

src/java/org/apache/cassandra/fql/FullQueryLogger.java
```diff
@@ -380,25 +395,33 @@ public Batch(BatchStatement.Type batchType,
...
             this.batchType = batchType;
 
-            int weight = super.weight();
-
-            // weight, queries, values, batch type
-            weight += Integer.BYTES +            // cached weight
-                      2 * EMPTY_LIST_SIZE +      // queries + values lists
-                      3 * OBJECT_REFERENCE_SIZE; // batchType and two lists references
-
+            // We assume that all the lists are ArrayLists and that the size of each underlying array is the one of the list
```
ekaterinadimitrova2, line 402: I did not find anything that was not an ArrayList so I don't think it is just an assumption, more like strong requirement.

src/java/org/apache/cassandra/fql/FullQueryLogger.java
ekaterinadimitrova2: I think we need also to add EMPTY_SIZE for the `BinAuditLogger#Message#weight()`

src/java/org/apache/cassandra/utils/ObjectSizes.java
```diff
@@ -105,15 +98,12 @@ public static long sizeOfReferenceArray(int length)
      */
     public static long sizeOfArray(Object[] objects)
     {
-        if (objects == null)
-            return 0;
-
-        return sizeOfReferenceArray(objects.length);
+        return meter.measureArray(objects);
```
ekaterinadimitrova2, line 102: If we substitute `sizeOfReferenceArray` with `meter.measureArray` here, why don't we do it everywhere?

Did you opt out of that as it will be easier for people to read it around the codebase, or you did not think of that? 
So effectively, we also have the null check in `sizeOfReferenceArray,` where it was not added before.

src/java/org/apache/cassandra/utils/ObjectSizes.java
```diff
@@ -105,15 +98,12 @@ public static long sizeOfReferenceArray(int length)
...
     }
 
-    private static long sizeOfArray(int length, long elementSize)
+    private static long sizeOfArray(int length, int elementSize)
```
ekaterinadimitrova2, line 105: So the `getReferenceSize` was always int, and long was for nothing here...

src/java/org/apache/cassandra/utils/ObjectSizes.java
```diff
@@ -134,65 +124,89 @@ public static long sizeOnHeapOf(ByteBuffer[] array)
...
 
-        // if we're only referencing a sub-portion of the ByteBuffer, don't count the array overhead (assume it is SLAB
-        // allocated - the overhead amortized over all the allocations is negligible and better to undercount than over)
-        if (arrayLen > bufLen)
-            return EMPTY_HEAP_BUFFER_SIZE + bufLen;
+        if (buffer.isDirect())
+            return DIRECT_BUFFER_DEEP_SIZE; // That might not be true if the buffer is a view of another buffer so we could undercount
 
-        return EMPTY_HEAP_BUFFER_SIZE + (arrayLen == 0 ? EMPTY_BYTE_ARRAY_SIZE : sizeOfArray(arrayLen, 1));
+        return HEAP_BUFFER_SHALLOW_SIZE + meter.measureArray(buffer.array());
```
ekaterinadimitrova2, line 174: We don't need EMPTY_BYTE_ARRAY_SIZE, anymore? Isn't it gonna return 0? Do we have a test case or so?

src/java/org/apache/cassandra/utils/ObjectSizes.java
```diff
@@ -134,65 +124,89 @@ public static long sizeOnHeapOf(ByteBuffer[] array)
...
+            return HEAP_BUFFER_SHALLOW_SIZE; // We ignore the array overhead
+        }
+
+        if (buffer.isDirect())
+            return DIRECT_BUFFER_DEEP_SIZE; // That might not be true if the buffer is a view of another buffer so we could undercount
 
-        // If buffers are dedicated, account for byte array size and any padding overhead
-        return EMPTY_HEAP_BUFFER_SIZE + (arrayLen == 0 ? EMPTY_BYTE_ARRAY_SIZE : (sizeOfArray(arrayLen, 1) - arrayLen));
+        byte[] bytes = buffer.array();
+        return HEAP_BUFFER_SHALLOW_SIZE + meter.measureArray(bytes) - bytes.length;
```
ekaterinadimitrova2, line 210: Same question about EMPTY_BYTE_ARRAY_SIZE

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
```diff
@@ -22,128 +22,81 @@
...
-
-        suffixBuf9.position(1);
-        suffixBuf9 = suffixBuf9.slice();
-
-        infixBuf7.limit(8);
-        infixBuf7.position(1);
-        infixBuf7 = infixBuf7.slice();
-    }
+    private static final MemoryMeter meter = MemoryMeter.builder()
+                                                        .withGuessing(Guess.INSTRUMENTATION, Guess.UNSAFE)
```
ekaterinadimitrova2, line 35: My assumption is you intentionally created the meter with `INSTRUMENTATION` and not `INSTRUMENTATION_AND_SPEC` as in `ObjectSizes`, for additional verification. Can we add a comment so people realize that and do not consider it a bug and change it to `INSTRUMENTATION_AND_SPEC`?

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
```diff
@@ -22,128 +22,81 @@
...
         // single non-empty direct buffer
-        buffers = new ByteBuffer[]{ ByteBuffer.allocateDirect(10) };
-        assertThat(ObjectSizes.sizeOnHeapExcludingData(buffers)).isEqualTo(REF_ARRAY_1_SIZE + EMPTY_OFFHEAP_BUFFER_RAW_SIZE);
+        checkBufferSizeExcludingData(ByteBuffer.allocateDirect(10), 0);
 
-        // two different empty byte buffers
-        buffers = new ByteBuffer[]{ ByteBuffer.allocate(0), ByteBuffer.allocateDirect(0) };
-        assertThat(ObjectSizes.sizeOnHeapExcludingData(buffers)).isEqualTo(REF_ARRAY_2_SIZE + EMPTY_HEAP_BUFFER_RAW_SIZE + BYTE_ARRAY_0_SIZE + EMPTY_OFFHEAP_BUFFER_RAW_SIZE);
+        // heap buffer being a prefix slab
+        ByteBuffer buffer = (ByteBuffer) ByteBuffer.allocate(10).duplicate().limit(8);
```
ekaterinadimitrova2, line 54:
```suggestion
        ByteBuffer buffer = ByteBuffer.allocate(10).duplicate().limit(8);
```

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
```diff
@@ -22,128 +22,81 @@
...
+        checkBufferSizeExcludingData(buffer, 8);
 
-        // two different non-empty byte buffers
-        buffers = new ByteBuffer[]{ buf10, ByteBuffer.allocateDirect(500) };
-        assertThat(ObjectSizes.sizeOnHeapExcludingData(buffers)).isEqualTo(REF_ARRAY_2_SIZE + EMPTY_HEAP_BUFFER_RAW_SIZE + BYTE_ARRAY_10_EXCEPT_DATA_SIZE + EMPTY_OFFHEAP_BUFFER_RAW_SIZE);
+        // heap buffer being a suffix slab
+        buffer = (ByteBuffer) ByteBuffer.allocate(10).duplicate().position(1);
```
ekaterinadimitrova2, line 58:
```suggestion
        buffer = ByteBuffer.allocate(10).duplicate().position(1);
```

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
```diff
@@ -22,128 +22,81 @@
...
-
-        // heap buffer being a suffix slice of other buffer
-        buffers = new ByteBuffer[]{ suffixBuf9 };
-        assertThat(ObjectSizes.sizeOnHeapExcludingData(buffers)).isEqualTo(REF_ARRAY_1_SIZE + EMPTY_HEAP_BUFFER_RAW_SIZE);
-
-        // heap buffer being an infix slice of other buffer
-        buffers = new ByteBuffer[]{ infixBuf7 };
-        assertThat(ObjectSizes.sizeOnHeapExcludingData(buffers)).isEqualTo(REF_ARRAY_1_SIZE + EMPTY_HEAP_BUFFER_RAW_SIZE);
+        // heap buffer being an infix slab
+        buffer = (ByteBuffer) ByteBuffer.allocate(10).duplicate().position(1).limit(8);
```
ekaterinadimitrova2, line 62:
```suggestion
        buffer = ByteBuffer.allocate(10).duplicate().position(1).limit(8);
```

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
```diff
@@ -22,128 +22,81 @@
...
         // single non-empty direct buffer
-        buffers = new ByteBuffer[]{ ByteBuffer.allocateDirect(10) };
-        assertThat(ObjectSizes.sizeOnHeapOf(buffers)).isEqualTo(REF_ARRAY_1_SIZE + EMPTY_OFFHEAP_BUFFER_RAW_SIZE);
+        checkBufferSize(ByteBuffer.allocateDirect(10));
 
-        // two different empty byte buffers
-        buffers = new ByteBuffer[]{ ByteBuffer.allocate(0), ByteBuffer.allocateDirect(0) };
-        assertThat(ObjectSizes.sizeOnHeapOf(buffers)).isEqualTo(REF_ARRAY_2_SIZE + EMPTY_HEAP_BUFFER_RAW_SIZE + BYTE_ARRAY_0_SIZE + EMPTY_OFFHEAP_BUFFER_RAW_SIZE);
+        // heap buffer being a prefix slab
+        ByteBuffer buffer = (ByteBuffer) ByteBuffer.allocate(10).duplicate().limit(8);
```
ekaterinadimitrova2, line 111:
```suggestion
        ByteBuffer buffer = ByteBuffer.allocate(10).duplicate().limit(8);
```

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
```diff
@@ -22,128 +22,81 @@
...
+        checkBufferSize(buffer);
 
-        // two different non-empty byte buffers
-        buffers = new ByteBuffer[]{ buf10, ByteBuffer.allocateDirect(500) };
-        assertThat(ObjectSizes.sizeOnHeapOf(buffers)).isEqualTo(REF_ARRAY_2_SIZE + EMPTY_HEAP_BUFFER_RAW_SIZE + BYTE_ARRAY_10_SIZE + EMPTY_OFFHEAP_BUFFER_RAW_SIZE);
+        // heap buffer being a suffix slab
+        buffer = (ByteBuffer) ByteBuffer.allocate(10).duplicate().position(1);
```
ekaterinadimitrova2, line 115:
```suggestion
        buffer = ByteBuffer.allocate(10).duplicate().position(1);
```

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
```diff
@@ -22,128 +22,81 @@
...
-
-        // heap buffer being a suffix slice of other buffer
-        buffers = new ByteBuffer[]{ suffixBuf9 };
-        assertThat(ObjectSizes.sizeOnHeapOf(buffers)).isEqualTo(REF_ARRAY_1_SIZE + EMPTY_HEAP_BUFFER_RAW_SIZE + 9);
-
-        // heap buffer being an infix slice of other buffer
-        buffers = new ByteBuffer[]{ infixBuf7 };
-        assertThat(ObjectSizes.sizeOnHeapOf(buffers)).isEqualTo(REF_ARRAY_1_SIZE + EMPTY_HEAP_BUFFER_RAW_SIZE + 7);
+        // heap buffer being an infix slab
+        buffer = (ByteBuffer) ByteBuffer.allocate(10).duplicate().position(1).limit(8);
```
ekaterinadimitrova2, line 119:
```suggestion
        buffer = ByteBuffer.allocate(10).duplicate().position(1).limit(8);
```

test/unit/org/apache/cassandra/db/memtable/MemtableSizeTestBase.java
```diff
@@ -39,16 +39,20 @@
...
 // value memtableClass is initialized with.
 @RunWith(Parameterized.class)
 public abstract class MemtableSizeTestBase extends CQLTester
 {
-    // Note: To see a printout of the usage for each object, add .enableDebug() here (most useful with smaller number of
+    // Note: To see a printout of the usage for each object, add .printVisitedTree() here (most useful with smaller number of
     // partitions).
-    static MemoryMeter meter = new MemoryMeter().ignoreKnownSingletons()
-                                                .withGuessing(MemoryMeter.Guess.FALLBACK_UNSAFE);
+    static MemoryMeter meter =  MemoryMeter.builder()
```
ekaterinadimitrova2, line 51: Why do we need a new meter, and we do not do `ObjectSizes.measureDeep` instead?  CC @blambov

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
```diff
@@ -22,128 +22,81 @@
...
-        buffers = new ByteBuffer[]{ ByteBuffer.allocateDirect(0) };
-        assertThat(ObjectSizes.sizeOnHeapExcludingData(buffers)).isEqualTo(REF_ARRAY_1_SIZE + EMPTY_OFFHEAP_BUFFER_RAW_SIZE);
+        checkBufferSizeExcludingData(ByteBuffer.allocateDirect(0), 0);
 
         // single non-empty direct buffer
-        buffers = new ByteBuffer[]{ ByteBuffer.allocateDirect(10) };
-        assertThat(ObjectSizes.sizeOnHeapExcludingData(buffers)).isEqualTo(REF_ARRAY_1_SIZE + EMPTY_OFFHEAP_BUFFER_RAW_SIZE);
+        checkBufferSizeExcludingData(ByteBuffer.allocateDirect(10), 0);
 
-        // two different empty byte buffers
```
ekaterinadimitrova2, line 85: Why not measuring arrays anymore?

src/java/org/apache/cassandra/fql/FullQueryLogger.java
```diff
@@ -441,6 +464,7 @@ private static abstract class AbstractLogEntry extends BinLog.ReleaseableWriteMa
     {
         private final long queryStartTime;
         private final int protocolVersion;
+        @Unmetered
         private final ByteBuf queryOptionsBuffer;
```
ekaterinadimitrova2: Is this `@Unmetered` some leftover? We measure `queryOptionsBuffer` by getting its capacity directly.

src/java/org/apache/cassandra/fql/FullQueryLogger.java
ekaterinadimitrova2: Plus I think it will be good to explain what and why measure in the javadoc for both methods.

src/java/org/apache/cassandra/utils/ObjectSizes.java
```diff
@@ -24,23 +24,25 @@
...
+import org.github.jamm.MemoryMeter.ByteBufferMode;
+import org.github.jamm.MemoryMeter.Guess;
+
+import static org.github.jamm.MemoryMeterStrategy.MEMORY_LAYOUT;
+import static org.github.jamm.utils.ArrayMeasurementUtils.computeArraySize;
 
 /**
  * A convenience class for wrapping access to MemoryMeter
  */
 public class ObjectSizes
```
ekaterinadimitrova2, line 38: So after seeing the InMemoryTrie not using the `ObjectSizes` wrapper I have a question... should we always use the wrapper and make it clear in the javadoc here?

ekaterinadimitrova2 (commented):
Review done. Overall it looks good; small questions and comments left. The main questions about `InMemoryTries` and the `FullQueryLogger` (Plus `BinAuditLogger`) were not updated here. 
One thing that is not mentioned in the code comments but I discussed a few days ago with another community member - do we need to measure actually at all the `ClusteringComparator `[CASSANDRA-16304](https://issues.apache.org/jira/browse/CASSANDRA-16304)? Now we can, as we fixed jamm for newer versions and hidden classes (which are used to implement lambdas), but should we measure this particular lambda?

I did not review the eclipse warnings commit as I believe it is out of scope for this ticket.
Perhaps you should mention it on CASSANDRA-18239, where we replace eclipse warnings with CheckerFramework.

ekaterinadimitrova2:
Almost forgot:
1) Fixed test here:
https://github.com/ekaterinadimitrova2/cassandra/commit/086800c8427e4c7211efdd898345377a385fcc3d
(without this, the branch will not compile)
2) Hacked jamm jar in this branch so we can test in CI (we won't commit it as part of cassandra/lib, this was only for test purposes until jamm 0.4.0 is released in maven): https://github.com/ekaterinadimitrova2/cassandra/tree/repeated-run2
3) We also need to add jamm version update to 0.4.0 in:
build.xml, redhat/cassandra.in.sh, conf/cassandra-env.sh, bin/cassandra.in.sh
4) CI run with JDK11 and JDK17 shows that all known jamm-related failures are fixed with the new jamm version, including the flakiness from CASSANDRA-17884.
The remaining Java 17 failures are already addressed in ECJ, netty update, and CASSANDRA-18180 (ready to commit when we drop JDK8).
CI runs: 
JDK11 - https://app.circleci.com/pipelines/github/ekaterinadimitrova2/cassandra/2418/workflows/e2a63ad4-4d51-45c0-a876-a45512b20073
JDK17 - https://app.circleci.com/pipelines/github/ekaterinadimitrova2/cassandra/2418/workflows/c386187e-dd3d-4e42-83c6-9602dd6759f1

src/java/org/apache/cassandra/utils/ObjectSizes.java
blerer: The biggest element size if 8. So even byte would work here. :-)

src/java/org/apache/cassandra/utils/ObjectSizes.java
blerer: If the array as a length of zero, `MemoryMeter.measureArray` will return the array based offset which is an empty array size. You avoid the miss-prediction cost for the if, as zero size array are rare, and get the same result. + it is easier to read.

src/java/org/apache/cassandra/utils/ObjectSizes.java
blerer: Same answer ;-)

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
blerer: I believe that it is needed for JDK 8 that we have not dropped yet.

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
blerer: I believe that it is needed for JDK 8 that we have not dropped yet.

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
blerer: I believe that it is needed for JDK 8 that we have not dropped yet.

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
blerer: I believe that it is needed for JDK 8 that we have not dropped yet.

test/unit/org/apache/cassandra/db/memtable/MemtableSizeTestBase.java
blerer: I believe that it better to have full control on the Meter even for debugging.

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
blerer: The important logic is the one for the ByteBuffer so I focussed on those. I can add a couple of tests for the arrays.

test/unit/org/apache/cassandra/db/memtable/MemtableSizeTestBase.java
ekaterinadimitrova2: I am not sure I fully agree on this, but we can leave it, it is not that it hurts too I guess. Closing this one

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
```diff
@@ -22,128 +22,81 @@
...
-        assertThat(ObjectSizes.sizeOnHeapOf(buffers)).isEqualTo(REF_ARRAY_1_SIZE + EMPTY_HEAP_BUFFER_RAW_SIZE + 7);
+        // heap buffer being an infix slab
+        buffer = (ByteBuffer) ByteBuffer.allocate(10).duplicate().position(1).limit(8);
+        checkBufferSize(buffer);
     }
+
+    private void checkBufferSize(ByteBuffer buffer)
+    {
+        assertEquals(meter.measureDeep(buffer, SLAB_ALLOCATION_NO_SLICE), ObjectSizes.sizeOnHeapOf(buffer));
+    };
```
ekaterinadimitrova2:
```suggestion
    }
```

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
```diff
@@ -22,128 +22,81 @@
...
-        assertThat(ObjectSizes.sizeOnHeapExcludingData(buffers)).isEqualTo(REF_ARRAY_1_SIZE + EMPTY_HEAP_BUFFER_RAW_SIZE);
+        // heap buffer being an infix slab
+        buffer = (ByteBuffer) ByteBuffer.allocate(10).duplicate().position(1).limit(8);
+        checkBufferSizeExcludingData(buffer, 7);
     }
 
+    private void checkBufferSizeExcludingData(ByteBuffer buffer, int dataSize)
+    {
+        assertEquals(meter.measureDeep(buffer, SLAB_ALLOCATION_NO_SLICE) - dataSize, ObjectSizes.sizeOnHeapExcludingDataOf(buffer));
+    };
```
ekaterinadimitrova2:
```suggestion
    }
```

test/unit/org/apache/cassandra/utils/ObjectSizesTest.java
ekaterinadimitrova2: Sure, but that is why you also added the different MemoryMeter strategy for the test MemoryMeter, which will measure the arrays in the non-optimized way so we can compare for correctness.

ekaterinadimitrova2 (commented):
There is a valid point that JDK8 is sill not dropped. Even if it will be dropped immediately after this ticket, it is worth it to confirm that test run too as jamm works with JDK8.
Started CI run here:
https://app.circleci.com/pipelines/github/ekaterinadimitrova2/cassandra/2422/workflows/ceb46a2f-c682-4830-bb14-3b0d75294f4a

ekaterinadimitrova2:
I mentioned it in Slack last week, but also adding here so it doesn't get lost.
We use `slice()` in `BufferPool` in `allocateDirectAligned.`

src/java/org/apache/cassandra/fql/FullQueryLogger.java
blerer: The original computation was ignoring in it computation a certain amount of things. You can sum up object shallow size to get the deep size but when computing the shallow size you need to be careful as the logic is not a simple sum.

blerer: I will update the documentation

src/java/org/apache/cassandra/fql/FullQueryLogger.java
blerer: should we put an assertion?

blerer: I think that it is too risky. Let's simply assume that it is ArrayList. We should be close enough from the reality.

src/java/org/apache/cassandra/utils/ObjectSizes.java
blerer: Not sure I understand your comment.

src/java/org/apache/cassandra/fql/FullQueryLogger.java
blerer: Yes it is a left over.

blerer:
The issue with `allocateDirectAligned` is that it is not what we call a SLAB but we will also not measure it correctly, we will slightly underestimate the size by the shallow size of a direct Buffer. I do not think that we can do much about it.

src/java/org/apache/cassandra/audit/BinAuditLogger.java
```diff
@@ -98,6 +99,11 @@ public void log(AuditLogEntry auditLogEntry)
     @VisibleForTesting
     public static class Message extends BinLog.ReleaseableWriteMarshallable implements WeightedQueue.Weighable
     {
+        /**
+         * The shallow size of a {@code Query} object.
```
ekaterinadimitrova2, line 103: * The shallow size of a {@code Message} object.

src/java/org/apache/cassandra/fql/FullQueryLogger.java
```diff
@@ -366,17 +366,21 @@ public void writeMarshallablePayload(WireOut wire)
...
     public static class Batch extends AbstractLogEntry
     {
         /**
-         * The shallow size of a {@code Batch} object.
+         * The shallow size of a {@code Batch} object (which includes primitive fields).
          */
         private static final long EMPTY_SIZE = ObjectSizes.measure(new Batch());
 
+        /**
+         * The way is pre-computed in the constructor and represent the object deep size.
```
ekaterinadimitrova2, line 382: * The weight is pre-computed in the constructor and represents the object's deep size.

src/java/org/apache/cassandra/fql/FullQueryLogger.java
ekaterinadimitrova2: Thanks for the additional comments, adding stuff separately is quite confusing even after looking into the MemoreMeter for months.... :(

src/java/org/apache/cassandra/utils/ObjectSizes.java
ekaterinadimitrova2: My understanding is measureArray is doing what sizeOfReferenceArray (thus the change you did here) is doing so I was wondering why then we do not replace all usages of sizeOfReferenceArray with measureArray? But now on a second read I see I misread, sizeOfReferenceArray is provided length...

ekaterinadimitrova2:
`The issue with allocateDirectAligned is that it is not what we call a SLAB but we will also not measure it correctly, we will slightly underestimate the size by the shallow size of a direct Buffer. I do not think that we can do much about it.`

I do believe the current patch just mimicked the same behavior we already have. Even if we decide to improve anything around that, it would be in a different ticket as an improvement.

ekaterinadimitrova2 (approved):
+1 ,just a few nits to be adressed on commit. All test results were published on the ticket and LGTM

test/unit/org/apache/cassandra/db/memtable/MemtableSizeTestBase.java
ekaterinadimitrova2: Ok, I just realized this is in a test, not a production class, my bad. It seems I misread the first time

PR #2490, opened 2023-07-14
CASSANDRA-18624 Interface to initialize the cryptographic provider, use ACCP as default

ayushisingh29:
* An interface ICryptoProvider is created. This will allow users to use custom providers if they don't prefer to use the default provider - ACCP. If installation/check of crypto provider fails then a fallback provider will be used and the exception will be logged as a warning message. 
 * Circle CI build: [java11](https://app.circleci.com/pipelines/github/ayushisingh29/cassandra/23/workflows/cbf44365-e3da-41c9-8fec-57175ffb4af0) [java8](https://app.circleci.com/pipelines/github/ayushisingh29/cassandra/23/workflows/b96ef0aa-609a-42d5-849f-bbb0736e5049)
 
Patch by @ayushisingh29; to be reviewed by @jrwest  and @clohfink for CASSANDRA-18624

[Cassandra-18624](https://issues.apache.org/jira/browse/CASSANDRA-18624)


.build/parent-pom-template.xml
```diff
@@ -774,6 +774,13 @@
       </dependency>
       <!-- end of chronicle-queue -->
 
+      <dependency>
+        <groupId>software.amazon.cryptools</groupId>
+        <artifactId>AmazonCorrettoCryptoProvider</artifactId>
+        <version>[2.0, 3.0)</version>
```
smiklosovic: This is most probably wrong, we can not use range versions. We need repeatable builds so we have to make it version specific.

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -79,6 +79,7 @@ public static Set<String> splitCommaDelimited(String src)
     public String authenticator;
     public String authorizer;
     public String role_manager;
+    public String crypto_provider;
```
smiklosovic: This should be of type `ParameterizedClass`. That way it will be configured same way as e.g. `seed_provider`.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1500,6 +1506,11 @@ private static IFailureDetector createFailureDetector(String detectorClassName)
         return detector;
     }
 
+    public static ICryptoProvider getCryptoProvider() {return cryptoProvider;}
+
+    public void setCryptoProvider(ICryptoProvider cryptoProvider) {
```
smiklosovic: The opening brace should be on a new line.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1500,6 +1506,11 @@ private static IFailureDetector createFailureDetector(String detectorClassName)
         return detector;
     }
 
+    public static ICryptoProvider getCryptoProvider() {return cryptoProvider;}
```
smiklosovic: braces / fromatting.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -877,6 +880,9 @@ else if (conf.commitlog_segment_size.toMebibytes() >= 2048)
         else if (conf.commitlog_segment_size.toKibibytes() < 2 * conf.max_mutation_size.toKibibytes())
             throw new ConfigurationException("commitlog_segment_size must be at least twice the size of max_mutation_size / 1024", false);
 
+        if (conf.crypto_provider != null)
```
smiklosovic: The initialization logic should be rewritten, try to look at `applySeedProvider();` how it is done to take an inspiration from there.

src/java/org/apache/cassandra/service/AmazonCorrettoCryptoProviderImpl.java
```diff
@@ -0,0 +1,58 @@
...
+
+import javax.crypto.Cipher;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.cassandra.exceptions.StartupException;
+import com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider;
+
+public class AmazonCorrettoCryptoProviderImpl implements ICryptoProvider
```
smiklosovic: I think this should be renamed to `DefaultCryptoProvider`. If we have different crypto providers available with some names, how do you know which one is the default one?

The class should be moved to `org.apache.cassandra.security`

src/java/org/apache/cassandra/service/AmazonCorrettoCryptoProviderImpl.java
```diff
@@ -0,0 +1,58 @@
...
+            AmazonCorrettoCryptoProvider.install();
+        }
+        catch(Exception e)
+        {
+            logger.warn("Amazon Corretto Crypto Provider is not available", e);
+        }
+    }
+
+    @Override
+    public void checkProvider() throws Exception
```
smiklosovic: Opening braces should be on new lines in the whole method.

src/java/org/apache/cassandra/service/AmazonCorrettoCryptoProviderImpl.java
```diff
@@ -0,0 +1,58 @@
...
+    {
+        try {
+            if (Cipher.getInstance("AES/GCM/NoPadding").getProvider().getName().equals(AmazonCorrettoCryptoProvider.PROVIDER_NAME)) {
```
smiklosovic: What if `Cipher.getInstance("AES/GCM/NoPadding").getProvider().getName()` returns no provider so it will be null? Can that happen?

src/java/org/apache/cassandra/service/AmazonCorrettoCryptoProviderImpl.java
```diff
@@ -0,0 +1,58 @@
...
+                AmazonCorrettoCryptoProvider.INSTANCE.assertHealthy();
+            } else {
+                logger.warn("ACCP is not the highest priority provider");
```
smiklosovic: ACCP should be more verbose, Use `AmazonCorrettoCryptoProvider.class.getName()` in this warn message.

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -259,6 +259,8 @@ protected void setup()
 
         CommitLog.instance.start();
 
+        installCryptoProvider();
```
smiklosovic, line 262: Why not to do this installation logic in `CassandraDaemon.activate()` method? There is `applyConfig()` at the very beginning. Then it will eventually propagate to `DatabaseDescriptor.applyAll()`.

src/java/org/apache/cassandra/service/AmazonCorrettoCryptoProviderImpl.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class AmazonCorrettoCryptoProviderImpl implements ICryptoProvider
+{
+    private static final Logger logger = LoggerFactory.getLogger(AmazonCorrettoCryptoProviderImpl.class);
+    @Override
+    public void installProvider() throws StartupException
+    {
+        try
+        {
+            AmazonCorrettoCryptoProvider.install();
```
smiklosovic: Why not to install the provider in `DatabaseDescriptor.applyAll` ? E.g. before `applySeedProvider`? After it is installed, there are startup checks done. So the startup check will be done on installed provider.

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -488,6 +490,18 @@ protected void setup()
         completeSetup();
     }
 
+    public void installCryptoProvider()
```
smiklosovic: As mentioned above, this is not necessary. You can install the provider as part of `DatabaseDescriptor.applyAll` and startup checks will be run on that afterwards.

test/unit/org/apache/cassandra/config/DatabaseDescriptorRefTest.java
```diff
@@ -258,7 +258,9 @@ public class DatabaseDescriptorRefTest
     "org.apache.cassandra.security.EncryptionContext",
     "org.apache.cassandra.security.ISslContextFactory",
     "org.apache.cassandra.security.SSLFactory",
+    "org.apache.cassandra.service.AmazonCorrettoCryptoProviderImpl",
```
smiklosovic: We will most probably not need this after suggested changes are incorporated.

src/java/org/apache/cassandra/service/ICryptoProvider.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package org.apache.cassandra.service;
+
+import org.apache.cassandra.exceptions.StartupException;
+
+/**
+ * Interface for managing the cryptographic provider
+ */
+public interface ICryptoProvider
```
smiklosovic: This should be moved to `org.apache.cassandra.security` package.

src/java/org/apache/cassandra/utils/FBUtilities.java
```diff
@@ -634,6 +635,13 @@ static IPartitioner newPartitioner(String partitionerClassName, Optional<Abstrac
         return FBUtilities.instanceOrConstruct(partitionerClassName, "partitioner");
     }
 
+    public static ICryptoProvider newCryptoProvider(String className) throws ConfigurationException
+    {
+        if (!className.contains("."))
+            className = "org.apache.cassandra.service." + className;
```
smiklosovic: dont forget to replace `service` with `security`. This method should also accept a map of parameters. Look into `newSslContextFactory` in this class how that is done.

src/java/org/apache/cassandra/service/AmazonCorrettoCryptoProviderImpl.java
```diff
@@ -0,0 +1,58 @@
...
+        }
```
xgerman: Please add any of the verifications as described here, too: https://github.com/corretto/amazon-corretto-crypto-provider#verification-optional

I would be ok with `AmazonCorrettoCryptoProvider.INSTANCE.assertHealthy();` which throws a runtime exception

src/java/org/apache/cassandra/config/Config.java
ayushisingh29: What's the benefit of making it` ParameterizedClass`? We don't need any other input other than the provider name.

src/java/org/apache/cassandra/service/AmazonCorrettoCryptoProviderImpl.java
ayushisingh29: I was not able to reproduce a case where it is null. It has always returned a provider name. Try-catch block will catch the exception in case of any.

src/java/org/apache/cassandra/service/AmazonCorrettoCryptoProviderImpl.java
xgerman: Thanks!!

.build/parent-pom-template.xml
```diff
@@ -774,6 +774,13 @@
       </dependency>
       <!-- end of chronicle-queue -->
 
+      <dependency>
+        <groupId>software.amazon.cryptools</groupId>
+        <artifactId>AmazonCorrettoCryptoProvider</artifactId>
+        <version>2.2.0</version>
+        <classifier>linux-x86_64</classifier>
```
smiklosovic, line 781: @ayushisingh29 what other classifiers there are? I think there is also classifier for arm? So what does this mean for the project? Because if we depend here on x86, what about people who want to run Cassandra on arm?

src/java/org/apache/cassandra/config/Config.java
smiklosovic: The benefit is that it is configurable. Are you 100% sure that there will never be a provider which is configurable? Your provider is not, are you sure you can say it about any possible provider yet to be implemented? 

Also, it is very easy to do that in terms of how it looks like in cassandra.yaml. Instead of this

    crypto_provider: this.is.my.ClassImpl

It would be like

    crypto_provider:
      - class_name: this.is.my.ClassImpl

I do not think it is a lot to ask given the fact that we open ourselves to a lot of possibilities when it is potentially configurable.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1217,6 +1221,25 @@ public static void applySslContext()
         }
     }
 
+    public static void applyCryptoProvider()
+    {
+        try
+        {
+            if (conf.crypto_provider == null)
+                conf.crypto_provider = new ParameterizedClass("org.apache.cassandra.security.DefaultCryptoProvider", null);
```
smiklosovic: @ayushisingh29 would it be possible to use 'DefaultCryptoProvider.class.getName()' instead of putting there just string? What if the name of that provider changes or the package it is located in changes? Then you would need to fix it here as well ... Using class.getName(), you have it for free.

Does it interfere with DatabaseDescriptorRefTest or something similar that it can not be used like that?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1217,6 +1221,25 @@ public static void applySslContext()
...
+
+            Class<?> cryptoProviderClass = Class.forName(conf.crypto_provider.class_name);
```
smiklosovic, line 1235: @ayushisingh29 what if `class_name` is null? Can't you have `conf.crypto_provider` non-null but `class_name` would be null? What happens if you have this in yaml?

    crypto_provider:
        - parameters:
            - a: "b"

I would check if `class_name` is null and if it is we either error out or we set it to default.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1217,6 +1221,25 @@ public static void applySslContext()
...
+            cryptoProvider = (ICryptoProvider)cryptoProviderClass.getConstructor(Map.class).newInstance(conf.crypto_provider.parameters);
+
+            cryptoProvider.installProvider();
+        }
+        catch(Exception e)
+        {
+            throw new ConfigurationException("Failed to initialize crypto Provider.", e);
+        }
+
```
smiklosovic: redundant empty line

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1217,6 +1221,25 @@ public static void applySslContext()
...
+        {
+            if (conf.crypto_provider == null)
+                conf.crypto_provider = new ParameterizedClass("org.apache.cassandra.security.DefaultCryptoProvider", null);
+
+            Class<?> cryptoProviderClass = Class.forName(conf.crypto_provider.class_name);
+            cryptoProvider = (ICryptoProvider)cryptoProviderClass.getConstructor(Map.class).newInstance(conf.crypto_provider.parameters);
+
+            cryptoProvider.installProvider();
+        }
+        catch(Exception e)
```
smiklosovic: `catch (Exception e)` - there should be space between `catch` and `(Exception`

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1217,6 +1221,25 @@ public static void applySslContext()
...
+        {
+            throw new ConfigurationException("Failed to initialize crypto Provider.", e);
```
smiklosovic: It would be nice to say what was the class (from class_name) of the provider we failed to initialize.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1500,6 +1523,12 @@ private static IFailureDetector createFailureDetector(String detectorClassName)
         return detector;
     }
 
+    public static ICryptoProvider getCryptoProvider() { return cryptoProvider; }
+
+    public void setCryptoProvider(ICryptoProvider cryptoProvider)
+    {
+        cryptoProvider = cryptoProvider;
+    }
```
smiklosovic, line 1537: separate these method by empty line between them

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
```diff
@@ -0,0 +1,68 @@
...
+    public DefaultCryptoProvider(Map<String, String> args) {}
+    @Override
+    public void installProvider() throws StartupException
+    {
+        try
+        {
+            AmazonCorrettoCryptoProvider.install();
+            AmazonCorrettoCryptoProvider.INSTANCE.assertHealthy();
+        }
+        catch(Exception e)
```
smiklosovic: put space between `catch` and `(Exception`

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.cassandra.exceptions.StartupException;
+import com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider;
+
+public class DefaultCryptoProvider implements ICryptoProvider
+{
+    private static final Logger logger = LoggerFactory.getLogger(DefaultCryptoProvider.class);
+
+    public DefaultCryptoProvider(Map<String, String> args) {}
+    @Override
+    public void installProvider() throws StartupException
```
smiklosovic: This method does not throw any `StartupException`, does it? I see that you are trying to catch the exception in `applyCryptoProvider`. That is OK but I do not see you are actually rethrowing `e` here after you catch it. You just log it. You need to rethrow.

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
```diff
@@ -0,0 +1,68 @@
...
+                AmazonCorrettoCryptoProvider.INSTANCE.assertHealthy();
+            }
+            else
+            {
+                logger.warn("{} is not the highest priority provider", AmazonCorrettoCryptoProvider.class.getName());
+            }
+        }
+        catch (Exception e)
+        {
+            logger.warn("Corretto Crypto Provider Error", e);
```
smiklosovic: @ayushisingh29 his method just swallows the exception and nothing is ever thrown from this method. You need to rethrow `e` if you want to catch it in the startup check.

src/java/org/apache/cassandra/security/ICryptoProvider.java
```diff
@@ -0,0 +1,40 @@
...
+ * Interface for managing the cryptographic provider
+ */
+public interface ICryptoProvider
+{
+
+    /**
+     * Installs a cryptographic provider of a specific type.
+     * @throws StartupException if an error occurs while installing the provider
+     */
+    public void installProvider() throws StartupException;
```
smiklosovic: This should probably throw `Exception` if we want to have it like checked exception a caller needs to catch every time.

.build/parent-pom-template.xml
ayushisingh29: The other classifier is `linux-aarch_64` which can be used to use ACCP on Linux ARM64 platforms. It is necessary to provide the classifier otherwise we will get an empty package. In this case it will not work for ARM.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
ayushisingh29: No issues with it. It is better to use class.getName(). I will update it.

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
ayushisingh29: It was done intentionally for cases where ACCP was not installed properly. We don't want to stop Cassandra from booting up, so it will use another provider as a fallback: "sun.security.provider.Sun". I have added exception to the interface to give flexibility to others in how they want to handle the implementations of their custom provider. I will update it to throw Exception and not StartupException.

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
ayushisingh29: Same reason as above.

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
smiklosovic: Interesting. I think that if we fail the installation of the provider, we should not just silently continue, we should truly fail the startup. A user asked for a crypto_provider he did not get.

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
smiklosovic: same as above

.build/parent-pom-template.xml
smiklosovic: yeah, check the discussion on the ticket, I am afraid we can not ship anything architecture specific. We need to keep vanilla Cassandra architecture agnostic.

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
xgerman: 100% yes, this is a pet peeve of mine. Something is not configured right and the startup continues and you will end up with a Cassandra which is substantially slower. If necessary make it an option so people can configure it the way they need it.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1217,6 +1222,27 @@ public static void applySslContext()
...
     }
 
+    public static void applyCryptoProvider()
+    {
+        try
+        {
+            if (conf.crypto_provider == null)
+                conf.crypto_provider = new ParameterizedClass(NoOpCryptoProvider.class.getName(), null);
+
+            if (conf.crypto_provider.class_name == null)
```
smiklosovic: @ayushisingh29 if you throw an exception because `class_name` is null, it will be caught in the catch block but there you are referencing `conf.crypto_provider.class_name` and you call `toString()` on it. So while you caught the exception, it may throw `NullPointerException` in that catch block again.

ayushisingh29:
Thanks @smiklosovic for the suggestions. I have cherry-picked the improvements you made and also created the flag to fail on error.

smiklosovic:
I am closing this in favor of https://github.com/apache/cassandra/pull/2505 where we are building on top of this work.

PR #2492, opened 2023-07-14
CASSANDRA-18670 Importer should build SSTable indexes successfully before making new …

maedhroz:
…SSTables readable

patch by Caleb Rackliffe; reviewed by ? for CASSANDRA-18670


maedhroz:
https://app.circleci.com/pipelines/github/maedhroz/cassandra?branch=CASSANDRA-18670

src/java/org/apache/cassandra/db/SSTableImporter.java
```diff
@@ -186,6 +186,8 @@ synchronized List<String> importNewSSTables(Options options)
         try (Refs<SSTableReader> refs = Refs.ref(newSSTables))
         {
             abortIfDraining();
+
+            cfs.indexManager.buildIndexesBlockingIncremental(newSSTables);
```
maedhroz: This is the crux of the change. When we have SSTable-attached indexes, they should always be build here, before we add the new SSTables to the tracker. That way we a.) fail the import before the SSTables become readable if there's an index building failure and b.) we have the index view update almost immediately after the base table view (which isn't perfect, as there could still be a very small window of inconsistency, but it's a huge improvement).

TODO: This probably would benefit from a comment to this effect.

src/java/org/apache/cassandra/index/Index.java
```diff
@@ -341,13 +341,21 @@ default public Callable<?> getPreJoinTask(boolean hadBootstrap)
...
      * SSTable data.
      * @return if the index should be included in the set which processes SSTable data, false otherwise.
      */
-    public boolean shouldBuildBlocking();
+    boolean shouldBuildBlocking();
+    
+    /**
+     * @return true if the index builds SSTable-attached on-disk components
+     */
+    default boolean isSSTableAttached()
```
maedhroz, line 362: This method allows the SIM to figure out which indexes it should try to build incrementally, and which it should ignore for partial rebuilds actuated via `SSTableAdded` notifications.

TODO: A comment that explains what SSTable-attached indexes are expected to do. In other words, they must be using an `SSTableFlushObserver` to make sure they build during streaming and be able to build SSTable-specific indexes incrementally in the build tasks they provide.

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
```diff
@@ -490,10 +490,58 @@ public static String getIndexName(String cfName)
         assert isIndexColumnFamily(cfName);
         return StringUtils.substringAfter(cfName, Directories.SECONDARY_INDEX_NAME_SEPARATOR);
     }
+    
+    public void buildIndexesBlockingIncremental(Collection<SSTableReader> sstables)
```
maedhroz: I'll JavaDoc this when we're closer having the patch approved, but basically it's a version of `buildIndexesBlocking()` that doesn't touch the index queryability/status. It is here to build indexes for new SSTables before they are actually added to the tracker, and so failures have no bearing on those things...except, of course, for failing the upstream operation and making sure the SSTables themselves never become readable.

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
```diff
@@ -490,10 +490,58 @@ public static String getIndexName(String cfName)
...
+                public void onFailure(Throwable t)
+                {
+                    logger.warn("Failed to build indexes {}", getIndexNames(groupedIndexes));
+                    build.tryFailure(t);
+                }
+
+                @Override
+                public void onSuccess(Object o)
+                {
+                    logger.info("Index build of {} completed", getIndexNames(groupedIndexes));
```
maedhroz:
```suggestion
                    logger.info("Incremental index build of {} completed", getIndexNames(groupedIndexes));
```

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
maedhroz: Open to any suggestions on the logging, structure, possibilities for deduplication, etc.

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
```diff
@@ -1706,6 +1754,7 @@ public void handleNotification(INotification notification, Object sender)
                                      indexes.values()
                                             .stream()
                                             .filter(Index::shouldBuildBlocking)
+                                            .filter(i -> !i.isSSTableAttached())
```
maedhroz, line 1797: If we're here, it's either after streaming or via `SSTableImporter`. SAI will already have written its indexes in the streaming case (entire file streaming, or writing incrementally via `SSTableFlushObserver`), and this patch redirects the importer case.

So, this is here to just make sure we ignore SAI indexes.

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexBuilder.java
```diff
@@ -205,7 +205,6 @@ private boolean indexSSTable(SSTableReader sstable, Set<StorageAttachedIndex> in
 
             if (t instanceof InterruptedException)
             {
-                // TODO: Is there anything that makes more sense than just restoring the interrupt?
```
maedhroz, line 208: No sure what else we would do here...

test/distributed/org/apache/cassandra/distributed/test/sai/ImportIndexedSSTablesTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+            Object[][] rs = first.executeInternal(withKeyspace("select pk from %s.test where pk = ?"), 1);
+            assertThat(rs.length).isEqualTo(1);
+
+            first.runOnInstance(() -> {
+                ColumnFamilyStore cfs = Keyspace.open(KEYSPACE).getColumnFamilyStore("test");
+                cfs.clearUnsafe();
+            });
+
+            cluster.schemaChange(withKeyspace("CREATE CUSTOM INDEX test_v_index ON %s.test(v) USING 'StorageAttachedIndex'"));
```
maedhroz: There's a failure in https://app.circleci.com/pipelines/github/maedhroz/cassandra/738/workflows/58e74f88-40d3-4776-9c20-9020a982966e/jobs/9901/tests#failed-test-0 that might be related to not waiting for the index to be ready here...

src/java/org/apache/cassandra/index/sai/disk/StorageAttachedIndexWriter.java
```diff
@@ -212,9 +208,10 @@ public void abort(Throwable accumulator, boolean fromIndex)
         // Mark the write operation aborted, so we can short-circuit any further operations on the component writers.
         aborted = true;
         
+        // TODO: This really shouldn't be necessary if we fail in the right places?
         // Make any indexes involved in this transaction non-queryable, as they will likely not match the backing table.
-        if (fromIndex)
-            indexes.forEach(StorageAttachedIndex::makeIndexNonQueryable);
+        //if (makeNonQueryable)
+        //    indexes.forEach(StorageAttachedIndex::makeIndexNonQueryable);
```
maedhroz: The changes in `abort()` here are probably what require some real discussion. The position I'm going to try to defend here is that making the index non-queryable is sub-optimal in all cases if our goal is to keep indexes and backing tables consistent w/ one another. (It is breaking a couple tests right now, but I'm working on that...)

(I don't see any problem w/ continuing to not throw when the abort is upstream is.)

@adelapena @mike-tr-adamson

src/java/org/apache/cassandra/index/sai/disk/StorageAttachedIndexWriter.java
```diff
@@ -209,13 +204,11 @@ public void abort(Throwable accumulator)
...
     {
+        if (aborted) return;
+
         // Mark the write operation aborted, so we can short-circuit any further operations on the component writers.
         aborted = true;
         
-        // Make any indexes involved in this transaction non-queryable, as they will likely not match the backing table.
-        if (fromIndex)
-            indexes.forEach(StorageAttachedIndex::makeIndexNonQueryable);
-
```
maedhroz, line 218: The changes in `abort()` here are probably what require some real discussion. The position I'm going to try to defend here is that making the index non-queryable is sub-optimal in all cases if our goal is to keep indexes and backing tables consistent w/ one another. (I've fixed `SegmentFlushingFailureTester` to reflect this, but more tests are probably necessary around Memtable flush...not hard to do.)

(I don't see any problem w/ continuing to not throw when the abort is upstream is generated from outside the index itself.)

For this to work, the companion changes in `SSTableWriter` are here: https://github.com/apache/cassandra/pull/2492/files#diff-27996cac323bd3bafcfd28e84caf649d44f3b4d83477874ae4d08c7513303209R234

@adelapena @mike-tr-adamson

src/java/org/apache/cassandra/io/sstable/format/SSTableWriter.java
```diff
@@ -226,12 +226,15 @@ public final Throwable commit(Throwable accumulate)
...
         }
-        finally
+        catch (Throwable t)
         {
-            observers.forEach(SSTableFlushObserver::complete);
+            // Return without advancing to COMMITTED, which will trigger abort() when the Transactional closes...
+            return Throwables.merge(accumulate, t);
         }
+
+        return txnProxy.commit(accumulate);
```
maedhroz, line 237: The change here in some way mirrors what already happens for streaming and what now happens for import in this patch. We make sure index components are complete before proceeding w/ making the related SSTables visible for reads, etc.

src/java/org/apache/cassandra/index/sai/disk/StorageAttachedIndexWriter.java
maedhroz: Haven't hit Memtable flush yet, but just added `IndexStreamingFailureTest`, which documents how SAI handles index build failures during streaming in the non-entire file case.

src/java/org/apache/cassandra/io/sstable/format/SSTableWriter.java
```diff
@@ -201,8 +201,8 @@ public void setOpenResult(boolean openResult)
     public SSTableReader finish(boolean openResult)
     {
         this.setOpenResult(openResult);
-        txnProxy.finish();
         observers.forEach(SSTableFlushObserver::complete);
+        txnProxy.finish();
```
maedhroz, line 205: If we get a failure in `complete()`, and we've already called `finish()`, we'll get an illegal state...i.e. that we'll try to abort after committing due to the error from `complete()`.

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
```diff
@@ -242,6 +246,19 @@ public void finished()
             }
             else
             {
+                if (receivedEntireSSTable)
```
maedhroz, line 251: I could add a bit more inline commentary to this, but when we don't stream the entire SSTable, validation doesn't make much sense, as the SAI components have just been successfully written when we arrive here. Any error doing that would have already aborted the streaming txn.

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
```diff
@@ -242,6 +246,19 @@ public void finished()
...
+                {
+                    try
+                    {
+                        // SSTable-attached indexes may have been streamed completely, and they must be validated:
+                        cfs.indexManager.validateStorageAttachedIndexes(readers);
+                    }
+                    catch (IOException e)
+                    {
+                        throw new RuntimeException(e);
```
maedhroz: At this point, I'm sure sure if there's a better alternative than `RuntimeException`, but open to suggestions...

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
```diff
@@ -491,9 +497,83 @@ public static String getIndexName(String cfName)
         return StringUtils.substringAfter(cfName, Directories.SECONDARY_INDEX_NAME_SEPARATOR);
     }
 
+    // TODO: JavaDoc if this approach survives...mention that this should not be used for live SSTables
+    public void validateStorageAttachedIndexes(Collection<SSTableReader> sstables) throws IOException
```
maedhroz: Right now, this method exists solely to validate SAI components streamed entirely before we allow a streaming transaction to commit. It therefore replaces any such validation that would occur via notification in `StorageAttachedIndexGroup` (which should only worry about adding the new SSTable indexes to the view, etc.).

This might also be useful for import, however, in the case where we bring in new SSTables that already have built index components. We'd just have to make sure we only did it in this case and not when we need to build them anyway...

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
```diff
@@ -491,9 +497,83 @@ public static String getIndexName(String cfName)
...
+        {
+            IndexDescriptor indexDescriptor = IndexDescriptor.create(sstable);
+
+            if (!indexDescriptor.isPerSSTableIndexBuildComplete())
+                throw new IllegalStateException(indexDescriptor.logMessage("Incomplete per-SSTable index"));
+
+            indexDescriptor.checksumPerSSTableComponents();
+
+            for (StorageAttachedIndex index : toValidate)
+                indexDescriptor.checksumPerIndexComponents(index.getIndexContext());
```
maedhroz: TODO: Might want to check the existence of the per-column components here to avoid any useless validation...

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
maedhroz: I think it would be possible to just add an argument indicating a failure mode to this method. Streaming could fail hard if missing anything in the entire SSTable case, and import could just validate what's there, given the rest will be built anyway immediately after.

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
```diff
@@ -491,9 +497,83 @@ public static String getIndexName(String cfName)
         return StringUtils.substringAfter(cfName, Directories.SECONDARY_INDEX_NAME_SEPARATOR);
     }
 
+    // TODO: JavaDoc if this approach survives...mention that this should not be used for live SSTables
+    public void validateStorageAttachedIndexes(Collection<SSTableReader> sstables) throws IOException
+    {
+        // TODO: A more generic approach via the Index interface rather than just working w/ SAI
```
adelapena: Indeed the entire method body seems SAI-specific, so I guess it should be in the Index interface.

maedhroz: Perhaps this is something I can push down to the `Group` interface, then have implemented by SAI and no-op'd by everything else for now...or validate only for SSTable-attached things. On `SIM` the method could be something like `validateSSTableAttachedIndexes()`? (Still need to access the `Group` somehow, so I guess `SIM` is the facade.)

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
```diff
@@ -0,0 +1,160 @@
...
+            assertThat(rs.length).isEqualTo(0);
+
+            // ...and the index should remain queryable, because from its perspective, the streaming never happened.
+            Boolean isQueryable = second.callOnInstance(() -> {
+                SecondaryIndexManager sim = Keyspace.open(KEYSPACE).getColumnFamilyStore("test").indexManager;
+                Index index = sim.getIndexByName("test_v_index");
+                return sim.isIndexQueryable(index);
+            });
+
+            assertTrue("The index should be queryable on restart!", isQueryable);
```
maedhroz: TODO: Given I'm querying the index immediately after this, the check for queryability is probably redundant.

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
```diff
@@ -249,9 +249,9 @@ public void handleNotification(INotification notification, Object sender)
         {
             SSTableAddedNotification notice = (SSTableAddedNotification) notification;
 
-            // Avoid validation for index files just written following Memtable flush.
-            IndexValidation validate = notice.memtable().isPresent() ? IndexValidation.NONE : IndexValidation.CHECKSUM;
-            onSSTableChanged(Collections.emptySet(), notice.added, indexes, validate);
+            // Avoid validation for index files just written following Memtable flush. Otherwise, the new SSTables have
+            // come either from import, streaming, or a standalone tool, where they have also already been validated.
+            onSSTableChanged(Collections.emptySet(), notice.added, indexes, IndexValidation.NONE);
```
maedhroz, line 254: With this change, I think we can no longer get into the trouble we might find ourselves in w/ https://github.com/apache/cassandra/pull/2460#discussion_r1253762041, and the comment explains why we're not validating at all.

Once we get here, index components should have been built and such, so the only thing we're doing is updating the views and potentially marking the index non-queryable, which I think is okay. (i.e. Marking the index non-queryable is the only thing that we can really do, but what I need to confirm is whether we're being strict enough around that.)

Thoughts?

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
adelapena: I think `RuntimeException` is fine.

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
adelapena: A comment saying that about not-entire sstables would probably be helpful.

src/java/org/apache/cassandra/index/Index.java
adelapena: A comment would make sense. If SASI or table-based indexes had more future than they have we could even subclass `Index` and have a generic `SSTableAttachedIndex` API, implemented by SAI and SASI. But it probably doesn't worth the additional complexity at this point.

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
adelapena: Probably SIM should call a validation method in each `Group`, and the `Group` would call another validation method on each of its indexes. That way the method on SIM doesn't need to know anything about `StorageAttachedIndex`, and can have a more generic name such as `validateStreamedIndexes` or something like that?

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
```diff
@@ -491,9 +497,108 @@ public static String getIndexName(String cfName)
...
+        byType.forEach((buildingSupport, groupedIndexes) ->
+        {
+            SecondaryIndexBuilder builder = buildingSupport.getIndexBuildTask(baseCfs, groupedIndexes, sstables, false);
+            AsyncPromise<Object> build = new AsyncPromise<>();
+            CompactionManager.instance.submitIndexBuild(builder).addCallback(new FutureCallback<Object>()
+            {
+                @Override
+                public void onFailure(Throwable t)
+                {
+                    logger.warn("Failed to build indexes {}", getIndexNames(groupedIndexes));
```
adelapena:
```suggestion
                    logger.warn("Failed to incrementally build indexes {}", getIndexNames(groupedIndexes));
```

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
adelapena: This will also ignore SASI, won't it?

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexBuilder.java
```diff
@@ -99,7 +99,8 @@ public class StorageAttachedIndexBuilder extends SecondaryIndexBuilder
     @Override
     public void build()
     {
-        logger.debug(logMessage("Starting full index build"));
+        logger.debug(logMessage(String.format("Starting %s %s index build...", isInitialBuild ? "initial" : "non-initial", isFullRebuild ? "full" : "partial")));
```
adelapena: Nit: can we break the long line?
```suggestion
        logger.debug(logMessage(String.format("Starting %s %s index build...",
                                              isInitialBuild ? "initial" : "non-initial",
                                              isFullRebuild ? "full" : "partial")));
```

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
```diff
@@ -491,9 +497,108 @@ public static String getIndexName(String cfName)
...
+        
+        boolean complete = true;
+
+        for (SSTableReader sstable : sstables)
+        {
+            IndexDescriptor indexDescriptor = IndexDescriptor.create(sstable);
+
+            if (indexDescriptor.isPerSSTableIndexBuildComplete())
+            {
+                indexDescriptor.checksumPerSSTableComponents();
```
adelapena: Couldn't this use `indexDescriptor#validatePerSSTableComponents(IndexValidation.CHECKSUM)`, so we don't need the special `IndexDescriptor#checksumPerSSTableComponents` method? And the same for per index components.

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
adelapena: Or maybe we want to add our own `CurruptIndexException`, but I don't think we have something particular to do with it.

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
```diff
@@ -0,0 +1,160 @@
...
+            rs = second.executeInternal(withKeyspace("select pk from %s.test where v = ?"), "v1");
+            assertThat(rs.length).isEqualTo(0);
```
adelapena: Maybe we could extend the test case a bit by deactivating the byteman rule and running a second repair. Then we would verify that this time the data arrives at the second node and it's queryable. Something like:
```java
second.runOnInstance(()-> ByteBuddyHelper.fail.set(false));
result = second.nodetoolResult("repair", KEYSPACE);
result.asserts().success();

rs = second.executeInternal(withKeyspace("select pk from %s.test where v = ?"), "v1");
assertThat(rs.length).isEqualTo(1);
```
This would be just to ensure we don't enter into some kind of irreversible state after a failure, if that makes sense.

src/java/org/apache/cassandra/index/Index.java
maedhroz: Yeah, I think we're all hesitant to make many more changes around this or the index API when we know those two are basically defunct.

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
maedhroz: Yeah...I think for the time being, I'm going to just have `SASIIndex` return false from `isSSTableAttached()` (the default). There's no point in changing it and worrying about whether SASI works the same way for `SSTableImporter` w/ more tests, etc.

By the way, `SSTableFlushObserver` also ensures that SAI indexes build when they should for the SSTable scrubber, upgrader, and splitter tools.

test/distributed/org/apache/cassandra/distributed/test/sai/ImportIndexedSSTablesTest.java
maedhroz: Added a wait.

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
maedhroz: It would be nice if Lucene's `CorruptIndexException` wasn't an `IOException`.

Actually, we could just wrap it in an `UncheckedIOException`.

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
maedhroz: This is what I ended up doing.

i.e. `boolean validateStorageAttachedIndexes(Collection<SSTableReader> sstables, boolean failIncomplete) throws IOException`

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
maedhroz: Those two methods don't preserve the `IOException` throw from the validate methods in `OnDiskFormat`. We could change those methods to throw as well, but then the builder and context manager that call them would have to handle the exception, etc.

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
maedhroz: This is basically what I ended up doing.

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
```diff
@@ -0,0 +1,181 @@
...
+    {
+        testAvailabilityAfterStreaming(ByteBuddyHelper::installValidationError, true);
+    }
+
+    private void testAvailabilityAfterStreaming(BiConsumer<ClassLoader, Integer> injectedError, boolean streamEntireSSTables) throws Exception
+    {
+        try (Cluster cluster = init(Cluster.build(2)
+                                           .withConfig(c -> c.with(NETWORK, GOSSIP).set("stream_entire_sstables", streamEntireSSTables))
+                                           .withInstanceInitializer(injectedError)
+                                           .start()))
```
maedhroz: TODO: This test might be a little on the expensive side, so I might try to have the two cases share the same cluster. It should be possible w/ a couple flags...

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
adelapena: `UncheckedIOException` looks like what we need here, +1

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
adelapena: Sharing the cluster would be great for CI costs and risk of timing out.

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
```diff
@@ -0,0 +1,181 @@
...
+    @Test
+    public void testAvailabilityAfterFailedEntireFileStreaming() throws Exception
+    {
+        testAvailabilityAfterStreaming(ByteBuddyHelper::installValidationError, true);
+    }
+
+    private void testAvailabilityAfterStreaming(BiConsumer<ClassLoader, Integer> injectedError, boolean streamEntireSSTables) throws Exception
+    {
+        try (Cluster cluster = init(Cluster.build(2)
+                                           .withConfig(c -> c.with(NETWORK, GOSSIP).set("stream_entire_sstables", streamEntireSSTables))
```
adelapena: Do we need `GOSSIP`? Removing it makes the test almost twice as fast, at least locally.

maedhroz:
There's one failure here I think I've caused: https://app.circleci.com/pipelines/github/maedhroz/cassandra/746/workflows/ac4f54f6-4cd6-4884-8204-366a8be43525/jobs/10985/tests#failed-test-0

Investigating...

EDIT: Fixes this and cleaned up the checked exceptions situation in the process :)

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
maedhroz: You're right. Removed...and changed things a bit so we can reuse the cluster...

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
maedhroz: Added JavaDoc.

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
mike-tr-adamson: This is close to the change that I added in my original patch for CASSANDRA-18653. At this point, we should already know that the indexes are built and valid. If they aren't then we aren't in a position to build them anyway. 

I'm wondering whether we always need to do this here or whether we only need to do it if a memtable is present. The SIM handles the notification as well and will trigger a build if the memtable is not present.

src/java/org/apache/cassandra/index/sai/disk/StorageAttachedIndexWriter.java
```diff
@@ -74,19 +71,17 @@ public static StorageAttachedIndexWriter createBuilderWriter(IndexDescriptor ind
...
     private StorageAttachedIndexWriter(IndexDescriptor indexDescriptor,
                                        Collection<StorageAttachedIndex> indexes,
                                        LifecycleNewTracker lifecycleNewTracker,
-                                       boolean perIndexComponentsOnly,
-                                       boolean propagateErrors) throws IOException
+                                       boolean perIndexComponentsOnly) throws IOException
     {
         this.indexDescriptor = indexDescriptor;
-        this.indexes = indexes;
+        //this.indexes = indexes;
```
mike-tr-adamson: nit: Can remove

src/java/org/apache/cassandra/index/sai/disk/StorageAttachedIndexWriter.java
mike-tr-adamson: I'm in favour of this change. I never felt that it should be down to the index to say whether or not it is queryable. Particularly at this point in the writer. Short of validating the index here we have no idea whether it is valid or not.

src/java/org/apache/cassandra/index/sai/disk/format/OnDiskFormat.java
```diff
@@ -115,22 +115,18 @@ PerColumnIndexWriter newPerColumnIndexWriter(StorageAttachedIndex index,
...
      * @param indexDescriptor The {@link IndexDescriptor} for the SSTable SAI index
      * @param indexContext The {@link IndexContext} holding the per-index information for the index
      * @param checksum {@code true} if the checksum should be tested as part of the validation
-     *
-     * @return true if all the per-column components are valid
      */
-    boolean validatePerColumnIndexComponents(IndexDescriptor indexDescriptor, IndexContext indexContext, boolean checksum);
-
+    void validatePerColumnIndexComponents(IndexDescriptor indexDescriptor, IndexContext indexContext, boolean checksum);
+
```
mike-tr-adamson: nit: empty space.

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
```diff
@@ -242,6 +245,12 @@ public void finished()
             }
             else
             {
+                // Validate SSTable-attached indexes that should have streamed in an already complete state. When we
+                // don't stream the entire SSTable, validation is unnecessary, as the indexes have just been written
+                // via the SSTable flush observer, and an error there would have aborted the streaming transaction.
+                if (receivedEntireSSTable)
+                    cfs.indexManager.validateSSTableAttachedIndexes(readers, true);
```
mike-tr-adamson, line 253: It's still not entirely clear here that the validation can throw here and abort the stream. It would be nice to be clear on this and what the consequences are.

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
```diff
@@ -492,16 +495,100 @@ public static String getIndexName(String cfName)
...
+     * This is similar to {@link #buildIndexesBlocking}, but it is designed to be used in cases where failure will
+     * cascade through to failing the containing operation that actuates the build. (ex. streaming and SSTable import)
+     * <p>
+     * It does not update index build status or queryablility on failure or success and does not call
+     * {@link #flushIndexesBlocking(Set, FutureCallback)}, as this is an artifact of the legacy non-SSTable-attached
+     * index implementation.
+     *
+     * @param sstables the SSTables for which indexes must be built
+     */
+    public void buildIndexesBlockingIncremental(Collection<SSTableReader> sstables)
```
mike-tr-adamson: Since this is only building attached indexes, I wonder whether this would be better named `buildAttachedIndexesBlockingIncremental`?

test/distributed/org/apache/cassandra/distributed/test/sai/ImportIndexedSSTablesTest.java
```diff
@@ -0,0 +1,283 @@
...
+            {
+                zuper.call();
+            }
+            catch (Exception e)
+            {
+                throw Throwables.unchecked(e);
+            }
+        }
+    }
+}
```
mike-tr-adamson: nit: Missing new line

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
```diff
@@ -0,0 +1,200 @@
...
+            {
+                zuper.call();
+            }
+            catch (Exception e)
+            {
+                throw Throwables.unchecked(e);
+            }
+        }
+    }
+}
```
mike-tr-adamson: nit: missing new line

src/java/org/apache/cassandra/index/Index.java
```diff
@@ -765,6 +779,22 @@ default void invalidate() { }
          * @return the SSTable components created by this group
          */
         Set<Component> getComponents();
+
+        /**
+         * Validates all indexes in the group against the specified SSTables. 
+         * 
+         * @param sstables SSTables for which indexes in the group should be built
+         * @param throwOnIncomplete whether to throw an error if any index in the group is incomplete
```
mike-tr-adamson, line 788: `or not valid` at the end?

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
maedhroz: SIM handles the notification, but it will no longer trigger builds for SAI, given SAI will now return true from `Index#isSSTableAttached()`. I think we need to update the view no matter what.

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
maedhroz: Yeah, I'll add an inline comment, but also add a `@throws` to the JavaDoc downstream to mention the `UncheckedIOException`.

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
maedhroz: What about `buildSSTableAttachedIndexesBlocking()`?

src/java/org/apache/cassandra/index/Index.java
maedhroz: We throw on a validation error whether or not `throwOnIncomplete == true`, and I've added `@throws` to the JavaDoc to indicate this a bit more clearly.

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
```diff
@@ -0,0 +1,200 @@
...
+    {
+        cluster.get(2).runOnInstance(()-> ByteBuddyHelper.failFlush = false);
+        cluster.get(2).runOnInstance(()-> ByteBuddyHelper.failValidation = true);
+        testAvailabilityAfterStreaming("entire_file_test", true);
+    }
+
+    private void testAvailabilityAfterStreaming(String table, boolean streamEntireSSTables) throws Exception
+    {
+        
+
```
adelapena: Nit: unneeded blank lines

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
```diff
@@ -0,0 +1,200 @@
...
+    public static class ByteBuddyHelper
+    {
+        volatile static boolean failFlush = false;
+        volatile static boolean failValidation = false;
+        
+        @SuppressWarnings("resource")
+        static void installErrors(ClassLoader loader, int node)
+        {
+            if (node == 2)
```
adelapena, line 151: The two `if` blocks can be merged, or we could just have `if (node != 2) return;` at the beginning of the method.

maedhroz: Done

maedhroz:
Committed as https://github.com/apache/cassandra/commit/83523ddd3d2aa5ddd9c9b455551a50cedcf8704a

PR #2495, opened 2023-07-17
Fix sai intersection count

jbellis:


src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
```diff
@@ -249,14 +249,24 @@ private void updateStatistics(Statistics statistics, KeyRangeIterator range)
...
-            count += range.getCount();
+            if (empty)
+            {
+                empty = false;
+                count = range.getCount();
+            }
+            else
+            {
+                count = Math.min(count, range.getCount());
+            }
```
adelapena, line 269: Can this be `count = count == 0 ? range.getCount() : Math.min(count, range.getCount());
`?

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIterator.java
```diff
@@ -81,7 +81,8 @@ public final long getCount()
      *
      * @param nextKey value to skip the iterator forward until matching
      *
-     * @return The next current key after the skip was performed
+     * @return The key skipped to, which will be the key returned by the
+     * next call to `next()`, i.e., we are "peeking" at the next key as part of the skip.
```
adelapena:
```suggestion
     * next call to {@link #next()}, i.e., we are "peeking" at the next key as part of the skip.
```

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIterator.java
```diff
@@ -106,13 +107,21 @@ else if (next.compareTo(nextKey) >= 0)
...
+     */
     protected abstract void performSkipTo(PrimaryKey nextKey);
 
-    protected PrimaryKey recomputeNext()
+    private PrimaryKey recomputeNext()
     {
         return tryToComputeNext() ? peek() : endOfData();
     }
 
+    // protected because inherited from Guava. We don't want to expose this method.
```
adelapena: We should probably annotate the method with `@Override`.

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
jbellis: I don't think so.  Consider adding two iterators A and B.  A has count of 0, B has count of 1.

If we add A first, then B, then we will end up with a count of 1, since it's treating 0 as "uninitialized" which is not correct.

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIterator.java
jbellis: done

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
adelapena: Ah, I see.

adelapena (approved):
Running CI: https://app.circleci.com/pipelines/github/adelapena/cassandra/3040/workflows/806c0ce9-3c55-4c86-89e0-8a7cd1dc3d7b

adelapena:
CI looks good. Committed to `cep-7-sai` as [21a30509b0ee95ce72c44c403f1b2804a0ade5af](https://github.com/apache/cassandra/commit/21a30509b0ee95ce72c44c403f1b2804a0ade5af).

PR #2498, opened 2023-07-19
CASSANDRA-18673: Reduce disk footprint for SAI on-disk per-SSTable components

mike-tr-adamson:
This PR is focused on reducing the on-disk footprint of SAI per-SSTable components.

The main work involves removing the token from the primary key components. By including the token in the sorted terms and trie, we lose the advantage of prefix compression because the tokens are unique. Because of the uniqueness and the order in which the bytes are ordered (`<token>/<partition key>/<clustering key>`), we do not get any prefix compression of the partition key and clustering key.

The downside of removing the token is that the trie needs to be segmented because the `<partition key>/<clustering key>` is not delivered in lexicographic order. Also, because there is no token in the trie, we need to use reverse lookup in the token long array to lookup rowIds for token only primary keys. These are used for data range skipping at the beginning of the query.

This last element (token array lookup) allows us to not have to create the primary key trie at all for non-clustering table indexes because we can lookup rowIds by token.


src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsMeta.java
```diff
@@ -19,30 +19,87 @@
...
     {
-        output.writeLong(trieFilePointer);
         output.writeLong(termCount);
         output.writeInt(maxTermLength);
+        output.writeInt(segments.size());
+        for (SortedTermsSegmentMeta segment : segments)
+            segment.write(output);
+    }
+
+    public static class SortedTermsSegmentMeta
```
adelapena: Nit: being nested into `SortedTermsMeta`, it can be renamed to `Segment` or `SegmentMeta`

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsTrieSearcher.java
```diff
@@ -0,0 +1,86 @@
...
+
+import org.apache.cassandra.index.sai.disk.v1.trie.TriePrefixSearcher;
+import org.apache.cassandra.io.tries.Walker;
+import org.apache.cassandra.io.util.Rebufferer;
+import org.apache.cassandra.utils.bytecomparable.ByteComparable;
+import org.apache.cassandra.utils.bytecomparable.ByteSource;
+
+public class SortedTermsTrieSearcher implements AutoCloseable
+{
+    List<TrieSegment> segments;
```
adelapena: Nit: can be `private final`

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsTrieSearcher.java
```diff
@@ -0,0 +1,86 @@
...
+        return Long.MAX_VALUE;
+    }
+
+    @Override
+    public void close() throws Exception
+    {
+        segments.forEach(TrieSegment::close);
+    }
+
+    private class TrieSegment implements AutoCloseable
```
adelapena: Nit: can be `static`

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
```diff
@@ -92,10 +92,10 @@ public void testIndexComponentStreaming() throws IOException
             ));
 
             int num_components = isLiteral ? sstableStreamingComponentsCount() +
-                                             V1OnDiskFormat.PER_SSTABLE_COMPONENTS.size() +
+                                             V1OnDiskFormat.SKINNY_PER_SSTABLE_COMPONENTS.size() +
```
adelapena: I think it would be interesting to use a wide table to see how all the possible components are streamed. Even better, we could parameterise whether the table is skinny or wide: https://github.com/adelapena/cassandra/commit/481c144105b886d1f8891011810c4b3e6cd9bf1b
If the tests turns out to be too slow or it OOMs, we can always split it.

src/java/org/apache/cassandra/index/sai/disk/format/OnDiskFormat.java
```diff
@@ -136,7 +136,7 @@ PerColumnIndexWriter newPerColumnIndexWriter(StorageAttachedIndex index,
      * This is a complete set of components that could exist on-disk. It does not imply that the
      * components currently exist on-disk.
      */
-    Set<IndexComponent> perSSTableIndexComponents();
+    Set<IndexComponent> perSSTableIndexComponents(boolean hasClustering);
```
adelapena, line 141: Nit: add `@param` JavaDoc for `hasClustering`

src/java/org/apache/cassandra/index/sai/disk/format/OnDiskFormat.java
```diff
@@ -152,7 +152,7 @@ PerColumnIndexWriter newPerColumnIndexWriter(StorageAttachedIndex index,
      * This is a static indication of the files that can be held open by an index
      * for queries. It is not a dynamic calculation.
      */
-    int openFilesPerSSTableIndex();
+    int openFilesPerSSTableIndex(boolean hasClustering);
```
adelapena, line 159: Nit: add `@param` JavaDoc for `hasClustering`

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
```diff
@@ -103,6 +103,11 @@ public static IndexDescriptor create(SSTableReader sstable)
                                    sstable.metadata().comparator);
     }
 
+    public boolean hasClustering()
+    {
+        return clusteringComparator != null && clusteringComparator.size() > 0;
```
adelapena: Can `clusteringComparator` be null?

src/java/org/apache/cassandra/index/sai/disk/v1/RowAwarePrimaryKeyMap.java
```diff
@@ -109,8 +111,14 @@ public RowAwarePrimaryKeyMapFactory(IndexDescriptor indexDescriptor, SSTableRead
         public PrimaryKeyMap newPerSSTablePrimaryKeyMap() throws IOException
         {
             LongArray rowIdToToken = new LongArray.DeferredLongArray(tokenReaderFactory::open);
+            if (clusteringComparator != null && clusteringComparator.size() > 0)
```
adelapena: Can `clusteringComparator` be null?

mike-tr-adamson:
Note: I have just pushed a couple of commits tidying the code a bit. I have separated the primary key map out into 2 classes to avoid the conditional code around clustering keys and I've pulled an `AbstractSortedTermsTester` out of the sorted terms test to get rid of duplication.

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingTest.java
mike-tr-adamson: I think this is a good idea for streaming particularly.

src/java/org/apache/cassandra/index/sai/disk/v1/SkinnyRowAwarePrimaryKeyMap.java
```diff
@@ -109,89 +102,71 @@ public RowAwarePrimaryKeyMapFactory(IndexDescriptor indexDescriptor, SSTableRead
         public PrimaryKeyMap newPerSSTablePrimaryKeyMap() throws IOException
         {
             LongArray rowIdToToken = new LongArray.DeferredLongArray(tokenReaderFactory::open);
-            return new RowAwarePrimaryKeyMap(rowIdToToken,
-                                             new TriePrefixSearcher(primaryKeyTrieFile.instantiateRebufferer(null), sortedTermsMeta.trieFilePointer),
```
adelapena: This leaves an unused import of `TriePrefixSearcher`

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
maedhroz: Second this...I don't think there's currently any way it can be, after looking at all the callers.

src/java/org/apache/cassandra/index/sai/utils/SegmentMemoryLimiter.java
```diff
@@ -0,0 +1,122 @@
...
+     */
+    public static final long DEFAULT_SEGMENT_BUILD_MEMORY_LIMIT = DatabaseDescriptor.getSAISegmentWriteBufferSpace().toBytes();
+
+    private static final AtomicLong bytesUsed = new AtomicLong(0);
+
+    private static volatile long limitBytes = DEFAULT_SEGMENT_BUILD_MEMORY_LIMIT;
+
+    private static volatile long minimumFlushSize = 0;
+
+    /** The number of column indexes being built globally. (Starts at one to avoid divide by zero.) */
```
maedhroz: nit: This starts at zero but we avoid divide-by-zero by incrementing before we get?

src/java/org/apache/cassandra/index/sai/utils/SegmentMemoryLimiter.java
```diff
@@ -0,0 +1,122 @@
...
+     * ex. If there is only one column index building, it can buffer up to segment_write_buffer_space_mb.
+     * <p>
+     * ex. If there is one column index building per table across 8 compactors, each index will be
+     *     eligible to flush once it reaches (segment_write_buffer_space_mb / 8) MBs.
+     */
+    public static final long DEFAULT_SEGMENT_BUILD_MEMORY_LIMIT = DatabaseDescriptor.getSAISegmentWriteBufferSpace().toBytes();
+
+    private static final AtomicLong bytesUsed = new AtomicLong(0);
+
+    private static volatile long limitBytes = DEFAULT_SEGMENT_BUILD_MEMORY_LIMIT;
```
maedhroz: The only thing that I'm not 100% sold on is that `limitBytes` is non-final here (and therefore requires volatile access) simply because of testing. Taking a quick look at it's usage in `SortedTermsWriter` at least, we could probably still avoid making everything here `static` and instead have a singleton limiter that's passed to the things that use it. It might actually make some of the testing easier that way? (i.e. We could just pass new limiters instead of having to reset the global one? ...and in theory, we could test things without having to hit DD?)

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
```diff
@@ -92,25 +93,17 @@ abstract class AbstractPrimaryKey implements PrimaryKey
...
                 ByteSource keyComparable = ByteSource.of(partitionKey().getKey(), version);
+                if (clusteringComparator.size() == 0)
+                    return keyComparable;
                 // It is important that the ClusteringComparator.asBytesComparable method is used
                 // to maintain the correct clustering sort order
-                ByteSource clusteringComparable = clusteringComparator == null ||
-                                                  clusteringComparator.size() == 0 ||
-                                                  clustering() == null ||
+                ByteSource clusteringComparable = clustering() == null ||
                                                   clustering().isEmpty() ? null
```
maedhroz, line 104: It's kind of funny that, because we have `@Nullable` on `clustering()`, IDEA throws up a bunch of warnings about this possibly causing an NPE...even though we have a short-circuiting `clustering() == null ||` immediately before it. We could always throw `@SuppressWarnings("DataFlowIssue")` on these methods...

src/java/org/apache/cassandra/index/sai/utils/SegmentMemoryLimiter.java
```diff
@@ -0,0 +1,122 @@
...
+    {
+        limitBytes = bytes;
+    }
+
+    @VisibleForTesting
+    public static void reset()
+    {
+        bytesUsed.set(0);
+        limitBytes = DEFAULT_SEGMENT_BUILD_MEMORY_LIMIT;
+        activeBuilderCount.set(0);
```
maedhroz: I obviously would rather move away from having everything here be static, but if we do leave it that way, for completeness, I guess adding `minimumFlushSize = 0;` here would be a good idea...

(see https://github.com/apache/cassandra/pull/2498/files#r1270102370)

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
```diff
@@ -161,18 +154,13 @@ private void copyBufferToBytesRef(ByteBuffer buffer, BytesRefBuilder stringBuffe
...
-    {
-        return ACTIVE_BUILDER_COUNT.get();
-    }
-
-    private SegmentBuilder(AbstractType<?> termComparator, NamedMemoryLimiter limiter)
+    private SegmentBuilder(AbstractType<?> termComparator)
     {
         this.termComparator = termComparator;
-        this.limiter = limiter;
+//        this.limiter = limiter;
```
maedhroz: TODO: cleanup, depending on how we finalize the structure of the limiter class/singleton

src/java/org/apache/cassandra/index/sai/disk/v1/LongArray.java
```diff
@@ -37,6 +37,13 @@ public interface LongArray extends Closeable
      */
     long length();
 
+    /**
+     * Using the given token returns the first row ID corresponding to the token.
+     * @param targetToken Token to lookup and it must not be smaller than previous value
+     * @return The row ID of the given token or negative value if target token is greater than all tokens
+     */
+    long findTokenRowID(long targetToken);
```
maedhroz: We've gone back and forth about naming the methods when we have tokens in some cases and not tokens in others, so I don't think you have to change anything at this point, especially if we're going to make further changes to potentially eliminate the sorted terms entirely. We name this method `indexOf(long value)` or something like that, kind of following the `get()`/`indexOf()` pattern you typically see in Java collections. (see `List#indexOf()`)

The callers in the `rowIdFromPrimaryKey()` methods make the usage pretty clear, as does the thing they access, usually called `tokenArray`.

`long rowId = tokenArray.findTokenRowID(key.token().getLongValue());`

...and we even immediately assign the result to something called `rowId` in one case, making the index/row ID equivalence reasonably obvious.

maedhroz: It also would make `PostingsReader.LongArrayReader` less awkward I guess.

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
```diff
@@ -92,25 +93,17 @@ abstract class AbstractPrimaryKey implements PrimaryKey
         {
             public ByteSource asComparableBytes(ByteComparable.Version version)
             {
-                ByteSource tokenComparable = token().asComparableBytes(version);
```
maedhroz: Nice that we can stop converting tokens to `ByteSource` here...

src/java/org/apache/cassandra/index/sai/disk/v1/WideRowAwarePrimaryKeyMap.java
```diff
@@ -0,0 +1,162 @@
...
+        this.clusteringComparator = clusteringComparator;
+    }
+
+    @Override
+    public long rowIdFromPrimaryKey(PrimaryKey key)
+    {
+        long rowId = tokenArray.findTokenRowID(key.token().getLongValue());
+        // If the key only has a token (initial range skip in the query) or the token is out of range
+        // the return the rowId from the token array.
+        if (key.isTokenOnly() || rowId == -1)
```
maedhroz: nit: `rowId < 0`, given the way we documented `findTokenRowID()`?

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
```diff
@@ -201,14 +234,89 @@ private void copyBytes(ByteComparable source, BytesRefBuilder dest)
...
+        private final BytesRefBuilder tempTerm = new BytesRefBuilder();
+
+        private BytesRef minTerm;
+        private long totalBytesAllocated;
+        private boolean flushed = false;
+        private boolean active = true;
+
+        TrieSegment()
+        {
+            trie = new InMemoryTrie<>(TrieMemtable.BUFFER_TYPE);
```
maedhroz: If we use `OFF_HEAP`, does the limiter just ignore that memory usage?

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
```diff
@@ -201,14 +234,89 @@ private void copyBytes(ByteComparable source, BytesRefBuilder dest)
...
+                    BytesRef termRef = tempTerm.get();
+
+                    if (minTerm == null)
+                        minTerm = new BytesRef(Arrays.copyOf(termRef.bytes, termRef.length));
+
+                    trieWriter.add(next.getKey(), next.getValue());
+                    copyBytes(next.getKey(), prevTerm);
+                }
+
+                trieFilePointer = trieWriter.complete();
```
maedhroz: So to summarize what's going on here...

We get terms (primary keys) in token order, add them to an in-memory trie, which implicitly sorts them, and then we feed that trie, in primary key order, to the incremental writer, which assumes primary key order?

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
```diff
@@ -124,26 +146,61 @@ public void add(final @Nonnull ByteComparable term) throws IOException
...
+        {
+            if (trieSegment == null)
+                trieSegment = new TrieSegment();
+            else if (SegmentMemoryLimiter.usageExceedsLimit(trieSegment.totalBytesAllocated))
+            {
+                segments.add(trieSegment.flush());
+                trieSegment = new TrieSegment();
+            }
+            trieSegment.add(term);
+        }
```
maedhroz: Just a code organization nit, but it might be nice to have something like a `TrieWriter` interface in here that we can implement w/ a no-op singleton and the real implementation. That way, I imagine we could avoid all the `if (trieOutputWriter != null)` checking stuff and keep all the trie writing state/header writing there.

For example, this whole block of code reverts back to something like `trieWriter.add(term, rowID);`

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsTrieSearcher.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public SortedTermsTrieSearcher(Rebufferer source, SortedTermsMeta meta)
+    {
+        this.segments = new ArrayList<>(meta.segments.size());
+        for (SortedTermsMeta.SortedTermsSegmentMeta segmentMeta : meta.segments)
+            this.segments.add(new TrieSegment(source, segmentMeta));
+
+    }
+
+    public long prefixSearch(ByteComparable term)
```
maedhroz: nit: Brief JavaDoc for this, especially the possible return values and what they mean.

src/java/org/apache/cassandra/index/sai/disk/v1/SkinnyRowAwarePrimaryKeyMap.java
```diff
@@ -109,89 +101,71 @@ public RowAwarePrimaryKeyMapFactory(IndexDescriptor indexDescriptor, SSTableRead
...
+        tokenBuffer.putLong(tokenArray.get(sstableRowId));
         tokenBuffer.rewind();
         return primaryKeyFactory.createDeferred(partitioner.getTokenFactory().fromByteArray(tokenBuffer), () -> supplier(sstableRowId));
     }
 
     @Override
     public long rowIdFromPrimaryKey(PrimaryKey key)
     {
-        return triePrefixSearcher.prefixSearch(key.asComparableBytes(ByteComparable.Version.OSS50));
+        return tokenArray.findTokenRowID(key.token().getLongValue());
```
maedhroz: Just to refresh my memory, this never had to be exact, because it's only used for skipping, right?

src/java/org/apache/cassandra/index/sai/disk/v1/WideRowAwarePrimaryKeyMap.java
```diff
@@ -0,0 +1,162 @@
...
+            return rowId;
+        // If the token for the rowId matches the key token then we need to search the keys in the trie
+        // for a prefix match to get the correct rowId for the clustering key.
+        if (tokenArray.get(rowId) == key.token().getLongValue())
+            return trieSearcher.prefixSearch(key::asComparableBytes);
```
maedhroz, line 132: I'm trying to remember if we actually need to do this for correctness, or it's just an optimization for partitions w/ tons of rows, given this method is just used for skipping. If it's the latter, I wonder how small the partition has to be for accessing the trie not to be worth it...

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
adelapena: It kind of makes sense because the IDE isn't clever enough to realize that `PrimaryKey#clustering()` always returns the same value. However, it wouldn't complain if the method was named as a getter, as in `PrimaryKey#getClustering()`. I guess we can also put the clustering into a local var to prevent the warning without relying on `@SuppressWarnings`:
```java
Clustering<?> clustering = clustering();
ByteSource clusteringComparable = clustering == null || clustering.isEmpty()
                                  ? null
                                  : clusteringComparator.asByteComparable(clustering)
                                                        .asComparableBytes(version);
```
There are multiple IDE warnings around `partitionKey()` and `clustering()` being null that could be gone with whichever of these approaches we prefer (suppress warnings, rename as getters or use local vars).

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsMeta.java
mike-tr-adamson: I have gone with `SegmentMeta`

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
mike-tr-adamson: Agreed. It is always passed a valid clustering comparator.

src/java/org/apache/cassandra/index/sai/utils/SegmentMemoryLimiter.java
mike-tr-adamson: Agreed. I've never been a big fan of loosening code for testing but hadn't really thought about this enough. It really needs a `MemoryLimiter interface` to allow a test implementation to be used where need. I'll have another look at this and see what can be done.

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
mike-tr-adamson: I have fixed these by adding `@SuppressWarnings("ConstantConditions")`.

src/java/org/apache/cassandra/index/sai/utils/SegmentMemoryLimiter.java
mike-tr-adamson: I have added this, but it no longer exists in `SegmentMemoryLimiter`. This is now a singleton and is overridden by `SAITester$TestMemoryLimiter` where this method now lives.

src/java/org/apache/cassandra/index/sai/disk/v1/LongArray.java
mike-tr-adamson: That naming does make more sense. It feels in place.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
mike-tr-adamson: No, only the payloads are stored off-heap, so we still have to report the on-heap memory usage of the trie nodes.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
mike-tr-adamson: Correct, or more to the point, the incremental writer assumes they are in lexicographic order.

src/java/org/apache/cassandra/index/sai/disk/v1/SkinnyRowAwarePrimaryKeyMap.java
mike-tr-adamson: I'm not sure what you mean by `exact` here, but, yes, it is only used for skipping so only needs to get to the first matching token.

src/java/org/apache/cassandra/index/sai/disk/v1/WideRowAwarePrimaryKeyMap.java
mike-tr-adamson: No, this is wrong. I think I got confused when splitting the primary key maps and left this in. If the key isn't token only then we need to full prefix search in all cases.

mike-tr-adamson:
@maedhroz @adelapena The latest commit addresses the current set of review comments

adelapena:
@mike-tr-adamson it seems there are multiple utest failures for collections: https://app.circleci.com/pipelines/github/adelapena/cassandra/3056/workflows/6a503495-83ae-417e-8c9b-a9458cab1f01/jobs/62722/tests

PR #2505, opened 2023-07-20
CASSANDRA-18624 Amazon Corretto Crypto Provider

smiklosovic:
 
```

```

### DIFF
--- a/.build/build-resolver.xml
+++ b/.build/build-resolver.xml
@@ -64,7 +64,7 @@
 
         <macrodef name="resolve">
             <!--
-              maven-resolver-ant-tasks's resolve logic doesn't have retry logic and does not respect settings.xml, 
+              maven-resolver-ant-tasks's resolve logic doesn't have retry logic and does not respect settings.xml,
               this causes issues when overriding maven central is required (such as when behind a corporate firewall);
               it is critical to always provide the 'all' remoterepos to override resolve's default hard coded logic.
 
@@ -74,7 +74,7 @@
             <element name="elements" implicit="yes"/>
             <sequential>
                 <retry retrycount="3">
-                  <resolver:resolve failonmissingattachments="@{failonmissingattachments}">
+                    <resolver:resolve failonmissingattachments="@{failonmissingattachments}">
                         <resolver:remoterepos refid="all"/>
                         <elements/>
                     </resolver:resolve>
@@ -88,8 +88,8 @@
             <sequential>
                 <retry retrycount="3">
                     <resolver:pom file="@{file}" id="@{id}">
-                          <remoterepos refid="all"/>
-                          <elements/>
+                        <remoterepos refid="all"/>
+                        <elements/>
                     </resolver:pom>
                 </retry>
             </sequential>
@@ -206,6 +206,26 @@
         </retry>
         <mkdir dir="${local.repository}/org/apache/cassandra/deps/sigar-bin"/>
         <mkdir dir="${build.lib}/sigar-bin"/>
+        <mkdir dir="${build.lib}/x86_64"/>
+        <mkdir dir="${build.lib}/aarch64"/> <!-- uname -m on arm prints aarch64 instead of aarch_64 -->
+
+        <!-- artifacts needs AmazonCorrettoCryptoProvider for multiple archs -->
+        <retry retrycount="3" retrydelay="10" >
+            <resolve>
+                <dependencies>
+                    <dependency groupId="software.amazon.cryptools" artifactId="AmazonCorrettoCryptoProvider" version="2.2.0" classifier="linux-x86_64" />
+                </dependencies>
+                <files dir="${build.lib}/x86_64" layout="{artifactId}-{version}-{classifier}.{extension}" />
+            </resolve>
+        </retry>
+        <retry retrycount="3" retrydelay="10" >
+            <resolve>
+                <dependencies>
+                    <dependency groupId="software.amazon.cryptools" artifactId="AmazonCorrettoCryptoProvider" version="2.2.0" classifier="linux-aarch_64" />
+                </dependencies>
+                <files dir="${build.lib}/aarch64" layout="{artifactId}-{version}-{classifier}.{extension}" />
+            </resolve>
+        </retry>
 
         <retry retrycount="3" retrydelay="10" >
             <antcall target="_resolver-dist-lib_get_files"/>
@@ -240,6 +260,10 @@
             <file file="${local.repository}/org/apache/cassandra/deps/sigar-bin/libsigar-x86-linux.so"/>
             <file file="${local.repository}/org/apache/cassandra/deps/sigar-bin/libsigar-x86-solaris.so"/>
         </copy>
+
+        <!-- as resolver will copy all dependencies into lib dir, and we are copying jars to lib/{x86_64|aarch64} as well, we would have duplicities -->
+        <delete file="${build.lib}/AmazonCorrettoCryptoProvider-2.2.0-linux-x86_64.jar" failonerror="false"/>
+        <delete file="${build.lib}/AmazonCorrettoCryptoProvider-2.2.0-linux-aarch_64.jar" failonerror="false"/>
     </target>
 
     <target name="_resolver-dist-lib_get_files">
--- a/.build/parent-pom-template.xml
+++ b/.build/parent-pom-template.xml
@@ -233,12 +233,53 @@
       <id>zznate</id>
       <name>Nate McCall</name>
     </developer>
+    <developer>
+      <id>smiklosovic</id>
+      <name>Stefan Miklosovic</name>
+    </developer>
   </developers>
   <scm>
     <connection>scm:https://gitbox.apache.org/repos/asf/cassandra.git</connection>
     <developerConnection>scm:https://gitbox.apache.org/repos/asf/cassandra.git</developerConnection>
     <url>https://gitbox.apache.org/repos/asf?p=cassandra.git;a=tree</url>
   </scm>
+
+  <profiles>
+    <profile>
+      <id>x86_64</id>
+      <activation>
+        <os>
+          <!-- we need something as a default even if it doesn't successfully load the .so files. -->
+          <arch>!aarch64</arch>
+        </os>
+      </activation>
+      <dependencies>
+        <dependency>
+          <groupId>software.amazon.cryptools</groupId>
+          <artifactId>AmazonCorrettoCryptoProvider</artifactId>
+          <classifier>linux-x86_64</classifier>
+          <version>2.2.0</version>
+        </dependency>
+      </dependencies>
+    </profile>
+    <profile>
+      <id>aarch_64</id>
+      <activation>
+        <os>
+          <arch>aarch64</arch>
+        </os>
+      </activation>
+      <dependencies>
+        <dependency>
+          <groupId>software.amazon.cryptools</groupId>
+          <artifactId>AmazonCorrettoCryptoProvider</artifactId>
+          <classifier>linux-aarch_64</classifier>
+          <version>2.2.0</version>
+        </dependency>
+      </dependencies>
+    </profile>
+  </profiles>
+
   <dependencyManagement>
     <!--
     Dependency metadata is specified here (version, scope, exclusions, etc.), then referenced in child POMs by groupId and
--- a/bin/cassandra.in.sh
+++ b/bin/cassandra.in.sh
@@ -84,6 +84,12 @@ JAVA_AGENT="$JAVA_AGENT -javaagent:$CASSANDRA_HOME/lib/jamm-0.4.0.jar"
 # Added sigar-bin to the java.library.path CASSANDRA-7838
 JAVA_OPTS="$JAVA_OPTS:-Djava.library.path=$CASSANDRA_HOME/lib/sigar-bin"
 
+platform=$(uname -m)
+if [ -d "$CASSANDRA_HOME"/lib/"$platform" ]; then
+    for jar in "$CASSANDRA_HOME"/lib/"$platform"/*.jar ; do
+        CLASSPATH="$CLASSPATH:${jar}"
+    done
+fi
 
 #
 # Java executable and per-Java version JVM settings
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -1375,6 +1375,21 @@ dynamic_snitch_reset_interval: 600000ms
 # until the pinned host was 20% worse than the fastest.
 dynamic_snitch_badness_threshold: 1.0
 
+# Configures Java crypto provider. By default, it will use DefaultCryptoProvider
+# which will install Amazon Correto Crypto Provider.
+#
+# Amazon Correto Crypto Provider works currently for x86_64 and aarch_64 platforms.
+# If this provider fails it will fall back to the default crypto provider in the JRE.
+#
+# To force failure when the provider was not installed properly, set the property "fail_on_missing_provider" to "true".
+#
+# To bypass the installation of a crypto provider use class 'org.apache.cassandra.security.JREProvider'
+#
+crypto_provider:
+  - class_name: org.apache.cassandra.security.DefaultCryptoProvider
+    parameters:
+      - fail_on_missing_provider: "false"
+
 # Configure server-to-server internode encryption
 #
 # JVM and netty defaults for supported SSL socket protocols and cipher suites can
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -98,7 +98,9 @@ import org.apache.cassandra.locator.IEndpointSnitch;
 import org.apache.cassandra.locator.InetAddressAndPort;
 import org.apache.cassandra.locator.Replica;
 import org.apache.cassandra.locator.SeedProvider;
+import org.apache.cassandra.security.AbstractCryptoProvider;
 import org.apache.cassandra.security.EncryptionContext;
+import org.apache.cassandra.security.JREProvider;
 import org.apache.cassandra.security.SSLFactory;
 import org.apache.cassandra.service.CacheService.CacheType;
 import org.apache.cassandra.service.paxos.Paxos;
@@ -123,10 +125,11 @@ import static org.apache.cassandra.config.CassandraRelevantProperties.SEARCH_CON
 import static org.apache.cassandra.config.CassandraRelevantProperties.SSL_STORAGE_PORT;
 import static org.apache.cassandra.config.CassandraRelevantProperties.STORAGE_DIR;
 import static org.apache.cassandra.config.CassandraRelevantProperties.STORAGE_PORT;
-import static org.apache.cassandra.config.CassandraRelevantProperties.TEST_STRICT_RUNTIME_CHECKS;
 import static org.apache.cassandra.config.CassandraRelevantProperties.SUN_ARCH_DATA_MODEL;
 import static org.apache.cassandra.config.CassandraRelevantProperties.TEST_FAIL_MV_LOCKS_COUNT;
 import static org.apache.cassandra.config.CassandraRelevantProperties.TEST_JVM_DTEST_DISABLE_SSL;
+import static org.apache.cassandra.config.CassandraRelevantProperties.TEST_SKIP_CRYPTO_PROVIDER_INSTALLATION;
+import static org.apache.cassandra.config.CassandraRelevantProperties.TEST_STRICT_RUNTIME_CHECKS;
 import static org.apache.cassandra.config.CassandraRelevantProperties.UNSAFE_SYSTEM;
 import static org.apache.cassandra.config.DataRateSpec.DataRateUnit.BYTES_PER_SECOND;
 import static org.apache.cassandra.config.DataRateSpec.DataRateUnit.MEBIBYTES_PER_SECOND;
@@ -174,6 +177,7 @@ public class DatabaseDescriptor
 
     private static Config.DiskAccessMode indexAccessMode;
 
+    private static AbstractCryptoProvider cryptoProvider;
     private static IAuthenticator authenticator;
     private static IAuthorizer authorizer;
     private static INetworkAuthorizer networkAuthorizer;
@@ -425,6 +429,8 @@ public class DatabaseDescriptor
     {
         //InetAddressAndPort cares that applySimpleConfig runs first
         applySSTableFormats();
+
+        applyCryptoProvider();
 
         applySimpleConfig();
 
@@ -1219,6 +1225,42 @@ public class DatabaseDescriptor
         }
     }
 
+    public static void applyCryptoProvider()
+    {
+        if (TEST_SKIP_CRYPTO_PROVIDER_INSTALLATION.getBoolean())
+            return;
+
+        if (conf.crypto_provider == null)
+            conf.crypto_provider = new ParameterizedClass(JREProvider.class.getName(), null);
+
+        // properties beat configuration
+        String classNameFromSystemProperties = CassandraRelevantProperties.CRYPTO_PROVIDER_CLASS_NAME.getString();
+        if (classNameFromSystemProperties != null)
+            conf.crypto_provider.class_name = classNameFromSystemProperties;
+
+        if (conf.crypto_provider.class_name == null)
+            throw new ConfigurationException("Failed to initialize crypto provider, class_name cannot be null");
+
+        if (conf.crypto_provider.parameters == null)
+            conf.crypto_provider.parameters = new HashMap<>();
+
+        Map<String, String> cryptoProviderParameters = new HashMap<>(conf.crypto_provider.parameters);
+        cryptoProviderParameters.putIfAbsent(AbstractCryptoProvider.FAIL_ON_MISSING_PROVIDER_KEY, "false");
+
+        try
+        {
+            cryptoProvider = FBUtilities.newCryptoProvider(conf.crypto_provider.class_name, cryptoProviderParameters);
+            cryptoProvider.install();
+        }
+        catch (Exception e)
+        {
+            if (e instanceof ConfigurationException)
+                throw (ConfigurationException) e;
+            else
+                throw new ConfigurationException(String.format("Failed to initialize crypto provider %s", conf.crypto_provider.class_name), e);
+        }
+    }
+
     public static void applySeedProvider()
     {
         // load the seeds for node contact points
@@ -1502,6 +1544,15 @@ public class DatabaseDescriptor
         return detector;
     }
 
+    public static AbstractCryptoProvider getCryptoProvider()
+    {
+        return cryptoProvider;
+    }
+
+    public static void setCryptoProvider(AbstractCryptoProvider cryptoProvider)
+    {
+        DatabaseDescriptor.cryptoProvider = cryptoProvider;
+    }
     public static IAuthenticator getAuthenticator()
     {
         return authenticator;
--- a/src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
+++ b/src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
@@ -1,0 +1,205 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.security;
+
+import org.apache.cassandra.exceptions.ConfigurationException;
+import org.apache.cassandra.utils.FBUtilities;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.security.Provider;
+import java.security.Security;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import static java.lang.String.format;
+import static org.apache.cassandra.config.CassandraRelevantProperties.FAIL_ON_MISSING_CRYPTO_PROVIDER;
+
+public abstract class AbstractCryptoProvider
+{
+    protected static final Logger logger = LoggerFactory.getLogger(AbstractCryptoProvider.class);
+    public static final String FAIL_ON_MISSING_PROVIDER_KEY = "fail_on_missing_provider";
+
+    protected final boolean failOnMissingProvider;
+    private final Map<String, String> properties;
+
+    public AbstractCryptoProvider(Map<String, String> args)
+    {
+        this.properties = args == null ? new HashMap<>() : args;
+        boolean failOnMissingProviderFromProperties = Boolean.parseBoolean(this.properties.getOrDefault(FAIL_ON_MISSING_PROVIDER_KEY, "false"));
+        failOnMissingProvider = FAIL_ON_MISSING_CRYPTO_PROVIDER.getBoolean(failOnMissingProviderFromProperties);
+    }
+
+    /**
+     * Returns unmodifiable properties of this crypto provider
+     *
+     * @return crypto provider properties
+     */
+    public Map<String, String> getProperties()
+    {
+        return Collections.unmodifiableMap(properties);
+    }
+
+    /**
+     * Returns name of the provider, as returned from {@link Provider#getName()}
+     *
+     * @return name of the provider
+     */
+    public abstract String getProviderName();
+
+    /**
+     * Returns the name of the class which installs specific provider of name {@link #getProviderName()}.
+     *
+     * @return name of class of provider
+     */
+    public abstract String getProviderClassAsString();
+
+    /**
+     * Returns a runnable which installs this crypto provider.
+     *
+     * @return runnable which installs this provider
+     */
+    protected abstract Runnable installator();
+
+    /**
+     * Returns boolean telling if this provider was installed properly.
+     *
+     * @return {@code true} if provider was installed properly, {@code false} otherwise.
+     */
+    protected abstract boolean isHealthyInstallation() throws Exception;
+
+    /**
+     * The default installation runs {@link AbstractCryptoProvider#installator()} and after that
+     * {@link AbstractCryptoProvider#isHealthyInstallation()}.
+     * <p>
+     * If any step fails, it will not throw an exception unless the parameter
+     * {@link AbstractCryptoProvider#FAIL_ON_MISSING_PROVIDER_KEY} is {@code true}.
+     */
+    public void install() throws Exception
+    {
+        String failureMessage = null;
+        Throwable t = null;
+        try
+        {
+            if (JREProvider.class.getName().equals(getProviderClassAsString()))
+            {
+                logger.info(format("Installation of a crypto provider was skipped as %s was used.", JREProvider.class.getName()));
+                return;
+            }
+
+            FBUtilities.classForName(getProviderClassAsString(), "crypto provider");
+
+            String providerName = getProviderName();
+            int providerPosition = getProviderPosition(providerName);
+            if (providerPosition > 0)
+            {
+                if (providerPosition == 1)
+                {
+                    logger.info("{} was already installed on position {}.", providerName, providerPosition);
+                }
+                else if (failOnMissingProvider)
+                {
+                    throw new IllegalStateException(String.format("%s was already installed on position %s.", providerName, providerPosition));
+                }
+                else
+                {
+                    logger.warn("{} was already installed on position {}. Check the configuration of " +
+                                "JRE and either remove the provider from java.security or do not install this provider " +
+                                "by Cassandra.", providerName, providerPosition);
+                    return;
+                }
+            }
+            else
+            {
+                Runnable r = installator();
+                if (r == null)
+                    throw new IllegalStateException("Installator runnable can not be null!");
+                else
+                    r.run();
+            }
+
+            if (isHealthyInstallation())
+                logger.info("{} health check OK.", getProviderName());
+            else
+                failureMessage = format("%s has not passed the health check. " +
+                                        "Check node's architecture (`uname -m`) is supported, see lib/<arch> subdirectories. " +
+                                        "The correct architecture-specific library for %s needs to be on the classpath. ",
+                                        getProviderName(),
+                                        getProviderClassAsString());
+        }
+        catch (ConfigurationException ex)
+        {
+            failureMessage = getProviderClassAsString() + " is not on the class path! " +
+                             "Check node's architecture (`uname -m`) is supported, see lib/<arch> subdirectories. " +
+                             "The correct architecture-specific library for needs to be on the classpath.";
+        }
+        catch (Throwable ex)
+        {
+            failureMessage = format("The installation of %s was not successful, reason: %s",
+                                    getProviderClassAsString(), ex.getMessage());
+            t = ex;
+        }
+
+        if (failureMessage != null)
+        {
+            // To be sure there is not any leftover, proactively remove this provider in case of any failure.
+            // This method returns silently if the provider is not installed or if name is null.
+            try
+            {
+                uninstall();
+            }
+            catch (Throwable throwable)
+            {
+                logger.warn("Uninstallation of {} failed", getProviderName(), throwable);
+            }
+
+            if (failOnMissingProvider)
+                throw new ConfigurationException(failureMessage, t);
+            else
+                logger.warn(failureMessage);
+        }
+    }
+
+    /**
+     * Uninstalls this crypto provider of name {@link #getProviderName()}
+     *
+     * @see Security#removeProvider(String)
+     */
+    public void uninstall()
+    {
+        Security.removeProvider(getProviderName());
+    }
+
+    private int getProviderPosition(String providerName)
+    {
+        Provider[] providers = Security.getProviders();
+
+        for (int i = 0; i < providers.length; i++)
+        {
+            if (providers[i].getName().equals(providerName))
+            {
+                return i + 1;
+            }
+        }
+
+        return -1;
+    }
+}
--- a/src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
+++ b/src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.security;
+
+import java.util.Map;
+import javax.crypto.Cipher;
+
+import com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider;
+
+/**
+ * Default crypto provider tries to install AmazonCorrettoCryptoProvider.
+ * <p>
+ * The implementation falls back to in-built crypto provider in JRE if the installation
+ * is not successful.
+ */
+public class DefaultCryptoProvider extends AbstractCryptoProvider
+{
+    public DefaultCryptoProvider(Map<String, String> args)
+    {
+        super(args);
+    }
+
+    @Override
+    public String getProviderName()
+    {
+        return "AmazonCorrettoCryptoProvider";
+    }
+
+    @Override
+    public String getProviderClassAsString()
+    {
+        return "com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider";
+    }
+
+    @Override
+    public Runnable installator()
+    {
+        return AmazonCorrettoCryptoProvider::install;
+    }
+
+    @Override
+    public boolean isHealthyInstallation() throws Exception
+    {
+        if (!getProviderName().equals(Cipher.getInstance("AES/GCM/NoPadding").getProvider().getName()))
+            return false;
+
+        AmazonCorrettoCryptoProvider.INSTANCE.assertHealthy();
+
+        return true;
+    }
+}
--- a/test/distributed/org/apache/cassandra/distributed/impl/Instance.java
+++ b/test/distributed/org/apache/cassandra/distributed/impl/Instance.java
@@ -923,6 +923,8 @@ public class Instance extends IsolatedExecutor implements IInvokableInstance
             
             error = parallelRun(error, executor, this::stopJmx);
 
+            error = parallelRun(error, executor, () -> DatabaseDescriptor.getCryptoProvider().uninstall());
+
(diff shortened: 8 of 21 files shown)


src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
```diff
@@ -39,11 +39,20 @@ public void installProvider()
     {
         try
         {
-            AmazonCorrettoCryptoProvider.install();
+            Class.forName("com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider");
```
michaelsembwever: what if we want something completely different? (like we load other things…)

.build/parent-pom-template.xml
```diff
@@ -239,6 +239,40 @@
     <developerConnection>scm:https://gitbox.apache.org/repos/asf/cassandra.git</developerConnection>
     <url>https://gitbox.apache.org/repos/asf?p=cassandra.git;a=tree</url>
   </scm>
+
+  <profiles>
+    <profile>
+      <id>x86_64</id>
+      <activation>
+        <activeByDefault>true</activeByDefault>
```
michaelsembwever: how will this not be enabled when `aarch64` is …? (profiles are not exclusive)

.build/build-resolver.xml
```diff
@@ -279,6 +288,15 @@
...
         </get>
+
+
+        <!-- Amazon Corretto Crypto Providers for various architectures -->
+        <get src="${artifact.github.release}/corretto/amazon-corretto-crypto-provider/releases/download/2.2.0/AmazonCorrettoCryptoProvider-2.2.0-linux-x86_64.jar"
+             dest="${local.repository}/software/amazon/cryptools/AmazonCorrettoCryptoProvider/2.2.0/AmazonCorrettoCryptoProvider-2.2.0-linux-x86_64.jar"
+             usetimestamp="true" quiet="true" skipexisting="true"/>
+        <get src="${artifact.github.release}/corretto/amazon-corretto-crypto-provider/releases/download/2.2.0/AmazonCorrettoCryptoProvider-2.2.0-linux-aarch_64.jar"
+             dest="${local.repository}/software/amazon/cryptools/AmazonCorrettoCryptoProvider/2.2.0/AmazonCorrettoCryptoProvider-2.2.0-linux-aarch_64.jar"
+             usetimestamp="true" quiet="true" skipexisting="true"/>
```
michaelsembwever: can we not resolve and pull by maven coordinates ?

.build/build-resolver.xml
```diff
@@ -246,6 +247,14 @@
...
             <file file="${local.repository}/org/apache/cassandra/deps/sigar-bin/libsigar-x86-solaris.so"/>
         </copy>
+
+        <copy todir="${build.lib}/corretto" quiet="true">
+            <file file="${local.repository}/software/amazon/cryptools/AmazonCorrettoCryptoProvider/2.2.0/AmazonCorrettoCryptoProvider-2.2.0-linux-x86_64.jar"/>
+            <file file="${local.repository}/software/amazon/cryptools/AmazonCorrettoCryptoProvider/2.2.0/AmazonCorrettoCryptoProvider-2.2.0-linux-aarch_64.jar"/>
+        </copy>
+
+        <delete file="${build.lib}/AmazonCorrettoCryptoProvider-2.2.0-linux-x86_64.jar" failonerror="false"/>
+        <delete file="${build.lib}/AmazonCorrettoCryptoProvider-2.2.0-linux-aarch_64.jar" failonerror="false"/>
```
michaelsembwever, line 266: why the two deletes?

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
smiklosovic: @michaelsembwever then you would have a completely different implementation of `ICryptoProvider` which you reference in `crypto_provider`.

I was thinking about abstracting this logic so the same pattern would be used, "Amazon provider" would just override few values, the class to load, how installation and check is going to look like, but otherwise it would be all same for all specific crypto providers. 

It looks like

    crypto_provider:
        - class_name: abc
           parameters:
              - a: b

.build/parent-pom-template.xml
smiklosovic: you would need to explicitly exclude it 

    mvn clean install -P !profile-to-exclude

If we are scripting this, if there is a build machine on ARM, we would detect this and we would construct maven command which would exclude `x86_64`.

.build/build-resolver.xml
smiklosovic: because when doing `ant jar`, since that dependency is in pom.xml, resolver will copy all libraries as dependencies to `libs`. But we are also copying x86 and aarch to `corretto` dir. So you would have there these files two times.

.build/build-resolver.xml
smiklosovic: probably possible, yeah...

.build/parent-pom-template.xml
michaelsembwever: yeah, we don't want to be playing around with profiles on the command line. 

let's 
```
 <profiles>
    <profile>
      <id>x86_64</id>
      <activation>
        <os><arch>!aarch64</arch></os>
      </activation>
      <dependencies>
        <dependency>
          <groupId>software.amazon.cryptools</groupId>
          <artifactId>AmazonCorrettoCryptoProvider</artifactId>
          <classifier>linux-x86_64</classifier>
          <version>2.2.0</version>
        </dependency>
      </dependencies>
    </profile>
    <profile>
      <id>aarch_64</id>
      <activation>
        <os><arch>aarch64</arch></os>
      </activation>
      …
```

and if `<os><arch>!aarch64</arch></os>` doesn't work, then `<property><os.arch>!amd64</os.arch></property >`

.build/build-resolver.xml
michaelsembwever: cool.  add a comment please :) 


and maybe… we need it in artifacts, not here… 🤷🏻‍♀️

.build/build-resolver.xml
smiklosovic: @michaelsembwever look, I am not going to do this. This is hell in Ant ... The thing is that we need to check if a file is present locally and if it is not we are going to call mvn plugin which downloads it, I tried to add the target to that but the logic around it is just ... meh. This is super straightforward.

michaelsembwever: yes it's a pita.

it is done like
```

        <resolve>
            <dependencies>
                <dependency groupId="software.amazon.cryptools" artifactId="AmazonCorrettoCryptoProvider" version="2.2.0" classifier="linux-x86_64" />
                <dependency groupId="software.amazon.cryptools" artifactId="AmazonCorrettoCryptoProvider" version="2.2.0" classifier="linux-aarch_64" />
            </dependencies>
        </resolve>
```

.build/build-resolver.xml
```diff
@@ -246,6 +247,15 @@
             <file file="${local.repository}/org/apache/cassandra/deps/sigar-bin/libsigar-x86-linux.so"/>
             <file file="${local.repository}/org/apache/cassandra/deps/sigar-bin/libsigar-x86-solaris.so"/>
         </copy>
+
+        <copy todir="${build.lib}/corretto" quiet="true">
+            <file file="${local.repository}/software/amazon/cryptools/AmazonCorrettoCryptoProvider/2.2.0/AmazonCorrettoCryptoProvider-2.2.0-linux-x86_64.jar"/>
```
xgerman: don't like the hard coded 2.2.0 -- can't there be vars like in maven?

.build/parent-pom-template.xml
```diff
@@ -239,6 +239,42 @@
...
+        <os>
+          <arch>!aarch64</arch>
+        </os>
+      </activation>
+      <dependencies>
+        <dependency>
+          <groupId>software.amazon.cryptools</groupId>
+          <artifactId>AmazonCorrettoCryptoProvider</artifactId>
+          <classifier>linux-x86_64</classifier>
+          <version>2.2.0</version>
```
xgerman, line 261: make 2.20 a variable to configure in one place

.build/parent-pom-template.xml
```diff
@@ -239,6 +239,42 @@
...
+        <os>
+          <arch>aarch64</arch>
+        </os>
+      </activation>
+      <dependencies>
+        <dependency>
+          <groupId>software.amazon.cryptools</groupId>
+          <artifactId>AmazonCorrettoCryptoProvider</artifactId>
+          <classifier>linux-aarch_64</classifier>
+          <version>2.2.0</version>
```
xgerman, line 277: same

.build/parent-pom-template.xml
```diff
@@ -770,6 +806,13 @@
       </dependency>
       <!-- end of chronicle-queue -->
 
+      <dependency>
+        <groupId>software.amazon.cryptools</groupId>
+        <artifactId>AmazonCorrettoCryptoProvider</artifactId>
+        <version>2.2.0</version>
```
xgerman: same

bin/cassandra.in.sh
```diff
@@ -84,6 +84,12 @@ JAVA_AGENT="$JAVA_AGENT -javaagent:$CASSANDRA_HOME/lib/jamm-0.3.2.jar"
 # Added sigar-bin to the java.library.path CASSANDRA-7838
 JAVA_OPTS="$JAVA_OPTS:-Djava.library.path=$CASSANDRA_HOME/lib/sigar-bin"
 
+operating_system=$(uname -s | tr '[:upper:]' '[:lower:]')
+corretto_lib_pattern="AmazonCorrettoCryptoProvider-*-$operating_system-$(uname -m).jar"
+corretto_lib="$(find "${CASSANDRA_HOME}"/lib/corretto -type f -name "$corretto_lib_pattern" -exec basename {} \; | head -n 1)"
+if [ -f "${CASSANDRA_HOME}/lib/corretto/${corretto_lib}" ]; then
+  CLASSPATH="$CLASSPATH${CASSANDRA_HOME}/lib/corretto/$corretto_lib";
+fi
```
xgerman, line 92: maybe an else to print a warning that the other architetcure (e.g. MIPS, power. or whatever) is not supported

xgerman (approved):
Some nits. Otherwise LGTM

bin/cassandra.in.sh
michaelsembwever: they are supported in the sense that it just falls back to normal provider.
i'm guessing such feedback is provided from java logging at startup …?

smiklosovic: yes it is logged

.build/build-resolver.xml
smiklosovic: @xgerman There is a lot of instances of this happening in that file, if you look closer. If we fixed it just here, it would be inconsistent with everything else. I would rather dedicate a separate ticket for cleaning it all up in one shot. How does that sound to you?

.build/parent-pom-template.xml
smiklosovic: same

.build/build-resolver.xml
xgerman: Sounds good to me.

smiklosovic:
@xgerman I removed the startup check logic and it is all installed as part of `appyCryptoProvider`. The reason behind this is that we should fail the start of the node as soon as possible.

.build/build-resolver.xml
```diff
@@ -290,13 +290,12 @@
...
-                <dependency groupId="software.amazon.cryptools" artifactId="AmazonCorrettoCryptoProvider" version="2.2.0" classifier="linux-x86_64" />
-                <dependency groupId="software.amazon.cryptools" artifactId="AmazonCorrettoCryptoProvider" version="2.2.0" classifier="linux-aarch_64" />
-            </dependencies>
-        </resolve>
+        <get src="${artifact.github.release}/corretto/amazon-corretto-crypto-provider/releases/download/2.2.0/AmazonCorrettoCryptoProvider-2.2.0-linux-x86_64.jar"
+             dest="${local.repository}/software/amazon/cryptools/AmazonCorrettoCryptoProvider/2.2.0/AmazonCorrettoCryptoProvider-2.2.0-linux-x86_64.jar"
+             usetimestamp="true" quiet="true" skipexisting="true"/>
+        <get src="${artifact.github.release}/corretto/amazon-corretto-crypto-provider/releases/download/2.2.0/AmazonCorrettoCryptoProvider-2.2.0-linux-aarch_64.jar"
+             dest="${local.repository}/software/amazon/cryptools/AmazonCorrettoCryptoProvider/2.2.0/AmazonCorrettoCryptoProvider-2.2.0-linux-aarch_64.jar"
+             usetimestamp="true" quiet="true" skipexisting="true"/>
```
michaelsembwever: why was this re-introduced? 

i'm also confused as to why the macro needed to be moved out… `resolver-init` defines a number of macros that are used in other places…

bin/cassandra.in.sh
```diff
@@ -86,9 +86,9 @@ JAVA_OPTS="$JAVA_OPTS:-Djava.library.path=$CASSANDRA_HOME/lib/sigar-bin"
 
 operating_system=$(uname -s | tr '[:upper:]' '[:lower:]')
 corretto_lib_pattern="AmazonCorrettoCryptoProvider-*-$operating_system-$(uname -m).jar"
-corretto_lib="$(find "${CASSANDRA_HOME}"/lib/corretto -type f -name "$corretto_lib_pattern" -exec basename {} \; | head -n 1)"
-if [ -f "${CASSANDRA_HOME}/lib/corretto/${corretto_lib}" ]; then
-  CLASSPATH="$CLASSPATH${CASSANDRA_HOME}/lib/corretto/$corretto_lib";
+corretto_lib="$(find "${CASSANDRA_HOME}"/lib/platform -type f -name "$corretto_lib_pattern" -exec basename {} \; | head -n 1)"
+if [ -f "${CASSANDRA_HOME}/lib/platform/${corretto_lib}" ]; then
+  CLASSPATH="$CLASSPATH${CASSANDRA_HOME}/lib/platform/$corretto_lib";
```
michaelsembwever, line 91: are we sure all distros have the `findutils` package installed ?

bin/cassandra.in.sh
```diff
@@ -86,9 +86,9 @@ JAVA_OPTS="$JAVA_OPTS:-Djava.library.path=$CASSANDRA_HOME/lib/sigar-bin"
 
 operating_system=$(uname -s | tr '[:upper:]' '[:lower:]')
 corretto_lib_pattern="AmazonCorrettoCryptoProvider-*-$operating_system-$(uname -m).jar"
-corretto_lib="$(find "${CASSANDRA_HOME}"/lib/corretto -type f -name "$corretto_lib_pattern" -exec basename {} \; | head -n 1)"
-if [ -f "${CASSANDRA_HOME}/lib/corretto/${corretto_lib}" ]; then
-  CLASSPATH="$CLASSPATH${CASSANDRA_HOME}/lib/corretto/$corretto_lib";
+corretto_lib="$(find "${CASSANDRA_HOME}"/lib/platform -type f -name "$corretto_lib_pattern" -exec basename {} \; | head -n 1)"
```
michaelsembwever: a cleaner approach here without the corretto specifics here would be
```
if [ -d "${CASSANDRA_HOME}/lib/${operating_system}" ]; then
  for jar in ${CASSANDRA_HOME}/lib/${operating_system}/*.jar ; then
    CLASSPATH="$CLASSPATH:${jar}"
  done
fi
```

.build/build-resolver.xml
smiklosovic: @michaelsembwever it didnt work :( I noticed that while doing build in circle. You can have another shot at that ... it just ... didnt download anything.

bin/cassandra.in.sh
smiklosovic: you mean rather this

    platform=$(uname -m)
    if [ -d "${CASSANDRA_HOME}/lib/${platform}" ]; then
      for jar in ${CASSANDRA_HOME}/lib/${platform}/*.jar ; do
        CLASSPATH="$CLASSPATH:${jar}"
      done
    fi

.build/build-resolver.xml
smiklosovic: @michaelsembwever 

I am getting this

_resolver-dist-lib_get_files:
    [retry] Attempt [0]: error occurred; retrying...
    [retry] Attempt [1]: error occurred; retrying...
    [retry] Attempt [2]: error occurred; retrying...
    [retry] Attempt [0]: error occurred; retrying after 10 ms...

_resolver-dist-lib_get_files:
    [retry] Attempt [0]: error occurred; retrying...
    [retry] Attempt [1]: error occurred; retrying...
    [retry] Attempt [2]: error occurred; retrying...
    [retry] Attempt [1]: error occurred; retrying after 10 ms...

_resolver-dist-lib_get_files:
    [retry] Attempt [0]: error occurred; retrying...
    [retry] Attempt [1]: error occurred; retrying...
    [retry] Attempt [2]: error occurred; retrying...
    [retry] Attempt [2]: error occurred; retrying after 10 ms...

_resolver-dist-lib_get_files:
    [retry] Attempt [0]: error occurred; retrying...
    [retry] Attempt [1]: error occurred; retrying...
    [retry] Attempt [2]: error occurred; retrying...

....

/home/fermat/dev/cassandra/cassandra-instaclustr/cassandra/.build/build-resolver.xml:76: Task [resolver:resolve] failed after [3] attempts; giving up.
Error messages:
Reference all not found.
Reference all not found.
Reference all not found.
Reference all not found.

michaelsembwever: i'll take a look

bin/cassandra.in.sh
michaelsembwever: yes :)

.build/build-resolver.xml
michaelsembwever: cherry-pick this bad boy
https://github.com/thelastpickle/cassandra/commit/58f5a9c04a3c32e5da5298de3638424614138f38

smiklosovic: That helped! Thank you.

conf/cassandra.yaml
```diff
@@ -1375,6 +1375,30 @@ dynamic_snitch_reset_interval: 600000ms
 # until the pinned host was 20% worse than the fastest.
 dynamic_snitch_badness_threshold: 1.0
 
+# Configures Java crypto provider. By default, it will use DefaultCryptoProvider
+# which will install Amazon Correto Crypto Provider for x86 architecture.
```
michaelsembwever:
```suggestion
# which will install Amazon Correto Crypto Provider.
```

src/java/org/apache/cassandra/security/NoOpCryptoProvider.java
```diff
@@ -0,0 +1,57 @@
...
+
+package org.apache.cassandra.security;
+
+import java.util.Map;
+
+/**
+ * Crypto provider which does nothing. Handy for situations when a user
+ * wants to completely bypass crypto provider installation.
+ */
+public class NoOpCryptoProvider extends AbstractCryptoProvider
```
michaelsembwever:
```suggestion
public final class NoOpCryptoProvider extends AbstractCryptoProvider
```

i can't imagine any reason anyone would want to subclass this.

conf/cassandra.yaml
```diff
@@ -1375,6 +1375,30 @@ dynamic_snitch_reset_interval: 600000ms
...
+# will be used.
+#
+# If you want to fail the startup of a node if the provider was not installed properly,
+# you have to set property "fail_on_missing_provider" to "true" which is by default "false".
+# This property is not specific for DefaultCryptoProvider, it will be present to every provider with same logic.
+
+# To install custom provider, you need to extend org.apache.cassandra.security.AbstractCryptoProvider,
+# put it to class path and reference here to use it.
+#
+# If you want to completely bypass the installation of a crypto provider, please use NoOpCryptoProvider.
```
michaelsembwever:
```suggestion
# Amazon Correto Crypto Provider works currently for x86_64 and aarch_64 platforms.  If this provider fails it will fallback to the default crypto provider in the JRE.
#
# To force failure when the provider was not installed properly, set the property "fail_on_missing_provider" to "true".
#
# To bypass the installation of a crypto provider use NoOpCryptoProvider.
```

my attempt at more concise and direct language.

smiklosovic:
thanks @michaelsembwever , all suggestions will be applied. It seems like I fixed the tests as well!

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
```diff
@@ -0,0 +1,144 @@
...
+        {
+            failureMessage = format("The installation of %s was not successful, reason: %s",
+                                    getProviderClassAsString(), e.getMessage());
+        }
+
+        if (failureMessage != null)
+            if (failOnMissingProvider)
+                throw new ConfigurationException(failureMessage);
+            else
+                logger.warn(failureMessage);
```
michaelsembwever: braces on the outer if ?

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
```diff
@@ -0,0 +1,144 @@
...
+                installator().run();
+            }
+
+            if (isHealthyInstallation())
+            {
+                logger.info("{} successfully passed the healthiness check", getProviderName());
+            }
+            else
+            {
+                failureMessage = format("%s is not the highest priority provider. " +
```
michaelsembwever: is this the situation where the fallback to the JRE crypto provider happens?

conf/cassandra.yaml
```diff
@@ -1375,6 +1375,21 @@ dynamic_snitch_reset_interval: 600000ms
...
+#
+# Amazon Correto Crypto Provider works currently for x86_64 and aarch_64 platforms.
+# If this provider fails it will fall back to the default crypto provider in the JRE.
+#
+# To force failure when the provider was not installed properly, set the property "fail_on_missing_provider" to "true".
+#
+# To bypass the installation of a crypto provider use NoOpCryptoProvider.
+#
+#crypto_provider:
+#  - class_name: org.apache.cassandra.security.DefaultCryptoProvider
```
michaelsembwever: what does one specific if they want only the default crypto provider in the JRE ?

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
smiklosovic: @michaelsembwever I am getting different opinion on this from different people. You say this, somebody else says that either braces everywhere or no braces at all.

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
smiklosovic: Yes. And that happens if `fail_on_missing_provider` is `false`, if `true`, we fail the startup.

conf/cassandra.yaml
smiklosovic: They use `NoOpCryptoProvider`.

smiklosovic: We may probably rename that to `InBuiltJREProvider` because `NoOpCryptoProvider` sounds like it does not "do crypto" at all ...

.build/build-resolver.xml
```diff
@@ -212,6 +212,26 @@
         </retry>
         <mkdir dir="${local.repository}/org/apache/cassandra/deps/sigar-bin"/>
         <mkdir dir="${build.lib}/sigar-bin"/>
+        <mkdir dir="${build.lib}/x86_64"/>
+        <mkdir dir="${build.lib}/aarch_64"/>
```
smiklosovic: This should be changed to `aarch64` because that is the output of `uname -m` we try to match. I have a commit for that but it is not pushed yet because pushing would cancel the build in Circle I am running right now. Will do later.

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
smiklosovic: The assumption is that if it does not pass the healthcheck, it is not installed. So the provider which was there (the default one) has not changed, so the "fallback" is kind of automatic. We are not returning to anything per se.

smiklosovic: https://github.com/corretto/amazon-corretto-crypto-provider#verification-optional

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
michaelsembwever: my understanding is no-braces is only when there is one line that idents at the next level (here there is two) 🤷

conf/cassandra.yaml
michaelsembwever: yes, that was how i was reading it…

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
```diff
@@ -0,0 +1,144 @@
...
+                return;
+            }
+            else
+            {
+                installator().run();
+            }
+
+            if (isHealthyInstallation())
+            {
+                logger.info("{} successfully passed the healthiness check", getProviderName());
```
michaelsembwever:
```suggestion
                logger.info("{} health check ok", getProviderName());
```

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
```diff
@@ -0,0 +1,144 @@
...
+            }
+            else
+            {
+                failureMessage = format("%s is not the highest priority provider. " +
+                                        "The most probable cause is that Cassandra node is not running on the same architecture " +
+                                        "the provider library is for. Please place the architecture-specific library " +
+                                        "for %s to the classpath and try again. ",
+                                        getProviderName(),
+                                        getProviderClassAsString());
```
michaelsembwever:
```suggestion
                failureMessage = format("%s health check failed . " +
                                        "Check node's architecture (`uname -m`) is supported, see lib/<arch> subdirectories. The correct architecture-specific library " +
                                        "for %s needs to be on the classpath. ",
                                        getProviderName(),
                                        getProviderClassAsString());
```

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
smiklosovic: I hardened this logic a bit. We may uninstall the provider in case it was installed but healthcheck has failed. There might be other crypto providers which do it differently than Corretto.

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
smiklosovic: No worries I ll change that. I think that will be better, I am uncomfortable to not have them there even it works ...

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
```diff
@@ -0,0 +1,171 @@
...
+import org.slf4j.LoggerFactory;
+
+import org.apache.cassandra.exceptions.ConfigurationException;
+
+import static java.lang.String.format;
+
+public abstract class AbstractCryptoProvider
+{
+    protected static final Logger logger = LoggerFactory.getLogger(AbstractCryptoProvider.class);
+    public static final String FAIL_ON_MISSING_PROVIDER_KEY = "fail_on_missing_provider";
```
Mmuzaf, line 39: I would also expose this property to the `CassandraRelevantProperties`, so that we can easily override it if necessary.

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
```diff
@@ -0,0 +1,171 @@
...
+        {
+            logger.debug("Installation of a crypto provider was skipped.");
+            return;
+        }
+
+        String failureMessage = null;
+        Throwable t = null;
+        try
+        {
+            Class.forName(getProviderClassAsString());
```
Mmuzaf: I have seen a lot of cases using the following `FBUtilities#classForName` I think here is a good place for that as well. It also throws `ConfigurationException` which is good for us.

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    public String getProviderName()
+    {
+        return AmazonCorrettoCryptoProvider.PROVIDER_NAME;
+    }
+
+    @Override
+    public String getProviderClassAsString()
+    {
+        return "com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider";
```
Mmuzaf, line 48: I think this should be `AmazonCorrettoCryptoProvider.PACKAGE_PREFIX + AmazonCorrettoCryptoProvider.PROVIDER_NAME` and placed as the class public constant.

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
```diff
@@ -0,0 +1,67 @@
...
+
+import com.amazon.corretto.crypto.provider.AmazonCorrettoCryptoProvider;
+
+/**
+ * Default crypto provider tries to install AmazonCorrettoCryptoProvider.
+ * <p>
+ * The implementation falls back to in-built crypto provider in JRE if the installation
+ * is not successful.
+ */
+public class DefaultCryptoProvider extends AbstractCryptoProvider
```
Mmuzaf, line 32: In my opinion, a better name would be `AmazonCorrettoCryptoProvider` or something like that, so that a user reading the configuration file can see exactly what the default provider is without having to go into the depths of the code.

src/java/org/apache/cassandra/security/InBuiltJREProvider.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    @Override
+    public String getProviderClassAsString()
+    {
+        return InBuiltJREProvider.class.getName();
+    }
+
+    @Override
+    protected Runnable installator() throws Exception
```
Mmuzaf: Neither implementation throws a checked exception, should we still keep the method signature as 'throws Exception`?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -426,6 +429,8 @@ private static void applyAll() throws ConfigurationException
         //InetAddressAndPort cares that applySimpleConfig runs first
         applySSTableFormats();
 
+        applyCryptoProvider();
```
Mmuzaf, line 433: Should it also be part of the tool initialisation and client initialisation? I'm referring to `toolInitialization` and `clientInitialization` methods.

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
```diff
@@ -923,6 +923,8 @@ public Future<Void> shutdown(boolean graceful)
             
             error = parallelRun(error, executor, this::stopJmx);
 
+            DatabaseDescriptor.getCryptoProvider().uninstall();
```
Mmuzaf: The `uninstall' method throws a `SecurityException'. If an exception is thrown everything below is skipped. You handle this case in the `AbstractCryptoProvider`, but not for `Instance#shutdown(boolean)`. Is this correct?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1500,6 +1534,15 @@ private static IFailureDetector createFailureDetector(String detectorClassName)
         return detector;
     }
 
+    public static AbstractCryptoProvider getCryptoProvider()
+    {
+        return cryptoProvider;
+    }
+
+    public static void setCryptoProvider(AbstractCryptoProvider cryptoProvider)
```
Mmuzaf, line 1552: Do you think we need this method even if it is not used?

conf/cassandra.yaml
```diff
@@ -1375,6 +1375,21 @@ dynamic_snitch_reset_interval: 600000ms
 # until the pinned host was 20% worse than the fastest.
 dynamic_snitch_badness_threshold: 1.0
 
+# Configures Java crypto provider. By default, it will use DefaultCryptoProvider
+# which will install Amazon Correto Crypto Provider.
+#
+# Amazon Correto Crypto Provider works currently for x86_64 and aarch_64 platforms.
```
Mmuzaf, line 1381: I think it is worth adding a link to the project with documentation pages. Wdyt?
https://github.com/corretto/amazon-corretto-crypto-provider

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1217,6 +1222,35 @@ public static void applySslContext()
...
+            conf.crypto_provider.parameters = new HashMap<>();
+
+        if (conf.crypto_provider.class_name == null)
+            throw new ConfigurationException("Failed to initialize crypto provider, class_name cannot be null");
+
+        conf.crypto_provider.parameters.putIfAbsent(AbstractCryptoProvider.FAIL_ON_MISSING_PROVIDER_KEY, "false");
+
+        try
+        {
+            Class<?> cryptoProviderClass = Class.forName(conf.crypto_provider.class_name);
```
Mmuzaf: I think we can use `FBUtilities#classForName` here.
OR
alternatively use `FBUtilities#construct(java.lang.String, java.lang.String)` and pass the `ParameterizedClass` instance from the `config` directly to the `install` method.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1217,6 +1222,35 @@ public static void applySslContext()
...
+            cryptoProvider = (AbstractCryptoProvider)cryptoProviderClass.getConstructor(Map.class).newInstance(conf.crypto_provider.parameters);
+            cryptoProvider.install();
```
Mmuzaf, line 1253: In my humble opinion, it is good for the install method to only throw runtime exceptions, so we don't need to handle them here and they are passed directly to `CassandraDaemon#activate()` for handling. In fact, I have checked the `install()` method and it throws only runtime exceptions.

src/java/org/apache/cassandra/security/InBuiltJREProvider.java
```diff
@@ -0,0 +1,57 @@
...
+
+package org.apache.cassandra.security;
+
+import java.util.Map;
+
+/**
+ * Crypto provider which does nothing. Handy for situations when a user
+ * wants to completely bypass crypto provider installation.
+ */
+public final class InBuiltJREProvider extends AbstractCryptoProvider
```
Mmuzaf: I would call it `NoOpCryptoProvider` instead, but it doesn't really matter.

.build/parent-pom-template.xml
```diff
@@ -233,12 +233,52 @@
...
     <developerConnection>scm:https://gitbox.apache.org/repos/asf/cassandra.git</developerConnection>
     <url>https://gitbox.apache.org/repos/asf?p=cassandra.git;a=tree</url>
   </scm>
+
+  <profiles>
+    <profile>
+      <id>x86_64</id>
+      <activation>
+        <os>
+          <arch>!aarch64</arch>
```
Mmuzaf, line 253: Should we also add `<family>unix</family>`? This will look more correct even if we never run on other operating systems.
Is the condition correct? Should we place here `x86_64` to avoid matching other types of `x86`?

.build/parent-pom-template.xml
```diff
@@ -233,12 +233,52 @@
...
+        </os>
+      </activation>
+      <dependencies>
```
Mmuzaf, line 256: This is the `cassandra-parent` pom, so it should probably be placed under the `<dependencyManagement>` section to avoid any problems for those projects that depend on this pom, but still don't want to see those dependencies in the classpath.
The `<dependency>` probably should be used in the `cassandra-deps-template.xml` which defines the `cassandra-all` artifact. So, the question here is - is it safe putting this dependency in the classpath for all?

test/unit/org/apache/cassandra/security/CryptoProviderTest.java
```diff
@@ -0,0 +1,225 @@
...
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.spy;
+
+public class CryptoProviderTest
```
Mmuzaf, line 57: I think we can add a test, that reads the configuration file something the same as the `cassandra-pem-sslcontextfactory.yaml` is used and check that `AmazonCorrettoCryptoProvider` is used in that case.

Mmuzaf:
`CASSANDRA-18624 various improvements #2505` 
The name of the pull request is a bit confusing :-)

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
smiklosovic: The reason it is `String` is that if it is not on the class path, then it would throw what ... `ClassNotFoundException`? So I made it String on purpose and I am using `Class.forName` (which can be changed to FBUtilities as you suggested)

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
smiklosovic: why not

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
smiklosovic: Seems like an overkill to me but why not.

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
smiklosovic: Matter of taste really. @michaelsembwever what do you think? I do not like the fact that we are explicitly depending on something "Amazon" in the config.. I know it might probably look like a weird reason though.

src/java/org/apache/cassandra/security/InBuiltJREProvider.java
smiklosovic: good catch, as it is just returning Runnable, not necessary.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
smiklosovic: I dont know yet. Does client or tool do crypto? I think so ... like ... nodetool? But one should verify if the library is going to be on the class path the same way Cassandra's class path has it. I think that scripts are executed somehow differently and putting this on CP of every tooling is a little bit tricky.

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
smiklosovic: If we catch it then what do you want to do with that ... to log it? So we do all the stuff but crypto provider will remain to be installed. Then on the next test this will fail horribly (been there, done that) because it will use old crypto provider with new classloader of new Instance. 

I would rather throw this so people are aware of the fact that something strange is going on with their setup. Otherwise it is just lost somewhere as a log which is easily overlooked.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
smiklosovic: Yes lets keep it there.

conf/cassandra.yaml
smiklosovic: Good idea but what if that project changes the name or something like that ... links tend to break.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
smiklosovic: I like the fact that this needs to be explicitly taken care of. I know it is probably overreach but installation might happen elsewhere over time, people can install / uninstall here and there. God knows how this is going to be used.

src/java/org/apache/cassandra/security/InBuiltJREProvider.java
smiklosovic: It was called like that :D But we came to conclusion with Mick that _NoOpCryptoProvider_ sounds like you are going to use a provider which is _not doing any crypto at all_. Calling it like this means that you will get same crypto which is in JRE which is _doing some crypto_.

.build/parent-pom-template.xml
smiklosovic: What other types, like 32 bits? Does Cassandra even run on 32 bits anyway? I would just keep it like this.

If we run this like ... on FreeBSD, it will put x86 lib on the class path, then it will try to load it and it will presumably fail and log will be printed.

I do not think anybody will ever build Cassandra with Maven on FreeBSD or any other unix for that matter. Like ...why? People are going to at most run it (if even that) and they are covered already as they will have it in `lib/x86_64` or `lib/aarch64`. So them running Cassandra on these platforms will not have any library added to the class path and again this is catched by the exception handling we have in place for the provider.

.build/parent-pom-template.xml
smiklosovic: We have been there. The problem with dependency management is that it can not be specified in profiles. I mean, you can technically put it there but then it is not visible in the child, you can not reference that from there.

 If we put that to dependency management outside profiles in parent, then we can not switch between the platform specific dependencies.

test/unit/org/apache/cassandra/security/CryptoProviderTest.java
smiklosovic: The tests are checking already that the provider is installed. `testCryptoProviderInstallation` checks that, for example.

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
Mmuzaf: I don't insist, just have a concern if we are adding a crypto library from the classpath it could cause some problems for the tools/clients. Perhaps it is better not to override the flag itself, but to override the `class_name` with a system property, so that if something goes wrong we can disable the 'Amazon' dependency and reset it back to the default. Wdyt?

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
Mmuzaf: This is up to you, just a minor suggestion :-)

I also feel something strange seeing something non-cassandra in the config, do we need such configuration in the yaml? Are there any cons of just using a simple system property with the class name instead?

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
Mmuzaf: I saw that the class already imports the `AmazonCorrettoCryptoProvider`, that's why suggested this change. If you want Strings then both `getProviderName` and `getProviderClassAsString` should be simple strings :-)

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
smiklosovic: but that can not be added here because it is "too late" (e.g. in the constructor). We would need to do that in DatabaseDescriptor because that is the place where this class is instantiated.

conf/cassandra.yaml
Mmuzaf: This is up to you :-)
I think if the link can save a user a few seconds, it's better to add it. In case it gets broken at some point in the future then we have an LHF for someone to fix.

src/java/org/apache/cassandra/security/InBuiltJREProvider.java
Mmuzaf: Naming is not my talent :-) so we can leave it as is.

.build/parent-pom-template.xml
Mmuzaf: My concern is about putting something in the classpath that doesn't match the architecture type, so even if the cassandra is able to handle this case it's good, but the other applications may not. 
If you think this will work, let's resolve this comment, I just wanted to point out a potential problem, but in fact, it may not even exist :-)

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
smiklosovic: It is good as it is ... I wont change that.

conf/cassandra.yaml
smiklosovic: I wont add it there. The link would be something for "ah it is this thing" rather than a link which is crucial for the configuration itself. There is plethora of cases where links like this might be added over all documentation. Are we going to add links to Facebook's Zstd so a user knows what Zstd is? etc etc

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
michaelsembwever: +1 for use of FBUtilities

src/java/org/apache/cassandra/security/DefaultCryptoProvider.java
michaelsembwever: 🤷

src/java/org/apache/cassandra/security/InBuiltJREProvider.java
michaelsembwever: i'm leaning on `JREProvider` 

NoOp means no crypto to me.

.build/parent-pom-template.xml
michaelsembwever: for now we need something as a default even if it doesn't successfully load the .so files. 

that's why it's `! aarch64` (as confusing as that might first read.)

let's add a comment explaining the contortion…

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
smiklosovic: classForName is fine. I like the fact that people have to use constructor with parameters (as map). Otherwise they would not see so easily that it is configurable and we are passing some map to some method. Parameters in constructors are common pattern in Cassandra configuration on ParameterizedClass.

src/java/org/apache/cassandra/security/InBuiltJREProvider.java
smiklosovic: sure, we can shorten to JREProvider

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
smiklosovic: @Mmuzaf btw why not to have both property for class as well as for that flag?

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
Mmuzaf: As a fix I was thinking about the following `error = parallelRun(error, executor, () -> DatabaseDescriptor.getCryptoProvider().uninstall());`, but I think as the uninstall is called only in tests it should be safe from all sides.

smiklosovic: @Mmuzaf yes that works! wrapping it in parallelRun, good catch.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
Mmuzaf: This is the hardest thing for me, so me neither. I need some time to get into the depths of the code to find a clear answer, so if we are trying to fit in release deadlines I would probably suggest adding this (or not) in another issue.

src/java/org/apache/cassandra/security/AbstractCryptoProvider.java
Mmuzaf: I was thinking of a simple fix that would allow us to disable a new crypto provider without having to update the configuration file itself, which might not be very fast in a production environment that usually requires additional coordination between teams. Both flags also look OK too, but all that is up to you.

smiklosovic: added both

michaelsembwever (approved):

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1219,6 +1225,42 @@ public static void applySslContext()
         }
     }
 
+    public static void applyCryptoProvider()
+    {
+        if (TEST_SKIP_CRYPTO_PROVIDER_INSTALLATION.getBoolean())
+            return;
+
+        if (conf.crypto_provider == null)
+            conf.crypto_provider = new ParameterizedClass(DefaultCryptoProvider.class.getName(), null);
```
michaelsembwever:
```suggestion
        if (conf.crypto_provider == null)
            conf.crypto_provider = new ParameterizedClass(JREProvider.class.getName(), null);
```

michaelsembwever (changes requested):
yaml provides accp default for new clusters.

DD provides jre default for existing clusters.

michaelsembwever (approved):

.build/build-resolver.xml
Mmuzaf: Perhaps it would be better to just move these jars to the lib/{$platform} folder after they have been resolved and downloaded into the `jar` folder, no?

PR #2506, opened 2023-07-21
Lift MessagingService.minimum_version to 40 (CASSANDRA-18314)

michaelsembwever:


src/java/org/apache/cassandra/db/filter/ColumnFilter.java
```diff
@@ -1004,8 +1003,6 @@ private void serializeRegularAndStaticColumns(RegularAndStaticColumns regularAnd
         public ColumnFilter deserialize(DataInputPlus in, int version, TableMetadata metadata) throws IOException
         {
             int header = in.readUnsignedByte();
-            // The meaning of isFetchAll is actually different for pre-4.0 versions and for 4.0+ versions
-            // In 4.0+ it meant is fetch all regulars
```
adelapena, line 1008: There is a similar obsolete comment on `FETCH_ALL_MASK`. We could also rename both `isFetchAll` and `FETCH_ALL_MASK` to `isFetchAllRegulars` and `FETCH_ALL_REGULARS_MASK`.

src/java/org/apache/cassandra/hints/HintsDescriptor.java
```diff
@@ -231,8 +231,6 @@ static int messagingVersion(int hintsVersion)
     {
         switch (hintsVersion)
         {
-            case VERSION_30:
-                return MessagingService.VERSION_30;
```
adelapena, line 235: This leaves `HintsDescriptor.VERSION_30` unused, it can probably be removed.

src/java/org/apache/cassandra/net/Message.java
```diff
@@ -711,15 +708,12 @@ private Serializer()
 
         public <T> void serialize(Message<T> message, DataOutputPlus out, int version) throws IOException
         {
-            if (version >= VERSION_40)
-                serializePost40(message, out, version);
-            else
-                serializePre40(message, out, version);
+            serializePost40(message, out, version);
```
adelapena: All the methods that were named `*Post40` can be renamed or inlined into the called methods, since they don't have a `*Pre40` counterpart anymore.

src/java/org/apache/cassandra/net/OutboundConnectionInitiator.java
```diff
@@ -350,68 +349,35 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out)
...
                                 break;
-                            case LZ4:
-                                frameEncoder = FrameEncoderLegacyLZ4.instance;
+                            case UNPROTECTED:
+                                frameEncoder = FrameEncoderUnprotected.instance;
                                 break;
                         }
 
-                        result = (Result<SuccessType>) messagingSuccess(ctx.channel(), requestMessagingVersion, frameEncoder.allocator());
+                        result = (Result<SuccessType>) messagingSuccess(ctx.channel(), useMessagingVersion, frameEncoder.allocator());
```
adelapena, line 373: Nit: we can add `@SuppressWarnings("unchecked")`

src/java/org/apache/cassandra/net/ParamType.java
```diff
@@ -46,19 +45,9 @@ public enum ParamType
     FORWARD_TO                  (0,  "FWD_TO",          ForwardingInfo.serializer),
     RESPOND_TO                  (1,  "FWD_FRM",         fwdFrmSerializer),
 
-    @Deprecated
-    FAILURE_RESPONSE            (2,  "FAIL",            LegacyFlag.serializer),
```
adelapena, line 50: Maybe we should add a comment on `ParamType.id` saying that the ids are nor continuous because some deprecated param types have been removed? Or maybe left a comment for every removed param type?

src/java/org/apache/cassandra/service/pager/PagingState.java
```diff
@@ -402,7 +402,7 @@ public static RowMark create(TableMetadata metadata, Row row, ProtocolVersion pr
             {
                 // We froze the serialization version to 3.0 as we need to make this this doesn't change (that is, it has to be
                 // fix for a given version of the protocol).
-                mark = Clustering.serializer.serialize(row.clustering(), MessagingService.VERSION_30, makeClusteringTypes(metadata));
+                mark = Clustering.serializer.serialize(row.clustering(), MessagingService.VERSION_40, makeClusteringTypes(metadata));
```
adelapena, line 405: The comment right above seems to need an update.

src/java/org/apache/cassandra/net/Message.java
michaelsembwever: i thought of that… but what about making forward-merge simpler? 

any harm in keeping it for another release cycle, cleaning it later?

src/java/org/apache/cassandra/service/pager/PagingState.java
michaelsembwever: what about
```
// We froze the serialization version to VERSION_30 as we need to make sure this this doesn't change
//  It got bumped to VERSION_40 when VERSION_30 got dropped, knowing it didn't change and the latter is not used in a 5.x cluster
```

src/java/org/apache/cassandra/db/filter/ColumnFilter.java
michaelsembwever: what about `isFetchAll` in `Builder.build()` ?   and the comments in that method …?

if we are removing the three `isUpgradingFromVersionLowerThan*()` methods, there's a few comments that need fixing.  is this in scope?

michaelsembwever: see what i did in the last commit https://github.com/apache/cassandra/pull/2506/commits/8782e8dda798b3fbe09dc01d6ecfdddfe5455b4e

adelapena: I guess `isFetchAll` in `Builder#build()` can also be renamed to `isFetchAllRegulars`, since it been true means that all regulars will be fetched, but fetching also statics (that is, all columns) depends on the queried columns and `returnStaticContentOnPartitionWithNoRows`

src/java/org/apache/cassandra/net/ParamType.java
```diff
@@ -40,25 +39,17 @@
  * will skip over any params it doesn't recognise.
  *
  * Please don't add boolean params here. Extend and use {@link MessageFlag} instead.
+ *
+ * Do not re-use old, of fill gaps in the sequence of, ids. New IDs must be higher.
```
adelapena: Nit: Shouldn't it be `or/nor` in `of fill gaps in...`?

src/java/org/apache/cassandra/service/pager/PagingState.java
```diff
@@ -400,9 +400,9 @@ public static RowMark create(TableMetadata metadata, Row row, ProtocolVersion pr
             }
             else
             {
-                // We froze the serialization version to 3.0 as we need to make this this doesn't change (that is, it has to be
-                // fix for a given version of the protocol).
-                mark = Clustering.serializer.serialize(row.clustering(), MessagingService.VERSION_30, makeClusteringTypes(metadata));
+                // We froze the serialization version to 3.0 as we need to make sure this this doesn't change
+                //  It got bumped to 4.0 when 3.0 got dropped, knowing it didn't chamge
```
adelapena: Nit: typo
```suggestion
                //  It got bumped to 4.0 when 3.0 got dropped, knowing it didn't change
```

michaelsembwever: updated existing commit.

src/java/org/apache/cassandra/net/ParamType.java
michaelsembwever: updated existing commit.

src/java/org/apache/cassandra/db/filter/ColumnFilter.java
```diff
@@ -267,29 +204,13 @@ public static ColumnFilter selection(RegularAndStaticColumns columns)
      * A filter that fetches all columns for the provided table, but returns
      * only the queried ones.
      */
+    @VisibleForTesting
```
michaelsembwever, line 199: i could not find anywhere this method was used but the tests…

src/java/org/apache/cassandra/net/Message.java
adelapena: If we leave it for later we will still be complicating merge. In general I prefer simpler code over simpler merges, and since the ticket is about cleaning up it seem that renaming the methods is in scope. I wouldn't be against doing it later in a separate ticket though, but I think in that case we should probably leave some TODOs about that so it's less confusing for readers. wdyt?

michaelsembwever: any objection to keeping the separate methods but `s/Post40/Impl/` ? 
 (where the caller is not a single-line delegating method, those i will inline)

michaelsembwever: see commit: https://github.com/apache/cassandra/pull/2506/commits/c786e6687d2ecb810fd5e22b9d34d54ca9076438

adelapena: I see that the methods have already inlined, which is what impacts merge the most. The remaining *Impl methods aren't implementations of other methods, I think, but private utility methods. So instead of renaming with `s/Post40/Impl/` we can just remove the `Post40`/`Impl` suffix. The methods I mean are:
* serializeHeaderImpl
* deserializeHeaderImpl
* skipHeaderImpl
* inferMessageSizeImpl
* extractParamsSizeImpl

There shouldn't be a lot of difference between the `s/Post40/Impl/` rename and just dropping the suffix when merging forward.

test/unit/org/apache/cassandra/db/filter/ColumnFilterTest.java
```diff
@@ -518,8 +518,6 @@ else if ("3.11".equals(clusterMinVersion) || (returnStaticContentOnPartitionWith
 
     private void testRoundTrips(ColumnFilter cf)
     {
-        testRoundTrip(cf, MessagingService.VERSION_30);
-        testRoundTrip(cf, MessagingService.VERSION_3014);
```
adelapena, line 522: The two signatures of `testRoundTrips` can be merged since we are always using the same MS versions.

test/unit/org/apache/cassandra/net/FramingTest.java
```diff
@@ -271,7 +270,7 @@ public void testRandomLegacy()
     private void testRandomLegacy(int count)
     {
         SecureRandom seeds = new SecureRandom();
-        for (int messagingVersion : new int[] { VERSION_30, VERSION_3014, current_version})
+        for (int messagingVersion : new int[] { VERSION_40 })
```
adelapena: We don't need to loop here, there is only a single value. Also, do we need the `FrameDecoderLegacy` class at all?

src/java/org/apache/cassandra/net/Message.java
michaelsembwever: 💯

test/unit/org/apache/cassandra/net/HandshakeTest.java
```diff
@@ -193,48 +193,31 @@ public void testSendCompatibleOldVersionPre40() throws InterruptedException
...
-
-    @Test
-    public void testSendIncompatibleOldVersion40() throws InterruptedException
+    @Test // fairly contrived case, but since we introduced logic for testing we need to be careful it doesn't make us worse
+    public void testSendToFuturePost40BelievedToBePre40() throws InterruptedException, ExecutionException
     {
         try
         {
-            Assert.fail(Objects.toString(handshake(VERSION_30, VERSION_30, VERSION_30, current_version, current_version)));
+            Assert.fail(Objects.toString(handshake(VERSION_30, VERSION_30, current_version, VERSION_30, current_version + 1)));
```
adelapena: Totally optional, but it can be simplified to:
```java
public void testSendToFuturePost40BelievedToBePre40()
{
    Assertions.assertThatThrownBy(() -> Objects.toString(handshake(VERSION_30, VERSION_30, current_version, VERSION_30, current_version + 1)))
              .hasRootCauseInstanceOf(ClosedChannelException.class);
}
```

adelapena: Same for `testSendCompatibleOldVersion40` and `testSendIncompatibleOldVersion40` right above.

test/unit/org/apache/cassandra/net/FramingTest.java
michaelsembwever: oh, that's a whole new unravel, give me an hour…

michaelsembwever: done.

src/java/org/apache/cassandra/net/InboundConnectionInitiator.java
```diff
@@ -321,78 +319,42 @@ void initiate(ChannelHandlerContext ctx, ByteBuf in) throws IOException
...
-                    if (initiate.type.isStreaming())
-                        setupStreamingPipeline(initiate.from, ctx);
-                    else
-                        setupMessagingPipeline(initiate.from, useMessagingVersion, initiate.acceptVersions.max, ctx.pipeline());
-                }
+            if (initiate.acceptVersions.min > accept.max)
+            {
+                logger.info("peer {} only supports messaging versions higher ({}) than this node supports ({})", ctx.channel().remoteAddress(), initiate.acceptVersions.min, current_version);
+                failHandshake(ctx);
+                return;
```
Mmuzaf: `return` is unnecessary now

src/java/org/apache/cassandra/net/InboundConnectionInitiator.java
```diff
@@ -321,78 +319,42 @@ void initiate(ChannelHandlerContext ctx, ByteBuf in) throws IOException
...
+            }
+            else if (initiate.acceptVersions.max < accept.min)
+            {
+                logger.info("peer {} only supports messaging versions lower ({}) than this node supports ({})", ctx.channel().remoteAddress(), initiate.acceptVersions.max, minimum_version);
+                failHandshake(ctx);
+                return;
```
Mmuzaf: `return` is unnecessary now

michaelsembwever:
all comments addressed.
rebased off trunk.
CI runs: https://app.circleci.com/pipelines/github/michaelsembwever/cassandra/201

src/java/org/apache/cassandra/net/FrameDecoderLegacy.java
```diff
@@ -1,184 +0,0 @@
...
-
-/**
- * {@link InboundMessageHandler} operates on frames that adhere to a certain contract
- * (see {@link FrameDecoder.IntactFrame} and {@link FrameDecoder.CorruptFrame} javadoc).
- *
- * Legacy (pre-4.0) messaging protocol does not natively support framing, however. The job
- * of {@link FrameDecoderLegacy} is turn a raw stream of messages, serialized back to back,
- * into a sequence of frames that adhere to 4.0+ conventions.
- */
-class FrameDecoderLegacy extends FrameDecoder
```
adelapena, line 36: The removal of the legacy frame decoders leaves some obsolete comments  in the class JavaDoc for `FrameDecoder`: https://github.com/apache/cassandra/blob/530750f780b1b4a683263337574f2fa5cabd4dce/src/java/org/apache/cassandra/net/FrameDecoder.java#L54-L57

adelapena: One of the `if` branches in `FrameDecoder#channelRead(ChannelHandlerContext, Object)` also seems related with the disappeared legacy decoders: https://github.com/apache/cassandra/blob/530750f780b1b4a683263337574f2fa5cabd4dce/src/java/org/apache/cassandra/net/FrameDecoder.java#L279-L282

src/java/org/apache/cassandra/net/InboundConnectionInitiator.java
```diff
@@ -533,26 +478,17 @@ void setupMessagingPipeline(InetAddressAndPort from, int useMessagingVersion, in
             {
                 case LZ4:
                 {
-                    if (useMessagingVersion >= VERSION_40)
-                        frameDecoder = FrameDecoderLZ4.fast(allocator);
-                    else
-                        frameDecoder = new FrameDecoderLegacyLZ4(allocator, useMessagingVersion);
+                    frameDecoder = FrameDecoderLZ4.fast(allocator);
```
adelapena, line 480: The comments for the `OutboundConnectionSettings.Framing` enum values have some references to the MS version that I think we don't need anymore: https://github.com/apache/cassandra/blob/530750f780b1b4a683263337574f2fa5cabd4dce/src/java/org/apache/cassandra/net/OutboundConnectionSettings.java#L53-L60

src/java/org/apache/cassandra/net/OutboundConnectionInitiator.java
```diff
@@ -350,68 +350,35 @@ protected void decode(ChannelHandlerContext ctx, ByteBuf in, List<Object> out)
...
-                    else if (peerMessagingVersion < settings.acceptVersions.min)
-                        result = incompatible(-1, peerMessagingVersion);
                     else
-                        result = retry(peerMessagingVersion);
-
-                    if (result.isSuccess())
                     {
-                        ConfirmOutboundPre40 message = new ConfirmOutboundPre40(settings.acceptVersions.max, settings.from);
-                        AsyncChannelPromise.writeAndFlush(ctx, message.encode());
+                        result = (Result<SuccessType>) streamingSuccess(ctx.channel(), useMessagingVersion);
```
adelapena, line 377: I think `OutboundConnectionSettings.maybeWithSecurePort`, at the end of this method, can be simplified now that the min MS version is 4.0.

adelapena:
I think another thing that can be slightly simplified now that we don't expect 3.0 nodes is `QueryProcessor#useNewPreparedStatementBehaviour()`, where we don't need `QueryProcessor.NEW_PREPARED_STATEMENT_BEHAVIOUR_SINCE_30` nor `QueryProcessor.NEW_PREPARED_STATEMENT_BEHAVIOUR_SINCE_3X`.

The JavaDoc for `QueryProcessor#prepare` should also be updated, since it contains some references to upgrades from 3.0.26 and 3.11.

test/unit/org/apache/cassandra/db/filter/ColumnFilterTest.java
```diff
@@ -86,7 +86,7 @@ public class ColumnFilterTest
     @Parameterized.Parameters(name = "{index}: clusterMinVersion={0}")
     public static Collection<Object[]> data()
     {
-        return Arrays.asList(new Object[]{ "3.0" }, new Object[]{ "3.11" }, new Object[]{ "4.0-rc1" }, new Object[]{ "4.0" });
+        return (Collection) Arrays.asList(new Object[]{ "4.0" });
```
adelapena: It seems that this test doesn't need to be parameterized anymore, since there is only a single parameter. The var `clusterMinVersion` can be removed.

src/java/org/apache/cassandra/db/filter/ColumnFilter.java
```diff
@@ -182,67 +180,6 @@ boolean areAllFetchedColumnsQueried()
...
     }
 
-    /**
-     * Returns {@code true} if there are pre-4.0-rc2 nodes in the cluster, {@code false} otherwise.
-     *
-     * <p>ColumnFilters from 4.0 releases before RC2 wrongly assumed that fetching all regular columns and not
-     * the static columns was enough. That was not the case for queries that needed to return rows for empty partitions.
-     * See CASSANDRA-16686 for more details.</p>
-     */
-    private static boolean isUpgradingFromVersionLowerThan40RC2()
```
adelapena, line 192: The JavaDoc of `ColumnFilter.FetchingStrategy.ALL_COLUMNS` refers to "mixed version clusters where some nodes have a version lower than 4.0", that reference can now be removed.

src/java/org/apache/cassandra/db/filter/ColumnFilter.java
```diff
@@ -182,67 +180,6 @@ boolean areAllFetchedColumnsQueried()
...
-    {
-        if (Gossiper.instance.isUpgradingFromVersionLowerThan(CassandraVersion.CASSANDRA_4_0_RC2))
-        {
-            logger.trace("ColumnFilter conversion has been applied so that static columns will not be fetched because there are pre 4.0-rc2 nodes in the cluster");
```
adelapena, line 196: After getting rid of this code `ColumnFilter.logger` is not used anymore. We can get rid of it too.

src/java/org/apache/cassandra/db/filter/ColumnFilter.java
adelapena: Me neither, +1

test/unit/org/apache/cassandra/db/filter/ColumnFilterTest.java
michaelsembwever: what about leaving it in, since there will be new versions 

(simple code i would clean, but the parameterisation of the test is nice reminder to keep in place)

michaelsembwever:
> I think another thing that can be slightly simplified now that we don't expect 3.0 nodes is QueryProcessor

can we do that in a separate ticket, not related to MessagingService (see parent ticket)

src/java/org/apache/cassandra/db/filter/ColumnFilter.java
michaelsembwever: done

src/java/org/apache/cassandra/db/filter/ColumnFilter.java
michaelsembwever: done

src/java/org/apache/cassandra/net/OutboundConnectionInitiator.java
michaelsembwever: i've done this. but it's not strictly due to the MessagingService version (but related and consequential)

src/java/org/apache/cassandra/net/InboundConnectionInitiator.java
michaelsembwever: done

src/java/org/apache/cassandra/net/OutboundConnectionSettings.java
```diff
@@ -508,20 +500,4 @@ static boolean shouldCompressConnection(IEndpointSnitch snitch, InetAddressAndPo
                || ((DatabaseDescriptor.internodeCompression() == Config.InternodeCompression.dc) && !isInLocalDC(snitch, localHost, remoteHost));
     }
 
-    private static InetAddressAndPort maybeWithSecurePort(InetAddressAndPort address, int messagingVersion, boolean isEncrypted)
-    {
-        if (!isEncrypted || messagingVersion >= VERSION_40)
```
adelapena, line 513: Nit: this leaves an unused import of `MessagingService.VERSION_40` that will make the build fail.

test/unit/org/apache/cassandra/db/filter/ColumnFilterTest.java
adelapena: Future messaging versions won't necessarily affect `ColumnFilter`. For example, `VERSION_50` doesn't change anything about column filters. In any case, I think we should remove the four `"4.0".equals(clusterMinVersion)` checks since they are no-op.

adelapena:
> I think another thing that can be slightly simplified now that we don't expect 3.0 nodes is QueryProcessor

> can we do that in a separate ticket, not related to MessagingService (see parent ticket)

Yeah, it can be a separate ticket since the min cluster version doesn't come from `MessagingService`.

Edit: Created CASSANDRA-18695 for not considering 3.x nodes in `QueryProcessor#useNewPreparedStatementBehaviour`

Mmuzaf:
These methods are not used and I think can be removed:
- `FwdFrmSerializer#pre40DeserializeWithLength`
- `Verb#toPre40Verb`

Do we need to generate benchmarks for a 'pre40'?
`MessageOutBench#serializePre40`

Mmuzaf:
```java
            // if the other node is pre-4.0, it will respond only with its maxMessagingVersion
            if (maxMessagingVersion < VERSION_40 || handshakeMessagingVersion < VERSION_40)
                return new Accept(useMessagingVersion, maxMessagingVersion);        
```

Should we return `null` for this case? This is the `HandshakeProtocol` class.

test/unit/org/apache/cassandra/db/filter/ColumnFilterTest.java
michaelsembwever: yeah you're right, my thinking applies really only to `*.net.*` tests

michaelsembwever:
> This is the HandshakeProtocol class.

changing that breaks HandshakeTest 🤷

src/java/org/apache/cassandra/net/OutboundConnectionSettings.java
michaelsembwever: done

test/unit/org/apache/cassandra/db/filter/ColumnFilterTest.java
michaelsembwever: done

test/unit/org/apache/cassandra/db/filter/ColumnFilterTest.java
```diff
@@ -52,14 +52,10 @@
 
 import static org.junit.Assert.assertEquals;
 
-@RunWith(Parameterized.class)
 public class ColumnFilterTest
 {
     private static final ColumnFilter.Serializer serializer = new ColumnFilter.Serializer();
 
-    @Parameterized.Parameter
```
adelapena, line 60: This leaves a few unused imports that will break CI

adelapena:
```
 // if the other node is pre-4.0, it will respond only with its maxMessagingVersion
if (maxMessagingVersion < VERSION_40 || handshakeMessagingVersion < VERSION_40)
   return new Accept(useMessagingVersion, maxMessagingVersion);        
```

> Should we return null for this case? This is the HandshakeProtocol class.

> changing that breaks HandshakeTest 🤷

Returning null there doesn't break `HandshakeTest` for me, at least locally. What failure are you finding?

adelapena:
> Do we need to generate benchmarks for a 'pre40'?
> `MessageOutBench#serializePre40`

Good catch. That benchmark is actually broken, it throws `IllegalStateException: Unkown serialization version 10`.

src/java/org/apache/cassandra/net/Message.java
```diff
@@ -1287,7 +986,7 @@ private Map<ParamType, Object> deserializeParams(DataInputPlus in, int version)
          */
         private Map<ParamType, Object> extractParams(ByteBuffer buf, int readerIndex, int version) throws IOException
         {
-            long count = version >= VERSION_40 ? getUnsignedVInt(buf, readerIndex) : buf.getInt(readerIndex);
+            long count = getUnsignedVInt(buf, readerIndex);
```
adelapena, line 981: Nit: The JavaDoc for this method has a mention of `post-4.0` that we can remove. Also, the class JavaDoc for `MessageSerializer` has a list of "Legacy notes" that can also be removed.

src/java/org/apache/cassandra/net/ParamType.java
```diff
@@ -40,25 +39,17 @@
  * will skip over any params it doesn't recognise.
  *
  * Please don't add boolean params here. Extend and use {@link MessageFlag} instead.
+ *
+ * Do not re-use old, nor fill gaps in the sequence of, ids.  New IDs must be higher.
```
adelapena, line 41: `ParamType#lookUpByAlias` is not used anymore, so we can get rid of it. With this method gone, `ParamType#legacyAlias` and `ParamType.aliasToTypeMap` can also be removed, and the constructor can take only the id and the serializer.

michaelsembwever:
> Returning null there doesn't break HandshakeTest for me, at least locally. What failure are you finding?

On the outbound (encode) side, you get a npe. 
Patch to deal with this (but still catch old versions so we can test ClosedChannelException is still correctly thrown) en route…

src/java/org/apache/cassandra/net/Message.java
michaelsembwever: done

src/java/org/apache/cassandra/net/ParamType.java
michaelsembwever: done

test/unit/org/apache/cassandra/db/filter/ColumnFilterTest.java
michaelsembwever: done

michaelsembwever:
> Patch to deal with this (but still catch old versions so we can test ClosedChannelException is still correctly thrown) en route…

Done.  See https://github.com/apache/cassandra/pull/2506/commits/168e986e1b6d4a7cef1f8ec582b223ead3f40a05

We don't test listening (outbound) on unsupported versions anymore, but preconditions are in place to prevent it from happening.

Mmuzaf:
> We don't test listening (outbound) on unsupported versions anymore, but preconditions are in place to prevent it from happening.

Thanks, the idea was exactly as you mentioned in the comment: close connections on a handshake if the node version is obsolete! :-)

test/unit/org/apache/cassandra/net/HandshakeTest.java
```diff
@@ -146,94 +144,38 @@ public void testSendIncompatibleFutureVersion() throws InterruptedException, Exe
...
+                                                expectSuccess ? Result.Outcome.SUCCESS : Result.Outcome.INCOMPATIBLE, result.outcome);
+
+                                if (expectSuccess)
+                                {
+                                    Assert.assertEquals(String.format("wrong agreed messagingVersion for outMin %s outMax %s inMin %s inMax %s", outMin.value, outMax.value, inMin.value, inMax.value),
+                                                    Math.min(outMax.value, inMax.value), result.success().messagingVersion);
+                                    result.success().channel.close();
+                                }
+                            }
         }
```
Mmuzaf, line 178: nit: wrong indentation

test/unit/org/apache/cassandra/streaming/async/StreamingInboundHandlerTest.java
```diff
@@ -56,7 +56,6 @@
 
 public class StreamingInboundHandlerTest
 {
-    private static final int VERSION = MessagingService.current_version;
```
adelapena, line 59: Supernit: unneeded blank line before `streamingChannel`

adelapena (approved):
Looks good to me, +1. Do we have a CI run for the last changes?

michaelsembwever:
> Do we have a CI run for the last changes?

CI: https://app.circleci.com/pipelines/github/michaelsembwever/cassandra/207 

failures known in trunk (and one timeout)

Mmuzaf:
Only the following constant is used in this class `LegacyLZ4Constants#HEADER_LENGTH`, should we move it to the `OutboundConnections`? It's worth noting that since we've removed the `FrameEncoderLegacyLZ4` and `FrameDecoderLegacyLZ4` classes, it seems we can inline this constant as it is no longer required.

Mmuzaf:
The `FrameDecoder` class the following links in java docs are broken:

```
 * 4. {@link FrameDecoderLegacy}:
          no compression; no integrity protection; turns unframed streams of legacy messages (< 4.0) into frames
 * 5. {@link FrameDecoderLegacyLZ4}
 *        LZ4 compression using standard LZ4 frame format; groups legacy messages (< 4.0) into frames
```

adelapena:
Sorry to come with this so late, but maybe we can get rid of `RequestCallbacks.CallbackInfo#responseVerb` and `Message#getPayloadSerializer(Verb, long, InetAddressAndPort)`. The JavaDoc for the latter seems to indicate that we don't need it anymore, unless I'm missing something.

Mmuzaf:
Should we remove the following configuration value as well? (maybe it is better to fix it in another issue).
Configuration: `EncryptionOptions.ServerEncryptionOptions#legacy_ssl_storage_port_enabled`


I know it is not marked as `@Deprecated`, but it seems the value released within the original commit here https://github.com/apache/cassandra/commit/310a48efcfdc0fb77fbd5fd6e734bf17667f617d and the ticket [CASSANDRA-15066](https://issues.apache.org/jira/browse/CASSANDRA-15066). The configuration value was previously called: `enable_legacy_ssl_storage_port` (here is the link to it https://github.com/apache/cassandra/commit/310a48efcfdc0fb77fbd5fd6e734bf17667f617d#diff-8c05296e94c174af68ee14e9c4a84982338aabf6d926a78608adf0f81775418bR240).
 
According to the following error message in `InboundConnectionSettings` class the SSL port was used in 3.0, right?
https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/net/InboundConnectionSettings.java#L161

michaelsembwever:
> Configuration: EncryptionOptions.ServerEncryptionOptions#legacy_ssl_storage_port_enabled

this is related to something else than the messaging service. a separate ticket under the epic CASSANDRA-18306 is warranted.

src/java/org/apache/cassandra/net/Message.java
```diff
@@ -729,20 +728,27 @@ public <T> Message<T> deserialize(DataInputPlus in, InetAddressAndPort peer, int
...
+            size += serializedHeader(message.header, version);
+            int payloadSize = message.payloadSize(version);
+            size += sizeofUnsignedVInt(payloadSize) + payloadSize;
+            return Ints.checkedCast(size);
         }
 
         /**
          * Size of the next message in the stream. Returns -1 if there aren't sufficient bytes read yet to determine size.
          */
         int inferMessageSize(ByteBuffer buf, int index, int limit, int version) throws InvalidLegacyProtocolMagic
```
Mmuzaf: `version` and `InvalidLegacyProtocolMagic` are not used and probably can be removed from the method signature.

Mmuzaf:
I have found a few if-else statements `if (version >= MessagingService.VERSION_40)` where the _else_ part can be safely removed as this condition is always true from now on, right? 

The production code classes:

- https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/db/ReadResponse.java#L343
- https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/locator/InetAddressAndPort.java#L356
- https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/locator/InetAddressAndPort.java#L431
- https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/locator/InetAddressAndPort.java#L459
- https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/locator/InetAddressAndPort.java#L475
- https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/locator/InetAddressAndPort.java#L506
- https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/schema/CompressionParams.java#L618

Mmuzaf:
The `ForwardingInfo` class:
- all of the if-else statements `version >= VERSION_40` can be simplified, right?
- according to the TODO statement ['in the next protocol version only serialize peers, message id will become redundant once 3.0 is out of the picture'](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/net/ForwardingInfo.java#L42) we can get rid of `messageIds` field or am I missing something?

michaelsembwever:
all comments addressed.

src/java/org/apache/cassandra/net/Message.java
michaelsembwever: done

Mmuzaf:
Sorry, the last one :-)

The `RequestFailureReason` class - the condition `version < VERSION_40` can be simplified as well.

Mmuzaf:
`DropCompactStorage#validateCanDropCompactStorage`

I think this can be removed as it is always `false`.

```java
                if (MessagingService.instance().versions.knows(node) &&
                    MessagingService.instance().versions.getRaw(node) < MessagingService.VERSION_40)
                {
                    before4.add(node);
                    continue;
                }
```

Mmuzaf:
`ReadResponseTest#roundTripSerialization`

Can remove the following:

```java
            if (version < MessagingService.VERSION_40)
            {
                assertFalse(deser.mayIncludeRepairedDigest());
                // even though that means they should never be used, verify that the default values are present
                assertEquals(ByteBufferUtil.EMPTY_BYTE_BUFFER, deser.repairedDataDigest());
                assertTrue(deser.isRepairedDigestConclusive());
            }
```

Mmuzaf:
For the `Verifier` class we can simplify the methods:
- willProcessOnEventLoop
- expiresAtNanos

Mmuzaf:
The `MessageGenerator` class the methods can be simplified as well:
- writeLength
- readHeader

michaelsembwever:
> DropCompactStorage#validateCanDropCompactStorage

this is its own kettle of fish, separate ticket please.

michaelsembwever:
comments addressed.

Mmuzaf (approved):

Mmuzaf:
@michaelsembwever Thank you, re-checked the comments and sources - OK 
The build with all checks passes locally.

test/unit/org/apache/cassandra/locator/InetAddressAndPortSerializerTest.java
```diff
@@ -37,8 +37,10 @@ public void testRoundtrip() throws Exception
         InetAddressAndPort ipv4 = InetAddressAndPort.getByName("127.0.0.1:42");
         InetAddressAndPort ipv6 = InetAddressAndPort.getByName("[2001:db8:0:0:0:ff00:42:8329]:42");
 
-        testAddress(ipv4, MessagingService.VERSION_30);
-        testAddress(ipv6, MessagingService.VERSION_30);
+        testAddress(ipv4, MessagingService.VERSION_40);
+        testAddress(ipv6, MessagingService.VERSION_40);
+        testAddress(ipv4, MessagingService.minimum_version);
+        testAddress(ipv6, MessagingService.minimum_version);
```
adelapena: This will test `VERSION_40` three times, since it's the value of `MessagingService.minimum_version` and `MessagingService.current_version`. However, `VERSION_50` won't be tested. I think we could modify the test to simply use `VERSION_40` and `VERSION_50`, and maybe make sure that `minimum_version` and `current_version` are among them.

test/unit/org/apache/cassandra/net/ForwardingInfoTest.java
```diff
@@ -41,12 +41,6 @@ public void testCurrent() throws Exception
         testVersion(MessagingService.current_version);
     }
 
-    @Test
-    public void test30() throws Exception
```
adelapena, line 45: Now `ForwardingInfoTest#testVersion(int)` is always called with the same value. I think it can be either inlined into `ForwardingInfoTest#testCurrent()` or annotated with `@SuppressWarnings("SameParameterValue")`.

test/unit/org/apache/cassandra/db/ReadResponseTest.java
```diff
@@ -162,12 +162,8 @@ public void makeDigestDoesntConsiderRepairedDataInfo()
     }
 
     private void verifySerDe(ReadResponse response) {
-        // check that roundtripping through ReadResponse.serializer behaves as expected.
-        // ReadResponses from pre-4.0 nodes will never contain repaired data digest
-        // or pending session info, but we run all messages through both pre/post 4.0
-        // serde to check that the defaults are correctly applied
+        // check that roundtripping through ReadResponse.serializer behaves as expected
         roundTripSerialization(response, MessagingService.current_version);
```
adelapena: `roundTripSerialization` now always takes the same version argument. I'd either inline it or annotate the method with `@SuppressWarnings("SameParameterValue")`.

test/burn/org/apache/cassandra/net/MessageGenerator.java
```diff
@@ -159,15 +158,13 @@ static Header readHeader(DataInputPlus in, int messagingVersion) throws IOExcept
...
-        else
-            out.writeUnsignedVInt32(payload.length);
+        assert messagingVersion >= VERSION_40;
+        out.writeUnsignedVInt32(payload.length);
     }
 
     static long serializedSize(byte[] payload, int messagingVersion)
     {
-        return payload.length + (messagingVersion < VERSION_40 ? 4 : VIntCoding.computeUnsignedVIntSize(payload.length));
+        return payload.length + VIntCoding.computeUnsignedVIntSize(payload.length);
```
adelapena, line 167: `serializedSize` doesn't use the `messagingVersion` parameter anymore, we can remove it.

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -605,6 +605,7 @@ static class Serializer implements IVersionedSerializer<CompressionParams>
     {
         public void serialize(CompressionParams parameters, DataOutputPlus out, int version) throws IOException
         {
+            assert version >= MessagingService.VERSION_40;
```
adelapena, line 607: It's not related to the changes but, since we are here, we can remove the unneeded class-level `@SuppressWarnings("deprecation")` annotation. That existed due to the deprecated properties `SSTABLE_COMPRESSION`, `CHUNK_LENGTH_KB` and `CRC_CHECK_CHANCE`, which aren't here anymore. Feel free to ignore.

src/java/org/apache/cassandra/locator/InetAddressAndPort.java
```diff
@@ -347,17 +345,10 @@ public void serialize(InetSocketAddress endpoint, DataOutputPlus out, int versio
...
-                out.writeByte(address.length + 2);
-                out.write(address);
-                out.writeShort(port);
-            }
-            else
-            {
-                out.writeByte(address.length);
-                out.write(address);
-            }
+            assert version >= MessagingService.VERSION_40;
```
adelapena, line 347: Totally optional since it's unrelated to the changes, but we can remove the unneeded `@SuppressWarnings("UnstableApiUsage")` class-level annotation.

test/unit/org/apache/cassandra/locator/InetAddressAndPortSerializerTest.java
michaelsembwever: i can see value in testing both explicit values and the minimum and current defined.

cost is irrelevant here.

so… i'd rather add VERSION_50 and not remove any. wdyt?

michaelsembwever: actually… i should loop through supportedVersions()

test/unit/org/apache/cassandra/net/ForwardingInfoTest.java
michaelsembwever: going to loop through supportedVersions()

test/unit/org/apache/cassandra/db/ReadResponseTest.java
michaelsembwever: going to loop through supportedVersions()

michaelsembwever:
@adelapena comments addressed

test/unit/org/apache/cassandra/locator/InetAddressAndPortSerializerTest.java
adelapena: Yes, `supportedVersions()` is perfect for this.

adelapena (approved):
Last changes look good to me, +1. Do we have a CI run for them?

michaelsembwever:
CI
 - https://app.circleci.com/pipelines/github/michaelsembwever/cassandra/214


SSL broken tests are from jdk17.
upgrade timestamp tests are from the deletion serialisation merge.
SSTablesSystemViewTest is from the SAI merge.

michaelsembwever:
latest CI
- https://app.circleci.com/pipelines/github/michaelsembwever/cassandra/215

all failures are not related.

adelapena:
CI results look good to me. There are failures for `SSTablesSystemViewTest#testVirtualTableThroughIndexLifeCycle` with tries that aren't reported, but they can be reproduced on trunk:

* https://app.circleci.com/pipelines/github/adelapena/cassandra/3068/workflows/17760f03-9807-41cc-9cb3-a1fdf3e60f77/jobs/64069/tests

I have created CASSANDRA-18703 for that test failure.

As for the j17 workflow, I've started a few tests here: https://app.circleci.com/pipelines/github/adelapena/cassandra/3069/workflows/613644ed-b9d6-4d81-9ef3-ee9d13625fd5

michaelsembwever:
Committed https://github.com/apache/cassandra/commit/4bbfd64fcd7ff208c8cc3833ffff00aea015d2a8

src/java/org/apache/cassandra/hints/HintsDescriptor.java
michaelsembwever: turned out, HintsDescriptor is part of deserialisation, and we needed to keep this to maintain readability of old hint files.

see https://issues.apache.org/jira/browse/CASSANDRA-18314?focusedCommentId=17783522&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17783522

PR #2534, opened 2023-07-28
CASSANDRA-18580: Accord metrics

jacek-lewandowski:


src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+
+    /**
+     * The number of timed out transactions on this coordinator.
+     */
+    public final Meter coordinatorTimedoutTrx;
+
+    /**
+     * The time between the start of the transaction and the start of the recovery, if the transaction is recovered.
+     */
+    public final Timer coordinatorRecoveryDelay;
```
henrikingo: Note to self: How can we know the start time of the trx, if we are in recovery? By definition in recovery we have lost the original coordinator, which would be the only node to know this information?

Note that t0 is not the start of the transaction, it is the earliest possible commit time / trx id.

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+
+    /**
+     * The time between the start of the recovery and the execution of the transaction, if the transaction is recovered.
+     */
+    public final Timer coordinatorRecoveryTime;
```
henrikingo: The two counters above seem a bit arbitrary in their naming.
Oh, okay, now that I write this I even understand what they mean...

Maybe with the last one you rather mean `coordinatorRecoveryDuration`? In other words Time = point in time, duration = interval.

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+
+    /**
+     * The number of slow path transactions executed on this coordinator.
+     */
+    public final Meter coordinatorSlowPathTrx;
+
+    /**
+     * The number of preempted transactions on this coordinator.
+     */
+    public final Meter coordinatorPreemptedTrx;
```
henrikingo: This name threw me off a bit... "pre-empted" is not a word used in the Accord paper. Is it used in the implementation? I assume this means number of conflicted transactions... Why is this metric not the same as coordinatorSlowPathTrx?

henrikingo (approved):
As usual, I guess I mostly comment on naming variables... (sorry)
But it was an educational read for me. Thanks for inviting.

src/java/org/apache/cassandra/metrics/AccordMetrics.java
jacek-lewandowski: as far as I understand, preemption pertains to the recovery

test/distributed/org/apache/cassandra/distributed/test/accord/AccordMetricsTest.java
```diff
@@ -0,0 +1,258 @@
...
+    {
+        IMessageFilters.Filter commitFilter = SHARED_CLUSTER.filters().outbound().verbs(Verb.ACCORD_COMMIT_REQ.id).from(1).to(2).drop();
+        commitFilter.on();
+
+        countingMetrics0 = getMetrics();
+        try
+        {
+            SHARED_CLUSTER.coordinator(1).executeWithResult(writeCql(), ConsistencyLevel.ALL, 0, 0, 0, 0);
+        }
+        catch (RuntimeException t)
```
maedhroz: nit: I wonder if it would make more sense to catch `CassandraException` in places like this, so things like NPE just fail without even worrying about the following assertion...

EDIT: I realized right after postings this that you can't do that because of classpath issues. Never mind :D

test/distributed/org/apache/cassandra/distributed/test/accord/AccordMetricsTest.java
```diff
@@ -0,0 +1,258 @@
...
+        {
+            assertThat(t.getCause().getClass().getName()).isEqualTo(WritePreemptedException.class.getName());
```
maedhroz: This fails w/ an NPE when I run locally, and I think it's because `t` itself is a `WPE`...so `getCause()` will just return `null`.

In any case, change this to...

```
assertThat(t.getClass().getName()).isEqualTo(WritePreemptedException.class.getName());
```

...and everything is green for me.

test/distributed/org/apache/cassandra/distributed/test/accord/AccordMetricsTest.java
```diff
@@ -0,0 +1,258 @@
...
+    private void assertReplicaMetrics(int node, String scope, long commits, long executions, long applications)
+    {
+        DefaultNameFactory nameFactory = new DefaultNameFactory("accord-replica", scope);
+        Map<String, Long> metrics = diff(countingMetrics0).get(node);
+        Function<String, Long> metric = n -> metrics.get(nameFactory.createMetricName(n).getMetricName());
+        assertThat(metric.apply("CommitLatency")).isEqualTo(commits);
+        assertThat(metric.apply("ExecuteLatency")).isEqualTo(executions);
+        assertThat(metric.apply("ApplyLatency")).isEqualTo(applications);
+        assertThat(metric.apply("ApplyDuration")).isEqualTo(applications);
+        assertThat(metric.apply("Dependencies")).isEqualTo(executions);
```
maedhroz: nit: For all of these metric names, we could make constants out of them in `AccordMetrics` to keep things DRY.

test/unit/org/apache/cassandra/service/accord/AccordStateCacheTest.java
```diff
@@ -162,13 +185,15 @@ public void testAcquisitionAndRelease()
 
         Assert.assertSame(safeString1.global, cache.head());
         Assert.assertSame(safeString2.global, cache.tail());
+
+        assertCacheMetrics(cache.metrics, 0, 2, 2);
```
maedhroz, line 213: The goal is just to test the global metrics here, correct?

src/java/org/apache/cassandra/service/accord/AccordStateCache.java
```diff
@@ -66,22 +60,27 @@ static class Stats
...
+    final AccordStateCacheMetrics metrics;
 
-    public AccordStateCache(ExecutorPlus loadExecutor, ExecutorPlus saveExecutor, long maxSizeInBytes)
+    public AccordStateCache(ExecutorPlus loadExecutor, ExecutorPlus saveExecutor, long maxSizeInBytes, String metricsScope)
     {
         this.loadExecutor = loadExecutor;
         this.saveExecutor = saveExecutor;
         this.maxSizeInBytes = maxSizeInBytes;
+        this.metricsScope = metricsScope;
+        this.metrics = new AccordStateCacheMetrics(String.format("%s-global", metricsScope), this);
```
maedhroz: Assuming we only need global stats, or global stats and stats broken down by instance, would it be easier to just create 3 singleton instances of `AccordStateCacheMetrics` and reuse them across all command stores? I'm not sure whether this actually ends up constituting any correctness problems w/ multiple stores active (although I don't think we have any tests that cover this yet, as `AccordStateCacheTest` is more of a unit test), but it does seem like we would be doing a lot of duplicate MBean registration (or at least needless attempts at re-registration) in `registerMBean()`...

test/distributed/org/apache/cassandra/distributed/test/accord/AccordMetricsTest.java
```diff
@@ -0,0 +1,258 @@
...
+        Map<String, Long> metrics = diff(countingMetrics0).get(node);
+        logger.info("Metrics for node {} / {}: {}", node, scope, metrics);
+        Function<String, Long> metric = n -> metrics.get(nameFactory.createMetricName(n).getMetricName());
+        assertThat(metric.apply("FastPathTrx")).isEqualTo(fastPathTrx);
+        assertThat(metric.apply("SlowPathTrx")).isEqualTo(slowPathTrx);
+        assertThat(metric.apply("PreemptedTrx")).isEqualTo(preemptions);
+        assertThat(metric.apply("TimedoutTrx")).isEqualTo(timeouts);
+        assertThat(metric.apply("RecoveryDelay")).isEqualTo(recoveries);
+        assertThat(metric.apply("RecoveryTime")).isEqualTo(recoveries);
+        assertThat(metric.apply("Dependencies")).isEqualTo(fastPathTrx + slowPathTrx);
```
maedhroz: nit: Should we have perhaps a smoke test to ensure the `FastPathToTotalX` bits are published?

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+     * Duration of applying changes.
+     */
+    public final Timer applyDuration;
+
+    /**
+     * The number of dependencies of the transaction on this replica.
+     */
+    public final Histogram deps;
+
+    public final Meter progressLogSize;
```
maedhroz, line 88: Should this be a `Meter`, or is it supposed to be the current number of items in the progress log (a `Counter`)? ...or the rate at which items are added to the progress log?

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+
+    /**
+     * Duration of applying changes.
+     */
+    public final Timer applyDuration;
+
+    /**
+     * The number of dependencies of the transaction on this replica.
+     */
+    public final Histogram deps;
```
maedhroz: Alternate naming -> `partialDependencies`

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+
+    public final Meter progressLogSize;
+
+    /**
+     * The number of dependencies of the transaction.
+     */
+    public final Histogram coordinatorDeps;
```
maedhroz: Alternate naming -> just `dependencies`, although this depends (heh) on the renaming above for `deps` -> `partialDependencies`

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+     */
+    public final Timer applyLatency;
+
+    /**
+     * Duration of applying changes.
+     */
+    public final Timer applyDuration;
+
+    /**
+     * The number of dependencies of the transaction on this replica.
```
maedhroz:
```suggestion
     * A histogram of the number of dependencies per partial transaction at this replica.
```

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+     */
+    public final Histogram deps;
+
+    public final Meter progressLogSize;
+
+    /**
+     * The number of dependencies of the transaction.
```
maedhroz:
```suggestion
     * A histogram of the number of dependencies per transaction at this coordinator.
```

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+                }
+            }
+        }
+        catch (IllegalAccessException e)
+        {
+            throw new RuntimeException(e);
+        }
+        builder.append("]");
+        return builder.toString();
+    }
```
maedhroz, line 181: Are there any logging bits where we use this `toString()`, or is it mostly here for debugging?

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+        applyDuration = Metrics.timer(replica.createMetricName("ApplyDuration"));
+        deps = Metrics.histogram(replica.createMetricName("Dependencies"), true);
+        progressLogSize = Metrics.meter(replica.createMetricName("ProgressLogSize"));
+
+        DefaultNameFactory coordinator = new DefaultNameFactory("accord-coordinator", scope);
+        coordinatorDeps = Metrics.histogram(coordinator.createMetricName("Dependencies"), true);
+        coordinatorFastPathTrx = Metrics.meter(coordinator.createMetricName("FastPathTrx"));
+        coordinatorSlowPathTrx = Metrics.meter(coordinator.createMetricName("SlowPathTrx"));
+        coordinatorPreemptedTrx = Metrics.meter(coordinator.createMetricName("PreemptedTrx"));
+        coordinatorTimedoutTrx = Metrics.meter(coordinator.createMetricName("TimedoutTrx"));
```
maedhroz:
```suggestion
        coordinatorTimedoutTrx = Metrics.meter(coordinator.createMetricName("Timeouts"));
```
This is what you see in places like `ClientRequestMetrics`

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+        applyLatency = Metrics.timer(replica.createMetricName("ApplyLatency"));
+        applyDuration = Metrics.timer(replica.createMetricName("ApplyDuration"));
+        deps = Metrics.histogram(replica.createMetricName("Dependencies"), true);
+        progressLogSize = Metrics.meter(replica.createMetricName("ProgressLogSize"));
+
+        DefaultNameFactory coordinator = new DefaultNameFactory("accord-coordinator", scope);
+        coordinatorDeps = Metrics.histogram(coordinator.createMetricName("Dependencies"), true);
+        coordinatorFastPathTrx = Metrics.meter(coordinator.createMetricName("FastPathTrx"));
+        coordinatorSlowPathTrx = Metrics.meter(coordinator.createMetricName("SlowPathTrx"));
+        coordinatorPreemptedTrx = Metrics.meter(coordinator.createMetricName("PreemptedTrx"));
```
maedhroz:
```suggestion
        coordinatorPreemptedTrx = Metrics.meter(coordinator.createMetricName("Preempts"));
```

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+        executeLatency = Metrics.timer(replica.createMetricName("ExecuteLatency"));
+        applyLatency = Metrics.timer(replica.createMetricName("ApplyLatency"));
+        applyDuration = Metrics.timer(replica.createMetricName("ApplyDuration"));
+        deps = Metrics.histogram(replica.createMetricName("Dependencies"), true);
+        progressLogSize = Metrics.meter(replica.createMetricName("ProgressLogSize"));
+
+        DefaultNameFactory coordinator = new DefaultNameFactory("accord-coordinator", scope);
+        coordinatorDeps = Metrics.histogram(coordinator.createMetricName("Dependencies"), true);
+        coordinatorFastPathTrx = Metrics.meter(coordinator.createMetricName("FastPathTrx"));
+        coordinatorSlowPathTrx = Metrics.meter(coordinator.createMetricName("SlowPathTrx"));
```
maedhroz:
```suggestion
        coordinatorSlowPathTrx = Metrics.meter(coordinator.createMetricName("SlowPath"));
```

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+        commitLatency = Metrics.timer(replica.createMetricName("CommitLatency"));
+        executeLatency = Metrics.timer(replica.createMetricName("ExecuteLatency"));
+        applyLatency = Metrics.timer(replica.createMetricName("ApplyLatency"));
+        applyDuration = Metrics.timer(replica.createMetricName("ApplyDuration"));
+        deps = Metrics.histogram(replica.createMetricName("Dependencies"), true);
+        progressLogSize = Metrics.meter(replica.createMetricName("ProgressLogSize"));
+
+        DefaultNameFactory coordinator = new DefaultNameFactory("accord-coordinator", scope);
+        coordinatorDeps = Metrics.histogram(coordinator.createMetricName("Dependencies"), true);
+        coordinatorFastPathTrx = Metrics.meter(coordinator.createMetricName("FastPathTrx"));
```
maedhroz:
```suggestion
        coordinatorFastPathTrx = Metrics.meter(coordinator.createMetricName("FastPath"));
```

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+     */
+    public final Histogram deps;
+
+    public final Meter progressLogSize;
+
+    /**
+     * The number of dependencies of the transaction.
+     */
+    public final Histogram coordinatorDeps;
+
```
maedhroz, line 94: We have a handful of metrics in the following section, and there are a couple general things:

1.) I'm not sure we need the `Trx` suffixes. They don't seem to add much information?

2.) The concepts are all coordinator-level concepts, AFAICT, so the `coordinator` prefix doesn't seem necessary? (i.e. There's no concept of a "replica slow path"...)

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+    /**
+     * The number of preempted transactions on this coordinator.
+     */
+    public final Meter coordinatorPreemptedTrx;
+
+    /**
+     * The number of timed out transactions on this coordinator.
+     */
+    public final Meter coordinatorTimedoutTrx;
```
maedhroz:
```suggestion
    public final Meter timeouts;
```
Similar to what we have in `ClientRequestMetrics`?

src/java/org/apache/cassandra/metrics/AccordMetrics.java
maedhroz:
```suggestion
    public final Meter preempts;
```

src/java/org/apache/cassandra/metrics/AccordMetrics.java
maedhroz: Wait, after looking at this again, what is the difference between the timeout and preempted metrics here and the ones in `AccordClientRequestMetrics` recorded in `AccordService` and split into read/write?

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+    public final RatioGaugeSet fastPathToTotal;
+
+    private AccordMetrics(String scope)
+    {
+        DefaultNameFactory replica = new DefaultNameFactory("accord-replica", scope);
+        commitLatency = Metrics.timer(replica.createMetricName("CommitLatency"));
+        executeLatency = Metrics.timer(replica.createMetricName("ExecuteLatency"));
+        applyLatency = Metrics.timer(replica.createMetricName("ApplyLatency"));
+        applyDuration = Metrics.timer(replica.createMetricName("ApplyDuration"));
+        deps = Metrics.histogram(replica.createMetricName("Dependencies"), true);
```
maedhroz:
```suggestion
        deps = Metrics.histogram(replica.createMetricName("PartialDependencies"), true);
```
Maybe the replica scope makes the "Partial" but unnecessary...but figured I'd throw it out there...

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+
+    /**
+     * The number of fast path transactions executed on this coordinator.
+     */
+    public final Meter coordinatorFastPathTrx;
+
+    /**
+     * The number of slow path transactions executed on this coordinator.
+     */
+    public final Meter coordinatorSlowPathTrx;
```
maedhroz: Trying to think of alternate naming here...

`fastPath` / `slowPath`
`fastPaths` / `slowPaths`
`fastPathsTaken` / `slowPathsTaken`

src/java/org/apache/cassandra/metrics/RatioGaugeSet.java
```diff
@@ -0,0 +1,119 @@
...
+
+import java.util.function.DoubleSupplier;
+
+import com.codahale.metrics.Metered;
+import com.codahale.metrics.RatioGauge;
+import org.apache.cassandra.metrics.CassandraMetricsRegistry.MetricName;
+
+import static org.apache.cassandra.metrics.CassandraMetricsRegistry.Metrics;
+
+public class RatioGaugeSet
```
maedhroz, line 29:
```suggestion
public class MeteredRatioGauge
```
I guess it doesn't explicitly implement `Metered`, but it's close :)

src/java/org/apache/cassandra/metrics/RatioGaugeSet.java
```diff
@@ -0,0 +1,119 @@
...
+    public final RatioGauge fiveMinuteRatio;
+    public final RatioGauge fifteenMinuteRatio;
+    public final RatioGauge allTimeRatio;
+
+    public RatioGaugeSet(Metered numerator, Metered denominator, MetricNameFactory factory, String namePattern)
+    {
+        this.oneMinuteRatio = ratioGauge(factory.createMetricName(String.format(namePattern, "OneMinuteRatio")), numerator::getOneMinuteRate, denominator::getOneMinuteRate);
+        this.fiveMinuteRatio = ratioGauge(factory.createMetricName(String.format(namePattern, "FiveMinuteRatio")), numerator::getFiveMinuteRate, denominator::getFiveMinuteRate);
+        this.fifteenMinuteRatio = ratioGauge(factory.createMetricName(String.format(namePattern, "FifteenMinuteRatio")), numerator::getFifteenMinuteRate, denominator::getFifteenMinuteRate);
+        this.allTimeRatio = ratioGauge(factory.createMetricName(String.format(namePattern, "AllTimeRatio")), numerator::getMeanRate, denominator::getMeanRate);
```
maedhroz:
```suggestion
        this.allTimeRatio = ratioGauge(factory.createMetricName(String.format(namePattern, "MeanRatio")), numerator::getMeanRate, denominator::getMeanRate);
```
...or do you think "mean" is to vague in this context?

src/java/org/apache/cassandra/metrics/AccordMetrics.java
```diff
@@ -0,0 +1,279 @@
...
+
+        DefaultNameFactory coordinator = new DefaultNameFactory("accord-coordinator", scope);
+        coordinatorDeps = Metrics.histogram(coordinator.createMetricName("Dependencies"), true);
+        coordinatorFastPathTrx = Metrics.meter(coordinator.createMetricName("FastPathTrx"));
+        coordinatorSlowPathTrx = Metrics.meter(coordinator.createMetricName("SlowPathTrx"));
+        coordinatorPreemptedTrx = Metrics.meter(coordinator.createMetricName("PreemptedTrx"));
+        coordinatorTimedoutTrx = Metrics.meter(coordinator.createMetricName("TimedoutTrx"));
+        coordinatorRecoveryDelay = Metrics.timer(coordinator.createMetricName("RecoveryDelay"));
+        coordinatorRecoveryDuration = Metrics.timer(coordinator.createMetricName("RecoveryTime"));
+        fastPathToTotal = new RatioGaugeSet(coordinatorFastPathTrx, RatioGaugeSet.sum(coordinatorFastPathTrx, coordinatorSlowPathTrx), coordinator, "FastPathToTotal%s");
```
maedhroz:
```suggestion
        fastPathToTotal = new RatioGaugeSet(coordinatorFastPathTrx, RatioGaugeSet.sum(coordinatorFastPathTrx, coordinatorSlowPathTrx), coordinator, "FastPathToTotal.%s");
```
Would something like `FastPathToTotal.OneMinuteRatio` be a little easier to read?

src/java/org/apache/cassandra/metrics/RatioGaugeSet.java
```diff
@@ -0,0 +1,119 @@
...
+    {
+        return Metrics.register(name, ratioGauge(numerator, denominator));
+    }
+
+    public static Metered sum(Metered... meters)
+    {
+        return new SumMeter(meters);
+    }
+
+    private static class SumMeter implements Metered
```
maedhroz:
```suggestion
    private static class SummingMeter implements Metered
```

maedhroz: Would it make sense to just move this class inside `AccordMetrics`? That would make the static factory unnecessary? I mean, it could be used outside either of these classes and be its own thing as well...

src/java/org/apache/cassandra/metrics/RatioGaugeSet.java
```diff
@@ -0,0 +1,119 @@
...
+        }
+
+        @Override
+        public long getCount()
+        {
+            long count = 0;
+            for (Metered meter : meters)
+                count += meter.getCount();
+            return count;
+        }
```
maedhroz, line 87: You could cut down on some of the repetition here w/ something like...

```
@Override
public long getCount()
{
    return getMetric(Metered::getCount);
}
...
private long getMetric(Function<Metered, Long> f)
{
    long count = 0;
    for (Metered meter : meters)
        count += f.apply(meter);
    return count;
}
```

src/java/org/apache/cassandra/metrics/AccordMetrics.java
jacek-lewandowski: It is updated with deltas so while the `xxxRate` metrics provide progress log size change rates, the `count` metric should provide its current size - so we have both, right?

src/java/org/apache/cassandra/metrics/AccordMetrics.java
jacek-lewandowski: It is mostly for debugging, would you like me to change it somehow?

test/distributed/org/apache/cassandra/distributed/test/accord/AccordMetricsTest.java
jacek-lewandowski: yup, fixed, I probably made some change when squashing and forgot to retest, thanks

test/distributed/org/apache/cassandra/distributed/test/accord/AccordMetricsTest.java
jacek-lewandowski: Do you mean just those strings or full mbean names?

src/java/org/apache/cassandra/metrics/AccordMetrics.java
jacek-lewandowski: 1) Ok
2) It is just for the ease of use - I believe that having them grouped under distinct categories makes it easier to navigate in simple JMX clients. However, I'm not adamantly opposed to changing that if you wish

src/java/org/apache/cassandra/metrics/AccordMetrics.java
jacek-lewandowski: Indeed. I was going to ask about that - `ACRM` are tied to the statement processing initiated by the client, on the coordinator hit by the client request. What about the events at the recovery coordinator?

src/java/org/apache/cassandra/metrics/RatioGaugeSet.java
jacek-lewandowski: A good candidate to reuse this functionality is `CacheMetrics` (I mean, the whole `RatioGaugeSet`, including `SummingMeter`) - however, that would follow a few metric renamings, which is outside of the scope of this story.

src/java/org/apache/cassandra/service/accord/AccordStateCache.java
jacek-lewandowski: Agreed

jacek-lewandowski: hmm... it can be difficult because the `CacheMetrics` requires a reference to the `CacheSize` instance, a cache provider implemented by the cache itself. Although it seems justified to have a single metric for hits / misses / requests, regardless the number of cache instances, it does not apply to the cache size gauges, which we would have to sum up manually. If I understand that code correctly, we create only one cache instance per class per node

test/unit/org/apache/cassandra/service/accord/AccordStateCacheTest.java
jacek-lewandowski: yes, I've just added also assertions against instance metrics

test/distributed/org/apache/cassandra/distributed/test/accord/AccordMetricsTest.java
maedhroz: Just the string literals here...but looks like you've already handled this.

src/java/org/apache/cassandra/metrics/AccordMetrics.java
maedhroz: As long as it doesn't end up being called on a hot path somewhere, it's probably fine. The alternate would, I guess, be adding the `Counting` instances to a collection and using that in `toString()` (sort of like how we track all metrics in `TableMetrics`), but using reflection isn't necessarily worse for this purpose.

src/java/org/apache/cassandra/metrics/AccordMetrics.java
maedhroz: I like the naming scheme overall after the last couple commits, FWIW

src/java/org/apache/cassandra/metrics/AccordMetrics.java
maedhroz: Yeah, it seems like the overlap is only partial? Trying to remember the details of the protocol, but the hooks we've put in `CoordinatePreAccept` would only apply to transactions initiated by the client, and would be the double-counted items (well, counted in two places), while the hooks in `Recover` would register in `AccordMetrics` but not an `AccordClientRequestMetrics` instance.

@dcapwell Do you have any preference in terms of how we evolve this?

We could leave `AccordClientRequestMetrics` intact, remove any bits from `AccordMetrics` that duplicate those, and revert to `AccordMetrics` (or some other specific metrics class) only tracking timeouts/preempts/etc. related to recovery. We could also leave things as they are, but then we don't have any idea what the breakdown is in `AccordMetrics` between recovery and client requests.

src/java/org/apache/cassandra/metrics/AccordMetrics.java
maedhroz: Sorry. You're right. It is both. (I wasn't expecting us to want to rate items were being added, but maybe someone will want to know...)

src/java/org/apache/cassandra/service/accord/AccordStateCache.java
maedhroz: `CacheSize` is just an interface though. We can still have our metrics collected in one object managed by `AccordCommandStores`, which can then delegate to the stores it manages to get an overall cache size. Check out `refreshCacheSizes()` where it interacts w/ the cache sizes already, albeit to set their size, not get it. Does that make more sense?

src/java/org/apache/cassandra/metrics/AccordMetrics.java
dcapwell: > What about the events at the recovery coordinator?

This gets tricky as every node will try to recover from time to time, and it won't know if it actually did anything in some cases[1].  We also get into the situation where multi key txn runs and could double count as we see the events in multiple `CommandStore`, so not sure how you would handle dedup in these cases (or do we just double count?).  I would need to look at this patch, but even having metrics in `Recover` is tricky as our callback doesn't mean we recovered, only that we scheduled things[1]...  The `TransactionStatement` is simple as its logically contained, but that isn't true with recover, so we would need to be careful...

Now, I guess my main question is "how would an operator use these metrics"?  If the answer is "I don't know" then prob best to punt or try to flesh that out.  


[1] `accord.coordinate.Recover#recover()`

```
Deps committedDeps = tryMergeCommittedDeps();
                    node.withEpoch(executeAt.epoch(), () -> {
                        // TODO (required, consider): when writes/result are partially replicated, need to confirm we have quorum of these
                        if (committedDeps != null) Persist.persistMaximal(node, txnId, route, txn, executeAt, committedDeps, acceptOrCommit.writes, acceptOrCommit.result);
                        else CollectDeps.withDeps(node, txnId, route, txn.keys(), executeAt, (deps, fail) -> {
                            if (fail != null) accept(null, fail);
                            else Persist.persistMaximal(node, txnId, route, txn, executeAt, deps, acceptOrCommit.writes, acceptOrCommit.result);
                        });
                    });
                    accept(acceptOrCommit.result, null);
```

Here we trigger the callback saying success, but later on it might actually fail, which will then require another repair to try to push things along...

maedhroz: Perhaps, for now, we should just leave `AccordClientRequestMetrics` as it is and remove the overlapping bits in `AccordMetrics`?

src/java/org/apache/cassandra/service/accord/AccordStateCache.java
jacek-lewandowski: I think I got what you mean;

src/java/org/apache/cassandra/metrics/AccordMetrics.java
jacek-lewandowski: Is the statement here equally valid for timeouts, preempts and invalidations?

maedhroz: I think so. We can always add more metrics around recovery later (especially if they're not going to be well-understood).

src/java/org/apache/cassandra/cache/CacheSize.java
```diff
@@ -31,4 +31,42 @@ public interface CacheSize
...
+    class Summing implements CacheSize
+    {
+        private final Iterable<CacheSize> cacheSizes;
+
+        public Summing(Iterable<CacheSize> cacheSizes)
+        {
+            this.cacheSizes = cacheSizes;
+        }
+
+        public long capacity()
```
maedhroz: nit: Throw `@Override` annotations on these 4 methods?

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
```diff
@@ -0,0 +1,90 @@
...
+        {
+            this(ACCORD_STATE_CACHE);
+        }
+
+        @VisibleForTesting
+        public Provider(String type)
+        {
+            this.type = type;
+            globalCacheSizes = new CopyOnWriteArraySet<>();
+            globalMetrics = new AccordStateCacheMetrics(type, new CacheSize.Summing(globalCacheSizes));
```
maedhroz: I'm a little bit confused about how we're managing the global metrics here. `AccordCommandStores` implements `CacheSize` now, but then we also have this `Summing` component for `globalMetrics` that has to receive and bookkeep the `globalCacheSizes`. Can't we just do something like pass `AccordCommandStores` to the provider on construction, where it can then create a `AccordStateCacheMetrics(type, stores)` (i.e. `stores` is the `AccordCommandStores`) for the globals? That at least gets rid of the `globalCacheSizes` set. (Then `forGlobal()` is simpler and no longer has to be `synchronized` as well.)

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
```diff
@@ -0,0 +1,90 @@
...
+            instanceMetrics = new HashMap<>(2);
+            instanceCacheSizes = new HashMap<>(2);
+        }
+
+        public synchronized AccordStateCacheMetrics forInstance(Class<?> klass, CacheSize commandStoreInstanceCacheSize)
+        {
+            Set<CacheSize> cacheSizes = instanceCacheSizes.computeIfAbsent(klass, k -> new CopyOnWriteArraySet<>());
+            cacheSizes.add(commandStoreInstanceCacheSize);
+            return instanceMetrics.computeIfAbsent(klass, k -> new AccordStateCacheMetrics(String.format("%s-%s", type, k.getSimpleName()), new CacheSize.Summing(cacheSizes)));
```
maedhroz: Looking at how the caches work at the instance level, there may be a problem w/ the summing approach here.

```
@Override
public long capacity()
{
    return AccordStateCache.this.capacity();
}

@Override
public void setCapacity(long capacity)
{
    AccordStateCache.this.setCapacity(capacity);
}

@Override
public int size()
{
    return AccordStateCache.this.size();
}

@Override
public long weightedSize()
{
    return bytesCached;
}
```
If we take the sum of all the `Instance` `size()` methods, won't we be taking the same shared size from `cache` in the top level `AccordStateCache`? (i.e. We're still actually looking at a global size?)

At this point, I'm not sure we should have `Instance` implement `CacheSize` at all, and I'm not sure having instance-level metrics is worth the complexity. The `Instance` caches share the same physical cache, and are less caches than they are cache accessors...

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
```diff
@@ -0,0 +1,90 @@
...
+import com.google.common.annotations.VisibleForTesting;
+
+import com.codahale.metrics.Histogram;
+import org.apache.cassandra.cache.CacheSize;
+
+import static org.apache.cassandra.metrics.CassandraMetricsRegistry.Metrics;
+
+public class AccordStateCacheMetrics extends CacheMetrics
+{
+    public final Histogram objectSize;
```
maedhroz, line 33: For the global case, is this `objectSize` metric different than the exposed `weightedSize()`? I overlooked this in my earlier reading...

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
maedhroz: Or perhaps here's another approach...

One of our problems here is that the instances caches are just views on an actual physical cache. `CacheMetrics` is really only appropriate for the global case. We could split out the bits of `CacheMetrics` (hits, misses, etc.) that are totally unambiguous for the instance "caches" and avoiding having to worry about any summing logic around those. (Would probably have to be a super-class of `CacheMetrics` w/o size information, like `CacheAccessMetrics`. Could have the same type name...)

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
jacek-lewandowski: I'm sorry for that, `AccordCommandStores` implementing `CacheSize` is a leftover and is redundant.

I was thinking about passing the list of command stores explicitly but I chose this solution as it seemed cleaner to me - I mean, there is no mutual dependency between the `Provider` and `AccordCommandStores` and it is easier to unit test.

The synchronization is not a pain point here because it happens once per store initialization.

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
jacek-lewandowski: Indeed, it is broken. I've made a mistake because `size()` for instance cache should return `itemsCached` rather than refer to the parent. It's a shame that my tests do not cover that.

jacek-lewandowski: While I agree that `CacheMetrics` could be split into `CacheAccessMetrics` being a super-class and the metrics including the cache size, I'm wondering what about `size` and `weightedSize` which express how much data is there. How we could avoid summing that across different command stores?

jacek-lewandowski: It actually works in a way that for the `AccordCommandStore` global metrics, the provider always returns the same metrics object; a similar approach was taken for the instance cache - for the same instance type, the same metrics are returned. We do not have to worry about summing up cache access metrics (hits, misses, requests) across different command stores because they all are reported to the same metrics instances.

For `CacheSize`, I believe we need to do summing on demand because the cache size can be individually set on each `AccordCommandStore`. Also, `size` and `weightedSize` are computed internally and separately for each command store and instance. So, whenever the operator requests the current number of items or the current number of bytes caches, for, say, instance X or global, we need to go through all the command stores (implementing `CacheSize`) and sum up the outputs.

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
jacek-lewandowski: While `weightedSize` is a gauge on the number of bytes stored in the cache, `objectSize` is a histogram filled with size updates - that is, it is focused on size of the individual objects added or removed from the cache. The only redundant thing would be `objectSize.getCount()` which should be equal to `weightedSize`. 

I was actually going to ask whether it should be updated only with the sizes of objects added to the cache, ignoring what was dropped?

src/java/org/apache/cassandra/cache/CacheSize.java
jacek-lewandowski: sure

src/java/org/apache/cassandra/metrics/AccordMetrics.java
jacek-lewandowski: I hesitate - preemption is something mostly tied to the recovery - isn't it? Let me at least rename those metrics to recoveryTimeouts, ... , and report only what happens in the recover to them? At the same time, such metrics present in `AccordClientRequestMetrics`  report just client facing events?

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
jacek-lewandowski: The same thing seems to apply to slow/fast path selection as well...

jacek-lewandowski: talked to @belliottsmith about that - his point is to collect combined and recovery only stats, so yet another idea...

src/java/org/apache/cassandra/metrics/AccordMetrics.java
jacek-lewandowski: That seems to apply to slow/fast path metrics as well...

talked to @belliottsmith about that - his point is to collect combined and recovery only stats, so yet another idea...

(sorry, previously this comment was accidently added to a wrong thread)

belliottsmith: My point was only that if we are thinking of separating the stats we should make it easy to view a combined stat since I think this is likely to be the most useful single set of numbers. Separate initial and recovery coordinator stats aren’t independently as useful as a combined stat, though I can see value in wanting to separate them on occasion.

I don’t think the multi key issue that has been raised is anything to worry about - only one key can become coordinator and anyway there’s the same number of coordinators.

statistics we gather per key will be duplicated but presumably these are expected to be duplicated and aren’t coordination level stats.

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
maedhroz: Okay, that makes sense. In that case, yes, I would think that we'd only want to update the metric on add, as removal should be symmetric?

(To be honest, I have no idea what the distribution of sizes is going to be here. If sizes are all in a pretty narrow range, I'm not sure what the metric buys an operator.)

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
maedhroz: I mean, having `AccordCommandStores` implement `CacheSize` was kind of nice, in any case. Generally, it seems nice to have a situation where metric information can be provided by the classes managing the relevant subjects. Like in this case, `AccordCommandStores` exposing size/capacity/etc. with no knowledge of the fact that we populate metrics w/ that information. It also seemed desirable in the sense that we didn't need the registration element of that in `forGlobal()`. Using `Summing` and registration via the `Provider` doesn't seem incorrect as implemented here...just trying to convince myself it's the simplest/most reasonable approach...

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
maedhroz: > We do not have to worry about summing up cache access metrics (hits, misses, requests) across different command stores because they all are reported to the same metrics instances.

Right. Understand that part.

> For CacheSize, I believe we need to do summing on demand because the cache size can be individually set on each AccordCommandStore

Again, agreed. I think where we still have things to resolve are:

1.) What's the most straightforward way to expose the summed sizes (see this thread https://github.com/apache/cassandra/pull/2534/files#r1317595176)

2.) Whether we really want to worry about exposing the `CacheSize` bits for the instance views at all. Again, we're talking about the same physical cache for all instances, right? Even if we don't do that right now in this patch, we could always add it later if there was a really compelling reason. This is why I suggested something like a `CacheAccessMetrics` that could just worry about exposing the hist/misses/etc. items for the `Instance` caches, which don't have to be summed at all.

src/java/org/apache/cassandra/metrics/AccordMetrics.java
maedhroz: If the most useful single thing we can do is expose combined statistics for both initial and recovery coordinators, why not just keep `AccordMetrics`, remove `AccordClientRequestMetrics`, and proceed from there?

jacek-lewandowski: agreed

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
belliottsmith: > they all are reported to the same metrics instances.

Should we be? We can manage much cheaper metrics per-command store. Whether we do that immediately or not, we should probably consider structuring our API to support this.

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
jacek-lewandowski: This is how I understood the requested "Size of CommandsForKey when loaded" mentioned in the Jira ticket;

"should be symmetric" - eventually - yes

I'll update it to report only on addition to the cache

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
jacek-lewandowski: So do you mean individual metrics per command store instead of aggregated?

belliottsmith: Whether we actually instantiate separate metrics per command store or not (we probably don't have optimally-efficient ones today for that in C*), it would be nice to structure the API to support the obviously more efficient approach, and move Cassandra to such an approach when ready.

belliottsmith: So by that I mean probably provide the CommandStore its own metrics callback on construction; but in our C* implementation this might be the same callback.

src/java/org/apache/cassandra/metrics/AccordMetrics.java
jacek-lewandowski: @dcapwell regarding your comment, the recovery code updates the metrics in this method:

```java
    @Override
    public void accept(Result result, Throwable failure)
    {
        isDone = true;
        if (failure == null)
        {
            callback.accept(ProgressToken.APPLIED, null);
            node.agent().metricsEventsListener().onRecover(txnId, ballot);
        }
        else
        {
            callback.accept(null, failure);
            if (failure instanceof Preempted)
                node.agent().metricsEventsListener().onPreempted(txnId);
            else if (failure instanceof Timeout)
                node.agent().metricsEventsListener().onTimeout(txnId);
            else if (failure instanceof Invalidated)
                node.agent().metricsEventsListener().onInvalidated(txnId);
        }

        node.agent().onRecover(node, result, failure);
    }
```

As far as I understand, the recovery initiated by this coordinator ends in that method. I don't think we need to deal with deduplication on the coordinator side because the coordinator operates on whole transactions rather than shards. It does not use command stores at all (in fact, it uses them just as `AgentExecutor`, a separate story)

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
jacek-lewandowski: actually, I was wrong, it is already implemented to report only additions, so nothing to do here

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
jacek-lewandowski: The problem with `CacheMetrics` is that they include both "push" and "pull" items. So perhaps it is a good idea to separate the concepts of `CacheSize` and `CacheAccessMetrics`. Will think about that over the weekend;

jacek-lewandowski: Ok, so I refactored things a bit (tests do not compile yet, I need to get your opinion first). I separated cache size metrics and cache access metrics and now it looks simpler to me. `AccordCommandStores` and `AccordCommandStore` both implement `CacheSize`. `AccordCommandStores` include cache size metrics and cache access metrics. The instance of the later is passed to each new `AccordCommandStore`. So in short, we have cache size metrics only for global `AccordCommandStores` while cache access metrics are both global and per instance. I hope this is sufficient

src/java/org/apache/cassandra/metrics/CacheAccessMetrics.java
```diff
@@ -0,0 +1,73 @@
...
+    public CacheAccessMetrics(MetricNameFactory factory)
+    {
+        this.factory = factory;
+
+        this.hits = Metrics.meter(factory.createMetricName("Hits"));
+        this.misses = Metrics.meter(factory.createMetricName("Misses"));
+        this.requests = Metrics.meter(factory.createMetricName("Requests"));
+
+        this.hitRate = new RatioGaugeSet(hits, requests, factory, "%sHitRate");
+        this.missRate = new RatioGaugeSet(misses, requests, factory, "%sMissRate");
```
maedhroz, line 59: So we're using RGS here, but the metrics should be published w/ the exact same names and locations, right?

src/java/org/apache/cassandra/metrics/AccordStateCacheMetrics.java
maedhroz: The new changes look reasonable to me...simpler and clearer.

src/java/org/apache/cassandra/metrics/CacheAccessMetrics.java
jacek-lewandowski: I've modified the naming scheme so that it should be identical to the original one

jacek-lewandowski:
@maedhroz, @dcapwell - so is it now ok? can I squash the commits?

maedhroz:
@jacek-lewandowski Things are looking pretty good. The only thing I forgot about WRT preempts in `AccordClientRequestMetrics` is that it's segregated for write/read-only transactions. If that's useful, we might want to add it back. (Hopefully not a big deal...just a few lines.)

Otherwise, +1 to squash...

maedhroz (approved):

jacek-lewandowski:
@maedhroz all the metrics in `AccordMetrics` are already separated to read-only and read-write.

Thanks for the thorough review guys!

jacek-lewandowski:
@dcapwell do you have more comments ?

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/880/workflows/b74b629c-8a40-417a-82c0-1b4d3c75aff3

dcapwell:
im good

PR #2540, opened 2023-08-01
Reduce size of per-SSTable index components for SAI

mike-tr-adamson:
This patch removes the PRIMARY_KEY_TRIE component and adds SortedTermsReader.Cursor.partitionedSeekToTerm to search for primary keys within a partition of the sorted terms. To do this, a new on-disk component PARTITION_SIZES has been added that holds the size of each partition in the SSTable

patch by Mike Adamson; reviewed by <REVIEWERS> for CASSANDRA-18673


.build/parent-pom-template.xml
```diff
@@ -611,13 +611,13 @@
...
         <artifactId>jmh-core</artifactId>
-        <version>1.21</version>
+        <version>1.36</version>
         <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.openjdk.jmh</groupId>
         <artifactId>jmh-generator-annprocess</artifactId>
-        <version>1.21</version>
+        <version>1.36</version>
```
maedhroz: nit: Any chance we can move these to 1.37? I'm getting failures pulling the dependency w/ 1.36...

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -435,6 +435,11 @@ public enum CassandraRelevantProperties
      */
     SAI_POSTINGS_SKIP("cassandra.sai.postings_skip", "3"),
 
+    /**
+     * Used to determine the block size and block mask for sorted terms.
+     */
+    SAI_SORTED_TERMS_BLOCK_SHIFT("cassandra.sai.sorted_terms_block_shift", "4"),
```
maedhroz: Yea. Better than hard-coding...

test/microbench/org/apache/cassandra/test/microbench/sai/SortedTermsBench.java
```diff
@@ -0,0 +1,193 @@
...
+import static org.mockito.Mockito.when;
+
+@BenchmarkMode({Mode.Throughput})
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Warmup(iterations = 5, time = 1)
+@Measurement(iterations = 5, time = 10)
+@Fork(value = 1, jvmArgsAppend = "-Xmx512M")
+@Threads(1)
+@State(Scope.Benchmark)
+public class SortedTermsBench
```
maedhroz: nit: It was a little annoying to run this w/ all the DEBUG messages everywhere, so I changed the level to INFO in `logback-test.xml`

maedhroz: My results look something like this...

```
Benchmark                       (blockShift)  (partitionSize)   Mode  Cnt    Score    Error   Units
SortedTermsBench.advanceToTerm             3               10  thrpt    5  971.451 ± 21.372  ops/ms
SortedTermsBench.advanceToTerm             3              100  thrpt    5  889.358 ± 30.618  ops/ms
SortedTermsBench.advanceToTerm             3             1000  thrpt    5  705.471 ±  8.090  ops/ms
SortedTermsBench.advanceToTerm             3            10000  thrpt    5  710.408 ±  9.807  ops/ms

SortedTermsBench.advanceToTerm             4               10  thrpt    5  975.907 ± 28.423  ops/ms
SortedTermsBench.advanceToTerm             4              100  thrpt    5  916.109 ±  9.445  ops/ms
SortedTermsBench.advanceToTerm             4             1000  thrpt    5  831.456 ±  3.727  ops/ms
SortedTermsBench.advanceToTerm             4            10000  thrpt    5  728.207 ± 10.269  ops/ms

SortedTermsBench.advanceToTerm             5               10  thrpt    5  652.437 ± 20.507  ops/ms
SortedTermsBench.advanceToTerm             5              100  thrpt    5  868.473 ±  8.995  ops/ms
SortedTermsBench.advanceToTerm             5             1000  thrpt    5  756.048 ± 60.404  ops/ms
SortedTermsBench.advanceToTerm             5            10000  thrpt    5  786.428 ±  8.536  ops/ms

SortedTermsBench.advanceToTerm             6               10  thrpt    5  675.240 ± 16.895  ops/ms
SortedTermsBench.advanceToTerm             6              100  thrpt    5  631.230 ± 10.666  ops/ms
SortedTermsBench.advanceToTerm             6             1000  thrpt    5  594.402 ±  4.730  ops/ms
SortedTermsBench.advanceToTerm             6            10000  thrpt    5  529.220 ± 11.949  ops/ms
```

I haven't looked deeply at the methodology, but 4 seems like the clear winner?

test/unit/org/apache/cassandra/index/sai/disk/v1/bitpack/NumericValuesTest.java
```diff
@@ -106,11 +106,11 @@ private void writeTokens(boolean monotonic, IndexDescriptor indexDescriptor, lon
...
-             final NumericValuesWriter numericWriter = new NumericValuesWriter(indexDescriptor,
-                                                                               IndexComponent.TOKEN_VALUES,
-                                                                               metadataWriter,
-                                                                               monotonic,
-                                                                               blockSize))
+             final NumericValuesWriter numericWriter = NumericValuesWriter.create(indexDescriptor,
+                                                                                  IndexComponent.TOKEN_VALUES,
+                                                                                  metadataWriter,
+                                                                                  monotonic,
+                                                                                  blockSize))
```
maedhroz: One thing I noticed here...

Above in `doTest()`, we have this:

```
writeTokens(monotonic, indexDescriptor, array, prev -> monotonic ? prev + nextInt(100) : nextInt(100));
```

In the monotonic case, this is probably fine, but otherwise, I wonder if it would be more appropriate to generate random longs rather than random ints. (i.e. Are there any cases we're missing using just int?)

test/unit/org/apache/cassandra/index/sai/utils/SAIRandomizedTester.java
```diff
@@ -80,6 +80,29 @@ public static IndexDescriptor newIndexDescriptor() throws IOException
...
+                                                                        randomSimpleString(3, 17),
+                                                                        new SequenceBasedSSTableId(getRandom().nextIntBetween(0, 128))),
+                                                         metadata,
+                                                         SequentialWriterOption.newBuilder()
+                                                                               .bufferSize(getRandom().nextIntBetween(17, 1 << 13))
+                                                                               .bufferType(getRandom().nextBoolean() ? BufferType.ON_HEAP : BufferType.OFF_HEAP)
+                                                                               .trickleFsync(getRandom().nextBoolean())
+                                                                               .trickleFsyncByteInterval(nextInt(1 << 10, 1 << 16))
+                                                                               .finishOnClose(true)
+                                                                               .build());
```
maedhroz: Is there any reason we don't just use `IndexFileUtils.DEFAULT_WRITER_OPTION` here (and in `newIndexDescriptor()`)? Callers of `newClusteringIndexDescriptor()` and `newIndexDescriptor()` don't seem to really benefit much from the randomization, as they just perform a single test run with an arbitrary ` SequentialWriterOption` rather than performing multiple iterations. (I guess you could, for example, run `TermsReaderTest` many times manually or wait for it to flake over time in CI...but I don't imagine that's the intended outcome.)

maedhroz: This would, of course, remove some duplication between `newClusteringIndexDescriptor()` and `newIndexDescriptor()`...

test/unit/org/apache/cassandra/index/sai/utils/PrimaryKeyTest.java
```diff
@@ -360,22 +348,4 @@ private void compareToAndEqualsTests(PrimaryKey.Factory factory, PrimaryKey... k
...
-            assertByteComparison(key, key, 0);
-            assertByteComparison(tokenOnlyKey, tokenOnlyKey, 0);
-
-            for (int comparisonIndex = index + 1; comparisonIndex < keys.length; comparisonIndex++)
-            {
-                assertByteComparison(key, keys[comparisonIndex], -1);
-                assertByteComparison(tokenOnlyKey, keys[comparisonIndex], -1);
-            }
-        }
-    }
```
maedhroz, line 380: So I'm trying to understand this in the context of the changes in `PrimaryKey`, and it might just be late, but I'm a bit confused. Instead of deleting this, I tried something else...

```
private void byteComparisonTests(PrimaryKey.Factory factory, PrimaryKey... keys)
{
    for (int index = 0; index < keys.length - 1; index++)
    {
        PrimaryKey key = keys[index];
        assertByteComparison(key, key, 0);

        for (int comparisonIndex = index + 1; comparisonIndex < keys.length; comparisonIndex++)
        {
            assertTrue(compareTerms(readBytes(key), readBytes(keys[comparisonIndex])) < 0);
        }
    }
}

private int compareTerms(BytesRef left, BytesRef right)
{
    return FastByteOperations.compareUnsigned(left.bytes, left.offset, left.offset + left.length,
                                              right.bytes, right.offset, right.offset + right.length);
}

private BytesRef readBytes(ByteComparable source)
{
    BytesRefBuilder builder = new BytesRefBuilder();

    ByteSource byteSource = source.asComparableBytes(ByteComparable.Version.OSS50);
    int val;
    while ((val = byteSource.next()) != ByteSource.END_OF_STREAM)
        builder.append((byte) val);
    return builder.get();
}
```
This seems to be more or less what `SortedTermsReader` does, and we've got a sequence of strictly increasing integers for our simple primary keys, but it fails. Is there no way to compare them in unsigned byte space?

maedhroz: Okay, so I think I see what happened! The sort takes the token into account, and that reorders the byte-comparables. So you could do this, and it would work just fine:

```
@Test
    public void singlePartitionTest()
    {
        PrimaryKey.Factory factory = new PrimaryKey.Factory(simplePartition.comparator);
        int rows = nextInt(10, 100);
        PrimaryKey[] keys = new PrimaryKey[rows];
        for (int index = 0; index < rows; index++)
            keys[index] = factory.create(makeKey(simplePartition, index), Clustering.EMPTY);

        byteComparisonTests(keys);
        Arrays.sort(keys);
        compareToAndEqualsTests(factory, keys);
    }
```
...and then...
```
private void byteComparisonTests(PrimaryKey... keys)
    {
        for (int index = 0; index < keys.length - 1; index++)
        {
            PrimaryKey key = keys[index];
            assertByteComparison(key, key, 0);

            for (int comparisonIndex = index + 1; comparisonIndex < keys.length; comparisonIndex++)
            {
                assertByteComparison(key, keys[comparisonIndex], -1);
            }
        }
    }
```
I suppose your removal says that you felt this wasn't really that useful, given we're not in the business of testing byte-comparable types themselves, and our usage of this is covered in `SortedTermsTest`?

test/unit/org/apache/cassandra/index/sai/utils/AbstractPrimaryKeyTester.java
```diff
@@ -154,7 +154,7 @@ DecoratedKey makeKey(TableMetadata table, Object...partitionKeys)
         return table.partitioner.decorateKey(key);
```
maedhroz, line 154: nit: It might be nicer to use `key = table.partitionKeyType.decomposeUntyped(partitionKeys[0]);` above on 153 to avoid some unnecessary `String` conversions?

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
```diff
@@ -163,19 +159,61 @@ public String toString()
...
+            {
+                return null;
+            }
+
+            @Nullable
+            @Override
+            public Clustering<?> clustering()
+            {
+                return null;
+            }
```
maedhroz, line 199: nit: I'd consider an override of `asComparableBytes()` here that throw `UnsupportedOperationException` instead of NPE.

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
```diff
@@ -243,6 +287,8 @@ private void loadDeferred()
         }
     }
 
+    boolean isTokenOnly();
```
maedhroz: nit: Perhaps this could just be a default method that returns false, then only the token-only case would have to override?

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
```diff
@@ -90,32 +90,27 @@ public PrimaryKey createDeferred(Token token, Supplier<PrimaryKey> primaryKeySup
...
+            @SuppressWarnings("ConstantConditions")
             public ByteSource asComparableBytes(ByteComparable.Version version)
             {
-                ByteSource tokenComparable = token().asComparableBytes(version);
-                if (partitionKey() == null)
-                    return ByteSource.withTerminator(version == ByteComparable.Version.LEGACY ? ByteSource.END_OF_STREAM
-                                                                                              : ByteSource.TERMINATOR,
-                                                     tokenComparable,
-                                                     null,
-                                                     null);
```
maedhroz, line 101: So much like the previous version of this patch, we ditch the token, since there's just nothing we need to do with it in byte-comparable space. (All we do is write entries to the sorted terms structure, etc.) Unlike the previous patch, we don't even have to buffer anything, since the trie is gone. Good times all around!

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
```diff
@@ -49,9 +43,9 @@
...
  * For documentation of the underlying on-disk data structures, see the package documentation.
  * <p>
- * The TERMS_DICT_ constants allow for quickly determining the id of the current block based on a point id
- * or to check if we are exactly at the beginning of the block.
- * Terms data are organized in blocks of (2 ^ {@link #TERMS_DICT_BLOCK_SHIFT}) terms.
+ * The {@code cassandra.sai.sorted_terms_block_shift} property is used to quickly determine the id of the current block
+ * based on a point id or to check if we are exactly at the beginning of the block.
+ * Terms data are organized in blocks of (2 ^ {@link #blockShift}) terms.
  * The blocks should not be too small because they allow prefix compression of
  * the terms except the first term in a block.
```
maedhroz: One of the things I've been mulling around with respect to the whole sorted terms structure is that it seems like we miss some opportunities for compression as partition size increases. Let's say we have 16 terms per block, and a simple `(k, c) PRIMARY KEY`. If there are 16 rows in a partition at `k = 5`, we compress away those partitions optimally. If we have 64 rows in that partition, we're now storing `5` four times. Of course, the identical partition across the 4 blocks means we don't degrade potential prefix compression for `c`, but there might be some ways we can do better?

This obviously complicates some things, but let's say we wrote two structures: one for the partition key, and one for the clustering keys. We would be able to control block size separately, if that had potential to prefix compress partition keys better. The other thing, though, is that when we seek/skip within a partition, we could read from only the clustering key structure, given it's indexed the same way in row ID space.

WDYT?

mike-tr-adamson: I have had exactly the same thoughts. Since we now know exactly how long each partition is, there is no real need for these to be combined. We also know the partition ID for the row ID, so we wouldn't need to store the partition key for each row, we'd only need to store it for the partition. 

Let me try this and see how it looks.

src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
```diff
@@ -68,19 +68,23 @@ public enum IndexComponent
...
+     * An on-disk block packed index containing the starting and ending rowIds for each partition.
      */
-    PRIMARY_KEY_TRIE("PrimaryKeyTrie"),
+    PARTITION_SIZES("PartitionSizes"),
 
     /**
      * Prefix-compressed blocks of primary keys used for rowId to partition key lookups
      */
-    PRIMARY_KEY_BLOCKS("PrimaryKeyBlocks"),
+    PARTITION_KEY_BLOCKS("PartitionKeyBlocks"),
```
adelapena, line 78: The JavaDoc of `PARTITION_KEY_BLOCKS` and `PARTITION_KEY_BLOCK_OFFSETS` would need to be updated to talk about partition keys instead of primary keys. Also, the new components for clustering keys would also need some JavaDoc.

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
```diff
@@ -219,10 +219,10 @@ public Set<Component> getComponents()
         return getComponents(indexes);
     }
 
-    static Set<Component> getComponents(Collection<StorageAttachedIndex> indices)
+    Set<Component> getComponents(Collection<StorageAttachedIndex> indices)
```
adelapena: Nit: can be `private`

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
```diff
@@ -49,9 +44,9 @@
  * <p>
  * For documentation of the underlying on-disk data structures, see the package documentation.
  * <p>
- * The TERMS_DICT_ constants allow for quickly determining the id of the current block based on a point id
- * or to check if we are exactly at the beginning of the block.
- * Terms data are organized in blocks of (2 ^ {@link #TERMS_DICT_BLOCK_SHIFT}) terms.
+ * The {@code cassandra.sai.sorted_terms_block_shift} property is used to quickly determine the id of the current block
```
adelapena: Nit: This can be a reference to `{@link CassandraRelevantProperties#SAI_SORTED_TERMS_BLOCK_SHIFT}`

test/microbench/org/apache/cassandra/test/microbench/sai/SortedTermsBench.java
```diff
@@ -0,0 +1,205 @@
...
+
+    protected TableMetadata metadata;
+    protected IndexDescriptor indexDescriptor;
+
+    private PrimaryKeyMap primaryKeyMap;
+
+    private PrimaryKey primaryKey;
+
+    @Param({"3", "4", "5"})
+    private int partitionBlockShift;
```
adelapena: Nit: All the attributes annotated with `@Param` can be `public` to prevent warnings

test/microbench/org/apache/cassandra/test/microbench/sai/SortedTermsBench.java
```diff
@@ -0,0 +1,205 @@
...
+
+        indexDescriptor = IndexDescriptor.create(descriptor, metadata.comparator);
+
+        CassandraRelevantProperties.SAI_SORTED_TERMS_PARTITION_BLOCK_SHIFT.setInt(partitionBlockShift);
+        CassandraRelevantProperties.SAI_SORTED_TERMS_CLUSTERING_BLOCK_SHIFT.setInt(clusteringBlockShift);
+        SSTableComponentsWriter writer = new SSTableComponentsWriter(indexDescriptor);
+
+        PrimaryKey.Factory factory = new PrimaryKey.Factory(metadata.comparator);
+
+        int rows = 1000000;
```
adelapena: This could be a constant. I would also write it as `1_000_000` for readability.

test/microbench/org/apache/cassandra/test/microbench/sai/SortedTermsBench.java
```diff
@@ -0,0 +1,205 @@
...
+        writer.complete();
+
+        SSTableReader sstableReader = mock(SSTableReader.class);
+        when(sstableReader.metadata()).thenReturn(metadata);
+
+        PrimaryKeyMap.Factory mapFactory = new WideRowAwarePrimaryKeyMap.WideRowAwarePrimaryKeyMapFactory(indexDescriptor, sstableReader);
+
+        primaryKeyMap = mapFactory.newPerSSTablePrimaryKeyMap();
+
+        primaryKey = primaryKeys[500000];
```
adelapena: Why `500000 `?

test/microbench/org/apache/cassandra/test/microbench/sai/SortedTermsBench.java
```diff
@@ -0,0 +1,205 @@
...
+        CassandraRelevantProperties.SAI_SORTED_TERMS_CLUSTERING_BLOCK_SHIFT.setInt(clusteringBlockShift);
+        SSTableComponentsWriter writer = new SSTableComponentsWriter(indexDescriptor);
+
+        PrimaryKey.Factory factory = new PrimaryKey.Factory(metadata.comparator);
+
+        int rows = 1000000;
+
+        PrimaryKey[] primaryKeys = new PrimaryKey[rows];
+        int partition = 0;
+        int partitionCounter = 0;
```
adelapena: `partitionCounter` might look like we are counting partitions, instead of rows. Perhaps this could be name `partitionRows`, or simply `rows`?

test/microbench/org/apache/cassandra/test/microbench/sai/SortedTermsBench.java
```diff
@@ -0,0 +1,205 @@
...
+        primaryKey = primaryKeys[500000];
+    }
+
+    @Benchmark
+    public long advanceToTerm()
+    {
+        return primaryKeyMap.rowIdFromPrimaryKey(primaryKey);
+    }
+
+    protected DecoratedKey makeKey(TableMetadata table, Object...partitionKeys)
```
adelapena: Nit: can be `private static`

test/microbench/org/apache/cassandra/test/microbench/sai/SortedTermsBench.java
```diff
@@ -0,0 +1,205 @@
...
+    {
+        ByteBuffer key;
+        if (TypeUtil.isComposite(table.partitionKeyType))
+            key = ((CompositeType)table.partitionKeyType).decompose(partitionKeys);
+        else
+            key = table.partitionKeyType.fromString((String)partitionKeys[0]);
+        return table.partitioner.decorateKey(key);
+    }
+
+    protected Clustering<?> makeClustering(TableMetadata table)
```
adelapena: Nit: can be `private`

src/java/org/apache/cassandra/index/sai/disk/v1/LongArray.java
```diff
@@ -37,6 +37,13 @@ public interface LongArray extends Closeable
      */
     long length();
 
+    /**
+     * Using the given value returns the first index corresponding to the value.
```
adelapena, line 41:
```suggestion
     * Using the given value returns the first index corresponding to the value.
     *
```

src/java/org/apache/cassandra/index/sai/disk/v1/SkinnyRowAwarePrimaryKeyMap.java
```diff
@@ -0,0 +1,184 @@
...
+ *     <li>A sorted-terms structure for rowId to {@link PrimaryKey} and {@link PrimaryKey} to rowId lookups using
+ *     {@link SortedTermsReader}. Uses components {@link IndexComponent#PARTITION_KEY_BLOCKS} and
+ *     {@link IndexComponent#PARTITION_KEY_BLOCK_OFFSETS}</li>
+ * </ul>
+ *
+ * While the {@link SkinnyRowAwarePrimaryKeyMapFactory} is threadsafe, individual instances of the {@link SkinnyRowAwarePrimaryKeyMap}
+ * are not.
+ */
+@NotThreadSafe
+public class SkinnyRowAwarePrimaryKeyMap implements PrimaryKeyMap
```
adelapena: I wonder if the row-aware concept should be mentioned in the name of the class and the first line of the JavaDoc. We don't have any other type of primary key maps, and row awareness seems meaningless for the skinny tables where `SkinnyRowAwarePrimaryKeyMap` is meant to be used.

So maybe the classes could be named just `SkinnyPrimaryKeyMap` and `WidePrimaryKeyMap`?

src/java/org/apache/cassandra/index/sai/disk/v1/SkinnyRowAwarePrimaryKeyMap.java
```diff
@@ -0,0 +1,184 @@
...
+{
+    @ThreadSafe
+    public static class SkinnyRowAwarePrimaryKeyMapFactory implements Factory
```
adelapena: Being an inner class, it can be named `Factory`, implementing `PrimaryKeyMap.Factory`.

src/java/org/apache/cassandra/index/sai/disk/v1/WideRowAwarePrimaryKeyMap.java
```diff
@@ -0,0 +1,178 @@
...
+ * An extension of the {@link SkinnyRowAwarePrimaryKeyMap} for wide tables (those with clustering columns).
+ * <p>
+ * While the {@link WideRowAwarePrimaryKeyMapFactory} is threadsafe, individual instances of the {@link WideRowAwarePrimaryKeyMap}
+ * are not.
+ */
+@NotThreadSafe
+public class WideRowAwarePrimaryKeyMap extends SkinnyRowAwarePrimaryKeyMap
+{
+    @ThreadSafe
+    public static class WideRowAwarePrimaryKeyMapFactory extends SkinnyRowAwarePrimaryKeyMapFactory
```
adelapena: Being an inner class, it can be named `Factory`, implementing `SkinnyRowAwarePrimaryKeyMap.Factory`.

src/java/org/apache/cassandra/index/sai/disk/v1/WideRowAwarePrimaryKeyMap.java
```diff
@@ -0,0 +1,178 @@
...
+    {
+        private final ClusteringComparator clusteringComparator;
+        protected final LongArray.Factory partitionReaderFactory;
+        protected final SortedTermsReader clusteringKeyReader;
+        protected final SortedTermsMeta clusteringKeyMeta;
+
+        protected FileHandle partitionsFile;
+        protected FileHandle clusteringKeyBlockOffsetsFile;
+        protected FileHandle clustingingKeyBlocksFile;
```
adelapena: All these attributes can be `private`. Also, `clusteringKeyMeta` can be a local var of the constructor.

src/java/org/apache/cassandra/index/sai/disk/v1/SkinnyRowAwarePrimaryKeyMap.java
```diff
@@ -0,0 +1,184 @@
...
+    {
+        protected final MetadataSource metadataSource;
+        protected final LongArray.Factory tokenReaderFactory;
+        protected final LongArray.Factory partitionReaderFactory;
+        protected final SortedTermsReader partitionKeyReader;
+        protected final SortedTermsMeta partitionKeysMeta;
```
adelapena: `partitionKeysMeta` can be a local var of the constructor.

src/java/org/apache/cassandra/index/sai/disk/v1/SkinnyRowAwarePrimaryKeyMap.java
```diff
@@ -0,0 +1,184 @@
...
+        protected final IPartitioner partitioner;
+        protected final PrimaryKey.Factory primaryKeyFactory;
+
+        protected FileHandle tokensFile;
+        protected FileHandle partitionsFile;
+        protected FileHandle partitionKeyBlockOffsetsFile;
+        protected FileHandle partitionKeyBlocksFile;
```
adelapena: These four attributes can be `private`.

src/java/org/apache/cassandra/index/sai/disk/v1/bitpack/NumericValuesWriter.java
```diff
@@ -22,79 +22,98 @@
...
-    private final MetadataWriter metadataWriter;
-    private final String componentName;
-    private final int blockSize;
-    private long count = 0;
-
-    public NumericValuesWriter(String componentName,
-                               IndexOutput indexOutput,
-                               MetadataWriter metadataWriter,
-                               boolean monotonic) throws IOException
+    NumericValuesWriter NOOP_WRITER = new NumericValuesWriter()
```
adelapena: This doesn't seem used anywhere.

src/java/org/apache/cassandra/index/sai/disk/v1/WideRowAwarePrimaryKeyMap.java
```diff
@@ -0,0 +1,178 @@
...
+                                                 partitioner,
+                                                 primaryKeyFactory,
+                                                 clusteringComparator);
+        }
+
+        @Override
+        public void close()
+        {
+            super.close();
+            FileUtils.closeQuietly(Arrays.asList(partitionsFile, clustingingKeyBlocksFile, clusteringKeyBlockOffsetsFile));
```
maedhroz: `partitionsFile` is already closed in the `super.close()`, right?

src/java/org/apache/cassandra/index/sai/disk/v1/bitpack/NumericValuesWriter.java
maedhroz: hmmm, yeah, the interface probably collapses if nothing actually needs it

src/java/org/apache/cassandra/index/sai/disk/v1/bitpack/AbstractBlockPackedReader.java
```diff
@@ -34,6 +34,9 @@ public abstract class AbstractBlockPackedReader implements LongArray
     private final byte[] blockBitsPerValue;
     private final SeekingRandomAccessInput input;
 
+    private long prevTokenValue = Long.MIN_VALUE;
```
maedhroz:
```suggestion
    private long prevValue = Long.MIN_VALUE;
```
?

test/unit/org/apache/cassandra/index/sai/iterators/KeyRangeConcatIteratorTest.java
```diff
@@ -33,8 +33,8 @@
 public class KeyRangeConcatIteratorTest extends AbstractKeyRangeIteratorTester
 {
     private static final String MUST_BE_SORTED_ERROR = "RangeIterator must be sorted, previous max: PrimaryKey: " +
-                                                       "{ token: %d, partition: null, clustering: null:null} , " +
-                                                       "next min: PrimaryKey: { token: %d, partition: null, clustering: null:null} ";
+                                                       "{ token: %d }, " +
+                                                       "next min: PrimaryKey: { token: %d }";
```
maedhroz: nit: Not terribly important, but could create a constant in `KeyRangeConcatIterator` and reuse here...modulo the `PrimaryKey` bits...

test/microbench/org/apache/cassandra/test/microbench/sai/SortedTermsBench.java
mike-tr-adamson: I have extended the test for the separate clustering terms and 4 still comes out the winner. I still think it's worth keeping the size configurable, for when we have more time and can, potentially, look at ways of improving the compression.

test/unit/org/apache/cassandra/index/sai/disk/v1/bitpack/NumericValuesTest.java
mike-tr-adamson: No, I think this was just laziness by whoever wrote the test. I've changed it to generate longs.

test/unit/org/apache/cassandra/index/sai/utils/SAIRandomizedTester.java
mike-tr-adamson: Agreed. A lot of this randomization doesn't always make sense. It definitely doesn't make sense for us to be testing these aspects of the writers. They should be tested elsewhere.

test/unit/org/apache/cassandra/index/sai/utils/PrimaryKeyTest.java
mike-tr-adamson: I tried various options here but, as you say, we aren't really in the business of testing byte comparables so it makes no sense to try and do it here. By removing the token from the byte comparable, the primary keys no longer maintain any order. 

I have considered not using byte comparable at all for the primary keys and this may be option for the future. We could just serialize the keys and use byte buffers in place of the comparables. This would allow us to remove some of the array copying that we currently do and remove our dependency of the `BytesRef` in the sorted terms.

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
mike-tr-adamson: Agreed.

This should also get simplified further in CASSANDRA-18734 because we will no longer need the `createPartitionKeyOnly` method. This will allow us to remove the `@Nullable` from the `partitionKey` and `clustering` methods.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
mike-tr-adamson: This now changes to `{@link #blockShift}` because the value is now passed to the constructor.

test/microbench/org/apache/cassandra/test/microbench/sai/SortedTermsBench.java
mike-tr-adamson: It was a random decision. Realistically we should test at different points to test the impact of the lookup being at different parts of the partition and block. I'll give this some thought and see what is possible.

test/microbench/org/apache/cassandra/test/microbench/sai/SortedTermsBench.java
mike-tr-adamson: I've changed this to `partitionRowCounter`

src/java/org/apache/cassandra/index/sai/disk/v1/SkinnyRowAwarePrimaryKeyMap.java
mike-tr-adamson: This sounds better. Some of the class names are getting verbose and we do only support row awareness.

mike-tr-adamson: I have also sorted out the javadoc on both classes to make them more correct.

src/java/org/apache/cassandra/index/sai/disk/v1/bitpack/NumericValuesWriter.java
mike-tr-adamson: This was put in when we were only, optionally, creating the partition size structure but isn't needed any more. I'll remove it and get rid of it and the interface.

src/java/org/apache/cassandra/index/sai/disk/v1/WideRowAwarePrimaryKeyMap.java
mike-tr-adamson: Correct, I've fixed this.

src/java/org/apache/cassandra/index/sai/disk/v1/bitpack/AbstractBlockPackedReader.java
mike-tr-adamson: Or, `previousValue`. I've changed it to that in my mission to make this all more readable.

test/unit/org/apache/cassandra/index/sai/utils/PrimaryKeyTest.java
maedhroz: > I have considered not using byte comparable at all for the primary keys and this may be option for the future.

I'd wondered that reading this patch as well. Assuming it doesn't have any real impact on compression ratio, that would be the way to go. Not opposed to it being a separate issue, but you can also make a case we never would have used byte-comparable stuff without the trie. To someone not familiar w/ the evolution of the sorted terms stuff, it will look a bit odd :D

maedhroz: There is this bit in the package info...

```
 * Terms are stored in <code>ByteComparable</code> strictly ascending order.
 * Duplicates are not allowed.
```

Using `ByteComparable` does ensure that the values are ordered by their unsigned representation, and `SortedTermsReader` relies on that, right?

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
```diff
@@ -49,9 +43,9 @@
  * <p>
```
maedhroz: Thinking through one last time how we explain what's going on here...

Rather than saying terms must be added in lexicographical order, would it make more sense to say that they must be added in their unsigned byte-comparable order, given that's what the reader seems to assume?

Aside/tangent: Just in general, the use of "term" to describe the values here is almost more confusing than just calling them "blobs" or something. Maybe this was a holdover from a structure in Lucene this was based on originally or something, but what it has become is an ordered sequence of unsigned byte-comparable blobs :D

maedhroz: Any terminology we **do** change here probably bears updating in the package JavaDoc thing too...

maedhroz: Aside from the first comment about the "lexicographical" bit, I wouldn't worry about my tangent, to be clear. Continuing to call these "terms" is fine.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
```diff
@@ -18,28 +18,22 @@
...
 import org.apache.lucene.store.IndexOutput;
 import org.apache.lucene.util.BytesRef;
 import org.apache.lucene.util.BytesRefBuilder;
 import org.apache.lucene.util.StringHelper;
 
-import static org.apache.cassandra.index.sai.disk.v1.trie.TrieTermsDictionaryReader.trieSerializer;
+import javax.annotation.Nonnull;
+import javax.annotation.concurrent.NotThreadSafe;
+import java.io.Closeable;
+import java.io.IOException;
```
maedhroz: nit: Looks like `java` and `javax` imports got auto-formatted down here?

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
```diff
@@ -90,30 +80,29 @@ public class SortedTermsWriter implements Closeable
...
+        this.blockMask = (1 << this.blockShift) - 1;
+        this.termsOutput = termsOutput;
+        this.termsOutput.writeVInt(blockShift);
+        this.bytesStartFP = termsOutput.getFilePointer();
         this.offsetsWriter = termsDataBlockOffsets;
     }
 
     /**
      * Appends a term at the end of the sequence.
-     * Terms must be added in lexicographic order.
```
maedhroz, line 116: Are they still added in unsigned byte-comparable order? If not, they'd have to be sorted again?

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
```diff
@@ -179,36 +173,29 @@ public void close() throws IOException
...
-            SortedTermsMeta.write(output, trieFilePointer, pointId, maxLength);
+            SortedTermsMeta.write(output, pointId, maxTermLength);
             // Don't close the offsets writer quietly because of the work it does
             // during its close. We need to propagate the error.
             offsetsWriter.close();
         }
         finally
         {
-            FileUtils.closeQuietly(Arrays.asList(trieWriter, trieOutput, termsOutput));
+            FileUtils.closeQuietly(termsOutput);
```
maedhroz: Should we close normally rather than quietly here? i.e. We need the `SequentialWriter` to close properly?

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
```diff
@@ -152,20 +136,30 @@ private void writeTermData(BytesRef term) throws IOException
...
-            int prefixLength = StringHelper.bytesDifference(prevTerm.get(), term);
-            int suffixLength = term.length - prefixLength;
-            assert suffixLength > 0: "terms must be unique";
-
+            int prefixLength = 0;
+            int suffixLength = 0;
+
+            // If the term is the same as the previous term then we use prefix and suffix lengths of 0.
+            // This means that we store a byte of 0 and don't write any data for the term.
+            if (compareTerms(prevTerm.get(), term) != 0)
```
maedhroz: If this is `> 0`, we are in trouble, right?

maedhroz: Oh, will `bytesDifference()` throw `IllegalArgumentException` if we do that?

src/java/org/apache/cassandra/index/sai/disk/v1/WidePrimaryKeyMap.java
```diff
@@ -0,0 +1,177 @@
...
+
+    private Clustering<?> readClusteringKey(long sstableRowId)
+    {
+        ByteSource.Peekable peekable = ByteSource.peekable(clusteringKeyCursor.seekForwardToPointId(sstableRowId)
+                                                                              .asComparableBytes(ByteComparable.Version.OSS50));
+
+        Clustering<?> clustering = clusteringComparator.clusteringFromByteComparable(ByteBufferAccessor.instance, v -> peekable);
+
+        if (clustering == null)
+            clustering = Clustering.EMPTY;
```
maedhroz, line 165: nit: Apologies...I keep forgetting what tests hits this code path (empty clustering) every time I review and run code coverage :D

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
```diff
@@ -112,123 +108,259 @@ public SortedTermsReader(@Nonnull FileHandle termsDataFileHandle,
...
+        }
+
+        // Tries to load the starting value of the next block into nextBlockTerm. This will return false
+        // if the pointId is in the last block.
+        private boolean peekNextBlock(long pointId)
+        {
+            long blockIndex = (pointId >>> blockShift) + 1;
+
+            if (blockIndex >= blockOffsets.length())
+                return false;
```
maedhroz: Trying to hit this line in a test, without success.

`notInCurrentBlock()` checks `inLastBlock()`...and then also checks that (albeit in an open-ended way) here in `peekNextBlock()`. Assuming we can do earlier validation to make sure we aren't trying to go to a block that doesn't exist, could we just rewrite this chunk of the code like this?

```
private boolean notInCurrentBlock(long pointId, BytesRef term)
{
    if (inLastBlock(pointId))
        return false;

    peekNextBlock(pointId);
    resetPosition();

    return compareTerms(term, nextBlockTerm) >= 0;
}
...
private void peekNextBlock(long pointId)
{
    long blockIndex = (pointId >>> blockShift) + 1;
    termsInput.seek(blockOffsets.get(blockIndex) + termsDataFp);
    readTerm(blockIndex << blockShift, nextBlockTerm);
}
```

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
```diff
@@ -112,123 +108,259 @@ public SortedTermsReader(@Nonnull FileHandle termsDataFileHandle,
...
+         * <p>
+         * It is allowed to position the cursor before the first item or after the last item;
+         * in these cases the internal buffer is cleared.
+         *
+         * @param nextPointId point id to lookup
+         * @return The {@link ByteComparable} containing the term
+         * @throws IndexOutOfBoundsException if the target point id is less than -1 or greater than the number of terms
          */
-        public long pointId()
+        public @Nonnull ByteComparable seekForwardToPointId(long nextPointId)
```
maedhroz:
```suggestion
        public @Nonnull ByteComparable seekForwardToPointId(long pointId)
```

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
```diff
@@ -112,123 +108,259 @@ public SortedTermsReader(@Nonnull FileHandle termsDataFileHandle,
...
         {
-            return pointId;
-        }
+            if (nextPointId < 0 || nextPointId > meta.termCount)
```
maedhroz: Shouldn't this be `nextPointId >= meta.termCount`? (Corresponding change to the error message, etc.)

Fun thing is that if you add...

```
assertThatThrownBy(() -> cursor.seekForwardToPointId(4000L)).isInstanceOf(IndexOutOfBoundsException.class);
```

...to `testSeekToPointIdOutOfRange()`, the test passes, but only because the `LongArray` in `resetPosition()` throws `IOOBE` :)

src/java/org/apache/cassandra/index/sai/disk/v1/WidePrimaryKeyMap.java
```diff
@@ -0,0 +1,177 @@
...
+        this.clusteringKeyCursor = clusteringKeyCursor;
+    }
+
+    @Override
+    public long rowIdFromPrimaryKey(PrimaryKey primaryKey)
+    {
+        long rowId = tokenArray.indexOf(primaryKey.token().getLongValue());
+
+        // If the key only has a token (initial range skip in the query), the token is out of range,
+        // or we have skipped a token, return the rowId from the token array.
```
maedhroz, line 134: > or we have skipped a token

This means we skipped past the token of the provided primary key to a new partition whose token is greater?

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
```diff
@@ -112,123 +108,259 @@ public SortedTermsReader(@Nonnull FileHandle termsDataFileHandle,
...
-         *
-         * @param pointId point id to lookup
-         * @throws IOException if a seek and read from the terms file fails
-         * @throws IndexOutOfBoundsException if the target point id is less than -1 or greater than the number of terms
+         * If the term is not in the block containing the start of the range a binary search is done to find
+         * the block containing the search. That block is then searched to return the pointId that corresponds
+         * to the term that either equal to or next highest to the term.
          */
-        public void seekToPointId(long pointId) throws IOException
+        public long partitionedSeekToTerm(ByteComparable term, long startingPointId, long endingPointId)
```
maedhroz: Is `endingPointId` exclusive? Would perhaps mention that in the method JavaDoc...

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
```diff
@@ -112,123 +108,259 @@ public SortedTermsReader(@Nonnull FileHandle termsDataFileHandle,
...
-            // added to the existing prefix.
-            termsInput.readBytes(currentTerm.bytes, prefixLength, suffixLength);
-            return true;
+        private int compareTerms(BytesRef left, BytesRef right)
+        {
+            return FastByteOperations.compareUnsigned(left.bytes, left.offset, left.offset + left.length,
+                                                      right.bytes, right.offset, right.offset + right.length);
+        }
+
+        private BytesRef readBytes(ByteComparable source)
```
maedhroz:
```suggestion
        private BytesRef asBytesRef(ByteComparable source)
```
nit

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
```diff
@@ -112,123 +108,259 @@ public SortedTermsReader(@Nonnull FileHandle termsDataFileHandle,
...
+
+            termsInput.seek(blockOffsets.get(blockIndex) + termsDataFp);
+            readTerm(blockIndex << blockShift, nextBlockTerm);
+
+            return true;
+        }
+
+        // Reset currentPointId and currentTerm to be at the start of the block
+        // pointed to by currentBlockIndex.
+        private void resetPosition()
```
maedhroz: Trying to come up w/ some alternate names for this method...

`goToCurrentBlock()`, `resetToCurrentBlock()`, `loadCurrentBlock()`, `loadAtCurrentBlock()`... WDYT?

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
```diff
@@ -112,123 +108,259 @@ public SortedTermsReader(@Nonnull FileHandle termsDataFileHandle,
...
+        {
+            termsInput.seek(blockOffsets.get(currentBlockIndex) + termsDataFp);
+            currentPointId = currentBlockIndex << blockShift;
+            readTerm(currentPointId, currentTerm);
```
maedhroz: nit: Might be nice to have a `loadCurrentTerm()` shortcut for this.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
```diff
@@ -112,123 +108,259 @@ public SortedTermsReader(@Nonnull FileHandle termsDataFileHandle,
...
+            // Move forward to the ending point ID, returning the point ID if we find our term
+            while (currentPointId < endingPointId)
+            {
+                if (compareTerms(currentTerm, skipTerm) >= 0)
+                    return currentPointId;
+                currentPointId++;
+                if (currentPointId == meta.termCount)
+                    return -1;
+                readTerm(currentPointId, currentTerm);
+                currentBlockIndex = currentPointId >>> blockShift;
```
maedhroz: nit: We do this in a couple places...might be nice to have an `updateCurrentBlockIndex()` or an `advanceBlockIndex(pointId)` or something similar...

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
```diff
@@ -112,123 +108,259 @@ public SortedTermsReader(@Nonnull FileHandle termsDataFileHandle,
...
+            }
+            return endingPointId < meta.termCount ? endingPointId : -1;
```
maedhroz: nit: throw a brief `@return` in the method JavaDoc?

maedhroz: ...and if we don't find the term, we can also just return the ending point ID, if it hasn't run off the end of the term sequence?

In any case, we don't actually care about exact matching here, since the use case is skipping, etc.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
```diff
@@ -112,123 +108,259 @@ public SortedTermsReader(@Nonnull FileHandle termsDataFileHandle,
...
+                    if (cmp == 0)
+                        return currentPointId;
+
+                    if (cmp < 0)
+                        splitPointId = currentPointId;
+
+                    split /= 2;
+                }
+                // After we finish the binary search we need to move the block back till we hit a block that has
+                // a starting term that is less than or equals to the skip term
```
maedhroz:
```suggestion
                // a starting term that is less than or equal to the skip term
```

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
```diff
@@ -112,123 +108,259 @@ public SortedTermsReader(@Nonnull FileHandle termsDataFileHandle,
...
+
+            currentBlockIndex = startingPointId >>> blockShift;
+            resetPosition();
+
+            if (compareTerms(currentTerm, skipTerm) == 0)
+                return startingPointId;
+
+            if (notInCurrentBlock(startingPointId, skipTerm))
+            {
+                long split = (endingPointId - startingPointId) >>> blockShift;
```
maedhroz: So the `split` is essentially the number of blocks we skip forward, and halving it (all the way to 1/2 truncating to zero) on each iteration is what gives us a binary search aesthetic?

test/unit/org/apache/cassandra/index/sai/utils/PrimaryKeyTest.java
mike-tr-adamson: That bit needs to come out. Neither the partition keys nor the clustering keys are in any order, and we do now allow duplicates.

mike-tr-adamson: I would prefer not to remove the `ByteComparable` in this patch. I'd like to combine that in another patch along with removing our reliance on Lucene `DataOutput` and `IndexOutput`. I will raise a ticket to remove this usage so that we can write `ByteBuffer`s directly without having to move them around.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
mike-tr-adamson: We have two separate uses for the sorted terms. The partition keys version is only ever used for getting partition keys from row Ids. This doesn't require any order at all. The clustering keys version can look up row ids from clustering keys but only within a partition where the clustering keys are ordered. 

So, we have a partitioned case where keys must be ordered within a partition and a non-partitioned case where they don't need to be ordered at all. 

This is leading me to think that we have a case here that is similar to the `BlockPackedReader` and the `MonotonicBlockPackedReader` (not literally), where we may need to enforce different rules. In the case of the clustering keys, we should be forcing order within partitions, and the partition keys should not allow calls to the partitioned lookup at all.

I'll get these changes in. 

I have no real objection to the use of 'term' here. This is a generic package that is storing terms. If we want to stop using term we'll need to do it across SAI.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
mike-tr-adamson: No, I have changed this.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
mike-tr-adamson: Agreed. I have also moved the closing of the `offsetsWriter` into the `finally` block because, currently, if the footer write or metadata write fails, it doesn't get closed.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
mike-tr-adamson: If the terms are equal, then we only want to write a byte zero. If the terms are not equal, then `bytesDifference` returns the number of bytes of matching prefix. The term does not need to be in any sorted order. The exception thrown by the `bytesDifference` method is not correct. It will get `-1` back from `Arrays.mismatch` if the terms are equal.

src/java/org/apache/cassandra/index/sai/disk/v1/WidePrimaryKeyMap.java
mike-tr-adamson: I'm not sure about a specific test for it, but this will disappear with CASSANDRA-18734. The only time we have a null clustering is when we create a partition-only key for the segment metadata.

I'd like to defer this to 18734.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
mike-tr-adamson: I have merged `peekNextBlock` into `notInCurrentBlock` because it is not safe without the check. This also allows the `resetPosition` call to be removed in favour or reading the file pointer before moving to the next block and then seeking to that file pointer after. This could save a few term reads by not moving to the beginning of the current block each time this is called.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
mike-tr-adamson: Yes, I've fixed this and added a test for this condition. I've also added `hasMessage` check to the tests to make sure we are getting the exception from the right place.

src/java/org/apache/cassandra/index/sai/disk/v1/WidePrimaryKeyMap.java
mike-tr-adamson: Yes, if we have moved to the next partition then we want to return first row id in the partition. We only need to search the clustering keys if we are in the same partition as the original primary key.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
mike-tr-adamson: I think `resetToCurrentBlock` works best. I've changed it to that.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
mike-tr-adamson: If  `endingPointId == meta.termCount`, this is the last partition. Otherwise, we haven't found the term in the partition, so we return the `endingPointId`, which represents the first row id in the next partition.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
mike-tr-adamson: `updateCurrentBlockIndex(long pointId)` works.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
mike-tr-adamson: Correct, and if we end up at a term less than the term we are looking for we move the `splitPointId = currentPointId`.

src/java/org/apache/cassandra/index/sai/disk/v1/SkinnyPrimaryKeyMap.java
```diff
@@ -0,0 +1,185 @@
...
+                SortedTermsMeta partitionKeysMeta = new SortedTermsMeta(metadataSource.get(indexDescriptor.componentName(IndexComponent.PARTITION_KEY_BLOCKS)));
+                this.partitionKeyBlockOffsetsFile = indexDescriptor.createPerSSTableFileHandle(IndexComponent.PARTITION_KEY_BLOCK_OFFSETS);
+                this.partitionKeyBlocksFile = indexDescriptor.createPerSSTableFileHandle(IndexComponent.PARTITION_KEY_BLOCKS);
+                this.partitionKeyReader = new SortedTermsReader(partitionKeyBlocksFile, partitionKeyBlockOffsetsFile, partitionKeysMeta, partitionKeyBlockOffsetsMeta);
+                this.partitioner = sstable.metadata().partitioner;
+                this.primaryKeyFactory = indexDescriptor.primaryKeyFactory;
+            }
+            catch (Throwable t)
+            {
+                throw Throwables.unchecked(Throwables.close(t, Arrays.asList(tokensFile, partitionsFile, partitionKeyBlocksFile, partitionKeyBlockOffsetsFile)));
```
maedhroz: It would be nice to make these 4 fields `final`. I think it'll compile if we do...

```
close();
throw Throwables.unchecked(t);
```

Is it essential to collect the suppressed exceptions, or is it fine to just log the warnings?

maedhroz (approved):

test/unit/org/apache/cassandra/index/sai/utils/PrimaryKeyTest.java
maedhroz: > I will raise a ticket to remove this usage so that we can write ByteBuffers directly without having to move them around.

+1

src/java/org/apache/cassandra/index/sai/disk/v1/SkinnyPrimaryKeyMap.java
mike-tr-adamson: It works fine if we move all the `FileHandle` creations out of the `try` block. That looks much nicer.

I don't see an issue with leaving the close/throw as is after that change.

src/java/org/apache/cassandra/index/sai/disk/v1/WidePrimaryKeyMap.java
maedhroz: Cool, no worries if it's going away anyway...

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsReader.java
maedhroz: Not sure if it's worth an inline comment, but it makes sense at least.

src/java/org/apache/cassandra/index/sai/disk/v1/SkinnyPrimaryKeyMap.java
mike-tr-adamson: I think I have an elegant solution to this problem. I have added a `cleanup` mechanism to the `IndexDescriptor.createPer*FileHandle` methods that allow a cleanup to be done if the file handle creation fails. In this case, this allows us to pass `this::close` to the method, so any previously opened file handles that need closing will be closed.

I was going to leave it there but realised we had a similar issue in `PerColumnIndexFiles`, so I made the change for per-SSTable and per-index versions of this method.

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
```diff
@@ -251,36 +257,72 @@ public IndexOutputWriter openPerIndexOutput(IndexComponent component, IndexConte
...
+                try
+                {
+                    cleanup.perform();
+                }
+                catch (Exception e)
+                {
+                    throw Throwables.unchecked(Throwables.merge(t, e));
+                }
+            }
+            throw t;
```
maedhroz: nit: I guess you could throw this bit into a helper method so it isn't repeated in both createX methods.

src/java/org/apache/cassandra/index/sai/disk/v1/SkinnyPrimaryKeyMap.java
```diff
@@ -0,0 +1,185 @@
...
+import org.apache.cassandra.utils.bytecomparable.ByteSourceInverse;
+
+import javax.annotation.concurrent.NotThreadSafe;
+import javax.annotation.concurrent.ThreadSafe;
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+
+/**
+ * A row-aware {@link PrimaryKeyMap} for skinny tables (those with no clustering columns).
```
adelapena:
```suggestion
 * A {@link PrimaryKeyMap} for skinny tables (those with no clustering columns).
```

src/java/org/apache/cassandra/index/sai/disk/v1/V1OnDiskFormat.java
```diff
@@ -236,11 +248,14 @@ public Set<IndexComponent> perColumnIndexComponents(IndexContext indexContext)
...
     {
-        // For the V1 format there are always 4 open files per SSTable - token values, primary key trie,
-        // primary key blocks, primary key block offsets
-        return 4;
+        // For the V1 format the number of open files depends on whether the table has clustering. For wide tables
+        // the number of open files will be 6 per SSTable - token values, partition sizes index, partition key blocks,
+        // partition key block offsets, clustering key blocks & clustering key block offsets and for skinny tables
+        // the number of files will be 4 per SSTable - token values, partition key sizes, partition key blocks &
+        // partition key block offsets.
+        return hasClustering ? 6 : 4;
```
adelapena, line 258:
```suggestion
        return hasClustering ? WIDE_PER_SSTABLE_COMPONENTS.size() : SKINNY_PER_SSTABLE_COMPONENTS.size();
```

adelapena: Not directly related to the changes, but I have noticed that `IndexDescriptor#sizeOnDiskOfPerIndexComponents(IndexContext)` a few lines below is unused, we could remove it.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
```diff
@@ -20,38 +20,40 @@
...
 /**
- * Writes an ordered sequence of terms for use with {@link SortedTermsReader}.
+ * Writes a sequence of terms for use with {@link SortedTermsReader}.
  * <p>
- * Terms must be added in lexicographical ascending order.
- * Terms can be of varying lengths.
+ * Terms can be written in two distinct way
+ * <ul>
+ *     <li>
+ *         Non-partitioned - terms can be written in any order
```
adelapena: If the terms can be in any order when non-partitioned, does it still make sense calling these classes `SortedTerms*`?

src/java/org/apache/cassandra/index/sai/disk/v1/V1OnDiskFormat.java
mike-tr-adamson: It's used in `SSTableIndex.sizeOfPerColumnComponents`

mike-tr-adamson: The problem with using the size of the component sets is that not all the components in the sets are opened. Some of them represent readers that use the opened components but have separate metadata.

src/java/org/apache/cassandra/index/sai/disk/v1/sortedterms/SortedTermsWriter.java
mike-tr-adamson: After a discussion on slack we have gone with calling the package `keystore`, with `KeyStoreWriter` and `KeyLookup`.

src/java/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookup.java
```diff
@@ -0,0 +1,377 @@
...
+            this.keysFilePointer = this.keysInput.getFilePointer();
+            this.blockOffsets = new LongArray.DeferredLongArray(blockOffsetsFactory::open);
+            this.currentKey = new BytesRef(keyLookupMeta.maxKeyLength);
+            this.nextBlockKey = new BytesRef(keyLookupMeta.maxKeyLength);
+            keysInput.seek(keysFilePointer);
+            readKey(currentPointId, currentKey);
+        }
+
+        /**
+         * Positions the cursor on the target point id and reads the keyat target to the current key buffer.
```
adelapena:
```suggestion
         * Positions the cursor on the target point id and reads the key at target to the current key buffer.
```

test/unit/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookupTest.java
```diff
@@ -0,0 +1,413 @@
...
+        KeyLookupMeta sortedTermsMeta = new KeyLookupMeta(metadataSource.get(indexDescriptor.componentName(IndexComponent.PARTITION_KEY_BLOCKS)));
+        try (FileHandle termsData = indexDescriptor.createPerSSTableFileHandle(IndexComponent.PARTITION_KEY_BLOCKS, null);
+             FileHandle blockOffsets = indexDescriptor.createPerSSTableFileHandle(IndexComponent.PARTITION_KEY_BLOCK_OFFSETS, null))
+        {
+            KeyLookup reader = new KeyLookup(termsData, blockOffsets, sortedTermsMeta, blockPointersMeta);
+            testCode.accept(reader);
+        }
+    }
+
+    private void withSortedTermsCursor(ThrowingConsumer<KeyLookup.Cursor> testCode) throws IOException
```
adelapena: This could be named `withKeyLookupCursor`

test/unit/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookupTest.java
```diff
@@ -0,0 +1,413 @@
...
+        }
+    }
+
+    @FunctionalInterface
+    public interface ThrowingConsumer<T>
+    {
+        void accept(T t) throws IOException;
+    }
+
+    private void withSortedTermsReader(ThrowingConsumer<KeyLookup> testCode) throws IOException
```
adelapena: This could be named `withKeyLookup`

test/unit/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookupTest.java
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    private byte[] makeTerm(int value)
+    {
+        ByteBuffer buffer = Int32Type.instance.decompose(value);
+        ByteSource byteSource = Int32Type.instance.asComparableBytes(buffer, ByteComparable.Version.OSS50);
+        return ByteSourceInverse.readBytes(byteSource);
+    }
+
+    private void doTestSortedTerms(List<byte[]> terms) throws Exception
```
adelapena: This could be named `doTestKeyLookup`

adelapena (approved):

test/unit/org/apache/cassandra/index/sai/cql/TokenCollisionTest.java
```diff
@@ -0,0 +1,120 @@
...
+    public void skinnyPartitionLastRowTest()
+    {
+        doSkinnyPartitionTest(49, 9);
+    }
+
+    private void doSkinnyPartitionTest(int v1Match, int v2Match)
+    {
+        createTable("CREATE TABLE %s (pk blob, v1 int, v2 int, PRIMARY KEY (pk))");
+        createIndex("CREATE CUSTOM INDEX ON %s(v1) USING 'StorageAttachedIndex'");
+        createIndex("CREATE CUSTOM INDEX ON %s(v2) USING 'StorageAttachedIndex'");
```
maedhroz:
```suggestion
        createIndex("CREATE INDEX ON %s(v2) USING 'sai'");
```

adelapena (approved):
The changes look good to me. We will need a PR for `cassandra-5.0`, which probably will be identical, and CI runs. The CI runs will require running `.circleci/generate.sh -p` again, so they can pick the new and renamed tests for the repeated test runs.

maedhroz:
Committed as https://github.com/apache/cassandra/commit/655a2455ac29395b0a303e6ad7fc4d458b18932d

PR #2546, opened 2023-08-03
CASSANDRA-18722 Support Dynamic Port Allocation for in-jvm dtest framework

frankgh:
Currently, `INodeProvisionStrategy` supports two strategies `OneNetworkInterface` and `MultipleNetworkInterfaces`. However the `seedPort`, `storagePorts`, `nativeTransportPorts`, and `jmxPorts` are always fixed or a function of the node number.

In order to better support parallel test runs, this commit introduces support for dynamic port allocation for the `seedPort`, `storagePorts`, `nativeTransportPorts`, and `jmxPorts`.

When enabled, the port allocation will be dynamic, an available port for the given bind address will be used instead of the previously statically allocated port number. This would allow us to run multiple clusters within the same test, or it will enable us to run in-jvm dtests in parallel given that the tests do not have other inter-test dependencies.

A new option in the cluster builder is introduced `.withDynamicPortAllocation(boolean)`. To enable the new feature one must request dynamic port allocation while building the cluster.

[CASSANDRA-18722](https://issues.apache.org/jira/browse/CASSANDRA-18722)

### DIFF
--- a/test/distributed/org/apache/cassandra/distributed/impl/AbstractCluster.java
+++ b/test/distributed/org/apache/cassandra/distributed/impl/AbstractCluster.java
@@ -155,6 +155,7 @@ public abstract class AbstractCluster<I extends IInstance> implements ICluster<I
     private final Map<Integer, NetworkTopology.DcAndRack> nodeIdTopology;
     private final Consumer<IInstanceConfig> configUpdater;
     private final int broadcastPort;
+    private final Map<String, Integer> portMap;
 
     // mutated by starting/stopping a node
     private final List<I> instances;
@@ -184,6 +185,7 @@ public abstract class AbstractCluster<I extends IInstance> implements ICluster<I
     {
         private INodeProvisionStrategy.Strategy nodeProvisionStrategy = INodeProvisionStrategy.Strategy.MultipleNetworkInterfaces;
         private ShutdownExecutor shutdownExecutor = DEFAULT_SHUTDOWN_EXECUTOR;
+        private boolean dynamicPortAllocation = false;
 
         {
             // Indicate that we are running in the in-jvm dtest environment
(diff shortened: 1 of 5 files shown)


test/unit/org/apache/cassandra/net/SocketUtils.java
```diff
@@ -19,37 +19,52 @@
 package org.apache.cassandra.net;
 
 import java.io.IOException;
+import java.net.InetAddress;
 import java.net.ServerSocket;
+import java.net.UnknownHostException;
 
 public class SocketUtils
 {
-    public static synchronized int findAvailablePort() throws RuntimeException
```
frankgh, line 26: this method was unused

test/distributed/org/apache/cassandra/distributed/impl/AbstractCluster.java
```diff
@@ -199,16 +201,38 @@ public AbstractBuilder(Factory<I, C, B> factory)
...
+         * When {@code dynamicPortAllocation} is {@code true}, it will ask {@link INodeProvisionStrategy} to provision
+         * available storage, native and JMX ports in the given interface. When {@code dynamicPortAllocation} is
+         * {@code false} (the default behavior), it will use statically allocated ports based on the number of
+         * interfaces available and the node number.
+         *
+         * @param dynamicPortAllocation {@code true} for dynamic port allocation, {@code false} for static port
+         *                              allocation
+         * @return a reference to this Builder
+         */
+        public B withDynamicPortAllocation(boolean dynamicPortAllocation)
```
frankgh, line 233: another option is to provide a new interface `PortAllocationStrategy` and we'd have default implementations as well as the `DynamicPortAllocationStrategy`. I didn't go with this approach because I don't foresee a different port allocation strategy other than the default and dynamic in the near future.

dineshjoshi (approved):
+1, thanks.

test/distributed/org/apache/cassandra/distributed/impl/INodeProvisionStrategy.java
```diff
@@ -18,97 +18,174 @@
...
 
+                    @Override
                     public int seedPort()
                     {
+                        if (portMap != null)
+                        {
+                            return portMap.computeIfAbsent("seedPort", key -> SocketUtils.findAvailablePort(seedIp(), 7012));
+                        }
                         return 7012;
                     }
```
jonmeredith: Instead of checking for nodeNum == 1 in `storagePort`, you could just define default methods on the interface for `seedIP()` and `seedPort()`, just calling `ipAddress(1)` and `storagePort(1)` respectively and reduce the amount of methods and make the portMap more regular with entries for `storagePort` for all.

```portMap = {ConcurrentHashMap@16060}  size = 6
 "nativeTransportPort:2" -> {Integer@17131} 63721
 "nativeTransportPort:1" -> {Integer@17107} 63719
 "jmxPort:2" -> {Integer@17133} 63722
 "jmxPort:1" -> {Integer@17109} 63720
 "storagePort:2" -> {Integer@16549} 63709
 "seedPort" -> {Integer@16550} 63708```

**Review by jonmeredith — APPROVED** [review]

File: test/distributed/org/apache/cassandra/distributed/impl/INodeProvisionStrategy.java [hunk 4]
```diff
@@ -18,97 +18,154 @@
...
                     {
                         return ipPrefix + nodeNum;
                     }
 
+                    @Override
                     public int storagePort(int nodeNum)
                     {
+                        if (portMap != null)
+                        {
+                            return portMap.computeIfAbsent("storagePort" + nodeNum, key -> SocketUtils.findAvailablePort(ipAddress(nodeNum), 7012));
```
yifan-c: This key does not contain ":", while all the others contain ":".
Are the keys going to be printed. If so, we can go a bit for verbose to have `node1` instead of `1` and probably `@` suits better than `:`.

yifan-c (approved):
I have one nit. 
+1 on the patch.

frankgh:
Closed via 6ffa43f68b8d10ca84d4a00bf81269527b4e14df

PR #2547, opened 2023-08-03
Cassandra 17684

vgali7:


bschoening (commented):
This should not open the web browser here, instead it needs to setup the URL as Data URL:
help_file = base64.b64encode(file.encode()).decode()
url = 'data:text/plain,str(help_file)

PR #2555, opened 2023-08-07
CASSANDRA-18730 5.0: Add guardrail for vector dimensions

adelapena:


src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -385,6 +385,18 @@ public final class Guardrails implements GuardrailsMBean
...
+    /**
+     * Guardrail on the number of dimensions of vector columns.
+     */
+    public static final MaxThreshold vectorDimensions =
+    new MaxThreshold("vector_dimensions",
+                     null,
+                     state -> CONFIG_PROVIDER.getOrCreate(state).getVectorDimensionsWarnThreshold(),
+                     state -> CONFIG_PROVIDER.getOrCreate(state).getVectorDimensionsFailThreshold(),
+                     (isWarning, what, value, threshold) ->
+                     format("%s has a vector of %s dimensions, this exceeds the %s threshold of %s.",
```
Maxwell-Guo, line 397: what about change the messge to : %s has a vector of %s dimensions, which exceeds the %s threshold of %s.

Maxwell-Guo (approved):

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
adelapena: Assuming it's grammatically correct, I'd prefer to leave it as it is since there are nine other guardrail messages with the same form.

PR #2556, opened 2023-08-08
Cep 29 not operator

pkolaczk:
* NOT IN
* NOT CONTAINS / NOT CONTAINS KEY
* Inequality operator (!=) for map entries
* Inequality operator (!=) for clustering columns
* Both ALLOW FILTERING and indexing (SAI) supported


src/antlr/Parser.g
```diff
@@ -1789,6 +1793,16 @@ relation[WhereClause.Builder clauses]
...
               { $clauses.add(MultiColumnRelation.createInRelation(ids, markers)); }
           )
+      | K_NOT K_IN
+          ( '(' ')'
+              { $clauses.add(MultiColumnRelation.createNotInRelation(ids, new ArrayList<Tuples.Literal>())); }
+          | tupleInMarker=inMarkerForTuple /* (a, b, c) IN ? */
+              { $clauses.add(MultiColumnRelation.createSingleMarkerNotInRelation(ids, tupleInMarker)); }
+          | literals=tupleOfTupleLiterals /* (a, b, c) IN ((1, 2, 3), (4, 5, 6), ...) */
+              { $clauses.add(MultiColumnRelation.createNotInRelation(ids, literals)); }
+          | markers=tupleOfMarkersForTuples /* (a, b, c) IN (?, ?, ...) */
```
adelapena: The three comments should use `NOT IN` instead of only `IN`

src/java/org/apache/cassandra/cql3/MarkerOrList.java
```diff
@@ -0,0 +1,140 @@
...
+import static org.apache.cassandra.cql3.statements.RequestValidations.checkBindValueSet;
+import static org.apache.cassandra.cql3.statements.RequestValidations.checkNotNull;
+
+/**
+ * Represents either a single marker for IN values, or a list of IN values, like:
+ * "SELECT ... WHERE x IN ?"
+ * "SELECT ... WHERE x IN (?, ?, ?)"
+ * "SELECT ... WHERE x IN (a, b, c)"
+ */
+public class MarkerOrList
```
adelapena: Maybe this could be named `MarkerOrTerms`, since without context it's not very clear what kind of list it is. Also, the constructor and most methods are branched depending on whether it's a single marker or a list of terms. This makes me think that the class could benefit from inheritance, more or less this way: https://github.com/adelapena/cassandra/commit/1a2361db14e4f1e2c32fbda4a0a3b8dd8543ea66

src/java/org/apache/cassandra/cql3/MultiColumnRelation.java
```diff
@@ -95,6 +96,11 @@ public static MultiColumnRelation createInRelation(List<ColumnIdentifier> entiti
         return new MultiColumnRelation(entities, Operator.IN, null, inValues, null);
     }
 
+    public static MultiColumnRelation createNotInRelation(List<ColumnIdentifier> entities, List<? extends Term.MultiColumnRaw> inValues)
```
adelapena: This could have JavaDoc like the method `createInRelation` right above.

src/java/org/apache/cassandra/cql3/Relation.java
```diff
@@ -169,6 +214,17 @@ public final Restriction toRestriction(TableMetadata table, VariableSpecificatio
      */
     protected abstract Restriction newEQRestriction(TableMetadata table, VariableSpecifications boundNames);
 
+    /**
+     * Creates a new NEQ restriction instance.
+     *
+     * @param table the table meta data
+     * @param boundNames the variables specification where to collect the bind variables
+     * @return a new EQ restriction instance.
```
adelapena:
```suggestion
     * @return a new NEQ restriction instance.
```

src/java/org/apache/cassandra/cql3/Relation.java
```diff
@@ -203,7 +270,7 @@ protected abstract Restriction newSliceRestriction(TableMetadata table,
      * @return a new Contains <code>Restriction</code> instance
      * @throws InvalidRequestException if the <code>Relation</code> is not valid
      */
-    protected abstract Restriction newContainsRestriction(TableMetadata table, VariableSpecifications boundNames, boolean isKey);
+    protected abstract Restriction newContainsRestriction(TableMetadata table, VariableSpecifications boundNames, boolean isKey, boolean isNot);
```
adelapena: The new `isNot` parameter should probably be documented. However, given that EQ/NEQ and IN/NOT_INT pairs have separate `new*Restriction` methods for the "not" variant, perhaps we should do the same here and have a separate `newNotContains` method instead of the parameter?

src/java/org/apache/cassandra/cql3/restrictions/ClusteringColumnRestrictions.java
```diff
@@ -92,61 +92,46 @@ public ClusteringColumnRestrictions mergeWith(Restriction restriction, @Nullable
...
+        MultiCBuilder builder = MultiCBuilder.create(comparator);
         for (SingleRestriction r : restrictions)
         {
             r.appendTo(builder, options);
 
             if (hasIN() && Guardrails.inSelectCartesianProduct.enabled(state))
                 Guardrails.inSelectCartesianProduct.guard(builder.buildSize(), "clustering key", false, state);
 
-            if (builder.hasMissingElements())
+            if (builder.buildSize() == 0)
```
adelapena: There are four calls to `builder.buildSize() == 0`, maybe we could add an `MultiCBuilder#isEmpty()` method?

src/java/org/apache/cassandra/cql3/restrictions/ClusteringColumnRestrictions.java
```diff
@@ -92,61 +92,46 @@ public ClusteringColumnRestrictions mergeWith(Restriction restriction, @Nullable
...
-
     public NavigableSet<Clustering<?>> valuesAsClustering(QueryOptions options, ClientState state) throws InvalidRequestException
     {
-        MultiCBuilder builder = MultiCBuilder.create(comparator, hasIN());
+        MultiCBuilder builder = MultiCBuilder.create(comparator);
         for (SingleRestriction r : restrictions)
         {
             r.appendTo(builder, options);
 
             if (hasIN() && Guardrails.inSelectCartesianProduct.enabled(state))
```
adelapena, line 102: I understand that the guardrails `partitionKeysInSelect` and `inSelectCartesianProduct` shouldn't consider the new `NOT IN` restrictions. That operator uses filtering and as such doesn't have the same performance issues the guardrails try to protect us from.

src/java/org/apache/cassandra/cql3/restrictions/SingleColumnRestriction.java
```diff
@@ -414,38 +398,56 @@ protected boolean isSupportedBy(Index index)
...
         private final List<Term> values = new ArrayList<>(); // for CONTAINS
-        private final List<Term> keys = new ArrayList<>(); // for CONTAINS_KEY
-        private final List<Term> entryKeys = new ArrayList<>(); // for map[key] = value
+        private final List<Term> negativeValues = new ArrayList<>(); // for NOT_CONTAINS
+        private List<Term> keys = new ArrayList<>(); // for CONTAINS_KEY
+        private final List<Term> negativeKeys = new ArrayList<>(); // for NOT_CONTAINS_KEY
+        private List<Term> entryKeys = new ArrayList<>(); // for map[key] = value
         private final List<Term> entryValues = new ArrayList<>(); // for map[key] = value
+        private List<Term> negativeEntryKeys = new ArrayList<>(); // for map[key] != value
+        private List<Term> negativeEntryValues = new ArrayList<>(); // for map[key] != value
```
adelapena: All can be `final`.

src/java/org/apache/cassandra/cql3/restrictions/SingleColumnRestriction.java
```diff
@@ -414,38 +398,56 @@ protected boolean isSupportedBy(Index index)
...
 
-    // This holds CONTAINS, CONTAINS_KEY, and map[key] = value restrictions because we might want to have any combination of them.
+
+    // This holds CONTAINS, CONTAINS_KEY, NOT CONTAINS, NOT CONTAINS KEY and map[key] = value restrictions because we might want to have any combination of them.
     public static final class ContainsRestriction extends SingleColumnRestriction
     {
         private final List<Term> values = new ArrayList<>(); // for CONTAINS
-        private final List<Term> keys = new ArrayList<>(); // for CONTAINS_KEY
-        private final List<Term> entryKeys = new ArrayList<>(); // for map[key] = value
+        private final List<Term> negativeValues = new ArrayList<>(); // for NOT_CONTAINS
```
adelapena, line 411: Nit: can we move the private constructor at the bottom of the class to the top, so it's together with the other constructors? That way I think it would be easier to understand that there can be instances with both positive and negative term lists (from `doMergeWith`).

src/java/org/apache/cassandra/db/MultiCBuilder.java
```diff
@@ -19,104 +19,155 @@
...
  */
-public abstract class MultiCBuilder
+public class MultiCBuilder
 {
+    /**
+     * Represents a building block of a clustering.
+     * Either a point (single value) or a bound.
+     * Knowing the kind of elements greatly simplifies the process of building clustering bounds.
+     */
+    public static class Element
```
adelapena: Nice utility class! It's not related to the patch but, since we are here, I wonder if we could/should rename the class to something more explicit, like `MultiClusteringBuilder`.

src/java/org/apache/cassandra/db/MultiCBuilder.java
```diff
@@ -19,104 +19,155 @@
...
+        public static Element TOP = new Element(Collections.emptyList(), Kind.INCL_END);
+        public static Element ROOT = new Element(Collections.emptyList(), Kind.POINT);
+
+        final List<ByteBuffer> value;
+        final Kind kind;
+
+
+        private Element(List<ByteBuffer> values, Kind kind)
+        {
+            this.value = values;
```
adelapena: Nit: the attribute and the parameter should probably have the same name

src/java/org/apache/cassandra/db/MultiCBuilder.java
```diff
@@ -19,104 +19,155 @@
...
     /**
      * Creates a new empty {@code MultiCBuilder}.
      */
-    public static MultiCBuilder create(ClusteringComparator comparator, boolean forMultipleValues)
+    public static MultiCBuilder create(ClusteringComparator comparator)
     {
-        return forMultipleValues
-             ? new MultiClusteringBuilder(comparator)
-             : new OneClusteringBuilder(comparator);
+        return new MultiCBuilder(comparator);
```
adelapena: Nit: double blank line after this method

src/java/org/apache/cassandra/db/MultiCBuilder.java
```diff
@@ -126,15 +177,9 @@ protected void checkUpdateable()
      */
     public int remainingCount()
     {
-        return comparator.size() - size;
+        return comparator.size() - columns.size();
```
adelapena, line 250: Nit: double blank line after this method

src/java/org/apache/cassandra/db/MultiCBuilder.java
```diff
@@ -157,358 +202,159 @@ public boolean containsUnset()
...
+     * Checks if some elements can still be added to the clusterings.
      *
-     * @return the clusterings
+     * @return <code>true</code> if it is possible to add more elements to the clusterings, <code>false</code> otherwise.
      */
-    public abstract NavigableSet<Clustering<?>> build();
+    public boolean hasRemaining()
+    {
+        return remainingCount() > 0;
+    }
```
adelapena, line 299: Nit: double blank line after this method

src/java/org/apache/cassandra/index/sai/disk/EmptyIndex.java
```diff
@@ -0,0 +1,109 @@
...
+
+import org.apache.cassandra.db.PartitionPosition;
+import org.apache.cassandra.db.virtual.SimpleDataSet;
+import org.apache.cassandra.dht.AbstractBounds;
+import org.apache.cassandra.index.sai.IndexContext;
+import org.apache.cassandra.index.sai.QueryContext;
+import org.apache.cassandra.index.sai.SSTableContext;
+import org.apache.cassandra.index.sai.iterators.KeyRangeIterator;
+import org.apache.cassandra.index.sai.plan.Expression;
+import org.apache.cassandra.utils.ByteBufferUtil;
```
adelapena: Nit: unused import. Running `ant check` shows nine other unused imports, I guess CI hasn't been run yet for this patch?

test/unit/org/apache/cassandra/index/sai/cql/CollectionIndexingTest.java
```diff
@@ -188,6 +253,7 @@ private void createPopulatedMap(String createIndex)
             put(1, "v1");
             put(2, "v3");
         }});
+
```
adelapena: Nit: unneeded blank line

src/java/org/apache/cassandra/index/sai/IndexContext.java
```diff
@@ -180,6 +181,11 @@ public boolean isNonFrozenCollection()
         return TypeUtil.isNonFrozenCollection(columnMetadata.type);
     }
 
+    public boolean isCollection()
```
adelapena: This new method seems unused

src/java/org/apache/cassandra/index/sai/disk/IndexSearchResultIterator.java
```diff
@@ -61,6 +62,27 @@ public static IndexSearchResultIterator build(Expression expression,
                                                   Collection<SSTableIndex> sstableIndexes,
                                                   AbstractBounds<PartitionPosition> keyRange,
                                                   QueryContext queryContext)
+    {
+        KeyRangeIterator keyIterator = buildKeyIterator(expression, sstableIndexes, keyRange, queryContext);
+
+        // For NOT CONTAINS or NOT CONTAINS KEY it is not enought to just return the primary keys
```
adelapena:
```suggestion
        // For NOT CONTAINS or NOT CONTAINS KEY it is not enough to just return the primary keys
```

src/java/org/apache/cassandra/index/sai/disk/PrimaryKeyMapIterator.java
```diff
@@ -0,0 +1,91 @@
...
+
+import org.apache.cassandra.db.PartitionPosition;
+import org.apache.cassandra.dht.AbstractBounds;
+import org.apache.cassandra.dht.Token;
+import org.apache.cassandra.index.sai.SSTableContext;
+import org.apache.cassandra.index.sai.iterators.KeyRangeIterator;
+import org.apache.cassandra.index.sai.utils.PrimaryKey;
+
+/**
+ * Iterates keys in the PrimaryKeyMap of an SSTable.
```
adelapena:
```suggestion
 * Iterates keys in the {@link PrimaryKeyMap} of a SSTable.
```

src/java/org/apache/cassandra/index/sai/disk/SSTableIndex.java
```diff
@@ -36,7 +37,9 @@
 import org.apache.cassandra.index.sai.IndexContext;
 import org.apache.cassandra.index.sai.QueryContext;
 import org.apache.cassandra.index.sai.SSTableContext;
+import org.apache.cassandra.index.sai.disk.format.IndexDescriptor;
 import org.apache.cassandra.index.sai.disk.format.Version;
+import org.apache.cassandra.index.sai.disk.v1.V1OnDiskFormat;
```
adelapena: Nit: unused imports breaking CI (`ant check`).

src/java/org/apache/cassandra/index/sai/disk/v1/PerColumnIndexFiles.java
```diff
@@ -22,6 +22,7 @@
 import java.util.EnumMap;
 import java.util.Map;
 
+import jnr.ffi.annotations.In;
```
adelapena: Nit: Unneeded change

src/java/org/apache/cassandra/index/sai/disk/v1/RowAwarePrimaryKeyMap.java
```diff
@@ -161,6 +161,11 @@ public long rowIdFromPrimaryKey(PrimaryKey key)
         return triePrefixSearcher.prefixSearch(key.asComparableBytes(ByteComparable.Version.OSS50));
     }
 
+    public long count()
```
adelapena, line 165: Nit: add `@Override`

src/java/org/apache/cassandra/index/sai/disk/v1/V1SSTableIndex.java
```diff
@@ -33,7 +34,10 @@
 import org.apache.cassandra.index.sai.IndexContext;
 import org.apache.cassandra.index.sai.QueryContext;
 import org.apache.cassandra.index.sai.SSTableContext;
+import org.apache.cassandra.index.sai.disk.PrimaryKeyMap;
+import org.apache.cassandra.index.sai.disk.PrimaryKeyMapIterator;
 import org.apache.cassandra.index.sai.disk.SSTableIndex;
+import org.apache.cassandra.index.sai.disk.format.IndexDescriptor;
```
adelapena: Nit: unused imports

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeAntiJoinIterator.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    public static KeyRangeAntiJoinIterator create(KeyRangeIterator left, KeyRangeIterator right)
+    {
+        AntiJoinStatistics statistics = new AntiJoinStatistics();
+        statistics.update(left);
+        return new KeyRangeAntiJoinIterator(statistics, left, right);
+    }
+
+    protected void performSkipTo(PrimaryKey nextKey)
```
adelapena, line 53: Nit: add `@Override`

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeAntiJoinIterator.java
```diff
@@ -0,0 +1,107 @@
...
+    {
+        left.performSkipTo(nextKey);
+        right.performSkipTo(nextKey);
+    }
+
+    public void close() throws IOException
```
adelapena, line 60: Nit: add `@Override`

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeAntiJoinIterator.java
```diff
@@ -0,0 +1,107 @@
...
+    {
+        FileUtils.closeQuietly(left);
+        FileUtils.closeQuietly(right);
+    }
+
+
+    protected PrimaryKey computeNext()
```
adelapena, line 68: Nit: add `@Override`

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeAntiJoinIterator.java
```diff
@@ -0,0 +1,107 @@
...
+            {
+                nextKeyToSkip = right.skipTo(key);
+            }
+            cmp = compare(key, nextKeyToSkip);
+        }
+
+        return key != null ? key : endOfData();
+    }
+
+    private int compare(PrimaryKey key1, PrimaryKey key2)
```
adelapena: Nit: can be `static`

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeAntiJoinIterator.java
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * An iterator wrapper that wraps two iterators (left and right) and returns the primary keys from the left iterator
+ * that do not match the primary keys from the right iterator. The keys returned by the wrapped iterators must
+ * follow token-clustering order.
+ */
+public class KeyRangeAntiJoinIterator extends KeyRangeIterator
+{
+    final KeyRangeIterator left;
+    final KeyRangeIterator right;
```
adelapena: Nit: Can be `private final`

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIterator.java
```diff
@@ -131,6 +131,11 @@ protected final boolean tryToComputeNext()
         return hasNext;
     }
 
+    public PrimaryKey nextOrNull()
```
adelapena: Nit: add `@Nullable`

src/java/org/apache/cassandra/index/sai/memory/MemtableIndex.java
```diff
@@ -30,18 +30,23 @@
 import org.apache.cassandra.index.sai.plan.Expression;
 import org.apache.cassandra.index.sai.utils.PrimaryKeys;
 import org.apache.cassandra.index.sai.iterators.KeyRangeIterator;
+import org.apache.cassandra.index.sai.utils.TypeUtil;
+import org.apache.cassandra.utils.ByteBufferUtil;
```
adelapena: Nit: unused imports

src/java/org/apache/cassandra/index/sai/memory/MemtableIndex.java
```diff
@@ -30,18 +30,23 @@
...
 import org.apache.cassandra.utils.Pair;
 import org.apache.cassandra.utils.bytecomparable.ByteComparable;
 
 public class MemtableIndex
 {
     private final TrieMemoryIndex index;
+
+    private final IndexContext indexContext;
```
adelapena: Unused attribute

src/java/org/apache/cassandra/index/sai/memory/MemtableKeyRangeIterator.java
```diff
@@ -0,0 +1,142 @@
...
+    {
+        PrimaryKey.Factory pkFactory = new PrimaryKey.Factory(memtable.metadata().comparator);
+        return new MemtableKeyRangeIterator(memtable, pkFactory, keyRange);
+    }
+
+    @Override
+    protected void performSkipTo(PrimaryKey nextKey)
+    {
+        AbstractBounds<PartitionPosition> keyRange = AbstractBounds.bounds(nextKey.partitionKey(),
+                                                                true,
```
adelapena: Nit: misaligned argument

src/java/org/apache/cassandra/index/sai/memory/MemtableKeyRangeIterator.java
```diff
@@ -0,0 +1,142 @@
...
+                                                                           this.keyRange.right,
+                                                                           this.keyRange.inclusiveRight());
+        DataRange dataRange = new DataRange(keyRange, new ClusteringIndexSliceFilter(Slices.ALL, false));
+        this.partitionIterator = memtable.partitionIterator(columns, dataRange, null);
```
adelapena:
```suggestion
       partitionIterator = memtable.partitionIterator(columns, dataRange, null);
```

src/java/org/apache/cassandra/index/sai/memory/MemtableKeyRangeIterator.java
```diff
@@ -0,0 +1,142 @@
...
+        if (partitionIterator.hasNext())
+        {
+            this.rowIterator = partitionIterator.next();
+            if (!nextKey.hasEmptyClustering() && rowIterator.partitionKey().equals(nextKey.partitionKey()))
+            {
+                Slice slice = Slice.make(nextKey.clustering(), Clustering.EMPTY);
+                Slices slices = Slices.with(this.memtable.metadata().comparator, slice);
+                this.rowIterator = memtable.rowIterator(nextKey.partitionKey(), slices, this.columns, false, null);
```
adelapena:
```suggestion
                Slices slices = Slices.with(memtable.metadata().comparator, slice);
                rowIterator = memtable.rowIterator(nextKey.partitionKey(), slices, columns, false, null);
```

src/java/org/apache/cassandra/index/sai/memory/MemtableKeyRangeIterator.java
```diff
@@ -0,0 +1,142 @@
...
+public class MemtableKeyRangeIterator extends KeyRangeIterator
+{
+    private final Memtable memtable;
+    private final PrimaryKey.Factory pkFactory;
+    private final AbstractBounds<PartitionPosition> keyRange;
+    private final ColumnFilter columns;
+    private UnfilteredPartitionIterator partitionIterator;
+    private UnfilteredRowIterator rowIterator;
+
+
```
adelapena:
```suggestion

```

src/java/org/apache/cassandra/index/sai/memory/TrieMemoryIndex.java
```diff
@@ -80,6 +80,8 @@ public TrieMemoryIndex(IndexContext indexContext)
         this.analyzerFactory = indexContext.getAnalyzerFactory();
         this.validator = indexContext.getValidator();
         this.isLiteral = TypeUtil.isLiteral(validator);
+
+
```
adelapena: Unneeded change

src/java/org/apache/cassandra/index/sai/plan/Operation.java
```diff
@@ -41,7 +41,8 @@ public class Operation
 {
     public enum BooleanOperator
     {
-        AND((a, b) -> a & b);
+        AND((a, b) -> a & b),
+        AND_NOT((a, b) -> a & !b);
```
adelapena: `AND_NOT` is not used anywhere.

src/java/org/apache/cassandra/utils/btree/BTree.java
```diff
@@ -347,8 +347,8 @@ public static <Compare, Existing extends Compare, Insert extends Compare> Object
 
         if (isEmpty(toUpdate))
         {
-            if (isSimple(updateF))
-                return insert; // if update is empty and updateF is trivial, return our new input
+//            if (isSimple(updateF))
+//                return insert; // if update is empty and updateF is trivial, return our new input
```
adelapena: Can this be removed?

src/java/org/apache/cassandra/cql3/restrictions/TermSlice.java
```diff
@@ -28,6 +28,12 @@
 
 final class TermSlice
 {
+    /**
```
adelapena, line 31: `TermSlice#isSupportedBy(ColumnMetadata, Index)` seems to reject `NEQ` expressions. Thus, if we have:
```
CREATE TABLE t (k int PRIMARY KEY, v int);
CREATE INDEX ON t(v) USING 'sai';
```
The query `SELECT * FROM t WHERE v!=0;` will require `ALLOW FILTERING`, whereas for example the query `SELECT * FROM t WHERE v>0;` will be accepted without AF. I think both queries should be accepted without AF, shouldn't they?

We should probably include the new operators in `AllowFilteringTest`.

src/java/org/apache/cassandra/utils/btree/BTree.java
pkolaczk: Quite likely not. I just accidentally commeted it out during testing. Good catch!

src/java/org/apache/cassandra/index/sai/plan/Operation.java
pkolaczk: Yeah, it turned out not needed. I'll remove.

src/antlr/Parser.g
pkolaczk: Fixed

src/java/org/apache/cassandra/cql3/MarkerOrList.java
pkolaczk: I'm not a very big fan of inheritance / subclassing but indeed in this case it is cleaner. Thank you. I merged your commit.

src/java/org/apache/cassandra/cql3/MultiColumnRelation.java
pkolaczk: Fixed

src/java/org/apache/cassandra/cql3/Relation.java
pkolaczk: I added `newNotContainsRestriction` and removed the `isNot` param.

src/java/org/apache/cassandra/cql3/restrictions/ClusteringColumnRestrictions.java
pkolaczk: Added

src/java/org/apache/cassandra/cql3/restrictions/ClusteringColumnRestrictions.java
pkolaczk: `NOT IN` creates slice restrictions not `IN` resrictions. Slice restrictions on different columns cannot be combined with a cartesian multiplication. There is only one slice restriction allowed, and the slice restriction must be present on the last restricted clustering column. 

I.e you cannot do this:
`SELECT * FROM x WHERE clustering_col_1 NOT IN (...) AND clustering_column_2 NOT IN (...)`

`NOT IN` behaves a lot more like `<, > etc` in this regard. 

However, now I noticed that `IN` can be combined with one `NOT IN`, so `NOT IN` could blow up the build size here because `NOT IN` can specify multiple values. But It looks like it is already covered - the `hasIn` check would be true in that case.

src/java/org/apache/cassandra/cql3/restrictions/SingleColumnRestriction.java
pkolaczk: Applied.

src/java/org/apache/cassandra/cql3/restrictions/SingleColumnRestriction.java
pkolaczk: Indeed. Applied.

src/java/org/apache/cassandra/db/MultiCBuilder.java
pkolaczk: I agree. I like `MultiClusteringBuilder` name much more than `MultiCBuilder`.

src/java/org/apache/cassandra/db/MultiCBuilder.java
pkolaczk: Fixed. Renamed field `value` to `values`. I also renamed `Element` to `ClusteringElements` as it can contain values for more than one column.

src/java/org/apache/cassandra/db/MultiCBuilder.java
pkolaczk: Done

src/java/org/apache/cassandra/index/sai/disk/EmptyIndex.java
pkolaczk: Fixed all unused imports. `ant check` passes now.

src/java/org/apache/cassandra/db/MultiCBuilder.java
pkolaczk: fixed

src/java/org/apache/cassandra/db/MultiCBuilder.java
pkolaczk: fixed

src/java/org/apache/cassandra/db/MultiCBuilder.java
pkolaczk: fixed

src/java/org/apache/cassandra/index/sai/IndexContext.java
pkolaczk: Removed

test/unit/org/apache/cassandra/index/sai/cql/CollectionIndexingTest.java
pkolaczk: fixed

src/java/org/apache/cassandra/index/sai/memory/TrieMemoryIndex.java
pkolaczk: fixed

src/java/org/apache/cassandra/cql3/restrictions/TermSlice.java
pkolaczk: Indeed, both should be accepted.

adelapena: It seems that `SELECT * FROM t WHERE v!=0` still requires `ALLOW FILTERING` after the last changes. The new check on `AllowFilteringTest#testAllowFilteringOnFirstClusteringKeyColumn` makes the test fail.

Also, `AllowFilteringTest` has test methods for different schemas. We should probably add the `NEQ` check to all the test methods, not only in `testAllowFilteringOnFirstClusteringKeyColumn`.

src/java/org/apache/cassandra/db/MultiClusteringBuilder.java
```diff
@@ -0,0 +1,437 @@
...
+
+import org.apache.cassandra.cql3.statements.Bound;
+import org.apache.cassandra.db.marshal.AbstractType;
+import org.apache.cassandra.schema.ColumnMetadata;
+import org.apache.cassandra.utils.ByteBufferUtil;
+import org.apache.cassandra.utils.UniqueComparator;
+import org.apache.cassandra.utils.btree.BTreeSet;
+
+/**
+ * Builder that allows to build multiple {@link Clustering}/{@link ClusteringBound} at the same time.
```
adelapena, line 34: Nice doc!

src/java/org/apache/cassandra/db/MultiClusteringBuilder.java
```diff
@@ -0,0 +1,437 @@
...
+ * point: (a1, b1)
+ * point: (a1, b2)
+ * point: (a1, b3)
+ * point: (a2, b1)
+ * point: (a2, b2)
+ * point: (a2, b3)
+ * </pre>
+ *
+ * Finally, we add the slice of column {@code c} by specifying the lower and upper bound
+ * (we use {@code TOP} for the upper bound), and we get the final set of clustering bounds:
```
adelapena:
```suggestion
 * (we use {@link ClusteringElements#TOP} for the upper bound), and we get the final set of clustering bounds:
```
Shouldn't we include `TOP` in the inclusive ends below?

src/java/org/apache/cassandra/cql3/restrictions/SingleRestriction.java
```diff
@@ -19,7 +19,7 @@
 
 import org.apache.cassandra.cql3.QueryOptions;
 import org.apache.cassandra.cql3.statements.Bound;
-import org.apache.cassandra.db.MultiCBuilder;
+import org.apache.cassandra.db.MultiClusteringBuilder;
```
adelapena, line 22: Nit: The JavaDoc of some methods on this class still has some mentions of `<code>MultiCBuilder</code>` that should be updated to `<code>MultiClusteringBuilder</code>`. Or, even better, `{@link MultiClusteringBuilder}`.

src/java/org/apache/cassandra/db/MultiClusteringBuilder.java
```diff
@@ -0,0 +1,437 @@
...
+    private boolean containsSliceBound;
+
+
+    private MultiClusteringBuilder(ClusteringComparator comparator)
+    {
+        this.comparator = comparator;
+    }
+
+    /**
+     * Creates a new empty {@code MultiCBuilder}.
```
adelapena, line 228:
```suggestion
     * Creates a new empty {@code MultiClusteringBuilder}.
```

src/java/org/apache/cassandra/db/MultiClusteringBuilder.java
```diff
@@ -0,0 +1,437 @@
...
+ * incl end:   (a1, b3)
+ * excl start: (a2, b1, c1)
+ * incl end:   (a2, b1)
+ * excl start: (a2, b2, c1)
+ * incl end:   (a2, b2)
+ * excl start: (a2, b3, c1)
+ * incl end:   (a2, b3)
+ * </pre>
+ */
+public class MultiClusteringBuilder
```
adelapena, line 103: Nit: There are a couple of mentions of `MultiCBuilder` in `InsertInvalidateSizedRecordsTest`.

src/java/org/apache/cassandra/cql3/MultiColumnRelation.java
```diff
@@ -95,6 +96,17 @@ public static MultiColumnRelation createInRelation(List<ColumnIdentifier> entiti
         return new MultiColumnRelation(entities, Operator.IN, null, inValues, null);
     }
 
+    /**
+     * Creates a multi-column NOT IN relation with a list of NOT IN values or markers.
+     * For example: "SELECT ... WHERE (a, b) NOT IN ((0, 1), (2, 3))"
```
adelapena:
```suggestion
     * For example: {code SELECT ... WHERE (a, b) NOT IN ((0, 1), (2, 3))}
     *
```

src/java/org/apache/cassandra/cql3/MultiColumnRelation.java
```diff
@@ -107,6 +119,18 @@ public static MultiColumnRelation createSingleMarkerInRelation(List<ColumnIdenti
         return new MultiColumnRelation(entities, Operator.IN, null, null, inMarker);
     }
 
+    /**
+     * Creates a multi-column NOT IN relation with a marker for the NOT IN values.
+     * For example: "SELECT ... WHERE (a, b) NOT IN ?"
```
adelapena:
```suggestion
     * For example: {@code SELECT ... WHERE (a, b) NOT IN ?}
     *
```

src/java/org/apache/cassandra/cql3/restrictions/TermSlice.java
pkolaczk: Indeed, both should be accepted. Fixed.

src/java/org/apache/cassandra/db/MultiClusteringBuilder.java
pkolaczk: Thank you, I hope it will make it easier for developers coming after me. I remember it took me a while until I understood what this component is for.

src/java/org/apache/cassandra/cql3/restrictions/SingleRestriction.java
pkolaczk: Renamed everywhere and created links.

src/java/org/apache/cassandra/db/MultiClusteringBuilder.java
pkolaczk: Fixed

src/java/org/apache/cassandra/cql3/restrictions/SingleColumnRestriction.java
```diff
@@ -403,49 +381,80 @@ public void addToRowFilter(RowFilter filter, IndexRegistry indexRegistry, QueryO
...
+            }
         }
 
         @Override
         protected boolean isSupportedBy(Index index)
         {
-            return slice.isSupportedBy(columnDef, index);
+            boolean supportsSlice = slice.isSupportedBy(columnDef, index);
+            boolean supportsNeq = index.supportsExpression(columnDef, Operator.NEQ);
+            return  supportsSlice || !skippedValues.isEmpty() && supportsNeq;
```
adelapena: The implementations of `SingleColumnRestriction.SliceRestriction#isSupportedBy` and `MultiColumnRestriction.SliceRestriction#isSupportedBy` are identical. Also, both implementations are the only callers of `TermSlice#isSupportedBy(ColumnMetadata, Index)`. 

Maybe we could pass the list of skipped values to `TermSlice#isSupportedBy` and delegate the NEQ check to it, putting similar things together and preventing some duplication:
```java
public boolean isSupportedBy(ColumnMetadata column, Index index, List<MarkerOrTerms> skippedValues)
{
    boolean supported = false;

    if (hasBound(Bound.START))
        supported |= isInclusive(Bound.START) ? index.supportsExpression(column, Operator.GTE)
                : index.supportsExpression(column, Operator.GT);
    if (hasBound(Bound.END))
        supported |= isInclusive(Bound.END) ? index.supportsExpression(column, Operator.LTE)
                : index.supportsExpression(column, Operator.LT);

    supported |= !skippedValues.isEmpty() && index.supportsExpression(column, Operator.NEQ);

    return supported;
}
```

adelapena:
@pkolaczk there is a bunch of unresolved nits above that only show up if you click on the `12 hidden conversations Load more...` link on the GH web UI, maybe you have missed them?

src/java/org/apache/cassandra/cql3/MultiColumnRelation.java
```diff
@@ -95,6 +96,18 @@ public static MultiColumnRelation createInRelation(List<ColumnIdentifier> entiti
...
     }
 
+    /**
+     * Creates a multi-column NOT IN relation with a list of NOT IN values or markers.
+     * For example: {code SELECT ... WHERE (a, b) NOT IN ((0, 1), (2, 3))}
+     *
+     * @param entities the columns on the LHS of the relation
+     * @param inValues a list of Tuples.Literal instances or a Tuples.Raw markers
+     * @return a new <code>MultiColumnRelation</code> instance
+     */public static MultiColumnRelation createNotInRelation(List<ColumnIdentifier> entities, List<? extends Term.MultiColumnRaw> inValues)
```
adelapena: Nit: missing line break
```suggestion
     */
     public static MultiColumnRelation createNotInRelation(List<ColumnIdentifier> entities, List<? extends Term.MultiColumnRaw> inValues)
```

adelapena:
I have noticed that this test returns more rows than expected:
```java
@Test
public void notInTest()
{
    createTable("CREATE TABLE %s (k int PRIMARY KEY, v1 int, v2 int)");
    createIndex("CREATE INDEX ON %s(v1) USING 'sai'");
    createIndex("CREATE INDEX ON %s(v2) USING 'sai'");

    execute("INSERT INTO %s (k, v1, v2) VALUES (1, 1, 4)");
    execute("INSERT INTO %s (k, v1, v2) VALUES (2, 2, 3)");
    execute("INSERT INTO %s (k, v1, v2) VALUES (3, 3, 2)");
    execute("INSERT INTO %s (k, v1, v2) VALUES (4, 4, 1)");

    waitForTableIndexesQueryable(currentTable());

    assertRowsIgnoringOrder(execute("SELECT * FROM %s"),
            row(1, 1, 4),
            row(2, 2, 3),
            row(3, 3, 2),
            row(4, 4, 1));

    assertRowsIgnoringOrder(execute("SELECT * FROM %s WHERE v1 NOT IN (1)"),
            row(2, 2, 3),
            row(3, 3, 2),
            row(4, 4, 1));

    assertRowsIgnoringOrder(execute("SELECT * FROM %s WHERE v1 NOT IN (4)"),
            row(1, 1, 4),
            row(2, 2, 3),
            row(3, 3, 2));

    assertRowsIgnoringOrder(execute("SELECT * FROM %s WHERE v1 NOT IN (1, 4)"),
            row(2, 2, 3),
            row(3, 3, 2)); // Got 1 extra row(s) in result: (k=1, v1=1, v2=4)

    assertRowsIgnoringOrder(execute("SELECT * FROM %s WHERE v1 NOT IN (1, 4) AND v2 NOT IN (3, 2)")); // 2 extra rows
}
```
However, the test passes if we don't create the indexes and use `ALLOW FILTERING` instead:
```java
@Test
public void notInTest()
{
    createTable("CREATE TABLE %s (k int PRIMARY KEY, v1 int, v2 int)");

    execute("INSERT INTO %s (k, v1, v2) VALUES (1, 1, 4)");
    execute("INSERT INTO %s (k, v1, v2) VALUES (2, 2, 3)");
    execute("INSERT INTO %s (k, v1, v2) VALUES (3, 3, 2)");
    execute("INSERT INTO %s (k, v1, v2) VALUES (4, 4, 1)");

    assertRowsIgnoringOrder(execute("SELECT * FROM %s"),
            row(1, 1, 4),
            row(2, 2, 3),
            row(3, 3, 2),
            row(4, 4, 1));

    assertRowsIgnoringOrder(execute("SELECT * FROM %s WHERE v1 NOT IN (1) ALLOW FILTERING"),
            row(2, 2, 3),
            row(3, 3, 2),
            row(4, 4, 1));

    assertRowsIgnoringOrder(execute("SELECT * FROM %s WHERE v1 NOT IN (4) ALLOW FILTERING"),
            row(1, 1, 4),
            row(2, 2, 3),
            row(3, 3, 2));

    assertRowsIgnoringOrder(execute("SELECT * FROM %s WHERE v1 NOT IN (1, 4) ALLOW FILTERING"),
            row(2, 2, 3),
            row(3, 3, 2));

    assertRowsIgnoringOrder(execute("SELECT * FROM %s WHERE v1 NOT IN (1, 4) AND v2 NOT IN (3, 2) ALLOW FILTERING"));
}
```

pkolaczk:
> there is a bunch of unresolved nits above that only show up if you click on the 12 hidden conversations Load more... link on the GH web UI, maybe you have missed them?

lol, indeed, I missed them.

src/java/org/apache/cassandra/cql3/restrictions/SingleColumnRestriction.java
pkolaczk: I don't agree on this one. This is just coincidental duplication - accidentally both single and multi restrictions work the same way here, but there is no fundamental reason they have to. The amount of duplication is also very tiny. And introducing an unrelated parameter to `TermSlice` is mixing responsibilities which is IMHO worse.

I thought that maybe we could move the skipped values to the `TermSlice` itself. Then it would make more sense to have `TermSlice` handle the `isSupportedBy` logic as it would have all the information it needs. WDYT?

src/java/org/apache/cassandra/index/sai/disk/v1/RowAwarePrimaryKeyMap.java
pkolaczk: fixed

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeAntiJoinIterator.java
pkolaczk: fixed

src/java/org/apache/cassandra/index/sai/memory/MemtableKeyRangeIterator.java
pkolaczk: lol, the inlay hint from Idea fooled me on this one. I stared at the code for good 10 seconds thinking wtf, they *are* aligned, and then noticed the hint that was present only at this line but not others ;)

src/java/org/apache/cassandra/db/MultiClusteringBuilder.java
adelapena: I think this nit hasn't actually been resolved, it still says `{@code MultiCBuilder}`.

src/java/org/apache/cassandra/cql3/restrictions/SingleColumnRestriction.java
adelapena: Agree, moving the skipped values to the `TermSlice` sounds good to me.

pkolaczk:
I reverted the support for `!=` and `NOT IN` on indexed columns for now as current index search does not allow multiple exclusions.

adelapena:
Albeit the reversal of the support for `!=` and `NOT IN` on indexed columns has made it necessary to use `ALLOW FILTERING` again, it seems that the indexes are still involved in the query and producing wrong results:
```java
@Test
public void notInTest()
{
    createTable("CREATE TABLE %s (k int PRIMARY KEY, v int)");

    execute("INSERT INTO %s (k, v) VALUES (1, 1)");
    execute("INSERT INTO %s (k, v) VALUES (2, 2)");
    execute("INSERT INTO %s (k, v) VALUES (3, 3)");
    execute("INSERT INTO %s (k, v) VALUES (4, 4)");

    String query = "SELECT * FROM %s WHERE v NOT IN (1, 4) ALLOW FILTERING";
    Object[][] rows = rows(row(2, 2), row(3, 3));

    assertRowsIgnoringOrder(execute(query), rows);

    createIndex("CREATE INDEX ON %s(v) USING 'sai'");
    waitForTableIndexesQueryable();

    assertRowsIgnoringOrder(execute(query), rows); // Got 1 extra row(s) in result: (k=1, v1=1)
}
```
I suggest we add some additional tests around the use of unsupported operators with SAI indexes and `ALLOW FILTERING`.

src/java/org/apache/cassandra/db/CBuilder.java
```diff
@@ -28,9 +28,9 @@
 /**
  * Allows to build ClusteringPrefixes, either Clustering or ClusteringBound.
  */
-public abstract class CBuilder
```
Mmuzaf, line 31: Hi, I prefer to make these sorts of changes in a dedicated issue, so it looks like we can significantly reduce the number of changed files if we revert it, and keep only necessary changes here.

See the discussion below, we can ignore code formatting changes by listing the relevant commit hashes in a special file (this is not used by the project right now). So if we do this change in another issue and put that commit hash in a special file, it won't pollute the git blame in our IDEs.
https://github.com/orgs/community/discussions/5033
https://git-scm.com/docs/git-blame#Documentation/git-blame.txt-blameignoreRevsFile

This is my humble argument for such not making such a change here :-)

src/java/org/apache/cassandra/index/sai/disk/PrimaryKeyMapIterator.java
```diff
@@ -0,0 +1,91 @@
...
+
+/**
+ * Iterates keys in the {@link PrimaryKeyMap} of a SSTable.
+ * Iterating keys in the primary key map is faster than reading them from the SSTable data component
+ * because we don't deserialize all the other columns except the primary key columns.
+ * The primary key map components are also likely much smaller than the whole SSTable data component.
+ * <p>
+ * The keys are returned in token-clustering order.
+ */
+public final class PrimaryKeyMapIterator extends KeyRangeIterator
```
Mmuzaf, line 38: The iterator looks very similar to this one, do we still need to keep it?

```
    private SSTableRowIdKeyRangeIterator(PrimaryKey min,
                                         PrimaryKey max,
                                         long count,
                                         PrimaryKeyMap primaryKeyMap,
                                         QueryContext queryContext,
                                         PostingList postingList)
```

PR #2561, opened 2023-08-09
Cassandra 17667 

vgali7:


pylib/cqlshlib/cqlshmain.py
```diff
@@ -799,29 +798,11 @@ def cmdloop(self, intro=None):
                     self.reset_statement()
                     print('')
```
fruch, line 800: I'm not sure this is the best approach

This drops the feature of being able to use comments and inside CQL statements


I think it's an important enough feature

PR #2563, opened 2023-08-09
Fix for CASSANDRA-18688: Limit Java runtime in 5.0 to JDK17 in scripts; add a flag to opt out of that

shyla226:
Limits Cassandra 5.0 startup to JDK11 & JDK17. An option -J has been added to bypass this limit


bin/cassandra
```diff
@@ -276,6 +280,19 @@ if [ "x$allow_root" != "xyes" ] ; then
...
+if  [ "$short" \< "11" ];then
+    echo "Cassandra 5.0 requires Java 11 or Java 17."
+    exit 1;
+fi
+
+java_versions_supported=11,17
+supported=$(echo "$java_versions_supported" | tr "," '\n' | grep -F -x "$short")
+if [ "x$allow_experimental_java_ver" != "xyes" ] && [ "x$supported" = "x" ] ; then
+    echo "Experimental Java version not allowed: Cassandra 5.0 requires Java 11 or Java 17."
+    echo "If you would like to test with this version, use -J command line option"
```
ekaterinadimitrova2:
```suggestion
    echo "Cassandra 5.0 requires Java 11 or Java 17."
    echo "If you would like to test running Cassandra on non-LTS post JDK11 version, you can start it by adding -J command line option"
```

bin/cassandra.in.sh
```diff
@@ -125,9 +125,6 @@ short=$(echo "${jvmver}" | cut -c1-2)
 JAVA_VERSION=17
 if [ "$short" = "11" ]  ; then
      JAVA_VERSION=11
-elif [ "$JVM_VERSION" \< "17" ] ; then
-    echo "Cassandra 5.0 requires Java 11 or Java 17."
-    exit 1;
```
ekaterinadimitrova2: The side effect of this removal is that for any version between Java 11 and 17, JAVA_VERSION will be set to 17, no?
Then the jvm options files for Java 17 will be used for versions pre-17. In my opinion we should use the Java 11 files for those JDK versions and the J17 files for any >= J17 version. Look here - https://github.com/apache/cassandra/blob/trunk/bin/cassandra.in.sh#L152-L158

ekaterinadimitrova2 (commented):
Overall LGTM, just two small commens for your consideration

ekaterinadimitrova2:
I almost forgot, we need similar changes for tools/bin/cassandra.in.sh

bin/cassandra.in.sh
bereng: +1 use j11 options

bin/cassandra
```diff
@@ -276,6 +280,19 @@ if [ "x$allow_root" != "xyes" ] ; then
     fi
 fi
 
+if  [ "$short" \< "11" ];then
+    echo "Cassandra 5.0 requires Java 11 or Java 17."
+    exit 1;
+fi
+
+java_versions_supported=11,17
```
bereng: Wouldn't we like to have this in `cassandra.in.sh`? Seem like a generally useful bit of info that could come handy in the future for other scripts.

bin/cassandra.in.sh
michaelsembwever: yes, if JVM_VERSION=16 then it should be JAVA_VERSION=16 so that the jvm11*.options file will be then used.

bin/cassandra.in.sh
```diff
@@ -125,9 +125,6 @@ short=$(echo "${jvmver}" | cut -c1-2)
 JAVA_VERSION=17
 if [ "$short" = "11" ]  ; then
      JAVA_VERSION=11
-elif [ "$JVM_VERSION" \< "17" ] ; then
```
michaelsembwever: are there any existing checks/conditions for when JVM_VERSION is greater than 17 ? 

instead of introducing a flag, i'd be ok with 
- preventing JDKs 12-16
- printing a loud warning when 18+ is used

I'm ~ok when extending the latter to failure, and using the `-J` to make it work (with the loud warning)

bin/cassandra.in.sh
```diff
@@ -122,13 +122,22 @@ jvmver=`echo "$java_ver_output" | grep '[openjdk|java] version' | awk -F'"' 'NR=
...
 elif [ "$JVM_VERSION" \< "17" ] ; then
     echo "Cassandra 5.0 requires Java 11 or Java 17."
     exit 1;
 fi
+# Allow execution if a supported Java version is used or CASSANDRA_USE_ALL_JDK argument is set
+java_versions_supported=11,17
+supported_version=$(echo "$java_versions_supported" | tr "," '\n' | grep -F -x "$short")
+
+if [ "x$CASSANDRA_USE_ALL_JDK" != "xtrue" ] && [ "x$supported_version" = "x" ] ; then
+    echo "Cassandra 5.0 requires Java 11, Java 17 (or newer LTS)."
```
michaelsembwever:
```suggestion
    echo "Cassandra 5.0 requires Java 11 or Java 17."
```


comment applies to all places.

bin/cassandra.in.sh
```diff
@@ -122,13 +122,22 @@ jvmver=`echo "$java_ver_output" | grep '[openjdk|java] version' | awk -F'"' 'NR=
...
      JAVA_VERSION=11
 elif [ "$JVM_VERSION" \< "17" ] ; then
     echo "Cassandra 5.0 requires Java 11 or Java 17."
     exit 1;
 fi
+# Allow execution if a supported Java version is used or CASSANDRA_USE_ALL_JDK argument is set
+java_versions_supported=11,17
+supported_version=$(echo "$java_versions_supported" | tr "," '\n' | grep -F -x "$short")
+
+if [ "x$CASSANDRA_USE_ALL_JDK" != "xtrue" ] && [ "x$supported_version" = "x" ] ; then
```
michaelsembwever:
```suggestion
if [ "x$CASSANDRA_USE_ALL_JDK" != "x" ] && [ "x$supported_version" = "x" ] ; then
```

just take any non-blank value as "the env variable is activated".  better to catch all the possible legit values than worry about the using trying to get "=false" working.

comment applies to all places.

michaelsembwever (changes requested):
Don't forget this file also exists (patch or whole) in

- `debian/patches/cassandra_in.sh_dirs.diff` 
- `redhat/cassandra.in.sh`

tools/bin/cassandra.in.sh
```diff
@@ -82,13 +82,24 @@ jvmver=`echo "$java_ver_output" | grep '[openjdk|java] version' | awk -F'"' 'NR=
 JVM_VERSION=${jvmver%_*}
 short=$(echo "${jvmver}" | cut -c1-2)
 
-JAVA_VERSION=17
 if [ "$short" = "11" ] ; then
      JAVA_VERSION=11
 elif [ "$JVM_VERSION" \< "17" ] ; then
     echo "Cassandra 5.0 requires Java 11 or Java 17(or newer)."
```
smiklosovic: there should be space between number `17` and `(or newer` like this: `Java 17 (or newer).`

tools/bin/cassandra.in.sh
```diff
@@ -82,13 +82,24 @@ jvmver=`echo "$java_ver_output" | grep '[openjdk|java] version' | awk -F'"' 'NR=
...
+    exit
 fi
 
+# Allow execution if a supported Java version is used or CASSANDRA_USE_ALL_JDK argument is set
+java_versions_supported=11,17
+supported_version=$(echo "$java_versions_supported" | tr "," '\n' | grep -F -x "$short")
+
+if [ "x$CASSANDRA_USE_ALL_JDK" != "xtrue" ] && [ "x$supported_version" = "x" ] ; then
+    echo "Cassandra 5.0 requires Java 11, Java 17 (or newer LTS)."
+    echo "If you would like to test with other Java versions >17, set \$CASSANDRA_USE_ALL_JDK=true"
```
smiklosovic: there should be space between `Java versions >` and `17` like this: `Java versions > 17`.

bin/cassandra.in.sh
```diff
@@ -122,13 +122,22 @@ jvmver=`echo "$java_ver_output" | grep '[openjdk|java] version' | awk -F'"' 'NR=
...
+    echo "Cassandra 5.0 requires Java 11, Java 17 (or newer LTS)."
+    echo "If you would like to test with other Java versions >17, set \$CASSANDRA_USE_ALL_JDK=true"
```
smiklosovic: there should be space between `Java versions` > and `17` like this: `Java versions > 17`.

shyla226: Added

tools/bin/cassandra.in.sh
shyla226: added

bin/cassandra.in.sh
shyla226: - I have added the changes to `redhat/cassandra.in.sh`. 
- `debian/patches/cassandra_in.sh_dirs.diff` does not check for Java version. Should I add?
How do I test these?

redhat/cassandra.in.sh
```diff
@@ -80,13 +80,22 @@ jvmver=`echo "$java_ver_output" | grep '[openjdk|java] version' | awk -F'"' 'NR=
...
+# Allow execution if a supported Java version is used or CASSANDRA_USE_ALL_JDK argument is set
+java_versions_supported=11,17
+supported_version=$(echo "$java_versions_supported" | tr "," '\n' | grep -F -x "$short")
+
+if [ "x$CASSANDRA_USE_ALL_JDK" != "x" ] && [ "x$supported_version" = "x" ] ; then
+    echo "Cassandra 5.0 requires Java 11, Java 17."
+    echo "If you would like to test with other Java versions > 17, set \$CASSANDRA_USE_ALL_JDK=true"
+    exit 1;
+fi
+JAVA_VERSION=$short
```
ekaterinadimitrova2, line 104: Something has shifted, as I can now start Cassandra with JDK20 and JDK21 without using the new variable. @shyla226, can you, please, check?
Also, in my understanding, @michaelsembwever was asking for a warning even if we use the variable - just a print message: "You are using JDKx. Cassandra is currently tested with 11 and 17 by the community. Use version x at your own risk" or something like that. I am on a new computer, and there is no jdk arm64 for macos, no non-LTS between 11 and 17. I hope someone else could test that.

bin/cassandra.in.sh
ekaterinadimitrova2: I believe we agreed not to allow any versions between 11 and 17 at all on the ticket so I am resolving this conversation.

bin/cassandra.in.sh
ekaterinadimitrova2: @michaelsembwever, @smiklosovic,  please correct me if I am wrong, but I think after https://github.com/apache/cassandra/commit/e76b63d08146a3918b8315b9f0a11a04dca66da7, we allow only 11 and 17 with Debian packages? 
Also, regarding testing, I think we need to build the artifacts and then try to run cassandra?

bin/cassandra.in.sh
```diff
@@ -122,13 +122,22 @@ jvmver=`echo "$java_ver_output" | grep '[openjdk|java] version' | awk -F'"' 'NR=
...
      JAVA_VERSION=11
 elif [ "$JVM_VERSION" \< "17" ] ; then
     echo "Cassandra 5.0 requires Java 11 or Java 17."
     exit 1;
 fi
+# Allow execution if a supported Java version is used or CASSANDRA_USE_ALL_JDK argument is set
+java_versions_supported=11,17
+supported_version=$(echo "$java_versions_supported" | tr "," '\n' | grep -F -x "$short")
+
+if [ "x$CASSANDRA_USE_ALL_JDK" != "x" ] && [ "x$supported_version" = "x" ] ; then
```
ekaterinadimitrova2:
```suggestion
if [ -z "$CASSANDRA_USE_ALL_JDK" ] && [ -z "$supported_version" ] ; then
```
My testing shows that something like that should fix your startup problem. Tested with JDK11 and 17 - C* starts; JDK20 and 21 prints correctly the below message. 
So for the suggested warning, we need to add a print when the CASSANDRA_USE_ALL_JDK is set.

ekaterinadimitrova2 (commented):
Left a few comments, it seems something shifted and I can start Cassandra with JDK 20 and 21 
I made a suggestion for a fix on one of the places where you check CASSANDRA_USE_ALL_JDK, applies to all places where we are handling it.

tools/bin/cassandra.in.sh
shyla226: Removed (or newer) to make it consistent with other messages

bin/cassandra.in.sh
michaelsembwever: Both debian and redhat packages require jdk11 or jdk17 dependencies.

But it's easy to get around package deps, so enforcing it here still applies 👍

michaelsembwever:
@shyla226 this is looking good.  i have further questions about tightening up the messages, and avoiding us having to edit hardcoded numbers.

these are best expressed in the following suggested patch.  does it make sense ? are you willing to squash it into your work ?  https://github.com/apache/cassandra/commit/edecd68d277ae0deb645ae160077c3e9341be6ed

(if you agree, it would need to be re-applied to each of the modified files)

michaelsembwever (approved):

bin/cassandra.in.sh
```diff
@@ -116,19 +116,34 @@ if [ -z $JAVA ] ; then
...
-elif [ "$JVM_VERSION" \< "17" ] ; then
-    echo "Cassandra 5.0 requires Java 11 or Java 17."
+# Unsupported JDKs below the upper supported version are not allowed
+if [ "$short" != "$(echo "$java_versions_supported" | cut -d, -f1)" ] && [ "$JVM_VERSION" \< "$(echo "$java_versions_supported" | cut -d, -f2)" ] ; then
+    echo "Unsupported Java $JVM_VERSION. Supported are $java_versions_supported"
     exit 1;
 fi
+# Allow execution of supported Java versions, and newer if CASSANDRA_JDK_UNSUPPORTED is set
+supported_version=$(echo "$java_versions_supported" | tr "," '\n' | grep -F -x "$short")
+if [ -z "$supported_version" ] ; then
```
ekaterinadimitrova2: nit: `supported_versions`

bin/cassandra.in.sh
```diff
@@ -116,19 +116,34 @@ if [ -z $JAVA ] ; then
...
+    if [ -z "$CASSANDRA_JDK_UNSUPPORTED" ] ; then
+        echo "Unsupported Java $JVM_VERSION. Supported are $java_versions_supported"
+        echo "If you would like to test with newer Java versions set CASSANDRA_JDK_UNSUPPORTED=true"
+        exit 1;
+    else
+        echo "######################################################################"
+        echo "Warning! You are using JDK$short. Cassandra currently only supports $java_versions_supported."
```
ekaterinadimitrova2: nit: `Warning! You are using JDK$short. This Cassandra version only supports $java_versions_supported.`


If we do not want to pull/hardcore the version of Cassandra we can just stress on `this version`
I hear many times people ask about JDK version and assume that it will be added for all versions, so it is good to stress on "this version" in my opinion

redhat/cassandra.in.sh
```diff
@@ -74,19 +74,34 @@ if [ -z $JAVA ] ; then
...
-elif [ "$JVM_VERSION" \< "17" ] ; then
-    echo "Cassandra 5.0 requires Java 11 or Java 17."
+# Unsupported JDKs below the upper supported version are not allowed
+if [ "$short" != "$(echo "$java_versions_supported" | cut -d, -f1)" ] && [ "$JVM_VERSION" \< "$(echo "$java_versions_supported" | cut -d, -f2)" ] ; then
+    echo "Unsupported Java $JVM_VERSION. Supported are $java_versions_supported"
     exit 1;
 fi
+# Allow execution of supported Java versions, and newer if CASSANDRA_JDK_UNSUPPORTED is set
+supported_version=$(echo "$java_versions_supported" | tr "," '\n' | grep -F -x "$short")
+if [ -z "$supported_version" ] ; then
```
ekaterinadimitrova2: nit:`supported_versions`

tools/bin/cassandra.in.sh
```diff
@@ -76,18 +76,35 @@ if [ -z $JAVA ] ; then
...
+# Unsupported JDKs below the upper supported version are not allowed
+if [ "$short" != "$(echo "$java_versions_supported" | cut -d, -f1)" ] && [ "$JVM_VERSION" \< "$(echo "$java_versions_supported" | cut -d, -f2)" ] ; then
+    echo "Unsupported Java $JVM_VERSION. Supported are $java_versions_supported"
+    exit 1;
+fi
+
+# Allow execution of supported Java versions, and newer if CASSANDRA_JDK_UNSUPPORTED is set
+supported_version=$(echo "$java_versions_supported" | tr "," '\n' | grep -F -x "$short")
+if [ -z "$supported_version" ] ; then
+    if [ -z "$CASSANDRA_JDK_UNSUPPORTED" ] ; then
```
ekaterinadimitrova2:
```suggestion
   if [ -z "$CASSANDRA_JDK_UNSUPPORTED" ] || [ "$CASSANDRA_JDK_UNSUPPORTED" = "false" ] ; then
```

Otherwise I can set it to false and it still accepts JDK21 which is weird

ekaterinadimitrova2: This comment is valid for all places where this check occurs

ekaterinadimitrova2 (commented):
+1 after addressing the env variable check to be not only whether it is set or not, but also if it is set to false.
Plus we need to test the packaging in Jenkins

michaelsembwever:
> Plus we need to test the packaging in Jenkins

Jenkins does not work for pre-commit on 5.0 and trunk.
Jenkins just uses the in-tree scripts, and it has already been tested thoroughly with these (separately by both @shyla226 and myself)

tools/bin/cassandra.in.sh
shyla226: Makes sense. I have added this. Will update rest of the files

bin/cassandra.in.sh
shyla226: @ekaterinadimitrova2, that variable was intended to be treated as a boolean. That is to say, Java version supported or not. Will it be better if I said version_supported?

bin/cassandra.in.sh
shyla226: Updated this

bin/cassandra.in.sh
michaelsembwever: yes @shyla226 , `supported_version` is correct.  it should not be plural, nothing needs to be changed here IMHO.

redhat/cassandra.in.sh
michaelsembwever: see above.  `supported_version`  is correct and nothing should be changed here (IMHO)

bin/cassandra.in.sh
ekaterinadimitrova2: Yes, my bad, I missed on the last read that this is boolean. Those are most of the time starting with `is_*` or so. 
It is all good the way it is, no changes needed, thanks.

pmcfadin:
Closing as [CASSANDRA-18688](https://issues.apache.org/jira/browse/CASSANDRA-18688) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #2577, opened 2023-08-11
 CASSANDRA-18725: IsolatedJmx should only remove the endpoints it created when stopping (3.11)

JeetKunDoug:

    potential issues like this.

### DIFF
--- a/src/java/org/apache/cassandra/utils/RMIClientSocketFactoryImpl.java
+++ b/src/java/org/apache/cassandra/utils/RMIClientSocketFactoryImpl.java
@@ -21,8 +21,11 @@ package org.apache.cassandra.utils;
 import java.io.IOException;
 import java.io.Serializable;
 import java.net.InetAddress;
+import java.net.ServerSocket;
 import java.net.Socket;
 import java.rmi.server.RMIClientSocketFactory;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
 
 /**
@@ -32,6 +35,7 @@ import java.util.Objects;
  */
 public class RMIClientSocketFactoryImpl implements RMIClientSocketFactory, Serializable
 {
+    List<Socket> sockets = new ArrayList<>();
     private final InetAddress localAddress;
 
     public RMIClientSocketFactoryImpl(InetAddress localAddress)
@@ -42,7 +46,24 @@ public class RMIClientSocketFactoryImpl implements RMIClientSocketFactory, Seria
     @Override
     public Socket createSocket(String host, int port) throws IOException
     {
-        return new Socket(localAddress, port);
+        Socket socket = new Socket(localAddress, port);
+        sockets.add(socket);
+        return socket;
+    }
+
+    public void close() throws IOException
+    {
+        for (Socket socket: sockets)
+        {
+            try
+            {
+                socket.close();
+            }
+            catch (IOException ignored)
+            {
+                // intentionally ignored
+            }
+        }
     }
 
     @Override
--- a/src/java/org/apache/cassandra/utils/ReflectionUtils.java
+++ b/src/java/org/apache/cassandra/utils/ReflectionUtils.java
@@ -1,0 +1,101 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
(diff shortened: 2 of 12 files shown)


jonmeredith (approved):

src/java/org/apache/cassandra/utils/RMIClientSocketFactoryImpl.java
```diff
@@ -42,7 +46,21 @@ public RMIClientSocketFactoryImpl(InetAddress localAddress)
...
     {
-        return new Socket(localAddress, port);
+        Socket socket = new Socket(localAddress, port);
+        sockets.add(socket);
+        return socket;
+    }
+
+    public void close() throws IOException
+    {
+        for (Socket socket: sockets) {
```
smiklosovic: @JeetKunDoug please put braces on new lines in this method

src/java/org/apache/cassandra/utils/ReflectionUtils.java
```diff
@@ -0,0 +1,94 @@
...
+                // but for debugging purposes it is much easier to keep it like this.
+                Iterator<Map.Entry<K,V>> it = map.entrySet().iterator();
+                while (it.hasNext()) {
+                    Map.Entry<K,V> entry = it.next();
+                    if (shouldRemove.test(entry)) {
+                        it.remove();
+                    }
+                }
+            }
+        } catch (NoSuchFieldException | IllegalAccessException ex) {
```
smiklosovic: @JeetKunDoug brace

test/distributed/org/apache/cassandra/distributed/impl/IsolatedJmx.java
```diff
@@ -127,45 +131,26 @@ public void startJmx()
...
-            }
-            catch (Throwable thrown)
-            {
-                lastThrown = thrown;
-            }
-            inInstancelogger.info("Could not connect to JMX on {} after {} attempts. Will retry.", url, attempts);
-            Thread.sleep(1000);
+        try (JMXConnector ignored = JMXUtil.getJmxConnector(config, 20, env)) {
+            // Do nothing - JMXUtil now retries
+        } catch (IOException iex) {
```
smiklosovic: @JeetKunDoug here too

test/distributed/org/apache/cassandra/distributed/test/jmx/JMXGetterCheckTest.java
```diff
@@ -29,44 +29,67 @@
...
+     * Useful for more than just testing getters, it also is used in JMXFeatureTest
+     * to make sure we've touched the complete JMX code path.
+     * @param cluster the cluster to test
+     * @throws Exception several kinds of exceptions can be thrown, mostly from JMX infrastructure issues.
+     */
+    public static void testAllValidGetters(Cluster cluster) throws Exception
+    {
+        for (IInvokableInstance instance: cluster)
+        {
+            if (instance.isShutdown()) {
```
smiklosovic: @JeetKunDoug this actually does not need to be enclosed with `{ }`.

src/java/org/apache/cassandra/utils/ReflectionUtils.java
```diff
@@ -0,0 +1,94 @@
...
+            // noinspection unchecked
+            Map<K, V> map = (Map<K, V>) mapField.get(instance);
+            // Because multiple instances can be shutting down at once,
+            // synchronize on the map to avoid ConcurrentModificationException
+            synchronized (map)
+            {
+                // This could be done with a simple `map.entrySet.removeIf()` call
+                // but for debugging purposes it is much easier to keep it like this.
+                Iterator<Map.Entry<K,V>> it = map.entrySet().iterator();
+                while (it.hasNext()) {
```
smiklosovic: here too, `it.remove()` does not need to be put in `{ }` as it is the only statement in that `if`.

JeetKunDoug: WRT the braces where they aren't needed, the code style states they _may_ be elided not that they _must_ be - by all means, I'll fix the placement (although obviously my personal habit is to have them on the end of the line not a new line) but I think keeping single-line bodies of if statements in braces is accepted by the code style and (for me at least) helps with readability, as it's clear there's a block of code there and not just some bad indentation.

PR #2580, opened 2023-08-11
CASSANDRA-18725: IsolatedJmx should only remove the endpoints it created when stopping (5.0)

JeetKunDoug:

    potential issues like this.


jonmeredith (approved):
same Uninterruptibles issue as trunk, once fixed +1

PR #2581, opened 2023-08-11
CASSANDRA-18725: IsolatedJmx should only remove the endpoints it created when stopping (trunk)

JeetKunDoug:

    potential issues like this.


test/distributed/org/apache/cassandra/distributed/impl/IsolatedJmx.java
```diff
@@ -18,38 +18,40 @@
...
 import javax.management.remote.JMXConnector;
-import javax.management.remote.JMXConnectorFactory;
 import javax.management.remote.JMXConnectorServer;
 import javax.management.remote.JMXServiceURL;
 import javax.management.remote.rmi.RMIConnectorServer;
 import javax.management.remote.rmi.RMIJRMPServerImpl;
 
 import org.slf4j.Logger;
 
+import com.google.monitoring.runtime.instrumentation.common.util.concurrent.Uninterruptibles;
```
jonmeredith: Should this be `com.google.common.util.concurrent.Uninterruptibles;`?

jonmeredith (approved):

test/distributed/org/apache/cassandra/distributed/impl/IsolatedJmx.java
JeetKunDoug: Yup - hit Option->Enter to quickly it seems. Fixed.

PR #2593, opened 2023-08-15
Ensure unit tests honor tmp.dir

dchenbecker:
Modify unit tests so that any files created are rooted in the directory specified by the "tmp.dir" Ant property. Ant already passes the value of tmp.dir to the JVM's java.io.tmpdir parameter, so the primary changes are to make use of Files.createTempDirectory to generate per-test work directories.

Patch by Derek Chen-Becker; reviewed by TBD for CASSANDRA-18750

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -90,7 +90,10 @@
     <property name="test.jvm.args" value="" />
     <property name="testtag.extra" value="" />
     <property name="dist.dir" value="${build.dir}/dist"/>
-    <property name="tmp.dir" value="${java.io.tmpdir}"/>
+
+    <!-- Use build/tmp for temp files if not otherwise specified. Because Ant properties are immutable, this has no effect if
+         the user specifies the tmp.dir property -->
+    <property name="tmp.dir" value="${build.dir}/tmp"/>
 
     <property name="doc.dir" value="${basedir}/doc"/>
 
@@ -381,6 +384,7 @@
 
     <target name="clean" description="Remove all locally created artifacts">
         <delete dir="${build.test.dir}" />
+        <delete dir="${build.dir}/tmp" />
         <delete dir="${build.classes}" />
         <delete dir="${build.src.gen-java}" />
         <delete dir="${version.properties.dir}" />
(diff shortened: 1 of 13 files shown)


build.xml
```diff
@@ -90,7 +90,10 @@
     <property name="test.jvm.args" value="" />
     <property name="testtag.extra" value="" />
     <property name="dist.dir" value="${build.dir}/dist"/>
-    <property name="tmp.dir" value="${java.io.tmpdir}"/>
+
+    <!-- Use build/tmp for temp files if not otherwise specified. Because Ant properties are immutable, this has no effect if
+         the user specifies the tmp.dir property -->
+    <property name="tmp.dir" value="${build.dir}/tmp"/>
```
bereng, line 96: Does this conflict, override, etc. [this](https://github.com/apache/cassandra/blob/97d23f40c9c7c3485b453fad5030f6427db5ca46/build.xml#L1127). Basically how they both play together?

dchenbecker: Basically, the property defines what is used for the `java.io.tmpdir` property in testing. Ant properties are immutable, so if someone provides the property on the command line (e.g. `ant -Dtmp.dir=/tmp/mytests ...`) then line 96 has no effect and the tests will use the provided directory. Now that you've mentioned it, though, I should make a change so that this honors the previous definition, which would use any provided `java.io.tmpdir` property on the command line. I can make it conditional so that we'll still use `${build.dir}/tmp` if `java.io.tmpdir` is not provided

dchenbecker: Actually, that won't work since we're running inside `ant` and the property will always be set since it's the JVM. I can't think of a way to differentiate between a JVM default `java.io.tmpdir` and one that was passed on the ant command line...

test/unit/org/apache/cassandra/audit/AuditLoggerTest.java
```diff
@@ -59,17 +59,17 @@
...
         options.logger = new ParameterizedClass("InMemoryAuditLogger", null);
         DatabaseDescriptor.setAuditLoggingOptions(options);
         requireNetwork();
     }
 
     @Before
-    public void beforeTestMethod()
+    public void beforeTestMethod() throws IOException
     {
         AuditLogOptions options = new AuditLogOptions();
```
driftx, line 74: Should this and the other instance like it also use getBaseAuditLogOptions for consistency?

dchenbecker: I actually started that refactor at first and then stopped because I wasn't sure whether consistency trumped minimizing changes. Happy to do that work if we all agree it's the right thing to do

driftx: I definitely think both consistency and having one way to do it are the way to go here.

bereng: +1

dchenbecker: PR is updated with the change for consistency

PR #2613, opened 2023-08-19
CASSNDRA-18780 Fixed ConnectionBurnTest fails with NullPointerException

Mmuzaf:
 
```

```

### DIFF
--- a/test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
+++ b/test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
@@ -51,7 +51,6 @@ import java.io.IOException;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
@@ -72,7 +71,6 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import static java.lang.Math.min;
-import static java.util.Collections.synchronizedList;
 import static org.apache.cassandra.net.ConnectionType.LARGE_MESSAGES;
 import static org.apache.cassandra.net.MessagingService.current_version;
 import static org.apache.cassandra.utils.Clock.Global.nanoTime;
--- a/test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
+++ b/test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
@@ -22,6 +22,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Queues;
 import com.google.common.util.concurrent.Uninterruptibles;
 import io.netty.channel.Channel;
+import io.netty.util.concurrent.Future;
 import net.openhft.chronicle.core.util.ThrowingBiConsumer;
 import net.openhft.chronicle.core.util.ThrowingRunnable;
 import org.apache.cassandra.config.DatabaseDescriptor;
@@ -50,6 +51,7 @@ import java.io.IOException;
 import java.net.UnknownHostException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.LinkedList;
@@ -57,6 +59,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -69,11 +72,13 @@ import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
 import static java.lang.Math.min;
+import static java.util.Collections.synchronizedList;
 import static org.apache.cassandra.net.ConnectionType.LARGE_MESSAGES;
 import static org.apache.cassandra.net.MessagingService.current_version;
 import static org.apache.cassandra.utils.Clock.Global.nanoTime;
 import static org.apache.cassandra.utils.MonotonicClock.Global.approxTime;
 import static org.apache.cassandra.utils.MonotonicClock.Global.preciseTime;
+import static org.apache.cassandra.utils.Throwables.maybeFail;
 
 public class ConnectionBurnTest
 {
@@ -159,7 +164,7 @@ public class ConnectionBurnTest
             public void serialize(byte[] payload, DataOutputPlus out, int version) throws IOException
             {
                 if (cancelled)
-                    throw new IOException("cancelled");
+                    return;
                 long id = MessageGenerator.getId(payload);
                 forId(id).serialize(id, payload, out, version);
             }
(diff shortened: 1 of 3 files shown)


test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
```diff
@@ -256,19 +271,19 @@ List<Connection> getConnections(InetAddressAndPort endpoint, boolean inbound)
         /**
          * Test connections with broken messages, live in-flight bytes updates, reconnect
          */
-        public void run() throws ExecutionException, InterruptedException, NoSuchFieldException, IllegalAccessException, TimeoutException
+        public void run() throws Exception
         {
             Reporters reporters = new Reporters(endpoints, connections);
+            List<Exception> exceptions = new LinkedList<>();
```
belliottsmith: why a linked list?

test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
```diff
@@ -467,26 +491,42 @@ public void run() throws ExecutionException, InterruptedException, NoSuchFieldEx
                     }
                 });
 
-                while (deadline > nanoTime() && failed.getCount() > 0)
+                int added = 0;
```
belliottsmith: at most we need a boolean, but I'd say this should be a part of the loop guard; perhaps use a do/while loop and simply check 0 == Queues.drainUninterruptibly()

test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
```diff
@@ -467,26 +491,42 @@ public void run() throws ExecutionException, InterruptedException, NoSuchFieldEx
...
+                while (deadline > nanoTime() && added == 0)
                 {
                     reporters.update();
                     reporters.print();
-                    Uninterruptibles.awaitUninterruptibly(failed, 30L, TimeUnit.SECONDS);
+                    added = Queues.drainUninterruptibly(unexpectedExceptions, exceptions, 10, 30L, TimeUnit.SECONDS);
```
belliottsmith: why 10 items, rather than 1?

test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
```diff
@@ -467,26 +491,42 @@ public void run() throws ExecutionException, InterruptedException, NoSuchFieldEx
...
                 }
 
                 executor.shutdownNow();
                 ExecutorUtils.awaitTermination(5L, TimeUnit.MINUTES, executor);
+
+                if (!exceptions.isEmpty())
+                    throw exceptions.get(0);
```
belliottsmith: should probably addSuppressed any items > index 0

test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
belliottsmith: also, consider declaring more proximal to usage

test/burn/org/apache/cassandra/net/Verifier.java
```diff
@@ -1326,9 +1339,11 @@ Chunk ensureChunk(long chunkSequenceId) throws InterruptedException
                 Map.Entry<Long, Chunk> e;
                 while ( null != (e = chunkList.firstEntry()) && chunkSequenceId - e.getKey() > 1 << 12)
                 {
+                    if (cancelled)
+                        throw new InterruptedException();
                     WaitQueue.Signal signal = writerWaiting.register();
                     if (null != (e = chunkList.firstEntry()) && chunkSequenceId - e.getKey() > 1 << 12)
-                        signal.await();
+                        signal.await(5L, TimeUnit.SECONDS);
```
belliottsmith: why is this now timed?

Mmuzaf: This is for safety's sake and in accordance with the following comment which gave me the feeling that we'd get the next chunk within a reasonable timeframe (5 seconds), and if not - we could fail the test, but of course I could be missing something.

https://github.com/apache/cassandra/blob/trunk/test/burn/org/apache/cassandra/net/Verifier.java#L668
```
// TODO: even 2s or 5s are unreasonable periods of time without _any_ movement on a message waiting to arrive
//       this seems to happen regularly on MacOS, but we should confirm this does not happen on Linux
```

test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
Mmuzaf: Fixed.

test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
Mmuzaf: Good suggestion, fixed.

test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
Mmuzaf: This is just a local buffer to fetch results from the `unexpectedExceptions` queue, so since it is only used locally it should be safe to use non-thread-safe collection, or did you mean something else? 

I think we can tie `drainUninterruptibly` and this exceptions buffer, so that they use the same buffer size via constant. I set it to `1` since we are only interested in the first exception that occurred, but we could increase it if you'd like.

test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
Mmuzaf: Fixed, and move this value to a constant value for visibility.

Mmuzaf:
@belliottsmith I've addressed your comments, please, take a look again :-)

Mmuzaf:
@belliottsmith Hi Benedict, have you had a chance to look at the comments that have been fixed? :-)

Mmuzaf:
@belliottsmith Hi Benedict, I think we are really close to fixing it! :-) Would you mind taking one last look?

test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
```diff
@@ -467,26 +493,43 @@ public void run() throws ExecutionException, InterruptedException, NoSuchFieldEx
                     }
                 });
 
-                while (deadline > nanoTime() && failed.getCount() > 0)
+                while (deadline > nanoTime() &&
+                        0 == Queues.drainUninterruptibly(unexpectedExceptions,
```
belliottsmith: I'm not sure this is the desired behaviour? It looks like the semantics of `drainUninterruptibly` is that it will wait for `exceptionBufferCapacity` before returning, whereas we want to return as soon as non-empty. This isn't super important here due to the timeout, but it is good to get this kind of thing right.

We should either wait for just 1 item, or use a different method for waiting, IMO.

test/burn/org/apache/cassandra/net/ConnectionBurnTest.java
belliottsmith: I meant that `LinkedList` is almost always inferior to `ArrayList`, and there appears to be no reason to prefer it here. It's not an important point, particularly here the distinction is unimportant. Thanks for making the change, though I'm not sure why `Lists.newArrayList` is preferred here to simply `new ArrayList<>()`.

Regarding declaring it more proximal to usage, I meant I don't see why it is declared at the start of the method rather than above the loop that uses it. It's helpful IMO to declare variables close to where they are logically introduced in the code.

test/burn/org/apache/cassandra/net/Verifier.java
belliottsmith: Sure, ok. Though given the earlier comment might be worth making this timeout a bit bigger to avoid false positives, or perhaps condition the wait time on the system under test.

PR #2616, opened 2023-08-21
CASSANDRA-18782 5.0: Forbid analyzed SAI indexes on primary key columns

adelapena:


test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
```diff
@@ -417,6 +421,33 @@ public void shouldEnableAsciiSearch()
...
+            }
+        }
+    }
+
+    private void assertRejectsAnalysisOnPrimaryKeyColumns(String column, Map<String, String> optionsMap)
+    {
+        String options = new CqlBuilder().append(optionsMap).toString();
+        Assertions.assertThatThrownBy(() -> createIndex("CREATE INDEX ON %s(" + column + ") USING 'sai' WITH OPTIONS = " + options))
+                  .hasRootCauseInstanceOf(InvalidRequestException.class)
+                  .hasRootCauseMessage("Analysis options are not supported on primary key columns, but found " + options);
```
maedhroz: nit: We could always make a constant error message string in `StorageAttachedIndex` so we don't duplicate here, and if it changes slightly, we don't have to change it both places...

maedhroz (approved):

PR #2620, opened 2023-08-22
CASSANDRA-18785: Add support for Sonar analysis

jacek-lewandowski:


.build/build-sonar.xml
```diff
@@ -0,0 +1,158 @@
...
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<project basedir="." name="apache-cassandra-sonar-tasks" xmlns:if="ant:if" xmlns:unless="ant:unless">
+    <property name="sonar-scanner-cli.version" value="5.0.1.3006-linux"/>
+    <property name="sonar.workdir" value="${build.dir}/sonar"/>
+    <property name="sonar.download.dir" value="${user.home}/.cassandra-cache/tools"/>
+    <property name="sonar-scanner.home" value="${sonar.download.dir}/sonar-scanner-${sonar-scanner-cli.version}"/>
+    <property name="sonar.projectKey" value="cassandra"/>
+    <property name="sonar.mainBranch" value="trunk"/>
```
smiklosovic, line 25: This would be nice to have automatically resolved based on what branch this file is in. I get that we will probably use this from now on only but once trunk is cassandra-5.1 and we are doing some patches for it, would not it be nice if we had a way how to run Sonar against that branch too without tweaking this property every time?

.build/build-sonar.xml
```diff
@@ -0,0 +1,158 @@
...
+            </setpermissions>
+        </sequential>
+    </target>
+
+    <macrodef name="check-server-status">
+        <attribute name="url"/>
+        <sequential>
+            <exec executable="bash" failifexecutionfails="true" failonerror="true" logerror="false">
+                <arg value="-c"/>
+                <arg value="curl -s &quot;@{url}/api/system/status&quot; | jq -r &quot;.status&quot; | grep UP"/>
```
smiklosovic: I am not saying that depending on jq is wrong, I am just saying that people will probably have to install this on their machines beforehand. If we dont replace it with something else, maybe some "sanity check" target to check if all required tooling is available before advancing this target as a whole would be nice so it fails as soon as possible.

.build/build-sonar.xml
jacek-lewandowski: everytime in this case means every year (once we branch off a new version); but to be honest, this is not used atm, because we do not have a reference analysis to compare the local project with

.build/build-sonar.xml
jacek-lewandowski: `jq` will be probably needed for OWASP as well... we can maybe use this: https://github.com/eiiches/jackson-jq

smiklosovic: +1 yeah probably we can use that one so it would be downloaded ...

.build/build-sonar.xml
```diff
@@ -0,0 +1,167 @@
...
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<project basedir="." name="apache-cassandra-sonar-tasks" xmlns:if="ant:if" xmlns:unless="ant:unless">
+    <property name="sonar-scanner-cli.version" value="5.0.1.3006-linux"/>
+    <property name="sonar.workdir" value="${build.dir}/sonar"/>
+    <property name="sonar.download.dir" value="${user.home}/.cassandra-cache/tools"/>
+    <property name="sonar-scanner.home" value="${sonar.download.dir}/sonar-scanner-${sonar-scanner-cli.version}"/>
```
Mmuzaf, line 24: I think we should put this directory somewhere under the project root, right? I would not expect the project to change anything in my home directory.

.build/build-sonar.xml
Mmuzaf: +1 to Steafan's comment. 

We have branch resolving in our shell scripts, is there any reason to keep it hardcoded in ant scripts? :-)

branch=`git symbolic-ref -q --short HEAD 2>/dev/null || true`

.build/build-sonar.xml
```diff
@@ -0,0 +1,167 @@
...
+            <echo>Downloading Sonar-Scanner CLI...</echo>
+            <retry retrycount="3" retrydelay="5000" >
+                <get src="https://binaries.sonarsource.com/Distribution/sonar-scanner-cli/sonar-scanner-cli-${sonar-scanner-cli.version}.zip"
+                     dest="${sonar.download.dir}/sonar-scanner-cli-${sonar-scanner-cli.version}.zip"/>
+            </retry>
+            <unzip src="${sonar.download.dir}/sonar-scanner-cli-${sonar-scanner-cli.version}.zip" dest="${sonar.download.dir}" overwrite="true"/>
+            <setpermissions mode="755">
+                <!-- unzip task in Ant does not preserve file permissions, so we need to make those files executable explicitly -->
+                <file file="${sonar-scanner.home}/bin/sonar-scanner" />
+                <file file="${sonar-scanner.home}/jre/bin/java" />
```
Mmuzaf: I've got the following during my first run, the directory seems to have the correct permissions. There is a couple of thing that confuses me:
- it fails
- it reports that the build is successful

```
m@MBP cassandra % ls -la ~/.cassandra-cache/tools/sonar-scanner-5.0.1.3006-linux/jre/bin/java 
-rwxr-xr-x@ 1 m  s  16320 Apr 18 21:49 /Users/m/.cassandra-cache/tools/sonar-scanner-5.0.1.3006-linux/jre/bin/java
```

```
init-sonar-server:
 [exec] UP

sonar:
 [exec] /Users/m/.cassandra-cache/tools/sonar-scanner-5.0.1.3006-linux/bin/sonar-scanner: line 66: /Users/m/.cassandra-cache/tools/sonar-scanner-5.0.1.3006-linux/jre/bin/java: cannot execute binary file
 [exec] /Users/m/.cassandra-cache/tools/sonar-scanner-5.0.1.3006-linux/bin/sonar-scanner: line 66: /Users/m/.cassandra-cache/tools/sonar-scanner-5.0.1.3006-linux/jre/bin/java: Undefined error: 0
 [exec] Result: 1

BUILD SUCCESSFUL
Total time: 2 minutes 20 seconds
m@MBP cassandra % 
```

I'll check what is wrong with it and how can it be fixed.

.build/build-sonar.xml
```diff
@@ -0,0 +1,167 @@
...
+
+        <retry retrycount="60" retrydelay="1000">
+            <check-server-status url="${sonar.host.url}"/>
+        </retry>
+
+        <exec executable="curl" failifexecutionfails="true" failonerror="true" logerror="false">
+            <arg line="-u admin:admin"/>
+            <arg line="-X POST"/>
+            <arg line="-s"/>
+            <arg value="${sonar.host.url}/api/users/change_password?login=admin&amp;password=password&amp;previousPassword=admin"/>
```
Mmuzaf, line 178: What is wrong with the `admin:admin`? Why do we need to change it? This is the requirement of the sonar itself, right?

.build/build-sonar.xml
```diff
@@ -0,0 +1,167 @@
...
+
+        <exec executable="bash" failifexecutionfails="true" failonerror="true" logerror="false">
+            <arg line="-c"/>
+            <arg value="curl -s -u admin:password -X POST '${sonar.host.url}/api/user_tokens/generate?name=cassandra&amp;type=GLOBAL_ANALYSIS_TOKEN' | jq -r '.token'"/>
+            <redirector outputproperty="sonar.token"/>
+        </exec>
+    </target>
+
+    <target name="start-sonarqube" depends="sonar-init">
+        <exec executable="docker" dir="${sonar.workdir}" failifexecutionfails="true" failonerror="true" logerror="true">
```
Mmuzaf, line 150: The script that was created here: [In-tree build scripts](https://issues.apache.org/jira/browse/CASSANDRA-18133) runs and stops the docker image. Should we reuse this approach here as well? wdyt? And here we can just call the sh we need.

Mmuzaf:
I think we need to do a tiny fix :-)

```
1 Unknown Licenses

*******************************

Unapproved licenses:

  /Users/mmuzaf/IdeaProjects/cassandra/.build/sonar-quality-profile.xml
```

.build/build-sonar.xml
jacek-lewandowski: I'm not sure which branch you guys refer to.

As far as I understand, the community edition of Sonar cannot compare analysis results between branches. It can only compare to the previous analysis. In the paid developer edition, the analysis results of the current branch can be compared to the reference branch analysis results. In such case, for a feature branch, say `CASSANDRA-18785-4.1`, based on `cassandra-4.1`, `mainBranch` would be `cassandra-4.1`, and for the analysis, we would provide additionally `sonar.branch.name=CASSANDRA-18785-4.1`. Since the reference analysis results are unavailable anywhere, the developer could run the analysis manually on `cassandra-4.1` to collect the reference results and then use the same server to run the analysis on its feature branch. 

But, as said, this is a feature of the paid developer edition and is inaccessible in the community edition. Therefore, it really does not matter whether we use `trunk`, `main` (which is the default), or the current branch name for `mainBranch`. We can only make it consider a new code to be something between this and the previous run.

.build/build-sonar.xml
jacek-lewandowski: thinking about that more... I'm rather reluctant to that. It is not very uncommon thing to have `jq` installed when you are a developer. I'm ok with failing clearly if this prerequisite is not satisfied.

.build/build-sonar.xml
jacek-lewandowski: This is a cache. Putting that under the build directory in the project invalidates that. That is, it would no longer be a cache, and everything would have to be fetched again whenever one does `ant clean`. We will have a similar story with OWASP analysis which needs to fetch much more data to run. I'd even consider putting the Sonar volumes there because it is probably not the best thing to have them wiped on `ant clean`, especially when the server is still running. Though, it is something that deserves more discussion...

.build/build-sonar.xml
jacek-lewandowski: linux version of sonar...

.build/build-sonar.xml
jacek-lewandowski: Yes, it forces the user to change the password as the first thing after startup. It does not allow for using `admin` as a password again.

.build/build-sonar.xml
jacek-lewandowski: We don't want to stop the server because we would not be able to browse the results.

.build/build-sonar.xml
jacek-lewandowski: Can you try again, I've pushed a fix for that

jacek-lewandowski: Note that I decided for the platform specific version, that is the one with attached JVM because the Scanner requires j17 and it is not obvious it is available in all the environments; it would be much stronger requirement than installing `jq`.

.build/build-sonar.xml
Mmuzaf: The first point,

I think we should be able to clean everything related to the project with one command, currently, I can do this with `ant realclean`. Tracking and cleaning all these kinds of caches manually is a really complex problem (you have to be aware of all the locations), the developer may not even know anything about them. We have a contract now - `realclean` cleans everything, and we have to keep this contract.

For example, checkstyle has its own cache located at `build/checkstyle/checkstyle_cachefile` which is cleaned by `ant realclean` and NOT cleaned by `ant clean`.

The second point,

I think you are assuming that you have only one copy of cassandra's upstream, but this is generally not true. For example, I have 3 copies of cassandra, and each of them has a slightly different trunk (upstream copy, local copy, and company's internal). This will lead to conflicts when the sonar cache is localized to my home directory. This is actually a bad way to control everything related to the project, just imagine if every open source project created something in your home directory how bad would that be to see such a messy home directory.

The third point,

I use two copies of the Cassandra source files in parallel - one for development and one for testing/reviewing patches. Now I am able to build and work with these projects in parallel, but after these changes, I wouldn't be able to, right? This sounds really bad to me :-(

.build/build-sonar.xml
smiklosovic: OK do it like that then.

.build/build-sonar.xml
smiklosovic: @jacek-lewandowski maybe putting it all to `/tmp` would be OK too? That would still cache it, kind of, even though it would not survive the restart of the developer's PC.

@Mmuzaf I think that we can do an exception like this. OWASP cache literally just downloads data from the OWASP server. It just downloads the repository of CVEs. There is nothing project-specific (as in Cassandra project) which we need to have `realclean` for.

Checkstyle is something different, isnt it? That cache is project specific. I mean ... its content depends on how the sources look like.

Anyway I think that moving it to /tmp is a nice compromise.

driftx: I am -1 on polluting the home directory and you can't rely on literal '/tmp'  on all machines, but tmp.dir may work.

jacek-lewandowski: I see your point and I don't want to make anything that would break working with multiple project worktrees, hence we are discussing.

Though:
1. the server needs to be running in order to browse the results, there can be an Ant task to stop it
2. When working with multiple worktrees, starting a new server for each worktree could be confusing, we can easily confuse which port was used for which project worktree
3. We do not necessarily setup volumes assuming that we will not reuse the results later, this would fix the problem of storing data in the build directory and exposing to problems when trying to clean the build directory while the server is still working. We would not need to store the data outside the project worktree. 
4. In order to reuse the server for multiple project worktrees, we could create a project in Sonar per project worktree. The only question would be the name of the project - branch name would not be sufficient because, as you mentioned, you operate on more than one `trunk`, which is totally justified. Perhaps branch name and sha? Or Git worktree name? 
5. OWASP cache will be more like Maven repository, I don't think we would want to clean it because it is branch-agnostic. It also ~ 500MB database.

wdyt?

.build/build-sonar.xml
Mmuzaf: Can we report the prepared result to the console? While the docker is running I think we can fetch all the results from the API and push them to the console. 

The two things that concern me the most: 
- We leave unclosed resources after the build process. To me, the `ant` is a build tool, and not designed to be a server-like script when we initiate something to start like a daemon;
- Another point is if it leaves the resources unclosed, it will probably only be used on local machines, which, in turn, concerns why not just use simple `bash` instead, like in-tree scripts.

.build/build-sonar.xml
smiklosovic: Running Sonar locally is a corner case after all, isnt it? The primary usage is to connect to already running Sonar server instance _somewhere_.

Anyway, 4. seems reasonable but how am I operating on more than one trunk? Each of my worktree is on a dedicated branch different from each other.

5. +1

.build/build-sonar.xml
```diff
@@ -0,0 +1,174 @@
...
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+<project basedir="." name="apache-cassandra-sonar-tasks" xmlns:if="ant:if" xmlns:unless="ant:unless">
+    <condition property="sonar-scanner-cli.platform" value="linux">
+        <os family="unix"/>
+    </condition>
+    <condition property="sonar-scanner-cli.platform" value="maxosx">
```
Mmuzaf:
```suggestion
    <condition property="sonar-scanner-cli.platform" value="maсosx">
```

.build/build-sonar.xml
jacek-lewandowski: Some ideas @smiklosovic for 4 then?

.build/build-sonar.xml
jacek-lewandowski: I think this is out of the scope. We could probably export list of issues in JSON, but that is a bit unreadable without some processing.

jacek-lewandowski:
The recent commits should fix the issue with:
- running on Mac M1
- the Scanner waits for analysis results from the server and reports quality gate status (passed/failed), fails the build if the quality gate fails
- since the quality gate is defined on the new-code and we do not have any reference, I did some dirty hack to run the quality gate on all code by initially running empty analysis (on an empty sources dir), and then the real analysis. In this case all files are considered new

If you guys know a better way to run the quality gate on all files rather than only those considered new, please share...

.build/build-sonar.xml
smiklosovic: branch name and SHA seems like a good combination, SHA might be trimmed to e.g. first 6 characters for brevity?

.build/build-sonar.xml
```diff
@@ -0,0 +1,257 @@
...
+                <outputfilterchain>
+                    <tokenfilter>
+                        <replaceregex pattern=".*&quot;token&quot;:&quot;([^&quot;]+)&quot;.*" replace="\1"/>
+                    </tokenfilter>
+                </outputfilterchain>
+            </redirector>
+        </exec>
+    </target>
+
+    <target name="stop-sonar-server" depends="sonar-init">
```
smiklosovic: @jacek-lewandowski I would rename these targets to "sonar-server-start/stop", when I am in the console, and I type "ant son<tab>", it does not offer this target and then I need to figure out which one it is ... it is easier for a user if all targets are grouped under "sonar-something".

smiklosovic: @jacek-lewandowski also, if I do:

    ant start-sonar-server
    ant stop-sonar-server
    ant start-sonar-server

The last command fails, because there is still a container of name "sonarqube" but it is in exited state and it collides.

Would it be possible to invoke `docker rm sonarqube` upon `stop-sonar-server`?

tools/stress/resources/org/apache/cassandra/stress/graph/graph.html
smiklosovic, line 1: @jacek-lewandowski what is this change about?

.build/build-sonar.xml
jacek-lewandowski: I deliberately didn't remove the container - the message at the end should says that. Until it is removed you can start it again and navigate the results. Also, I'd avoid removing the container because someone can already have some sonarqube container in the system and our script should not remove it

tools/stress/resources/org/apache/cassandra/stress/graph/graph.html
jacek-lewandowski: the location of that file was wrong - resources should be outside of the source root; that caused Sonar to use some HTML analyzer

.build/build-sonar.xml
smiklosovic: OK, so ... do you think that "sonar-xyz" targets make sense?

jacek-lewandowski: ah, that - yes, I've already changed that, though haven't committed yet

.build/build-sonar.xml
michaelsembwever: Agree with @driftx , we can't be polluting home. (And it's not even in $HOME/.cache anyway)

This should follow the existing precedences, see build-owasp.xml and build-resolver.xml .  Use the build dir or the correct path under ~/.m2/repository

If *you* would like it elsewhere, that's what `sonar.download.dir` is for.

jacek-lewandowski: Those are the default cache directories for different OSes:

```
Mac OS X: ~/Library/Caches/<AppName>
Linux: ~/. cache/<AppName>
Win XP: C:\\Documents and Settings\\<username>\\Local Settings\\Application Data\\<AppAuthor>\\<AppName>\\Cache.
```

would it work if the we store the data under `<default cache dir>/cassandra/builds` ?

build.xml
```diff
@@ -27,8 +27,9 @@
     </fail>
 
     <property environment="env"/>
-    <property file="build.properties" />
-    <property file="build.properties.default" />
+    <property file="build.properties"/>
+    <property file="build.properties.default"/>
```
michaelsembwever: nit: a few lines in this file change for only whitespace. remove from patch please.

.build/build-sonar.xml
jacek-lewandowski: @driftx I've applied your suggestion to use `tmp.dir`, but I've also added loading `~/.ant/build.properties` so that one can override globally `tmp.dir` for all projects, as suggested by @smiklosovic

build.xml
jacek-lewandowski: shouldn't we fix the formatting of affected files?

Mmuzaf: I agree with Mick, the patch should only contain changes related to the problem. Formatting is definitely not on that list.

Mmuzaf (approved):

.build/build-sonar.xml
driftx: That works for me.

.build/build-sonar.xml
jacek-lewandowski: JSON report is saved in the build directory after analysis; in a separate ticket we can try to convert it to JUnit compatible report so that Jenkins can display it properly

michaelsembwever (approved):

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1072/workflows/d4a3fc43-c1fa-4dae-9bd9-f5d85f42b79c ⚠️ - rerunning one job

https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1072/workflows/b23a13d7-fd67-4dec-a1f5-5f27f606b07e

PR #2624, opened 2023-08-23
Refactor assertFalse(equals()) with assertNotEquals

Taher-Ghaleb:
I am working on research that investigates test smell refactoring in which we identify alternative implementations of test cases, study how commonly used these refactorings are, and assess how acceptable they are in practice.

This smell occurs when inappropriate assertions are used, while there exist better alternatives. For example, in [DiskBoundaryManagerTest.java](https://github.com/apache/cassandra/commit/bf2c5a44e86ef1b36c9ce756ad5fac1ec8e200e6#diff-2c0a8482948514ed8944fec2a26aa6af798d25351fd87c36c5f318181f30db71), I refactored `assertFalse(dbv1.equals(dbv2));` with `assertNotEquals(dbv1, dbv2);`.

I would like to get your feedback on this particular test smell and its refactoring. Thanks in advance for your input.

### DIFF
--- a/test/unit/org/apache/cassandra/cql3/KeyCacheCqlTest.java
+++ b/test/unit/org/apache/cassandra/cql3/KeyCacheCqlTest.java
@@ -46,6 +46,7 @@ import org.apache.cassandra.service.StorageService;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotEquals;
 import static org.junit.Assert.assertTrue;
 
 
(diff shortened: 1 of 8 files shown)


suenalaba (commented):
This Looks clean and is certainly more intuitive. But I can see the motivation of using assertFalse on an equality check. ```assertNotEquals``` is known for not providing helpful debugging messages to the developer when the assertion fails.

Taher-Ghaleb:
Thanks for your input, @suenalaba. I actually found other cases in this project where `assertNotEquals` was used. In your opinion, what scenarios should `assertNotEquals` instead of `assertFalse(equals())`, and vice versa? Thanks.

Taher-Ghaleb:
Hi @suenalaba, I would like to highlight that `assertNotEquals` also improves the readability of a test case and makes equality checks more explicit and intuitive, unlike `assertFalse` combined with `equals`, which indirectly checks if two objects are not equal. Also, `assertNotEquals` can actually provide more informative failure messages, as it can include details about the expected and actual values if the assertion fails, making it easier to diagnose and fix the issue. Moreover, the implementation of `assertNotEquals` is also designed to handle a variety of data types, as its implementation typically includes type checks and appropriate comparison logic to determine if the values or objects are unequal, regardless of their types, unlike `assertFalse`, which typically relies on the `equals()` method of the objects being compared.

Taher-Ghaleb:
Hi, any updates, please? Thanks.

suenalaba (approved):

maoling (approved):
LGTM. +1(no binding)

PR #2625, opened 2023-08-23
CASSANDRA-18773 4.0 compaction speedup

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/CompactionIterator.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionIterator.java
@@ -154,6 +154,17 @@ public class CompactionIterator extends CompactionInfo.Holder implements Unfilte
     {
         return new UnfilteredPartitionIterators.MergeListener()
         {
+            private boolean rowProcessingNeeded()
+            {
+                return type == OperationType.COMPACTION && controller.cfs.indexManager.hasIndexes();
+            }
+
+            @Override
+            public boolean preserveOrder()
+            {
+                return rowProcessingNeeded();
+            }
+
             public UnfilteredRowIterators.MergeListener getRowMergeListener(DecoratedKey partitionKey, List<UnfilteredRowIterator> versions)
             {
                 int merged = 0;
@@ -169,7 +180,7 @@ public class CompactionIterator extends CompactionInfo.Holder implements Unfilte
 
                 CompactionIterator.this.updateCounterFor(merged);
 
-                if (type != OperationType.COMPACTION || !controller.cfs.indexManager.hasIndexes())
+                if (!rowProcessingNeeded())
                     return null;
 
                 Columns statics = Columns.NONE;
--- a/src/java/org/apache/cassandra/db/partitions/UnfilteredPartitionIterators.java
+++ b/src/java/org/apache/cassandra/db/partitions/UnfilteredPartitionIterators.java
@@ -45,10 +45,30 @@ public abstract class UnfilteredPartitionIterators
 
     public interface MergeListener
     {
+        /**
+         * Returns true if the merger needs to preserve the position of sources within the merge when passing data to
+         * the listener. If false, the merger can avoid creating empty sources for non-present partitions and
+         * significantly speed up processing.
+         *
+         * @return True to preserve position of source iterators.
+         */
+        public default boolean preserveOrder() { return true; }
         public UnfilteredRowIterators.MergeListener getRowMergeListener(DecoratedKey partitionKey, List<UnfilteredRowIterator> versions);
         public default void close() {}
 
-        public static MergeListener NOOP = (partitionKey, versions) -> UnfilteredRowIterators.MergeListener.NOOP;
+        public static MergeListener NOOP = new MergeListener()
+        {
+            @Override
+            public boolean preserveOrder()
+            {
+                return false;
+            }
+
+            public UnfilteredRowIterators.MergeListener getRowMergeListener(DecoratedKey partitionKey, List<UnfilteredRowIterator> versions)
+            {
+                return UnfilteredRowIterators.MergeListener.NOOP;
+            }
+        };
     }
 
     @SuppressWarnings("resource") // The created resources are returned right away
@@ -108,6 +128,8 @@ public abstract class UnfilteredPartitionIterators
 
         final TableMetadata metadata = iterators.get(0).metadata();
 
+        final boolean preserveOrder = listener != null && listener.preserveOrder();
+
         final MergeIterator<UnfilteredRowIterator, UnfilteredRowIterator> merged = MergeIterator.get(iterators, partitionComparator, new MergeIterator.Reducer<UnfilteredRowIterator, UnfilteredRowIterator>()
         {
             private final List<UnfilteredRowIterator> toMerge = new ArrayList<>(iterators.size());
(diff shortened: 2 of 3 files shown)


src/java/org/apache/cassandra/db/partitions/UnfilteredPartitionIterators.java
```diff
@@ -45,6 +45,7 @@ private UnfilteredPartitionIterators() {}
 
     public interface MergeListener
     {
+        public default boolean preserveOrder() { return true; }
```
blambov, line 55: This needs a short javadoc/explanation.

`NOOP` below should be returning `false` for this.

src/java/org/apache/cassandra/db/partitions/UnfilteredPartitionIterators.java
```diff
@@ -120,9 +121,16 @@ public void reduce(int idx, UnfilteredRowIterator current)
                 partitionKey = current.partitionKey();
                 isReverseOrder = current.isReverseOrder();
 
-                // Note that because the MergeListener cares about it, we want to preserve the index of the iterator.
-                // Non-present iterator will thus be set to empty in getReduced.
-                toMerge.set(idx, current);
+                if (listener != null && listener.preserveOrder())
```
blambov: As nothing stops `preserveOrder` from changing between this call and the one in `getReduced` and `onKeyChange`, I would store the decision in a boolean and use that in the latter calls.

Better still, could we call `preserveOrder` just once at construction?

src/java/org/apache/cassandra/db/compaction/CompactionIterator.java
```diff
@@ -154,6 +154,12 @@ private UnfilteredPartitionIterators.MergeListener listener()
     {
         return new UnfilteredPartitionIterators.MergeListener()
         {
+            @Override
+            public boolean preserveOrder()
+            {
+                return type == OperationType.COMPACTION && controller.cfs.indexManager.hasIndexes();
```
blambov: Make sure to update this to match the `return null` path in `listener` for the other versions -- perhaps extract this to a method that both sites call?

src/java/org/apache/cassandra/db/partitions/UnfilteredPartitionIterators.java
```diff
@@ -132,17 +140,20 @@ protected UnfilteredRowIterator getReduced()
                                                                  ? null
                                                                  : listener.getRowMergeListener(partitionKey, toMerge);
 
-                // Make a single empty iterator object to merge, we don't need toMerge.size() copiess
-                UnfilteredRowIterator empty = null;
-
-                // Replace nulls by empty iterators
-                for (int i = 0; i < toMerge.size(); i++)
+                if (listener != null && listener.preserveOrder())
```
blambov: We could also skip the empty iterators if `rowListener` turns out to be null.

As this is a bit more involved (getting this listener and setting up the array needs to happen on the first `reduce` call), feel free to ignore this comment.

src/java/org/apache/cassandra/db/partitions/UnfilteredPartitionIterators.java
grom358: Updated NOOP and added javadoc comment

src/java/org/apache/cassandra/db/partitions/UnfilteredPartitionIterators.java
grom358: Completed

src/java/org/apache/cassandra/db/compaction/CompactionIterator.java
grom358: Didn't understand this one, would you mind expanding on this one?

src/java/org/apache/cassandra/db/partitions/UnfilteredPartitionIterators.java
grom358: yeah going ignore this one for now. I was trying todo that initially when creating the pull request but due to complexity I skipped it.

src/java/org/apache/cassandra/db/compaction/CompactionIterator.java
smiklosovic: @blambov would be great to have more feedback on this

blambov: Apologies, the comment is not clear at all. The code for choosing to return null is different in C* 5, and at the moment there's no way to see that the code for `preserveOrder` should be updated to match it (the patch will merge cleanly but won't be correct).

We should extract this `type == OperationType.COMPACTION && controller.cfs.indexManager.hasIndexes()` decision into a method, and call it from here as well as `listener()`. That way we will get a merge conflict when this is being ported up.

grom358: Okay I have updated this.

src/java/org/apache/cassandra/db/compaction/CompactionIterator.java
```diff
@@ -175,7 +175,7 @@ public UnfilteredRowIterators.MergeListener getRowMergeListener(DecoratedKey par
 
                 CompactionIterator.this.updateCounterFor(merged);
 
-                if (type != OperationType.COMPACTION || !controller.cfs.indexManager.hasIndexes())
+                if (!preserveOrder())
```
blambov: This would be extremely confusing to readers. Why does order preservation determine if we should create a row listener?

Create a new method called e.g. `rowProcessingNeeded()`, and call it from `preserveOrder()`  as well as here.

src/java/org/apache/cassandra/db/partitions/UnfilteredPartitionIterators.java
```diff
@@ -45,10 +45,27 @@ private UnfilteredPartitionIterators() {}
 
     public interface MergeListener
     {
+        /**
+         * Check wether the merge listener requires rreservation of iterator index.
```
blambov: Nit: wether -> whether, rreservation -> preservation

"Index" here is a bit confusing as it may refer to the secondary index, which is the usual reason to preserve order. I would phrase this as "Returns true if the merger needs to preserve the position of sources within the merge when passing data to the listener."

I would also add why it matters, e.g.: "If false, the merger can avoid creating empty sources for non-present partitions and significantly speed up processing."

src/java/org/apache/cassandra/db/compaction/CompactionIterator.java
grom358: done

src/java/org/apache/cassandra/db/partitions/UnfilteredPartitionIterators.java
grom358: whoops at typos.. updated

smiklosovic:
@blambov is there anything else you would like to address? Otherwise I would approach the builds. We also need to do this from 4.0 to trunk which is a work in itself.

src/java/org/apache/cassandra/db/compaction/CompactionIterator.java
```diff
@@ -169,7 +180,7 @@ public UnfilteredRowIterators.MergeListener getRowMergeListener(DecoratedKey par
 
                 CompactionIterator.this.updateCounterFor(merged);
 
-                if (type != OperationType.COMPACTION || !controller.cfs.indexManager.hasIndexes())
+                if (!rowProcessingNeeded())
```
smiklosovic, line 183: @grom358 should not this be `if (!preserveOrder())` ?

blambov: No. Order preservation does not determine if we return here, the need for row processing determines both if we return here and if we should preserve order.

blambov (approved):

src/java/org/apache/cassandra/db/partitions/UnfilteredPartitionIterators.java
```diff
@@ -45,10 +45,29 @@ private UnfilteredPartitionIterators() {}
...
+         */
+        public default boolean preserveOrder() { return true; }
         public UnfilteredRowIterators.MergeListener getRowMergeListener(DecoratedKey partitionKey, List<UnfilteredRowIterator> versions);
         public default void close() {}
 
-        public static MergeListener NOOP = (partitionKey, versions) -> UnfilteredRowIterators.MergeListener.NOOP;
+        public static MergeListener NOOP = new MergeListener()
+        {
+            @Override
+            public boolean preserveOrder() {
```
dcapwell: style, should move `{` to the next line

dcapwell (commented):
my review is not a formal review.  There was a discuss thread about back porting this so quickly skimmed to see if it made sense

PR #2635, opened 2023-08-26
Ensure unit tests honor tmp.dir

dchenbecker:
Modify unit tests so that any files created are rooted in the directory specified by the "tmp.dir" Ant property. Ant already passes the value of tmp.dir to the JVM's java.io.tmpdir parameter, so the primary changes are to make use of Files.createTempDirectory to generate per-test work directories. Also update the test checkstyle rules to catch any future usage of hardcoded "/tmp".

Patch by Derek Chen-Becker; reviewed by Brandon Williams for CASSANDRA-18750


dchenbecker:
Closing since it was already merged separately

PR #2645, opened 2023-08-29
SAI Result retriever is filtering too many rows

mike-tr-adamson:
This patch fixes a bug in the SegmentMetadata that was only storing the partition key for min and max primary keys for a segment. 

It also contains some refactoring of the PrimaryKey to remove the deferred loading of PrimaryKeys by the PrimaryKeyMaps.

Patch by Mike Adamson; reviewed by <reviewer> for CASSANDRA-18734


src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
```diff
@@ -17,34 +17,56 @@
...
+     * See the javadoc for {@link #kind()} for how this enum is used.
+      */
+    enum Kind
+    {
+        Token(false),
+        Skinny(false),
+        Wide(true),
+        Static(true);
+
+        public final boolean isClustering;
```
maedhroz:
```suggestion
        public final boolean hasClustering;
```
?

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
```diff
@@ -145,172 +208,226 @@ public boolean equals(Object obj)
...
+    Clustering<?> clustering();
 
     /**
      * Returns the {@link PrimaryKey} as a {@link ByteSource} byte comparable representation.
      * <p>
      * It is important that these representations are only ever used with byte comparables using
      * the same elements. This means that {@code asComparableBytes} responses can only be used
      * together from the same {@link PrimaryKey} implementation.
+     * <p>
+     * Note: This method can throw {@link UnsupportedOperationException} by some {@link PrimaryKey} implementations.
```
maedhroz: nit: Perhaps just use a `@throws` section for this bit?

```
@throws UnsupportedOperationException for {@link PrimaryKey} implementations that are not byte-comparable
```

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
```diff
@@ -74,66 +100,103 @@ public PrimaryKey createPartitionKeyOnly(DecoratedKey partitionKey)
...
+            public int compareTo(PrimaryKey o)
+            {
+                return token().compareTo(o.token());
             }
 
             @Override
             public int hashCode()
             {
-                return Objects.hash(token(), partitionKey(), clustering(), clusteringComparator);
+                return Objects.hash(token(), 16, 32, clusteringComparator);
```
maedhroz: What do the 16 and 32 represent?

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
```diff
@@ -145,172 +208,226 @@ public boolean equals(Object obj)
...
 
             @Override
-            public Clustering<?> clustering()
+            public int compareTo(PrimaryKey o)
             {
-                throw new UnsupportedOperationException();
+                int cmp = super.compareTo(o);
+
+                // If the tokens don't match then we don't need to compare any more of the key.
+                // Otherwise, if the other key is token only we can only compare tokens
```
maedhroz, line 253: I wonder if it's worth mentioning that this is necessary because of `KeyRangeIterator#skipTo()`

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
```diff
@@ -61,11 +83,15 @@ public PrimaryKey createTokenOnly(Token token)
             return new TokenOnlyPrimaryKey(token);
         }
 
+        /**
+         * Create a {@link PrimaryKey} for tables without clustering columns
+         */
         public PrimaryKey createPartitionKeyOnly(DecoratedKey partitionKey)
```
maedhroz: nit: For `createTokenOnly()` and `createPartitionKeyOnly()`, I wouldn't be sad if we just overloaded `create()`, given the JavaDoc and the arguments make it pretty clear what we can possibly create. Not a big deal either way though...

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
```diff
@@ -215,7 +216,12 @@ private ClusteringIndexFilter makeFilter(PrimaryKey key)
     {
         ClusteringIndexFilter clusteringIndexFilter = command.clusteringIndexFilter(key.partitionKey());
 
-        if (key.hasEmptyClustering())
+        assert cfs.metadata().comparator.size() == 0 && !key.kind().isClustering ||
+               cfs.metadata().comparator.size() > 0 && key.kind().isClustering :
+               "PrimaryKey " + key + " clustering does not match table. There should be a clustering of size " + cfs.metadata().comparator.size();
+
+        // TODO How do we avoid iterating the whole partition for static rows?
```
maedhroz: I've been looking at `StorageAttachedIndexSearcher#applyIndexFilter()`, and one thing I notice is that, when a static row is present, and we only have index expressions against it, we could avoid post-filtering on every row in the partition, even if we have to return every row in the partition. In other words, `QueryWriteLifecycleTest` passes if you rewrite `applyIndexFilter()` like this:

```
private static UnfilteredRowIterator applyIndexFilter(UnfilteredRowIterator partition, FilterTree tree, QueryContext queryContext)
{
    Row staticRow = partition.staticRow();

    queryContext.rowsFiltered++;
    if (tree.isSatisfiedBy(partition.partitionKey(), staticRow, staticRow))
        return partition;

    List<Unfiltered> clusters = new ArrayList<>();

    while (partition.hasNext())
    {
        Unfiltered row = partition.next();

        queryContext.rowsFiltered++;
        if (tree.isSatisfiedBy(partition.partitionKey(), row, staticRow))
            clusters.add(row);
    }

    if (clusters.isEmpty())
        return null;

    return new PartitionIterator(partition, staticRow, Iterators.filter(clusters.iterator(), u -> !((Row)u).isStatic()));
}
```
(Note that I've left out the comments in the method, which may need some tweaking.)

Maybe there's some other problem w/ this I'm not seeing, but it does short circuit a lot of work if we're just querying static columns?

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentMetadata.java
```diff
@@ -201,6 +204,19 @@ private static void writeBytes(ByteBuffer buf, IndexOutput out)
...
+    private static void writeBytes(byte[] bytes, IndexOutput out)
+    {
+        try
+        {
+            out.writeInt(bytes.length);
+            out.writeBytes(bytes, 0, bytes.length);
+        }
+        catch (IOException ioe)
+        {
+            throw new RuntimeException(ioe);
```
maedhroz, line 217:
```suggestion
            throw new UncheckedIOException(ioe);
```
Maybe in the method above too?

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
mike-tr-adamson: Ah, I think that got left in. I added them to try and avoid hash collisions but I'm not sure that is possible here or even needed. AFAICT token only keys are only used for comparison. 

I have removed these and replaced the magic number in the `SkinnyPrimaryKey` with `Clustering.EMPTY`.

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
mike-tr-adamson: I've been more specific because it is only used by the `ResultRetriever` to skip to key range start positions.

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
mike-tr-adamson: Well, I initially had them all as `create` and then changed them to this. I am more than happy to have them as all create overloads. I'll change it.

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
mike-tr-adamson: I think this is exactly what I wanted. The comment was really for me to think about the impact of querying on a static column index. Of course, the result is going to be that the whole partition matches and should be returned.

I'll put this change in and see how the testing goes.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentMetadata.java
mike-tr-adamson: I have changed this to `Throwables.unchecked`. This is what we do with these everywhere else so seems the right thing to do here.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentMetadata.java
```diff
@@ -30,12 +31,15 @@
 
 import com.google.common.collect.ImmutableMap;
 
-import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.index.sai.disk.format.IndexComponent;
 import org.apache.cassandra.index.sai.disk.v1.MetadataSource;
 import org.apache.cassandra.index.sai.disk.v1.MetadataWriter;
 import org.apache.cassandra.index.sai.utils.PrimaryKey;
 import org.apache.cassandra.utils.ByteBufferUtil;
+import org.apache.cassandra.utils.Throwables;
```
maedhroz: nit: `Throwables` is unused now

src/java/org/apache/cassandra/index/sai/plan/StorageAttachedIndexSearcher.java
```diff
@@ -380,6 +380,13 @@ private UnfilteredRowIterator queryStorageAndFilter(PrimaryKey key)
         private static UnfilteredRowIterator applyIndexFilter(UnfilteredRowIterator partition, FilterTree tree, QueryContext queryContext)
         {
             Row staticRow = partition.staticRow();
+
+            // We want to short-circuit the filtering of the whole partition if the static row
+            // satisfies the filter. If that is the case we just need to return the whole partition.
+            queryContext.rowsFiltered++;
+            if (tree.isSatisfiedBy(partition.partitionKey(), staticRow, staticRow))
+                return partition;
+
```
maedhroz, line 389: We can also remove the block below on L403 now, unless I'm missing something. If the `isSatisfiedBy()` check here fails, it won't suddenly pass later, right?

maedhroz (approved):
Other than the minor bit to clean up [here](https://github.com/apache/cassandra/pull/2645/files#r1313190721), everything LGTM

Thanks!

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
```diff
@@ -17,34 +17,56 @@
...
 /**
  * Representation of the primary key for a row consisting of the {@link DecoratedKey} and
  * {@link Clustering} associated with a {@link org.apache.cassandra.db.rows.Row}.
  */
 public interface PrimaryKey extends Comparable<PrimaryKey>, ByteComparable
 {
+    /**
+     * See the javadoc for {@link #kind()} for how this enum is used.
+      */
+    enum Kind
```
adelapena, line 49: All `Kind` enums in the codebase are written in upper case, I think we should use the same convention here.

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
```diff
@@ -17,34 +17,56 @@
...
+    {
+        Token(false),
```
adelapena: It might be a bit confusing that a token-only object can be called a primary key. I think we could add a short clarification on the class JavaDoc about why we can have this kind of keys.

adelapena (approved):
Looks good to me, +1. We should probably reabase and give it a final CI round including the j17 pre-commit workflow, just in case.

src/java/org/apache/cassandra/index/sai/utils/PrimaryKey.java
maedhroz: Yeah, I had that thought as well. Agree on the JavaDoc in lieu of a better concept that encompasses all the types. Something like `IndexKey` that could really be a token, partition key, or real compound primary key...but calling it a `PrimaryKey` is technically correct in most cases and descriptive when it is.

The only other thing is we could follow the example of `RingPosition` or something and have like...`IndexPosition` which our `PrimaryKey` and also just the token implement for comparisons (which is the only reason we have the token-only thing...to be comparable for skipping, right?)

mike-tr-adamson: I'd like to leave this as is with the clarification in the javadoc. I will raise a separate ticket to improve this later. CEP-30 makes some changes that may lead to this not being needed at all but if it still exists in some form, I will look at something like an `IndexPosition`.

PR #2648, opened 2023-08-29
Caffeine upgrade

ekaterinadimitrova2:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/auth/CIDRPermissionsCache.java
+++ b/src/java/org/apache/cassandra/auth/CIDRPermissionsCache.java
@@ -54,10 +54,11 @@ public class CIDRPermissionsCache extends AuthCache<RoleResource, CIDRPermission
      */
     public boolean invalidateCidrPermissions(String roleName)
     {
-        if (cache.getIfPresent(RoleResource.role(roleName)) == null)
+        RoleResource role = RoleResource.role(roleName);
+        if (cache.getIfPresent(role) == null)
             return false;
 
-        invalidate(RoleResource.role(roleName));
+        invalidate(role);
         return true;
     }
 
--- a/src/java/org/apache/cassandra/auth/CIDRPermissionsCache.java
+++ b/src/java/org/apache/cassandra/auth/CIDRPermissionsCache.java
@@ -54,7 +54,7 @@ public class CIDRPermissionsCache extends AuthCache<RoleResource, CIDRPermission
      */
     public boolean invalidateCidrPermissions(String roleName)
     {
-        if (cache.getIfPresent(roleName) == null)
+        if (cache.getIfPresent(RoleResource.role(roleName)) == null)
             return false;
 
         invalidate(RoleResource.role(roleName));
--- a/src/java/org/apache/cassandra/auth/CassandraCIDRAuthorizer.java
+++ b/src/java/org/apache/cassandra/auth/CassandraCIDRAuthorizer.java
@@ -154,10 +154,4 @@ public class CassandraCIDRAuthorizer extends AbstractCIDRAuthorizer
                                                           TimeUnit.NANOSECONDS);
         return hasAccess;
     }
-
-    @VisibleForTesting
-    public boolean has(RoleResource role)
-    {
-        return cidrPermissionsCache.cache.getIfPresent(role) != null;
-    }
 }
--- a/src/java/org/apache/cassandra/auth/CassandraCIDRAuthorizer.java
+++ b/src/java/org/apache/cassandra/auth/CassandraCIDRAuthorizer.java
@@ -154,4 +154,10 @@ public class CassandraCIDRAuthorizer extends AbstractCIDRAuthorizer
                                                           TimeUnit.NANOSECONDS);
         return hasAccess;
     }
+
+    @VisibleForTesting
+    public boolean has(RoleResource role)
+    {
+        return cidrPermissionsCache.cache.getIfPresent(role) != null;
+    }
 }
--- a/src/java/org/apache/cassandra/metrics/ChunkCacheMetrics.java
+++ b/src/java/org/apache/cassandra/metrics/ChunkCacheMetrics.java
@@ -21,6 +21,7 @@ import java.util.concurrent.TimeUnit;
 import javax.annotation.Nonnull;
 
 import com.codahale.metrics.Timer;
+import com.github.benmanes.caffeine.cache.RemovalCause;
 import com.github.benmanes.caffeine.cache.stats.CacheStats;
 import com.github.benmanes.caffeine.cache.stats.StatsCounter;
 import org.apache.cassandra.cache.ChunkCache;
@@ -72,7 +73,7 @@ public class ChunkCacheMetrics extends CacheMetrics implements StatsCounter
     }
 
     @Override
-    public void recordEviction()
+    public void recordEviction(int weight, RemovalCause cause)
     {
     }
 
@@ -80,6 +81,6 @@ public class ChunkCacheMetrics extends CacheMetrics implements StatsCounter
     @Override
     public CacheStats snapshot()
     {
-        return new CacheStats(hits.getCount(), misses.getCount(), missLatency.getCount(), 0L, missLatency.getCount(), 0L, 0L);
+        return CacheStats.of(hits.getCount(), misses.getCount(), missLatency.getCount(), 0L, missLatency.getCount(), 0L, 0L);
     }
 }
--- a/src/java/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCache.java
+++ b/src/java/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCache.java
@@ -39,8 +39,8 @@ public class InvalidateCIDRPermissionsCache extends NodeToolCmd
     {
         if (args.isEmpty())
         {
-            if(probe.invalidateCidrPermissionsCache(""))
-                probe.output().out.println("Invalidated CIDR permissions cache");
+            probe.invalidateCidrPermissionsCache("");
+            probe.output().out.println("Invalidated CIDR permissions cache");
         }
         else
         {
--- a/src/java/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCache.java
+++ b/src/java/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCache.java
@@ -39,8 +39,8 @@ public class InvalidateCIDRPermissionsCache extends NodeToolCmd
     {
         if (args.isEmpty())
         {
-            probe.invalidateCidrPermissionsCache("");
-            probe.output().out.println("Invalidated CIDR permissions cache");
+            if(probe.invalidateCidrPermissionsCache(""))
+                probe.output().out.println("Invalidated CIDR permissions cache");
         }
         else
         {
--- a/test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
+++ b/test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
@@ -20,13 +20,10 @@ package org.apache.cassandra.tools.nodetool;
 
 import java.net.InetSocketAddress;
 
+import org.apache.cassandra.auth.*;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
-import org.apache.cassandra.auth.AuthCacheService;
-import org.apache.cassandra.auth.AuthTestUtils;
-import org.apache.cassandra.auth.AuthenticatedUser;
-import org.apache.cassandra.auth.IRoleManager;
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.cql3.CQLTester;
 import org.apache.cassandra.tools.ToolRunner;
@@ -118,10 +115,13 @@ public class InvalidateCIDRPermissionsCacheTest extends CQLTester
         // ensure cidr permission is cached
         role.hasAccessFromIp(ipAddr);
         assertThat(originalReadsCount).isEqualTo(getCidrPermissionsReadCount());
+        assertThat(((CassandraCIDRAuthorizer) DatabaseDescriptor.getCIDRAuthorizer()).has(ROLE_A)).isTrue();
 
         // invalidate cidr permission
         ToolRunner.ToolResult tool = ToolRunner.invokeNodetool("invalidatecidrpermissionscache", ROLE_A.getRoleName());
         tool.assertOnCleanExit();
+        assertThat(tool.getStdout()).contains("Invalidated the role role_a from CIDR permissions cache");
+        assertThat(((CassandraCIDRAuthorizer) DatabaseDescriptor.getCIDRAuthorizer()).has(ROLE_A)).isFalse();
 
         // ensure cidr permission is reloaded
         assertThat(role.hasAccessFromIp(new InetSocketAddress("127.0.0.0", 0))).isTrue();
@@ -142,11 +142,15 @@ public class InvalidateCIDRPermissionsCacheTest extends CQLTester
         assertThat(roleA.hasAccessFromIp(ipAddr)).isTrue();
         assertThat(roleB.hasAccessFromIp(ipAddr)).isTrue();
         assertThat(originalReadsCount).isEqualTo(getCidrPermissionsReadCount());
+        assertThat(((CassandraCIDRAuthorizer) DatabaseDescriptor.getCIDRAuthorizer()).has(ROLE_A)).isTrue();
+        assertThat(((CassandraCIDRAuthorizer) DatabaseDescriptor.getCIDRAuthorizer()).has(ROLE_B)).isTrue();
 
         // invalidate both cidr permissions
         ToolRunner.ToolResult tool = ToolRunner.invokeNodetool("invalidatecidrpermissionscache");
         tool.assertOnCleanExit();
         assertThat(tool.getStdout()).contains("Invalidated CIDR permissions cache");
+        assertThat(((CassandraCIDRAuthorizer) DatabaseDescriptor.getCIDRAuthorizer()).has(ROLE_A)).isFalse();
+        assertThat(((CassandraCIDRAuthorizer) DatabaseDescriptor.getCIDRAuthorizer()).has(ROLE_B)).isFalse();
 
         // ensure cidr permission for roleA is reloaded
         assertThat(roleA.hasAccessFromIp(ipAddr)).isTrue();
--- a/test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
+++ b/test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
@@ -40,6 +40,7 @@ public class InvalidateCIDRPermissionsCacheTest extends CQLTester
     @BeforeClass
     public static void setup() throws Exception
     {
+        DatabaseDescriptor.setRolesValidity(Integer.MAX_VALUE-1);
         CQLTester.setUpClass();
         CQLTester.requireAuthentication();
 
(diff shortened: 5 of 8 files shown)


src/java/org/apache/cassandra/auth/CIDRPermissionsCache.java
```diff
@@ -54,7 +54,7 @@ public CIDRPermissionsCache(Function<RoleResource, CIDRPermissions> loadFunction
      */
     public boolean invalidateCidrPermissions(String roleName)
     {
-        if (cache.getIfPresent(roleName) == null)
+        if (cache.getIfPresent(RoleResource.role(roleName)) == null)
```
yifan-c: Thanks for the fix! It looks good. 
The original code does not actually invalidate from cache. (And I wrote a test that verifies it)

ekaterinadimitrova2: Thanks @yifan-c. There is actually this test class - InvalidateCIDRPermissionsCacheTest but it doesn't seem to test properly. It completes successfully with and without the change made. I think it doesn't really assert that invalidation happens. What test did you write?

yifan-c: yeah. Specifically, I changed the test, `testInvalidateSingleCidrPermission`. The assertions before and after invalidation were missing. The updated test fails before your patch. 

```java
        assertThat(((CassandraCIDRAuthorizer) DatabaseDescriptor.getCIDRAuthorizer()).has(ROLE_A)).isTrue();

        // invalidate cidr permission
        ToolRunner.ToolResult tool = ToolRunner.invokeNodetool("invalidatecidrpermissionscache", ROLE_A.getRoleName());
        tool.assertOnCleanExit();

        assertThat(((CassandraCIDRAuthorizer) DatabaseDescriptor.getCIDRAuthorizer()).has(ROLE_A)).isFalse();
```

ekaterinadimitrova2: Hmmm, there is no such method "has" :-) And I do not think you meant "hasAccessFromIp".  (I was looking to update the test class)

yifan-c: I added a `has` method for testing...

ekaterinadimitrova2: Updated the tests and added small condition in the nodetool command so we are sure we do not regress again in the future. Can you, please, take a look?

src/java/org/apache/cassandra/auth/CassandraCIDRAuthorizer.java
```diff
@@ -154,4 +154,10 @@ public boolean hasAccessFromIp(RoleResource role, InetAddress ipAddress)
                                                           TimeUnit.NANOSECONDS);
         return hasAccess;
     }
+
+    @VisibleForTesting
+    public boolean has(RoleResource role)
```
ekaterinadimitrova2: Not sure whether this is the best place for this method, but I did not find any better one so far. Open for ideas

src/java/org/apache/cassandra/auth/CIDRPermissionsCache.java
ben-manes:
```suggestion
        if (!cache.asMap().containsKey(RoleResource.role(roleName)))
```
this has slightly less cost (as does not record a policy access) but would reintroduce the type safety hole; either is perfectly fine.

test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
```diff
@@ -118,10 +115,13 @@ public void testInvalidateSingleCidrPermission()
         // ensure cidr permission is cached
         role.hasAccessFromIp(ipAddr);
         assertThat(originalReadsCount).isEqualTo(getCidrPermissionsReadCount());
+        assertThat(((CassandraCIDRAuthorizer) DatabaseDescriptor.getCIDRAuthorizer()).has(ROLE_A)).isTrue();
 
         // invalidate cidr permission
         ToolRunner.ToolResult tool = ToolRunner.invokeNodetool("invalidatecidrpermissionscache", ROLE_A.getRoleName());
         tool.assertOnCleanExit();
+        assertThat(tool.getStdout()).contains("Invalidated the role role_a from CIDR permissions cache");
```
skoppu22, line 123: @ekaterinadimitrova2 @yifan-c As we set expiry on cache entries, it could be possible that while tests run on circleCI/Jenkins etc, there may be delay after the entry is loaded into the cache and before invalidate runs, resulting in entry being expired before the actual invalidation. In that case, checks at line 118 and 123 may fail, causing flaky test.

src/java/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCache.java
```diff
@@ -39,8 +39,8 @@ public void execute(NodeProbe probe)
     {
         if (args.isEmpty())
         {
-            probe.invalidateCidrPermissionsCache("");
-            probe.output().out.println("Invalidated CIDR permissions cache");
+            if(probe.invalidateCidrPermissionsCache(""))
```
skoppu22: This change doesn't hurt, but works fine without this change as well. Invalidating entire CIDR permissions cache doesn't go through invalidateCidrPermissions(String roleName) we are changing above. i.e, we always get return value true here. 

```
    @Override
    public boolean invalidateCidrPermissionsCache(String roleName)
    {
        if (roleName == null || roleName.isEmpty())
        {
            cidrPermissionsCache.invalidate();
            return true;
        }

        return cidrPermissionsCache.invalidateCidrPermissions(roleName);
    }
```

ekaterinadimitrova2: But `return cidrPermissionsCache.invalidateCidrPermissions(roleName);` can return false? 
```
public boolean invalidateCidrPermissions(String roleName)
    {
        if (cache.getIfPresent(RoleResource.role(roleName)) == null)
            return false;

        invalidate(RoleResource.role(roleName));
        return true;
    }
```

test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
ekaterinadimitrova2: You are right, and this repeated run proves it - very rarely, but it can be flaky - https://app.circleci.com/pipelines/github/ekaterinadimitrova2/cassandra/2496/workflows/6ec742b1-7b6d-4ea9-b81a-36486da34e14/jobs/39036

One way is probably to use the Util#flakyTest to rerun it once if it fails?
Another is now when we have https://github.com/apache/cassandra/pull/2648/files#diff-02dc3c7a2a64aef7633ac085c24deddfed1b32b469c9958f2c04d5e228e6144fR42 to rely on that message but that one is not always generated and we don't check the particular roles were invalidated. Open for other ideas, too.

src/java/org/apache/cassandra/auth/CassandraCIDRAuthorizer.java
skoppu22: CassandraCIDRAuthorizer has two caches in it - CIDRPermissionsCache and CIDRGroupsMappingCache. Naming this function has() doesn't indicate which cache are we looking into. Also, the same function won't be available when AllowAllCIDRAuthorizer is used. I prefer moving this has() into CIDRPermissionsManager, so we can make this function available regardless of CIDR authorizer selected, also available for nodetool commands through CIDRPermissionsManager mbean. I mean similar to the way CIDRPermissionsManager::invalidateCidrPermissionsCache implemented. If you don't mind, I can make those changes and send them to you to include here.

src/java/org/apache/cassandra/auth/CIDRPermissionsCache.java
ekaterinadimitrova2: That's awesome, thank you @ben-manes ! I appreciate your feedback. Do you mind checking if there might be anything else I might have missed with this upgrade?

src/java/org/apache/cassandra/auth/CassandraCIDRAuthorizer.java
ekaterinadimitrova2: SGTM, thanks :-)

src/java/org/apache/cassandra/auth/CIDRPermissionsCache.java
ben-manes: lgtm, thanks for upgrading!

src/java/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCache.java
skoppu22: Yes, but this line 42 gets executed when args.isEmpty, i.e, user wants to invalidate entire cache, so cidrPermissionsCache.invalidateCidrPermissions(roleName) is not called in this case. cidrPermissionsCache.invalidateCidrPermissions(roleName) is called in the below else condition and we check for return value there - line 49.

test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
skoppu22: I was making the test depend only on table read count to avoid being flaky. Table read count tells us that a cache entry was loaded from the table, without tightly depending on when/how the entry was invalidated. I believe that should be sufficient in the unit test. Also, this particular check cache.getIfPresent(<role>) was only to print descriptive message to the user when nodetool command is run. It doesn't impact the functionality. So I wouldn't be comfortable creating a flaky test disturbing circleCI/Jenkins builds and impacting developers time/efforts just for this minor message.

src/java/org/apache/cassandra/auth/CassandraCIDRAuthorizer.java
skoppu22: We wouldn't need this has() function if we plan to not check for entry existence in the cache in unit tests to avoid flaky test. I will wait for your confirmation on the final approach before making changes.

ekaterinadimitrova2: I am fine if we do not use that method, but how do we confirm the validation? The tests are passing with/without the fix. I was looking to cover the invalidation fully and ensure that we do not regress in the future.

test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
ekaterinadimitrova2: `I believe that should be sufficient in the unit test. `
My only concern is that it wasn't to catch the bug we found here

src/java/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCache.java
ekaterinadimitrova2: Oh I see what you mean about the full cache invalidation

src/java/org/apache/cassandra/auth/CassandraCIDRAuthorizer.java
skoppu22: I am finding a way to control cache expiration in unit tests, either using ticker or larger cache validity value. Please give me some time, I will let you know my findings

ekaterinadimitrova2: Makes sense, thanks for looking into that

src/java/org/apache/cassandra/auth/CIDRPermissionsCache.java
jacek-lewandowski: please create `RoleResource.role` only once

src/java/org/apache/cassandra/metrics/ChunkCacheMetrics.java
```diff
@@ -72,14 +73,14 @@ public void recordLoadFailure(long loadTime)
...
+    public void recordEviction(int weight, RemovalCause cause)
     {
     }
 
     @Nonnull
     @Override
     public CacheStats snapshot()
     {
-        return new CacheStats(hits.getCount(), misses.getCount(), missLatency.getCount(), 0L, missLatency.getCount(), 0L, 0L);
+        return CacheStats.of(hits.getCount(), misses.getCount(), missLatency.getCount(), 0L, missLatency.getCount(), 0L, 0L);
```
jacek-lewandowski, line 84: please validate this snapshot - I suppose that what is passed as `loadSuccessCount` is not what it should be...

src/java/org/apache/cassandra/auth/CIDRPermissionsCache.java
ekaterinadimitrova2: good point, thanks

test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
skoppu22: Found a way. We can set high validity for cache, in the setup() before cache gets created. Rest of the setup() code remains as it is.

```
    @BeforeClass
    public static void setup() throws Exception
    {
        DatabaseDescriptor.setRolesValidity(Integer.MAX_VALUE-1);

```

Once we do that, cache's expireAfterWrite gets set to 2147483646 ms, which is 596+ hours, i.e, 24+ days. 

- We do not need has(role) function. We can just verify that `assertThat(tool.getStdout()).contains("Invalidated the role role_a from CIDR permissions cache")` after invalidating a role in the test testInvalidateSingleCidrPermission.

- Next unit test testInvalidateAllCidrPermissions doesn't go through the code path we are changing as we discussed in thread https://github.com/apache/cassandra/pull/2648/files#r1321766712

I will run the test under circle CI and let you know how it goes.

test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
```diff
@@ -142,11 +142,15 @@ public void testInvalidateAllCidrPermissions()
         assertThat(roleA.hasAccessFromIp(ipAddr)).isTrue();
         assertThat(roleB.hasAccessFromIp(ipAddr)).isTrue();
         assertThat(originalReadsCount).isEqualTo(getCidrPermissionsReadCount());
+        assertThat(((CassandraCIDRAuthorizer) DatabaseDescriptor.getCIDRAuthorizer()).has(ROLE_A)).isTrue();
```
skoppu22: As mentioned above, this unit test just invokes AuthCache's invalidate() function which re-inits the cache. So we don't need has(role) function.

test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
```diff
@@ -118,10 +115,13 @@ public void testInvalidateSingleCidrPermission()
         // ensure cidr permission is cached
         role.hasAccessFromIp(ipAddr);
         assertThat(originalReadsCount).isEqualTo(getCidrPermissionsReadCount());
+        assertThat(((CassandraCIDRAuthorizer) DatabaseDescriptor.getCIDRAuthorizer()).has(ROLE_A)).isTrue();
```
skoppu22: `"Invalidated the role role_a from CIDR permissions cache"` gets printed only when entry exists before invalidation. i.e, `assertThat(tool.getStdout()).contains("Invalidated the role role_a from CIDR permissions cache");` you added below at line 123 ensures that invalidation of a role was successful. So we do not need has(role) function.

src/java/org/apache/cassandra/auth/CassandraCIDRAuthorizer.java
skoppu22: Mentioned a possible solution here https://github.com/apache/cassandra/pull/2648/files#r1332985152, will run through circle CI and will confirm.

test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
ekaterinadimitrova2: Great, let me know how it goes with the CI. Can you also post a link to your patch so I can consolidate everything for the final patch, please? Thank you in advance.

src/java/org/apache/cassandra/metrics/ChunkCacheMetrics.java
ekaterinadimitrova2: I think you are right. According to the javadoc:

```
[loadSuccessCount](https://www.javadoc.io/static/com.github.ben-manes.caffeine/caffeine/2.9.2/com/github/benmanes/caffeine/cache/stats/CacheStats.html#loadSuccessCount--)()
Returns the *number of times* [Cache](https://www.javadoc.io/static/com.github.ben-manes.caffeine/caffeine/2.9.2/com/github/benmanes/caffeine/cache/Cache.html) lookup methods have successfully loaded a new value.
```
So it should be a number of times, but we provide latency/time... I am not sure why that was done this way. 
It seems this was added in 4.0+ in [CASSANDRA-14628](https://issues.apache.org/jira/browse/CASSANDRA-14628)
@iamaleksey, any comments?

test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
skoppu22: @ekaterinadimitrova2 here is the patch https://github.com/skoppu22/cassandra/pull/1/files

src/java/org/apache/cassandra/metrics/ChunkCacheMetrics.java
skoppu22: My understanding is: missLatency is recorded whenever there is a cache miss and an entry is loaded into the cache. missLatency.getCount() returns number of times successful cache load happened. Which seems correct based on this understanding.

skoppu22:
+1 LGTM

skoppu22 (approved):

src/java/org/apache/cassandra/metrics/ChunkCacheMetrics.java
jacek-lewandowski: you are right, though `missLatency.getCount()` is also used for load time, and in this case it is probably wrong; anyway, it is out of the scope

jacek-lewandowski (approved):

test/unit/org/apache/cassandra/tools/nodetool/InvalidateCIDRPermissionsCacheTest.java
```diff
@@ -20,13 +20,10 @@
 
 import java.net.InetSocketAddress;
 
+import org.apache.cassandra.auth.*;
```
yifan-c, line 23: I thought we do not allow wild import, but it was my false memory :)

ekaterinadimitrova2: Good point, it looks like our checkstyle does not enforce that, at least not in tests.
I will replace the import on commit with:
```
import org.apache.cassandra.auth.AuthCacheService;
import org.apache.cassandra.auth.AuthTestUtils;
import org.apache.cassandra.auth.AuthenticatedUser;
import org.apache.cassandra.auth.IRoleManager;
```

ekaterinadimitrova2:
This was merged

src/java/org/apache/cassandra/metrics/ChunkCacheMetrics.java
ekaterinadimitrova2: I will open a follow-up ticket

ekaterinadimitrova2: [CASSANDRA-18908](https://issues.apache.org/jira/browse/CASSANDRA-18908)

PR #2649, opened 2023-08-30
CASSANDRA-18781 Add guardrail to be able to disable bulk load of SSTables

Runtian:
```
Add config to be able to disable bulk load

```

The [Cassandra Jira](https://issues.apache.org/jira/browse/CASSANDRA-18781)

### DIFF
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -1889,6 +1889,9 @@ drop_compact_storage_enabled: false
 # Guardrail to allow/disallow DROP KEYSPACE statements
 # drop_keyspace_enabled: true
 #
+# Guardrail to allow/disallow bulk load of SSTables
+# bulk_load_enabled: true
+#
 # Guardrail to warn or fail when using a page size greater than threshold.
 # The two thresholds default to -1 to disable.
 # page_size_warn_threshold: -1
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -860,6 +860,7 @@ public class Config
     public volatile boolean user_timestamps_enabled = true;
     public volatile boolean alter_table_enabled = true;
     public volatile boolean group_by_enabled = true;
+    public volatile boolean bulk_load_enabled = true;
     public volatile boolean drop_truncate_table_enabled = true;
     public volatile boolean drop_keyspace_enabled = true;
     public volatile boolean secondary_indexes_enabled = true;
--- a/src/java/org/apache/cassandra/config/GuardrailsOptions.java
+++ b/src/java/org/apache/cassandra/config/GuardrailsOptions.java
@@ -365,6 +365,20 @@ public class GuardrailsOptions implements GuardrailsConfig
     }
 
     @Override
+    public boolean getBulkLoadEnabled()
+    {
+        return config.bulk_load_enabled;
+    }
+
+    public void setBulkLoadEnabled(boolean enabled)
+    {
+        updatePropertyWithLogging("bulk_load_enabled",
+                                  enabled,
+                                  () -> config.bulk_load_enabled,
+                                  x -> config.bulk_load_enabled = x);
+    }
+
+    @Override
     public boolean getSecondaryIndexesEnabled()
     {
         return config.secondary_indexes_enabled;
--- a/src/java/org/apache/cassandra/db/guardrails/Guardrails.java
+++ b/src/java/org/apache/cassandra/db/guardrails/Guardrails.java
@@ -185,6 +185,15 @@ public final class Guardrails implements GuardrailsMBean
                     "DROP KEYSPACE functionality");
 
     /**
+     * Guardrail disabling bulk loading of SSTables
+     */
+    public static final EnableFlag bulkLoadEnabled =
+    (EnableFlag) new EnableFlag("bulk_load_enabled",
+                   "Bulk loading of SSTables might potentially destabilize the node.",
+                   state -> CONFIG_PROVIDER.getOrCreate(state).getBulkLoadEnabled(),
+                   "Bulk loading of SSTables").throwOnNullClientState(true);
+
+    /**
      * Guardrail disabling user's ability to turn off compression
      */
     public static final EnableFlag uncompressedTablesEnabled =
@@ -777,6 +786,18 @@ public final class Guardrails implements GuardrailsMBean
     public void setDropKeyspaceEnabled(boolean enabled)
     {
         DEFAULT_CONFIG.setDropKeyspaceEnabled(enabled);
+    }
+
+    @Override
+    public boolean getBulkLoadEnabled()
+    {
+        return DEFAULT_CONFIG.getBulkLoadEnabled();
+    }
+
+    @Override
+    public void setBulkLoadEnabled(boolean enabled)
+    {
+        DEFAULT_CONFIG.setBulkLoadEnabled(enabled);
     }
 
     @Override
--- a/src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
+++ b/src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
@@ -176,6 +176,13 @@ public interface GuardrailsConfig
     boolean getDropKeyspaceEnabled();
 
     /**
+     * Returns whether bulk load of SSTables is allowed
+     *
+     * @return {@code true} if allowed, {@code false} otherwise.
+     */
+    boolean getBulkLoadEnabled();
+
+    /**
      * @return The threshold to warn when page size exceeds given size.
      */
     int getPageSizeWarnThreshold();
--- a/src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
+++ b/src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
@@ -317,6 +317,18 @@ public interface GuardrailsMBean
     void setDropKeyspaceEnabled(boolean enabled);
 
     /**
+     * Returns whether bulk load of SSTables is allowed
+     *
+     * @return {@code true} if allowed, {@code false} otherwise.
+     */
+    boolean getBulkLoadEnabled();
+
+    /**
+     * Sets whether bulk load of SSTables is allowed
+     */
+    void setBulkLoadEnabled(boolean enabled);
+
+    /**
      * @return The threshold to warn when requested page size greater than threshold.
      * -1 means disabled.
      */
--- a/src/java/org/apache/cassandra/streaming/StreamDeserializingTask.java
+++ b/src/java/org/apache/cassandra/streaming/StreamDeserializingTask.java
@@ -22,6 +22,7 @@ import com.google.common.annotations.VisibleForTesting;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.cassandra.db.guardrails.Guardrails;
 import org.apache.cassandra.locator.InetAddressAndPort;
 import org.apache.cassandra.streaming.messages.KeepAliveMessage;
 import org.apache.cassandra.streaming.messages.StreamMessage;
@@ -69,6 +70,9 @@ public class StreamDeserializingTask implements Runnable
 
                 if (session == null)
                     session = deriveSession(message);
+
+                if (session.getStreamOperation() == StreamOperation.BULK_LOAD)
+                    Guardrails.bulkLoadEnabled.ensureEnabled("Bulk load of SSTables", null);
 
                 if (logger.isDebugEnabled())
                     logger.debug("{} Received {}", createLogTag(session, channel), message);
--- a/test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
+++ b/test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
@@ -1,0 +1,179 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.distributed.test.guardrails;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.List;
+
+import org.apache.commons.io.FileUtils;
+import org.junit.AfterClass;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.db.Keyspace;
+import org.apache.cassandra.db.guardrails.Guardrails;
+import org.apache.cassandra.distributed.Cluster;
+import org.apache.cassandra.io.util.File;
+import org.apache.cassandra.service.StorageService;
+import org.apache.cassandra.tools.BulkLoader;
+import org.apache.cassandra.tools.ToolRunner;
+
+import static com.google.common.collect.Lists.transform;
+import static java.util.stream.Collectors.toList;
+import static org.apache.cassandra.db.ColumnFamilyStore.FlushReason.UNIT_TESTS;
+import static org.apache.cassandra.distributed.api.Feature.GOSSIP;
+import static org.apache.cassandra.distributed.api.Feature.NATIVE_PROTOCOL;
+import static org.apache.cassandra.distributed.api.Feature.NETWORK;
+import static org.apache.cassandra.distributed.shared.AssertUtils.assertRows;
+import static org.apache.cassandra.distributed.shared.AssertUtils.row;
+import static org.apache.cassandra.distributed.test.ExecUtil.rethrow;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests the guardrail for bulk load, {@link Guardrails#bulkLoadEnabled}.
+ */
+public class GuardrailBulkLoadEnabledTest extends GuardrailTester
+{
+    private static Path tempDir;
+    private static Cluster cluster;
+    private static String nodes;
+    private static int nativePort;
+    private static int storagePort;
+
+    public static Path tempDir() throws Exception
+    {
+        return Files.createTempDirectory("GuardrailBulkLoadEnabledTest");
+    }
+
+    @Override
+    protected Cluster getCluster()
+    {
+        return cluster;
+    }
+
+    @BeforeClass
+    public static void setupCluster() throws Exception
+    {
+        tempDir = tempDir();
+        cluster = init(Cluster.build(1).withConfig(c -> c.with(NATIVE_PROTOCOL, NETWORK, GOSSIP)).start());
+        nodes = cluster.get(1).config().broadcastAddress().getHostString();
+        nativePort = cluster.get(1).callOnInstance(DatabaseDescriptor::getNativeTransportPort);
+        storagePort = cluster.get(1).callOnInstance(DatabaseDescriptor::getStoragePort);
+    }
+
+    @AfterClass
+    public static void teardownCluster()
+    {
+        if (cluster != null)
+            cluster.close();
+
+        for (File f : new File(tempDir).tryList())
+        {
+            f.tryDelete();
+        }
+    }
+
+    @Test
+    public void bulkLoaderEnabled() throws Throwable
+    {
+        File sstablesToUpload = prepareSstablesForUpload();
+        // bulk load SSTables work as expected
+        ToolRunner.ToolResult tool = loadData(sstablesToUpload);
+        tool.assertOnCleanExit();
+        assertTrue(tool.getStdout().contains("Summary statistics"));
+        assertRows(cluster.get(1).executeInternal("SELECT count(*) FROM bulk_load_tables.test"), row(22L));
+
+        // truncate table
+        truncateGeneratedTables();
+        assertRows(cluster.get(1).executeInternal("SELECT * FROM bulk_load_tables.test"), EMPTY_ROWS);
+
+        // Disable bulk load, stream should fail and no data should be loaded
+        cluster.get(1).runOnInstance(() -> Guardrails.instance.setBulkLoadEnabled(false));
+        long mark = cluster.get(1).logs().mark();
+        tool = loadData(sstablesToUpload);
+
+        cluster.get(1).logs().watchFor(mark, "Bulk load of SSTables is not allowed");
+        tool.asserts().failure().errorContains("Stream failed");
+        assertRows(cluster.get(1).executeInternal("SELECT * FROM bulk_load_tables.test"), EMPTY_ROWS);
+
+        // Enable bulk load again, data should be loaded successfully
+        cluster.get(1).runOnInstance(() -> Guardrails.instance.setBulkLoadEnabled(true));
+        tool = loadData(sstablesToUpload);
+        tool.assertOnCleanExit();
+        assertTrue(tool.getStdout().contains("Summary statistics"));
+        assertRows(cluster.get(1).executeInternal("SELECT count(*) FROM bulk_load_tables.test"), row(22L));
+    }
+
+    private static ToolRunner.ToolResult loadData(File sstablesToUpload)
+    {
+        return ToolRunner.invokeClass(BulkLoader.class,
+                                      "--nodes", nodes,
+                                      "--port", Integer.toString(nativePort),
+                                      "--storage-port", Integer.toString(storagePort),
+                                      sstablesToUpload.absolutePath());
+    }
+
+    private static File prepareSstablesForUpload() throws IOException
+    {
+        generateSSTables();
+        File sstableDir = copySStablesFromDataDir("test");
+        truncateGeneratedTables();
+        return sstableDir;
+    }
+
+    private static File copySStablesFromDataDir(String table) throws IOException
+    {
+        File cfDir = new File(tempDir +  File.pathSeparator() + "bulk_load_tables" + File.pathSeparator() + table);
+        cfDir.tryCreateDirectories();
+        List<String> keyspaceDirPaths = cluster.get(1).callOnInstance(
+        () -> Keyspace.open("bulk_load_tables")
+                      .getColumnFamilyStore(table)
+                      .getDirectories()
+                      .getCFDirectories()
+                      .stream()
+                      .map(File::absolutePath)
+                      .collect(toList())
+        );
+        for (File srcDir : transform(keyspaceDirPaths, File::new))
+        {
+            for (File file : srcDir.tryList(File::isFile))
+                FileUtils.copyFileToDirectory(file.toJavaIOFile(), cfDir.toJavaIOFile());
+        }
+        return cfDir;
+    }
+
+    private static void generateSSTables()
+    {
+        cluster.schemaChange("CREATE KEYSPACE IF NOT EXISTS bulk_load_tables WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}");
+        cluster.schemaChange("CREATE TABLE IF NOT EXISTS bulk_load_tables.test (pk int, val text, PRIMARY KEY (pk))");
+        for (int i = 0; i < 22; i++)
+        {
+            cluster.get(1).executeInternal(String.format("INSERT INTO bulk_load_tables.test (pk, val) VALUES (%s, '%s')", i, i));
+        }
+        cluster.get(1).runOnInstance(rethrow(() -> StorageService.instance.forceKeyspaceFlush("bulk_load_tables", UNIT_TESTS)));
+    }
+
+    private static void truncateGeneratedTables()
+    {
+        cluster.get(1).executeInternal("TRUNCATE bulk_load_tables.test");
+    }
+}
--- a/src/java/org/apache/cassandra/db/guardrails/Guardrail.java
+++ b/src/java/org/apache/cassandra/db/guardrails/Guardrail.java
@@ -64,6 +64,9 @@ public abstract class Guardrail
     /** Time of last failure in milliseconds. */
     private volatile long lastFailInMs = 0;
 
+    /** Should throw exception if null client state is provided. */
+    private volatile boolean throwOnNullClientState = false;
+
     Guardrail(String name, @Nullable String reason)
     {
         this.name = name;
@@ -136,7 +139,7 @@ public abstract class Guardrail
             GuardrailsDiagnostics.failed(name, decorateMessage(redactedMessage));
         }
 
-        if (state != null)
+        if (state != null || throwOnNullClientState)
             throw new GuardrailViolatedException(message);
     }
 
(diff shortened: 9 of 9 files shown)


src/java/org/apache/cassandra/config/Config.java
```diff
@@ -397,6 +397,8 @@ public static class SSTableConfig
     @Replaces(oldName = "max_mutation_size_in_kb", converter = Converters.KIBIBYTES_DATASTORAGE, deprecated = true)
     public DataStorageSpec.IntKibibytesBound max_mutation_size;
 
+    public boolean bulk_load_enabled = true;
```
smiklosovic: @Runtian you should introduce and document this property in `cassandra.yml` as well

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -3974,6 +3974,17 @@ public static void setGCWarnThreshold(int threshold)
         conf.gc_warn_threshold = new DurationSpec.IntMillisecondsBound(threshold);
     }
 
+    public static boolean isBulkLoadEnabled()
+    {
+        return conf.bulk_load_enabled;
+    }
+
+    @VisibleForTesting
+    public static void setBulkLoadEnabled(boolean bulk_load_enabled)
```
smiklosovic: do not use snake-case here, just use camel-case, `boolean bulkLoadEnabled`

src/java/org/apache/cassandra/streaming/StreamDeserializingTask.java
```diff
@@ -106,6 +107,11 @@ public StreamSession deriveSession(StreamMessage message)
         // to be established a priori
         StreamSession streamSession = message.getOrCreateAndAttachInboundSession(channel, messagingVersion);
 
+        if (streamSession.getStreamOperation() == StreamOperation.BULK_LOAD && !DatabaseDescriptor.isBulkLoadEnabled())
+        {
+            throw new RuntimeException("Bulk Load is disabled");
```
smiklosovic: `Bulk load is disabled`.

src/java/org/apache/cassandra/config/Config.java
smiklosovic: also, I think this variable should be `volatile`.

conf/cassandra.yaml
```diff
@@ -1889,6 +1889,9 @@ drop_compact_storage_enabled: false
 # Guardrail to allow/disallow DROP KEYSPACE statements
 # drop_keyspace_enabled: true
 #
+# Guardrail to allow/disallow BULK LOAD
```
smiklosovic: why do you use capitals? Why is not "bulk load" enough?

Guardrail to allow/disallow bulk loading of SSTables

We should mention what we actually want to bulk load ... bulk load of what?

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -397,6 +397,8 @@ public static class SSTableConfig
     @Replaces(oldName = "max_mutation_size_in_kb", converter = Converters.KIBIBYTES_DATASTORAGE, deprecated = true)
     public DataStorageSpec.IntKibibytesBound max_mutation_size;
 
+    public volatile boolean bulk_load_enabled = true;
```
smiklosovic: There is a section down below in this file where all guardrail configuration fields are specified. Could you move it down there so they are all logically coupled together?

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -184,6 +184,15 @@ public final class Guardrails implements GuardrailsMBean
                     state -> CONFIG_PROVIDER.getOrCreate(state).getDropKeyspaceEnabled(),
                     "DROP KEYSPACE functionality");
 
+    /**
+     * Guardrail disabling BULK LOAD behavior
```
smiklosovic: Guardrail disabling bulk loading of SSTables

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -184,6 +184,15 @@ public final class Guardrails implements GuardrailsMBean
...
+     */
+    public static final EnableFlag bulkLoadEnabled =
+    new EnableFlag("bulk_load_enabled",
+                   null,
+                   state -> CONFIG_PROVIDER.getOrCreate(state).getBulkLoadEnabled(),
+                   "BULK LOAD functionality");
```
smiklosovic: "bulk load", maybe we can be more descriptive here like:

"bulk loading of SSTables"

src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
```diff
@@ -175,6 +175,13 @@ public interface GuardrailsConfig
      */
     boolean getDropKeyspaceEnabled();
 
+    /**
+     * Returns whether BULK LOAD is allowed
```
smiklosovic: bulk load of SSTables

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -311,6 +311,18 @@ public interface GuardrailsMBean
      */
     boolean getDropKeyspaceEnabled();
 
+    /**
+     * Sets whether BULK LOAD is allowed
```
smiklosovic: same and next method again

test/unit/org/apache/cassandra/streaming/async/StreamingInboundHandlerTest.java
```diff
@@ -105,13 +108,53 @@ public void StreamDeserializingTask_deriveSession_StreamInitMessage()
...
+    @Test
+    public void StreamDeserializingTask_deriveSession_BulkLoad_Disabled_Do_Not_Affect_Non_Bulk_Streaming() throws Exception
+    {
+        try
+        {
+            Guardrails.instance.setBulkLoadEnabled(false);
+            // verify that other operation is not affected
+            StreamInitMessage msg = new StreamInitMessage(REMOTE_ADDR, 0, nextTimeUUID(), StreamOperation.REPAIR, nextTimeUUID(), PreviewKind.ALL);
+            StreamDeserializingTask task = new StreamDeserializingTask(null, streamingChannel, MessagingService.current_version);
+            StreamSession session = task.deriveSession(msg);
```
smiklosovic: can be done like `Assert.assertNotNull(task.deriveSession(msg))`

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,144 @@
...
+import static com.google.common.collect.Lists.transform;
+import static org.apache.cassandra.distributed.shared.AssertUtils.assertRows;
+import static org.apache.cassandra.distributed.shared.AssertUtils.row;
+import static org.apache.cassandra.distributed.test.ExecUtil.rethrow;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests the guardrails for bulk load, {@link Guardrails#bulkLoadEnabled}.
+ */
+public class GuardrailBulkLoadEnabledTest extends GuardrailTester
```
smiklosovic, line 54: You are not testing that enabled bulk loading is loading sstables and that we can enable and disable (go back and forth) You only test that disabled bulk loading is not loading them.

I would do:

1) start a node
2) create tables, add data, backup sstables, truncate
3) bulk load sstables and verify data are there, truncate again
4) disable bulk load (by calling a respective method on `Guardrails` on server side)
5) bulk load sstables and verify that it is not loading it
6) enable bulk load by same mean as in 4)
7) load sstables and verify it is loading it

4-7 is important as we know that we can actually go with enable -> disable -> enable so we can alter this configuration and it works as expected.

src/java/org/apache/cassandra/streaming/StreamDeserializingTask.java
```diff
@@ -109,6 +111,12 @@ public StreamSession deriveSession(StreamMessage message)
         // Attach this channel to the session: this only happens upon receiving the first init message as a follower;
         // in all other cases, no new control channel will be added, as the proper control channel will be already attached.
         streamSession.attachInbound(channel);
+
+        if (streamSession.getStreamOperation() == StreamOperation.BULK_LOAD && !Guardrails.bulkLoadEnabled.isEnabled(null))
```
smiklosovic: We tend to omit braces when bodies of `if`s contain just one statement. In this case you do not need braces here.

test/unit/org/apache/cassandra/streaming/async/StreamingInboundHandlerTest.java
```diff
@@ -105,13 +108,53 @@ public void StreamDeserializingTask_deriveSession_StreamInitMessage()
...
+                      .as("Derive session should fail");
+        }
+        finally
+        {
+            Guardrails.instance.setBulkLoadEnabled(true);
+        }
+    }
+
+    @Test
+    public void StreamDeserializingTask_deriveSession_BulkLoad_Disabled_Do_Not_Affect_Non_Bulk_Streaming() throws Exception
```
smiklosovic: should be actually `Disabled_Does_Not_Affect`

Do_Not -> Does_Not

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,144 @@
...
+    public static void teardownCluster()
+    {
+        if (CLUSTER != null)
+            CLUSTER.close();
+    }
+
+    @Test
+    public void bulkLoaderFailedWithBulkLoadDisabled() throws Throwable
+    {
+        File sstables_to_upload = prepareSstablesForUpload();
```
smiklosovic: Why are you using snake case? Should be camel case as everywhere, there are multiple instances of this happening in this file.

smiklosovic:
@Runtian the current code does not pass `ant check`. Your code contains unused imports both in production as well as in tests.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,171 @@
...
+    {
+        generateSstables();
+        File sstable_dir = copySstablesFromDataDir("test");
+        truncateGeneratedTables();
+        return sstable_dir;
+    }
+
+    private static File copySstablesFromDataDir(String table) throws IOException
+    {
+        File cfDir = new File("build/test/cassandra/bulk_load_tables", table);
```
smiklosovic: @Runtian please move this to a temporary directory. Also, if we are going to run this test in a multiplexer (e.g. 500 times in a loop to see if a test is stable), you are not deleting this directory after the test. That means that we are polluting the disk space unnecessarily.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,171 @@
...
+                                      sstablesToUpload.absolutePath());
+        tool.assertOnCleanExit();
+        assertTrue(tool.getStdout().contains("Summary statistics"));
+        assertRows(CLUSTER.get(1).executeInternal("SELECT count(*) FROM bulk_load_tables.test"), row(22L));
+    }
+
+    private static File prepareSstablesForUpload() throws IOException
+    {
+        generateSstables();
+        File sstable_dir = copySstablesFromDataDir("test");
```
smiklosovic: @Runtian this is still on snake case

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,171 @@
...
+    {
+        return CLUSTER;
+    }
+
+    @BeforeClass
+    public static void setupCluster() throws IOException
+    {
+        CLUSTER = init(Cluster.build(1)
+                              .withConfig(c -> {
+                                  c.with(Feature.NATIVE_PROTOCOL, Feature.NETWORK, Feature.GOSSIP);
```
smiklosovic: could you statically import these enums? You also do not need to wrap this in `{` and `}` as that closure is a one liner.

If you do all of that, you realize that whole  `build(1).withConfig(c -> c.with(...).start()` can fit on one line below 120 characters. You might statically import `build` method as well.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,171 @@
...
+        if (CLUSTER != null)
+            CLUSTER.close();
+    }
+
+    @Test
+    public void bulkLoaderEnabled() throws Throwable
+    {
+        File sstablesToUpload = prepareSstablesForUpload();
+        // bulk load SSTables work as expected
+        ToolRunner.ToolResult tool = ToolRunner.invokeClass(BulkLoader.class,
```
smiklosovic: I would create a helper method for this. You are calling it 3 times with same parameters.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,171 @@
...
+        File sstable_dir = copySstablesFromDataDir("test");
+        truncateGeneratedTables();
+        return sstable_dir;
+    }
+
+    private static File copySstablesFromDataDir(String table) throws IOException
+    {
+        File cfDir = new File("build/test/cassandra/bulk_load_tables", table);
+        cfDir.tryCreateDirectories();
+        List<String> keyspaceDirPaths = CLUSTER.get(1).callOnInstance(() -> {
```
smiklosovic: you dont need `transform` for this:

        CLUSTER.get(1).callOnInstance(() -> Keyspace.open("bulk_load_tables")
                                                               .getColumnFamilyStore("test")
                                                               .getDirectories()
                                                               .getCFDirectories()
                                                               .stream()
                                                               .map(File::absolutePath)
                                                               .collect(toList()));

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,171 @@
...
+            List<File> cfDirs = Keyspace.open("bulk_load_tables").getColumnFamilyStore(table).getDirectories().getCFDirectories();
+            return transform(cfDirs, (d) -> d.absolutePath());
+        });
+        for (File srcDir : transform(keyspaceDirPaths, (p) -> new File(p)))
+        {
+            for (File file : srcDir.tryList((file) -> file.isFile()))
+            {
+                FileUtils.copyFileToDirectory(file.toJavaIOFile(), cfDir.toJavaIOFile());
```
smiklosovic, line 159: the rule of "one liner does not need to be wrapped in braces" holds for `for`s too.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,171 @@
...
+    {
+        File cfDir = new File("build/test/cassandra/bulk_load_tables", table);
+        cfDir.tryCreateDirectories();
+        List<String> keyspaceDirPaths = CLUSTER.get(1).callOnInstance(() -> {
+            List<File> cfDirs = Keyspace.open("bulk_load_tables").getColumnFamilyStore(table).getDirectories().getCFDirectories();
+            return transform(cfDirs, (d) -> d.absolutePath());
+        });
+        for (File srcDir : transform(keyspaceDirPaths, (p) -> new File(p)))
+        {
+            for (File file : srcDir.tryList((file) -> file.isFile()))
```
smiklosovic: `tryList(File::isFile)`

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,171 @@
...
+    private static void generateSstables() throws IOException
+    {
+        CLUSTER.schemaChange("CREATE KEYSPACE IF NOT EXISTS bulk_load_tables WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}");
+        CLUSTER.schemaChange("CREATE TABLE IF NOT EXISTS bulk_load_tables.test (pk int, val text, PRIMARY KEY (pk))");
+        for (int i = 0; i < 22; i++)
+        {
+            CLUSTER.get(1).executeInternal(String.format("INSERT INTO bulk_load_tables.test (pk, val) VALUES (%s, '%s')", i, i));
+        }
+        CLUSTER.get(1).runOnInstance(rethrow(() -> StorageService.instance.forceKeyspaceFlush("bulk_load_tables",
+                                                                                              ColumnFamilyStore.FlushReason.UNIT_TESTS)));
```
smiklosovic: If you statically import `UNIT_TESTS` you might fit it on one line.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,171 @@
...
+        {
+            for (File file : srcDir.tryList((file) -> file.isFile()))
+            {
+                FileUtils.copyFileToDirectory(file.toJavaIOFile(), cfDir.toJavaIOFile());
+            }
+        }
+        return cfDir;
+    }
+
+    private static void generateSstables() throws IOException
```
smiklosovic: This method actually does not throw `IOException`. It does not need to throw anything.

smiklosovic: could you rename this to`generateSSTables` ?

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+    private static File prepareSstablesForUpload() throws IOException
+    {
+        generateSstables();
+        File sstable_dir = copySstablesFromDataDir("test");
+        truncateGeneratedTables();
+        return sstable_dir;
+    }
+
+    private static File copySstablesFromDataDir(String table) throws IOException
```
smiklosovic: `copySStablesFromDataDir`

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -364,6 +364,19 @@ public void setDropKeyspaceEnabled(boolean enabled)
                                   x -> config.drop_keyspace_enabled = x);
     }
 
+    public boolean getBulkLoadEnabled()
```
adelapena, line 368: Nit: add `@Override`

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -184,6 +184,15 @@ public final class Guardrails implements GuardrailsMBean
                     state -> CONFIG_PROVIDER.getOrCreate(state).getDropKeyspaceEnabled(),
                     "DROP KEYSPACE functionality");
 
+    /**
+     * Guardrail disabling bulk loading of SSTables
+     */
+    public static final EnableFlag bulkLoadEnabled =
+    new EnableFlag("bulk_load_enabled",
+                   null,
```
adelapena: Can we add a reason for disabling bulk loading instead of `null`?

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -311,6 +311,18 @@ public interface GuardrailsMBean
      */
     boolean getDropKeyspaceEnabled();
 
+    /**
+     * Sets whether bulk load of SSTables is allowed
+     */
+    void setBulkLoadEnabled(boolean enabled);
```
adelapena: The setter/getters for `bulk_load_enabled` are inserted between the getters and setters for `drop_keyspace_enabled`, setters and getters for the same property should remain together. Also, on this file we usually write first the getter, then the setter.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -184,6 +184,15 @@ public final class Guardrails implements GuardrailsMBean
...
+                   state -> CONFIG_PROVIDER.getOrCreate(state).getBulkLoadEnabled(),
+                   "bulk loading of SSTables");
```
adelapena:
```suggestion
                   "Bulk loading of SSTables");
```

src/java/org/apache/cassandra/streaming/StreamDeserializingTask.java
```diff
@@ -109,6 +110,10 @@ public StreamSession deriveSession(StreamMessage message)
...
+            throw new StreamReceiveException(streamSession, new RuntimeException("Bulk load is disabled"));
```
adelapena: This isn't a proper usage of guardrails. They should be checked with `ensureEnabled()`, so they throw the formatted `GuardrailViolatedException` defined in their constructor. That would also trigger other associated behaviours, like emitting a log message and a diagnostic event.

Instead, the usage of `EnableFlag#isEnabled` is ignoring all that and using it as a regular boolean config property.

An issue here might be that we don't have a `ClientState`, and guardrails don't throw an exception but a warning on `null` client states: https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/db/guardrails/Guardrail.java#L139-L140

I think we could add a new `Guardrail#throwOnNullClientState()` method to instruct guardrails to throw an exception when triggered with a null `ClientState`. We usually don't throw on null client states to not break background processes such as compaction, but it can make sense for unauthenticated tools like this one.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
Runtian: Do you have any suggestions on the reason for this feature?

adelapena: That should be the reason why this guardrail exists, to be displayed in the error message. It should be something of the form `Bulk loading is risky because in X conditions Y might happen`, or something like that.

Runtian: This is different from other features as this is disabling the bulk load streaming feature. There is no client state for this and the error message won't be displayed.

adelapena: With the last changes that use `ensureEnabled` the error message with the reason is included in the message, isn't it?

Runtian:
@smiklosovic I moved the check part out of the `deriveSession` function so we always have a session when throwing error. I removed the unit test since we have the jvm dtest to test this feature and it is not easy to test if it is moved out of the `deriveSession` function.

src/java/org/apache/cassandra/streaming/StreamDeserializingTask.java
Runtian: Made the change please review it again. Thanks.

src/java/org/apache/cassandra/db/guardrails/EnableFlag.java
```diff
@@ -117,13 +117,33 @@ public void ensureEnabled(@Nullable ClientState state)
...
+     * @param featureName            The feature that is guarded by this guardrail (for reporting in error messages).
+     * @param state                  The client state, used to skip the check if the query is internal or is done by a
+     *                               superuser. A {@code null} value means that the check should be done regardless of
+     *                               the query, although it won't throw any exception if the failure threshold isexceeded.
+     *                               This is so because checks without an associated client come from asynchronous processes
+     *                               such as compaction, and we don't want to interrupt such processes.
+     * @param throwOnNullClientState If this is true, an exception will be thrown even if client state is null. This is
+     *                               mainly used for features not assoiate with client state such as bulk load of SSTables.
+     */
+    public void ensureEnabled(String featureName, @Nullable ClientState state, boolean throwOnNullClientState)
```
adelapena: I think this boolean flag shouldn't be added in an overload of `ensureEnabled`, but set once with a method analogous to `Guardrail#minNotifyIntervalInMs(long)`, probably `Guardrail#throwOnNullClientState(boolen)`. 

Same as `minNotifyIntervalInMs`, that method would be called once in the `Guardrails` class, so it's defined once for the guardrail instance.

Adding that on the `Guardrail` class would have the benefit of working for all its subclasses, without the need to modify them with additional method overloads.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
Runtian: I mean the reason won't be sent back to the "client" because there is not client state. Client will only get the "stream failed" message.

Runtian: The reason for disabling bulk load is solely the decision of the Cassandra operator. In our situation, we want to activate this feature only for users who truly require it.

src/java/org/apache/cassandra/db/guardrails/EnableFlag.java
Runtian: sure, makes sense.

src/java/org/apache/cassandra/streaming/StreamDeserializingTask.java
smiklosovic: @adelapena thank you for catching this! It is easy to overlook the difference between `isEnabled` and `ensureEnabled` without deeper look when doing a review just from a browser. On face value it looks like it does the correct job, even returns `boolean`.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
smiklosovic: @Runtian you just wrote the message which should be present in an exception thrown when it is disabled :)

`Bulk loading of SSTables was disabled by Cassandra operator. We activate this feature only for users who truly require it to not make nodes potentially unstable.`

Something like that ... 

Is not it true that guardrail's exception will be propagated to StreamSession#onError? You can probably append `t.getMessage()` via SessionFailedMessage

        if (channel.connected())
        {
            state(State.FAILED); // make sure subsequent error handling sees the session in a final state 
            channel.sendControlMessage(new SessionFailedMessage()).awaitUninterruptibly();
        }

`new SessionFailedMessage(t)`

and then append it in toString when it is not null.

Is @adelapena cool with that?


However, I am not sure how that would work with serialization / deserialization logic. It would probably not work ...

But I think that description is still valuable to have. Maybe in future all these guardrails will be present somewhere for a user so he might see what each guardrail does and what it is good for.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -189,10 +189,15 @@ public final class Guardrails implements GuardrailsMBean
      */
     public static final EnableFlag bulkLoadEnabled =
     new EnableFlag("bulk_load_enabled",
-                   null,
+                   "Bulk loading of SSTables might potentially destabilize the node.",
                    state -> CONFIG_PROVIDER.getOrCreate(state).getBulkLoadEnabled(),
                    "Bulk loading of SSTables");
 
+    static
```
smiklosovic: @Runtian I dont like this. Can't you just call that method as you instantiate `bulkLoadEnabled`? Like this:

    new EnableFlag("bulk_load_enabled",
                   null,
                   "Bulk loading of SSTables might potentially destabilize the node.",
                   state -> CONFIG_PROVIDER.getOrCreate(state).getBulkLoadEnabled(),
                   "Bulk loading of SSTables").throwOnNullClientState(true);

That method returns `this`. There is no reason to not call it on that.

adelapena: +1, although it would require a trivial cast to `EnableFlag` since `Guardrail#throwOnNullClientState` returns a `Guardrail`:
```java
public static final EnableFlag bulkLoadEnabled =
(EnableFlag) new EnableFlag("bulk_load_enabled",
                            "Bulk loading of SSTables might potentially destabilize the node.",
                            state -> CONFIG_PROVIDER.getOrCreate(state).getBulkLoadEnabled(),
                            "Bulk loading of SSTables").throwOnNullClientState(true);
```

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,179 @@
...
+import static org.apache.cassandra.distributed.api.Feature.GOSSIP;
+import static org.apache.cassandra.distributed.api.Feature.NATIVE_PROTOCOL;
+import static org.apache.cassandra.distributed.api.Feature.NETWORK;
+import static org.apache.cassandra.distributed.shared.AssertUtils.assertRows;
+import static org.apache.cassandra.distributed.shared.AssertUtils.row;
+import static org.apache.cassandra.distributed.test.ExecUtil.rethrow;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests the guardrails for bulk load, {@link Guardrails#bulkLoadEnabled}.
```
adelapena:
```suggestion
 * Tests the guardrail for bulk load, {@link Guardrails#bulkLoadEnabled}.
```

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,179 @@
...
+ */
+public class GuardrailBulkLoadEnabledTest extends GuardrailTester
+{
+    private static Path tempDir;
+    private static Cluster CLUSTER;
+    static String NODES;
+    static int NATIVE_PORT;
+    static int STORAGE_PORT;
```
adelapena: Nit: These aren't constants, so they should be camelCased instead of upper case. They can also be `private`.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailBulkLoadEnabledTest.java
```diff
@@ -0,0 +1,179 @@
...
+        List<String> keyspaceDirPaths = CLUSTER.get(1).callOnInstance(
+        () -> Keyspace.open("bulk_load_tables")
+                      .getColumnFamilyStore(table)
+                      .getDirectories()
+                      .getCFDirectories()
+                      .stream()
+                      .map(File::absolutePath)
+                      .collect(toList())
+        );
+        for (File srcDir : transform(keyspaceDirPaths, (p) -> new File(p)))
```
adelapena:
```suggestion
        for (File srcDir : transform(keyspaceDirPaths, File::new))
```

conf/cassandra.yaml
```diff
@@ -1889,6 +1889,9 @@ drop_compact_storage_enabled: false
 # Guardrail to allow/disallow DROP KEYSPACE statements
 # drop_keyspace_enabled: true
 #
+# Guardrail to allow/disallow bulk load of SSTables
+# bulk_load_enabled: true
```
adelapena, line 1893: Can we add a note about the new guardrail in `NEWS.txt`? It seems it would be the very first new feature in 5.1 :)

It could look like:
```
5.1
===

New features
------------
    - New Guardrails added:
      - Whether bulk loading of sstables is allowed.
```

smiklosovic: OK so I guess the decision to put this to 5.1 was done? It would be great to have it in 5.0 but I can live with this.

This can be done on commit. @Runtian no worries about that.

test/distributed/org/apache/cassandra/distributed/test/SSTableLoaderEncryptionOptionsTest.java
```diff
@@ -49,16 +49,16 @@
 
 public class SSTableLoaderEncryptionOptionsTest extends AbstractEncryptionOptionsImpl
 {
-    static Cluster CLUSTER;
-    static String NODES;
-    static int NATIVE_PORT;
-    static int STORAGE_PORT;
-    static int SSL_STORAGE_PORT;
+    static Cluster cluster;
```
smiklosovic: @Runtian I think you fixed the wrong test. `GuardrailBulkLoadEnabledTest` is untouched when it comes to these changes. Would you mind to keep this one as it was?

Runtian: oops, let me fix this.

PR #2652, opened 2023-08-30
CASSANDRA-18811 Set right client auth for creating SSL context in mTLS optional mode

jyothsnakonisa:


src/java/org/apache/cassandra/config/EncryptionOptions.java
```diff
@@ -67,6 +68,42 @@ public String description()
         }
     }
 
+    public enum ClientAuth
+    {
+        REQUIRED("true"),
+        NOT_REQUIRED("false"),
+        OPTIONAL("optional");
+        private final String value;
+        private static Map<String, ClientAuth> values = new HashMap<>();
```
frankgh: I would prefer a case insensitive treemap here, given that we are dealing with user input, so both "Required" and "REQUIRED" would be valid configuration options. Also, I think it is preferred to use final and I would uppercase this field to denote it is a static final field
```suggestion
        private static final Map<String, ClientAuth> VALUES = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```

src/java/org/apache/cassandra/config/EncryptionOptions.java
```diff
@@ -67,6 +68,42 @@ public String description()
...
+        {
+            this.value = value;
+        }
+
+        public static ClientAuth from(String value)
+        {
+            if(values.containsKey(value.toLowerCase()))
+            {
+                return values.get(value.toLowerCase());
+            };
```
frankgh: dangling semicolon
```suggestion
            }
```

src/java/org/apache/cassandra/config/EncryptionOptions.java
```diff
@@ -67,6 +68,42 @@ public String description()
...
+    public enum ClientAuth
+    {
+        REQUIRED("true"),
+        NOT_REQUIRED("false"),
+        OPTIONAL("optional");
+        private final String value;
+        private static Map<String, ClientAuth> values = new HashMap<>();
+        static
+        {
+            for(final ClientAuth clientAuth: ClientAuth.values())
```
frankgh: NIT: Usage of final is discouraged in C* code style
```suggestion
            for (ClientAuth clientAuth : ClientAuth.values())
```

jonmeredith (approved):

src/java/org/apache/cassandra/config/EncryptionOptions.java
```diff
@@ -67,6 +68,42 @@ public String description()
...
+        }
+
+        ClientAuth(String value)
+        {
+            this.value = value;
+        }
+
+        public static ClientAuth from(String value)
+        {
+            if(values.containsKey(value.toLowerCase()))
```
frankgh: NIT spacing
```suggestion
            if (values.containsKey(value.toLowerCase()))
```

src/java/org/apache/cassandra/auth/MutualTlsWithPasswordFallbackAuthenticator.java
```diff
@@ -50,4 +55,15 @@ public SaslNegotiator newSaslNegotiator(InetAddress clientAddress, Certificate[]
         }
         return mutualTlsAuthenticator.newSaslNegotiator(clientAddress, certificates);
     }
+
+    @Override
+    public void validateConfiguration() throws ConfigurationException
+    {
+        Config config = DatabaseDescriptor.getRawConfig();
+        if(config.client_encryption_options.getClientAuth() == EncryptionOptions.ClientAuth.NOT_REQUIRED)
```
frankgh: NIT:
```suggestion
        if (config.client_encryption_options.getClientAuth() == EncryptionOptions.ClientAuth.NOT_REQUIRED)
```

frankgh (commented):
Looks good overall, I left some minor comments, regarding styling and one about usability.

src/java/org/apache/cassandra/config/EncryptionOptions.java
```diff
@@ -67,6 +69,42 @@ public String description()
         }
     }
 
+    public enum ClientAuth
+    {
+        REQUIRED("true"),
+        NOT_REQUIRED("false"),
+        OPTIONAL("optional");
+        private final String value;
+        private static final Map<String, ClientAuth> VALUES = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
```
frankgh: giving this a second thought. It looks like the values are always lowercased, so I think we should go back to the original HashMap implementation

frankgh (approved):
+1 I left a comment about a recent change that I suggested, but I think it's not needed

src/java/org/apache/cassandra/config/EncryptionOptions.java
jyothsnakonisa: Reverted it back

pmcfadin:
Closing as [CASSANDRA-18811](https://issues.apache.org/jira/browse/CASSANDRA-18811) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #2671, opened 2023-09-05
CASSANDRA-18820 trunk Fix throwing exception from coordinator in dtests

jacek-lewandowski:
Unwrapping an exception has been modified so that the enclosing stacktrace is attached as suppressed; hence, a developer can access and investigate it.

### DIFF
--- a/src/java/org/apache/cassandra/utils/Throwables.java
+++ b/src/java/org/apache/cassandra/utils/Throwables.java
@@ -277,8 +277,11 @@ public final class Throwables
      * return its argument untouched.
      * <p>
      * We call a "wrapping" exception in the context of that method an exception whose only purpose is to wrap another
-     * exception, and currently this method recognize only 2 exception as "wrapping" ones: {@link ExecutionException}
-     * and {@link CompletionException}.
+     * exception, and currently this method recognize only 3 exception as "wrapping" ones: {@link ExecutionException},
+     * {@link CompletionException}, and {@link InvocationTargetException}.
+     * <p>
+     * Note that this method will also add the provided throwable as a suppressed the returned one, so that the original
+     * stacktrace will be available for debugging.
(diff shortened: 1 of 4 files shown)


src/java/org/apache/cassandra/utils/Throwables.java
```diff
@@ -288,12 +288,17 @@ public static Throwable unwrapped(Throwable t)
...
-        // I don't think it make sense for those 2 exception classes to ever be used with null causes, but no point
-        // in failing here if this happen. We still wrap the original exception if that happen so we get a sign
-        // that the assumption of this method is wrong.
-        return unwrapped == null
-               ? new RuntimeException("Got wrapping exception not wrapping anything", t)
-               : unwrapped;
+        if (unwrapped == null)
+            throw new RuntimeException("Got wrapping exception not wrapping anything", t);
+
+        if (!t.getStackTrace()[t.getStackTrace().length - 1].getClassName().equals(Thread.class.getName()))
```
dcapwell: I don't follow this logic, can you add some documentation for what this is trying to do?

dcapwell (approved):
+1.  I would love to see documentation about the "if the last stack element's class isn't Thread, then don't loose the stack trace", I don't fully follow it, but I am ok with it and the rest of the patch

src/java/org/apache/cassandra/utils/Throwables.java
jacek-lewandowski: Shame on me but I probably accidentally committed some junk. I removed that condition, sorry for the confusion.

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/881/workflows/90dd1988-5b80-459b-abf9-67838dc1abad

PR #2672, opened 2023-09-06
CASSANDRA-18821: Fix too early attempts to connect when auth is enabled

jacek-lewandowski:


JeremiahDJordan:
Should the auth creation wait time be lowered as well to improve test times?

bereng:
- Can't you create `waitForExistingRoles` in `CQLTester` i.e. and then it can be called from everywhere? That would remove the new single method class for jvm and it could be reused from junits as well iiuc
- You don't seem to have reduced the delay to 0 for unit tests?
- Missing CI and repeats

jacek-lewandowski:
Indeed I forgot about setting the delay to 0 for unit tests. Fixed now.

I'll not move that method outside of the distributed "module" because it applies only to the in-jvm distributed tests, where the delay is non-zero. I'd like to not modify the delay for the distributed tests as they usually start multiple nodes and we may fall into another category of problems.

Repeated tests on CircleCI makes zero sense to me in this case because:
1. the problem has not popped up on CircleCI, so why would it now?
2. I found the problem and was able to reproduce it reliably - then after applying a fix, the problem went away

bereng:
- Does `ServerTestUtils.java` have unsued imports now?
- The change in `CQLTester` impacts 90% of junits so a CI run is mandatory here imo.
- The repeated runs are not optional AFAIK so we'll need them. Do a 100 or 500 if you're confident enough...

test/unit/org/apache/cassandra/ServerTestUtils.java
```diff
@@ -25,6 +25,10 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.apache.cassandra.auth.CassandraRoleManager;
```
JeremiahDJordan: No changes needed to this file now?

test/distributed/org/apache/cassandra/distributed/util/Auth.java
```diff
@@ -0,0 +1,40 @@
...
+import org.apache.cassandra.auth.CassandraRoleManager;
+import org.apache.cassandra.distributed.api.IInvokableInstance;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.awaitility.Awaitility.await;
+
+public class Auth
+{
+
+    public static void waitForExistingRoles(IInvokableInstance instance)
```
JeremiahDJordan, line 30: No existing utility classes this can fit in?  That would be preferable to me, more chance of others finding and using it rather than making their own version of it, if it is in a common location.

bereng: `CQLTester` imo is a good candidate. My 2cts.

jacek-lewandowski: `CQLTester` is unrelated to in-jvm dtests. There is no utility class in distributed in-jvm tests. I can only move it either Cluster or Instance implementation

test/unit/org/apache/cassandra/ServerTestUtils.java
jacek-lewandowski: indeed, that would be caught by checkstyle anyway

jacek-lewandowski:
@bereng I'll obviously run the CI

jacek-lewandowski:
- [ ] trunk [j11](https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/868/workflows/1a65d5e7-c3ea-415f-bf6a-eb7048775d22)
- [x] 5.0 [j11](https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/870/workflows/b9a4d2b6-1d42-4f02-a071-73f463e0094f)
- [x] 4.1 [j8](https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/869/workflows/edae89f2-9374-46b3-9e12-c8412771e325)
- [x] 4.1 [j11](https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/869/workflows/aa17d4f6-772e-4d94-92c0-66354106f51c)

test/distributed/org/apache/cassandra/distributed/util/Auth.java
JeremiahDJordan: “waitForAuthInitialization” seems very reasonable to be on the cluster object actually?  Or possibly even have cluster start by default not return until that has happened if auth was enabled?

just trying to figure out how we avoid every test that gets added in the futures and uses auth from being flaky until a month later when someone remembers “oh yeah, you have auth you have to add this extra wait call”

jacek-lewandowski: moved to `Cluster`. I don't want to call it automatically because that would introduce some implicit semantic. I didn't really investigated what those tests are doing...

jacek-lewandowski:
Rerun dtest after moving the method https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/873/workflows/f5346b96-0650-4af6-8205-2139d37f784b

JeremiahDJordan (approved):

jacek-lewandowski:
@JeremiahDJordan @bereng It seems like I cannot move that method to `Cluster` as we get class loader problems.

test/distributed/org/apache/cassandra/distributed/util/Auth.java
bereng: `SSTableGeneratorTest` in the distributed package already uses `CQLTester` so it wouldn't be the first case...

bereng:
Ok the current PRs seem to all have the code moved into `Cluster`. You need to push new ones yes?

Just for reference the old CI runs you have previous to the move [here](https://github.com/apache/cassandra/pull/2672#issuecomment-1711554643) are fine. You're hitiing c16778 and c18360 on trunk

test/distributed/org/apache/cassandra/distributed/util/Auth.java
jacek-lewandowski: This is not a good example. That test is a fuzz test implemented with Harry and has nothing to do with in-jvm dtests as far as I can see (except being under the `distributed` root)

jacek-lewandowski:
Reverted to the `Auth` approach.

jacek-lewandowski:
@bereng can we merge that?

bereng:
Right I see you reverted all PRs. +1 though I still dislike the single utility class... :shrug:

jacek-lewandowski:
Me too @bereng, believe me.

Thanks for review!

PR #2675, opened 2023-09-07
CASSANDRA-18317 Properly synchronize CQLSSTableWriter#build on the Sc…

frankgh:
…hema.instance object

In this commit the `org.apache.cassandra.io.sstable.CQLSSTableWriter#build` method synchronizes on the `Schema.instance` object (instead of the `CQLSSTableWriter.class`) to prevent concurrent schema operations to fail when the offline tool also updates the schema.

For example, a table creation operation, which modifies the keyspaces tables metadata, might end up missing the update when a concurrent call to the `CQLSSTableWriter#build` method is accessing the singleton Schema instance.

In 5.0, the synchronization on Schema.instance is no longer needed. This is only an issue in the 4.0 branch

[CASSANDRA-18317](https://issues.apache.org/jira/browse/CASSANDRA-18317)


test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterConcurrencyTest.java
```diff
@@ -0,0 +1,132 @@
...
+                catch (Throwable throwable)
+                {
+                    logger.error("Error while processing element number {}", finalI, throwable);
+                    errorCount.incrementAndGet();
+                }
+            });
+        }
+
+        pool.shutdown();
+        assertThat(pool.awaitTermination(1, TimeUnit.MINUTES)).isTrue();
```
Maxwell-Guo: what about if ut just occurs a gc and cost more than one minute and  this just faild?

frankgh: a GC in the test? the test will fail. What is an acceptable value?

yifan-c: Timeout of closing the executor pool may not qualify as a failure condition of the test. How about logging a warning message and let it go?

frankgh: pushed suggestion to all branches

Maxwell-Guo: +1

Maxwell-Guo (approved):

frankgh:
Closed via https://github.com/apache/cassandra/commit/26c374da4f03e4a6b64e414805cd92f3eb0a36c6

PR #2680, opened 2023-09-12
CASSANDRA-18838: make stop-server shell out of the box

maoling:
- Making `stop-server` shell out of the box. The original shell needs users to uncomment the snippet to work for them, which's not user-friendly.
- The original shotgun approach is also not suitable for the scenario when we have cassandra sidecar process which also has the naming keyword: cassandra. By using the new added `./stop-server -e`, we can kill the `CassandraDaemon` exactly.
- Tested by `MacOS 12.1` and `Linux version 5.10.124`
- more details in the [CASSANDRA-18838](https://issues.apache.org/jira/browse/CASSANDRA-18838)

### DIFF
--- a/bin/stop-server
+++ b/bin/stop-server
@@ -1,3 +1,5 @@
+#!/bin/sh
+
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
 # distributed with this work for additional information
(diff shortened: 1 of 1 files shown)


bin/stop-server
```diff
@@ -14,18 +16,55 @@
...
+        ;;
+    -l)
+        # you can run something like this, but
+        # this is a shotgun approach and will kill other processes
+        # with cassandra in their name or arguments too:
+        user=$(whoami)
+        pgrep -u "$user" -f cassandra | xargs kill -9
+        ;;
+    -e)
+        pids=$(ps ax | grep -i 'org.apache.cassandra.service.CassandraDaemon' | grep java | grep -v grep | awk '{print $1}')
```
smiklosovic, line 50: @maoling I think you can use `pgrep` here too

PR #2682, opened 2023-09-13
CASSANDRA-18844: Fix Debian packaging forcing users to use OpenJDK

jfleming-ic:
Fixes https://issues.apache.org/jira/browse/CASSANDRA-18844

Instead of using OpenJDK as a dependency, use the virtual packages such as java11-runtime / java17-runtime

eg https://www.linuxtopia.org/online_books/linux_system_administration/debian_linux_guides/debian_linux_faq/ch-pkg_basics.en_007.html

Note for reviewer -  I've been able to test installation of a JDK 11 version of Cassandra but not JDK 17, I've checked the virtual package and it should work. Will try it later tonight my time and see if it works

```
root@2b3021cccbb0:/# apt-cache showpkg java17-runtime
Package: java17-runtime
Versions:

Reverse Depends:
Dependencies:
Provides:
Reverse Provides:
openjdk-17-jre 17.0.7+7-1~deb11u1 (= )
openjdk-17-jre 17.0.6+10-1~deb11u1 (= )

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,7 +3,7 @@ Section: misc
 Priority: extra
 Maintainer: Eric Evans <eevans@apache.org>
 Uploaders: Sylvain Lebresne <slebresne@apache.org>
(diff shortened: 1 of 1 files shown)


jfleming-ic:
Managed to test now in a Corretto 17 container as well, Cassandra installed and started ok

michaelsembwever (approved):

michaelsembwever (changes requested):
only 11 and 17 are permitted on 5.0

PR #2686, opened 2023-09-14
CASSANDRA-18841 - Fix InstanceClassloader leak

JeetKunDoug:
By running the `postStartup` logic in a `sync` block, we ensure that any ThreadLocal instances are created on the appropriate, instance-isolated thread. In this case, ThreadLocal objects were holding references to the InstanceClassLoader which created the ThreadLocal's value, thereby preventing the classloader from being collected.


jonmeredith (approved):

frankgh (approved):

JeetKunDoug:
Merged in 4.1/5.0-trunk - closing.

Thanks @jonmeredith !

PR #2689, opened 2023-09-15
CASSANDRA-18737: Do not create sstable files before registering in txn

jacek-lewandowski:


src/java/org/apache/cassandra/io/sstable/format/SSTableWriter.java
```diff
@@ -104,11 +107,21 @@ protected SSTableWriter(Builder<?, ?> builder, LifecycleNewTracker lifecycleNewT
         this.mmappedRegionsCache = builder.getMmappedRegionsCache();
         this.lifecycleNewTracker = lifecycleNewTracker;
 
-        Uninterruptibles.sleepUninterruptibly(1, TimeUnit.SECONDS);
+        List<Component> existingsComponents = components.stream().filter(c -> descriptor.fileFor(c).exists()).collect(Collectors.toList());
+        if (!existingsComponents.isEmpty())
```
blambov: We could use `anyMatch` instead of making a collection during the normal path (if necessary, gathering for the error message).

src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java
```diff
@@ -80,14 +80,23 @@ public class BigTableWriter extends SortedTableWriter<BigFormatPartitionWriter>
...
-                                     && lifecycleNewTracker instanceof ILifecycleTransaction
-                                     && !((ILifecycleTransaction) lifecycleNewTracker).isOffline();
+
+        try
+        {
+            this.rowIndexEntrySerializer = builder.getRowIndexEntrySerializer();
+            this.indexWriter = builder.getIndexWriter().get();
+
+            checkNotNull(builder.getRowIndexEntrySerializer());
+            checkNotNull(builder.getIndexWriter());
```
blambov: I don't think this is doing the intended check, `this.indexWriter` is the one that should not be null.

src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java
```diff
@@ -405,13 +414,13 @@ public MmappedRegionsCache getMmappedRegionsCache()
         }
 
         @Override
-        public SequentialWriter getDataWriter()
+        public LazyCloseableRef<SequentialWriter> getDataWriter()
```
blambov: I would switch the interface back to returning the objects, but create them here (if necessary, implemented as calling the reference's `get` method).

src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java
```diff
@@ -439,14 +448,14 @@ protected BigTableWriter buildInternal(LifecycleNewTracker lifecycleNewTracker,
...
                 rowIndexEntrySerializer = new RowIndexEntry.Serializer(descriptor.version, getSerializationHeader(), owner != null ? owner.getMetrics() : null);
-                dataWriter = DataComponent.buildWriter(descriptor,
-                                                       getTableMetadataRef().getLocal(),
-                                                       getIOOptions().writerOptions,
-                                                       getMetadataCollector(),
-                                                       lifecycleNewTracker.opType(),
-                                                       getIOOptions().flushCompression);
-                indexWriter = new IndexWriter(this);
-                partitionWriter = new BigFormatPartitionWriter(getSerializationHeader(), dataWriter, descriptor.version, rowIndexEntrySerializer.indexInfoSerializer());
+                dataWriter = LazyCloseableRef.create(() -> DataComponent.buildWriter(descriptor,
```
blambov: Could this logic be in the `getDataWriter` method? (Renamed to `openDataWriter` for clarity.)

src/java/org/apache/cassandra/utils/LazyCloseableRef.java
```diff
@@ -0,0 +1,98 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.utils;
+
+import java.util.function.Supplier;
+import javax.annotation.concurrent.NotThreadSafe;
+
+public abstract class LazyCloseableRef<T extends AutoCloseable> implements Closeable, Supplier<T>
```
blambov: Why do we need this to be closeable? The builders don't manage these resources.

I think changing the getters to `open...` as described above will add clarity and make it less likely for incorrect resource management to be introduced later.

src/java/org/apache/cassandra/io/sstable/format/big/BigTableWriter.java
jacek-lewandowski: right, my bad, moved checkNotNull but forgot to modify the arg

src/java/org/apache/cassandra/utils/LazyCloseableRef.java
jacek-lewandowski: thanks for the idea, will try and see how it looks

src/java/org/apache/cassandra/io/sstable/format/bti/BtiTableWriter.java
```diff
@@ -358,25 +315,52 @@ public MmappedRegionsCache getMmappedRegionsCache()
...
+        protected BtiFormatPartitionWriter openPartitionWriter()
         {
+            checkState(partitionWriter == null, "Partition writer has been already opened.");
+            partitionWriter = new BtiFormatPartitionWriter(getSerializationHeader(),
+                                                           getTableMetadataRef().getLocal().comparator,
+                                                           getDataWriter(),
+                                                           getIndexWriter().rowIndexWriter,
+                                                           descriptor.version);
+
             return ensuringInBuildInternalContext(partitionWriter);
```
blambov: This call is no longer necessary.

src/java/org/apache/cassandra/io/sstable/format/bti/BtiTableWriter.java
```diff
@@ -358,25 +315,52 @@ public MmappedRegionsCache getMmappedRegionsCache()
...
+
+        IndexWriter getIndexWriter()
         {
             return ensuringInBuildInternalContext(indexWriter);
         }
 
         private <T> T ensuringInBuildInternalContext(T value)
         {
-            Preconditions.checkState(value != null, "This getter can be used only during the lifetime of the sstable constructor. Do not use it directly.");
+            checkState(value != null, "This getter can be used only during the lifetime of the sstable constructor. Do not use it directly.");
```
blambov: It appears that we will most likely get this if `openPartitionWriter` is called before `openDataWriter` or `openIndexWriter`. Perhaps change the message to say it?

src/java/org/apache/cassandra/io/sstable/format/bti/BtiTableWriter.java
```diff
@@ -358,25 +315,52 @@ public MmappedRegionsCache getMmappedRegionsCache()
...
+        }
+
+        SequentialWriter getDataWriter()
         {
             return ensuringInBuildInternalContext(dataWriter);
         }
 
         @Override
-        public BtiFormatPartitionWriter getPartitionWriter()
+        protected BtiFormatPartitionWriter openPartitionWriter()
```
blambov: I see the problem now... it appears to me that it would be safer to pass index and data writers as arguments, but leave as is if you prefer.

src/java/org/apache/cassandra/io/sstable/format/bti/BtiTableWriter.java
```diff
@@ -358,25 +315,52 @@ public MmappedRegionsCache getMmappedRegionsCache()
...
         {
+            checkState(partitionWriter == null, "Partition writer has been already opened.");
+            partitionWriter = new BtiFormatPartitionWriter(getSerializationHeader(),
```
blambov: Note: Generally, if the only purpose of this field is to forbid second opening, I would replace it with a boolean, so that there is no chance that this reference causes the garbage collector to keep the object.

This should not be a real problem here as the builder reference is not something that is kept around.

blambov (approved):

src/java/org/apache/cassandra/io/sstable/format/bti/BtiTableWriter.java
jacek-lewandowski: no, not only that - it is also to close the resources if the constructor fails. And, also to access the already opened resource.

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/895/workflows/bb8f2560-4e21-442d-8073-cd0865e60e00

src/java/org/apache/cassandra/io/sstable/format/SSTableWriter.java
```diff
@@ -102,9 +108,38 @@ protected SSTableWriter(Builder<?, ?> builder, LifecycleNewTracker lifecycleNewT
         this.mmappedRegionsCache = builder.getMmappedRegionsCache();
         this.lifecycleNewTracker = lifecycleNewTracker;
 
+        // We need to ensure that no sstable components exist before the lifecycle transaction starts tracking it.
+        // Otherwise, it means that we either want to overwrite some existing sstable, which is not allowed, or some
+        // sstable files were created before the sstable is registered in the lifecycle transaction, which may lead
+        // to a race such that the sstable is listed as completed due to the lack of the transaction file before
+        // anything is actually written to it.
+        Supplier<List<Component>> existingComponents = Suppliers.memoize(() -> components.stream()
```
blambov: Nit:
I think I gave you the wrong impression what I was looking for here. We are normally working with assertions enabled, so optimizing for the other case is not something I consider important; I was wondering about the amount of work we do in the case where the assertion is checked but does not fire, and this change is making that case worse.

To me it is preferable to not materialize the resulting set, e.g. by doing:
```
        Set<Component> existingComponents = Sets.filter(components, c -> descriptor.fileFor(c).exists());

        assert existingComponents.isEmpty() : String.format("Cannot create a new SSTable in directory %s as a files for components %s already exist there", 
                                                            descriptor.directory, 
                                                            existingComponents);
```
where we create only the filter function and set wrapper at the expense of having to walk and filter the set twice when the assertion fires.

blambov (approved):
LGTM

src/java/org/apache/cassandra/io/sstable/format/SSTableWriter.java
jacek-lewandowski: I was under incorrect impression that we run without assertions enabled

src/java/org/apache/cassandra/io/sstable/format/SSTableWriter.java
```diff
@@ -460,26 +515,17 @@ private static Set<Component> indexComponents(Collection<Index.Group> indexGroup
...
-                if (observer != null)
-                {
-                    observer.begin();
-                    observers.add(observer);
-                }
-            }
+            if (this.indexGroups == null)
+                this.indexGroups = ImmutableList.copyOf(indexGroups);
+            else
+                this.indexGroups = ImmutableList.copyOf(Iterables.concat(this.indexGroups, indexGroups));
```
mike-tr-adamson: I'm not sure why we need to do this. I can't see any cases where `addSecondaryIndexGroups` is called twice on a builder. This could be potentially dangerous as we could end up calling the same index group twice (or more).

src/java/org/apache/cassandra/io/sstable/format/SSTableWriter.java
```diff
@@ -98,11 +107,63 @@ protected SSTableWriter(Builder<?, ?> builder, LifecycleNewTracker lifecycleNewT
         this.isTransient = builder.isTransientSSTable();
         this.metadataCollector = builder.getMetadataCollector();
         this.header = builder.getSerializationHeader();
-        this.observers = builder.getFlushObservers();
         this.mmappedRegionsCache = builder.getMmappedRegionsCache();
         this.lifecycleNewTracker = lifecycleNewTracker;
 
+        ArrayList<SSTableFlushObserver> observers = new ArrayList<>();
+        this.observers = Collections.unmodifiableList(observers);
```
mike-tr-adamson: nit: Couldn't these 2 lines move down to above the `try` below that populates the observers. I find reading the code that flicking up and down is painful.

src/java/org/apache/cassandra/io/sstable/format/SSTableWriter.java
```diff
@@ -98,11 +107,63 @@ protected SSTableWriter(Builder<?, ?> builder, LifecycleNewTracker lifecycleNewT
...
+
+        // We need to ensure that no sstable components exist before the lifecycle transaction starts tracking it.
+        // Otherwise, it means that we either want to overwrite some existing sstable, which is not allowed, or some
+        // sstable files were created before the sstable is registered in the lifecycle transaction, which may lead
+        // to a race such that the sstable is listed as completed due to the lack of the transaction file before
+        // anything is actually written to it.
+        Set<Component> existingComponents = Sets.filter(components, c -> descriptor.fileFor(c).exists());
+
+        assert existingComponents.isEmpty() : String.format("Cannot create a new SSTable in directory %s as a files for components %s already exist there",
```
mike-tr-adamson: nit: `Cannot create a new SSTable in directory %s as component files %s already exist there` would read better.

mike-tr-adamson (approved):

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/911/workflows/7a876be0-ab14-4800-b19c-3506e831417f

PR #2693, opened 2023-09-15
Internode legacy SSL storage port certificate is not hot reloaded on …

jonmeredith:
See CASSANDRA-18681

### DIFF
--- a/src/java/org/apache/cassandra/net/MessagingServiceMBeanImpl.java
+++ b/src/java/org/apache/cassandra/net/MessagingServiceMBeanImpl.java
@@ -24,8 +24,6 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
-import org.apache.cassandra.config.DatabaseDescriptor;
-import org.apache.cassandra.config.EncryptionOptions;
 import org.apache.cassandra.locator.InetAddressAndPort;
 import org.apache.cassandra.metrics.InternodeOutboundMetrics;
 import org.apache.cassandra.metrics.MessagingMetrics;
@@ -278,10 +276,7 @@ public class MessagingServiceMBeanImpl implements MessagingServiceMBean
     @Override
     public void reloadSslCertificates() throws IOException
     {
-        final EncryptionOptions.ServerEncryptionOptions serverOpts = DatabaseDescriptor.getInternodeMessagingEncyptionOptions();
-        final EncryptionOptions clientOpts = DatabaseDescriptor.getNativeProtocolEncryptionOptions();
-        SSLFactory.validateSslCerts(serverOpts, clientOpts);
-        SSLFactory.checkCertFilesForHotReloading(serverOpts, clientOpts);
+        SSLFactory.forceCheckCertFiles();
     }
 
     @Override
--- a/src/java/org/apache/cassandra/net/MessagingServiceMBeanImpl.java
+++ b/src/java/org/apache/cassandra/net/MessagingServiceMBeanImpl.java
@@ -274,7 +274,7 @@ public class MessagingServiceMBeanImpl implements MessagingServiceMBean
     }
 
     @Override
-    public void reloadSslCertificates() throws IOException
+    public void reloadSslCertificates()
     {
         SSLFactory.forceCheckCertFiles();
     }
--- a/src/java/org/apache/cassandra/net/MessagingServiceMBeanImpl.java
+++ b/src/java/org/apache/cassandra/net/MessagingServiceMBeanImpl.java
@@ -17,7 +17,6 @@
  */
 package org.apache.cassandra.net;
 
-import java.io.IOException;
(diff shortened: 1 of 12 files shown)


src/java/org/apache/cassandra/transport/PipelineConfigurator.java
```diff
@@ -198,7 +201,9 @@ protected void decode(ChannelHandlerContext channelHandlerContext, ByteBuf byteB
                 return channel -> {
                     SslContext sslContext = SSLFactory.getOrCreateSslContext(encryptionOptions,
                                                                              encryptionOptions.require_client_auth,
-                                                                             ISslContextFactory.SocketType.SERVER);
+                                                                             ISslContextFactory.SocketType.SERVER,
+                                                                             SSL_FACTORY_CONTEXT_DESCRIPTION);
+                    InetSocketAddress peer = encryptionOptions.require_endpoint_verification ? (InetSocketAddress) channel.remoteAddress() : null;
```
frankgh: is this needed for something?

src/java/org/apache/cassandra/security/SSLFactory.java
```diff
@@ -175,42 +177,52 @@ static SslContext createNettySslContext(EncryptionOptions options, boolean verif
...
-        {
-            if (options.sslContextFactoryInstance.shouldReload())
+        List<CacheKey> keys = new ArrayList<>(Collections.list(cachedSslContexts.keys()));
+        Set<EncryptionOptions> checked = new HashSet<>();
+
+        keys.forEach(key -> {
+            final EncryptionOptions opts = key.encryptionOptions;
+            if (checked.contains(opts)) // only check each encryption option once as shared for different types
+                return;
+            checked.add(key.encryptionOptions);
```
frankgh: we can simplify this a bit:
```suggestion
            if (!checked.add(key.encryptionOptions)) // only check each encryption option once as shared for different types
                return;
```

src/java/org/apache/cassandra/security/SSLFactory.java
```diff
@@ -175,42 +177,52 @@ static SslContext createNettySslContext(EncryptionOptions options, boolean verif
...
             {
-                logger.info("SSL certificates have been updated for {}. Resetting the ssl contexts for new " +
-                            "connections.", options.getClass().getName());
-                validateSslContext(contextDescription, options, verifyPeerCertificate, false);
-                clearSslContextCache(options);
+                try
+                {
+                    validateSslContext(key.contextDescription, opts,
+                                       opts instanceof EncryptionOptions.ServerEncryptionOptions ? true
+                                                                                                 : opts.require_client_auth, false);
```
frankgh: This should simplify a bit:
```suggestion
                                       opts instanceof EncryptionOptions.ServerEncryptionOptions || opts.require_client_auth, false);
```

src/java/org/apache/cassandra/net/MessagingServiceMBeanImpl.java
```diff
@@ -278,10 +276,7 @@ public boolean isBackPressureEnabled()
     @Override
     public void reloadSslCertificates() throws IOException
```
frankgh: `IOException` is no longer thrown, so it can be removed.

test/unit/org/apache/cassandra/security/SSLFactoryTest.java
```diff
@@ -296,9 +311,38 @@ public void testCacheKeyInequalityForCustomSslContextFactory() {
...
         .withProtocol("TLSv1.1");
 
-        SSLFactory.CacheKey cacheKey2 = new SSLFactory.CacheKey(encryptionOptions2, ISslContextFactory.SocketType.SERVER
+        SSLFactory.CacheKey cacheKey2 = new SSLFactory.CacheKey(encryptionOptions2, ISslContextFactory.SocketType.SERVER, "test"
         );
 
         Assert.assertNotEquals(cacheKey1, cacheKey2);
     }
+
+    public static class TestFileBasedSSLContextFactory extends FileBasedSslContextFactory {
```
frankgh: unused?

src/java/org/apache/cassandra/security/SSLFactory.java
jonmeredith: agreed, better

src/java/org/apache/cassandra/transport/PipelineConfigurator.java
jonmeredith: no, some left over debugging, nice catch.

test/unit/org/apache/cassandra/security/SSLFactoryTest.java
jonmeredith: accidentally included in back port, used by other tests in 5.0/trunk

frankgh (approved):
I prefer this approach over the previous iteration where we had the custom equality check in the `clearSslContextCache` method. +1

PR #2708, opened 2023-09-20
CASSANDRA-18871: JMH improvements - faster build and async profiler

jacek-lewandowski:


test/microbench/org/apache/cassandra/test/microbench/instance/ReadTestSmallPartitionsBench.java
```diff
@@ -39,7 +39,7 @@
 @Fork(value = 1)
 @Threads(1)
 @State(Scope.Benchmark)
-public class ReadTestSmallPartitions extends ReadTest
+public class ReadTestSmallPartitionsBench extends ReadTest
```
blambov: Nit: I would get rid of `Test` here and in the next file: `ReadSmallPartitionsBench`, `ReadWidePartitionsBench`. Consider also renaming the base class from `ReadTest` to `ReadBench`.

blambov (approved):

.build/build-bench.xml
```diff
@@ -0,0 +1,137 @@
...
+                <arg value="-rff"/>
+                <arg value="${build.test.dir}/jmh-result.json"/>
+                <arg value="-v"/>
+                <arg value="EXTRA"/>
+
+                <arg line="${jmh.args}"/>
+
+                <extra-args/>
+
+                <!-- Broken: ZeroCopyStreamingBench,MutationBench,FastThreadLocalBench  (FIXME) -->
```
Mmuzaf: Do we need to keep this comment? It doesn't shed any light on what exactly is broken. We can also create and issue and leave the issue number here.

Mmuzaf (approved):
@jacek-lewandowski I have tested the latest changes on my Mac M1, and everything now works fine now. I can see the profiler results here: `build/test/profiler`.

I have left a minor comment, I think the `TODO` comments should be added with a reference issue number.

PR #2713, opened 2023-09-21
CASSANDRA-18747-4.1: Fix race condition in Schema

jacek-lewandowski:


jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/909/workflows/8c169ac2-a284-426b-9a64-510e966e63d5

src/java/org/apache/cassandra/cql3/statements/DescribeStatement.java
```diff
@@ -132,8 +132,13 @@ public final ResultMessage execute(QueryState state, QueryOptions options, long
     @Override
     public ResultMessage executeLocally(QueryState state, QueryOptions options)
     {
-        Keyspaces keyspaces = Schema.instance.distributedAndLocalKeyspaces();
-        UUID schemaVersion = Schema.instance.getVersion();
+        UUID schemaVersion;
+        Keyspaces keyspaces;
+        do
+        {
+            schemaVersion = Schema.instance.getVersion();
```
bereng: So the error was coming from sthg running in between these 2 lines of code getting the schema versions not match?

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -77,8 +77,19 @@ public class Schema implements SchemaProvider
...
     public static final Schema instance = new Schema();
 
-    private volatile Keyspaces distributedKeyspaces = Keyspaces.none();
-    private volatile Keyspaces distributedAndLocalKeyspaces;
+    private static class Holder
+    {
+        private final Keyspaces distributedKeyspaces;
+        private final Keyspaces distributedAndLocalKeyspaces;
+
+        private Holder(Keyspaces distributedKeyspaces, Keyspaces distributedAndLocalKeyspaces)
```
bereng: This is just to force both to be updated at the same time and avoid them getting out of sync?

bereng:
Ar you planning on adding a test for the specific scenario? Also you can run the pre-commit CI job rather than the separate one which is more cumbersome and some jobs got missed.

src/java/org/apache/cassandra/cql3/statements/DescribeStatement.java
jacek-lewandowski: The error was not coming from this place - this is a bug I discovered while inspecting the code. Although the update of version and list of keyspaces is synchronized, they are not updated atomically. In particular, while the list of keyspaces is updated incrementally with each added/modified/dropped keyspace, the version is updated at the of the entire schema transformation. It is because we do not know the intermediate schema versions, and also for the performance reasons.

Therefore, this simple loop ensures that the schema did not change while we requested list of keyspaces. In other words, schema version and the keyspaces match.

src/java/org/apache/cassandra/schema/Schema.java
jacek-lewandowski: yes, instead of making getter synchronized as it would be slower and lead to deadlocks.

jacek-lewandowski:
I don't think I can add a test for this scenario due to changes I've made. To prove that, I needed to add a sleep between updating one and the other field. With those changes such a test makes no sense.

bereng:
^Couldn't you inject a sleep + getVersionUnsafe() + getVersion() to test the former is out of sync and the latter is safe to use?

jacek-lewandowski:
So you mean that piece of code in `DescriptionStatement` right? yeah, I can try

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/939/workflows/d6ef66f0-18b6-420d-a264-6cc6932ac0ae

bereng:
It seems like `MigrationCoordinatorTest` failed on the repeats. I checked JIRA and Butler and they both seem clean. So it's worth investigating imo

jacek-lewandowski:
Yeah, I know, I'll run repeated test for 4.1 today

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/941/workflows/46fc6cb7-135e-4862-b9d3-6996c0993de8

bereng:
^Ok so it failed on 4.1 as well. A ticket for that one needs to be created and I'll get back to this review shortly

test/unit/org/apache/cassandra/cql3/statements/DescribeStatementTest.java
```diff
@@ -377,6 +393,52 @@ public void testDescribe() throws Throwable
...
+            {
+                return executeNetWithPaging(ProtocolVersion.CURRENT, "DESCRIBE KEYSPACES", 1);
+            }
+            catch (Throwable e)
+            {
+                throw new RuntimeException(e);
+            }
+        });
+        // wait for the describe statement to be in progress
+        Awaitility.await().atMost(Duration.ofSeconds(5)).until(() -> describeStatementLatch.count() == 1);
```
bereng: Schema cahnge operations can take >60s in jenkins. We have 120s in dtests i.e. and I am raising this to 120s for jvm dtests. We should raise it here as well to prevent flakies.

bereng:
+1 pending increasing timeouts

test/unit/org/apache/cassandra/cql3/statements/DescribeStatementTest.java
jacek-lewandowski: Though, this is not a dtest and there is no networking. Repeated run with this test already completed without problems. If you insist, I can change that, however, I cannot justify such a long timeout given what it waits for.

bereng: _jvm_ dtest, not py dtests. And repeated runs worked on circle, not jenkins. I know it sounds crazy but on jenkins schema changes can take >60s easily and sbdy will have to read logs/debug if we don't go with more realistic, though crazy, timeouts :shrug:

jacek-lewandowski: this is cqltester based unit test....

bereng: Yes I know.

jacek-lewandowski: changed to 20 seconds, it does not make sense to increase it more for those tests. Also renamed "blablabla" rule 🤦🏻

bereng: It does indeed make a lot of sense to raise it. The pain of having to continuously debug flakies, read the logs, chase the code, etc. which in the end happen to be just timeouts it's not negligible and very demotivating imo.

20s sounds like a magic number. You mentioned `test.driver.read_timeout_ms` is at 24s. I wonder if we can access that programmatically?

jacek-lewandowski:
is it ok now?

bereng:
+1 now we need the rest of branches and CI :-)

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/955/workflows/dee3d689-6795-4089-bb78-539b837bdb1f

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/955/workflows/dee3d689-6795-4089-bb78-539b837bdb1f

https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/961/workflows/d8e74ab8-f3d8-404f-9ca7-df4a0ccbde29

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/961/workflows/ccb113fc-7a0e-4fa6-bdb9-4c284bd7e97a

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -496,27 +509,23 @@ public Optional<Function> findFunction(FunctionName name, List<AbstractType<?>>
...
-    public boolean isSameVersion(UUID schemaVersion)
+    public synchronized UUID getVersion()
     {
-        return schemaVersion != null && schemaVersion.equals(version);
-    }
-
-    /**
-     * Checks whether the current schema is empty.
-     */
-    public boolean isEmpty()
```
JeremiahDJordan: Why remove the helper function?

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -496,27 +509,23 @@ public Optional<Function> findFunction(FunctionName name, List<AbstractType<?>>
...
     {
         return version;
     }
 
     /**
-     * Checks whether the given schema version is the same as the current local schema.
+     * Returns the curren schema version. If the schema is in the middle of the update, the method will wait until
+     * the update is completed and return the new version.
      */
-    public boolean isSameVersion(UUID schemaVersion)
```
JeremiahDJordan: Why remove the helper function?

JeremiahDJordan (approved):
Seems reasonable to me. Race conditions are hard.
Only comment is I wouldn’t remove the helper functions if possible.

src/java/org/apache/cassandra/schema/Schema.java
jacek-lewandowski: Keeping them would require safe and unsafe versions of them, which I presumed to be clumsy; I don't have a strong opinion on that though

src/java/org/apache/cassandra/schema/Schema.java
ekaterinadimitrova2: I wouldn't remove public function, especially on a patch release.

src/java/org/apache/cassandra/schema/Schema.java
ekaterinadimitrova2: I do not recommend removing public methods on patch release

test/unit/org/apache/cassandra/cql3/statements/DescribeStatementTest.java
```diff
@@ -377,6 +393,54 @@ public void testDescribe() throws Throwable
...
+
+        // at this point, the describe statement got the schema version and is about to get the keyspaces
+        // we will now create a keyspace, release the lock, and let the describe statement continue
+        String ks = createKeyspace("CREATE KEYSPACE %s WITH REPLICATION = {'class' : 'SimpleStrategy', 'replication_factor' : 1};");
+        describeStatementLatch.decrement();
+
+        // the paginated describe statement should complete successfully and should include the new keyspace even though
+        // the keyspace was created after the describe statement started; this proves that when the describe statement
+        // finds itself in a situation where the schema being changed, it will wait for completion before actually
+        // starting execution; otherwise we would get an exception
```
ekaterinadimitrova2: Isn't this a change of behavior for Describe in a patch version?
While it seems more correct, it is an improvement. Also, I suspect this was the same even before 4.1, no? Was this checked? Also, there is still synchronization cost in place, unfortunately, added through the synchronized getVersion.

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -496,25 +509,43 @@ public Optional<Function> findFunction(FunctionName name, List<AbstractType<?>>
...
      */
-    public UUID getVersion()
+    public synchronized UUID getVersion()
     {
         return version;
     }
 
     /**
      * Checks whether the given schema version is the same as the current local schema.
+     *
```
ekaterinadimitrova2:
```suggestion
     * <p>
```

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -496,25 +509,43 @@ public Optional<Function> findFunction(FunctionName name, List<AbstractType<?>>
...
+     * version is unknown for the partially applied changes and is updated after the entire schema change is applied.
+     * @see #getVersion()
+     */
+    public UUID getVersionUnsafe()
+    {
+        return version;
+    }
+
+    /**
+     * Returns the curren schema version. If the schema is in the middle of the update, the method will wait until
```
ekaterinadimitrova2:
```suggestion
     * Returns the current schema version. If the schema is in the middle of the update, the method will wait until
```

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -496,25 +509,43 @@ public Optional<Function> findFunction(FunctionName name, List<AbstractType<?>>
...
-    public boolean isSameVersion(UUID schemaVersion)
+    @Deprecated(since = "4.1", forRemoval = true)
+    public synchronized boolean isSameVersion(UUID schemaVersion)
     {
         return schemaVersion != null && schemaVersion.equals(version);
     }
 
     /**
      * Checks whether the current schema is empty.
+     *
```
ekaterinadimitrova2:
```suggestion
     * <p>
```

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -496,25 +509,43 @@ public Optional<Function> findFunction(FunctionName name, List<AbstractType<?>>
...
     }
 
     /**
      * Checks whether the given schema version is the same as the current local schema.
+     *
+     * @deprecated please use {@link #getVersion()} or {@link #getVersionUnsafe()} depending on the use case
      */
-    public boolean isSameVersion(UUID schemaVersion)
+    @Deprecated(since = "4.1", forRemoval = true)
+    public synchronized boolean isSameVersion(UUID schemaVersion)
```
ekaterinadimitrova2: Technically you replace it with the `equals` method being chained to the appropriate getVersion method.

ekaterinadimitrova2 (commented):
Left some nit picks.

Looks correct but I have one concern, while DescribeStatement will be improved and this could be considered a bug, we are changing behavior in a patch version and incurring synchronization cost. No? Also, wasn't this working that way even before 4.1? 

An explicit test for the new holder is missing, but we will see the Jenkins failures disappearing after commit. I believe that is all we need in this case. 
CI LGTM
My comments on the PR to be considered for all branches (except the 4.1 explicit one around change of behavior)

src/java/org/apache/cassandra/schema/Schema.java
jacek-lewandowski: yes, would you like me to add that to the comment? I believe the API should be sufficient and minimal

test/unit/org/apache/cassandra/cql3/statements/DescribeStatementTest.java
jacek-lewandowski: It is a bug if we can return schema and schema version where version does not correspond to the schema, which will result in paging failure; You are right that the bug exists in 4.0 as well; 3.11 and prior do not have server-side `DescribeStatement`. I'll create a separate ticket to fix that starting from 4.0 (https://issues.apache.org/jira/browse/CASSANDRA-18921)

src/java/org/apache/cassandra/schema/Schema.java
ekaterinadimitrova2: `would you like me to add that to the comment? `
I will leave the final decision to you,  just mentioned that it is a bit more than that :-)

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -246,9 +253,10 @@ private Keyspace maybeRemoveKeyspaceInstance(String keyspaceName, Consumer<Keysp
         }
     }
 
+    @Deprecated(since = "4.1", forRemoval = true)
```
blerer: I think that you can simply remove it.

src/java/org/apache/cassandra/db/Keyspace.java
```diff
@@ -770,12 +770,12 @@ public static Stream<Keyspace> allExisting()
 
     public static Iterable<Keyspace> nonSystem()
```
blerer: This one seems to be use only in Keyspace.drain where we should use `nonLocalStrategy()` so we should be able to remove it.

src/java/org/apache/cassandra/metrics/TableMetrics.java
```diff
@@ -280,7 +280,7 @@ private static Pair<Long, Long> totalNonSystemTablesSize(Predicate<SSTableReader
     {
         long total = 0;
         long filtered = 0;
-        for (String keyspace : Schema.instance.getNonSystemKeyspaces().names())
+        for (String keyspace : Schema.instance.distributedKeyspaces().names())
```
blerer, line 283: It seems that we are lying for TableMetrics too :-(

jacek-lewandowski: can we address that in 5.0 and change metrics a bit?

blerer: It was just a statement. We can let that for later.

blerer (approved):

ekaterinadimitrova2 (approved):

jacek-lewandowski:
j8 - https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/977/workflows/31134b2b-11d9-4992-b2e8-04ea4ed9ef4c 🆗 
j11 - https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/977/workflows/48fd2a77-cff7-4bd4-ae7a-a4dd9f824149 🆗

PR #2723, opened 2023-09-22
CASSANDRA-18876 5.0: Vector type shouldn't support empty value

adelapena:


src/java/org/apache/cassandra/db/marshal/VectorType.java
```diff
@@ -174,8 +176,6 @@ public <V> V decomposeAsFloat(ValueAccessor<V> accessor, float[] value)
     {
         if (!(elementType instanceof FloatType))
             throw new IllegalStateException("Attempted to read as float, but element type is " + elementType.asCQL3Type());
-        if (value == null)
```
dcapwell, line 177: if this happens we NPE, so should we throw an error with a useful msg if we don't want to return `null`?

This change is also not consistent with `fromComparableBytes`

src/java/org/apache/cassandra/db/marshal/VectorType.java
```diff
@@ -629,8 +621,6 @@ public <V> List<T> deserialize(V input, ValueAccessor<V> accessor)
         @Override
         public <V> void validate(V input, ValueAccessor<V> accessor) throws MarshalException
         {
-            if (accessor.isEmpty(input))
```
dcapwell, line 632: don't we need to reject this case?

src/java/org/apache/cassandra/db/marshal/VectorType.java
```diff
@@ -597,8 +591,6 @@ public <V> V serializeRaw(List<V> value, ValueAccessor<V> accessor)
         @Override
         public ByteBuffer serialize(List<T> value)
         {
-            if (value == null)
```
dcapwell, line 600: don't we need to reject this case?

src/java/org/apache/cassandra/db/marshal/VectorType.java
```diff
@@ -491,8 +492,6 @@ public <V> List<T> deserialize(V input, ValueAccessor<V> accessor)
         @Override
         public <V> void validate(V input, ValueAccessor<V> accessor) throws MarshalException
         {
-            if (accessor.isEmpty(input))
```
dcapwell, line 494: don't we need to reject this case?

src/java/org/apache/cassandra/db/marshal/VectorType.java
```diff
@@ -457,8 +460,6 @@ public <V> V serializeRaw(List<V> value, ValueAccessor<V> accessor)
         @Override
         public ByteBuffer serialize(List<T> value)
         {
-            if (value == null)
```
dcapwell, line 460: don't we need to reject this case?

src/java/org/apache/cassandra/db/marshal/VectorType.java
```diff
@@ -442,8 +447,6 @@ public <V> List<V> split(V buffer, ValueAccessor<V> accessor)
         @Override
         public <V> V serializeRaw(List<V> value, ValueAccessor<V> accessor)
         {
-            if (value == null)
```
dcapwell, line 445: don't we need to reject this case?

src/java/org/apache/cassandra/db/marshal/VectorType.java
```diff
@@ -583,8 +579,6 @@ public <V> List<V> split(V buffer, ValueAccessor<V> accessor)
...
```
dcapwell, line 586: don't we need to reject this case?

dcapwell (approved):
Overall LGTM, few non-blocking questions about `VectorType` change

Maxwell-Guo (approved):

adelapena:
Thanks for the review. I have added an explicit exception about the empty/null value in those places where we used to manage the empty values. The error message is simply `Invalid empty vector value`, I don't know if there is a better idea for it.

dcapwell (approved):

PR #2746, opened 2023-09-27
CASSANDRA-18876 trunk: Vector type shouldn't support empty value

adelapena:


src/java/org/apache/cassandra/db/marshal/VectorType.java
```diff
@@ -215,7 +217,8 @@ public <V> ByteSource asComparableBytes(ValueAccessor<V> accessor, V value, Byte
     public <V> V fromComparableBytes(ValueAccessor<V> accessor, ByteSource.Peekable comparableBytes, ByteComparable.Version version)
     {
         if (comparableBytes == null)
-            return accessor.empty();
+            rejectEmptyValue();
```
Maxwell-Guo: why do not throw  a MarshalException directly instead of execute a function which  just only throw an exception ?

Beside If the purpose is to use common code , I think we can use some code like this [throw ire](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java#L307) 
which can express the exception throwing and rejection behaviors more intuitively. wdyt ?

adelapena: The purpose here is to make sure that all similar null/empty errors throw the same exception and message. Throwing inside the utility method or returning the exception to be immediately thrown seems down to personal preferences.

Maxwell-Guo: ok, that is all right.~~~~ But I still have to explain the reason why I ask this question , because I have checked the cassandra's code , but don't  find  any function which only throw exception and the function name(rejectEmptyValue)  are also a bit less straightforward in explaining behavior to me .

Maxwell-Guo: Besides, what about rename the funtion to something like rejectNullOrEmptyValue ? as you have said "The purpose here is to make sure that all similar null/empty errors throw the same exception and message."

adelapena: Yes, why not. I have renamed it to `rejectNullOrEmptyValue`.

PR #2747, opened 2023-09-27
CASSANDRA-18889 5.0: Fix vector type to support DDM's mask_default function

adelapena:


src/java/org/apache/cassandra/db/marshal/VectorType.java
```diff
@@ -359,6 +365,13 @@ private <V> void checkConsumedFully(V buffer, ValueAccessor<V> accessor, int off
             throw new MarshalException("Unexpected " + remaining + " extraneous bytes after " + asCQL3Type() + " value");
     }
 
+    @Override
+    public ByteBuffer getMaskedValue()
+    {
+        List<ByteBuffer> values = Collections.nCopies(dimension, elementType.getMaskedValue());
```
Maxwell-Guo, line 379: Do we need to add a doc to tell the users how vectortype's mask_default was implemented ? let users know whether the usage expectations are met.

test/unit/org/apache/cassandra/cql3/functions/masking/ColumnMaskQueryWithDefaultTest.java
```diff
@@ -39,7 +42,11 @@ public static Collection<Object[]> options()
             options.add(new Object[]{ order, "DEFAULT", "text", "abc", "****" });
             options.add(new Object[]{ order, "DEFAULT", "int", 123, 0 });
             options.add(new Object[]{ order, "mask_default()", "text", "abc", "****" });
-            options.add(new Object[]{ order, "mask_default()", "int", 123, 0, });
+            options.add(new Object[]{ order, "mask_default()", "int", 123, 0 });
+            // TODO: the driver version that we use doesn't support vectors, so we have to use raw values by now
+            options.add(new Object[]{ order, "mask_default()", "vector<int, 2>",
```
Maxwell-Guo, line 47: Is it necessary to add any other element type's case  like float ?

test/unit/org/apache/cassandra/cql3/functions/masking/ColumnMaskQueryWithReplaceTest.java
```diff
@@ -43,6 +46,10 @@ public static Collection<Object[]> options()
             options.add(new Object[]{ order, "mask_replace(0)", "int", 123, 0 });
             options.add(new Object[]{ order, "mask_replace(0)", "bigint", 123L, 0L });
             options.add(new Object[]{ order, "mask_replace(0)", "varint", BigInteger.valueOf(123), BigInteger.ZERO });
+            // TODO: the driver version that we use doesn't support vectors, so we have to use raw values by now
+            options.add(new Object[]{ order, "mask_replace([0, 0])", "vector<int, 2>",
```
Maxwell-Guo, line 50: same with the upper case

src/java/org/apache/cassandra/db/marshal/VectorType.java
adelapena: I have added some doc lines and examples about how default masking values are generated.

test/unit/org/apache/cassandra/cql3/functions/masking/ColumnMaskQueryWithDefaultTest.java
adelapena: These tests focus on how different queries interact with masking rather than on what each particular function and data type combination does. It's useful to test different functions just in case, but we cannot add too many combinations here because each function/type variation multiplies the number of parameterized cases, making the test slower. That's why we only test with a single function and a few data types. At the moment the test takes 2-3 minutes to run, and adding more cases might make it timeout on CI.

Instead, I have added a specific randomized/fuzz test on `AbstractTypeTest` that would cover many data type combinations.

test/unit/org/apache/cassandra/cql3/functions/masking/ColumnMaskQueryWithReplaceTest.java
adelapena: Same as above.

doc/modules/cassandra/examples/CQL/ddm_create_table_with_udf.cql
```diff
@@ -2,7 +2,7 @@ CREATE FUNCTION redact(input text)
    CALLED ON NULL INPUT
    RETURNS text
    LANGUAGE java
-   AS 'return "redacted";
+   AS 'return "redacted";';
```
adelapena, line 5: This has been kindly found by @sarma1807 . I'm piggy backing this little fix here since it probably isn't worth a full ticket.

sarma1807: Thank you very much for the fix @adelapena

src/java/org/apache/cassandra/cql3/CQL3Type.java
```diff
@@ -566,6 +567,21 @@ public String toCQLLiteral(ByteBuffer buffer)
             return sb.toString();
         }
 
+        @Override
```
bereng, line 570: These should have been added when Vector was added correct?

adelapena: Yes, they should. Without proper hashcode/equals we can end up seeing messages such as `Masking function mask_default() return type is vector<int, 2>. This is different to the type of the masked column r2 of type vector<int, 2>.`

bereng:
+1

pmcfadin:
Closing as [CASSANDRA-18889](https://issues.apache.org/jira/browse/CASSANDRA-18889) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #2766, opened 2023-10-03
CASSANDRA-17092 CEP-15 Accord Beta

maedhroz:


accord_demo.txt
```diff
@@ -0,0 +1,19 @@
+
+ccm create accord-cql-poc -n 3
+ccm start
+
+bin/cqlsh -e "create keyspace ks with replication={'class':'SimpleStrategy', 'replication_factor':3};"
```
Maxwell-Guo, line 5: what about rename this to general_purpose_transactions_demo.txt ?

conf/cassandra.yaml
```diff
@@ -1818,6 +1818,9 @@ drop_compact_storage_enabled: false
 # Whether or not USE <keyspace> is allowed. This is enabled by default to avoid failure on upgrade.
 #use_statements_enabled: true
 
+# Enables the execution of Accord (multi-key) transactions on this node.
```
Maxwell-Guo: what about change to "Enables the execution of General Purpose Transactions (multi-partition) transactions on this node." ,Together with the "accord_transactions_enabled" configuration .

maedhroz:
Last known CircleCI run: https://app.circleci.com/pipelines/github/maedhroz/cassandra/774/workflows/ca42a1cd-8d3b-4309-b25a-997648a28730

accord_demo.txt
maedhroz: It's very likely that, before `cep-15-accord` merges, we'll remove this document entirely and replace it w/ something like the Markdown readme we have for SAI. In particular, the unsafe epoch creation is only a crutch while TCM/CEP-21 integration is still pending.

conf/cassandra.yaml
maedhroz: We can certainly revisit this before `cep-15-accord` merges.

PR #2769, opened 2023-10-04
Correctly remove Index.Group from IndexRegistry

mike-tr-adamson:
The Index.Group was being left in the list indexGroups in the SecondaryIndexManager because the incorrect key was being used to remove it from the map

patch by Mike Adamson; reviewed by <reviewer> for CASSANDRA-18905

```

```

### DIFF
--- a/src/java/org/apache/cassandra/index/Index.java
+++ b/src/java/org/apache/cassandra/index/Index.java
@@ -21,7 +21,11 @@
 package org.apache.cassandra.index;
 
 import java.io.UncheckedIOException;
-import java.util.*;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.function.Function;
 import java.util.function.Predicate;
@@ -52,8 +56,8 @@ import org.apache.cassandra.index.transactions.IndexTransaction;
 import org.apache.cassandra.io.sstable.Component;
 import org.apache.cassandra.io.sstable.Descriptor;
 import org.apache.cassandra.io.sstable.ReducingKeyIterator;
-import org.apache.cassandra.io.sstable.SSTableFlushObserver;
 import org.apache.cassandra.io.sstable.SSTable;
+import org.apache.cassandra.io.sstable.SSTableFlushObserver;
 import org.apache.cassandra.io.sstable.format.SSTableReader;
 import org.apache.cassandra.schema.ColumnMetadata;
 import org.apache.cassandra.schema.IndexMetadata;
@@ -730,14 +734,16 @@ public interface Index
          *
          * @param index the index to be added
          */
-        void addIndex(Index index);
+        default void addIndex(Index index)
+        {}
 
         /**
          * Removes the specified {@link Index} from the members of this group.
          *
          * @param index the index to be removed
          */
-        void removeIndex(Index index);
+        default void removeIndex(Index index)
+        {}
 
         /**
          * Returns if this group contains the specified {@link Index}.
@@ -746,6 +752,16 @@ public interface Index
          * @return {@code true} if this group contains {@code index}, {@code false} otherwise
          */
         boolean containsIndex(Index index);
+
+        /**
+         * Returns whether this group can only ever contain a single index.
+         *
+         * @return {@code true} if this group only contains a single index, {@code false} otherwise
+         */
+        default boolean isSingleton()
+        {
+            return true;
+        }
 
         /**
          * Creates an new {@code Indexer} object for updates to a given partition.
--- a/src/java/org/apache/cassandra/index/Index.java
+++ b/src/java/org/apache/cassandra/index/Index.java
@@ -21,10 +21,7 @@
 package org.apache.cassandra.index;
 
 import java.io.UncheckedIOException;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.Callable;
 import java.util.function.Function;
 import java.util.function.Predicate;
@@ -274,6 +271,17 @@ public interface Index
      * @param registry the index registry to register the instance with
      */
     public void register(IndexRegistry registry);
+
+    /**
+     * Unregister current index when it's removed from system
+     *
+     * @param registry the index registry to register the instance with
+     */
+    default void unregister(IndexRegistry registry)
+    {
+        // for singleton index, the group key is the index itself
+        registry.unregisterIndex(this, new Index.Group.Key(this));
+    }
 
     /**
      * If the index implementation uses a local table to store its index data, this method should return a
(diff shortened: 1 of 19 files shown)


test/unit/org/apache/cassandra/index/sai/cql/IndexUnregistrationTest.java
```diff
@@ -0,0 +1,66 @@
...
+import static org.junit.Assert.assertNotSame;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
+public class IndexUnregistrationTest extends SAITester
+{
+    @Before
+    public void setup()
+    {
+        requireNetwork();
```
maedhroz: nit: Does the test need this? Seems like it passes without...

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
```diff
@@ -1182,7 +1182,8 @@ public void droppingIndexStopInitialIndexBuild() throws Throwable
 
         delayIndexBuilderCompletion.disable();
 
-        verifySSTableIndexes(indexName, 0);
+        assertNull(getCurrentIndexGroup());
+//        verifySSTableIndexes(indexName, 0);
```
maedhroz: nit: Clean up?

maedhroz (approved):
Left a couple small nits, but otherwise LGTM

src/java/org/apache/cassandra/index/Index.java
```diff
@@ -21,10 +21,7 @@
 package org.apache.cassandra.index;
 
 import java.io.UncheckedIOException;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
maedhroz: nit: revert to non-wildcard imports?

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
```diff
@@ -75,10 +75,12 @@ public class StorageAttachedIndexGroup implements Index.Group, INotificationCons
 {
     private static final Logger logger = LoggerFactory.getLogger(StorageAttachedIndexGroup.class);
 
+    public static final Index.Group.Key GROUP_KEY = new Index.Group.Key(StorageAttachedIndexGroup.class);
+
     private final TableQueryMetrics queryMetrics;
     private final TableStateMetrics stateMetrics;
     private final IndexGroupMetrics groupMetrics;
-    private final Set<StorageAttachedIndex> indexes = Sets.newConcurrentHashSet();
+    private final Set<StorageAttachedIndex> indexes = ConcurrentHashMap.newKeySet();
```
maedhroz, line 83: This is safe because `SIM#createIndex()` is `synchronized`?

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
```diff
@@ -128,14 +130,13 @@ public void removeIndex(Index index)
...
             deletePerSSTableFiles(baseCfs.getLiveSSTables());
-            baseCfs.getTracker().unsubscribe(this);
         }
     }
 
     @Override
     public void invalidate()
     {
-        // in case of dropping table, sstable contexts should already been removed by SSTableListChangedNotification.
+        // in case of removing last index from group, sstable contexts should already been removed by removeIndex
```
maedhroz, line 139: ...and per-column index files are removed via `SIM#reload()`, correct?

src/java/org/apache/cassandra/index/SecondaryIndexManager.java
```diff
@@ -1312,41 +1315,25 @@ public void registerIndex(Index index, Object groupKey, Supplier<Index.Group> gr
...
-            if (index.getBackingTable().isPresent())
-                throw new InvalidRequestException("Indexes belonging to a group of indexes shouldn't have a backing table");
-
-            group.addIndex(index);
-        }
+        group.addIndex(index);
     }
 
-    private Index unregisterIndex(String name)
+    public void unregisterIndex(Index removed, Index.Group.Key groupKey)
```
maedhroz, line 1322: nit: Throw `@Override` on there I guess?

src/java/org/apache/cassandra/index/SingletonIndexGroup.java
```diff
@@ -42,13 +44,11 @@
  */
 public class SingletonIndexGroup implements Index.Group
 {
-    private final Index delegate;
-    private final Set<Index> indexes;
+    private volatile Index delegate;
```
maedhroz: I've been mulling over the change here around `delegate`, `registerIndex()`, and `unregisterIndex()` (in `SIM`)...

I'd propose something like this:

1.) Revert all the changes in the singleton group here, and return to having the index passed to it at construction.
2.) Just have `addIndex()` and `removeIndex()` no-op, or even inherit default implementations from `Index`. (no-op might be useful even if just to have a comment there that explains why we no-op)
3.) In `SIM#unregisterIndex()`, take advantage of the fact that if the group contains the index to remove and only has 1 index anyway, we can remove the group without special-casing the logic for the singleton case. (There are other options, like having an `isSingleton()` method on `Group` that defaults to returning `true`.)

The benefits would be:
1.) The singleton group gets less complex again, and is trivially thread-safe again.
2.) I know we probably aren't worried too much about this, but any minor performance penalty for volatile access goes away again.
3.) Some existing comments in `SIM` become correct enough again that we don't have to alter them. (ex. "Remove the index from non-singleton groups...")

maedhroz (commented):
Left a few nits, and one non-trivial suggestion around the changes in `SingletonIndexGroup`

CC @jasonstack

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
jasonstack: it could be creating new index while iterating current index list for notifications

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
mike-tr-adamson: Hmm, there are a few comments floating around that I don't think are strictly correct. When you say "per-column index files are removed", do you mean when they are deleted? 

Apart from index rebuilds and index corruption, the only time per-column index files are deleted is when their associated `SSTableIndex` is released and has been marked as obsolete. This happens when the `IndexViewManager` associated with an index (via the `IndexContext`) is invalidated. This happens as part of the index invalidate task which is called from `Keyspace.unloadCf` during drop table.

src/java/org/apache/cassandra/index/SingletonIndexGroup.java
mike-tr-adamson: I have reverted the `SingletonIndexGroup` to the simple case and have added the `isSingleton` method. I tried some other routes but always ended up with needing some form of synchronization. The `isSingleton` means that `unregisterIndex` works for either case without special knowledge of the group.

maedhroz (approved):

jasonstack (approved):

maedhroz:
Committed as https://github.com/apache/cassandra/commit/e45c1092f91edd63591f562b2120ea6a5fd3edd5

PR #2776, opened 2023-10-05
[DO NOT MERGE] C-18798 ListType and Accord

henrikingo:
WIP sharing to get comments from @jacek-lewandowski and @blambov


blambov:
Could we do this by adding an `updatePathTimestamps` method in `AbstractType` that does nothing by default but is implemented by `ListType` to adjust all the timestamp part of its path UUIDs, and call it from `ColumnData.updateAllTimestamps`?

henrikingo:
Okay so I pulled all the new cep-15-accord code and I think this has been fixed in https://github.com/apache/cassandra/blob/79b3dc07e7f01a315df8db3724db4c1064b8eac9/src/java/org/apache/cassandra/service/accord/AccordKeyspace.java#L361-L363

Jacek had kindly helped with providing a simpler repro than what I was shooting for here. It no longer fails with the most recent cep-15-accord branch. On the other hand if a force 

                updateTimestamps=false;

...in the above, then the repro fails again.

Next steps:

* Rerun the original Elle consistency test to verify
* Seems like we can close C-18798 after confirming ^
* Maybe @jacek-lewandowski you want to submit your unit test as an independent patch

PR #2777, opened 2023-10-05
Enable Direct-IO feature for CommitLog files using Java native API's.

amitdpawar:
Enable Direct-IO feature for CommitLog files using Java native API's.

Cassandra by default uses memory mapped segments for commitlog files. This implementation is unable to scale on large core count machines for heavy insert type workload(s). This PR proposes Direct-IO feature for Commitlog files to improve the write speed. This feature known to provide larger bandwidth for IO activity. Java version 11 supports this feature by providing new API's.

### DIFF
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -589,9 +589,14 @@ commitlog_segment_size: 32MiB
 #     parameters:
 #         -
 
-# Configure commitlog disk access mode preferred order:direct_io, mmap, standard.
-# Default access mode is mmap. Direct I/O uses minimum page size to flush to disk.
-commitlog_disk_access_mode : mmap
+# CommitLog segments can use mmap (default) and direct (new) I/O modes to flush
+# the data to disk. Direct I/O is a new feature that minimizes cache effects by
+# using user space buffers. This helps in transferring data from/to disk at high
+# speed. Java enabled support for Direct-I/O feature from version 10 onwards.
+#   Refer: https://bugs.openjdk.org/browse/JDK-8164900
+# Direct-I/O is not enabled yet for compressed and encrypted based segments and
+# for replay (reading) commitlog files.
+commitlog_disk_access_mode_for_write: mmap
 
 # Compression to apply to SSTables as they flush for compressed tables.
 # Note that tables without compression enabled do not respect this flag.
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -589,6 +589,10 @@ commitlog_segment_size: 32MiB
 #     parameters:
 #         -
 
+# Configure commitlog disk access mode preferred order:direct_io, mmap, standard.
+# Default access mode is mmap. Direct I/O uses minimum page size to flush to disk.
+commitlog_disk_access_mode : mmap
+
 # Compression to apply to SSTables as they flush for compressed tables.
 # Note that tables without compression enabled do not respect this flag.
 #
@@ -2111,5 +2115,3 @@ drop_compact_storage_enabled: false
 #
 storage_compatibility_mode: CASSANDRA_4
 
-# Enable Direct I/O feature for Commitlog files. Java-11 forces minimum alignement to 4K and minimum disk write to 4K.
-direct_io_for_commitlog_enabled: true
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -2114,4 +2114,3 @@ drop_compact_storage_enabled: false
 #   a stable cluster. If a node from a previous version was started by accident we won't any longer toggle behaviors as when UPGRADING.
 #
 storage_compatibility_mode: CASSANDRA_4
-
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -2111,5 +2111,5 @@ drop_compact_storage_enabled: false
 #
 storage_compatibility_mode: CASSANDRA_4
 
-# Enable Direct IO feature for Commitlog files. Java-11 forces minimum alignement to 4K and minimum disk write to 4K.
-use_directio_for_commitlog: false
+# Enable Direct I/O feature for Commitlog files. Java-11 forces minimum alignement to 4K and minimum disk write to 4K.
+direct_io_for_commitlog_enabled: true
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -2110,3 +2110,6 @@ drop_compact_storage_enabled: false
 #   a stable cluster. If a node from a previous version was started by accident we won't any longer toggle behaviors as when UPGRADING.
 #
 storage_compatibility_mode: CASSANDRA_4
+
+# Enable Direct IO feature for Commitlog files. Java-11 forces minimum alignement to 4K and minimum disk write to 4K.
+use_directio_for_commitlog: false
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -589,14 +589,15 @@ commitlog_segment_size: 32MiB
 #     parameters:
 #         -
 
-# CommitLog segments can use mmap (default) and direct (new) I/O modes to flush
-# the data to disk. Direct I/O is a new feature that minimizes cache effects by
-# using user space buffers. This helps in transferring data from/to disk at high
-# speed. Java enabled support for Direct-I/O feature from version 10 onwards.
-#   Refer: https://bugs.openjdk.org/browse/JDK-8164900
-# Direct-I/O is not enabled yet for compressed and encrypted based segments and
+# CommitLog segments can use legacy (default), direct (new) and 'auto' (optimized
+# configuration) I/O modes to flush the data to disk. Direct I/O is a new feature
+# that minimizes cache effects by using user space buffers. This helps in
+# transferring data from/to disk at high speed. Java enabled support for
+# Direct-I/O feature from version 10 # onwards.
+# Refer: https://bugs.openjdk.org/browse/JDK-8164900
+# The same is not enabled yet for compressed and encrypted based segments and
 # for replay (reading) commitlog files.
-commitlog_disk_access_mode_for_write: mmap
+# commitlog_disk_access_mode: legacy
 
 # Compression to apply to SSTables as they flush for compressed tables.
 # Note that tables without compression enabled do not respect this flag.
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -390,8 +390,7 @@ public class Config
     public ParameterizedClass commitlog_compression;
     public FlushCompression flush_compression = FlushCompression.fast;
     public int commitlog_max_compression_buffers_in_pool = 3;
-    public boolean direct_io_for_commitlog_enabled = false;
-    public CommitLogDiskAccessMode commitlog_disk_access_mode = CommitLogDiskAccessMode.mmap;
+    public DiskAccessMode commitlog_disk_access_mode_for_write = DiskAccessMode.mmap;
     @Replaces(oldName = "periodic_commitlog_sync_lag_block_in_ms", converter = Converters.MILLIS_DURATION_INT, deprecated = true)
     public DurationSpec.IntMillisecondsBound periodic_commitlog_sync_lag_block;
     public TransparentDataEncryptionOptions transparent_data_encryption_options = new TransparentDataEncryptionOptions();
@@ -1149,13 +1148,14 @@ public class Config
         mmap,
         mmap_index_only,
         standard,
+        direct,
     }
 
     public enum CommitLogDiskAccessMode
     {
         standard,
         mmap,
-        direct_io,
+        direct,
     }
 
     public enum MemtableAllocationType
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -390,7 +390,7 @@ public class Config
     public ParameterizedClass commitlog_compression;
     public FlushCompression flush_compression = FlushCompression.fast;
     public int commitlog_max_compression_buffers_in_pool = 3;
-    public boolean use_directio_for_commitlog = false;
+    public boolean direct_io_for_commitlog_enabled = false;
     @Replaces(oldName = "periodic_commitlog_sync_lag_block_in_ms", converter = Converters.MILLIS_DURATION_INT, deprecated = true)
     public DurationSpec.IntMillisecondsBound periodic_commitlog_sync_lag_block;
     public TransparentDataEncryptionOptions transparent_data_encryption_options = new TransparentDataEncryptionOptions();
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -391,6 +391,7 @@ public class Config
     public FlushCompression flush_compression = FlushCompression.fast;
     public int commitlog_max_compression_buffers_in_pool = 3;
     public boolean direct_io_for_commitlog_enabled = false;
+    public CommitLogDiskAccessMode commitlog_disk_access_mode = CommitLogDiskAccessMode.mmap;
     @Replaces(oldName = "periodic_commitlog_sync_lag_block_in_ms", converter = Converters.MILLIS_DURATION_INT, deprecated = true)
     public DurationSpec.IntMillisecondsBound periodic_commitlog_sync_lag_block;
     public TransparentDataEncryptionOptions transparent_data_encryption_options = new TransparentDataEncryptionOptions();
@@ -1148,6 +1149,13 @@ public class Config
         mmap,
         mmap_index_only,
         standard,
+    }
+
+    public enum CommitLogDiskAccessMode
+    {
+        standard,
+        mmap,
+        direct_io,
     }
 
     public enum MemtableAllocationType
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -390,6 +390,7 @@ public class Config
     public ParameterizedClass commitlog_compression;
     public FlushCompression flush_compression = FlushCompression.fast;
     public int commitlog_max_compression_buffers_in_pool = 3;
+    public boolean use_directio_for_commitlog = false;
     @Replaces(oldName = "periodic_commitlog_sync_lag_block_in_ms", converter = Converters.MILLIS_DURATION_INT, deprecated = true)
     public DurationSpec.IntMillisecondsBound periodic_commitlog_sync_lag_block;
     public TransparentDataEncryptionOptions transparent_data_encryption_options = new TransparentDataEncryptionOptions();
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -390,7 +390,7 @@ public class Config
     public ParameterizedClass commitlog_compression;
     public FlushCompression flush_compression = FlushCompression.fast;
     public int commitlog_max_compression_buffers_in_pool = 3;
-    public DiskAccessMode commitlog_disk_access_mode_for_write = DiskAccessMode.mmap;
+    public DiskAccessMode commitlog_disk_access_mode = DiskAccessMode.legacy;
     @Replaces(oldName = "periodic_commitlog_sync_lag_block_in_ms", converter = Converters.MILLIS_DURATION_INT, deprecated = true)
     public DurationSpec.IntMillisecondsBound periodic_commitlog_sync_lag_block;
     public TransparentDataEncryptionOptions transparent_data_encryption_options = new TransparentDataEncryptionOptions();
@@ -1148,14 +1148,8 @@ public class Config
         mmap,
         mmap_index_only,
         standard,
-        direct,
-    }
-
-    public enum CommitLogDiskAccessMode
-    {
-        standard,
-        mmap,
-        direct,
+        legacy,
+        direct // Direct-I/O is enabled for commitlog disk only.
     }
 
     public enum MemtableAllocationType
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -511,6 +511,32 @@ public class DatabaseDescriptor
             logger.debug("Syncing log with a period of {}", conf.commitlog_sync_period.toString());
         }
 
+        // Only mmap & Direct-I/O access modes are supported. Standard access mode is preassumed if Compressed or Encrypted segment is used.
+        // This ensures default behavior is unchanged.
+        if  (conf.commitlog_disk_access_mode_for_write != Config.DiskAccessMode.mmap
+             && conf.commitlog_disk_access_mode_for_write != Config.DiskAccessMode.direct)
+        {
+            throw new ConfigurationException(String.format("Commitlog access mode '%s' is not supported. Supported modes are '%s' and '%s'", conf.commitlog_disk_access_mode_for_write,
+                                                                              Config.DiskAccessMode.mmap, Config.DiskAccessMode.direct));
+        } else {
+            Config.DiskAccessMode current_disk_access_mode = conf.commitlog_disk_access_mode_for_write;
+
+            if (getCommitLogCompression() != null || (getEncryptionContext() != null && getEncryptionContext().isEnabled()))
+            {
+                if (current_disk_access_mode == Config.DiskAccessMode.mmap)
+                {
+                    current_disk_access_mode = Config.DiskAccessMode.standard;
+                }
+                else if (current_disk_access_mode == Config.DiskAccessMode.direct)
+                {
+                    // Compressed and Encrypted segments are not yet supported with Direct-I/O feature.
+                    throw new ConfigurationException("Direct I/O does not support Compressed or Encrypted segment yet.");
+                }
+            }
+
+            logger.info("Commitlog disk using '{}' access mode for write and '{}' mode for replay.", current_disk_access_mode, Config.DiskAccessMode.standard);
+        }
+
         /* evaluate the DiskAccessMode Config directive, which also affects indexAccessMode selection */
         if (conf.disk_access_mode == Config.DiskAccessMode.auto)
         {
@@ -2652,19 +2678,19 @@ public class DatabaseDescriptor
     /**
      * Return commitlog disk access mode.
      */
-    public static Config.CommitLogDiskAccessMode getCommitLogDiskAccessMode()
+    public static Config.DiskAccessMode getCommitLogDiskAccessMode()
     {
-        return conf.commitlog_disk_access_mode;
+        return conf.commitlog_disk_access_mode_for_write;
     }
 
     public static boolean isCommitLogUsingDirectIO()
     {
-        return getCommitLogDiskAccessMode() == Config.CommitLogDiskAccessMode.direct_io;
+        return getCommitLogDiskAccessMode() == Config.DiskAccessMode.direct;
     }
 
-    public static void setCommitLogDiskAccessMode(Config.CommitLogDiskAccessMode new_mode)
+    public static void setCommitLogDiskAccessMode(Config.DiskAccessMode new_mode)
     {
-        conf.commitlog_disk_access_mode = new_mode;
+        conf.commitlog_disk_access_mode_for_write = new_mode;
     }
 
     public static String getSavedCachesLocation()
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -2649,6 +2649,11 @@ public class DatabaseDescriptor
         conf.commitlog_segment_size = new DataStorageSpec.IntMebibytesBound(sizeMebibytes);
     }
 
+    public static boolean getDirectIOStatus()
+    {
+        return conf.use_directio_for_commitlog;
+    }
+
     public static String getSavedCachesLocation()
     {
         return conf.saved_caches_directory;
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -513,21 +513,30 @@ public class DatabaseDescriptor
 
         // Only mmap & Direct-I/O access modes are supported. Standard access mode is preassumed if Compressed or Encrypted segment is used.
         // This ensures default behavior is unchanged.
-        if  (conf.commitlog_disk_access_mode_for_write != Config.DiskAccessMode.mmap
-             && conf.commitlog_disk_access_mode_for_write != Config.DiskAccessMode.direct)
+        if  (conf.commitlog_disk_access_mode != Config.DiskAccessMode.auto
+             && conf.commitlog_disk_access_mode != Config.DiskAccessMode.legacy
+             && conf.commitlog_disk_access_mode != Config.DiskAccessMode.direct)
         {
-            throw new ConfigurationException(String.format("Commitlog access mode '%s' is not supported. Supported modes are '%s' and '%s'", conf.commitlog_disk_access_mode_for_write,
-                                                                              Config.DiskAccessMode.mmap, Config.DiskAccessMode.direct));
-        } else {
-            Config.DiskAccessMode current_disk_access_mode = conf.commitlog_disk_access_mode_for_write;
+            throw new ConfigurationException(String.format("Commitlog access mode '%s' is not supported. Supported modes are '%s', '%s' and '%s'",
+                                                            conf.commitlog_disk_access_mode, Config.DiskAccessMode.auto,
+                                                            Config.DiskAccessMode.legacy, Config.DiskAccessMode.direct));
+        }
+        else
+        {
+            Config.DiskAccessMode current_disk_access_mode = conf.commitlog_disk_access_mode;
+
+            if (current_disk_access_mode == Config.DiskAccessMode.auto)
+            {
+                current_disk_access_mode = conf.disk_optimization_strategy == Config.DiskOptimizationStrategy.ssd ? Config.DiskAccessMode.direct
+                                                                                                                    : Config.DiskAccessMode.legacy;
+                // This is necessary for other functions to get the new mode determined through auto.
+                conf.commitlog_disk_access_mode = current_disk_access_mode;
+                logger.info("DiskAccessMode 'auto' determined to be '{}' for Commitlog disk", current_disk_access_mode);
+            }
 
             if (getCommitLogCompression() != null || (getEncryptionContext() != null && getEncryptionContext().isEnabled()))
             {
-                if (current_disk_access_mode == Config.DiskAccessMode.mmap)
-                {
-                    current_disk_access_mode = Config.DiskAccessMode.standard;
-                }
-                else if (current_disk_access_mode == Config.DiskAccessMode.direct)
+                if (current_disk_access_mode == Config.DiskAccessMode.direct)
                 {
                     // Compressed and Encrypted segments are not yet supported with Direct-I/O feature.
                     throw new ConfigurationException("Direct I/O does not support Compressed or Encrypted segment yet.");
@@ -549,6 +558,10 @@ public class DatabaseDescriptor
             conf.disk_access_mode = Config.DiskAccessMode.standard;
             indexAccessMode = Config.DiskAccessMode.mmap;
             logger.info("DiskAccessMode is {}, indexAccessMode is {}", conf.disk_access_mode, indexAccessMode);
+        }
+        else if (conf.disk_access_mode == Config.DiskAccessMode.direct || conf.disk_access_mode == Config.DiskAccessMode.legacy)
+        {
+            throw new ConfigurationException(String.format("DiskAccessMode '%s' and '%s' are not supported", Config.DiskAccessMode.direct, Config.DiskAccessMode.legacy));
         }
         else
         {
@@ -2680,7 +2693,7 @@ public class DatabaseDescriptor
      */
     public static Config.DiskAccessMode getCommitLogDiskAccessMode()
     {
-        return conf.commitlog_disk_access_mode_for_write;
+        return conf.commitlog_disk_access_mode;
     }
 
     public static boolean isCommitLogUsingDirectIO()
@@ -2690,7 +2703,7 @@ public class DatabaseDescriptor
 
     public static void setCommitLogDiskAccessMode(Config.DiskAccessMode new_mode)
     {
-        conf.commitlog_disk_access_mode_for_write = new_mode;
+        conf.commitlog_disk_access_mode = new_mode;
     }
 
     public static String getSavedCachesLocation()
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -2649,9 +2649,12 @@ public class DatabaseDescriptor
         conf.commitlog_segment_size = new DataStorageSpec.IntMebibytesBound(sizeMebibytes);
     }
 
-    public static boolean getDirectIOStatus()
+    /**
+     * Return user configured input for Commitlog files.
+     */
+    public static boolean getDirectIOForCommitlogEnabled()
     {
-        return conf.use_directio_for_commitlog;
+        return conf.direct_io_for_commitlog_enabled;
     }
 
     public static String getSavedCachesLocation()
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -2696,11 +2696,6 @@ public class DatabaseDescriptor
         return conf.commitlog_disk_access_mode;
     }
 
-    public static boolean isCommitLogUsingDirectIO()
-    {
-        return getCommitLogDiskAccessMode() == Config.DiskAccessMode.direct;
-    }
-
     public static void setCommitLogDiskAccessMode(Config.DiskAccessMode new_mode)
     {
         conf.commitlog_disk_access_mode = new_mode;
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -2650,11 +2650,21 @@ public class DatabaseDescriptor
     }
 
     /**
-     * Return user configured input for Commitlog files.
+     * Return commitlog disk access mode.
      */
-    public static boolean getDirectIOForCommitlogEnabled()
+    public static Config.CommitLogDiskAccessMode getCommitLogDiskAccessMode()
     {
-        return conf.direct_io_for_commitlog_enabled;
+        return conf.commitlog_disk_access_mode;
+    }
+
+    public static boolean isCommitLogUsingDirectIO()
+    {
+        return getCommitLogDiskAccessMode() == Config.CommitLogDiskAccessMode.direct_io;
+    }
+
+    public static void setCommitLogDiskAccessMode(Config.CommitLogDiskAccessMode new_mode)
+    {
+        conf.commitlog_disk_access_mode = new_mode;
     }
 
     public static String getSavedCachesLocation()
--- a/src/java/org/apache/cassandra/db/commitlog/AbstractCommitLogSegmentManager.java
+++ b/src/java/org/apache/cassandra/db/commitlog/AbstractCommitLogSegmentManager.java
@@ -117,8 +117,9 @@ public abstract class AbstractCommitLogSegmentManager
                                     ? BufferType.ON_HEAP
                                     : commitLog.configuration.getCompressor().preferredBufferType();
 
-        // Identify minimum block size used for Direct-I/O. allocateDirect does not return from aligned space.
-        // To avoid write errors additional block size is requested and next aligned it minimum block size.
+        // The direct buffer must be aligned with the file system block size. We cannot enforce that during
+        // allocation, but we can get an aligned slice from the allocated buffer. The buffer must be oversized by the
+        // alignment unit to make it possible.
         int minimumAllowedAlign = !commitLog.configuration.isDirectIOEnabled()
                                   ? 0 : CommitLogSegment.getDirectIOMinimumAlignement(storageDirectory, "CommitLog-DirectIO-Test.log");
 
--- a/src/java/org/apache/cassandra/db/commitlog/AbstractCommitLogSegmentManager.java
+++ b/src/java/org/apache/cassandra/db/commitlog/AbstractCommitLogSegmentManager.java
@@ -111,12 +111,19 @@ public abstract class AbstractCommitLogSegmentManager
     {
         // For encrypted segments we want to keep the compression buffers on-heap as we need those bytes for encryption,
         // and we want to avoid copying from off-heap (compression buffer) to on-heap encryption APIs
-        BufferType bufferType = commitLog.configuration.useEncryption() || !commitLog.configuration.useCompression()
-                                ? BufferType.ON_HEAP
-                                : commitLog.configuration.getCompressor().preferredBufferType();
+        BufferType bufferType = commitLog.configuration.isDirectIOEnabled()
+                                ? BufferType.OFF_HEAP
+                                  : commitLog.configuration.useEncryption() || !commitLog.configuration.useCompression()
+                                    ? BufferType.ON_HEAP
+                                    : commitLog.configuration.getCompressor().preferredBufferType();
+
+        // Identify minimum block size used for Direct-I/O. allocateDirect does not return from aligned space.
+        // To avoid write errors additional block size is requested and next aligned it minimum block size.
+        int minimumAllowedAlign = !commitLog.configuration.isDirectIOEnabled()
+                                  ? 0 : CommitLogSegment.getDirectIOMinimumAlignement(storageDirectory, "CommitLog-DirectIO-Test.log");
 
         this.bufferPool = new SimpleCachedBufferPool(DatabaseDescriptor.getCommitLogMaxCompressionBuffersInPool(),
-                                                     DatabaseDescriptor.getCommitLogSegmentSize(),
+                                                     DatabaseDescriptor.getCommitLogSegmentSize() + minimumAllowedAlign,
                                                      bufferType);
 
 
--- a/src/java/org/apache/cassandra/db/commitlog/CommitLog.java
+++ b/src/java/org/apache/cassandra/db/commitlog/CommitLog.java
@@ -130,9 +130,7 @@ public class CommitLog implements CommitLogMBean
         // register metrics
         metrics.attach(executor, segmentManager);
 
-	if (configuration.isDirectIOEnabled()) {
-          logger.info("Direct-IO feature enabled for Commitlog files.");
-	}
+        logger.info("Direct I/O feature for Commitlog files is : " + configuration.isDirectIOEnabled() + ".");
     }
 
     /**
@@ -638,7 +636,7 @@ public class CommitLog implements CommitLogMBean
             this.compressorClass = compressorClass;
             this.compressor = compressorClass != null ? CompressionParams.createCompressor(compressorClass) : null;
             this.encryptionContext = encryptionContext;
-            this.useDirectIO = DatabaseDescriptor.getDirectIOStatus();
+            this.useDirectIO = DatabaseDescriptor.getDirectIOForCommitlogEnabled();
         }
 
         /**
--- a/src/java/org/apache/cassandra/db/commitlog/CommitLog.java
+++ b/src/java/org/apache/cassandra/db/commitlog/CommitLog.java
@@ -40,6 +40,7 @@ import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import org.apache.cassandra.config.Config;
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.config.ParameterizedClass;
 import org.apache.cassandra.db.Mutation;
@@ -104,7 +105,8 @@ public class CommitLog implements CommitLogMBean
     CommitLog(CommitLogArchiver archiver, Function<CommitLog, AbstractCommitLogSegmentManager> segmentManagerProvider)
     {
         this.configuration = new Configuration(DatabaseDescriptor.getCommitLogCompression(),
-                                               DatabaseDescriptor.getEncryptionContext());
+                                               DatabaseDescriptor.getEncryptionContext(),
+                                               DatabaseDescriptor.getCommitLogDiskAccessMode());
         DatabaseDescriptor.createAllDirectories();
 
         this.archiver = archiver;
@@ -130,7 +132,24 @@ public class CommitLog implements CommitLogMBean
         // register metrics
         metrics.attach(executor, segmentManager);
 
-        logger.info("Direct I/O feature for Commitlog files is : " + configuration.isDirectIOEnabled() + ".");
+        String diskMode = " ";
+
+        switch (DatabaseDescriptor.getCommitLogDiskAccessMode())
+        {
+            case standard:
+                diskMode = "Standard (buffered)";
+                break;
+            case mmap:
+                diskMode = "MMap(memory mmaped)";
+                break;
+            case direct_io:
+                diskMode = "Direct (non-buffered)";
+                break;
+            default:
+                throw new IllegalArgumentException("Unknown commitlog disk access mode type: " + DatabaseDescriptor.getCommitLogDiskAccessMode());
+        }
+
+	logger.info("Using " + diskMode + " I/O for CommitLog files.");
     }
 
     /**
@@ -523,7 +542,8 @@ public class CommitLog implements CommitLogMBean
     synchronized public void resetConfiguration()
     {
         configuration = new Configuration(DatabaseDescriptor.getCommitLogCompression(),
-                                          DatabaseDescriptor.getEncryptionContext());
+                                          DatabaseDescriptor.getEncryptionContext(),
+                                          DatabaseDescriptor.getCommitLogDiskAccessMode());
     }
 
     /**
@@ -614,7 +634,7 @@ public class CommitLog implements CommitLogMBean
         /**
          * Flag used to shows user configured Direct-IO status.
          */
-        private final boolean useDirectIO;
+        private final Config.CommitLogDiskAccessMode diskMode;
 
         /**
          * The compressor class.
@@ -631,12 +651,13 @@ public class CommitLog implements CommitLogMBean
          */
         private EncryptionContext encryptionContext;
 
-        public Configuration(ParameterizedClass compressorClass, EncryptionContext encryptionContext)
+        public Configuration(ParameterizedClass compressorClass, EncryptionContext encryptionContext,
+                             Config.CommitLogDiskAccessMode diskAccessMode)
         {
             this.compressorClass = compressorClass;
             this.compressor = compressorClass != null ? CompressionParams.createCompressor(compressorClass) : null;
             this.encryptionContext = encryptionContext;
-            this.useDirectIO = DatabaseDescriptor.getDirectIOForCommitlogEnabled();
+            this.diskMode = diskAccessMode;
         }
 
         /**
@@ -694,12 +715,30 @@ public class CommitLog implements CommitLogMBean
         }
 
         /**
-         * Returns DirectIO/non-buffer IO status
-         * @return DirectIO/non-buffer IO status
+         * Returns Direct-IO/non-buffer used for CommitLog IO.
+         * @return Direct-IO used for CommitLog IO
          */
         public boolean isDirectIOEnabled()
         {
-            return useDirectIO;
+            return diskMode == Config.CommitLogDiskAccessMode.direct_io ;
+        }
+
+        /**
+         * Returns MMAP used for CommitLog IO.
+         * @return MMAP used for CommitLog IO
+         */
+        public boolean isMMAPEnabled()
+        {
+            return diskMode == Config.CommitLogDiskAccessMode.mmap ;
+        }
+
+        /**
+         * Returns Standard or buffered I/O used for CommitLog IO.
+         * @return Standard or buffered I/O used for CommitLog IO
+         */
+        public boolean isBufferedIOEnabled()
+        {
+            return diskMode == Config.CommitLogDiskAccessMode.standard ;
         }
     }
 }
--- a/src/java/org/apache/cassandra/db/commitlog/CommitLog.java
+++ b/src/java/org/apache/cassandra/db/commitlog/CommitLog.java
@@ -72,7 +72,7 @@ import static org.apache.cassandra.utils.FBUtilities.updateChecksumInt;
  */
 public class CommitLog implements CommitLogMBean
 {
-    private static final Logger logger = LoggerFactory.getLogger(CommitLog.class);
+    public static final Logger logger = LoggerFactory.getLogger(CommitLog.class);
 
     public static final CommitLog instance = CommitLog.construct();
 
@@ -129,6 +129,10 @@ public class CommitLog implements CommitLogMBean
 
         // register metrics
         metrics.attach(executor, segmentManager);
+
+	if (configuration.isDirectIOEnabled()) {
+          logger.info("Direct-IO feature enabled for Commitlog files.");
+	}
     }
 
     /**
@@ -609,6 +613,7 @@ public class CommitLog implements CommitLogMBean
 
     public static final class Configuration
     {
+        private final boolean useDirectIO;
         /**
          * The compressor class.
          */
@@ -629,6 +634,7 @@ public class CommitLog implements CommitLogMBean
             this.compressorClass = compressorClass;
             this.compressor = compressorClass != null ? CompressionParams.createCompressor(compressorClass) : null;
             this.encryptionContext = encryptionContext;
+            this.useDirectIO = DatabaseDescriptor.getDirectIOStatus();
         }
 
         /**
@@ -683,6 +689,15 @@ public class CommitLog implements CommitLogMBean
         public EncryptionContext getEncryptionContext()
         {
             return encryptionContext;
+        }
+
+        /**
+         * Returns DirectIO/non-buffer IO status
+         * @return DirectIO/non-buffer IO status
+         */
+        public boolean isDirectIOEnabled()
+        {
+            return useDirectIO;
         }
     }
 }
--- a/src/java/org/apache/cassandra/db/commitlog/CommitLog.java
+++ b/src/java/org/apache/cassandra/db/commitlog/CommitLog.java
@@ -613,7 +613,11 @@ public class CommitLog implements CommitLogMBean
 
     public static final class Configuration
     {
+        /**
+         * Flag used to shows user configured Direct-IO status.
+         */
         private final boolean useDirectIO;
+
         /**
          * The compressor class.
          */
--- a/src/java/org/apache/cassandra/db/commitlog/CommitLog.java
+++ b/src/java/org/apache/cassandra/db/commitlog/CommitLog.java
@@ -131,25 +131,6 @@ public class CommitLog implements CommitLogMBean
 
         // register metrics
         metrics.attach(executor, segmentManager);
-
-        String diskMode = " ";
-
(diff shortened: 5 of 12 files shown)


src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2649,6 +2649,11 @@ public static void setCommitLogSegmentSize(int sizeMebibytes)
         conf.commitlog_segment_size = new DataStorageSpec.IntMebibytesBound(sizeMebibytes);
     }
 
+    public static boolean getDirectIOStatus()
```
Maxwell-Guo: I think we can rename the function name to enableDirectIOForCommitlog

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -129,6 +129,10 @@ private static CommitLog construct()
 
         // register metrics
         metrics.attach(executor, segmentManager);
+
+	if (configuration.isDirectIOEnabled()) {
```
Maxwell-Guo: I think we may do not need this logging ~~~ or just  logger.info("Use direcrtor io for commit log " +  configuration.isDirectIOEnabled())

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -609,6 +613,7 @@ private static String addAdditionalInformationIfPossible(String msg)
 
     public static final class Configuration
     {
+        private final boolean useDirectIO;
```
Maxwell-Guo: Add comment like : the flag to show wether use dir io for commitlog

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
```diff
@@ -138,7 +139,8 @@ static CommitLogSegment createSegment(CommitLog commitLog, AbstractCommitLogSegm
         Configuration config = commitLog.configuration;
         CommitLogSegment segment = config.useEncryption() ? new EncryptedSegment(commitLog, manager)
                                                           : config.useCompression() ? new CompressedSegment(commitLog, manager)
-                                                                                    : new MemoryMappedSegment(commitLog, manager);
+                                                                                    : ! config.isDirectIOEnabled() ? new MemoryMappedSegment(commitLog, manager)
```
Maxwell-Guo: what about : config.isDirectIOEnabled() ? new DirectIOSegment(commitLog, manager) : new MemoryMappedSegment(commitLog, manager);

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
```diff
@@ -175,7 +177,10 @@ static long getNextId()
 
         try
         {
-            channel = FileChannel.open(logFile.toPath(), StandardOpenOption.WRITE, StandardOpenOption.READ, StandardOpenOption.CREATE);
+            if(commitLog.configuration.isDirectIOEnabled())
+                channel = FileChannel.open(logFile.toPath(), StandardOpenOption.WRITE, StandardOpenOption.READ, StandardOpenOption.CREATE,ExtendedOpenOption.DIRECT);
```
Maxwell-Guo: `channel = FileChannel.open(logFile.toPath(), StandardOpenOption.WRITE, StandardOpenOption.READ, StandardOpenOption.CREATE, ExtendedOpenOption.DIRECT);`

Maxwell-Guo: miss a space before ExtendedOpenOption.DIRECT

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,153 @@
...
+
+        // mark the initial sync marker as uninitialised
+        int firstSync = buffer.position();
+        buffer.putInt(firstSync + 0, 0);
+        buffer.putInt(firstSync + 4, 0);
+
+        // Testing shows writing initial bytes takes some time. During peak load, it helps
+        // lot and Syncer thread can actually do flush activity. Making this initial
+        // slow operation to be executed by Allocator thread here.
+        int oldLastSyncedOffset=lastSyncedOffset;
```
Maxwell-Guo: int oldLastSyncedOffset = lastSyncedOffset;

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,153 @@
...
+        // 8 bytes are written above.
+        lastSyncedOffset=8;
```
Maxwell-Guo: lastSyncedOffset = 8;

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,153 @@
...
+        flush(0,lastSyncedOffset);
```
Maxwell-Guo: flush(0, lastSyncedOffset);

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,153 @@
...
+        lastSyncedOffset = oldLastSyncedOffset;
+    }
+
+    ByteBuffer createBuffer(CommitLog commitLog)
+    {
+        if (minimumAllowedAlign ==0)
```
Maxwell-Guo: minimumAllowedAlign == 0

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+
+        int cl_size = DatabaseDescriptor.getCommitLogSegmentSize();
+        original = ByteBuffer.allocateDirect(cl_size + minimumAllowedAlign);
+
+        ByteBuffer alignedBuffer = original.alignedSlice(minimumAllowedAlign);
+        assert alignedBuffer.limit() >= cl_size : String.format("Bytebuffer slicing failed to get required buffer size (required=%d,current size=%d", cl_size, alignedBuffer.limit());
+
+        assert alignedBuffer.alignmentOffset(0, minimumAllowedAlign) ==0 : "Index 0 should be aligned to 4K page size";
```
Maxwell-Guo: minimumAllowedAlign) == 0 :

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,153 @@
...
+        assert alignedBuffer.alignmentOffset(alignedBuffer.limit(), minimumAllowedAlign) ==0 : "Limit should be aligned to 4K page size" ;
```
Maxwell-Guo: minimumAllowedAlign) == 0 :

Maxwell-Guo (commented):
Besides, I want to know how to make a test ?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
amitdpawar: name "enableDirectIOForCommitlog" wont it be confusing to get the current user configure value of Direct-IO ?

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
amitdpawar: OK.

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
amitdpawar: will update as per the suggestion.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
Maxwell-Guo: But I don't think a boolean value can accurately describe the status  as your function name is getDirectIOStatus , so what does true and false stand for ？
And this is only for commit log , right ?

amitdpawar: yes, it returns the Direct-IO user configured input give in Cassandra.yaml as an example given below. true = enabled and false = disabled.

use_directio_for_commitlog: <true or false>

Maxwell-Guo: it is only for commitlog , and besides I think the function name can describe status more accurately than getDirectIOStatus, as a boolean value has only true or false. And I don't known what does true /false stands for in IO status.

Maxwell-Guo: so I think useDirectIOForCommitlog is also better than getDirectIOStatus , just in my mind.

amitdpawar: I will change as per your suggestion.

amitdpawar: I changed similar to what is followed in Cassandra.yaml.

amitdpawar:
> Besides, I want to know how to make a test ?

Thanks to "Ariel Weisberg" for writing the initial testcase to test the Direct I/O feature. This test [https://pastebin.com/JXJv0tSd](https://pastebin.com/JXJv0tSd) was very useful and showed how fast Direct I/O feature is able to write. Right now, it fails to build so trying to fix it.  Also, I am trying to understand how to do unit testing without affecting existing testcases. Any help will be appreciated.

amitdpawar:
Direct I/O testing can be done using existing testcase now. "test/long/org/apache/cassandra/db/commitlog/CommitLogStressTest.java" modified and following table is test result for "test/long/org/apache/cassandra/db/commitlog/PeriodicCommitLogStressTest.java" testcase. It can be changed or improved based on feedback.
image

amitdpawar:
I could not add new testcase for Direct I/O feature so modified existing one. Build was reporting error so thought to provide this change to allow testing.

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
```diff
@@ -175,7 +177,10 @@ static long getNextId()
 
         try
         {
-            channel = FileChannel.open(logFile.toPath(), StandardOpenOption.WRITE, StandardOpenOption.READ, StandardOpenOption.CREATE);
+            if(commitLog.configuration.isDirectIOEnabled())
+                channel = FileChannel.open(logFile.toPath(), StandardOpenOption.WRITE, StandardOpenOption.READ, StandardOpenOption.CREATE, ExtendedOpenOption.DIRECT);
```
Maxwell-Guo, line 185: seems have not make an implementation for commitlog reader ?

this patch only work for write path , right ?

Maxwell-Guo:
> I could not add new testcase for Direct I/O feature so modified existing one. Build was reporting error so thought to provide this change to allow testing.

for data validation test , I think we can make a test that write using direct io, and using commitlogreader to read the file,then we can just make the data validation test.  also how can we make sure that we are using direct-io ? @blambov Any suggestions?

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
amitdpawar: Yes. Reading not done yet.

conf/cassandra.yaml
```diff
@@ -2110,3 +2110,6 @@ drop_compact_storage_enabled: false
 #   a stable cluster. If a node from a previous version was started by accident we won't any longer toggle behaviors as when UPGRADING.
 #
 storage_compatibility_mode: CASSANDRA_4
+
+# Enable Direct I/O feature for Commitlog files. Java-11 forces minimum alignement to 4K and minimum disk write to 4K.
+direct_io_for_commitlog_enabled: true
```
blambov: This setting should move with the other commit log settings, e.g. [here](https://github.com/apache/cassandra/blob/624af90bab0d7f5f08de0b9d30cb212186c6bbdf/conf/cassandra.yaml#L582), and the value needs to be an enum (with values `direct` and `mmap`), so that users can understand what each option means.
See e.g. the implementation of `disk_access_mode`.

It also makes sense to unit test that the option is properly understood (i.e. that the object created by `CommitLog.createSegment` is of the correct type).

While I personally would prefer to switch to direct by default, please ask on dev mailing list or at least in the JIRA ticket if anyone is opposed to that as it might catch someone by surprise.

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
blambov, line 185: Does this also change the compressed and encrypted segments to use non-cached access? Could we also benchmark LZ4-compressed commit log with and without the flag?

blambov (commented):
For a change of this sort the proper test is to run the full battery of unit tests with the flag having the two different values. To do this, find a test configuration that is not already changing the commit log options (e.g. [this one](https://github.com/apache/cassandra/blob/624af90bab0d7f5f08de0b9d30cb212186c6bbdf/build.xml#L1272)), and switch it to the non-default value of the flag.

Maxwell-Guo:
@blambov thanks , Can you make some modifications based on @blambov 's feedback? @amitdpawar

test/long/org/apache/cassandra/db/commitlog/CommitLogStressTest.java
```diff
@@ -48,9 +48,9 @@
 import org.apache.cassandra.db.marshal.UTF8Type;
 import org.apache.cassandra.db.partitions.PartitionUpdate;
 import org.apache.cassandra.db.rows.*;
-import org.apache.cassandra.io.compress.DeflateCompressor;
-import org.apache.cassandra.io.compress.LZ4Compressor;
-import org.apache.cassandra.io.compress.SnappyCompressor;
+//import org.apache.cassandra.io.compress.DeflateCompressor;
```
smiklosovic: watch out here, remove this upon commit

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
amitdpawar: Direct I/O failed to run with LZ4-compressed and encrypted segments as expected due to alignment. Good to fix once this PR is upstreamed.

conf/cassandra.yaml
amitdpawar: I added commitlog disk access mode flag to take care existing "standard" for "Compressed" and "Encrypted" segments. Then, mmap for current default and direct_io as a new mode for this PR. These options made it easier to test with existing testcases only. If still required, then will change as per the earlier suggestion.

test/long/org/apache/cassandra/db/commitlog/CommitLogStressTest.java
amitdpawar: Reverted. thanks.

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
amitdpawar: Screenshot of reported error.

image

conf/cassandra.yaml
```diff
@@ -589,6 +589,10 @@ commitlog_segment_size: 32MiB
 #     parameters:
 #         -
 
+# Configure commitlog disk access mode preferred order:direct_io, mmap, standard.
+# Default access mode is mmap. Direct I/O uses minimum page size to flush to disk.
+commitlog_disk_access_mode : mmap
```
Maxwell-Guo: I think we can rename this to commitlog_disk_access_mode_for_write or some more concise words, or 
may be better add some more detailed descriptions, to tell the users that this configuration only work for commitlog write(sync) not for commitlog read(replay)
I would prefer to add some description , so that when read commitlog at replay stage, people will not need to modify this configuration .
And I think you should also give some description on mmap, stand(buffer),direct_io , the difference between them.

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
Maxwell-Guo: I have a new suggestion here, for commitlog log replay , if the configuration commitlog_disk_access_mode is direct_io , we can just logging here : do not support directo io for commitlog replay, use default (stand) here, so when we support direct io in the future , we will not need to introduce a new configuration for read.

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -390,6 +390,8 @@ public static class SSTableConfig
     public ParameterizedClass commitlog_compression;
     public FlushCompression flush_compression = FlushCompression.fast;
     public int commitlog_max_compression_buffers_in_pool = 3;
+    public boolean direct_io_for_commitlog_enabled = false;
```
Maxwell-Guo: what about remove this variable？And if we want to know wether we can use direct ion ,just retun ommitLogDiskAccessMode.direct_io == commitlog_disk_access_mode  
?

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -129,6 +131,25 @@ private static CommitLog construct()
...
+                diskMode = "Standard (buffered)";
+                break;
+            case mmap:
+                diskMode = "MMap(memory mmaped)";
+                break;
+            case direct_io:
+                diskMode = "Direct (non-buffered)";
+                break;
+            default:
+                throw new IllegalArgumentException("Unknown commitlog disk access mode type: " + DatabaseDescriptor.getCommitLogDiskAccessMode());
```
Maxwell-Guo: `   logger.info("Disk access mode for commitlog write is " + DatabaseDescriptor.getCommitLogDiskAccessMode() );`
And if we add some detailed descriptions in the cassandra.yaml as what I described  upper , I think there may have no needs for this part here. Just tell the user what disk mode we are using is enough.

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -684,5 +713,32 @@ public EncryptionContext getEncryptionContext()
...
+        public boolean isDirectIOEnabled()
+        {
+            return diskAccessMode == Config.CommitLogDiskAccessMode.direct_io;
+        }
+
+        /**
+         * Returns MMAP used for CommitLog IO.
+         * @return MMAP used for CommitLog IO
+         */
+        public boolean isMMAPEnabled()
```
Maxwell-Guo: Do we need this method here ?

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -684,5 +713,32 @@ public EncryptionContext getEncryptionContext()
...
+        }
+
+        /**
+         * Returns Standard or buffered I/O used for CommitLog IO.
+         * @return Standard or buffered I/O used for CommitLog IO
+         */
+        public boolean isBufferedIOEnabled()
+        {
+            return diskAccessMode == Config.CommitLogDiskAccessMode.standard;
+        }
```
Maxwell-Guo, line 714: Do we need this method here ?

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
Maxwell-Guo: as I have said before ，what about add a log for read ? if commitlog_disk_access_mode is direct_io , and for read here, we can add a warn log ,and tell the user, the direct io for read is not supported now, use the default or other instand.

test/conf/cassandra.yaml
```diff
@@ -8,6 +8,7 @@ memtable_allocation_type: offheap_objects
 commitlog_sync: batch
 commitlog_segment_size: 5MiB
 commitlog_directory: build/test/cassandra/commitlog
+commitlog_disk_access_mode : mmap
```
Maxwell-Guo: mmap ? Aren't we going to test direct_io ?

conf/cassandra.yaml
jacek-lewandowski: Given what @blambov mentioned and the test results - is there any reason to not immediately deprecate this setting in favour of hardcoding it to direct in the next release?

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1149,6 +1151,13 @@ public enum DiskAccessMode
         standard,
     }
 
+    public enum CommitLogDiskAccessMode
```
jacek-lewandowski: Maybe we can avoid creating another enum? The existing `DiskAccessMode` can be probably used as well with added `direct` option and some validation in places where it is used whether the selected access mode is supported or not? (TBH, I'd rather remove `mmap_index_only` from `DiskAccessMode` to make it more generic - disk access mode for indexes could be specified explicitly instead)

Maxwell-Guo: Is this flexible? As directio is only work for commitlog write, It May not be supported  for normal sstable read /write and commitlog read and others situation yet if DiskAccessMode is direct_io.

Maxwell-Guo: or if we can just add a configuration(may be we can call commitlogAccessMode) like indexAccessMode so that we may use DiskAccessMode's configuration

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,163 @@
...
+ * ByteBuffer.position, ByteBuffer.limit and FileChannel.position to page size (4K).
+ * Java-11 forces minimum page size to be written to disk with Direct-IO.
+ */
+public class DirectIOSegment extends CommitLogSegment
+{
+    ByteBuffer original;
+    static int minimumAllowedAlign;
+
+    // Needed to track number of bytes written to disk in multiple of page size.
+    int lastWritten = 0;
```
Maxwell-Guo: int -> long ?

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,163 @@
...
+            {
+                minimumAllowedAlign = (int)Files.getFileStore(logFile.toPath()).getBlockSize();
+            }
+            catch (IOException e)
+            {
+                throw new FSWriteError(e, logFile);
+            }
+        }
+
+        int cl_size = DatabaseDescriptor.getCommitLogSegmentSize();
```
Maxwell-Guo: I think we can rename this to  segmentSize

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,163 @@
...
+    protected void flush(int startMarker, int nextMarker)
+    {
+
+        try
+        {
+            // lastSyncedOffset is synced to disk. Align lastSyncedOffset to start of 4K page
+            // and nextMarker to end of 4K page to avoid write errors.
+            long filePosition = lastSyncedOffset;
+            ByteBuffer duplicate = buffer.duplicate();
+            long file_old_position = channel.position();
```
Maxwell-Guo: file_old_position  is not used ?

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,163 @@
...
+        lastSyncedOffset = oldLastSyncedOffset;
+    }
+
+    ByteBuffer createBuffer(CommitLog commitLog)
+    {
+        if (minimumAllowedAlign == 0)
+        {
+            try
+            {
+                minimumAllowedAlign = (int)Files.getFileStore(logFile.toPath()).getBlockSize();
```
Maxwell-Guo: why you cast the long value to int ?

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,163 @@
...
+        catch (IOException e)
+        {
+            throw new FSWriteError(e, getPath());
+        }
+    }
+
+    @Override
+    public long onDiskSize()
+    {
+        return lastWritten ;
```
Maxwell-Guo: Remove trailing spaces

src/java/org/apache/cassandra/config/Config.java
jacek-lewandowski: I meant we could simply validate the configuration either in `DatabaseDescriptor` or where the files are opened.

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
amitdpawar: Agree and will add a warn log for read case.

conf/cassandra.yaml
amitdpawar: > Given what @blambov mentioned and the test results - is there any reason to not immediately deprecate this setting in favour of hardcoding it to direct in the next release?

As per the earlier suggestion, I have emailed to dev-community to understand how to proceed. Link for the thread is [https://lists.apache.org/thread/t6v0p10737p0joob2vcsdt0r3g8zt94q](url). Some replies showed interest in Mmap as default to prevent any surprises. I want to share another point here. Direct-IO didn't work with Ramdisk while doing performance study. So, I feel it is better to have configuration option and default can be changed based on community feedback.

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
amitdpawar: I will remove it.

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
amitdpawar: It was needed to test Compressed and Encrypted to compare standard vs Direct-IO. It might be required in future but not now.

test/conf/cassandra.yaml
amitdpawar: Default kept as it is and Direct-IO testing is enabled in "CommitLogStressTest.java" by adding a new entry. This entry will test "PeriodicCommitLogStressTest.java", "GroupCommitLogTest.java" and "BatchCommitLogTest.java". 

Let me know if this didn't solve the testing.

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
Maxwell-Guo: So what about rename to isStandardModeEnable() ?

It seems that there is a little gap between standard and bufferIo, it may need to take some time for you to explain that .
But in my mind , the code should and could explain by itself if you rename to isStandardModeEnable.

src/java/org/apache/cassandra/config/Config.java
amitdpawar: Good idea. Shall I add direct_io to existing enum and change the relavant functions to verify the correct input?

test/conf/cassandra.yaml
Maxwell-Guo: commit log is mmap by default, so I don't think we need this , am I right ?

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
amitdpawar: I think this is required and right place to show disk access mode and replay warn log when Direct-IO is used. Is this OK ?

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
amitdpawar: > So what about rename to isStandardModeEnable() ?
> 
> It seems that there is a little gap between standard and bufferIo, it may need to take some time for you to explain that . But in my mind , the code should and could explain by itself if you rename to isStandardModeEnable.

Agree and self-explanation is good.

test/conf/cassandra.yaml
amitdpawar: > commit log is mmap by default, so I don't think we need this , am I right ?

I included for future usage if someone wants to change and test. Please suggest and will change as per the suggestion.

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
amitdpawar: I changed in flush method too.

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
amitdpawar: getBlockSize() returns long but variables used for flush operation are defined in int. These variables are "lastSyncedOffset", "startMarker" and "nextMarker".

test/conf/cassandra.yaml
Maxwell-Guo: `I included for future usage if someone wants to change and test.`
I recommend deleting it and leaving it for future revisionists to do the modifications

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
Maxwell-Guo: My suggestion is that you can describe these configurations in detail in the yaml file. For logs, we may need the information to be concise, and the information needs to reflect the key points. I think diskAccessMode（mmap，direct io，standard） is enough, see [here](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/config/DatabaseDescriptor.java#L525), these logs do not logging like what you have done. If you want to know more information, it can be explained in the description of the configuration file.

And I think we may logging here https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/config/DatabaseDescriptor.java#L525 as these codes have done, just tell the user we are using what kind of mode for diskAccess, for indexAccess, for commitLogAccess .

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
Maxwell-Guo: block size is always 4k or something value that is not so big. But what if there are some very extreme situations? Causes the value of int to overflow.

amitdpawar: A quick search on the internet didn't help to prove that it won't cross max int value. Most of the link suggest minimum required is page size (4K). Linux further reduces this to minimum 512 bytes and link for the same is [https://lwn.net/Articles/12032/](url). I think It is very unlikely that it can cross max int value.

Maxwell-Guo: of course， if the value reach Int.Max then the blocksize will be more than 2G.
What I have said is some extreme situations. what if some one set his own block size manually, then there may be a possibility, and as [CEP-36](https://mail.google.com/mail/u/0/#inbox/FMfcgzGtxKXMXDwqwZjcCbKGnQpfWklX)  described  If the distributed file system is connected in the future, the blocksize can be big .

src/java/org/apache/cassandra/config/Config.java
jacek-lewandowski: I think that would be clear - one enum, just validate whether we support the selected mode for the exact use case

conf/cassandra.yaml
blambov: The name of the option is good for me, but it is quite important to have a more precise description.

> Some replies showed interest in Mmap as default to prevent any surprises.

This is what I expected. Take a look at CASSANDRA-18753 as a potential way forward for making direct the default for new users.

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1150,6 +1151,13 @@ public enum DiskAccessMode
...
+    {
+        standard,
+        mmap,
+        direct_io,
```
blambov: `standard` and `mmap` are also I/O modes, I would drop the `_io` part to `direct`.

conf/cassandra.yaml
amitdpawar: Received another feedback yesterday regarding changing default to Direct-IO email and that makes it to 7 responses.

Summary:
1. mmap     :3/7
2. Direct-IO: 4/7

@blambov and @jacek-lewandowski your input might be helpful or please suggest what should be the default?

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
blambov: Okay, I see that I did not understand how Java NIO's direct mode works and we need to implement the paging/buffering ourselves, which is not as trivial.
One thing we can do for now is to only apply the direct mode if compression or encryption are not enabled, falling back to standard access otherwise (possibly with a message).

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1149,6 +1151,13 @@ public enum DiskAccessMode
         standard,
     }
 
+    public enum CommitLogDiskAccessMode
+    {
+        standard,
```
blambov: I see three problems with the `standard` option:
- there is no implementation of standard mode for uncompressed log (we probably can use [this simple one](https://github.com/datastax/cassandra/blob/ds-trunk/src/java/org/apache/cassandra/db/commitlog/UncompressedSegment.java) if we need to have it);
- the current code requires the user to change from `direct` to `standard` when enabling compression/encryption, which is not nice, and would be a deal breaker for `direct` being the default;
- we understand `mmap` to silently translate to `standard` for compressed/encrypted logs, which is legacy behaviour that we do not want to change, but which is at odds with the above.

The simplest way to solve this is to explain in the option's comment that it currently only applies to uncompressed/unencrypted logs and remove `standard` from this enum. But if we are planning also to implement direct writing of compressed/encrypted logs (valuable in its own right), it look like we should change the alternative to `mmap_or_standard` or maybe something like `legacy`.

Maxwell-Guo:
@amitdpawar  do you have time to continue this work ? I can help to finish if you have no time.

amitdpawar:
> @amitdpawar do you have time to continue this work ? I can help to finish if you have no time.

I was waiting for feedback and will update the PR with suggested changes.

conf/cassandra.yaml
blambov: My personal preference would be not only to change the default, but to scrap memory-mapped logs altogether. However,  for things like these we can't change the default unless everyone agrees or there's no expected downside. In this case a downside is the fact that a memory-mapped log might effectively persist data more often in the default periodic mode.

`mmap` should thus remain default, with a switch to `direct` as one of the items in the optimized configuration envisioned by CASSANDRA-18753.

amitdpawar:
@Maxwell-Guo changes are done as per the feedback. Only "mmap" and "direct" I/O modes are allowed in cassandra.yaml and this avoids confusion related to 'standard' mode. Please let me know for any new changes.

conf/cassandra.yaml
jacek-lewandowski: I agree with @blambov

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -511,6 +511,32 @@ else if (conf.commitlog_sync_period.toMilliseconds() != 0)
...
         }
 
+        // Only mmap & Direct-I/O access modes are supported. Standard access mode is preassumed if Compressed or Encrypted segment is used.
+        // This ensures default behavior is unchanged.
+        if  (conf.commitlog_disk_access_mode_for_write != Config.DiskAccessMode.mmap
+             && conf.commitlog_disk_access_mode_for_write != Config.DiskAccessMode.direct)
+        {
+            throw new ConfigurationException(String.format("Commitlog access mode '%s' is not supported. Supported modes are '%s' and '%s'", conf.commitlog_disk_access_mode_for_write,
+                                                                              Config.DiskAccessMode.mmap, Config.DiskAccessMode.direct));
+        } else {
```
Maxwell-Guo: code format: keep "else" alone in a line , like line 505

src/java/org/apache/cassandra/config/Config.java
jacek-lewandowski: I'm still at the position we should have one enum which cover all the needed cases and just validate that for certain uses.

Here in particular, it would be useful to add, as suggested by @blambov - `legacy` or assume undefined value would denote that. On the other hand, it would be useful to add `auto` for the new, optimized behavior. 

When access mode is specified explicitly, and it is invalid for certain use, for example - `direct` for compressed/encrypted commitlog, I would like to fail the startup with configuration error rather than do something implicitly. In other words, I wish Cassandra does not configure something implicitly unless we explicitly specify something that suggest default, automatic or legacy behavior.

That applies to both commitlog and sstable data files, and maybe would be useful to separate that configuration for read and write.

jacek-lewandowski:
One more idea from me is to put sstable disk access mode under sstable options and deprecate the current top-level `disk_access_mode`. That can be the default for now, but we can let it be overwritten in sstable options. In particular, there is an `sstable` of the `SSTableConfig` type. This is used to select the default format and define per-format reader and writer options.

It would be nice if, for a certain format, one could define in its options things like:
`data_read_access_mode`, `data_write_access_mode`, `index_read_access_mode`, `index_write_access_mode`. For the new TrieIndex format, that could even be separated for partition and row index. All of those options are optional and hidden by default. However, for the new TrieIndex format, they could be configured with the optimized values instead of legacy defaults (in particular, their implicit values could be `auto` instead of inheriting the value of deprecated `disk_access_mode` global option).

conf/cassandra.yaml
```diff
@@ -589,6 +589,15 @@ commitlog_segment_size: 32MiB
...
 #         -
 
+# CommitLog segments can use mmap (default) and direct (new) I/O modes to flush
+# the data to disk. Direct I/O is a new feature that minimizes cache effects by
+# using user space buffers. This helps in transferring data from/to disk at high
+# speed. Java enabled support for Direct-I/O feature from version 10 onwards.
+#   Refer: https://bugs.openjdk.org/browse/JDK-8164900
+# Direct-I/O is not enabled yet for compressed and encrypted based segments and
+# for replay (reading) commitlog files.
+commitlog_disk_access_mode_for_write: mmap
```
Maxwell-Guo: sorry @amitdpawar ,my fault, @blambov is +1 on the original name: commitlog_disk_access_mode, and I'm + 1 too. As when reading is support, we can make the access mode at commitlog level not the read/write operation level for commitlog .
but the description should tell the user that read is not support.

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+
+        int segmentSize = DatabaseDescriptor.getCommitLogSegmentSize();
+        original = ByteBuffer.allocateDirect(segmentSize + minimumAllowedAlign);
+
+        ByteBuffer alignedBuffer = original.alignedSlice(minimumAllowedAlign);
+        assert alignedBuffer.limit() >= segmentSize : String.format("Bytebuffer slicing failed to get required buffer size (required=%d,current size=%d", segmentSize, alignedBuffer.limit());
+
+        assert alignedBuffer.alignmentOffset(0, minimumAllowedAlign) == 0 : "Index 0 should be aligned to 4K page size";
```
Maxwell-Guo: 4k should change to minimumAllowedAlign length ? am I right ? the default may be 4k, But there may exist some other possibilities, right?

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,162 @@
...
+            {
+                filePosition = filePosition & ~(minimumAllowedAlign -1);
+                channel.position(filePosition);
+            }
+            duplicate.position(filePosition);
+
+            int flushSizeInBytes = nextMarker;
+
+            // Align last byte to end of 4K page.
+            if (flushSizeInBytes % minimumAllowedAlign !=0)
```
Maxwell-Guo: code format: != 0

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,162 @@
...
+                flushSizeInBytes = (flushSizeInBytes + minimumAllowedAlign) & ~(minimumAllowedAlign -1);
+
+            duplicate.limit(flushSizeInBytes);
+
+            channel.write(duplicate);
+
+            // Direct I/O always writes flushes in block size and writes more than the flush size.
+            // File size on disk will always multiple of page size and taking this into account
+            // helps testcases to pass.
+            if (flushSizeInBytes > lastWritten) {
```
Maxwell-Guo: code formt: { should start a new line

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1147,6 +1148,14 @@ public enum DiskAccessMode
         mmap,
         mmap_index_only,
         standard,
+        direct,
```
Maxwell-Guo: As you have add direct in DiskAccessMode, I think you should add some asserts or throw ConfigurationException [here ](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/config/DatabaseDescriptor.java#L515) if some one make a configuration for diskaccess mode to be direct.

amitdpawar:
> One more idea from me is to put sstable disk access mode under sstable options and deprecate the current top-level `disk_access_mode`. That can be the default for now, but we can let it be overwritten in sstable options. In particular, there is an `sstable` of the `SSTableConfig` type. This is used to select the default format and define per-format reader and writer options.
> 
> It would be nice if, for a certain format, one could define in its options things like: `data_read_access_mode`, `data_write_access_mode`, `index_read_access_mode`, `index_write_access_mode`. For the new TrieIndex format, that could even be separated for partition and row index. All of those options are optional and hidden by default. However, for the new TrieIndex format, they could be configured with the optimized values instead of legacy defaults (in particular, their implicit values could be `auto` instead of inheriting the value of deprecated `disk_access_mode` global option).

I think it is good to come up with such changes to have better control for read/write to/from disk. 

```
public static class SSTableConfig
{
    public String selected_format = BigFormat.NAME;
    public Map<String, Map<String, String>> format = new HashMap<>();
    DiskAccessMode data_read_access_mode = DiskAccessMode.auto;
    DiskAccessMode data_write_access_mode = DiskAccessMode.auto;
    DiskAccessMode index_read_access_mode = DiskAccessMode.auto;
    DiskAccessMode index_write_access_mode = DiskAccessMode.auto;
}

public final SSTableConfig sstable = new SSTableConfig();
```
OK with this change? I am not sure whether this will affect other thing. Please let me know.

amitdpawar:
PR is udpate. Please check now.

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1147,6 +1148,8 @@ public enum DiskAccessMode
         mmap,
         mmap_index_only,
         standard,
+        legacy,
```
Maxwell-Guo, line 1151: what does legacy in diskaccess mode stand for ? @jacek-lewandowski @amitdpawar ,seems lack of description at least.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -511,6 +511,41 @@ else if (conf.commitlog_sync_period.toMilliseconds() != 0)
             logger.debug("Syncing log with a period of {}", conf.commitlog_sync_period.toString());
         }
 
+        // Only mmap & Direct-I/O access modes are supported. Standard access mode is preassumed if Compressed or Encrypted segment is used.
+        // This ensures default behavior is unchanged.
+        if  (conf.commitlog_disk_access_mode != Config.DiskAccessMode.auto
+             && conf.commitlog_disk_access_mode != Config.DiskAccessMode.legacy
```
Maxwell-Guo, line 517: I'm a little confused about the logic of this code. What do you want to do here ?
if commitlog access mode is direct or mmap then exception throw?

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,163 @@
...
+        // lot and Syncer thread can actually do flush activity. Making this initial
+        // slow operation to be executed by Allocator thread here.
+        int oldLastSyncedOffset = lastSyncedOffset;
+        // 8 bytes are written above.
+        lastSyncedOffset = 8;
+        flush(0, lastSyncedOffset);
+        lastSyncedOffset = oldLastSyncedOffset;
+    }
+
+    ByteBuffer createBuffer(CommitLog commitLog)
```
Maxwell-Guo, line 54: Hello , a new question here, and that is what I want to know for a long time: how can we ensure  the correctness of this code, does it need to be tested， I mean the class of this Segment ?  @blambov , @jacek-lewandowski ，wdyt ？

src/java/org/apache/cassandra/config/Config.java
jacek-lewandowski: I consider `legacy` to: use the same defaults as in the previous release unless something was explicitly specified.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
jacek-lewandowski: out of curiosity, why standard must be used for writing compressed/encrypted? Doesn't it write data sequentially anyway?

jacek-lewandowski: The condition should rather unconditionally prevent `mmap_index_only`

Other options may be valid depending on the context - say whether we have encryption/compression enabled or not, where valid does not necessarily mean efficient - that is another story;

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,163 @@
...
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.io.FSWriteError;
+import org.apache.cassandra.io.util.FileUtils;
+
+/*
+ * Direct-IO segment. Allocates ByteBuffer using ByteBuffer.allocateDirect and align
+ * ByteBuffer.position, ByteBuffer.limit and FileChannel.position to page size (4K).
+ * Java-11 forces minimum page size to be written to disk with Direct-IO.
+ */
+public class DirectIOSegment extends CommitLogSegment
```
jacek-lewandowski, line 32: Why it does not inherit from `FileDirectSegment` ?

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
jacek-lewandowski: I was also wondering why don't we use `FileHandle` for commit log?

amitdpawar:
@blambov, @Maxwell-Guo, @jacek-lewandowski any more suggestions?

Maxwell-Guo:
> @blambov, @Maxwell-Guo, @jacek-lewandowski any more suggestions?

I think I have stated most of what I wanted to express. looking forward for your feedback of the code modification.

amitdpawar:
> > @blambov, @Maxwell-Guo, @jacek-lewandowski any more suggestions?
> 
> I think I have stated most of what I wanted to express. looking forward for your feedback of the code modification.

Thanks @Maxwell-Guo. I will update the PR with changes discussed related to first flush for each commitlog segment.

amitdpawar:
Summary of changes done to this PR.
1. Moved file initialization from DirectIOSegment constructor to CommitLogSegment.writeLogHeader function. This is required from correctness and performance point of view for each file.
2. Enabled "SimpleCachedBufferPool" pool to re-use the buffers as suggested by "Ariel Weisberg" during initial review of this ticket.  Need to identify Java minimum alignment to use this feature during startup and need to pass this value to same class to allocate the additional bytes to slice the buffer.

Please review.

amitdpawar:
@Maxwell-Guo, when does these PR gets tested completely? during review or after the upstreaming?

jacek-lewandowski:
@amitdpawar we will need to run CircleCI tests before merging

jacek-lewandowski:
I have a timeslot today so I can devote it to reviewing and running tests for you

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2649,6 +2688,24 @@ public static void setCommitLogSegmentSize(int sizeMebibytes)
...
 
+    /**
+     * Return commitlog disk access mode.
+     */
+    public static Config.DiskAccessMode getCommitLogDiskAccessMode()
+    {
+        return conf.commitlog_disk_access_mode;
+    }
+
+    public static boolean isCommitLogUsingDirectIO()
```
jacek-lewandowski: Since we use enums, I think this method is redundant

amitdpawar:
> I have a timeslot today so I can devote it to reviewing and running tests for you

Thank you @jacek-lewandowski for your time and will wait for your next feedback.

conf/cassandra.yaml
```diff
@@ -589,6 +589,16 @@ commitlog_segment_size: 32MiB
 #     parameters:
 #         -
 
+# CommitLog segments can use legacy (default), direct (new) and 'auto' (optimized
```
jacek-lewandowski, line 592: The comment looks good for release notes however I'd prefer to rephrase it for yaml doc. I think that the essential information which should be there are:
- this property applies to writing the commit log
- the allowed values are: ... (with meaning)
- the default value is: ...

I think we should include the current explanation in the NEWS.txt

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -511,6 +511,41 @@ else if (conf.commitlog_sync_period.toMilliseconds() != 0)
...
+                // This is necessary for other functions to get the new mode determined through auto.
+                conf.commitlog_disk_access_mode = current_disk_access_mode;
+                logger.info("DiskAccessMode 'auto' determined to be '{}' for Commitlog disk", current_disk_access_mode);
+            }
+
+            if (getCommitLogCompression() != null || (getEncryptionContext() != null && getEncryptionContext().isEnabled()))
+            {
+                if (current_disk_access_mode == Config.DiskAccessMode.direct)
+                {
+                    // Compressed and Encrypted segments are not yet supported with Direct-I/O feature.
```
jacek-lewandowski, line 541: It looks like we will get there when `auto` is set and the commit log is compressed/encrypted - shouldn't `auto` mean no configuration errors?

jacek-lewandowski: tbh, I think the check should be as follows:

Allowed options when no compression/encryption is enabled:
- `legacy`, `auto`, `mmap`, `direct`, (I don't see why standard should not be allowed, but let's leave it forbidden for now)

Allowed options when compression/encryption is enabled:
- `legacy`, `auto`, `standard`, (at least for now, in the future we can try to use `mmap` if needed, support for reading compressed files with `mmap` has been implemented as part of CASSANDRA-17056, at least for sstable files)

`auto` is resolved to:
- `direct` if ssd and no compression
- `legacy` otherwise

`legacy` is resolved to:
- `mmap` if no compression (right?)
- `standard` otherwise

conf/cassandra.yaml
```diff
@@ -589,6 +589,16 @@ commitlog_segment_size: 32MiB
...
+# configuration) I/O modes to flush the data to disk. Direct I/O is a new feature
+# that minimizes cache effects by using user space buffers. This helps in
+# transferring data from/to disk at high speed. Java enabled support for
+# Direct-I/O feature from version 10 # onwards.
+# Refer: https://bugs.openjdk.org/browse/JDK-8164900
+# The same is not enabled yet for compressed and encrypted based segments and
+# for replay (reading) commitlog files.
+# commitlog_disk_access_mode: legacy
```
jacek-lewandowski, line 600: nit, please at least state that it is for writing, like `commitlog_write_disk_acccess_mode`; I'd really prefer to have hierarchical structure as it is the direction we are slowing moving towards:

```
commitlog:
  write_disk_access_policy: legacy
```

test/long/org/apache/cassandra/db/commitlog/CommitLogStressTest.java
```diff
@@ -190,10 +194,11 @@ private void testLog() throws IOException, InterruptedException
     }
 
     private void testLog(CommitLog commitLog) throws IOException, InterruptedException {
-        System.out.format("\nTesting commit log size %.0fmb, compressor: %s, encryption enabled: %b, sync %s%s%s\n",
+        System.out.format("\nTesting commit log size %.0fmb, compressor: %s, encryption enabled: %b, direct I/O enabled: %b, sync %s%s%s\n",
```
jacek-lewandowski: maybe mention the disk policy explicitly

src/java/org/apache/cassandra/db/commitlog/CommitLog.java
```diff
@@ -624,11 +632,13 @@ public static final class Configuration
...
         private EncryptionContext encryptionContext;
 
-        public Configuration(ParameterizedClass compressorClass, EncryptionContext encryptionContext)
+        public Configuration(ParameterizedClass compressorClass, EncryptionContext encryptionContext,
+                             Config.DiskAccessMode diskAccessMode)
         {
             this.compressorClass = compressorClass;
             this.compressor = compressorClass != null ? CompressionParams.createCompressor(compressorClass) : null;
             this.encryptionContext = encryptionContext;
+            this.diskAccessMode = diskAccessMode;
```
jacek-lewandowski, line 641: I think we should have a precondition to validate the disk access mode here as well

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
```diff
@@ -714,6 +729,29 @@ public CDCState setCDCState(CDCState newState)
         }
     }
 
+    /**
+     * Direct requires minimum alignement and block size to be used for read and writes. Identify it now.
+     */
+    static public int getDirectIOMinimumAlignement(String storageDirectory, String fileName)
```
jacek-lewandowski, line 729: `fileName` is unused.

Also I think you should rather do `Files.createTempFile(Paths.get(storageDirectory), ...)` and try-with-resources

src/java/org/apache/cassandra/db/commitlog/AbstractCommitLogSegmentManager.java
```diff
@@ -111,12 +111,19 @@ void start()
...
-                                ? BufferType.ON_HEAP
-                                : commitLog.configuration.getCompressor().preferredBufferType();
+        BufferType bufferType = commitLog.configuration.isDirectIOEnabled()
+                                ? BufferType.OFF_HEAP
+                                  : commitLog.configuration.useEncryption() || !commitLog.configuration.useCompression()
+                                    ? BufferType.ON_HEAP
+                                    : commitLog.configuration.getCompressor().preferredBufferType();
+
+        // Identify minimum block size used for Direct-I/O. allocateDirect does not return from aligned space.
+        // To avoid write errors additional block size is requested and next aligned it minimum block size.
```
jacek-lewandowski: Could you rephrase it? I found it difficult to understand - maybe something like:

_The direct buffer must be aligned with the file system block size. We cannot enforce that during allocation, but we can get an aligned slice from the allocated buffer. The buffer must be oversized by the alignment unit to make it possible._

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
```diff
@@ -197,6 +206,12 @@ void writeLogHeader()
         lastSyncedOffset = lastMarkerOffset = buffer.position();
         allocatePosition.set(lastSyncedOffset + SYNC_MARKER_SIZE);
         headerWritten = true;
+
+        // Testing shows writing initial bytes takes some time for Direct I/O. During peak load,
+        // it is better to make "COMMIT-LOG-ALLOCATOR" thread to write these few bytes of each
+        // file and this helps syncer thread to speedup the flush activity.
+        if (CommitLog.instance.configuration.isDirectIOEnabled())
```
jacek-lewandowski: `writeLogHeader` methods should be rather overridden in `DirectIOSegment` and call this statement there

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,154 @@
...
+
+        // write previous sync marker to point to next sync marker
+        // we don't chain the crcs here to ensure this method is idempotent if it fails
+        writeSyncMarker(id, buffer, startMarker, startMarker, nextMarker);
+    }
+
+    @Override
+    protected void flush(int startMarker, int nextMarker)
+    {
+
```
jacek-lewandowski: nit: blank line to be removed

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.cassandra.io.FSWriteError;
+
+/*
+ * Direct-IO segment. Allocates ByteBuffer using ByteBuffer.allocateDirect and align
+ * ByteBuffer.position, ByteBuffer.limit and FileChannel.position to page size (4K).
+ * Java-11 forces minimum page size to be written to disk with Direct-IO.
+ */
+public class DirectIOSegment extends CommitLogSegment
+{
+    ByteBuffer original;
```
jacek-lewandowski: should be private I think

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,154 @@
...
+        int firstSync = buffer.position();
+        buffer.putInt(firstSync + 0, 0);
+        buffer.putInt(firstSync + 4, 0);
+    }
+
+    ByteBuffer createBuffer(CommitLog commitLog)
+    {
+        int segmentSize = DatabaseDescriptor.getCommitLogSegmentSize();
+
+        original =  manager.getBufferPool().createBuffer();
```
jacek-lewandowski, line 58: assert `original == null`

src/java/org/apache/cassandra/db/commitlog/CommitLogSegment.java
```diff
@@ -83,6 +85,9 @@ public enum CDCState
         replayLimitId = idBase = Math.max(currentTimeMillis(), maxId + 1);
     }
 
+    // Minimum alignement and block size required by direct I/O feature.
+    public static int minimumDirectIOAlignement = 0;
```
jacek-lewandowski, line 89: this should either land in `DirectIOSegment` and should not be static or it should be computed as part of `DatabaseDescriptor` initialization and stored there; Also, I'd rename it to `commitLogFileBlockSize` (or something like that)

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    protected void flush(int startMarker, int nextMarker)
+    {
+
+        try
+        {
+            // lastSyncedOffset is synced to disk. Align lastSyncedOffset to start of 4K page
+            // and nextMarker to end of 4K page to avoid write errors.
```
jacek-lewandowski: you mean pages of file block size rather than 4K right?

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,154 @@
...
+            int filePosition = lastSyncedOffset;
+            ByteBuffer duplicate = buffer.duplicate();
+
+            // Aligned file position if not aligned to start of 4K page.
+            if (filePosition % minimumDirectIOAlignement != 0)
+            {
+                filePosition = filePosition & ~(minimumDirectIOAlignement -1);
```
jacek-lewandowski: `~(minimumDirectIOAlignement -1)` - if you stored the block size in a final field or `DatabaseDescriptor` you could store a mask in final field as well

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,154 @@
...
+                channel.position(filePosition);
+            }
+            duplicate.position(filePosition);
+
+            int flushSizeInBytes = nextMarker;
```
jacek-lewandowski: name `flushSizeInBytes` is misleading, it is the `flushLimit` right?

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,154 @@
...
+
+    @Override
+    protected void flush(int startMarker, int nextMarker)
+    {
+
+        try
+        {
+            // lastSyncedOffset is synced to disk. Align lastSyncedOffset to start of 4K page
+            // and nextMarker to end of 4K page to avoid write errors.
+            int filePosition = lastSyncedOffset;
```
jacek-lewandowski: would be good to named it `flushPosition`

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,154 @@
...
+                flushSizeInBytes = (flushSizeInBytes + minimumDirectIOAlignement) & ~(minimumDirectIOAlignement -1);
+
+            duplicate.limit(flushSizeInBytes);
+
+            channel.write(duplicate);
+
+            // Direct I/O always writes flushes in block size and writes more than the flush size.
+            // File size on disk will always multiple of page size and taking this into account
+            // helps testcases to pass.
+            if (flushSizeInBytes > lastWritten)
```
jacek-lewandowski: I'm not sure about this. It looks like `flushSizeInBytes` is always > `lastWritten`, isn't it?

src/java/org/apache/cassandra/db/commitlog/DirectIOSegment.java
```diff
@@ -0,0 +1,154 @@
...
+            {
+                manager.addSize(flushSizeInBytes - lastWritten);
```
jacek-lewandowski: Why don't call `manager.addSize(segment size)` on buffer creation just like it is done for mapped segment? We know the size upfront, don't we?

jacek-lewandowski (changes requested):
I've left some comments and questions. Other than that, just as it was already mentioned, we need a test for the `DirectIOSegment` - it could be a unit test with mocking most of the stuff. We really need to have a coverage of that alignment related stuff. It looks correct as I was reviewing, but additional test is required. I suggest implement it using quick theories so that we catch the edge cases.

Additionally, it would be great to have some test which can verify different configuration paths.

We are open to help you with testing and other stuff, feel free to ask.

I'm going to modify the default configuration to use the direct IO for commit log and run the tests then so that we can be sure that new functionality is in use.

amitdpawar:
> I've left some comments and questions. Other than that, just as it was already mentioned, we need a test for the `DirectIOSegment` - it could be a unit test with mocking most of the stuff. We really need to have a coverage of that alignment related stuff. It looks correct as I was reviewing, but additional test is required. I suggest implement it using quick theories so that we catch the edge cases.
> 
> Additionally, it would be great to have some test which can verify different configuration paths.
> 
> We are open to help you with testing and other stuff, feel free to ask.
> 
> I'm going to modify the default configuration to use the direct IO for commit log and run the tests then so that we can be sure that new functionality is in use.

I didn't realize and it may take more time to do all these changes. Definitely need your help. I request you to do all the necessary changes required to upstream this patch. If anything is required from my side, then please let me know.

jacek-lewandowski:
@amitdpawar - I'll help you with testing

jacek-lewandowski:
@amitdpawar - are you going to fix those remarks which you marked as resolved or you'd like me to help with it as well?

Maxwell-Guo:
> @amitdpawar - are you going to fix those remarks which you marked as resolved or you'd like me to help with it as well?

@amitdpawar I can also do some help to finish this task  @jacek-lewandowski , just ping me as you want .

jacek-lewandowski:
Thanks @Maxwell-Guo - but first, let me know if you agree with my comments - if we are going to finish this, our changes should be reviewed as well, so I'd like hear your feedback on that :)

amitdpawar:
> @amitdpawar - are you going to fix those remarks which you marked as resolved or you'd like me to help with it as well?

I have fixed the marked one's as resolved but was thinking how to do common code related. It will be great if you can help for non-marked cases as these are related to common code and might be required for future case. So, feeling taking help is ideal.

jacek-lewandowski:
@amitdpawar ok, but please submit what you have so far so that we work on your most actual version

amitdpawar:
@jacek-lewandowski, I have pushed the changes and please check.

amitdpawar:
> > @amitdpawar - are you going to fix those remarks which you marked as resolved or you'd like me to help with it as well?
> 
> @amitdpawar I can also do some help to finish this task @jacek-lewandowski , just ping me as you want .

Thanks @Maxwell-Guo. I will seek your help after @jacek-lewandowski review.

jacek-lewandowski:
@amitdpawar could you answer my question https://github.com/apache/cassandra/pull/2777#discussion_r1389237529 ?

I'll create a PR against your branch with changes I meant for the rest of the comments

amitdpawar:
> @amitdpawar could you answer my question [#2777 (comment)](https://github.com/apache/cassandra/pull/2777#discussion_r1389237529) ?
> 
> I'll create a PR against your branch with changes I meant for the rest of the comments

@jacek-lewandowski, I am OK. Otherwise, if you share those changes as a file then I can apply and push your change. Please do whatever is easy for you.

jacek-lewandowski:
@amitdpawar here is the commit with the changes I propose - https://github.com/jacek-lewandowski/cassandra/commit/1c51c0db1fce3979e89b8d22c64119f32258d2d5

Feel free to check it, I'll work on tests now (@Maxwell-Guo, any thoughts?)

Maxwell-Guo:
> @amitdpawar here is the commit with the changes I propose - [jacek-lewandowski@1c51c0d](https://github.com/jacek-lewandowski/cassandra/commit/1c51c0db1fce3979e89b8d22c64119f32258d2d5)
> 
> Feel free to check it, I'll work on tests now (@Maxwell-Guo, any thoughts?)

Thank @jacek-lewandowski , I just left some comments , I feel that most of your commits are great. There are only a few points where I may still be a little confused. For example, I prefer the hierarchical structure you mentioned, if this is not accept this time , I think commitlog_disk_access_mode may be better , As  direct io for log read when commitlog doing replay may finally be supported in the future , and different configuration for different disk access mode seems a bit redundant. As they seems are two big kind of configuration but hierarchical structure seem is only one .

As for correctness testing of DirectIOSegment, what about create a  log file ,and just write data , and using reader to fetch data and verify the correctness ?
Also I think we can add some stress test .
Let me know if you if i can do something for this , I think this is a great feature for c* .

src/java/org/apache/cassandra/db/commitlog/AbstractCommitLogSegmentManager.java
```diff
@@ -111,12 +111,20 @@ void start()
     {
         // For encrypted segments we want to keep the compression buffers on-heap as we need those bytes for encryption,
         // and we want to avoid copying from off-heap (compression buffer) to on-heap encryption APIs
-        BufferType bufferType = commitLog.configuration.useEncryption() || !commitLog.configuration.useCompression()
-                                ? BufferType.ON_HEAP
-                                : commitLog.configuration.getCompressor().preferredBufferType();
+        BufferType bufferType = commitLog.configuration.isDirectIOEnabled()
```
smiklosovic, line 114: wow to read this is quite a challenge! Could you rewrite this somehow with if / else?

jacek-lewandowski:
@amitdpawar @Maxwell-Guo I've created a separate PR to make it easier for reviewing
https://github.com/apache/cassandra/pull/2894

While implementing a unit test I realized we need more refactoring to make it a bit more testable. I hope you will find my changes useful and acceptable - please have a look.

@Maxwell-Guo - I haven't overwritten the branch, so it has only that one additional commit compared to what you have already reviewed

@smiklosovic this is already refactored

jacek-lewandowski:
I've pushed one more commit with DatabaseDescriptor test case;

jacek-lewandowski:
The mailing list decided that by default we should use a legacy disk access mode (the same as it was selected in Cassandra 4 and prior). What do you think about tying it to `storage_compatibility_mode`? I mean, if `storage_compatibility_mode` is 4, then we use `legacy`, otherwise we use `auto`

cc @michaelsembwever

blambov:
>  What do you think about tying it to `storage_compatibility_mode`? I mean, if `storage_compatibility_mode` is 4, then we use `legacy`, otherwise we use `auto`

Unless we have a clear understanding that SCM is not just compatibility but also default switching, I would not do that. That decision was something I was thinking of in CASSANDRA-18753, but it didn't appear to be something other people wanted, and there is no longer time to implement it.

Direct IO for the commit log will be part of the "latest" configuration provided by CASSANDRA-18753.

PR #2783, opened 2023-10-06
CASSANDRA-18913: Gossip NPE due to shutdown event corrupting empty statuses

dcapwell:


src/java/org/apache/cassandra/gms/Gossiper.java
```diff
@@ -600,11 +600,12 @@ public void convict(InetAddressAndPort endpoint, double phi)
      * This method is used to mark a node as shutdown; that is it gracefully exited on its own and told us about it
      * @param endpoint endpoint that has shut itself down
      */
+    @Deprecated // can remove once 4.x is not supported
```
Claudenw: Should this use the "since" property of the Deprecated annotation? 

i.e. @Deprecated(since="5.0")

src/java/org/apache/cassandra/gms/Gossiper.java
```diff
@@ -600,11 +600,12 @@ public void convict(InetAddressAndPort endpoint, double phi)
...
     protected void markAsShutdown(InetAddressAndPort endpoint)
     {
         checkProperThreadForStateMutation();
         EndpointState epState = endpointStateMap.get(endpoint);
-        if (epState == null)
+        if (epState == null || epState.isStateEmpty())
```
driftx: Is this enough to prevent the issue without sending the endpoint state?  If so maybe we can do this in lower branches?

src/java/org/apache/cassandra/gms/Gossiper.java
dcapwell: added locally; using 5.0.0 assuming this should be back ported there

src/java/org/apache/cassandra/gms/Gossiper.java
dcapwell: this blocks the known corruption issue, so back porting this to 4.x should be safe and fix this specific issue, but won't fix the other issue we talked about, that host replacement is missing TOKENS.  Given the jira is only about the NPE that feels like a good solution to fix the most branches without any protocol level change

pmcfadin:
Closing as [CASSANDRA-18913](https://issues.apache.org/jira/browse/CASSANDRA-18913) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #2784, opened 2023-10-09
Cassandra 18221 - Accord configuration

jacek-lewandowski:


src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1129,6 +1123,8 @@ public enum PaxosOnLinearizabilityViolation
      */
     public ParameterizedClass default_compaction = null;
 
+    public volatile AccordSpec accord = new AccordSpec();
```
dcapwell: don't need `volatile` here, `final` is better... we only mutate the leaf fields, not the complex fields

conf/cassandra.yaml
```diff
@@ -2113,3 +2110,15 @@ accord_transactions_enabled: false
 #   a stable cluster. If a node from a previous version was started by accident we won't any longer toggle behaviors as when UPGRADING.
 #
 storage_compatibility_mode: NONE
+
+accord:
```
dcapwell: need to comment this out; non of the fields are defined outside comments

conf/cassandra.yaml
```diff
@@ -2113,3 +2110,15 @@ accord_transactions_enabled: false
...
+  # Enables the execution of Accord (multi-key) transactions on this node.
+  # accord_transactions_enabled: false
```
dcapwell: think this is just `enabled`

dcapwell (changes requested):

conf/cassandra.yaml
```diff
@@ -2113,3 +2110,16 @@ accord_transactions_enabled: false
 #   a stable cluster. If a node from a previous version was started by accident we won't any longer toggle behaviors as when UPGRADING.
 #
 storage_compatibility_mode: NONE
+
+# accord:
+# Enables the execution of Accord (multi-key) transactions on this node.
```
dcapwell: spacing is off, and should we add a `#` in front so users can block uncomment and have this still work?  We don't really have many examples of nested properties with comments inside the nesting... so this could be the first time this could cause confusion I think.

src/java/org/apache/cassandra/config/AccordSpec.java
```diff
@@ -0,0 +1,30 @@
...
+
+public class AccordSpec
+{
+    public volatile boolean enabled = false;
+
+    public volatile String journal_directory;
+
+    public volatile OptionaldPositiveInt shard_count = OptionaldPositiveInt.UNDEFINED;
+
+    public volatile DurationSpec.IntSecondsBound progress_log_schedule_delay = new DurationSpec.IntSecondsBound(1);
```
dcapwell, line 29: unless we plan to have JMX/VTable overrides of these properties, we should not add `volatile`... For example `enabled` can not change during the life of the JVM as we only add the accord tables on boot...  same for most of these configs, they are init only

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1129,6 +1123,8 @@ public enum PaxosOnLinearizabilityViolation
      */
     public ParameterizedClass default_compaction = null;
 
+    public AccordSpec accord = new AccordSpec();
```
dcapwell: `final`

test/conf/cassandra-mtls.yaml
```diff
@@ -87,3 +86,5 @@ authenticator:
   class_name : org.apache.cassandra.auth.MutualTlsAuthenticator
   parameters :
     validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
+accord:
+  journal_directory: build/test/cassandra/accord_journal
```
dcapwell, line 90: FYI, you can also do

```
accord.journal_directory: build/test/cassandra/accord_journal
```

No need to change, was mostly saying as it might have been faster for you when you first changed these, so thought to mention

dcapwell (approved):
mostly small issues that can be fixed on merge, so +1

conf/cassandra.yaml
jacek-lewandowski: there is one: `repair` section but it is broken as well. 
will fix the accord section

src/java/org/apache/cassandra/config/AccordSpec.java
jacek-lewandowski: I was primarily thinking about tests

test/conf/cassandra-mtls.yaml
jacek-lewandowski: yeah, but now we have all the Accord related stuff grouped together

PR #2786, opened 2023-10-09
Make sstable format configurable per table for CASSANDRA-18534

Maxwell-Guo:
```
Make sstable format configurable per table for CASSANDRA-18534

patch by <Maxwell Guo>; reviewed by <Reviewers> for CASSANDRA-18534

```

### DIFF
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -363,10 +363,11 @@ public class Config
     public static class SSTableConfig
     {
         public String selected_format = BigFormat.NAME;
-        public Map<String, Map<String, String>> format = new HashMap<>();
+        public Map<String, Map<String, String>> sstable_format_options = new HashMap<>();
     }
 
     public final SSTableConfig sstable = new SSTableConfig();
+
 
     /**
      * The directory to use for storing the system keyspaces data.
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -33,6 +33,7 @@ import java.util.Collection;
 import java.util.Comparator;
 import java.util.Enumeration;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -53,6 +54,8 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.MapDifference;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.common.primitives.Ints;
 import com.google.common.primitives.Longs;
@@ -99,6 +102,8 @@ import org.apache.cassandra.locator.IEndpointSnitch;
 import org.apache.cassandra.locator.InetAddressAndPort;
 import org.apache.cassandra.locator.Replica;
 import org.apache.cassandra.locator.SeedProvider;
+import org.apache.cassandra.schema.SSTableFormatParams;
+import org.apache.cassandra.schema.SSTableFormatParams.Option;
 import org.apache.cassandra.security.AbstractCryptoProvider;
 import org.apache.cassandra.security.EncryptionContext;
 import org.apache.cassandra.security.JREProvider;
@@ -1441,14 +1446,62 @@ public class DatabaseDescriptor
 
     private static ImmutableMap<String, Supplier<SSTableFormat<?, ?>>> validateAndMatchSSTableFormatOptions(Iterable<SSTableFormat.Factory> factories, Map<String, Map<String, String>> options)
     {
+        ImmutableMap.Builder<String, Supplier<SSTableFormat<?, ?>>> providersBuilderLeft = ImmutableMap.builder();
+        ImmutableMap.Builder<String, Supplier<SSTableFormat<?, ?>>> providersBuilderRight = ImmutableMap.builder();
         ImmutableMap.Builder<String, Supplier<SSTableFormat<?, ?>>> providersBuilder = ImmutableMap.builder();
         if (options == null)
             options = ImmutableMap.of();
+
+        // add default sstable format options for bti and big
         for (SSTableFormat.Factory factory : factories)
         {
-            Map<String, String> formatOptions = options.getOrDefault(factory.name(), ImmutableMap.of());
-            providersBuilder.put(factory.name(), () -> factory.getInstance(ImmutableMap.copyOf(formatOptions)));
+            providersBuilderLeft.put(factory.name(), () -> factory.getInstance(SSTableFormatParams.DEFAULT_FORMAT_MAP.apply(factory.name())));
         }
+
+        for (String key : options.keySet())
+        {
+            Map<String, String> formatOptions = Maps.newLinkedHashMap();
+            Map<String, String> map = options.get(key);
+            Iterator<String> iterator = map.keySet().iterator();
+            // as the map's value parsed by yaml may be numeric type, but sstable format factory need string
+            // but what about change the type of sstable format factory to a map that value is Object?
+            while (iterator.hasNext()){
+                String k = iterator.next();
+                String v = String.valueOf(map.get(k));
+                formatOptions.put(k, v);
+            }
+            // TODO add validation for yaml configuration number that must meet all the sstable param options ?
+
+            String type = SSTableFormatParams.validateAndParseType(key);
+            if (StringUtils.isBlank(type))
+                throw new ConfigurationException("SSTable format type name is empty.");
+
+            boolean illegalFormat = true;
+            // add user defined sstable format type
+            for (SSTableFormat.Factory factory : factories)
+            {
+                if(factory.name().equalsIgnoreCase(type))
+                {
+                     formatOptions.putIfAbsent(Option.TYPE.getName(), key);
+                    // add option map to AbstractSSTableFormat in order to use it in the future
+                    providersBuilderRight.put(key, () -> factory.getInstance(formatOptions));
+                    illegalFormat = false;
+                }
+            }
+            if (illegalFormat)
+                throw new ConfigurationException("Configuration contains options of unknown sstable formats " + key );
+        }
+
+        ImmutableMap<String, Supplier<SSTableFormat<?, ?>>> providersLeft = providersBuilderLeft.build();
+        ImmutableMap<String, Supplier<SSTableFormat<?, ?>>> providersRight = providersBuilderRight.build();
+        MapDifference<String, Supplier<SSTableFormat<?, ?>>> difference = Maps.difference(providersLeft, providersRight);
+        providersBuilder.putAll(difference.entriesInCommon()).putAll(difference.entriesOnlyOnRight()).putAll(difference.entriesOnlyOnLeft());
+        ImmutableMap.Builder<String, Supplier<SSTableFormat<?, ?>>> differentMapProvider = ImmutableMap.builder();
+        difference.entriesDiffering().forEach((key, value) -> {
+            // if user defined the same name with the default factory name ,then we should use the user defined
+            differentMapProvider.put(key, value.rightValue());
+        });
+        providersBuilder.putAll(differentMapProvider.build());
         ImmutableMap<String, Supplier<SSTableFormat<?, ?>>> providers = providersBuilder.build();
         if (options != null)
         {
@@ -1488,7 +1541,7 @@ public class DatabaseDescriptor
             return;
 
         validateSSTableFormatFactories(factories);
-        ImmutableMap<String, Supplier<SSTableFormat<?, ?>>> providers = validateAndMatchSSTableFormatOptions(factories, sstableFormatsConfig.format);
+        ImmutableMap<String, Supplier<SSTableFormat<?, ?>>> providers = validateAndMatchSSTableFormatOptions(factories, sstableFormatsConfig.sstable_format_options);
 
         ImmutableMap.Builder<String, SSTableFormat<?, ?>> sstableFormatsBuilder = ImmutableMap.builder();
         providers.forEach((name, provider) -> {
@@ -4849,6 +4902,13 @@ public class DatabaseDescriptor
         sstableFormats = null;
         selectedSSTableFormat = null;
         applySSTableFormats(factories, config);
+    }
+
+    public static Map<String, Map<String, String>>  getSSTableFormatOptions()
+    {
+        if (conf == null || conf.sstable == null)
+            return null;
+        return conf.sstable.sstable_format_options;
     }
 
     public static ImmutableMap<String, SSTableFormat<?, ?>> getSSTableFormats()
--- a/src/java/org/apache/cassandra/cql3/statements/schema/TableAttributes.java
+++ b/src/java/org/apache/cassandra/cql3/statements/schema/TableAttributes.java
@@ -25,10 +25,12 @@ import com.google.common.collect.Sets;
 
 import org.apache.cassandra.cql3.statements.PropertyDefinitions;
 import org.apache.cassandra.exceptions.ConfigurationException;
+import org.apache.cassandra.exceptions.SyntaxException;
 import org.apache.cassandra.schema.CachingParams;
 import org.apache.cassandra.schema.CompactionParams;
 import org.apache.cassandra.schema.CompressionParams;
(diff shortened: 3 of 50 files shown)


src/java/org/apache/cassandra/cql3/statements/schema/TableAttributes.java
```diff
@@ -185,6 +208,26 @@ private Map<String, String> getMap(Option option)
         return value;
     }
 
+    private boolean hasStringValue(Option option)
+    {
+        Object val = properties.get(option.toString());
+        if (val != null && val instanceof String)
```
smiklosovic: @Maxwell-Guo why do you use if / else? the result of `val != null && val instanceof String` is a boolean as such, `hasMapValue` method same case

Maxwell-Guo:
as @blambov have described in CASSANDRA-18534 , we may support 
`CREATE TABLE ... WITH sstable_format = "bti-fast";`
and 
`CREATE TABLE ... WITH sstable_format = { "type" : "bti-fast", "bloom_filter_fp_chance": 0.001};`
So I just add these methods to see if the options for stable_format is a string or a map, and can correspondingly, we can parse out the corresponding object. 
For other methods , we may get a exception .

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -363,11 +363,12 @@ public MemtableOptions()
     public static class SSTableConfig
     {
         public String selected_format = BigFormat.NAME;
-        public Map<String, Map<String, String>> format = new HashMap<>();
+        public Map<String, Map<String, String>> sstable_format_options = new HashMap<>();
```
blambov, line 366: This is already under `sstable`, rename to `format_options`.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1441,14 +1446,62 @@ private static void validateSSTableFormatFactories(Iterable<SSTableFormat.Factor
...
         ImmutableMap.Builder<String, Supplier<SSTableFormat<?, ?>>> providersBuilder = ImmutableMap.builder();
         if (options == null)
             options = ImmutableMap.of();
+
+        // add default sstable format options for bti and big
         for (SSTableFormat.Factory factory : factories)
         {
-            Map<String, String> formatOptions = options.getOrDefault(factory.name(), ImmutableMap.of());
-            providersBuilder.put(factory.name(), () -> factory.getInstance(ImmutableMap.copyOf(formatOptions)));
+            providersBuilderLeft.put(factory.name(), () -> factory.getInstance(SSTableFormatParams.DEFAULT_FORMAT_MAP.apply(factory.name())));
```
blambov, line 1458: Why not do this after the custom-type construction, checking if `format.name()` has already been added?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1441,14 +1446,62 @@ private static void validateSSTableFormatFactories(Iterable<SSTableFormat.Factor
...
         }
+
+        for (String key : options.keySet())
+        {
+            Map<String, String> formatOptions = Maps.newLinkedHashMap();
+            Map<String, String> map = options.get(key);
```
blambov, line 1464: Iterate on `options.entrySet()` so that you can get the map together with the key without the work of finding the entry in the map.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1441,14 +1446,62 @@ private static void validateSSTableFormatFactories(Iterable<SSTableFormat.Factor
...
+            Iterator<String> iterator = map.keySet().iterator();
+            // as the map's value parsed by yaml may be numeric type, but sstable format factory need string
+            // but what about change the type of sstable format factory to a map that value is Object?
+            while (iterator.hasNext()){
+                String k = iterator.next();
+                String v = String.valueOf(map.get(k));
+                formatOptions.put(k, v);
+            }
+            // TODO add validation for yaml configuration number that must meet all the sstable param options ?
```
blambov, line 1473: Doesn't `factory.getInstance(formatOptions)` below do that?

src/java/org/apache/cassandra/db/memtable/Flushing.java
```diff
@@ -104,6 +104,7 @@ static FlushRunnable flushRunnable(ColumnFamilyStore cfs,
                                        Directories.DataDirectory flushLocation)
     {
         Memtable.FlushablePartitionSet<?> flushSet = memtable.getFlushSet(from, to);
+        // TODO make this per table ,use table's sstable_format type
```
blambov, line 107: This needs to be implemented.

src/java/org/apache/cassandra/io/sstable/format/SSTableWriter.java
```diff
@@ -275,7 +275,7 @@ public final void abort()
     protected Map<MetadataType, MetadataComponent> finalizeMetadata()
     {
         return metadataCollector.finalizeMetadata(getPartitioner().getClass().getCanonicalName(),
-                                                  metadata().params.bloomFilterFpChance,
+                                                  metadata().params.getBloomFilterFpChance(),
```
blambov, line 278: This should be set by `SortedTableWriter` (the descendant that creates the filter component) using a value given by the filter as the metadata may have changed between the filter construction and this call.

This can be done in a separate ticket.

src/java/org/apache/cassandra/schema/SchemaEvent.java
```diff
@@ -248,6 +249,14 @@ private HashMap<String, Serializable> repr(CompressionParams compr)
         return ret;
     }
 
+    private HashMap<String, Serializable> repr(SSTableFormatParams sstableFormat)
+    {
+        HashMap<String, Serializable> ret = new HashMap<>();
+        if (sstableFormat == null) return ret;
+        ret.putAll(sstableFormat.asMap());
```
blambov, line 256: This will put the expanded definition in the schema, won't it?

This would defeat one of the main objectives of the ticket. A user must be able to specify e.g. `bti-fast` format for a table, and have that treated as specified by the individual node's `cassandra.yaml` rather than the coordinator's.

src/java/org/apache/cassandra/schema/SSTableFormatParams.java
```diff
@@ -0,0 +1,383 @@
...
+
+    public static final String DEFAUL_BIG_TYPE = "big";
+    public static final double DEFAULT_CRC_CHECK_CHANCE = 1.0;
+    public static final double INVALID_BLOOM_FILTER_FP_CHANCE = -1;
+    public static final double DEFAULT_BLOOM_FILTER_FP_CHANCE = 0.01;
+    public static final int DEFAULT_MAX_INDEX_INTERVAL = 2048;
+    public static final int DEFAULT_MIN_INDEX_INTERVAL = 128;
+
+    private static final Map<String, SSTableFormatParams> CONFIGURATIONS = new HashMap<>();
+    public static final SSTableFormatParams DEFAULT_BIG_SSTABLE = SSTableFormatParams.create(ImmutableMap.of(TYPE, DEFAUL_BIG_TYPE,
```
blambov, line 66: This should be called just `DEFAULT`.

src/java/org/apache/cassandra/schema/SSTableFormatParams.java
```diff
@@ -0,0 +1,383 @@
...
+        }
+    };
+
+    // wether these deprecated properties are defined by ddl or not
+    private Map<Option, Object> schemaProperties = new HashMap<>();
+
+    public enum Option
+    {
+        TYPE("type"),
+        BLOOM_FILTER_FP_CHANCE("bloom_filter_fp_chance"),
```
blambov, line 93: Parsing the format parameters should be completely up to the format factory. In other words, the only option this needs to understand is `type`, so that it can expand and merge the schema options with the yaml ones.

`bloom_filter_fp_chance` and `crc_check_chance` should be parsed by the `AbstractSSTableFormat`, or better a new shared superclass of `Big`- and `BtiFormat` that parses and holds the processed values.

`min/max_index_interval` are properties relevant to `BigFormat` only and should be parsed by it.

src/java/org/apache/cassandra/schema/TableParams.java
```diff
@@ -71,20 +72,22 @@ public String toString()
...
-    public final int maxIndexInterval;
+    @Deprecated public final int minIndexInterval;
+    @Deprecated public final int maxIndexInterval;
     public final SpeculativeRetryPolicy speculativeRetry;
     public final SpeculativeRetryPolicy additionalWritePolicy;
     public final CachingParams caching;
     public final CompactionParams compaction;
     public final CompressionParams compression;
     public final MemtableParams memtable;
+    public final SSTableFormatParams ssTableFormatParams;
```
blambov, line 90: In line with the fields above, this should be just `ssTableFormat`.

src/java/org/apache/cassandra/schema/TableParams.java
```diff
@@ -194,13 +200,54 @@ public void validate()
...
 
-    private static void fail(String format, Object... args)
+    public static void fail(String format, Object... args)
     {
         throw new ConfigurationException(format(format, args));
     }
 
+    public double getBloomFilterFpChance()
+    {
+        return bloomFilterFpChance == compaction.defaultBloomFilterFbChance() ? ssTableFormatParams.getBloomFilterFpChance(compaction) : bloomFilterFpChance;
```
blambov, line 233: This means that we practically have two values for this parameter in the system, which is bound to cause confusion.

Can we not keep only the one in the format? E.g. call methods on the sstable format if we see one of the deprecated flags and let that both handle conflict and store the value? (Note: this way the schema will also be updated to use the format-level properties the next time it is modified.)

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -602,6 +604,11 @@ private static void addTableParamsToRowBuilder(TableParams params, Row.SimpleBui
         // incremental_backups is enabled, to avoid RTE in pre-4.2 versioned node during upgrades
         if (!params.incrementalBackups)
             builder.add("incremental_backups", false);
+
+        // As above, only add the sstable_format column if the table uses a non-default sstable_fromat configuration to avoid RTE
```
blambov, line 608: Now that we have a compatibility mode setting, it makes better sense to use the compatibility level to determine if the setting should be given in the schema.

This is better done in a separate ticket.

test/unit/org/apache/cassandra/cql3/validation/operations/AlterTest.java
```diff
@@ -954,4 +955,113 @@ public void testAlterKeyspaceWithIfExists() throws Throwable
 
         assertInvalidThrow(InvalidRequestException.class, "ALTER KEYSPACE ks1 WITH replication= { 'class' : 'SimpleStrategy', 'replication_factor' : 1 }");
     }
+
+    @Test
+    public void testAlterTableWitSSTableFormatParams() throws Throwable
+    {
+        //use test cassandra.yaml
+        createTable("CREATE TABLE %s (a text, b int, c int, primary key (a, b)) WITH sstable_format = 'bti-fast'");
+        assertSchemaOption("sstable_format", map("bloom_filter_fp_chance", "0.01", "crc_check_chance", "0.1", "max_index_interval", "2048", "min_index_interval", "128", "type", "bti-fast"));
```
blambov, line 964: We don't want to lose the per-node configurability of the format, i.e. the specific property values should only be given if the user specified them. Otherwise these will override the settings in `cassandra.yaml`.

We can't return just the type as the user gave it, but we `map("type", "bti-fast")` should work just as well.

Maxwell-Guo:
@blambov thanks for the feedback, I will update a new version for this.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
Maxwell-Guo: No, I don't think so, only if we can change the [factory name](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/io/sstable/format/big/BigFormat.java#L635) with is big or bti.
as we only have big and bti format. But the kind of  parameters for big and bti may have more than two, According to my understanding to CASSANDRA-18534（If my understanding is wrong, please help me point it out ,thanks) there may be many custom-type that can be defined in yaml , for example , we may got big-fast, big-small, big-slow, bti-one, bti-two, and so on. each of them with different sstable parameters(I think it is ok to have same parameters wit different type name), and users can define their sstable format through cql's ddl , but the format should be defined in yaml before. 
(1)My expectation might be there a map with many entries which key is the format name (bit-one/ big-fast) value is the sstable parameters;
(2)but as factory name are big or bti , so the finally map number is only two. 
I originally wanted to make modifications as (1) described, but I think it would be better to start after confirming with you, to confirm whether my modifications meet your expectations.
And you can see that I have left a comment  [TODO](https://github.com/Maxwell-Guo/cassandra/blob/CASSANDRA-18534/src/java/org/apache/cassandra/db/memtable/Flushing.java#L107) here , that is the same reason with (1). @blambov

src/java/org/apache/cassandra/schema/SchemaEvent.java
Maxwell-Guo: that's to say : 1. this should be removed or left empty (I got what you mean now);2. if node_a and node_b have different  parameters in cassandra.yaml for sstable_format with the same name bti-fast, and we create tableA in node_a with bit-fast, then there may be two situations:
a ) node_a's configuration for bti-fast,e.g. `create table xxx with sstable_format = 'bti-fast'` can just match the create table's ddl , we can use node_a's yaml, and node_b will use node_b's yaml as the type for ddl is bti-fast and fortunately we got bti-fast configuration in node_b' yaml;
b) node_a's  have bti-fast 's configurations but the ddl only set the type but the parameters are not same with configuration, so we will use user's defined parameter, and node_b will use the yaml's configuration of his own , if node_b don't have bti-fast's configuration , the default configuration will be used.
is that all right ? @blambov

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
Maxwell-Guo: what about in CASSANDRA-18533 ?

src/java/org/apache/cassandra/schema/SSTableFormatParams.java
Maxwell-Guo: I got it, the enum here is just to list the options that can be defined by yaml and cql.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
Maxwell-Guo: modify the code so we can get sstableFormats which is a map , where key is custom-type name , and value is the specific type instance(bti or big), as for the type instance , it is composed of type name and type options. 
I changed the type name for each type instance from the fixed value (big or bti) to the custom-type name, and the option is the sstable options.

src/java/org/apache/cassandra/db/memtable/Flushing.java
Maxwell-Guo: we may need to merge the configured value and cql defined value, not using the [sstableFormats](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/config/DatabaseDescriptor.java#L226) for different custom-type, if we want to sstable format options to be the latest.As the option value in sstableFormats may not be the latest. And what are the [options](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/io/sstable/format/AbstractSSTableFormat.java#L27) used for ?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
blambov: Currently the two formats do not implement format-specific parameters. This is mainly because BTI does not (at the moment) need new ones, and all the existing settings that apply to BIG exist in non-format-specific form. CASSANDRA-18533 was meant to move these settings into the format configuration. 

It makes better sense to implement that ticket first, and I have started to make comments which ask for the inclusion of 18533 in the implementation of this, for which I must apologise.

Taking a step back, how about pausing the development of per-table config for a while, and doing per-format config (CASSANDRA-18533) first? While there are no options for any of the formats, much of the machinery we are implementing here is useless and that makes it hard to understand and code.

Maxwell-Guo: I think it's okay, maybe that's why, during the development process,I discovered something subtle. Maybe it would be more appropriate to do 18533 first.

PR #2787, opened 2023-10-09
Fixes for in-tree scripts: offline mode, maybe-build for fqltool-test…

michaelsembwever:
…, jvm-dtest-upgrade

 - jvm-dtest-upgrade script running in ci-cassandra.a.o (needs default bash)
 - add offline mode (to avoid `git fetch` before building dtest-jars on other branches),
 - avoid rebuilding in fqltool-test target, as is standard for test targets.
 - python detection and override
 - FileTest cleanup (tmp directory was left unreadable and undeletable)
 - reuse ~/.ccm (like normal)
 - remove support for python 3.6

CASSANDRA-18665

### DIFF
--- a/.build/docker/run-tests.sh
+++ b/.build/docker/run-tests.sh
@@ -69,8 +69,11 @@ if [[ ! "${java_version}" =~ $regx_java_version ]]; then
     exit 1
 fi
 
-python_version="3.6"
-command -v python >/dev/null 2>&1 && python_version="$(python -V | awk '{print $2}' | awk -F'.' '{print $1"."$2}')"
+# allow python version override, otherwise default to current python version or 3.7
+if [ "x" == "x${python_version}" ] ; then
+    command -v python >/dev/null 2>&1 && python_version="$(python -V 2>&1 | awk '{print $2}' | awk -F'.' '{print $1"."$2}')"
+    python_version="${python_version:-3.7}"
+fi
 
 # print debug information on versions
 docker --version
@@ -150,6 +153,7 @@ elif [[ "${target}" =~ dtest* ]] ; then
 else
     docker_flags="--cpus=${docker_cpus} -m 5g --memory-swap 5g"
 fi
+
 docker_flags="${docker_flags} --env-file build/env.list -d --rm"
 
 # make sure build_dir is good
--- a/.build/docker/run-tests.sh
+++ b/.build/docker/run-tests.sh
@@ -85,6 +85,8 @@ dockerfile="ubuntu2004_test.docker"
 image_tag="$(md5sum docker/${dockerfile} | cut -d' ' -f1)"
 image_name="apache/cassandra-${dockerfile/.docker/}:${image_tag}"
 docker_mounts="-v ${cassandra_dir}:/home/cassandra/cassandra -v "${build_dir}":/home/cassandra/cassandra/build -v ${HOME}/.m2/repository:/home/cassandra/.m2/repository"
+# HACK hardlinks in overlay are buggy, the following mount prevents hardlinks from being used. ref $TMP_DIR in .build/run-tests.sh
+docker_mounts="${docker_mounts} -v "${build_dir}/tmp":/home/cassandra/cassandra/build/tmp"
 
 # Look for existing docker image, otherwise build
 if ! ( [[ "$(docker images -q ${image_name} 2>/dev/null)" != "" ]] ) ; then
--- a/.build/docker/ubuntu2004_test.docker
+++ b/.build/docker/ubuntu2004_test.docker
@@ -25,9 +25,6 @@ ENV PYTHONUNBUFFERED=true
 
 LABEL org.cassandra.buildenv=ubuntu_2004
 
-VOLUME ${CASSANDRA_DIR}
-VOLUME ${DIST_DIR}
-
 RUN echo "Building with arguments:" \
     && echo " - DIST_DIR=${DIST_DIR}" \
     && echo " - BUILD_HOME=${BUILD_HOME}" \
@@ -41,7 +38,7 @@ RUN echo 'Acquire::http::Timeout "60";' > /etc/apt/apt.conf.d/80proxy.conf
 RUN echo 'Acquire::ftp::Timeout "60";' >> /etc/apt/apt.conf.d/80proxy.conf
 
 # install our python dependencies and some other stuff we need
-# libev4 libev-dev are for the python driver / libssl-dev is for python3.6
+# libev4 libev-dev are for the python driver
 
 RUN export DEBIAN_FRONTEND=noninteractive && \
     apt-get update && \
@@ -51,20 +48,17 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
     add-apt-repository -y ppa:deadsnakes/ppa && \
     apt-get update && \
     apt-get install -y curl git-core python3-pip \
-        python3.6 python3.6-venv python3.6-dev \
         python3.7 python3.7-venv python3.7-dev \
         python3.8 python3.8-venv python3.8-dev \
         python3.11 python3.11-venv python3.11-dev \
-        virtualenv net-tools libev4 libev-dev wget gcc libssl-dev libxml2 libxslt1-dev \
+        virtualenv net-tools libev4 libev-dev wget gcc libxml2 libxslt1-dev \
         vim lsof sudo libjemalloc2 dumb-init locales \
         openjdk-8-jdk openjdk-11-jdk openjdk-17-jdk ant ant-optional
 
 
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.6 2
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.7 3
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.8 4
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.11 5
-RUN python3.6 -m pip install --upgrade pip
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.7 2
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.8 3
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.11 4
 RUN python3.7 -m pip install --upgrade pip
 RUN python3.8 -m pip install --upgrade pip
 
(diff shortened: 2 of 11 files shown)


michaelsembwever:
@Mmuzaf

.build/run-python-dtests.sh
```diff
@@ -73,7 +72,7 @@ if [ "${java_version}" -eq 17 ] && [[ "${target}" == "dtest-upgrade" ]] ; then
     exit 1
 fi
 
-python_version=$(python -V | awk '{print $2}' | awk -F'.' '{print $1"."$2}')
+python_version=$(python -V 2>&1 | awk '{print $2}' | awk -F'.' '{print $1"."$2}')
```
Mmuzaf, line 75: Since we removed the 3.6 version, should we add a constraint here that version >= 3.7?

Mmuzaf (commented):
.build/run-python-dtests.sh:

`sed -r "s/<[\/]?testsuites>//g" ${DIST_DIR}/test/output/nosetests.xml > /tmp/nosetests.xml`

Should we save an intermediate result to the `${TMPDIR}` ?

.build/run-python-dtests.sh
michaelsembwever: in bash how would you do that ?

Mmuzaf: `cassandra_regx_supported_versions="^(2.2|3.0|3.11|4.0|4.1)(.[0-9]+)?$"`

Like this one in the accompanying script in the [cassandra-builds](https://github.com/apache/cassandra-builds/pull/96/files#diff-f458d745bb6227c4f9161c75ffed2a98d0ee3cd7da734afe7e759670c5969b3cR23) :-)

michaelsembwever: done in https://github.com/apache/cassandra/compare/90182972ebc0dee7f936058f222d2df966893d1d..557f90ed5ccdaa1e86459e135017ef96cb64bc4b

michaelsembwever:
> Should we save an intermediate result to the ${TMPDIR} ? 

done in https://github.com/apache/cassandra/compare/90182972ebc0dee7f936058f222d2df966893d1d..557f90ed5ccdaa1e86459e135017ef96cb64bc4b

.build/docker/ubuntu2004_test.docker
```diff
@@ -146,11 +131,26 @@ RUN /bin/bash -c "export CASS_DRIVER_NO_CYTHON=1 CASS_DRIVER_NO_EXTENSIONS=1 \
...
-    ccm create -n 1 -v git:trunk test && ccm remove test && \
+RUN /bin/bash -c "source ${BUILD_HOME}/env3.7/bin/activate && \
     ccm create -n 1 -v git:cassandra-4.1 test && ccm remove test && \
     ccm create -n 1 -v git:cassandra-4.0 test && ccm remove test"
 
+# Initialize ccm versions.  right side of each sequence needs to be updated with new releases.
+#  this can be checked with:
+#  `curl -s https://downloads.apache.org/cassandra/ | grep -oP '(?<=href=\")[0-9]+\.[0-9]+\.[0-9]+(?=)' | sort -V | uniq -w 3`
+RUN bash -c 'source ~/env3.7/bin/activate && \
+    for i in {1..11} ; do echo $i ; ccm create --quiet -n 1 -v binary:4.0.$i test && ccm remove test ; done && \
```
Mmuzaf, line 142: I'm not sure if it's worth doing the way below, but I think we can get 4.x releases from the `downloads'.  WDYT?

```
m@MBP ~ % curl --silent https://downloads.apache.org/cassandra/ | grep -o '\[DIR\]\"> <a href=".*">' | sed 's/\[DIR\]\"> <a href="//;s/\/">//' | grep -E '^([4-4]|[0-9]{2,})\..*$' | 

The output:

4.0.11
4.1.3
```

```
m@MBP ~ % curl --silent https://downloads.apache.org/cassandra/ | grep -o '\[DIR\]\"> <a href=".*">' | sed 's/\[DIR\]\"> <a href="//;s/\/">//' | grep -E '^([4-4]|[0-9]{2,})\..*$' | awk -F. '{ for(i=1; i<=$NF; i++) system("echo Executing command for version "$0", iteration " i) }'

Executing command for version 4.0.11, iteration 1
Executing command for version 4.0.11, iteration 2
Executing command for version 4.0.11, iteration 3
Executing command for version 4.0.11, iteration 4
Executing command for version 4.0.11, iteration 5
Executing command for version 4.0.11, iteration 6
Executing command for version 4.0.11, iteration 7
Executing command for version 4.0.11, iteration 8
Executing command for version 4.0.11, iteration 9
Executing command for version 4.0.11, iteration 10
Executing command for version 4.0.11, iteration 11
Executing command for version 4.1.3, iteration 1
Executing command for version 4.1.3, iteration 2
Executing command for version 4.1.3, iteration 3
```

michaelsembwever: Can you get it into this line:
```
RUN bash -c 'source ~/env3.7/bin/activate && \
    for i in {1..11} ; do echo $i ; ccm create --quiet -n 1 -v binary:4.0.$i test && ccm remove test ; done && \
    for i in {1..3}  ; do echo $i ; ccm create --quiet -n 1 -v binary:4.1.$i test && ccm remove test ; done'
```

i've failed at every attempt so far…

Mmuzaf: I've managed to improve the latest version :-)

```
mmuzaf@Maxims-MBP ~ % curl --silent https://downloads.apache.org/cassandra/ | grep -o '\[DIR\]\"> <a href=".*">' | sed 's/\[DIR\]\"> <a href="//;s/\/">//' | grep -E '^([4-4]|[0-9]{2,})\..*$' | awk -F. '{ for(i=1; i<=$NF; i++) print $1 "." $2 "." i }' 
4.0.1
4.0.2
4.0.3
4.0.4
4.0.5
4.0.6
4.0.7
4.0.8
4.0.9
4.0.10
4.0.11
4.1.1
4.1.2
4.1.3
```

michaelsembwever: it's the `RUN bash -c ' … '` that's the biggest hurdle

Mmuzaf: Yeah, it seems this is a problem...
What if we create a script`_init_ccm.sh 4.0 4.1` and run it instead?

michaelsembwever: separate ticket for follow up work ?

Mmuzaf: +1 
:-)

Mmuzaf (approved):
I'm a little worried about the way we configure ccm, but I've seen that it's not a new code and it's actually been done before this. So, let's create an issue for that and continue with everything we have now. 

The rest of the changes are LGTM

.build/docker/run-tests.sh
```diff
@@ -69,8 +69,11 @@ if [[ ! "${java_version}" =~ $regx_java_version ]]; then
     exit 1
 fi
 
-python_version="3.6"
-command -v python >/dev/null 2>&1 && python_version="$(python -V | awk '{print $2}' | awk -F'.' '{print $1"."$2}')"
+# allow python version override, otherwise default to current python version or 3.7
+if [ "x" == "x${python_version}" ] ; then
+    command -v python >/dev/null 2>&1 && python_version="$(python -V 2>&1 | awk '{print $2}' | awk -F'.' '{print $1"."$2}')"
+    python_version="${python_version:-3.7}"
```
jmckenzie-dev, line 75: I don't have a horse in this race, but why 3.7? vs. 3.8, 3.11, etc.

.build/docker/ubuntu2004_test.docker
```diff
@@ -41,7 +38,7 @@ RUN echo 'Acquire::http::Timeout "60";' > /etc/apt/apt.conf.d/80proxy.conf
 RUN echo 'Acquire::ftp::Timeout "60";' >> /etc/apt/apt.conf.d/80proxy.conf
 
 # install our python dependencies and some other stuff we need
-# libev4 libev-dev are for the python driver / libssl-dev is for python3.6
+# libev4 libev-dev are for the python driver
```
jmckenzie-dev, line 41: Are we dropping python3.6? Asking for a friend. 😀

.build/docker/ubuntu2004_test.docker
jmckenzie-dev, line 142: So we're intentionally trying to init 4.0.1 to 4.0.11 right? And when a 4.0.12 hits we'll need to manually add that here?

If we want to be... clever? Hacky? We could do something like:
`git tag | grep -E 'cassandra-4.0.[0-9]+$' | cut -d '.' -f 3 | sort -nr | head -n 1`

to parse out the latest version at time of docker image creation and iterate through all up to that point to init them. Alternatively, or along with that, might be nice to add a comment to indicate why we have the top end of the sequence here since it's non-obvious.

.build/run-tests.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
```
jmckenzie-dev, line 1: Was this an accidental change or intentional? If intentional (for my own education) - why the change?

.build/run-tests.sh
```diff
@@ -84,7 +87,9 @@ _build_all_dtest_jars() {
     fi
 
     if [ -d ${TMP_DIR}/cassandra-dtest-jars ] && [ "https://github.com/apache/cassandra.git" == "$(git -C ${TMP_DIR}/cassandra-dtest-jars remote get-url origin)" ] ; then
-        until git -C ${TMP_DIR}/cassandra-dtest-jars fetch --quiet origin ; do echo "git pull failed… trying again… " ; done
+      if [ "x" == "x${OFFLINE}" ] ; then
+        until git -C ${TMP_DIR}/cassandra-dtest-jars fetch --quiet origin ; do echo "git fetch failed… trying again… " ; done
```
jmckenzie-dev, line 91: Can this infinite loop / hang? And if so maybe we constrain it and explain why the need?

jmckenzie-dev (commented):
Think it's fine to merge in its current form though there's a couple/few thing that might be nice to comment or clarify on commit.

jmckenzie-dev (approved):
... I hit the wrong radio button. Meant to approve. ;)

.build/docker/ubuntu2004_test.docker
Mmuzaf: I thought about it too, but this approach may fail the CI if the git tag is added before the actual release happens, right? That's why fetching from the archive is better, I think...

.build/docker/ubuntu2004_test.docker
michaelsembwever: yes.  it simply does not work at all anymore.

.build/docker/run-tests.sh
michaelsembwever: just a conservative/minimal bump after dropping 3.6

(am all for discussing removing 3.7 too (in a separate ticket))

.build/run-tests.sh
michaelsembwever: it can, but you get the echo statements so it's visible to the user. they can ctl-c it.

this poormans approach is repeated in a number of places.  improving it (which is smart) should be done together.  and maybe just rewriting bash altogether into something non-shit

.build/run-tests.sh
michaelsembwever: I saw it break in some environments (where it picked up custom bash command on the path).
this was the simple fix. i haven't yet heard of an env where /bin/bash does not exist.

this approach is also the standard for the other scripts.

michaelsembwever:
Committed as https://github.com/apache/cassandra/commit/987d03c142f086f05857f9c6a2eafef7e157eb15

PR #2795, opened 2023-10-11
Fix DescribeStatement

jacek-lewandowski:


src/java/org/apache/cassandra/cql3/statements/DescribeStatement.java
```diff
@@ -132,8 +132,13 @@ public final ResultMessage execute(QueryState state, QueryOptions options, long
...
     public ResultMessage executeLocally(QueryState state, QueryOptions options)
     {
-        Keyspaces keyspaces = Schema.instance.distributedAndLocalKeyspaces();
-        UUID schemaVersion = Schema.instance.getVersion();
+        UUID schemaVersion;
+        Keyspaces keyspaces;
+        do
+        {
+            schemaVersion = Schema.instance.getVersionSync();
+            keyspaces = Schema.instance.distributedAndLocalKeyspaces();
```
Maxwell-Guo: will the version always be inconsistent for a long time ,and we are in this loop for long time.

jacek-lewandowski: Not sure - is it a statement or question? I think we can skip this sync if paging is not requested; if paging was requested and there happened what you say, we would never be able to return results as paging would always fail due to concurrent schema changes. On the other hand - I don't know under what circumstances that could happen?

Maxwell-Guo: long time ago, in our cluster， with poor machine configuration, we some time saw that the cluster schema may be inconsistent, and it take a long time for them to reach a consistent state. Some times it due to the poor disk hung, where the system table just locate. Some time is other reasons.

So what about logging here when the loop  is more than some times like 3 ?

jacek-lewandowski: Actually, I don't think this is a case - it is not waiting here for the cluster schema to be consistent - it is only the local thing - when the schema gets transformed locally, the transformation is applied in a synchronized block. A single transformation may include multiple items which are applied one by one immediately updating the snapshot. On the other hand, scheme version is updated once after the whole transformation is completed. So, here we are just waiting for that synchronized block to complete and make sure that no other synchronized transformation has started in the meantime. In other words, waiting to make sure the metadata and version pertains to the same schema.

Maxwell-Guo: Oh, I see.

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1037/workflows/b371e061-4128-446d-b1a9-1adecf3972e9 ⚠️  - one simulation dtest failed, the test is already reported to be flaky: https://issues.apache.org/jira/browse/CASSANDRA-18944

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1037/workflows/7af077c4-d44c-461f-8c8a-750790d49b14 (j11) ⚠️  - JMXFeatureFailed with "Failed to bind port 38065 on 127.0.0.1" it does not seem related

(rerun succeeded https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1037/workflows/7af077c4-d44c-461f-8c8a-750790d49b14/jobs/42939/parallel-runs/4?filterBy=ALL )

PR #2804, opened 2023-10-13
Measure repair state caches by memory retention

aratno:
CASSANDRA-18904

### DIFF
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -914,7 +914,7 @@ public class Config
     // Only one of repair_state_size and repair_state_heap_size should be set
     @Deprecated
     public volatile Integer repair_state_size = null;
-    public volatile DataStorageSpec.IntBytesBound repair_state_heap_size = new DataStorageSpec.IntBytesBound(5, DataStorageSpec.DataStorageUnit.MEBIBYTES);
+    public volatile DataStorageSpec.IntBytesBound repair_state_heap_size = null;
 
     /** The configuration of timestamp bounds */
     public volatile DurationSpec.LongMicrosecondsBound maximum_timestamp_warn_threshold = null;
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -910,7 +910,11 @@ public class Config
     public volatile Map<StartupCheckType, Map<String, Object>> startup_checks = new HashMap<>();
 
     public volatile DurationSpec.LongNanosecondsBound repair_state_expires = new DurationSpec.LongNanosecondsBound("3d");
-    public volatile int repair_state_size = 100_000;
+
+    // Only one of repair_state_size and repair_state_heap_size should be set
+    @Deprecated
+    public volatile Integer repair_state_size = null;
+    public volatile DataStorageSpec.IntBytesBound repair_state_heap_size = new DataStorageSpec.IntBytesBound(5, DataStorageSpec.DataStorageUnit.MEBIBYTES);
 
     /** The configuration of timestamp bounds */
     public volatile DurationSpec.LongMicrosecondsBound maximum_timestamp_warn_threshold = null;
--- a/src/java/org/apache/cassandra/repair/CommonRange.java
+++ b/src/java/org/apache/cassandra/repair/CommonRange.java
@@ -21,21 +21,30 @@ package org.apache.cassandra.repair;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Objects;
 import java.util.Set;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
 
+import org.apache.cassandra.cache.IMeasurableMemory;
+import org.apache.cassandra.dht.Murmur3Partitioner;
 import org.apache.cassandra.dht.Range;
 import org.apache.cassandra.dht.Token;
 import org.apache.cassandra.locator.InetAddressAndPort;
+import org.apache.cassandra.utils.FBUtilities;
+import org.apache.cassandra.utils.ObjectSizes;
 
 /**
  * Groups ranges with identical endpoints/transient endpoints
  */
-public class CommonRange
+public class CommonRange implements IMeasurableMemory
 {
+    // REVIEW: I've heard from some that these should be on a single line, others prefer them split out into separate lines
+    // in a static block. I'm open to either.
+    private static final long EMPTY_SIZE = ObjectSizes.measure(new CommonRange(ImmutableSet.of(FBUtilities.getBroadcastAddressAndPort()), Collections.emptySet(), Range.rangeSet(new Range<>(new Murmur3Partitioner.LongToken(Long.MIN_VALUE), new Murmur3Partitioner.LongToken(Long.MIN_VALUE)))));
+
     public final ImmutableSet<InetAddressAndPort> endpoints;
     public final ImmutableSet<InetAddressAndPort> transEndpoints;
     public final Collection<Range<Token>> ranges;
@@ -91,5 +100,18 @@ public class CommonRange
                ", ranges=" + ranges +
                ", hasSkippedReplicas=" + hasSkippedReplicas +
                '}';
+    }
+
+    @Override
+    public long unsharedHeapSize()
+    {
+        long size = EMPTY_SIZE;
+        for (InetAddressAndPort addr : endpoints)
+            size += addr.unsharedHeapSize();
+        for (InetAddressAndPort addr : transEndpoints)
+            size += addr.unsharedHeapSize();
+        for (Range<Token> range : ranges)
+            size += ObjectSizes.sizeOf(range);
+        return size;
     }
 }
--- a/src/java/org/apache/cassandra/repair/CommonRange.java
+++ b/src/java/org/apache/cassandra/repair/CommonRange.java
@@ -41,8 +41,6 @@ import org.apache.cassandra.utils.ObjectSizes;
  */
 public class CommonRange implements IMeasurableMemory
 {
-    // REVIEW: I've heard from some that these should be on a single line, others prefer them split out into separate lines
-    // in a static block. I'm open to either.
     private static final long EMPTY_SIZE = ObjectSizes.measure(new CommonRange(ImmutableSet.of(FBUtilities.getBroadcastAddressAndPort()), Collections.emptySet(), Range.rangeSet(new Range<>(new Murmur3Partitioner.LongToken(Long.MIN_VALUE), new Murmur3Partitioner.LongToken(Long.MIN_VALUE)))));
 
     public final ImmutableSet<InetAddressAndPort> endpoints;
--- a/src/java/org/apache/cassandra/repair/state/ParticipateState.java
+++ b/src/java/org/apache/cassandra/repair/state/ParticipateState.java
@@ -18,10 +18,12 @@
 package org.apache.cassandra.repair.state;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 import java.util.stream.Collectors;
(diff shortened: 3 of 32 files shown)


src/java/org/apache/cassandra/repair/CommonRange.java
```diff
@@ -21,21 +21,30 @@
...
+import org.apache.cassandra.utils.FBUtilities;
+import org.apache.cassandra.utils.ObjectSizes;
 
 /**
  * Groups ranges with identical endpoints/transient endpoints
  */
-public class CommonRange
+public class CommonRange implements IMeasurableMemory
 {
+    // REVIEW: I've heard from some that these should be on a single line, others prefer them split out into separate lines
```
dcapwell: https://cassandra.apache.org/_/development/code_style.html#multiline-statements

> Where possible prefer keeping a logical action to a single line. Prefer introducing additional variables, or well-named methods encapsulating actions, to multi-line statements - unless this harms clarity (e.g. in an already short method).
> Try to keep lines under 120 characters, but use good judgment. It is better to exceed this limit, than to split a line that has no natural splitting points, particularly when the remainder of the line is boilerplate or easily inferred by the reader.

src/java/org/apache/cassandra/service/ActiveRepairService.java
```diff
@@ -245,16 +250,64 @@ public ActiveRepairService(SharedContext ctx)
...
+            logger.info("Storing repair state for {} or for {} retained size", duration, maxRetainedSize);
+            repairs = Caffeine.newBuilder()
+                              .expireAfterWrite(duration.quantity(), duration.unit())
+                              .weigher(new Weigher<TimeUUID, CoordinatorState>()
+                              {
+                                  @Override
+                                  public int weigh(TimeUUID id, CoordinatorState coordinatorState)
+                                  {
+                                      // REVIEW: This addition is unchecked - I was planning to not check memory accounting
+                                      // additions into long because the maximum before overflow is ~thousands of PB, but
```
dcapwell: sounds good to me

src/java/org/apache/cassandra/utils/ObjectSizes.java
```diff
@@ -47,6 +51,8 @@ public class ObjectSizes
 
     public static final long IPV6_SOCKET_ADDRESS_SIZE = ObjectSizes.measureDeep(new InetSocketAddress(getIpvAddress(16), 42));
 
+    private static final long UUID_SIZE = measure(new UUID(0L, 0L));
```
dcapwell: this is dead code

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -910,7 +910,11 @@ public static void setClientMode(boolean clientMode)
     public volatile Map<StartupCheckType, Map<String, Object>> startup_checks = new HashMap<>();
 
     public volatile DurationSpec.LongNanosecondsBound repair_state_expires = new DurationSpec.LongNanosecondsBound("3d");
-    public volatile int repair_state_size = 100_000;
+
+    // Only one of repair_state_size and repair_state_heap_size should be set
+    @Deprecated
+    public volatile Integer repair_state_size = null;
+    public volatile DataStorageSpec.IntBytesBound repair_state_heap_size = new DataStorageSpec.IntBytesBound(5, DataStorageSpec.DataStorageUnit.MEBIBYTES);
```
dcapwell: spoke in slack, this is a bug as we can not tell if the user set `5mb` or not... 

here is the semantic we would want in yamll

```
repair_state_size: 42
```

we are size based

```
repair_state_heap_size: 10mb
```

we are heap based

```
repair_state_size: 42
repair_state_heap_size: 10mb
```

reject

src/java/org/apache/cassandra/repair/state/WeightedHierarchy.java
```diff
@@ -0,0 +1,63 @@
...
+        {
+            Root root = root();
+            root.totalNestedRetainedSize().addAndGet(nested.independentRetainedSize());
+            root.onRetainedSizeUpdate();
+        }
+    }
+
+    interface Root extends InternalNode, IMeasurableMemory
+    {
+        AtomicLong totalNestedRetainedSize();
```
dcapwell, line 54: this is leaky... why return the `AtomicLong` rather than return `long`?

src/java/org/apache/cassandra/repair/state/WeightedHierarchy.java
```diff
@@ -0,0 +1,63 @@
...
+
+        default void onNestedStateRegistration(Node nested)
+        {
+            Root root = root();
+            root.totalNestedRetainedSize().addAndGet(nested.independentRetainedSize());
+            root.onRetainedSizeUpdate();
+        }
+    }
+
+    interface Root extends InternalNode, IMeasurableMemory
```
dcapwell, line 52: based off the usage, we really don't care about `totalNestedRetainedSize` and `independentRetainedSize`... we only care about `unsharedHeapSize`

src/java/org/apache/cassandra/repair/state/WeightedHierarchy.java
```diff
@@ -0,0 +1,63 @@
...
+    {
+        AtomicLong totalNestedRetainedSize();
+
+        default long unsharedHeapSize()
```
dcapwell, line 56:
```suggestion
        @Override
        default long unsharedHeapSize()
```

src/java/org/apache/cassandra/repair/state/WeightedHierarchy.java
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * Internal nodes must maintain a reference to the root, in order to notify it of any new registrations that may impact
+ * the total weight of the hierarchy.
+ */
+class WeightedHierarchy
+{
+    interface Node
+    {
+        // How much does this object retain without any of its nested state
+        long independentRetainedSize();
```
dcapwell, line 37: we spoke before but new readers won't know why you did this vs `IMeasurableMemory`... can you document this?

src/java/org/apache/cassandra/repair/state/WeightedHierarchy.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.cassandra.cache.IMeasurableMemory;
+
+/**
+ * A pattern for tracking the weight of a hierarchy of references. Registered objects are treated as though they are
+ * immutable, and are only measured on registration.
+ *
+ * Internal nodes must maintain a reference to the root, in order to notify it of any new registrations that may impact
+ * the total weight of the hierarchy.
+ */
+class WeightedHierarchy
```
dcapwell, line 32: really not a fan of this class, feels like its more complex than it really needs to be...  why did you go this route vs your original?

src/java/org/apache/cassandra/service/ActiveRepairService.java
```diff
@@ -245,16 +250,64 @@ public ActiveRepairService(SharedContext ctx)
...
+                                      // I check any casts into ints (4GB). Is that a reasonable policy?
+
+                                      long retained = coordinatorState.totalNestedRetainedSize().get() + coordinatorState.independentRetainedSize();
```
dcapwell, line 279:
```suggestion
                                      long retained = coordinatorState.unsharedHeapSize();
```

src/java/org/apache/cassandra/service/ActiveRepairService.java
```diff
@@ -245,16 +250,64 @@ public ActiveRepairService(SharedContext ctx)
...
+                              .executor(ImmediateExecutor.INSTANCE)
+                              .build();
+            participates = Caffeine.newBuilder()
+                                   .expireAfterWrite(duration.quantity(), duration.unit())
+                                   .weigher(new Weigher<TimeUUID, ParticipateState>()
+                                   {
+                                       @Override
+                                       public int weigh(TimeUUID id, ParticipateState participateState)
+                                       {
+                                           long retained = participateState.totalNestedRetainedSize().get() + participateState.independentRetainedSize();
```
dcapwell, line 294:
```suggestion
                                           long retained = participateState.unsharedHeapSize();
```

src/java/org/apache/cassandra/repair/state/WeightedHierarchy.java
aratno: Since CASSANDRA-18816 introduces more hierarchy (ParticipateState now contains Job, which contains validation and syncs) there are now two of these nested hierarchies, the other one being CoordinatorState-SessionState-JobState. I thought it might be easier to see which objects are roots and which are nested. If it's not your preference I'll remove it and put it back the way it used to be.

src/java/org/apache/cassandra/config/Config.java
aratno: Addressed - thanks.

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -910,7 +910,11 @@ public static void setClientMode(boolean clientMode)
     public volatile Map<StartupCheckType, Map<String, Object>> startup_checks = new HashMap<>();
 
     public volatile DurationSpec.LongNanosecondsBound repair_state_expires = new DurationSpec.LongNanosecondsBound("3d");
-    public volatile int repair_state_size = 100_000;
+
+    // Only one of repair_state_size and repair_state_heap_size should be set
+    @Deprecated
```
dcapwell, line 915: can you add a `since`?  I think we are trying to do that now...

src/java/org/apache/cassandra/repair/state/ParticipateState.java
```diff
@@ -53,14 +61,54 @@ public enum RegisterStatus
 
     public final Phase phase = new Phase();
 
+    public final ConcurrentMap<RepairJobDesc, Job> jobs = new ConcurrentHashMap<>();
+
+    private static final long EMPTY_SIZE = ObjectSizes.measure(new ParticipateState(SharedContext.Global.instance.clock(), null, new PrepareMessage(nextTimeUUID(), Collections.emptyList(), Collections.emptyList(), false, 0L, false, PreviewKind.NONE)));
```
dcapwell, line 66: should define the static as the first set of variables

src/java/org/apache/cassandra/repair/state/ParticipateState.java
```diff
@@ -81,7 +129,11 @@ public Job job(RepairJobDesc desc)
 
     public Job getOrCreateJob(RepairJobDesc desc)
     {
-        return jobs.computeIfAbsent(desc, d -> new Job(clock, d));
+        boolean isNew = !jobs.containsKey(desc);
+        Job job = jobs.computeIfAbsent(desc, d -> new Job(clock, this, d));
```
dcapwell, line 133: this isn't thread safe...

src/java/org/apache/cassandra/repair/state/ParticipateState.java
```diff
@@ -148,16 +200,43 @@ public void accept()
...
+        private final ParticipateState participateState;
         private final AtomicReference<ValidationState> validation = new AtomicReference<>(null);
         private final ConcurrentMap<SyncState.Id, SyncState> syncs = new ConcurrentHashMap<>();
+        private static final long EMPTY_SIZE;
+        static
+        {
+            PrepareMessage msg = new PrepareMessage(nextTimeUUID(), Collections.emptyList(), Collections.emptyList(), false, 0L, false, PreviewKind.NONE);
+            ParticipateState participateState = new ParticipateState(Clock.Global.clock(), FBUtilities.getBroadcastAddressAndPort(), msg);
+            RepairJobDesc desc = new RepairJobDesc(null, null, null, null, null);
+            EMPTY_SIZE = ObjectSizes.measure(new Job(Clock.Global.clock(), participateState, desc));
```
dcapwell, line 216:
```suggestion
            EMPTY_SIZE = ObjectSizes.measure(new Job(Clock.Global.clock(), null, null));
```

could cleanup some code / imports...

src/java/org/apache/cassandra/repair/state/ParticipateState.java
```diff
@@ -148,16 +200,43 @@ public void accept()
...
+
+        @Override
+        public long independentRetainedSize()
+        {
+            long size = EMPTY_SIZE;
+
+            // Excludes participateState since that's already measured as the parent, and both validation and syncs propagate
+            // their updates back to the root so do not need to be counted here.
+
+            return size;
```
dcapwell, line 239:
```suggestion
            // Excludes participateState since that's already measured as the parent, and both validation and syncs propagate
            // their updates back to the root so do not need to be counted here.

            return EMPTY_SIZE;
```

src/java/org/apache/cassandra/repair/state/SessionState.java
```diff
@@ -42,14 +54,18 @@ public enum State
     public final String keyspace;
     public final String[] cfnames;
     public final CommonRange commonRange;
+    private final CoordinatorState coordinator;
     private final ConcurrentMap<UUID, JobState> jobs = new ConcurrentHashMap<>();
 
     public final Phase phase = new Phase();
 
-    public SessionState(Clock clock, TimeUUID parentRepairSession, String keyspace, String[] cfnames, CommonRange commonRange)
+    private static final long EMPTY_SIZE = ObjectSizes.measure(new SessionState(new CoordinatorState(SharedContext.Global.instance.clock(), 0, "", new RepairOption(RepairParallelism.SEQUENTIAL, false, false, false, 8, Collections.emptySet(), false, false, false, PreviewKind.NONE, false, false, false, false)), SharedContext.Global.instance.clock(), "", new String[] {}, new CommonRange(ImmutableSet.of(FBUtilities.getBroadcastAddressAndPort()), Collections.emptySet(), Range.rangeSet(new Range<>(new Murmur3Partitioner.LongToken(Long.MIN_VALUE), new Murmur3Partitioner.LongToken(Long.MIN_VALUE))))));
```
dcapwell, line 62: since you do `ObjectSize.measure` it would be good to do `null` than define so much that isn't actually needed...

dcapwell (changes requested):

src/java/org/apache/cassandra/repair/state/WeightedHierarchy.java
dcapwell: ill leave to your discretion... not a fan and feel like the old solution of having just the counter in both classes with nested objects incrementing was cleaner and less code...

PR #2805, opened 2023-10-13
Measure repair state caches by memory retention

aratno:
CASSANDRA-18904

This PR is for 4.1, which slightly differs from trunk due to the exclusion of CASSANDRA-18816 on 4.1. See #2804 for the trunk PR.

### DIFF
--- a/src/java/org/apache/cassandra/repair/CommonRange.java
+++ b/src/java/org/apache/cassandra/repair/CommonRange.java
@@ -21,21 +21,30 @@ package org.apache.cassandra.repair;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Objects;
 import java.util.Set;
 
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableSet;
 
+import org.apache.cassandra.cache.IMeasurableMemory;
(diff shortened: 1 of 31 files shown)


src/java/org/apache/cassandra/repair/CommonRange.java
```diff
@@ -21,21 +21,30 @@
...
+import org.apache.cassandra.utils.FBUtilities;
+import org.apache.cassandra.utils.ObjectSizes;
 
 /**
  * Groups ranges with identical endpoints/transient endpoints
  */
-public class CommonRange
+public class CommonRange implements IMeasurableMemory
 {
+    // REVIEW: I've heard from some that these should be on a single line, others prefer them split out into separate lines
```
dcapwell, line 44: https://cassandra.apache.org/_/development/code_style.html#multiline-statements

> Where possible prefer keeping a logical action to a single line. Prefer introducing additional variables, or well-named methods encapsulating actions, to multi-line statements - unless this harms clarity (e.g. in an already short method).
> Try to keep lines under 120 characters, but use good judgment. It is better to exceed this limit, than to split a line that has no natural splitting points, particularly when the remainder of the line is boilerplate or easily inferred by the reader.

PR #2807, opened 2023-10-17
CEP-21 review

blerer:
The goal of this PR is to allow to collect the review/comments for the CEP-21 branch while allowing the addition of javadoc or minor code changes by the reviewers.
This will allow the reviewers to quickly benefit from the improvements in documentation to improve the understanding of the code.


src/java/org/apache/cassandra/tcm/ClusterMetadata.java
```diff
@@ -0,0 +1,986 @@
...
+            // Process extension first as a) these are actually mutable and b) they are added to the set of
+            // modified keys when added/updated/removed
+            for (MetadataKey key : modifiedKeys)
+            {
+                ExtensionValue<?> mutable = extensions.get(key);
+                if (null != mutable)
+                    mutable.withLastModified(epoch);
+            }
+
+            if (schema != base.schema)
```
jacek-lewandowski, line 572: why do we compare all those things by references instead of values (like dumpDiff does)?

src/java/org/apache/cassandra/tcm/membership/Directory.java
```diff
@@ -0,0 +1,749 @@
...
+import org.apache.cassandra.utils.Pair;
+import org.apache.cassandra.utils.UUIDSerializer;
+import org.apache.cassandra.utils.btree.BTreeBiMap;
+import org.apache.cassandra.utils.btree.BTreeMap;
+import org.apache.cassandra.utils.btree.BTreeMultimap;
+
+import static org.apache.cassandra.db.TypeSizes.sizeof;
+import static org.apache.cassandra.tcm.membership.NodeVersion.CURRENT;
+
+public class Directory implements MetadataValue<Directory>
```
jacek-lewandowski, line 55: consider renaming this class - maybe `NodeDirectory`? or `NodeIndex`? or `Nodes`?

src/java/org/apache/cassandra/schema/DistributedSchema.java
```diff
@@ -18,38 +18,261 @@
...
+            keyspaces = keyspaces.with(DistributedMetadataLogKeyspace.metadata());
+        return new DistributedSchema(keyspaces, Epoch.UPGRADE_GOSSIP);
+    }
+
+    public void initializeKeyspaceInstances(DistributedSchema prev)
+    {
+        initializeKeyspaceInstances(prev, true);
+    }
+
+    public void initializeKeyspaceInstances(DistributedSchema prev, boolean loadSSTables)
```
jacek-lewandowski, line 118: I think that managing `Keyspace` instances does not belong to this class - this is immutable object and as such should not modify itself. Maybe this could be moved back to `Schema`?

src/java/org/apache/cassandra/tcm/Period.java
```diff
@@ -0,0 +1,211 @@
...
+import org.apache.cassandra.schema.ColumnMetadata;
+import org.apache.cassandra.schema.TableMetadata;
+import org.apache.cassandra.utils.ByteBufferUtil;
+import org.apache.cassandra.utils.FBUtilities;
+
+/**
+ * Utility methods and classes to work with {@code Period}.
+ *
+ */
+public class Period
```
jacek-lewandowski, line 48: consider renaming this class, maybe `LogUtils`, `PeriodUtils`, `Periods`?

src/java/org/apache/cassandra/cql3/QueryProcessor.java
```diff
@@ -428,11 +455,16 @@ public static Prepared parseAndPrepare(String query, ClientState clientState, bo
...
         }
-
         // Note: if 2 threads prepare the same query, we'll live so don't bother synchronizing
         CQLStatement statement = raw.prepare(clientState);
         statement.validate(clientState);
 
+        // Set CQL string for AlterSchemaStatement as this is used to serialize the transformation
+        // in the cluster metadata log
+        if (statement instanceof AlterSchemaStatement)
+            ((AlterSchemaStatement)statement).setCql(query);
```
jacek-lewandowski, line 465: would it be possible to serialize schema transformation in more formalized way? I don't know if it is possible to have bound variables in alter schema statement, but if so, or it will be possible in the future, what would happen then?

src/java/org/apache/cassandra/tcm/log/Replication.java
```diff
@@ -0,0 +1,281 @@
...
+import org.apache.cassandra.tcm.Epoch;
+import org.apache.cassandra.tcm.membership.NodeVersion;
+import org.apache.cassandra.tcm.serialization.MetadataSerializer;
+import org.apache.cassandra.tcm.serialization.VerboseMetadataSerializer;
+import org.apache.cassandra.tcm.serialization.Version;
+
+/**
+ * A sorted set of log entries mainly used to replicate log entries within the cluster.
+ */
+public class Replication
```
jacek-lewandowski, line 49: consider renaming this class, some suggestions: LogEntries, LogEntryCollection...

src/java/org/apache/cassandra/tcm/log/SystemKeyspaceStorage.java
```diff
@@ -0,0 +1,168 @@
...
+            String query = String.format("INSERT INTO %s.%s (period, epoch, current_epoch, entry_id, transformation, kind) VALUES (?,?,?,?,?,?)",
+                                         SchemaConstants.SYSTEM_KEYSPACE_NAME, NAME);
+            executeInternal(query, period, entry.epoch.getEpoch(), entry.epoch.getEpoch(),
+                            entry.id.entryId, serializedTransformation, entry.transform.kind().toString());
+            // todo; should probably not flush every time, but it simplifies tests
+            Keyspace.open(SchemaConstants.SYSTEM_KEYSPACE_NAME).getColumnFamilyStore(NAME).forceBlockingFlush(ColumnFamilyStore.FlushReason.INTERNALLY_FORCED);
+        }
+        catch (Throwable t)
+        {
+            logger.error("Could not persist the entry {} proceeding with in-memory commit.", entry, t);
```
jacek-lewandowski, line 89: no feedback to the caller - why? can we ignore the error and continue?

src/java/org/apache/cassandra/schema/TableMetadata.java
```diff
@@ -181,14 +193,15 @@ public enum Kind
 
     // performance hacks; TODO see if all are really necessary
     public final DataResource resource;
+    public TableMetadataRef ref;
```
jacek-lewandowski, line 196: what is the purpose of ref initialized in TableMetadata?

src/java/org/apache/cassandra/db/SystemKeyspace.java
```diff
@@ -459,6 +468,45 @@ private SystemKeyspace()
           + "cfids set<uuid>, "
           + "PRIMARY KEY (parent_id))").build();
 
+    public static final TableMetadata LocalMetadataLog =
```
jacek-lewandowski, line 471: I'm not sure if I understand that correctly - there are four additional tables, but it looks like we need only one:
- we need to be able to store a transformation per period and epoch
- we need to store a snapshot per period
- we need to be able to go through transformations sorted by epochs
- we need to be able to find the latest period and the latest epoch

Given we only have LocalMetadataLog, which stores the ultimate cluster metadata snapshot in a static column, we can probably get rid of the snapshots table;

Old entries from the log should be evicted, so there should be little partitions representing periods. If so, we could fetch all period - min/max epoch tuples by performing a full table scan and memoizing the results. That would eliminate the need for the remaining two tables with sealed periods and the last sealed period. 

Even if the table turns out to be big, making full scan too costly, we can leverage that each sstable contains min/max clustering information in the statistics metadata. By duplicating the period column as the first clustering column, we could quickly select the relevant sstables by only inspecting their metadata?

Even if we need to store the last epoch and period, we could put them into `local` table.

src/java/org/apache/cassandra/schema/DistributedMetadataLogKeyspace.java
```diff
@@ -0,0 +1,283 @@
...
+        }
+        catch (Throwable t)
+        {
+            logger.error("Caught an exception while trying to CAS", t);
+            return false;
+        }
+    }
+
+    @VisibleForTesting
+    public static void truncateLogState()
```
ekaterinadimitrova2, line 165: This method is not used anywhere.

.build/parent-pom-template.xml
```diff
@@ -336,6 +336,11 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
+        <groupId>com.google.guava</groupId>
+        <artifactId>guava-testlib</artifactId>
+        <version>27.0-jre</version>
```
ekaterinadimitrova2, line 342: I suspect this should also b updated to 32.0.1 as guava, at least.

ci/harry_simulation.sh
```diff
@@ -0,0 +1,96 @@
...
+        -javaagent:$current_dir/../build/test/lib/jars/simulator-asm.jar
+        -Xbootclasspath/a:$current_dir/../build/test/lib/jars/simulator-bootstrap.jar
+        -XX:ActiveProcessorCount=4
+        -XX:-TieredCompilation
+        -XX:-BackgroundCompilation
+        -XX:CICompilerCount=1
+        -XX:Tier4CompileThreshold=1000
+        -XX:ReservedCodeCacheSize=256M
+        -Xmx16G
+        -Xmx4G
```
ekaterinadimitrova2, line 68: We should keep either 16 or 4 for Xmx or maybe it was meant to be Xmx and Xms? In our JVM options we recommend both set to 4G

ci/harry_simulation.sh
```diff
@@ -0,0 +1,96 @@
...
+        -Xss384k
+        --add-exports java.base/jdk.internal.misc=ALL-UNNAMED
```
ekaterinadimitrova2, line 70: Have you considered pulling all these add-exports and add-opens from one place for all tests? They are copy-pasted, and only the Java 11 batch, which I suspect will quickly regress when we are updating JDKs.

conf/cassandra.yaml
```diff
@@ -1184,16 +1184,16 @@ range_request_timeout: 10000ms
 # How long the coordinator should wait for writes to complete.
 # Lowest acceptable value is 10 ms.
 # Min unit: ms
-write_request_timeout: 2000ms
+write_request_timeout: 10000ms
```
ekaterinadimitrova2, line 1187: Was this change of default a mistake?
I can create a patch on confirmation to either revert it or update Config and the test suites config files.

conf/cassandra.yaml
```diff
@@ -1184,16 +1184,16 @@ range_request_timeout: 10000ms
...
 # How long the coordinator should wait for counter writes to complete.
 # Lowest acceptable value is 10 ms.
 # Min unit: ms
-counter_write_request_timeout: 5000ms
+counter_write_request_timeout: 1000ms
 # How long a coordinator should continue to retry a CAS operation
```
ekaterinadimitrova2, line 1192: Was this change of default a mistake?
I can create a patch on confirmation to either revert it or update Config and the test suites config files.

conf/cassandra.yaml
```diff
@@ -1184,16 +1184,16 @@ range_request_timeout: 10000ms
...
 # that contends with other proposals for the same row.
 # Lowest acceptable value is 10 ms.
 # Min unit: ms
-cas_contention_timeout: 1000ms
```
ekaterinadimitrova2, line 1196: Was this change of default a mistake?
I can create a patch on confirmation to either revert it or update Config and the test suites config files.

conf/cassandra.yaml
```diff
@@ -2109,4 +2109,4 @@ drop_compact_storage_enabled: false
 # - Do a rolling restart with all nodes starting with NONE. This sheds the extra cost of checking nodes versions and ensures 
 #   a stable cluster. If a node from a previous version was started by accident we won't any longer toggle behaviors as when UPGRADING.
 #
-storage_compatibility_mode: CASSANDRA_4
+storage_compatibility_mode: NONE
```
ekaterinadimitrova2, line 2112: Leaving this comment as a placeholder, as I see there is TODO in config, so we need to figure out this one.

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -348,7 +348,7 @@ public enum CassandraRelevantProperties
     MEMTABLE_OVERHEAD_SIZE("cassandra.memtable.row_overhead_size", "-1"),
     MEMTABLE_SHARD_COUNT("cassandra.memtable.shard.count"),
     MEMTABLE_TRIE_SIZE_LIMIT("cassandra.trie_size_limit_mb"),
-    MIGRATION_DELAY("cassandra.migration_delay_ms", "60000"),
+    METRICS_REPORTER_CONFIG_FILE("cassandra.metricsReporterConfigFile"),
```
ekaterinadimitrova2, line 351: I suspect this one was added back by mistake? We removed the metrics reporter recently.

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -348,7 +348,7 @@ public enum CassandraRelevantProperties
     MEMTABLE_OVERHEAD_SIZE("cassandra.memtable.row_overhead_size", "-1"),
     MEMTABLE_SHARD_COUNT("cassandra.memtable.shard.count"),
     MEMTABLE_TRIE_SIZE_LIMIT("cassandra.trie_size_limit_mb"),
-    MIGRATION_DELAY("cassandra.migration_delay_ms", "60000"),
```
ekaterinadimitrova2, line 351: Shouldn't this be deprecated before removal?

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -450,7 +450,6 @@ public enum CassandraRelevantProperties
     SAI_TEST_BALANCED_TREE_DEBUG_ENABLED("cassandra.sai.test.balanced_tree_debug_enabled", "false"),
     SAI_TEST_DISABLE_TIMEOUT("cassandra.sai.test.disable.timeout", "false"),
 
-    SCHEMA_PULL_INTERVAL_MS("cassandra.schema_pull_interval_ms", "60000"),
```
ekaterinadimitrova2, line 453: Shouldn't this be deprecated before removal?

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -122,7 +122,7 @@ public static Set<String> splitCommaDelimited(String src)
     public DiskFailurePolicy disk_failure_policy = DiskFailurePolicy.ignore;
     public CommitFailurePolicy commit_failure_policy = CommitFailurePolicy.stop;
 
-    public volatile boolean use_deterministic_table_id = false;
+    public volatile boolean use_deterministic_table_id = true;
```
ekaterinadimitrova2, line 125: Was this changed intentionally?

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -171,6 +171,20 @@ public static Set<String> splitCommaDelimited(String src)
 
     public volatile DurationSpec.LongMillisecondsBound stream_transfer_task_timeout = new DurationSpec.LongMillisecondsBound("12h");
 
+    public volatile DurationSpec.LongMillisecondsBound cms_await_timeout = new DurationSpec.LongMillisecondsBound("120000ms");
+    public volatile int cms_default_max_retries = 10;
+    public volatile DurationSpec.IntMillisecondsBound cms_default_retry_backoff = new DurationSpec.IntMillisecondsBound("50ms");
```
ekaterinadimitrova2, line 176: I think we can remove the "default_" part, no? It is a configurable parameter whose default value is set, no? Same for `cms_default_max_retries`

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -171,6 +171,20 @@ public static Set<String> splitCommaDelimited(String src)
 
     public volatile DurationSpec.LongMillisecondsBound stream_transfer_task_timeout = new DurationSpec.LongMillisecondsBound("12h");
 
+    public volatile DurationSpec.LongMillisecondsBound cms_await_timeout = new DurationSpec.LongMillisecondsBound("120000ms");
```
ekaterinadimitrova2, line 174: Out of curiosity, how were default values for all new parameters chosen? Methodology?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -4917,4 +4964,100 @@ public static RepairRetrySpec getRepairRetrySpec()
     {
         return conf == null ? new RepairRetrySpec() : conf.repair.retries;
     }
+
+    public static int getCmsDefaultRetryMaxTries()
+    {
+        return conf.cms_default_max_retries;
+    }
+
+    public static void setCmsDefaultRetryMaxTries(int value)
```
ekaterinadimitrova2, line 4973: This method seems to be not used and the name does not really match the parameter. Should I remove it?

src/java/org/apache/cassandra/tcm/log/SystemKeyspaceStorage.java
```diff
@@ -0,0 +1,168 @@
...
+
+    /**
+     * Checks if the metdata log table contains at least one row.
+     * @return {@code true} if the metdata log table contains at least one row, {@code false} otherwise.
+     */
+    public synchronized static boolean hasAnyEpoch()
+    {
+        String query = String.format("SELECT epoch FROM %s.%s LIMIT 1", SchemaConstants.SYSTEM_KEYSPACE_NAME, NAME);
+
+        for (UntypedResultSet.Row row : executeInternal(query))
```
ekaterinadimitrova2, line 101: executeInternal can return null?

src/java/org/apache/cassandra/tcm/sequences/ProgressBarrier.java
```diff
@@ -0,0 +1,568 @@
...
+ *
+ * If number of nodes in the cluster is smaller than the number of nodes specified in the replication factor, we will
+ * collect only n/2 + 1 nodes to avoid availability issues.
+ */
+public class ProgressBarrier
+{
+    private static final Logger logger = LoggerFactory.getLogger(ProgressBarrier.class);
+    private static final ConsistencyLevel MIN_CL = DatabaseDescriptor.getProgressBarrierMinConsistencyLevel();
+    private static final ConsistencyLevel DEFAULT_CL = DatabaseDescriptor.getProgressBarrierDefaultConsistencyLevel();
+    private static final long TIMEOUT_MILLIS = DatabaseDescriptor.getProgressBarrierTimeout(TimeUnit.MILLISECONDS);
```
ekaterinadimitrova2, line 78: How was 1 hour chosen?

src/java/org/apache/cassandra/tcm/transformations/CustomTransformation.java
```diff
@@ -0,0 +1,264 @@
...
+import org.apache.cassandra.tcm.Transformation;
+import org.apache.cassandra.tcm.extensions.ExtensionKey;
+import org.apache.cassandra.tcm.extensions.IntValue;
+import org.apache.cassandra.tcm.extensions.StringValue;
+import org.apache.cassandra.tcm.serialization.AsymmetricMetadataSerializer;
+import org.apache.cassandra.tcm.serialization.Version;
+import org.apache.cassandra.tcm.ClusterMetadata;
+import org.apache.cassandra.tcm.sequences.LockedRanges;
+
+public class CustomTransformation implements Transformation
```
ekaterinadimitrova2, line 35: Can you add a bit more information about this one, please?

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -511,8 +511,9 @@ public enum CassandraRelevantProperties
      */
     TCM_PROGRESS_BARRIER_BACKOFF_MILLIS("cassandra.progress_barrier_backoff_ms", "1000"),
     TCM_PROGRESS_BARRIER_TIMEOUT_MILLIS("cassandra.progress_barrier_timeout_ms", "3600000"),
+
     /**
-     * size of in-memory index of max epoch -> sealed period
+     * Maximum sized of the {@code RecentlySealedPeriods} in-memory index.
```
ifesdjeen: nit: sized -> size

src/java/org/apache/cassandra/config/Config.java
ifesdjeen: I think `default_` predicate makes sense here. I did think of calling it `start_` or `initial_` but it seems to be more confusing. We use the same term for indicating the protocol version nodes start negotiating by default.

src/java/org/apache/cassandra/schema/SchemaTransformation.java
```diff
@@ -37,7 +37,7 @@ public interface SchemaTransformation
     /**
      * Apply a statement transformation to a schema snapshot.
      * <p>
-     * Implementing methods should be side-effect free (outside of throwing exceptions if the transformation cannot
+     * Implementing methods should be side effect free (outside of throwing exceptions if the transformation cannot
```
ifesdjeen, line 40: `side-effect` is also a correct spelling: https://www.collinsdictionary.com/dictionary/english/side-effect.

I would say `side-effect` but `side-effect free`.

src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
```diff
@@ -125,6 +125,10 @@ public static ClusterMetadataService instance()
 
     private final AtomicBoolean commitsPaused = new AtomicBoolean();
 
+    /**
+     * Returns the state of the {@code ClusteMetadataService}.
```
ifesdjeen: This comment seems to be tautological. Maybe add "returns _current_ state" at least? I would say it could be useful to specify what gets returned, too. Ie whether or not current cluster metadata service is local (ie member of CMS), remote (ie a non-CMS member of the cluster), gossip (ie a node that has not yet been upgraded to TCM), or the node is forced to reset the local metadata.

blerer: I think that you answer the question that I was wondering. It was unclear to me what the different states means.
How does work the reset of the local metadata? 
As it is based on a property does it means that the node will stay on a reset state until it is restarted?

src/java/org/apache/cassandra/config/Config.java
krummas: temporarily yes, we'll post a patch that fixes deterministic table ids soon

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
krummas: it has no effect anymore, and for users setting it cassandra will still start without issue

src/java/org/apache/cassandra/config/Config.java
ekaterinadimitrova2: A quick look shows me we have diverged from using `initial_` to `default_` in the past year in the `Config` class. Now, we are inconsistent. The precedent was not created in this patch; sadly, as it could be confusing to people, it happened. I leave it up to you; naming is hard in general.

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
ekaterinadimitrova2: I am with you on this; it was just something people do in the community for one major release, even if the property has no effect. But on second thought, that makes more sense for the Config class than here. In this class, the Deprecated parameters can still be used. I just checked them.

.build/parent-pom-template.xml
jacek-lewandowski: yeah, perhaps use some internal property for guava version

src/java/org/apache/cassandra/tcm/ClusterMetadata.java
```diff
@@ -0,0 +1,986 @@
...
+ * @see MetadataSnapshots for more information about cluster metadata snapshots
+ */
+public class ClusterMetadata
+{
+    public static final Serializer serializer = new Serializer();
+
+    public final Epoch epoch;
+    public final long period;
+    public final boolean lastInPeriod;
+    public final IPartitioner partitioner;       // Set during (initial) construction and not modifiable via Transformer
```
jacek-lewandowski, line 104: so is it there just for validation that all nodes has the same partitioner?

src/java/org/apache/cassandra/tcm/listeners/SchemaListener.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.cassandra.db.SystemKeyspace;
+import org.apache.cassandra.gms.ApplicationState;
+import org.apache.cassandra.gms.Gossiper;
+import org.apache.cassandra.schema.DistributedSchema;
+import org.apache.cassandra.schema.Schema;
+import org.apache.cassandra.schema.SchemaDiagnostics;
+import org.apache.cassandra.service.StorageService;
+import org.apache.cassandra.tcm.ClusterMetadata;
+
+public class SchemaListener implements ChangeListener
```
jacek-lewandowski, line 30: Consider renaming this class to something more like `AlterSchemaListener`, `AlterSchemaTransformationLister`, `SchemaTransformationListener` to not confuse it with `SchemaChangeListener`

src/java/org/apache/cassandra/tcm/membership/NodeId.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    public static boolean isValidNodeId(UUID uuid)
+    {
+        return uuid.getMostSignificantBits() == 0 &&
+               uuid.getLeastSignificantBits() < Integer.MAX_VALUE;
+    }
+
+    @Deprecated
+    public UUID toUUID()
```
ekaterinadimitrova2, line 65: Brand new method, deprecated?

src/java/org/apache/cassandra/schema/DistributedSchema.java
blerer: Yes, this part confused me too.

src/java/org/apache/cassandra/tcm/extensions/AbstractExtensionValue.java
```diff
@@ -0,0 +1,80 @@
...
+public abstract class AbstractExtensionValue<V> implements ExtensionValue<V>
+{
+    private Epoch lastModified = Epoch.EMPTY;
+    private V value = null;
+
+    @Override
+    public V withLastModified(Epoch lastModified)
+    {
+        this.lastModified = lastModified;
+        return (V)this;
```
blerer, line 37: This method is broken for its sub-classes. For example `IntValue` will attempt to cast an `IntValue` object into an `Integer`.
There is also a change in the class expectation outside of `ExtensionValue` classes all others `MetadataValue` seems to be immutable. I think that it would make sense to keep that immutability everywhere.

src/java/org/apache/cassandra/tcm/log/Replication.java
ekaterinadimitrova2: Further ideas `ReplicationLogEntries`, `ReplicationSet`

src/java/org/apache/cassandra/tcm/transformations/Register.java
```diff
@@ -0,0 +1,222 @@
...
+        else
+        {
+            throw new IllegalStateException(String.format("A node with address %s already exists, cancelling join. Use cassandra.replace_address if you want to replace this node.", nodeAddresses.broadcastAddress));
+        }
+
+        logger.info("Registering with endpoint {}", nodeAddresses.broadcastAddress);
+        return nodeId;
+    }
+
+    private static NodeId register(boolean force)
```
ekaterinadimitrova2, line 131: `force` is always `false`?

test/distributed/org/apache/cassandra/distributed/test/log/RegisterTest.java
```diff
@@ -0,0 +1,169 @@
...
+                    throw new RuntimeException(e);
+                }
+                ClusterMetadataService.instance().commit(SealPeriod.instance);
+            });
+
+            cluster.get(2).runOnInstance(() -> {
+                try
+                {
+                    Field field = NodeVersion.class.getDeclaredField("CURRENT");
+                    Field modifiers = Field.class.getDeclaredField("modifiers");
```
ekaterinadimitrova2, line 120: This will not work with JDK 17 because of JDK-8210522 (rewrite reflection of "modifiers" field)
We fixed other in-jvm tests in [CASSANDRA-18181](https://issues.apache.org/jira/browse/CASSANDRA-18181)

src/java/org/apache/cassandra/tcm/log/SystemKeyspaceStorage.java
jacek-lewandowski: I suppose it returns null when the statement does not return rows (for example, a schema update, or grant/revoke)

src/java/org/apache/cassandra/cql3/statements/schema/AlterKeyspaceStatement.java
```diff
@@ -128,14 +142,21 @@ private void validateNoRangeMovements()
...
 
-        Stream<InetAddressAndPort> unreachableNotAdministrativelyInactive =
-            Gossiper.instance.getUnreachableMembers().stream().filter(endpoint -> !FBUtilities.getBroadcastAddressAndPort().equals(endpoint) &&
-                                                                      !Gossiper.instance.isAdministrativelyInactiveState(endpoint));
-        Stream<InetAddressAndPort> endpoints = Stream.concat(Gossiper.instance.getLiveMembers().stream(),
-                                                             unreachableNotAdministrativelyInactive);
-        List<InetAddressAndPort> notNormalEndpoints = endpoints.filter(endpoint -> !FBUtilities.getBroadcastAddressAndPort().equals(endpoint) &&
-                                                                                   !Gossiper.instance.getEndpointStateForEndpoint(endpoint).isNormalState())
-                                                               .collect(Collectors.toList());
+        ClusterMetadata metadata = ClusterMetadata.current();
```
jacek-lewandowski, line 145: Referring `ClusterMetadata.current()` from the transformation code is wrong, so as referring to the configuration, `DatabaseDescriptor`, `Gossiper` and others. This can be seen in many schema transformations and makes the transformations non-pure. In turn, it can make replaying the log lead to inconsistent metadata for the same epoch on different nodes.

src/java/org/apache/cassandra/tcm/membership/NodeId.java
```diff
@@ -0,0 +1,120 @@
...
+    {
+        if (!isValidNodeId(uuid))
+            throw new UnsupportedOperationException("Not a node id: " + uuid); // see RemoveTest#testBadHostId
+        return new NodeId(Ints.checkedCast(uuid.getLeastSignificantBits()));
+    }
+
+    public static boolean isValidNodeId(UUID uuid)
+    {
+        return uuid.getMostSignificantBits() == 0 &&
+               uuid.getLeastSignificantBits() < Integer.MAX_VALUE;
```
blerer, line 61: It seems that `uuid.getLeastSignificantBits()` can return a negative number that will not fit in an integer but will be smaller than `Integer.MAX_VALUE`.

src/java/org/apache/cassandra/tcm/membership/NodeId.java
```diff
@@ -0,0 +1,120 @@
...
+import org.apache.cassandra.io.util.DataInputPlus;
+import org.apache.cassandra.io.util.DataOutputPlus;
+import org.apache.cassandra.tcm.serialization.MetadataSerializer;
+import org.apache.cassandra.tcm.serialization.Version;
+
+public class NodeId implements Comparable<NodeId>
+{
+    public static final Serializer serializer = new Serializer();
+
+    private final int id;
```
blerer, line 37: Looking at the code it is unclear what are the valid values for id an if it should be considered as a signed or unsigned number.

src/java/org/apache/cassandra/tcm/transformations/AlterSchema.java
```diff
@@ -0,0 +1,280 @@
...
+            // Guard against an invalid SchemaTransformation supplying a TableMetadata with a future epoch
+            newKeyspaces = schemaTransformation.apply(prev);
+            newKeyspaces.forEach(ksm -> {
+               ksm.tables.forEach(tm -> {
+                   if (tm.epoch.isAfter(prev.nextEpoch()))
+                       throw new InvalidRequestException(String.format("Invalid schema transformation. " +
+                                                                       "Resultant epoch for table metadata of %s.%s (%d) " +
+                                                                       "is greater than for cluster metadata (%d)",
+                                                                       ksm.name, tm.name, tm.epoch.getEpoch(),
+                                                                       prev.nextEpoch().getEpoch()));
```
blerer, line 108: `InvalidRequestException` has always been used for error that were linked to a user request that was trying to do something that C* was not supporting (e.g. ordering by non-clustering columns). In the current patch, there are multiple places where we use that exception for errors on which the user has no control (e.g. some race conditions or internal errors). I believe that we need to change that because it can be really confusing from the user point of view.

src/java/org/apache/cassandra/tcm/log/SystemKeyspaceStorage.java
ekaterinadimitrova2: We need to take precautions with this loop.

src/java/org/apache/cassandra/locator/InetAddressAndPort.java
```diff
@@ -316,6 +328,26 @@ static int getDefaultPort()
         return defaultPort;
     }
 
+    public static final class MetadataSerializer implements org.apache.cassandra.tcm.serialization.MetadataSerializer<InetAddressAndPort>
+    {
+        public static final MetadataSerializer serializer = new MetadataSerializer();
```
ekaterinadimitrova2, line 333:
```suggestion
        public static final MetadataSerializer SERIALIZER = new MetadataSerializer();
```

src/java/org/apache/cassandra/db/SystemKeyspace.java
```diff
@@ -1926,4 +1969,109 @@ public static TopPartitionTracker.StoredTopPartitions getTopPartitions(TableMeta
...
+    /**
+     * Retrieves the cluster metadata snapshot for the specified epoch from the {@code metadata_snapshot} table.
+     *
+     * @param epoch the epoch for which the snapshot must be retrieved
+     * @return the cluster metadata snapshot for the specified epoch or {@code null} if no snapshot exists for the epoch.
+     */
+    public static ByteBuffer getSnapshot(Epoch epoch)
+    {
+        logger.info("Getting snapshot of epoch = {}", epoch);
+        String query = String.format("SELECT SNAPSHOT FROM %s.%s WHERE epoch = ?", SchemaConstants.SYSTEM_KEYSPACE_NAME, SNAPSHOT_TABLE_NAME);
```
ekaterinadimitrova2, line 1996:
```suggestion
        String query = String.format("SELECT snapshot FROM %s.%s WHERE epoch = ?", SchemaConstants.SYSTEM_KEYSPACE_NAME, SNAPSHOT_TABLE_NAME);
```

src/java/org/apache/cassandra/schema/ReplicationParams.java
```diff
@@ -17,21 +17,40 @@
...
+import org.apache.cassandra.utils.FBUtilities;
+
+import static org.apache.cassandra.db.TypeSizes.sizeof;
 
 public final class ReplicationParams
 {
+    private static final ReplicationParams META = new ReplicationParams(MetaStrategy.class, ImmutableMap.of());
+
+    public static final Serializer serializer = new Serializer();
+    public static final MessageSerializer messageSerializer = new MessageSerializer();
```
ekaterinadimitrova2, line 52:
```suggestion
    public static final Serializer SERIALIZER = new Serializer();
    public static final MessageSerializer MESSAGE_SERIALIZER = new MessageSerializer();
```

src/java/org/apache/cassandra/schema/Schema.java
```diff
@@ -75,235 +57,126 @@
  * This class should be the only entity used to query and manage schema. Internal details should not be access in
  * production code (would be great if they were not accessed in the test code as well).
  * <p>
- * TL;DR: All modifications are made using the implementation of {@link SchemaUpdateHandler} obtained from the provided
+ * TL;DR: All modifications are made using the implementation of SchemaUpdateHandle obtained from the provided
```
ekaterinadimitrova2, line 60: This doc needs update; To begin with - there are no `SchemaUpdateHandle` and `mergeAndUpdateVersion` or do I miss anything?

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -67,7 +67,7 @@
  * system_schema.* tables and methods for manipulating them.
  *
  * Please notice this class is _not_ thread safe and all methods which reads or updates the data in schema keyspace
- * should be accessed only from the implementation of {@link SchemaUpdateHandler} in synchronized blocks.
+ * should be accessed only from the implementation of {SchemaUpdateHandler} in synchronized blocks.
```
ekaterinadimitrova2, line 70: It seems Java doc around Schema needs a re-write at more places. There is no `SchemaUpdateHandler` anymore

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -287,7 +287,7 @@ public static KeyspaceMetadata metadata()
         return KeyspaceMetadata.create(SchemaConstants.SCHEMA_KEYSPACE_NAME, KeyspaceParams.local(), org.apache.cassandra.schema.Tables.of(ALL_TABLE_METADATA));
     }
 
-    static Collection<Mutation> convertSchemaDiffToMutations(KeyspacesDiff diff, long timestamp)
+    public static Collection<Mutation> convertSchemaDiffToMutations(KeyspacesDiff diff, long timestamp)
```
ekaterinadimitrova2, line 290: Does this really needs to be public?

src/java/org/apache/cassandra/schema/SchemaPullVerbHandler.java
```diff
@@ -32,6 +32,7 @@
  * Sends it's current schema state in form of mutations in response to the remote node's request.
  * Such a request is made when one of the nodes, by means of Gossip, detects schema disagreement in the ring.
  */
+@Deprecated
```
ekaterinadimitrova2, line 35: This will explode on rebase because of [CASSANDRA-18912](https://issues.apache.org/jira/browse/CASSANDRA-18912) We need to add `since` and maybe even a bit more info like why it was deprecated and when it is for removal (if ever)

src/java/org/apache/cassandra/schema/SchemaPushVerbHandler.java
```diff
@@ -35,6 +35,7 @@
  * (which is going to act as coordinator) and that node sends (pushes) it's updated schema state
  * (in form of mutations) to all the alive nodes in the cluster.
  */
+@Deprecated
```
ekaterinadimitrova2, line 38: This will explode on rebase because of [CASSANDRA-18912](https://issues.apache.org/jira/browse/CASSANDRA-18912) We need to add `since` and maybe even a bit more info like why it was deprecated and when it is for removal (if ever)

src/java/org/apache/cassandra/schema/SchemaTransformation.java
```diff
@@ -17,20 +17,76 @@
...
+     *   cluster metadata for the peer.
+     *   </li>
+     * </ul>
+     * At the moment, not all SchemaTransformation are entirely free of side effects, because both client warnings and
+     * guardrails are tightly coupled to the ST implementations and they do produce side effects. Ideally, ClientWarn
+     * and Guardrails can be reworked to decouple them from ClientState and threadlocals. In the meantime, we can hint
+     * to them that SchemaTransformation::apply is being called in the context of AlterSchema::execute and so they
+     * should not produce their side effects. The default impl is a no-op, but AlterSchemaStatement which has access to
+     * a ClientState, is able to pause both warnings and guardrails.
+     * @see org.apache.cassandra.cql3.statements.schema.AlterSchemaStatement#enterExecution()
```
ekaterinadimitrova2, line 76: What are the side effects you mean here, except duplicate warnings?
I saw the `org.apache.cassandra.cql3.statements.schema.AlterSchemaStatement#enterExecution()` method.

src/java/org/apache/cassandra/schema/SchemaTransformations.java
```diff
@@ -31,6 +35,14 @@
  */
 public class SchemaTransformations
 {
+    public static SchemaTransformation fromCql(String cql)
```
ekaterinadimitrova2, line 38: Not used anwyhere

src/java/org/apache/cassandra/schema/SchemaTransformations.java
```diff
@@ -128,8 +143,9 @@ public static SchemaTransformation addTypes(Types toAdd, boolean ignoreIfExists)
      */
     public static SchemaTransformation addView(ViewMetadata view, boolean ignoreIfExists)
```
ekaterinadimitrova2, line 144: Not used anywhere

src/java/org/apache/cassandra/schema/SchemaUpdateHandler.java
```diff
@@ -1,78 +0,0 @@
...
-import org.apache.cassandra.utils.concurrent.Awaitable;
-
-/**
- * Schema update handler is responsible for maintaining the shared schema and synchronizing it with other nodes in
- * the cluster, which means pushing and pulling changes, as well as tracking the current version in the cluster.
- * <p/>
- * The interface has been extracted to abstract out that functionality. It allows for various implementations like
- * Gossip based (the default), ETCD, offline, etc., and make it easier for mocking in unit tests.
- */
-public interface SchemaUpdateHandler
```
ekaterinadimitrova2, line 33: Isn't removal of an interface a breaking change?

src/java/org/apache/cassandra/schema/SystemDistributedKeyspace.java
```diff
@@ -228,10 +229,6 @@ public static void successfulParentRepair(TimeUUID parent_id, Collection<Range<T
 
     public static void startRepairs(TimeUUID id, TimeUUID parent_id, String keyspaceName, String[] cfnames, CommonRange commonRange)
     {
-        // Don't record repair history if an upgrade is in progress as version 3 nodes generates errors
-        // due to schema differences
-        boolean includeNewColumns = !Gossiper.instance.hasMajorVersion3Nodes();
```
ekaterinadimitrova2, line 233: This was leftover from 4? We do not support upgrades from 3 to 5.

src/java/org/apache/cassandra/schema/TableMetadata.java
```diff
@@ -59,25 +61,34 @@
...
 import static java.lang.String.format;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
+import static org.apache.cassandra.db.TypeSizes.sizeof;
 import static org.apache.cassandra.schema.IndexMetadata.isNameValid;
 
 @Unmetered
 public class TableMetadata implements SchemaElement
 {
+    public static final Serializer serializer = new Serializer();
```
ekaterinadimitrova2, line 90:
```suggestion
    public static final Serializer SERIALIZER = new Serializer();
```

src/java/org/apache/cassandra/schema/TableMetadataRef.java
```diff
@@ -15,33 +15,62 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.apache.cassandra.schema;
 
-import org.github.jamm.Unmetered;
```
ekaterinadimitrova2, line 20: Why `@Unmetered` was removed?

src/java/org/apache/cassandra/tcm/AtomicLongBackedProcessor.java
```diff
@@ -0,0 +1,182 @@
...
+                else
+                    return mid; // key found
+            }
+            return -(low + 1);  // key not found
+        }
+    }
+
+    public static class InMemoryMetadataSnapshots implements MetadataSnapshots
+    {
+        private SortedMap<Epoch, ClusterMetadata> snapshots;
```
ekaterinadimitrova2, line 152:
```suggestion
        private final SortedMap<Epoch, ClusterMetadata> snapshots;
```

src/java/org/apache/cassandra/tcm/Commit.java
```diff
@@ -0,0 +1,429 @@
...
+        {
+            return TypeSizes.sizeof(serializationVersion.asInt()) +
+                   Transformation.serializer.serializedSize(t.transform, serializationVersion) +
+                   Entry.Id.serializer.serializedSize(t.entryId, serializationVersion) +
+                   Epoch.serializer.serializedSize(t.lastKnown, serializationVersion);
+        }
+    }
+
+    static volatile Result.Serializer resultSerializerCache;
+    public static interface Result
```
ekaterinadimitrova2, line 124:
```suggestion
    public interface Result
```

src/java/org/apache/cassandra/tcm/Discovery.java
```diff
@@ -0,0 +1,249 @@
...
+
+    /**
+     * The states of the discovery process.
+     */
+    private enum State
+    {
+        NOT_STARTED,
+        IN_PROGRESS,
+        FINISHED,
+        FOUND_CMS
```
blerer, line 247: `FOUND_CMS` is not used anymore.

src/java/org/apache/cassandra/tcm/Discovery.java
```diff
@@ -0,0 +1,249 @@
...
+     * known CMS members. If a seed node knows about CMS members, it responds with a collection of only those nodes.
+     * Otherwise, the seed node responds with a collection of all known peers. They are accumulated as new target
+     * nodes to query for CMS members in the next round. The process completes when either:
+     * - a response containing the CMS members is received
+     * - a deadline or a specified maximum number of rounds is reached
+     */
+    public DiscoveredNodes discover(int rounds)
+    {
+        boolean res = state.compareAndSet(State.NOT_STARTED, State.IN_PROGRESS);
+        assert res : String.format("Can not start discovery as it is in state %s", state.get());
```
blerer, line 81: The code relies on this assertion to behave correctly. Should it not be a real check?

PR #2813, opened 2023-10-17
Remove completed coordinator sessions

aratno:
CASSANDRA-18903

### DIFF
--- a/src/java/org/apache/cassandra/repair/consistent/ConsistentSession.java
+++ b/src/java/org/apache/cassandra/repair/consistent/ConsistentSession.java
@@ -206,6 +206,12 @@ public abstract class ConsistentSession
         this.participants = ImmutableSet.copyOf(builder.participants);
     }
 
+    public boolean isCompleted()
+    {
+        State s = getState();
+        return s == State.FINALIZED || s == State.FAILED;
+    }
+
     public State getState()
     {
         return state;
--- a/src/java/org/apache/cassandra/repair/consistent/CoordinatorSession.java
+++ b/src/java/org/apache/cassandra/repair/consistent/CoordinatorSession.java
@@ -87,6 +87,7 @@ public class CoordinatorSession extends ConsistentSession
     public static class Builder extends AbstractBuilder
     {
         Consumer<CoordinatorSession> listener;
+
         public void withListener(Consumer<CoordinatorSession> listener)
         {
             this.listener = listener;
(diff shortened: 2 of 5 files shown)


src/java/org/apache/cassandra/repair/consistent/CoordinatorSession.java
```diff
@@ -82,6 +86,12 @@ public CoordinatorSession(Builder builder)
 
     public static class Builder extends AbstractBuilder
     {
+        Consumer<CoordinatorSession> listener;
```
maedhroz, line 89: nit: throw a newline in here, like the 4.1 patch?

src/java/org/apache/cassandra/repair/consistent/ConsistentSession.java
```diff
@@ -206,6 +206,12 @@ public static State valueOf(int ordinal)
         this.participants = ImmutableSet.copyOf(builder.participants);
     }
 
+    public boolean isCompleted()
```
maedhroz, line 209: Do we need to remove the `isCompleted()` in `LocalSession`?

test/distributed/org/apache/cassandra/distributed/test/IncRepairCoordinatorErrorTest.java
```diff
@@ -51,7 +51,7 @@ public void errorTest() throws Exception
             cluster.get(3).runOnInstance(() -> {
                 ActiveRepairService.instance.failSession(result.toString(), true);
             });
-            assertThat(cluster.get(1).logs().watchFor("Can't transition endpoints .* to FAILED").getResult()).isNotEmpty();
+            assertThat(cluster.get(1).logs().watchFor("Removing completed session .* with state FINALIZED").getResult()).isNotEmpty();
```
maedhroz, line 54: nit: Any way we could just call into node 1 and get the number of sessions via `ARS` rather than having to check the logs? If that's brutal, no biggie, but I don't even thing we'd have to change the visibility of anything, and we already have the `UUID`?

maedhroz (approved):

test/distributed/org/apache/cassandra/distributed/test/IncRepairCoordinatorErrorTest.java
aratno: Talked about this with Caleb - I actually prefer the logs check because it's more explicit than "nothing found in map" - it confirms that the specific action (removal from `ARS.consistent.coordinated`) was taken, rather than the more generic "no data found".

PR #2824, opened 2023-10-18
CASSANDRA-18941: Support max SSTable size in sorted CQLSSTableWriter

yifan-c:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
@@ -69,6 +69,8 @@ import org.apache.cassandra.tcm.ClusterMetadataService;
 import org.apache.cassandra.transport.ProtocolVersion;
 import org.apache.cassandra.utils.ByteBufferUtil;
 import org.apache.cassandra.utils.JavaDriverUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import static org.apache.cassandra.utils.Clock.Global.currentTimeMillis;
 
@@ -376,17 +378,19 @@ public class CQLSSTableWriter implements Closeable
      */
     public static class Builder
     {
-        private File directory;
+        private static final Logger logger = LoggerFactory.getLogger(Builder.class);
+        private static final long DEFAULT_BUFFER_SIZE_IN_MIB_FOR_UNSORTED = 128L;
 
         protected SSTableFormat<?, ?> format = null;
 
-        private CreateTableStatement.Raw schemaStatement;
         private final List<CreateTypeStatement.Raw> typeStatements;
+
+        private File directory;
+        private CreateTableStatement.Raw schemaStatement;
         private ModificationStatement.Parsed modificationStatement;
         private IPartitioner partitioner;
-
         private boolean sorted = false;
-        private long bufferSizeInMiB = 128;
+        private long maxSSTableSizeInMiB = -1L;
 
         protected Builder() {
             this.typeStatements = new ArrayList<>();
@@ -495,21 +499,21 @@ public class CQLSSTableWriter implements Closeable
         }
 
         /**
-         * The size of the buffer to use.
-         * <p>
-         * This defines how much data will be buffered before being written as
-         * a new SSTable. This correspond roughly to the data size that will have the created
-         * sstable.
-         * <p>
-         * The default is 128MiB, which should be reasonable for a 1GiB heap. If you experience
-         * OOM while using the writer, you should lower this value.
+         * Defines the maximum SSTable size in mebibytes when using the sorted writer.
+         * By default, i.e. not specified, there is no maximum size limit for the produced SSTable
          *
-         * @param size the size to use in MiB.
-         * @return this builder.
+         * @param size the maximum sizein mebibytes of each individual SSTable allowed
+         * @return this builder
          */
-        public Builder withBufferSizeInMiB(int size)
+        public Builder withMaxSSTableSizeInMiB(int size)
         {
-            this.bufferSizeInMiB = size;
+            if (size <= 0)
+            {
+                logger.warn("A non-positive value for maximum SSTable size is specified, " +
+                            "which disables the size limiting effectively. Please supply a positive value in order " +
+                            "to enforce size limiting for the produced SSTables.");
+            }
+            this.maxSSTableSizeInMiB = size;
             return this;
         }
 
@@ -517,7 +521,27 @@ public class CQLSSTableWriter implements Closeable
          * The size of the buffer to use.
          * <p>
          * This defines how much data will be buffered before being written as
-         * a new SSTable. This correspond roughly to the data size that will have the created
+         * a new SSTable. This corresponds roughly to the data size that will have the created
+         * sstable.
+         * <p>
+         * The default is 128MiB, which should be reasonable for a 1GiB heap. If you experience
+         * OOM while using the writer, you should lower this value.
+         *
+         * @deprecated This method is deprecated in favor of the new withMaxSSTableSizeInMiB(int size)
+         * @param size the size to use in MiB.
+         * @return this builder.
+         */
+        @Deprecated(since = "5.1")
+        public Builder withBufferSizeInMiB(int size)
+        {
+            return withMaxSSTableSizeInMiB(size);
+        }
+
+        /**
+         * The size of the buffer to use.
+         * <p>
+         * This defines how much data will be buffered before being written as
+         * a new SSTable. This corresponds roughly to the data size that will have the created
          * sstable.
          * <p>
          * The default is 128MiB, which should be reasonable for a 1GiB heap. If you experience
@@ -568,6 +592,13 @@ public class CQLSSTableWriter implements Closeable
             Preconditions.checkState(Sets.difference(SchemaConstants.LOCAL_SYSTEM_KEYSPACE_NAMES, Schema.instance.getKeyspaces()).isEmpty(),
                                      "Local keyspaces were not loaded. If this is running as a client, please make sure to add %s=true system property.",
                                      CassandraRelevantProperties.FORCE_LOAD_LOCAL_KEYSPACES.getKey());
+
+            // Assign the default max SSTable size if not defined in builder
+            if (isMaxSSTableSizeUnset())
+            {
+                maxSSTableSizeInMiB = sorted ? -1L : DEFAULT_BUFFER_SIZE_IN_MIB_FOR_UNSORTED;
+            }
+
             synchronized (CQLSSTableWriter.class)
             {
                 String keyspaceName = schemaStatement.keyspace();
@@ -599,14 +630,19 @@ public class CQLSSTableWriter implements Closeable
 
                 TableMetadataRef ref = tableMetadata.ref;
                 AbstractSSTableSimpleWriter writer = sorted
-                                                     ? new SSTableSimpleWriter(directory, ref, preparedModificationStatement.updatedColumns())
-                                                     : new SSTableSimpleUnsortedWriter(directory, ref, preparedModificationStatement.updatedColumns(), bufferSizeInMiB);
+                                                     ? new SSTableSimpleWriter(directory, ref, preparedModificationStatement.updatedColumns(), maxSSTableSizeInMiB)
+                                                     : new SSTableSimpleUnsortedWriter(directory, ref, preparedModificationStatement.updatedColumns(), maxSSTableSizeInMiB);
 
                 if (format != null)
                     writer.setSSTableFormatType(format);
 
                 return new CQLSSTableWriter(writer, preparedModificationStatement, preparedModificationStatement.getBindVariables());
             }
+        }
+
+        private boolean isMaxSSTableSizeUnset()
+        {
+            return maxSSTableSizeInMiB <= 0;
         }
 
         private Types createTypes(String keyspace)
--- a/src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
@@ -392,7 +392,8 @@ public class CQLSSTableWriter implements Closeable
         private boolean sorted = false;
         private long maxSSTableSizeInMiB = -1L;
 
-        protected Builder() {
+        protected Builder()
+        {
             this.typeStatements = new ArrayList<>();
         }
 
--- a/src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
+++ b/src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
@@ -19,6 +19,7 @@ package org.apache.cassandra.io.sstable;
 
 import java.io.IOException;
 
+import com.google.common.base.Preconditions;
 import com.google.common.base.Throwables;
 
 import org.apache.cassandra.db.DecoratedKey;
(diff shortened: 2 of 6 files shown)


src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
```diff
@@ -554,6 +555,18 @@ public Builder sorted()
...
 
+        /**
+         * Defines the maximum SSTable size when using the sorted writer.
+         * By default, i.e. not specified, there is no maximum size limit for the produced SSTable
+         * @param size the maximum size of each individual SSTable allowed
+         * @return this builder
+         */
+        public Builder withMaxSSTableSizeInMiBForSorted(int size)
+        {
+            this.maxSSTableSizeInMiBForSorted = size;
```
Maxwell-Guo: we should do a validation for size , it should be positive, if a negative number is set , then we should remind users explicitly.
If not set then only one sstable will generate, and it's meaningless to set a negative or add some description to make a clear statement that negative number will generate only one sstable.

src/java/org/apache/cassandra/io/sstable/AbstractSSTableSimpleWriter.java
```diff
@@ -110,6 +112,39 @@ private static SSTableId getNextId(File directory, final String columnFamily) th
         }
     }
 
+    /**
+     * Returns the partition update builder for the given key
```
Maxwell-Guo: Returns the partition update builder for the given key.

Missing a period (if you think it's needed), just align with the description of the method below

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
```diff
@@ -1233,6 +1235,77 @@ public void testWriteWithTimestampsAndTtl() throws Exception
...
+    public void testWriteWithSorted() throws Exception
+    {
+        String schema = "CREATE TABLE " + qualifiedTable + " ("
+                        + "  k int PRIMARY KEY,"
+                        + "  v blob )";
+        CQLSSTableWriter writer = CQLSSTableWriter.builder()
+                                                  .inDirectory(dataDir)
+                                                  .forTable(schema)
+                                                  .using("INSERT INTO " + qualifiedTable +
+                                                         " (k, v) VALUES (?,text_as_blob(?))" )
```
Maxwell-Guo: need a space before text_as_blob

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
```diff
@@ -1233,6 +1235,77 @@ public void testWriteWithTimestampsAndTtl() throws Exception
...
+    public void testWriteWithSortedAndMaxSize() throws Exception
+    {
+        String schema = "CREATE TABLE " + qualifiedTable + " ("
+                        + "  k int PRIMARY KEY,"
+                        + "  v blob )";
+        CQLSSTableWriter writer = CQLSSTableWriter.builder()
+                                                  .inDirectory(dataDir)
+                                                  .forTable(schema)
+                                                  .using("INSERT INTO " + qualifiedTable +
+                                                         " (k, v) VALUES (?,text_as_blob(?))" )
```
Maxwell-Guo: need a space before text_as_blob

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
```diff
@@ -36,64 +42,132 @@
...
  * rarely possible and SSTableSimpleUnsortedWriter should most of the time be
  * prefered.
+ * <p>
+ * Optionally, the writer can be configured with a max SSTable size for SSTables.
+ * The output will be a series of SSTables that do not exceed a specified size.
+ * By default, all sorted data are written into a single SSTable.
  */
 class SSTableSimpleWriter extends AbstractSSTableSimpleWriter
 {
+    private final long maxSSTableSize;
```
frankgh: can we either add a comment that mentions the units for the size, or better yet rename the variable to indicate it measures in bytes?
```suggestion
    private final long maxSSTableSizeInBytes;
```

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
```diff
@@ -36,64 +42,132 @@
...
+            Throwable e = writer.abort(t);
             Throwables.throwIfUnchecked(e);
             throw new RuntimeException(e);
         }
     }
 
     private void writePartition(PartitionUpdate update) throws IOException
     {
-        getOrCreateWriter().append(update.unfilteredIterator());
+        getOrSwitchWriter().append(update.unfilteredIterator());
```
frankgh: I think the original method name maintains the desired semantics `getOrCreateWriter` . The `create writer` portion is left for the implementation to decide when to create a new writer

src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
```diff
@@ -554,6 +555,18 @@ public Builder sorted()
             return this;
         }
 
+        /**
+         * Defines the maximum SSTable size when using the sorted writer.
+         * By default, i.e. not specified, there is no maximum size limit for the produced SSTable
+         * @param size the maximum size of each individual SSTable allowed
```
frankgh:
```suggestion
         * @param size the maximum size in mebibytes of each individual SSTable allowed
```

src/java/org/apache/cassandra/io/sstable/AbstractSSTableSimpleWriter.java
```diff
@@ -110,6 +112,39 @@ private static SSTableId getNextId(File directory, final String columnFamily) th
...
+    /**
+     * Returns the partition update builder that counts row sizes for the given key.
+     * @param key the partition key of the partition update
+     * @param onRow consumes the row to count the size
+     * @param onRowAdded runnale that runs after the row is added
+     * @return an update on partition {@code key} that is tied to this writer
+     */
+    protected PartitionUpdate.Builder createSizeCountingPartitionUpdateBuilder(DecoratedKey key,
+                                                                               Consumer<Row> onRow,
+                                                                               Runnable onRowAdded)
```
frankgh: to me it feels that all the logic can be encapsulated in the consumer lambda, and renders the runnable unnecessary. For the only calling site where we actually use the Runnable parameter I would refactor it as:

```
previous = createSizeCountingPartitionUpdateBuilder(key, row -> {
                countRow(row);
                maybeSync();
            });
```

```suggestion
    protected PartitionUpdate.Builder createSizeCountingPartitionUpdateBuilder(DecoratedKey key, Consumer<Row> onRow)
```

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
```diff
@@ -1233,6 +1235,77 @@ public void testWriteWithTimestampsAndTtl() throws Exception
...
+        int rowCount = 30_000;
+        // Max SSTable size is 1 MiB
+        // 30_000 rows should take 30_000 * (4 + 37) = 1.17 MiB > 1 MiB
+        for (int i = 0; i < rowCount; i++)
+        {
+            writer.addRow(i, UUID.randomUUID().toString());
+        }
+        writer.close();
+
+        File[] dataFiles = dataDir.list(f -> f.name().endsWith(BigFormat.Components.DATA.type.repr));
```
frankgh, line 1294: I ran this test and I am seeing 2 files which is correct. However, both files are a little over 600K , and I was expecting to see a file closer to 1MiB and the other one taking a few tens of bytes.

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
```diff
@@ -36,64 +42,132 @@
...
-            writer = createWriter(null);
-
-        return writer;
-    }
-
+    @Override
     PartitionUpdate.Builder getUpdateFor(DecoratedKey key) throws IOException
     {
-        assert key != null;
+        Preconditions.checkArgument(key != null, "Partition update cannot have null key");
```
frankgh, line 74: even though I like this change, we need to be mindful that this will have a performance difference if assertions are disabled. So before committing to this change, I think it'll be useful to consider whether the original assertion was desired as a way to have different performance characteristics depending on whether assertions are enabled or not. This change could have a noticeable impact to existing consumers of this class.

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
```diff
@@ -36,64 +42,132 @@
...
+            // and create a new update for the new key
+            currentSize += PartitionUpdate.serializer.serializedSize(createUpdateBuilder(key).build(),
+                                                                     format.getLatestVersion().correspondingMessagingVersion());
             currentKey = key;
-            update = new PartitionUpdate.Builder(metadata.get(), currentKey, columns, 4);
+            update = createSizeCountingPartitionUpdateBuilder(key, this::countRow, () -> {});
         }
 
-        assert update != null;
+        Preconditions.checkState(update != null, "Partition update to write cannot be null");
```
frankgh, line 87: same comment as above.

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
```diff
@@ -36,64 +42,132 @@
...
+        }
+        catch (Throwable t)
+        {
+            Throwable e = writer == null ? t : writer.abort(t);
+            Throwables.throwIfUnchecked(e);
+            throw new RuntimeException(e);
+        }
+    }
+
+    private void closeWriterIfPresent(SSTableTxnWriter writer)
```
frankgh: NIT for a more succinct method name
```suggestion
    private void maybeCloseWriter(SSTableTxnWriter writer)
```

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
yifan-c: For null key, the write fails elsewhere if `-ea` is off. It is desired to check for nullness, and stop right at this execution point.
And as you mentioned, `-ea` is not a reliable mean for assertion, as it can be disabled.

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
yifan-c: Same reply applies here.

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
yifan-c: fair.

src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
yifan-c: Good point. I would prefer to log a warning here and let the write process to proceed, since producing a single sstable is a valid behavior, but might not be expected. How about this message?

"Maximum SSTable size is specified for the sorted CQLSSTableWriter, but the value is negative, which disables the size limiting effectively. Please supply a positive value in order to enforce size limiting for the produced SSTables."

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
frankgh: correct, but my concern is about the performance impact that this change could potentially bring. I think it should be evaluated before making the change. Maybe it could be part of a different jira, where JMH benchmarks are considered?

yifan-c:
Hi folks, thanks for reviewing! 

I made a significant change in the implementation that improves the size counting accuracy by using the on-disk file pointer.

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
yifan-c: A nullness check is concerning to you? Maybe I am missing something? Can you elaborate?

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
yifan-c: Improved the accuracy

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
frankgh: These changes tend to make a difference when you are writing millions or records in a tight loop. It might not be an issue, but my concern is a performance degradation

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
frankgh: Great, the new approach is much much better.

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
yifan-c: Let's qualify the number of instructions.

Using `assert`
```
  org.apache.cassandra.db.partitions.PartitionUpdate$Builder getUpdateFor(org.apache.cassandra.db.DecoratedKey) throws java.io.IOException;
    descriptor: (Lorg/apache/cassandra/db/DecoratedKey;)Lorg/apache/cassandra/db/partitions/PartitionUpdate$Builder;
    flags: (0x0000)
    Code:
      stack=7, locals=2, args_size=2
         0: getstatic     #4                  // Field $assertionsDisabled:Z
         3: ifne          18
         6: aload_1
         7: ifnonnull     18
        10: new           #5                  // class java/lang/AssertionError
        13: dup
        14: invokespecial #6                  // Method java/lang/AssertionError."<init>":()V
        17: athrow
        18: aload_1
        19: aload_0
```

Using nullness check with `Precondition`
```
  org.apache.cassandra.db.partitions.PartitionUpdate$Builder getUpdateFor(org.apache.cassandra.db.DecoratedKey) throws java.io.IOException;
    descriptor: (Lorg/apache/cassandra/db/DecoratedKey;)Lorg/apache/cassandra/db/partitions/PartitionUpdate$Builder;
    flags: (0x0000)
    Code:
      stack=7, locals=2, args_size=2
         0: aload_1
         1: ifnull        8
         4: iconst_1
         5: goto          9
         8: iconst_0
         9: ldc           #5                  // String Partition update cannot have null key
        11: invokestatic  #6                  // Method com/google/common/base/Preconditions.checkArgument:(ZLjava/lang/Object;)V
        14: aload_0
```

Assume the input is not null.

In the case of `assert`, when with `-ea`, the method is going to check whether assertion is disabled and check whether the input is null. There are 4 instructions. When w/o `-ea`, there are 2 instructions to check whether assertion is disabled. 

In the case of `Preconditions`, there are 6 instructions, i.e. `0, 1, 4, 5, 9, 11` compare null, store comparison result, and invoke checkArgument method. It has 2 or 4 extra instructions comparing with `assert`, depending on whether it is enabled or not.

The method `getUpdateFor` is called per each row. Adding a row easily runs thousands of instructions. A couple more instruction is a negligible tradeoff, but provides a good improvement with the nicer descriptive error message and guarantees the check is always performed for correctness.

src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
Maxwell-Guo: ok

Maxwell-Guo (approved):

src/java/org/apache/cassandra/io/sstable/SSTableSimpleWriter.java
frankgh: Thanks for checking at the instructions level. +1 for using the new approach!

frankgh (approved):
+1 (nb) . Thanks for the patch and for addressing all the comments

yifan-c:
closing the committed patch

PR #2830, opened 2023-10-20
For Review CASSANDRA-18798 Appending to list in Accord transactions uses insertion timestamp

henrikingo:


src/java/org/apache/cassandra/db/rows/AbstractCell.java
```diff
@@ -114,14 +119,44 @@ public Cell<?> clone(ByteBufferCloner cloner)
...
     public Cell<?> updateAllTimestamp(long newTimestamp)
     {
-        return new BufferCell(column, isTombstone() ? newTimestamp - 1 : newTimestamp, ttl(), localDeletionTime(), buffer(), path());
+        CellPath newPath = null;
+        if (path() != null) {
+            byte[] newUUID = nextTimeUUIDAsBytes(newTimestamp);
+            newPath = CellPath.create(ByteBuffer.wrap(newUUID));
+            logger.debug("timestamp: {}    newPath: {}", newTimestamp, newPath.get(0));
+        }
+        return new BufferCell(column, isTombstone() ? newTimestamp - 1 : newTimestamp, ttl(), localDeletionTime(), buffer(), newPath);
```
blambov, line 128: Would this not change all cell paths, including the ones used in maps, breaking them?

Shouldn't this be done by `ListType` rather than the generic `Cell` implementation?

src/java/org/apache/cassandra/utils/TimeUUID.java
```diff
@@ -411,9 +437,28 @@ public static byte[] nextTimeUUIDAsBytes()
             return toBytes(rawTimestampToMsb(unixMicrosToRawTimestamp(nextUnixMicros())), clockSeqAndNode);
         }
 
+        public static byte[] nextTimeUUIDAsBytes(long unixMicrosAtLeast)
+        {
+            long useThisUnixMicros = unixMicrosAtLeast;
+            if(unixMicrosAtLeast>lastMicros.get())
+            {
+                lastMicros.set(unixMicrosAtLeast+1);
```
jacek-lewandowski, line 445: I think you should rather do CAS operation than just setting a value

src/java/org/apache/cassandra/db/rows/AbstractCell.java
```diff
@@ -114,14 +119,44 @@ public Cell<?> clone(ByteBufferCloner cloner)
...
+        // are keyed with a TimeUUID, which is based on the Cassandra timestamp. We want these to now also adhere to
+        // the accord executeAt timestamps, so that they are ordered the same as the transactions that wrote them.
+        // As an example, if a set of accord transactions all append an element to a list, then the order of the elements
+        // in the list should map to the order that the transactions were executed in.
+        CellPath newPath = null;
+        if (path() != null) {
+            // We need to use a next*() function, so that we get a new timestamp for each ListType element, in case
+            // one transaction would write several list elements. (Including writing into more than one list<> column
+            // within the same write.)
+            byte[] newUUID = nextTimeUUIDAsBytes(newTimestamp);
```
jacek-lewandowski: I'm wondering if this wouldn't make the same list items have different paths on each replica? After all, the `nextTimeUUIDAsBytes` function is not pure, is it?

src/java/org/apache/cassandra/db/rows/AbstractCell.java
henrikingo: Correct. Fixed in the most recent push, although I still didn't make it into a method of ListType class, just did `if(column.type...)` (Although this isn't "the right way" to write OO, I somehow instinctively dislike moving 2-4 lines to a completely separate file.)

I will look into Jacek's comment

src/java/org/apache/cassandra/utils/TimeUUID.java
```diff
@@ -411,9 +437,28 @@ public static byte[] nextTimeUUIDAsBytes()
             return toBytes(rawTimestampToMsb(unixMicrosToRawTimestamp(nextUnixMicros())), clockSeqAndNode);
         }
 
+        public static byte[] nextTimeUUIDAsBytes(long unixMicrosAtLeast)
```
dcapwell, line 440: this is unsafe in accord as it is:

1) based off the local node, doesn't act the same cross the cluster
2) shared by non-accord tables, so both are corrupting each other...

You really don't want this and instead want `unsafe(unixMicros)` which does not use any state...

Now, you still have the issue that the `clockSeqAndNode` isn't the same, so you have to solve that as well... if we used the instance that created the txn, then that would make sense... else we would need to have a global one used by all accord nodes so its consistent...

@belliottsmith thoughts?

src/java/org/apache/cassandra/db/rows/AbstractCell.java
```diff
@@ -114,14 +119,45 @@ public Cell<?> clone(ByteBufferCloner cloner)
...
+
+        final long newts = isTombstone() ? newTimestamp - 1 : newTimestamp;
+        final long localDeletionTime = localDeletionTime() != NO_DELETION_TIME ? newLocalDeletionTime : NO_DELETION_TIME;
+
+        // In addition to updating the timestamps for each row and partition, it turns out the elements of a ListType
+        // are keyed with a TimeUUID, which is based on the Cassandra timestamp. We want these to now also adhere to
+        // the accord executeAt timestamps, so that they are ordered the same as the transactions that wrote them.
+        // As an example, if a set of accord transactions all append an element to a list, then the order of the elements
+        // in the list should map to the order that the transactions were executed in.
+        CellPath newPath = path();
```
dcapwell, line 146: I have to check again but don't think this is an isolated problem, so might have to think about how best to handle this...

PR #2832, opened 2023-10-20
CASSANDRA-18952: Add metrics and logging to repair retries

dcapwell:


src/java/org/apache/cassandra/metrics/RepairMetrics.java
```diff
@@ -18,17 +18,41 @@
...
+import org.apache.cassandra.net.Verb;
 
 import static org.apache.cassandra.metrics.CassandraMetricsRegistry.Metrics;
 
 public class RepairMetrics
 {
     public static final String TYPE_NAME = "Repair";
     public static final Counter previewFailures = Metrics.counter(DefaultNameFactory.createMetricName(TYPE_NAME, "PreviewFailures", null));
+    private static final Histogram retries = Metrics.histogram(DefaultNameFactory.createMetricName(TYPE_NAME, "Retries", null), false);
+    private static final Map<Verb, Histogram> retriesByVerb = Collections.synchronizedMap(new EnumMap<>(Verb.class));
```
Mmuzaf: We have the thread-safety guarantees because `Verb` messages are never processed in parallel, right? `computeIfAbsent` doesn't look like a safe option for `synchronizedMap` because it's still not an atomic operation - so if messages for the `Verb` are not strictly ordered somewhere else we could lose metric values.

src/java/org/apache/cassandra/metrics/RepairMetrics.java
```diff
@@ -18,17 +18,41 @@
...
+    private static final Counter retryTimeout = Metrics.counter(DefaultNameFactory.createMetricName(TYPE_NAME, "RetryTimeout", null));
+    private static final Map<Verb, Counter> retryTimeoutByVerb = Collections.synchronizedMap(new EnumMap<>(Verb.class));
```
Mmuzaf: Are we guaranteed to see metrics in the JMX API if they are registered dynamically after the static context is initialized? I remember using the `DynamicMBean` for this case, but I couldn't find any mention of it in the source code, so I am concerned about metrics being static and I'm asking such a question :-)

src/java/org/apache/cassandra/metrics/RepairMetrics.java
```diff
@@ -18,17 +18,41 @@
...
+import com.codahale.metrics.Histogram;
+import org.apache.cassandra.net.Verb;
 
 import static org.apache.cassandra.metrics.CassandraMetricsRegistry.Metrics;
 
 public class RepairMetrics
 {
     public static final String TYPE_NAME = "Repair";
     public static final Counter previewFailures = Metrics.counter(DefaultNameFactory.createMetricName(TYPE_NAME, "PreviewFailures", null));
+    private static final Histogram retries = Metrics.histogram(DefaultNameFactory.createMetricName(TYPE_NAME, "Retries", null), false);
```
Mmuzaf: Do we need to write tests for these metrics?

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
```diff
@@ -165,13 +171,33 @@ public void onFailure(InetAddressAndPort from, RequestFailureReason failureReaso
...
                         finalCallback.onFailure(from, failureReason);
                         return;
                     default:
                         throw new AssertionError("Unknown error handler: " + allowed);
                 }
             }
 
+            private void maybeRecordRetry(boolean succcess)
+            {
+                if (attempt > 0) // if greater than 0 we know a retry happened...
```
Mmuzaf: nit: we can remove unnecessary identation 

```
if (attempt <= 0)
    return;
```

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
```diff
@@ -165,13 +171,33 @@ public void onFailure(InetAddressAndPort from, RequestFailureReason failureReaso
...
+                {
+                    // we don't know what the prefix kind is... so use NONE... this impacts logPrefix as it will cause us to use "repair" rather than "preview repair" which may not be correct... but close enough...
+                    String prefix = PreviewKind.NONE.logPrefix(request.parentRepairSession());
+                    RepairMetrics.retry(verb, attempt);
+                    if (succcess)
+                    {
+                        noSpam.info("{} Retry of repair verb " + verb + " was success after {} attempts", prefix, attempt);
```
frankgh: NIT
```suggestion
                        noSpam.info("{} Retry of repair verb " + verb + " was successful after {} attempts", prefix, attempt);
```

src/java/org/apache/cassandra/metrics/RepairMetrics.java
frankgh: yeah, I think `ConcurrentHashMap` is a better choice

dcapwell: > We have the thread-safety guarantees because Verb messages are never processed in parallel, right?

we can process multiple repairs in parallel, so that isn't true.

> computeIfAbsent doesn't look like a safe option for synchronizedMap because it's still not an atomic operation 

`java.util.Collections.SynchronizedMap#computeIfAbsent` is the following on JDK 11

```
public V computeIfAbsent(K key,
                Function<? super K, ? extends V> mappingFunction) {
            synchronized (mutex) {return m.computeIfAbsent(key, mappingFunction);}
        }
```

@Mmuzaf where did you see it not use the lock?  This is thread safe on JDK 11...

> yeah, I think ConcurrentHashMap is a better choice

This is thread safe on JDK 11 and should have less memory, so the fact that CHM is faster with contention isn't as valuable to me as having less memory.

src/java/org/apache/cassandra/metrics/RepairMetrics.java
dcapwell: Fair point... I didn't want to set that up upfront as that requires a list of supported verbs... which can be hard to maintain...

dcapwell: ok I moved the test set of verbs that allow retry to `RepairMessage` which enables me to pre-populate

src/java/org/apache/cassandra/metrics/RepairMetrics.java
Mmuzaf: @dcapwell My bad. I should have checked the `SynchronizedMap` implementation, not the `Map` interface itself, so you are right :-) How much RAM are we trying to save? I'm still concerned that we add the synchronized map to the hotpath, as my understanding is that metrics calculation should have no impact on the message processing as it is less important to the process itself.

This is anyway unrelated to the latest changes :-)

src/java/org/apache/cassandra/metrics/RepairMetrics.java
dcapwell: wrote tests and found a bug!  thanks for pushing me to do it!

The bug was that a failed request would trigger updating the timeout metric, so an `UNKNOWN` would increment `TIMEOUT` metric...  trivial fix, but tests helped to point that out

src/java/org/apache/cassandra/metrics/RepairMetrics.java
dcapwell: > I'm still concerned that we add the synchronized map to the hotpath

this isn't the hot path, this is retry messaging on the network thread... this is def the slow path =D

Also, your other feedback to preallocate avoids this problem as its now an immutable map...

dcapwell:
@frankgh , @Mmuzaf pushed all feedback, can you review again?

maedhroz (approved):
LGTM

The only minor thing we might want to add is a notion of retry failures in addition to retry timeouts, which would give us a more complete picture. At the moment, the total number of points in the retry histogram is the number of retry sequences we start, which counts the sequences that end in failure (but not timeout).

dcapwell:
@maedhroz we now have retry failure metric.  so we have `total_retries = success + timeout + failure`... we don't have a success metric so thats the only one that must be inferred

frankgh (approved):
Looks good to me

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
```diff
@@ -56,8 +60,23 @@
 public abstract class RepairMessage
 {
     private enum ErrorHandling { NONE, TIMEOUT, RETRY }
-    private static final CassandraVersion SUPPORTS_RETRY = new CassandraVersion("5.0.0-alpha2.SNAPSHOT");
+    @VisibleForTesting
+    static final CassandraVersion SUPPORTS_RETRY = new CassandraVersion("5.0.0-alpha2.SNAPSHOT");
     private static final Map<Verb, CassandraVersion> VERB_TIMEOUT_VERSIONS;
+    public static final Set<Verb> ALLOWS_RETRY;
+    private static final Set<Verb> SUPPORTS_RETRY_WITHOUT_VERSION_CHECK = Collections.unmodifiableSet(EnumSet.of(Verb.CLEANUP_MSG));
+    public static final RequestCallback<Object> NOOP_CALLBACK = new RequestCallback<>()
```
Mmuzaf, line 68: nit: I think we can move it to the `RequestCallback` interface e.g. 

```
public interface RequestCallback<T>
{
    RequestCallback<Object> NOOP_CALLBACK = msg -> {};
    
    ...
}
```

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
```diff
@@ -165,13 +188,37 @@ public void onFailure(InetAddressAndPort from, RequestFailureReason failureReaso
...
+                    noSpam.info("{} Retry of repair verb " + verb + " was successful after {} attempts", prefix, attempt);
+                }
+                else if (reason == RequestFailureReason.TIMEOUT)
+                {
+                    noSpam.warn("{} Timeout for repair verb " + verb + "; could not complete within {} attempts", prefix, attempt);
+                    RepairMetrics.retryTimeout(verb);
+                }
+                else
+                {
+                    noSpam.warn("{} Failure for repair verb " + verb + "; could not complete within {} attempts", prefix, attempt);
```
Mmuzaf, line 217: Is it worth logging the `reason` as well?

Mmuzaf (approved):
@dcapwell I went through the changes, and now it looks good!
I've left some minor comments, but we can live without fixing them :-)

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
dcapwell: switched to `WARN  [main] 2023-10-26 14:48:08,084 NoSpamLogger.java:107 - [repair #58395110-7449-11ee-b645-8b78869f39eb] UNKNOWN failure for repair verb VALIDATION_REQ; could not complete within 1 attempts`

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
dcapwell: in this case I rather not.  There isn't a reason to use a no-op callback in MessagingService as `null` is what you actually want (noop has higher cost than null...), but since we are doing retries we need a way to tell you we are done, so this should really only be used with retries, so feel its best to stay here

PR #2836, opened 2023-10-23
CASSANDRA-18945: Unified Compaction Strategy is creating too many sstables

ethan-brown2022:
- Added `min_sstable_size` and `sstable_growth` options
- Updated `getNumShards()` to include sstable growth
- When `(localDensity / minSize) < baseShardCount` and `baseShardCount` is not a power of 2, split to the highest power of 2 that is a divisor of `baseShardCount`
- Added and updated unit tests

```

```

[CASSANDRA-18945](https://issues.apache.org/jira/browse/CASSANDRA-18945)

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md
+++ b/src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md
@@ -156,7 +156,7 @@ Density levelling permits a much wider variety of splitting options including on
 be kept close to a selected target, and also allows UCS to understand the levelling structure of STCS (where size grows
 with each level) as well as LCS (where token share shrinks with each level).
 
-## Sharding
+## Basic sharding scheme
 
 Once density levelling is in place, we have a range of choices for splitting sstables. One is to simply split
 when a certain output size is reached (like LCS), forming non-overlapping sstable runs instead of individual
@@ -213,6 +213,66 @@ local token space, thus 4 for the 1/4 span that the compaction covers. Assuming 
 deletions, the resulting sstables will be of size 75 MiB, token share 1/16 and density 1200 MiB.
 
 This sharding mechanism is independent of the compaction specification.
+
+## Full sharding scheme
+
+This sharding scheme easily admits extensions. In particular, when the size of the data set is expected to grow very
+large, to avoid having to pre-specify a high enough target size to avoid problems with per-sstable overhead, we can
+apply an "sstable growth" parameter, which determines what part of the density growth should be assigned to increased
+SSTable size, reducing the growth of the number of shards (and hence non-overlapping sstables).
+
+Additionally, to allow for a mode of operation with a fixed number of shards, and splitting conditional on reaching
+a minimum size, we provide for a "minimum sstable size" that reduces the base shard count whenever that would result
+in sstables smaller than the provided minimum.
+
+Generally, the user can specify four sharding parameters:
+
+- base shard count $b$
+- target sstable size $t$
+- minimum sstable size $m$
+- sstable growth component $\lambda$
+
+The number of shards $S$ for a given density $d$ is then calculated as
+
+$$
+S =
+\begin{cases}
+1
+    & \text{if } d < m \\
+min(2^{\left\lfloor \log_2 \frac d m \right\rfloor}, x)
+    & \text{if } d < mb \text{, where } x \text{ is the largest power of 2 divisor of } b \\
+b
+    & \text{if } d < tb \\
+2^{\left\lfloor (1-\lambda) \cdot \log_2 \left( {\frac d t \cdot \frac 1 b}\right)\right\rceil} \cdot b
+    & \text{otherwise}
+\end{cases}
+$$
+
+Some useful combinations of these parameters:
+
+- The basic scheme above uses a sstable growth $\lambda=0$, and a minimum sstable size $m=0$. The graph below
+  illustrates it for base shard count $b=4$ and target sstable size $t=1\mathrm{GB}$:
+
+![Graph with lambda 0](unified/shards_graph_lambda_0.svg)
+
+- Using $\lambda = 0.5$ makes the strategy grow the shard count and sstable size evenly. When the density
+  quadruples, both the shard count and the expected sstable size for that density band will double. The example
+  below uses $b=8$, $t=1\mathrm{GB}$ and also applies a minimal size $m=100\mathrm{MB}$:
+
+![Graph with lambda 0.5](unified/shards_graph_lambda_0_5.svg)
+
+- Similarly, $\lambda = 1/3$ makes the sstable growth the cubic root of the density growth, i.e. the sstable size
+  grows with the square root of the growth of the shard count. The graph below uses $b=1$ and $t = 1\mathrm{GB}$
+  (note: when $b=1$ the minimal size has no effect):
+
+![Graph with lambda 0.33](unified/shards_graph_lambda_0_33.svg)
+
+- A growth component of 1 constructs a hierarchy with exactly $b$ shards at every level. Combined with a minumum
+  sstable size, this defines a mode of operation where we use a pre-specified
+  number of shards, but split only after reaching a minimum size. Illustrated below for $b=10$ and $m=100\mathrm{MB}$
+  (note: the target sstable size is irrelevant when $\lambda=1$):
+
+![Graph with lambda 1](unified/shards_graph_lambda_1.svg)
 
 ## Choosing sstables to compact
 
@@ -363,6 +423,23 @@ UCS accepts these compaction strategy parameters:
   minimum compaction concurrency for the lowest levels. A low number would result in larger L0 sstables but may limit
   the overall maximum write throughput (as every piece of data has to go through L0).  
   The default value is 4 (1 for system tables, or when multiple data locations are defined).
+* **sstable_growth** The sstable growth component $\lambda$, applied as a factor in the shard exponent calculation.
+  This is a number between 0 and 1 that controls what part of the density growth should apply to individual sstable
+  size and what part should increase the number of shards. Using a value of 1 has the effect of fixing the shard
+  count to the base value. Using 0.5 makes the shard count and sstable size grow with the square root of the density
+  growth.
+  This is useful to decrease the sheer number of sstables that will be created for very large data sets. For
+  example, without growth correction a data set of 10TiB with 1GiB target size would result in over 10k sstables,
+  which may present as too much overhead both as on-heap memory used by per-sstable structures as well as time to look
+  for intersecting sstables and tracking overlapping sets during compaction. Applying $\lambda=0.5$
+  in this scenario (with base count 4) will reduce the potential number of sstables to ~160 of ~64GiB, which is still
+  manageable both as memory overhead and individual compaction duration and space overhead. The balance between the
+  two can be further tweaked by increasing $\lambda$ to get fewer but bigger sstables on the top level, and decreasing
+  it to favour a higher count of smaller sstables. The default value is 0, corresponding to a fixed sstable target size.
+* **min_sstable_size** The minimum sstable size $m$, applicable when the base shard count will result is sstables
+  that are considered too small. If set, the strategy will split the space into fewer than the base count shards, to
+  make the estimated sstables size at least as large as this value.
+  The default value is 0, which disables this feature.
 * **expired_sstable_check_frequency_seconds**. Determines how often to check for expired SSTables.  
   The default value is 10 minutes.
 
--- a/src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md
+++ b/src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md
@@ -218,12 +218,12 @@ This sharding mechanism is independent of the compaction specification.
 
 This sharding scheme easily admits extensions. In particular, when the size of the data set is expected to grow very
 large, to avoid having to pre-specify a high enough target size to avoid problems with per-sstable overhead, we can
-apply an "sstable growth" parameter, which determines what part of the density growth should be assigned to increased
+apply an "SSTtable growth" parameter, which determines what part of the density growth should be assigned to increased
 SSTable size, reducing the growth of the number of shards (and hence non-overlapping sstables).
 
 Additionally, to allow for a mode of operation with a fixed number of shards, and splitting conditional on reaching
-a minimum size, we provide for a "minimum sstable size" that reduces the base shard count whenever that would result
-in sstables smaller than the provided minimum.
+a minimum size, we provide for a "minimum SSTable size" that reduces the base shard count whenever that would result
+in SSTables smaller than the provided minimum.
 
 Generally, the user can specify four sharding parameters:
 
--- a/src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md
+++ b/src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md
@@ -421,8 +421,8 @@ UCS accepts these compaction strategy parameters:
   The default value is 1 GiB.
 * **base_shard_count**. The minimum number of shards $b$, used for levels with the smallest density. This gives the
   minimum compaction concurrency for the lowest levels. A low number would result in larger L0 sstables but may limit
-  the overall maximum write throughput (as every piece of data has to go through L0).  
-  The default value is 4 (1 for system tables, or when multiple data locations are defined).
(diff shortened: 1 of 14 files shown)


src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -75,6 +85,32 @@ public class Controller
...
+     * Provision for growth of the constructed SSTables as the size of the data grows. By default the target SSTable
+     * size is fixed for all levels. In some scenarios is may be better to reduce the overall number of SSTables when
+     * the data size becomes larger to avoid using too much memory and processing for the corresponding structures.
+     * The setting enables such control and determines how much we reduce the growth of the number of split points as
+     * the data size grows. The number specifies the sstable growth part, and the difference from 1 is the shard count
+     * growth component, which is a multiplier applied to the logarithm of the data size, before it is rounded and
+     * applied as an exponent in the number of split points. In other words, the given value applies as a negative
+     * exponent in the calculation of the number of split points.
+     * <p>
+     * Using 0 (the default) applies no correction to the number of split points, resulting in SSTables close to the
```
blambov: This patch changes the default to 1/3.

Actually, could we introduce the extra machinery in one commit, and change the default in a separate one? (Both committed with this ticket, just separately so it's easy to roll back just one.)

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -202,36 +249,180 @@ public int getThreshold(int index) {
...
+                             localDensity / targetSize,
+                             FBUtilities.prettyPrintBinary(targetSize, "B", " "));
+            }
+            return shards;
+        }
+    }
+
+    private boolean isPowerOf2(int n)
+    {
+        return (int)(Math.ceil((Math.log(n) / Math.log(2)))) == (int)(Math.floor(((Math.log(n) / Math.log(2)))));
```
blambov: An easier (and more precise) way to do this is `return (x & (x-1)) == 0`.

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -202,36 +249,180 @@ public int getThreshold(int index) {
...
+     * This is either set by the user in the options or calculated by rounding up the first flush size to 50 MB.
+     *
+     * @return the minimum sstable size in bytes.
+     */
+    public long getMinSstableSizeBytes()
+    {
+        if (minSSTableSize >= 0)
+            return minSSTableSize;
+
+        synchronized (this)
```
blambov: We should only apply a minimum size if the user has specified it, there is no need to fall back to the flush size when none is given.

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -202,36 +249,180 @@ public int getThreshold(int index) {
...
+    {
+        return (int)(Math.ceil((Math.log(n) / Math.log(2)))) == (int)(Math.floor(((Math.log(n) / Math.log(2)))));
+    }
+
+    /**
+     * @param n number of shards needs to be divisible by n
+     * @param shards the maximum number of shards
+     * @return the largest power of 2 that is a divisor of n (e.g. when n = 10, return 2) and less than shards
+     */
+    private int getLargestPowerOf2Divisor(int n, int shards)
```
blambov: The largest power-of-2 divisor of `n` can be found via `Integer.numberOfTrailingZeros(n)`. 
I.e. this can be implemented as `Math.min(shards, 1 << Integer.numberOfTrailingZeros(n))`.

This would also work without an explicit `isPowerOf2` check.

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -202,36 +249,180 @@ public int getThreshold(int index) {
...
+        {
+            // How many we would have to aim for the target size. Divided by the base shard count, so that we can ensure
+            // the result is a multiple of it by multiplying back below.
+            double count = localDensity / (targetSSTableSize * baseShardCount);
+            // Take a logarithm of the count (in base 2), and adjust it by the given growth modifier.
+            // Adjust by 0.5 to round the exponent so that the result falls between targetSSTableSize * sqrt(0.5) and
+            // targetSSTableSize * sqrt(2). Finally, make sure the exponent is at least 0 and not greater than the
+            // fixed maximum.
+            // Note: This code also works correctly for the special cases of sstableGrowthModifier == 0 and 1,
+            //       but the above code avoids the floating point arithmetic for these common cases.
```
blambov: Nit: We can add "imprecise" before "floating point arithmetic" as the loss of precision is the main reason to avoid this path.

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -202,36 +249,180 @@ public int getThreshold(int index) {
...
      * This is calculated as a power-of-two multiple of baseShardCount, so that the expected size of resulting sstables
-     * is between sqrt(0.5) * targetSSTableSize and sqrt(2) * targetSSTableSize, with a minimum of baseShardCount shards
-     * for smaller sstables.
-     *
+     * is between sqrt(0.5) and sqrt(2) times the target size, which is calculated from targetSSTableSize to grow
+     * at the given sstableGrowthModifier of the exponential growth of the density.
+     * <p>
+     * Additionally, if a minimum sstable size is set, we can go below the baseShardCount when that would result in
+     * sstables smaller than that minimum. Note that in the case of a non-power-of-two base count this will cause
+     * smaller sstables to not be aligned with the ones whose size is enough for the base count.
```
blambov: This is not true as we will only split to divisors of `baseShardCount`.

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -53,6 +55,14 @@ public class Controller
     private final static String DEFAULT_SCALING_PARAMETERS =
         CassandraRelevantProperties.UCS_SCALING_PARAMETER.getString();
 
+    /**
+     * The minimum sstable size. Sharded writers split sstables over shard only if they are at least as large as the
+     * minimum size.
+     */
+    static final String MIN_SSTABLE_SIZE_OPTION = "min_sstable_size";
+
+    static final long DEFAULT_MIN_SSTABLE_SIZE = FBUtilities.parseHumanReadableBytes(System.getProperty(PREFIX + MIN_SSTABLE_SIZE_OPTION, "100MiB"));
```
blambov: All these properties must be retrieved through `CassandraRelevantProperties` (see e.g. `DEFAULT_SCALING_PARAMETER` above).

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -53,6 +55,14 @@ public class Controller
     private final static String DEFAULT_SCALING_PARAMETERS =
         CassandraRelevantProperties.UCS_SCALING_PARAMETER.getString();
 
+    /**
+     * The minimum sstable size. Sharded writers split sstables over shard only if they are at least as large as the
+     * minimum size.
+     */
+    static final String MIN_SSTABLE_SIZE_OPTION = "min_sstable_size";
```
blambov, line 59: We currently have special code to change the base shard count to 1 for system tables (see `DEFAULT_BASE_SHARD_COUNT` below) because they are expected to be small. The min size makes this unnecessary; can we remove that special case as part of the commit that changes the default min size?

blambov (commented):
The markdown documentation needs to be updated as well.

src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md
```diff
@@ -214,6 +214,66 @@ deletions, the resulting sstables will be of size 75 MiB, token share 1/16 and d
...
+    & \text{otherwise}
+\end{cases}
+$$
+
+Some useful combinations of these parameters:
+
+- The basic scheme above uses a sstable growth $\lambda=0$, and a minimum sstable size $m=0$. The graph below
+  illustrates it for base shard count $b=4$ and target sstable size $t=1\mathrm{GB}$:
+
+![Graph with lambda 0](unified/shards_graph_lambda_0.svg)
```
blambov, line 256: The graphs appear to be missing.

src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md
```diff
@@ -214,6 +214,66 @@ deletions, the resulting sstables will be of size 75 MiB, token share 1/16 and d
...
+- sstable growth component $\lambda$
+
+The number of shards $S$ for a given density $d$ is then calculated as
+
+$$
+S =
+\begin{cases}
+1
+    & \text{if } d < m \\
+2^{\left\lfloor \log_2 \frac d m \right\rfloor}
```
blambov: This is actually the minimum of this number and the highest power-of-2 divisor of `b`.

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -202,36 +247,110 @@ public int getThreshold(int index) {
...
+        {
+            double count = localDensity / minSSTableSize;
+            // Minimum size only applies if it is smaller than the base count.
+            // Note: the minimum size cannot be larger than the target size's minimum.
+            if (count < baseShardCount)
+            {
+                // Make it a power of two, rounding down so that sstables are greater in size than the min.
+                // Setting the bottom bit to 1 ensures the result is at least 1.
+                // If baseShardCount is not a power of 2, split only to powers of two that are divisors of baseShardCount so boundaries match higher levels
+                shards = Math.min(Integer.highestOneBit((int) count | 1), 1 << Integer.numberOfTrailingZeros(baseShardCount));
```
blambov: Nit: I found an even easier way to find the highest power-of-2 divisor / lowest one bit of a number: `n & -n`.

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -93,10 +92,12 @@ public class Controller
...
+     * Using 0 applies no correction to the number of split points, resulting in SSTables close to the
      * target size. Setting this number to 1 will make UCS never split beyong the base shard count. Using 0.5 will
      * make the number of split points a square root of the required number for the target SSTable size, making
-     * the number of split points and the size of SSTables grow in lockstep as the density grows.
+     * the number of split points and the size of SSTables grow in lockstep as the density grows. Using
+     * 0.333 (the default) makes the sstable growth the cubic root of the density growth, i.e. the sstable size
+     * grows with the square root of the growth of the shard count.
      * <p>
      * For example, given a data size of 1TiB on the top density level and 1GiB target size with base shard count of 1,
      * growth 0 would result in 1024 SSTables of ~1GiB each, 0.5 would yield 32 SSTables of ~32GiB each, and 1 would
```
blambov: Let's also add 1/3 to these examples: 128 SSTables of ~8 GiB each, which happens to be the case both here (for b=1) and in the next paragraph (for b = 4).

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -424,8 +425,7 @@ public static Controller fromOptions(ColumnFamilyStore cfs, Map<String, String>
         }
         else
         {
-            if (SchemaConstants.isSystemKeyspace(cfs.getKeyspaceName())
-                || (cfs.getDiskBoundaries().positions != null && cfs.getDiskBoundaries().positions.size() > 1))
+            if (cfs.getDiskBoundaries().positions != null && cfs.getDiskBoundaries().positions.size() > 1)
```
blambov: I would remove this special case too, the minimum size should handle it.

blambov:
Something else I forgot: we need an entry in CHANGES.txt about the patch, and the commit messages should use the agreed format (see "Tips" in the contributor guidelines). The UCS description in NEWS.txt does not appear to need any changes.

src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md
```diff
@@ -214,6 +214,66 @@ deletions, the resulting sstables will be of size 75 MiB, token share 1/16 and d
...
+![Graph with lambda 0.5](unified/shards_graph_lambda_0_5.svg)
+
+- Similarly, $\lambda = 1/3$ makes the sstable growth the cubic root of the density growth, i.e. the sstable size
+  grows with the square root of the growth of the shard count. The graph below uses $b=1$ and $t = 1\mathrm{GB}$
+  (note: when $b=1$ the minimal size has no effect):
+
+![Graph with lambda 0.33](unified/shards_graph_lambda_0_33.svg)
+
+- A growth component of 1 constructs a hierarchy with exactly $b$ shards at every level. Combined with a minumum
+  sstable size, this defines a mode of operation similar to UCS V1 (as used in DSE 6.8), where we use a pre-specified
```
blambov: Drop "similar to UCS V1 (as used in DSE 6.8),".

src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md
```diff
@@ -363,6 +423,24 @@ UCS accepts these compaction strategy parameters:
...
+  growth.
+  This is useful to decrease the sheer number of sstables that will be created for very large data sets. For
+  example, without growth correction a data set of 10TiB with 1GiB target size would result in over 10k sstables,
+  which may present as too much overhead both as on-heap memory used by per-sstable structures as well as time to look
+  for intersecting sstables and tracking overlapping sets during compaction. Applying $\lambda=0.5$
+  in this scenario (with base count 4) will reduce the potential number of sstables to ~160 of ~64GiB, which is still
+  manageable both as memory overhead and individual compaction duration and space overhead. The balance between the
+  two can be further tweaked by increasing $\lambda$ to get fewer but bigger sstables on the top level, and decreasing
+  it to favour a higher count of smaller sstables. The default value is 0.333 meaning the sstable size
+  grows with the square root of the growth of the shard count
```
blambov: End the sentence with a period.

blambov (approved):

src/java/org/apache/cassandra/utils/FBUtilities.java
```diff
@@ -1030,6 +1030,22 @@ public static long parseHumanReadableBytes(String value)
...
     }
 
+    /**
+     * Parse a double where both a direct value and a percentage are accepted.
+     * For example, for inputs "0.1" and "10%", this function will return 0.1.
+     */
+    public static double parsePercent(String value)
+    {
+        value = value.trim();
+        if (value.endsWith("%"))
```
smiklosovic, line 1040: @ethan-brown2022 what if the value ends on multiple "%" like this? "0.1%%%"? Then you will try to parse "0.1%%". It is questionable if we want to do something with it though.

ethan-brown2022: If the value ends on multiple "%", it will throw a `NumberFormatException`.  This is the same result as any other string that cannot be parsed as a double, so I think it makes sense for this case as well.

smiklosovic: ok

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -200,38 +246,112 @@ public int getThreshold(int index) {
...
+        }
+        else if (sstableGrowthModifier == 0)
+        {
+            // How many we would have to aim for the target size. Divided by the base shard count, so that we can ensure
+            // the result is a multiple of it by multiplying back below. Adjusted by sqrt(0.5) to calculate the split
+            // points needed for the minimum size.
+            double count = localDensity / (targetSSTableSize * INVERSE_SQRT_2 * baseShardCount);
+            if (count > MAX_SHARD_SPLIT)
+                count = MAX_SHARD_SPLIT;
+            assert !(count < 0);    // Must be positive, 0 or NaN, which should translate to baseShardCount
```
smiklosovic, line 306: @ethan-brown2022  `count >= 0` is more natural to me

src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md
```diff
@@ -363,6 +423,24 @@ UCS accepts these compaction strategy parameters:
   minimum compaction concurrency for the lowest levels. A low number would result in larger L0 sstables but may limit
   the overall maximum write throughput (as every piece of data has to go through L0).  
   The default value is 4 (1 for system tables, or when multiple data locations are defined).
```
blambov: This default is now 4 for all tables; the description needs to mention that this only applies after `min_sstable_size` is reached.

PR #2840, opened 2023-10-25
CASSANDRA-18956 3.0 Fix incorrect column identifier bytes problem when renaming a column

hanke580:
…CASSANDRA-18956

This PR solves CASSANDRA-18956.

Problem: when renaming a column to another one, it checks whether the column exists in the table. However, when handling legacy thrift-model data, the check is implemented incorrectly.
* FROM => TO
Cassandra checks whether the FROM column exists but not check whether the TO column exists. This PR adds the check for the TO column.


test/unit/org/apache/cassandra/schema/LegacySchemaMigratorThriftTest.java
```diff
@@ -0,0 +1,53 @@
...
+public class LegacySchemaMigratorThriftTest extends LegacySchemaMigratorBaseTest
+{
+    public static final String KEYSPACE_18956 = "ks18956";
+    public static final String TABLE_18956 = "table18956";
+
+    @Test
+    public void testMigrate18956() throws IOException
+    {
+        CQLTester.cleanupAndLeaveDirs();
+        Keyspaces expected = LegacySchemaMigratorBaseTest.keyspacesToMigrate18956();
```
smiklosovic: @hanke580 move that method into this class. Currently, in that method, you again depend on a test. Abstract class should not depend on this one.

test/unit/org/apache/cassandra/schema/LegacySchemaMigratorThriftTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.apache.cassandra.config.Schema;
+import org.apache.cassandra.cql3.CQLTester;
+import org.apache.cassandra.exceptions.InvalidRequestException;
+import static org.apache.cassandra.cql3.QueryProcessor.executeOnceInternal;
+
+@SuppressWarnings("deprecation")
+public class LegacySchemaMigratorThriftTest extends LegacySchemaMigratorBaseTest
+{
+    public static final String KEYSPACE_18956 = "ks18956";
```
smiklosovic: make these two private

test/unit/org/apache/cassandra/schema/LegacySchemaMigratorThriftTest.java
```diff
@@ -0,0 +1,53 @@
...
+    @Test
+    public void testMigrate18956() throws IOException
+    {
+        CQLTester.cleanupAndLeaveDirs();
+        Keyspaces expected = LegacySchemaMigratorBaseTest.keyspacesToMigrate18956();
+        expected.forEach(LegacySchemaMigratorBaseTest::legacySerializeKeyspace);
+        LegacySchemaMigrator.migrate();
+        Schema.instance.loadFromDisk();
+        LegacySchemaMigratorBaseTest.loadLegacySchemaTables();
+        try {
```
smiklosovic: braces on new line in whole try-catch

test/unit/org/apache/cassandra/schema/LegacySchemaMigratorThriftTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+import org.junit.Test;
+
+import org.apache.cassandra.config.Schema;
+import org.apache.cassandra.cql3.CQLTester;
+import org.apache.cassandra.exceptions.InvalidRequestException;
+import static org.apache.cassandra.cql3.QueryProcessor.executeOnceInternal;
+
+@SuppressWarnings("deprecation")
+public class LegacySchemaMigratorThriftTest extends LegacySchemaMigratorBaseTest
```
smiklosovic: @hanke580 btw what is "Thirft" about this? What was the problem that we could not have one class?

test/unit/org/apache/cassandra/schema/LegacySchemaMigratorThriftTest.java
hanke580: Fixed

test/unit/org/apache/cassandra/schema/LegacySchemaMigratorThriftTest.java
hanke580: The base class need to using table name to decide whether to use serializeKind or serializeKind18956. The bug needs `serializeKind18956` to reproduce.
```
        if (table.cfName.equals(LegacySchemaMigratorThriftTest.TABLE_18956)) {
            adder.add("validator", column.type.toString())
                    .add("type", serializeKind18956(column.kind))
                    .add("component_index", column.position());
        } else {
            adder.add("validator", column.type.toString())
                    .add("type", serializeKind(column.kind, table.isDense()))
                    .add("component_index", column.position());
        }
```
It doesn't look good in base method. I plan to remove the static tag for the base methods and abstract `serializeKind` function so that the two child tests can implement its own `serializeKind` function. Do you think that makes sense?

test/unit/org/apache/cassandra/schema/LegacySchemaMigratorThriftTest.java
hanke580: **Thrift** means the legacy table is in thrift model created with cassandra-cli.
The reason they cannot be put into a single class is that between two tests, it need to clean the *data&commit log* folder. Otherwise there will be conflicts when creating/loading legacy data.

smiklosovic: @hanke580 why dont we clean data and commit log folder then? :)

smiklosovic: @hanke580 leave it is it is now ... I ve checked and I do not have a clue why it throws even we clean it etc ... If separation into two classes helps, well ... that's just fine.

test/unit/org/apache/cassandra/schema/LegacySchemaMigratorThriftTest.java
smiklosovic: yes that makes sense.

test/unit/org/apache/cassandra/schema/LegacySchemaMigratorThriftTest.java
hanke580: Thanks! I also rechecked it. The reason is somehow related to a concurrent file descriptor read, but I am not sure.
I'll follow the previous methods.

test/unit/org/apache/cassandra/schema/LegacySchemaMigratorThriftTest.java
hanke580: I have modified them to private and abstracted the `serializeKind` function.

PR #40, opened 2014-08-05
Fix bug in setting up remote host min/max simul requests

stragulus:
Without this fix, using the CqlConfigHelper will run into an exception where it complains MaxRequests cannot be higher than MinRequests.

PR #2842, opened 2023-10-25
Cassandra-16565 trunk remove Sigar

Claudenw:
 
```

```

Removed Sigar references from the build scripts and library.
Removed SigarLibrary.java and replaced it with SystemInfo.java and updated the calls.

Switched to [OSHI library](https://github.com/oshi/oshi)
Modified `warnIfRunningInDegradedMode` as noted on the ticket CASSANDRA-16565


src/java/org/apache/cassandra/service/StartupChecks.java
```diff
@@ -364,7 +364,7 @@ public void execute(StartupChecksOptions options)
         {
             if (options.isDisabled(getStartupCheckType()))
                 return;
-            SigarLibrary.instance.warnIfRunningInDegradedMode();
+            new SystemInfo().warnIfRunningInDegradedMode();
```
smiklosovic: @Claudenw the startup check on line 360 is still called `initSigarLibrary`.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+    private Logger logger = LoggerFactory.getLogger(SystemInfo.class);
+
+    /* The oshi.SystemInfo has  the following note:
+     * Platform-specific Hardware and Software objects are retrieved via memoized suppliers. To conserve memory at the
+     * cost of additional processing time, create a new version of SystemInfo() for subsequent calls. To conserve
+     * processing time at the cost of additional memory usage, re-use the same {@link SystemInfo} object for future
+     * queries.
+     *
+     * We are opting for minimal memory footprint. */
+    private oshi.SystemInfo si ;
```
smiklosovic: @Claudenw can't be this `import`ed too?

EDIT: hah, I see that, because this class is called the same way ... well ... Do you know of any better name to not have it like that? It is a little bit confusing.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+    private static long INFINITY = -1l;
+    private static long EXPECTED_MIN_NOFILE = 10000l; // number of files that can be opened
+    private static long EXPECTED_NPROC = 32768l; // number of processes
+    private static long EXPECTED_AS = 0x7FFFFFFFl; // address space
+    /**
+     * The default number of processes that are reported if the actual value can not be retrieved.
+     */
+    public static long DEFAULT_MAX_PROCESSES = 1024;
+
+    private Logger logger = LoggerFactory.getLogger(SystemInfo.class);
```
smiklosovic: @Claudenw might be static and final? could you move it to the top of the class too?

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+import oshi.PlatformEnum;
+
+/**
+ * An abstraction of System information, this class provides access to system information without specifying how
+ * it is retrieved.
+ */
+public class SystemInfo
+{
+
+    private static long INFINITY = -1l;
```
smiklosovic: all might be final

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+    private static long EXPECTED_MIN_NOFILE = 10000l; // number of files that can be opened
+    private static long EXPECTED_NPROC = 32768l; // number of processes
+    private static long EXPECTED_AS = 0x7FFFFFFFl; // address space
+    /**
+     * The default number of processes that are reported if the actual value can not be retrieved.
+     */
+    public static long DEFAULT_MAX_PROCESSES = 1024;
```
smiklosovic: why does this have to be public and why is it not final?

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+     * Gets the maximum number of processes the user can create.
+     * Note: if not on a Linux system this always return the
+     * @{code DEFAULT_MAX_PROCESSES} value.
+     * @return The maximum number of processes.
+     * @see #DEFAULT_MAX_PROCESSES
+     */
+    long getMaxProcess() {
+        // this check only works on Linux systems.
+        if (oshi.SystemInfo.getCurrentPlatform() == PlatformEnum.LINUX) {
+            Path p =  FileSystems.getDefault().getPath("/proc", Long.toString(getPid()), "limits");
```
smiklosovic: @Claudenw you have `org.apache.cassandra.io.util.FileUtils.readLines` method for this.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+            try
+            {
+                List<String> lines = Files.readAllLines(p);
+                for (String s : lines)
+                {
+                    if (s.startsWith("Max processes"))
+                    {
+                        String[] parts = s.split("\s+");
+                        String limit = parts[2];
+                        if ("unlimited".equals(limit))
```
smiklosovic: cant be this just a simple ternary operator? `return "unlimited".equals(limit) ? INFINITY : Long.parse(limit);`

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+
+    /**
+     * @return the PID of the current system.
+     */
+    public long getPid() {
+        return si.getOperatingSystem().getProcessId();
+    }
+
+    /**
+     * Checks if a value if valide (i.e. value >= min or value == INFINITY..
```
smiklosovic: typo in `valide`. Also `INFINITY)`.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+    {
+        boolean degraded = false;
+        StringBuilder sb = new StringBuilder();
+
+        if (oshi.SystemInfo.getCurrentPlatform() == PlatformEnum.LINUX)
+        {
+            // only check proc on nproc linux
+            if (!valid(getMaxProcess(), EXPECTED_NPROC))
+            {
+                sb.append("Number of processes should be >= ").append(EXPECTED_NPROC).append("\n");
```
smiklosovic: I think you meant `<=` instead of `>=` here.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+            degraded = true;
+            sb.append("System is running ").append(platform()).append(", Linux OS is recommended");
+        }
+
+        if (getSwapSize() > 0)
+        {
+            sb.append("Swap should be disabled\n");
+            degraded = true;
+        }
+        if (!valid(getMaxProcess(), EXPECTED_NPROC))
```
smiklosovic: This is checked twice. Also, upon the first check above, you said that this should be checked on Linux only.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+        }
+        if (!valid(getMaxOpenFiles(), EXPECTED_MIN_NOFILE))
+        {
+            sb.append("Minimum value for max open files should be >=  ").append(EXPECTED_MIN_NOFILE).append("\n");
+            degraded = true;
+        }
+        if (degraded) {
+            if (CassandraRelevantProperties.TEST_IGNORE_SIGAR.getBoolean())
+            {
+                logger.warn("Cassandra server running in degraded mode.\n" + sb.toString());
```
smiklosovic: I am not sure all these individual appends via `sb` should ended on `\n`. Why can it not be just a simple sentence with a dot? Why new lines? A simple line is just fine.

.build/cassandra-deps-template.xml
```diff
@@ -244,8 +244,8 @@
     <!-- end of chronicle-queue -->
 
     <dependency>
-      <groupId>org.fusesource</groupId>
-      <artifactId>sigar</artifactId>
+      <groupId>com.github.oshi</groupId>
+      <artifactId>oshi-core</artifactId>
```
smiklosovic, line 248: this also fetches this into the project:

[resolver:resolve] Downloading https://repo1.maven.org/maven2/org/slf4j/slf4j-api/2.0.9/slf4j-api-2.0.9.pom
[resolver:resolve] Downloaded https://repo1.maven.org/maven2/org/slf4j/slf4j-api/2.0.9/slf4j-api-2.0.9.pom (3 KB at 152.8 KB/sec)
[resolver:resolve] Downloading https://repo1.maven.org/maven2/org/slf4j/slf4j-parent/2.0.9/slf4j-parent-2.0.9.pom
[resolver:resolve] Downloaded https://repo1.maven.org/maven2/org/slf4j/slf4j-parent/2.0.9/slf4j-parent-2.0.9.pom (16 KB at 947.1 KB/sec)
[resolver:resolve] Downloading https://repo1.maven.org/maven2/org/slf4j/slf4j-bom/2.0.9/slf4j-bom-2.0.9.pom
[resolver:resolve] Downloaded https://repo1.maven.org/maven2/org/slf4j/slf4j-bom/2.0.9/slf4j-bom-2.0.9.pom (5 KB at 300.3 KB/sec)

I am not sure we should depend on this. Maybe we should exclude it?

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+        if (oshi.SystemInfo.getCurrentPlatform() == PlatformEnum.LINUX) {
+            Path p =  FileSystems.getDefault().getPath("/proc", Long.toString(getPid()), "limits");
+            try
+            {
+                List<String> lines = Files.readAllLines(p);
+                for (String s : lines)
+                {
+                    if (s.startsWith("Max processes"))
+                    {
+                        String[] parts = s.split("\s+");
```
smiklosovic: `\s+` does not compile. It should be `\\s+` if you meant that.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    /**
+     * Gets the maximum number of processes the user can create.
+     * Note: if not on a Linux system this always return the
+     * @{code DEFAULT_MAX_PROCESSES} value.
+     * @return The maximum number of processes.
+     * @see #DEFAULT_MAX_PROCESSES
+     */
+    long getMaxProcess() {
```
smiklosovic: I am not sure why these methods are all package protected. If you want to have them visible, why in this package only? I would just make them public.

It is also questionable if the resolved value should be cached. However, I think it is possible to change the limits during runtime, no? So it might be possible to get different values each time you call this. Or is that a constant set upon processes start?

src/java/org/apache/cassandra/utils/SystemInfo.java
smiklosovic: this might be final too.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+        {
+            sb.append("Amount of available address space should be >= ").append(EXPECTED_AS).append("\n");
+            degraded = true;
+        }
+        if (!valid(getMaxOpenFiles(), EXPECTED_MIN_NOFILE))
+        {
+            sb.append("Minimum value for max open files should be >=  ").append(EXPECTED_MIN_NOFILE).append("\n");
+            degraded = true;
+        }
+        if (degraded) {
```
smiklosovic: I dont think it was ever a good idea to log it in this method. This method should just check the degradations and return messages. The logging logic should be in the check itself.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,214 @@
...
+     * Tests if the system is running in degraded mode.
+     * If the system is running in degraded mode this method will log information.
+     * @return @{code true} if the system is in degraded mode, @{code false} otherwise.
+     */
+    public boolean warnIfRunningInDegradedMode()
+    {
+        boolean degraded = false;
+        StringBuilder sb = new StringBuilder();
+
+        if (oshi.SystemInfo.getCurrentPlatform() == PlatformEnum.LINUX)
```
smiklosovic: I find this logic complicated. Ifs, elses ... degraded set to true etc ... We can do better.

src/java/org/apache/cassandra/service/StartupChecks.java
```diff
@@ -357,14 +357,26 @@ public void execute(StartupChecksOptions options) throws StartupException
...
             if (options.isDisabled(getStartupCheckType()))
                 return;
-            new SystemInfo().warnIfRunningInDegradedMode();
+
+            Optional<String> degradations = new SystemInfo().isDegraded();
+
+            if (degradations.isPresent())
+            {
+                String message = "Cassandra server running in degraded mode. " + degradations.get();
+                if (CassandraRelevantProperties.TEST_IGNORE_PROCESS_ENVIRONMENT_CHECK.getBoolean() || CassandraRelevantProperties.TEST_IGNORE_PROCESS_ENVIRONMENT_CHECK.getBoolean())
```
smiklosovic: @Claudenw there are two same checks in this if

src/java/org/apache/cassandra/utils/SystemInfo.java
Claudenw: I was looking for a class name for Cassandra that described what the class did.  SystemInfo seem like the proper name, I don't have a better one.

src/java/org/apache/cassandra/utils/SystemInfo.java
Claudenw: Static and final will keep the logger around after the class is finished being used.  This class is currently only used during a test at startup.  Hanging on to the memory does not make sense.  In addition, I recall an article a while back that indicated the timing between the static and non static calls were negligible.  I think the non static version is proper in this case.

src/java/org/apache/cassandra/service/StartupChecks.java
Claudenw: The more I look at this the more I think it is wrong.  The original code logged a warning if the system was running in degraded mode OR the flag was set.  Since the flag is only set in a test I think that perhaps we could remove the flag all together.    I will explore this today and see if there is something more reasonable to do here.  But this solution is incorrect as it will not log anything if there is no degradation and the flag is set.

smiklosovic: @Claudenw ah yes, I see that. Good point. The original code logs the message even there are no degradations. I do not see any value in it. I think this should be just simplified like 

`if (degradations.isPresent) logger.warn(message) else logger.info("optimal performance")`

That also means that we will not introduce any new system property and we can remove the old one.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,211 @@
...
+        {
+            check.get().map(sb::append);
+        }
+
+        String message = sb.toString();
+        return message.isEmpty() ? empty() : of(message);
+    }
+
+    /**
+     * Checks if a value if valide (i.e. value >= min or value == INFINITY..
```
smiklosovic: @Claudenw there is still messed up end of this line. Should not it end with `INFINITY)` ? Also `valide` should be `valid`. and `if valide` should be `is valid`. Basically, this whole javadoc should reflect the inverted "invalid" logic with the example code as well.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,211 @@
...
+     * @param value the value to check.
+     * @param min the minimum value.
+     * @return true if value is valid.
```
smiklosovic: @Claudenw `return true if value is invalid`

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,211 @@
...
+
+    /**
+     * @return the PID of the current system.
+     */
+    public long getPid() {
+        return si.getOperatingSystem().getProcessId();
+    }
+
+    /**
+     * Tests if the system is running in degraded mode.
```
smiklosovic, line 182: @Claudenw This javadoc is not accurate anymore, it will return possible degradations as one sentence and it is not logging them.

src/java/org/apache/cassandra/service/StartupChecks.java
Claudenw: I checked, and can find only one place in the tests where TEST_IGNORE_SIGAR is set and no place where any of the expected text is searched for.  I think it was used to report the status even if there was no issue so that it would be in the logs in case there was some data that could be used for debugging tests.

I think that we should remove the flag and any processing based on the flag.  If a need arises we can adjust the visibility of SystemInfo methods or add new methods as necessary.

smiklosovic: python dtests are not checking it either. +1 on the flag removal

Claudenw:
Should the startup check log a message if it is disabled?

```        public void execute(StartupChecksOptions options)
        {
            if (options.isDisabled(getStartupCheckType()))
                return;

            Optional<String> degradations = new SystemInfo().isDegraded();

            if (degradations.isPresent())
                logger.warn("Cassandra server running in degraded mode. " + degradations.get());
            else
                logger.info("Checked OS settings and found them configured for optimal performance.");
        }
```

.build/parent-pom-template.xml
```diff
@@ -990,13 +990,21 @@
         <scope>provided</scope>
       </dependency>
       <dependency>
-        <groupId>org.fusesource</groupId>
-        <artifactId>sigar</artifactId>
-        <version>1.6.4</version>
+        <groupId>com.github.oshi</groupId>
+        <artifactId>oshi-core</artifactId>
+        <version>6.4.6</version>
```
smiklosovic: update to 6.4.8

src/java/org/apache/cassandra/service/StartupChecks.java
```diff
@@ -136,7 +136,7 @@ public enum StartupCheckType
                                                                       checkJMXProperties,
                                                                       inspectJvmOptions,
                                                                       checkNativeLibraryInitialization,
-                                                                      initSigarLibrary,
+                                                                      checkProcessEnvironment,
```
smiklosovic, line 140: there is `initSigarLibrary` string in the comment of this class. You need to replace it with this check.

src/java/org/apache/cassandra/utils/SigarLibrary.java
```diff
@@ -1,187 +0,0 @@
...
-
-    private Sigar sigar;
-    private FileSystemMap mounts = null;
-    private boolean initialized = false;
-    private long INFINITY = -1;
-    private long EXPECTED_MIN_NOFILE = 10000l; // number of files that can be opened
-    private long EXPECTED_NPROC = 32768l; // number of processes
-    private long EXPECTED_AS = INFINITY; // address space
-
-    // TODO: Determine memlock limits if possible
```
smiklosovic, line 41: you have forgotten to move these todos to new implementation

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,210 @@
...
+{
+
+    private static final long INFINITY = -1l;
+    private static final long EXPECTED_MIN_NOFILE = 10000l; // number of files that can be opened
+    private static final long EXPECTED_NPROC = 32768l; // number of processes
+    private static final long EXPECTED_AS = 0x7FFFFFFFl; // address space
+    /**
+     * The default number of processes that are reported if the actual value can not be retrieved.
+     */
+    public static final long DEFAULT_MAX_PROCESSES = 1024;
```
smiklosovic: I would move public constants as first in the class.

src/java/org/apache/cassandra/utils/SystemInfo.java
smiklosovic: I dont know about this ... do we ever use this pattern elsewhere in the code? Seems like we are introducing something special and custom.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    /**
+     * Gets the maximum number of processes the user can create.
+     * Note: if not on a Linux system this always return the
+     * @{code DEFAULT_MAX_PROCESSES} value.
+     * @return The maximum number of processes.
+     * @see #DEFAULT_MAX_PROCESSES
+     */
+    long getMaxProcess()
```
smiklosovic: why are all these methods package protected and not public?

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,210 @@
...
+    {
+        // this check only works on Linux systems.
+        if (oshi.SystemInfo.getCurrentPlatform() == PlatformEnum.LINUX)
```
smiklosovic: I would do a check and returned when platform is not Linux. Then all other logic would not be wrapped in `if` which reduces complexity.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,210 @@
...
+    /* The oshi.SystemInfo has  the following note:
+     * Platform-specific Hardware and Software objects are retrieved via memoized suppliers. To conserve memory at the
+     * cost of additional processing time, create a new version of SystemInfo() for subsequent calls. To conserve
+     * processing time at the cost of additional memory usage, re-use the same {@link SystemInfo} object for future
+     * queries.
+     *
+     * We are opting for minimal memory footprint. */
+    private oshi.SystemInfo si ;
+
+    public SystemInfo() {
```
smiklosovic: braces on new line

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,210 @@
...
+    {
+        // ulimit -H -n
+        return si.getOperatingSystem().getCurrentProcess().getHardOpenFileLimit();
+    }
+
+    /**
+     * Gets the Virtual Memory Size (VSZ). Includes all memory that the process can access, including memory that is swapped out and memory that is from shared libraries.
+     * @return The amount of virtual memory allowed to be allocatedby the current process/user.
+     */
+    long getVirtualMemoryMax() {
```
smiklosovic: braces on new line

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,210 @@
...
+        return si.getOperatingSystem().getCurrentProcess().getVirtualSize();
+    }
+
+    /**
+     *
+     * @return The amount of swap space allocated on the system.
+     */
+    long getSwapSize() {
```
smiklosovic: braces on new line

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,210 @@
...
+        return si.getHardware().getMemory().getVirtualMemory().getSwapTotal();
+    }
+
+    /**
+     * @return the PID of the current system.
+     */
+    public long getPid() {
```
smiklosovic: braces on new line

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,210 @@
...
+                                                                ? of(format("Amount of available address space should be >= %s. ", EXPECTED_AS))
+                                                                : empty();
+
+        Supplier<Optional<String>> expectedMinNoFile = () -> invalid(getMaxOpenFiles(), EXPECTED_MIN_NOFILE)
+                                                             ? of(format("Minimum value for max open files should be >= %s. ", EXPECTED_MIN_NOFILE))
+                                                             : empty();
+
+        StringBuilder sb = new StringBuilder();
+
+        for (Supplier<Optional<String>> check : ImmutableList.of(expectedNumProc,
```
smiklosovic: From Java 9, this might be further simplified to `List.of`, I would also make it one line, not 4, and removed braces around this `for` as `check.get().map(sb::append);` is the only logic in it.

smiklosovic:
@Claudenw you also need to take care of this:

https://github.com/instaclustr/cassandra/commit/5aac06b7020c5884943b3b7b72a0cf494b491a4f#diff-b04a7dc5d2948794698a86fc8df005db9cf8c94d9dbc900f7613b13b0b7d1649

https://github.com/instaclustr/cassandra/commit/5aac06b7020c5884943b3b7b72a0cf494b491a4f#diff-b04a7dc5d2948794698a86fc8df005db9cf8c94d9dbc900f7613b13b0b7d1649

https://github.com/instaclustr/cassandra/commit/5aac06b7020c5884943b3b7b72a0cf494b491a4f#diff-32a8b4fa02cda1ef5a7216fe68d70d23a9784a4bb33b1230d2ec776140dae3d7

Claudenw:
@smiklosovic I rebased on trunk and picked up the Harry changes.  What do we need to do to move this forward?

src/java/org/apache/cassandra/utils/SigarLibrary.java
Claudenw: TODOs added

src/java/org/apache/cassandra/service/StartupChecks.java
```diff
@@ -417,14 +417,26 @@ public void execute(StartupChecksOptions options) throws StartupException
...
+            if (degradations.isPresent())
+            {
+                String message = "Cassandra server running in degraded mode. " + degradations.get();
+                if (CassandraRelevantProperties.TEST_IGNORE_PROCESS_ENVIRONMENT_CHECK.getBoolean())
+                    logger.warn(message);
+                else
+                    logger.info(message);
+            }
+            else
+                logger.info("Checked OS settings and found them configured for optimal performance.");
```
michaelsembwever, line 429: nit: why?  we didn't need this before.

ide/nbproject/project.xml
```diff
@@ -7,7 +7,7 @@
...
...0.jar:${project.dir}/build/test/lib/jars/mockito-inline-4.7.0.jar:${project.dir}/build/test/lib/jars/objenesis-3.2.jar:${project.dir}/build/test/lib/jars/org.jacoco.agent-0.8.8.jar:${project.dir}/build/test/lib/jars/org.jacoco.ant-0.8.8.jar:${project.dir}/build/test/lib/jars/org.jacoco.core-0.8.8.jar:${project.dir}/build/test/lib/jars/org.jacoco.report-0.8.8.jar:${project.dir}/build/test/lib/jars/picocli-4.7.4.jar:${project.dir}/build/test/lib/jars/quicktheories-0.26.jar:${project.dir}/build/test/lib/jars/randomizedtesting-runner-2.1.2.jar:${project.dir}/build/test/lib/jars/reflections-0.10.2.jar:${project.dir}/build/test/lib/jars/semver4j-3.1.0.jar:${project.dir}/build/test/lib/jars/simulator-asm.jar:${project.dir}/build/test/lib/jars/simulator-bootstrap.jar:${project.dir}/build/test/lib/jars/slf4j-api-1.7.32.jar:${project.dir}/build/test/lib/jars/wiremock-jre8-2.35.0.jar:${project.dir}/build/test/lib/jars/xmlresolver-5.1.2-data.jar:${project.dir}/build/test/lib/jars/xmlresolver-5.1.2.jar:${project.dir}/build/test/lib/jars/xmlunit-core-2.9.0.jar:${project.dir}/build/test/lib/jars/xmlunit-legacy-2.9.0.jar:${project.dir}/build/test/lib/jars/xmlunit-placeholders-2.9.0.jar:</property>
```
michaelsembwever: was this change create by using the `ide/nbproject/update-netbeans-classpaths.sh` script ?

Claudenw:
@michaelsembwever short answer is: no,  I did not use the script.  I have fixed the issues.

As I noted in Slack (and here for public record) 

I didn't know that what it was for.  I will have to update the 'change dependencies' documentation I recently submitted.

In addition when I tried to run the script on a `Linux 6.2.0-39-generic #40~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 16 10:53:04 UTC 2 x86_64 x86_64 x86_64 GNU/Linux` system it failed with 2 errors 
`sed: can't read s/cassandra\.classpath\.jars\">.*<\/property>/cassandra\.classpath\.jars\">NEW_CLASSPATH<\/property>/: No such file or directory` and a much longer one from the next line with the CLASSPATH but same `No such file or directory` error.  

The SED comands in the script have blank files for script names (the single quotes after the -i as noted below:
```
sed -i '' 's/cassandra\.classpath\.jars\">.*<\/property>/cassandra\.classpath\.jars\">NEW_CLASSPATH<\/property>/' $DIR/project.xml
sed -i '' "s@NEW_CLASSPATH@"$CLASSPATH"@" $DIR/project.xml
```

If I remove the single quotes it works on my system but I am reticent to include this in this fix.

ide/nbproject/project.xml
michaelsembwever: > https://github.com/apache/cassandra/pull/2842#issuecomment-1874990481

Please use the script.  This file isn't intended to be manually edited.

The failures are sed and macos related.  That completely slipped by me, thanks for catching it.

This is fixed by changing the two sed lines to
```
sed 's/cassandra\.classpath\.jars\">.*<\/property>/cassandra\.classpath\.jars\">NEW_CLASSPATH<\/property>/' $DIR/project.xml > $DIR/project.xml.tmp
mv $DIR/project.xml.tmp $DIR/project.xml
sed "s@NEW_CLASSPATH@"$CLASSPATH"@" $DIR/project.xml > $DIR/project.xml.tmp
mv $DIR/project.xml.tmp $DIR/project.xml
```

Can you too please update the dependencies docs accordingly 🙏🏼

Claudenw: It is also possible to fix the script by simply removing the `''`' as in:
```
sed -i 's/cassandra\.classpath\.jars\">.*<\/property>/cassandra\.classpath\.jars\">NEW_CLASSPATH<\/property>/' $DIR/project.xml
sed -i "s@NEW_CLASSPATH@"$CLASSPATH"@" $DIR/project.xml
```
While this looks cleaner to my eye, I am not certain this works across all platforms.

michaelsembwever: then it won't work on macos

src/java/org/apache/cassandra/service/StartupChecks.java
smiklosovic: @michaelsembwever I see same log here

https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/utils/SigarLibrary.java#L179

If we are all good, we just print this, as it was before.

On the other hand, I am not sure what is the logic behind if / else above and logging on `warn` or `info` level depending on what `TEST_IGNORE_PROCESS_ENVIRONMENT_CHECK` is set to. That seems to be just unnecessary distinction here. I would say that we should always log it on `warn`.  cc @Claudenw

michaelsembwever: Cool, thanks for pointing it out.

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,223 @@
...
+    { 
+        return new Semver(si.getOperatingSystem().getVersionInfo().getBuildNumber(), Semver.SemverType.LOOSE);
+    }
+
+    /**
+     * Tests if the system is running in degraded mode.
+     *
+     * @return non-empty optional with degradation messages if the system is in degraded mode, empty optional otherwise.
+     */
+    public Optional<String> isDegraded()
```
smiklosovic, line 186: @Claudenw I was thinking about adding some tests for this. As of now, we just don't test the logic in this method. SystemInfo might be mocked by Mockito on its methods to return whatever we want instead of oshi returning it and then we could make assertions on various corner cases and values so we might watch what degradation string we resolve here.

Another option is to not use Mockito but we might just extend this class and override its methods, returning whatever we want. Then we might do assertions on `isDegraded` output.

ide/nbproject/project.xml
```diff
@@ -7,7 +7,7 @@
...
...ne-4.7.0.jar:${project.dir}/build/test/lib/jars/objenesis-3.2.jar:${project.dir}/build/test/lib/jars/org.jacoco.agent-0.8.8.jar:${project.dir}/build/test/lib/jars/org.jacoco.ant-0.8.8.jar:${project.dir}/build/test/lib/jars/org.jacoco.core-0.8.8.jar:${project.dir}/build/test/lib/jars/org.jacoco.report-0.8.8.jar:${project.dir}/build/test/lib/jars/picocli-4.7.4.jar:${project.dir}/build/test/lib/jars/quicktheories-0.26.jar:${project.dir}/build/test/lib/jars/randomizedtesting-runner-2.1.2.jar:${project.dir}/build/test/lib/jars/reflections-0.10.2.jar:${project.dir}/build/test/lib/jars/Saxon-HE-12.2.jar:${project.dir}/build/test/lib/jars/semver4j-3.1.0.jar:${project.dir}/build/test/lib/jars/simulator-asm.jar:${project.dir}/build/test/lib/jars/simulator-bootstrap.jar:${project.dir}/build/test/lib/jars/slf4j-api-1.7.32.jar:${project.dir}/build/test/lib/jars/wiremock-jre8-2.35.0.jar:${project.dir}/build/test/lib/jars/xmlresolver-5.1.2-data.jar:${project.dir}/build/test/lib/jars/xmlresolver-5.1.2.jar:${project.dir}/build/test/lib/jars/xmlunit-core-2.9.0.jar:${project.dir}/build/test/lib/jars/xmlunit-legacy-2.9.0.jar:${project.dir}/build/test/lib/jars/xmlunit-placeholders-2.9.0.jar:</property>
```
jacek-lewandowski, line 11: I'm not a netbeans user - can't we use a wildcard?

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -567,7 +567,7 @@ public enum CassandraRelevantProperties
      * can be also done manually for that particular case: {@code flush(SchemaConstants.SCHEMA_KEYSPACE_NAME);}. */
     TEST_FLUSH_LOCAL_SCHEMA_CHANGES("cassandra.test.flush_local_schema_changes", "true"),
     TEST_HARRY_SWITCH_AFTER("cassandra.test.harry.progression.switch-after", "1"),
-    TEST_IGNORE_SIGAR("cassandra.test.ignore_sigar"),
```
jacek-lewandowski, line 570: where was that used?

src/java/org/apache/cassandra/utils/SystemInfo.java
```diff
@@ -0,0 +1,223 @@
...
+    {
+        Supplier<String> expectedNumProc = () -> {
+            // only check proc on nproc linux
+            if (oshi.SystemInfo.getCurrentPlatform() == PlatformEnum.LINUX)
```
smiklosovic: @Claudenw we can use `platform` method of this class instead of `oshi.SystemInfo.getCurrentPlatform()` and `platform` might return `PlatformEnum` instead of `String`.

That way, except reusing the code, it will also unblocks us when we will go to mock what this class is doing when testing various scenarios.

src/java/org/apache/cassandra/service/StartupChecks.java
jacek-lewandowski: I agree with you guys, I think we should just print warn if it is running in degraded mode and that's it. No info/warn distinction, no positive message when everything is ok.

jacek-lewandowski:
It looks ok, I have one general comment - I think that something like `SystemInfo` should have a kind of a provider so that we can easily mock it in tests. Maybe something similar to the clock provider.

ide/nbproject/project.xml
Claudenw: as in `<property name="cassandra.classpath.jars">${project.dir}/build/lib/jars/*</property>` ?

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
smiklosovic: @jacek-lewandowski 

in `SigarLibrary#warnIfRunningInDegradedMode` as well as in `SimulationRunner#beforeAll`.

Honestly, I don't see any reason for its usage in `warnIfRunningInDegradedMode`. If you check that logic, it does not make sense. Imagine that all checks pass. So the warning will be logged when `TEST_IGNORE_SIGAR` is set to true? Why would one want that?

https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/utils/SigarLibrary.java#L171

ide/nbproject/project.xml
jacek-lewandowski: If it is feasible, why not?

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
jacek-lewandowski: Remove that!

src/java/org/apache/cassandra/service/StartupChecks.java
smiklosovic: @jacek-lewandowski  @Claudenw 

I am fine with informational log when all is fine as it was before (otherwise we are changing the behavior here). Maybe  somebody parses their logs on this message and we will remove it so they will have nothing to grep on?

Also, if we indeed want to skip the system check, how would we go around this? I think that `TEST_IGNORE_PROCESS_ENVIRONMENT_CHECK` in its current form is not doing anything useful. It is not like if it is set to true, the check itself would be skipped ... 

Some time ago, I made startup checks to be configurable (CASSANDRA-17220). So if we indeed want to make this check skippable, we would need to change its StartupCheckType so this

            if (options.isDisabled(getStartupCheckType()))
                return;

will evaluate to `true`.

You could do this in cassandra.yaml then

        startup_checks:
            check_process_environment:
                enabled: false

So, if we do not want to ever skip this check, then whole TEST_IGNORE... property should be just removed. If we want to have a possibility to skip it, then we should make it configurable to set `enabled: false`. We could also introduce a system property (instead of `TEST_IGNORE...` and return from check's execution when it is `true`. That has same functionality as the change in cassandra.yaml but it is easier to set it e.g. for some tests when we do not need to mess with setting that in some testing cassandra.yaml file. Currently, I do not see we are in fact skipping this check in tests so this capability could be just omitted.

jacek-lewandowski: I consider this is just something which shouldn't be even a startup check because it does not stop the execution in certain cases. This is also a new Cassandra version so behaviour changes are ok. This adds extra complexity, extra log messages and I cannot really see any value in that. I think we should limit the log messages visible for the user rather than extending them and printing obvious information

ide/nbproject/project.xml
michaelsembwever: https://github.com/apache/netbeans/issues/6044

in the meantime we could add a simple test check that running this script does not change the project.xml

src/java/org/apache/cassandra/service/StartupChecks.java
smiklosovic: @jacek-lewandowski 

I think that making this a startup check is just fine for now. Let's just rewrite the guts of it for now ... There is as lot of startup checks which are not preventing a node from starting. For example `checkReadAheadKbSetting` just (at most) logs warnings ... so based on that logic, that one does not qualify for being a startup check either. Similar to `checkLz4Native` etc ...

I think for now this is just fine. We could then probably consolidate these checks so they indeed stop the bootstrap if they fail and we could extract the rest to some other logic. I think that we are dealing primarily with "how things were".

smiklosovic: @Claudenw  @jacek-lewandowski 

To wrap it up, I think whole logic might be just replaced by this: 

    new SystemInfo().isDegraded()
                    .ifPresent(msgs -> logger.warn("Cassandra server running in degraded mode. " + msgs));

and we remove `TEST_IGNORE....` altogether.

jacek-lewandowski: I think it should not be even a startup check

smiklosovic: @jacek-lewandowski 

Then we need to remove all these checks as well and find new place for them with whole machinery around that.

checkJemalloc
checkLz4Native
checkJMXPorts
checkJMXProperties
inspectJvmOptions
checkMaxMapCount
checkReadAheadKbSetting

Why not to just do it as proposed here and then we can elaborate on that in a different ticket?

**Keep in mind that we are rewriting the check, we are not implementing a new one.** This will save us couple MBs from the tarball too.

I understand your point but extracting this from startup checks is not the objective here at the moment ...

michaelsembwever: > Why not to just do it as proposed here and then we can elaborate on that in a different ticket?
>
> Keep in mind that we are rewriting the check, we are not implementing a new one. 

yes.  startup check for this ticket works for me.  let's not drag the ticket on forever.

jacek-lewandowski: ok, this is not objective of this ticket, indeed

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
Claudenw: It looks like it was originally used in
```
src/java/org/apache/cassandra/config/CassandraRelevantProperties.java:    TEST_IGNORE_SIGAR("cassandra.test.ignore_sigar", "false"),
src/java/org/apache/cassandra/utils/SigarLibrary.java:            if (swapEnabled || !goodAddressSpace || !goodFileLimits || !goodProcNumber || CassandraRelevantProperties.TEST_IGNORE_SIGAR.getBoolean())
test/simulator/main/org/apache/cassandra/simulator/SimulationRunner.java:import static org.apache.cassandra.config.CassandraRelevantProperties.TEST_IGNORE_SIGAR;
test/simulator/main/org/apache/cassandra/simulator/SimulationRunner.java:        TEST_IGNORE_SIGAR.setBoolean(true);
```

src/java/org/apache/cassandra/service/StartupChecks.java
Claudenw: Reading through this is looks like the consensus is that the message logging should read something like:

```

            Optional<String> degradations = new SystemInfo().isDegraded();

            if (degradations.isPresent())
                logger.warn("Cassandra server running in degraded mode. " + degradations.get());
            else
                logger.info("Checked OS settings and found them configured for optimal performance.");
```

Keeping both the warning and the info message.

ide/nbproject/project.xml
Claudenw: I added another comment below the one shown above that reads: `<!-- DO NOT EDIT THE FOLLOWING LINE DIRECTLY: Use the update-netbeans-classpath.sh script -->` so that my mistake should not be duplicated in the future.

src/java/org/apache/cassandra/service/StartupChecks.java
smiklosovic: Jacek wants to get rid of that info log as well. That is how I read it. I do not mind if it stays or not.

src/java/org/apache/cassandra/utils/SystemInfo.java
Claudenw: Switching to returning the PlatformEnum exposes the underlying classes of the SystemInfo class.  This will make it more complex to replace the functionality later.  I think that SystemInfo should isolate the developer from the underlying classes that provide the functionality.  We could make SystemInfo an interface and move the current code to a SystemInfoImpl class, then @jacek-lewandowski  suggestion (below) that  there be a provider, and your observation that Mockito might be used become easier to implement.

src/java/org/apache/cassandra/service/StartupChecks.java
jacek-lewandowski: yes, I'm -1 on keeping info; If we really need something like that, we should have a generic mechanism which would just say: "Running startup check <name>" for each check consistently (out of scope).

Claudenw:
The latest changes DO NOT implement a provider, and do not expose the underlying SystemInfo implementation.

I still need some clarity on whether or not SystemInfo should be an interface with the current code as an implementation.

jacek-lewandowski:
@Claudenw I'd say it is up to you because SystemInfo looks like something easy to mock.

src/java/org/apache/cassandra/utils/SystemInfo.java
smiklosovic: @Claudenw still, use that `platform()` method here and make it return String (as it is now) and just call `Platform.LINUX.name()` here and compare it with `equals`.

smiklosovic:
@Claudenw I dont think we need an interface for this. It is not like it is a hobby to change the guts of this every quarter or some custom impls would be necessary to justify an interface for that. I think that SystemInfo, as is, is mockable already. cc @jacek-lewandowski

jacek-lewandowski:
yup, just as said, I'm fine with either approach

src/java/org/apache/cassandra/utils/SystemInfo.java
smiklosovic: @Claudenw one nice advantage of returning enum is that you can use it in `case` and it is compile time safe ... We can just compare enums directly. I do not think it is a lot to ask for given how comfortable its usage will be. We are not planning to change the impl anytime soon again.

I can imagine a lot of people would ask "what platform I am at" quite frequently and dealing with "equals" and Linux string etc every time is ... nah.

src/java/org/apache/cassandra/service/StartupChecks.java
Claudenw: This would be an change in behavior and we are all OK  with that?  I want to verify before removing-  Personally, I think it should stay.

ide/nbproject/project.xml
Claudenw: Script was updated as per  @michaelsembwever  soluiton.

ide/nbproject/project.xml
Claudenw: No simple solution to this problem at this time.  https://github.com/apache/netbeans/issues/6044

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
Claudenw: Removed

src/java/org/apache/cassandra/service/StartupChecks.java
smiklosovic: I am for keeping it there until we find more robust way how to inform a user what startup check is being run. I can imagine all would log they are executed on DEBUG or similar, then this  log can go away too.

michaelsembwever: > This would be an change in behavior and we are all OK with that? I want to verify before removing- Personally, I think it should stay.

please keep existing behaviour (don't remove).
we never know who's using the logs for want, there needs to be good reason to change.

src/java/org/apache/cassandra/service/StartupChecks.java
```diff
@@ -417,20 +417,20 @@ public void execute(StartupChecksOptions options) throws StartupException
...
                 return;
-            SigarLibrary.instance.warnIfRunningInDegradedMode();
+            new SystemInfo().isDegraded().ifPresent( s -> logger.warn("Cassandra server running in degraded mode. " + s));
         }
     };
 
     public static final StartupCheck checkReadAheadKbSetting = new StartupCheck()
     {
-        // This value is in KB.
+        // This value is in KB.git
```
smiklosovic: @Claudenw typo?

src/java/org/apache/cassandra/service/StartupChecks.java
```diff
@@ -417,20 +417,20 @@ public void execute(StartupChecksOptions options) throws StartupException
...
-    public static final StartupCheck initSigarLibrary = new StartupCheck()
+    public static final StartupCheck checkProcessEnvironment = new StartupCheck()
     {
         @Override
         public void execute(StartupChecksOptions options)
         {
             if (options.isDisabled(getStartupCheckType()))
                 return;
-            SigarLibrary.instance.warnIfRunningInDegradedMode();
+            new SystemInfo().isDegraded().ifPresent( s -> logger.warn("Cassandra server running in degraded mode. " + s));
```
smiklosovic: @Claudenw was not the deal to keep info log there? https://github.com/apache/cassandra/pull/2842#discussion_r1441646580

src/java/org/apache/cassandra/utils/SystemInfo.java
smiklosovic: @Claudenw do you plan to use `platform()` here?

Claudenw: No, it is an internal call, so no need to go though the method.

src/java/org/apache/cassandra/service/StartupChecks.java
Claudenw: fixed

src/java/org/apache/cassandra/service/StartupChecks.java
Claudenw: I made the change too soon.  I have reverted it.

src/java/org/apache/cassandra/utils/SystemInfo.java
smiklosovic: @Claudenw how do you want to mock the operating system this is run on when testing various scenarios? If we used `platform()` here then we could mock it returns something else from "Linux".

jacek-lewandowski:
@Claudenw will you add a provider for SystemInfo?

Claudenw:
@michaelsembwever The dev documentation update mentioning the `ide/nbproject/update-netbeans-classpaths.sh` has been submitted as a pull request to @aratno change for CASSANDRA-17750 (https://github.com/apache/cassandra-website/pull/170).  I expect that will be updated shortly and perhaps we can get some focus on it to get it merged.

Claudenw:
The provider has been added.  I think this now implements all agreed suggestions from the reviews.

jacek-lewandowski:
Please at least memoize the value from the default provider. Also name `instance` does not suggest a provider - would be good to replace it with `get` or something like that.

jacek-lewandowski (approved):

PR #2843, opened 2023-10-25
Cassandra 18961 - record commit timestamps

jacek-lewandowski:


src/java/org/apache/cassandra/tcm/AbstractLocalProcessor.java
```diff
@@ -53,19 +54,22 @@ public final Commit.Result commit(Entry.Id entryId, Transformation transform, fi
...
                 !transform.allowDuringUpgrades() &&
-                previous.metadataSerializationUpgradeInProgress())
+                prevCommitted.metadata.metadataSerializationUpgradeInProgress())
             {
                 result = new Transformation.Rejected(ExceptionCode.INVALID, "Upgrade in progress, can't commit " + transform);
             }
             else
             {
-                result = transform.execute(previous);
+                while ((timestampMicros = nextUnixMicros()) <= prevCommitted.timestampMicros);
```
jakubzytka, line 71: Could we do something better than active wait? like sleep?
perhaps it makes sense to `WARN` if the difference is too big?

jacek-lewandowski: Indeed, we can either wait or assume the previous epoch timestamp + 1 in case the current timestamp is lower than the last epoch timestamp. In both cases we can warn that clocks of CMS nodes have diverged too much.

PR #2844, opened 2023-10-25
CASSANDRA-18962: Add retries to IR messages

dcapwell:


src/java/org/apache/cassandra/metrics/RepairMetrics.java
```diff
@@ -18,17 +18,75 @@
 
 package org.apache.cassandra.metrics;
 
+import java.util.Collections;
```
dcapwell: This is from https://issues.apache.org/jira/browse/CASSANDRA-18952 and no changes are from this PR... can ignore this file

src/java/org/apache/cassandra/repair/SharedContext.java
```diff
@@ -59,6 +59,16 @@ public interface SharedContext
     ScheduledExecutorPlus optionalTasks();
 
     MessageDelivery messaging();
+    default SharedContext withMessaging(MessageDelivery messaging)
```
dcapwell, line 62: This is from https://issues.apache.org/jira/browse/CASSANDRA-18952 and no changes are from this PR... can ignore this file

src/java/org/apache/cassandra/repair/RepairMessageVerbHandler.java
```diff
@@ -286,27 +285,28 @@ public void doVerb(final Message<RepairMessage> message)
...
                     break;
 
                 case FINALIZE_COMMIT_MSG:
-                    ctx.repair().consistent.local.handleFinalizeCommitMessage(message.from(), (FinalizeCommit) message.payload);
+                    ctx.repair().consistent.local.handleFinalizeCommitMessage(message);
                     break;
 
                 case FAILED_SESSION_MSG:
                     FailSession failure = (FailSession) message.payload;
+                    sendAck(message);
```
dcapwell, line 309: prob could nack if we don't know the session... ack vs nack really doesn't matter for IR as it doesn't wait for the result... the retries really are just for stability and not for the senders benefit.

src/java/org/apache/cassandra/repair/consistent/CoordinatorSession.java
```diff
@@ -154,34 +160,37 @@ synchronized boolean hasFailed()
...
+        InetAddressAndPort participant = msg.payload.participant;
+        boolean success = msg.payload.success;
         if (getState() == State.FAILED)
         {
             logger.trace("Incremental repair {} has failed, ignoring prepare response from {}", sessionID, participant);
+            sendFailureResponse(ctx, msg);
             return;
         }
+        sendAck(ctx, msg);
+        if (getParticipantState(participant) != State.PREPARING)
```
dcapwell, line 192: this also fixes a IR bug where we see the response *after* we get a fail session request...  in that case we fail and log a stack trace saying we can't go from `FAILURE -> PREPARED`

src/java/org/apache/cassandra/repair/consistent/CoordinatorSession.java
```diff
@@ -218,19 +227,29 @@ public synchronized Future<Void> finalizePropose()
...
+        InetAddressAndPort participant = message.payload.participant;
+        boolean success = message.payload.promised;
         if (getState() == State.FAILED)
         {
             logger.trace("Incremental repair {} has failed, ignoring finalize promise from {}", sessionID, participant);
+            sendFailureResponse(ctx, message);
+            return;
+        }
+        sendAck(ctx, message);
+        if (getParticipantState(participant) != State.REPAIRING)
```
dcapwell, line 249: this also fixes a IR bug where we see the response *after* we get a fail session request...  in that case we fail and log a stack trace saying we can't go from `FAILURE -> FINALIZE_PROMISED`

src/java/org/apache/cassandra/repair/consistent/LocalSessions.java
```diff
@@ -637,10 +642,13 @@ public LocalSession getSession(TimeUUID sessionID)
     }
 
     @VisibleForTesting
-    synchronized void putSessionUnsafe(LocalSession session)
+    synchronized boolean putSessionUnsafe(LocalSession session)
     {
+        if (sessions.containsKey(session.sessionID))
```
dcapwell, line 647: this actually makes it safe... but does have a big change in semantics... for retry logic we need ability to dedup, so blind overriding can't be done... I could always rename this to `putSession`....

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
```diff
@@ -67,11 +86,36 @@ private enum ErrorHandling { NONE, TIMEOUT, RETRY }
...
+        // IR messages
+        map.put(Verb.PREPARE_CONSISTENT_REQ, SUPPORTS_RETRY);
+        map.put(Verb.PREPARE_CONSISTENT_RSP, SUPPORTS_RETRY);
+        map.put(Verb.FINALIZE_PROPOSE_MSG, SUPPORTS_RETRY);
+        map.put(Verb.FINALIZE_PROMISE_MSG, SUPPORTS_RETRY);
+        map.put(Verb.FINALIZE_COMMIT_MSG, SUPPORTS_RETRY);
+        map.put(Verb.FAILED_SESSION_MSG, SUPPORTS_RETRY);
         VERB_TIMEOUT_VERSIONS = Collections.unmodifiableMap(map);
+
+        EnumSet<Verb> allowsRetry = EnumSet.noneOf(Verb.class);
```
dcapwell, line 98: this logic is from https://issues.apache.org/jira/browse/CASSANDRA-18952 but added IR messages as well

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
```diff
@@ -67,11 +86,36 @@ private enum ErrorHandling { NONE, TIMEOUT, RETRY }
...
+        allowsRetry.add(Verb.FINALIZE_PROPOSE_MSG);
+        allowsRetry.add(Verb.FINALIZE_PROMISE_MSG);
+        allowsRetry.add(Verb.FINALIZE_COMMIT_MSG);
+        allowsRetry.add(Verb.FAILED_SESSION_MSG);
+        ALLOWS_RETRY = Collections.unmodifiableSet(allowsRetry);
     }
-    private static final Set<Verb> SUPPORTS_RETRY_WITHOUT_VERSION_CHECK = Collections.unmodifiableSet(EnumSet.of(Verb.CLEANUP_MSG));
 
     private static final Logger logger = LoggerFactory.getLogger(RepairMessage.class);
+    private static final NoSpamLogger noSpam = NoSpamLogger.getLogger(logger, 1, TimeUnit.MINUTES);
```
dcapwell: from https://issues.apache.org/jira/browse/CASSANDRA-18952

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
```diff
@@ -121,27 +165,25 @@ public static <T> void sendMessageWithRetries(SharedContext ctx, RepairMessage r
...
+    public static void sendMessageWithRetries(SharedContext ctx, Supplier<Boolean> allowRetry, RepairMessage request, Verb verb, InetAddressAndPort endpoint)
+    {
+        sendMessageWithRetries(ctx, backoff(ctx, verb), allowRetry, request, verb, endpoint, NOOP_CALLBACK, 0);
     }
 
-    private static <T> void sendMessageWithRetries(SharedContext ctx, Backoff backoff, Supplier<Boolean> allowRetry, RepairMessage request, Verb verb, InetAddressAndPort endpoint, RequestCallback<T> finalCallback, int attempt)
+    @VisibleForTesting
+    static <T> void sendMessageWithRetries(SharedContext ctx, Backoff backoff, Supplier<Boolean> allowRetry, RepairMessage request, Verb verb, InetAddressAndPort endpoint, RequestCallback<T> finalCallback, int attempt)
     {
+        if (!ALLOWS_RETRY.contains(verb))
```
dcapwell: https://issues.apache.org/jira/browse/CASSANDRA-18952

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
```diff
@@ -165,13 +207,32 @@ public void onFailure(InetAddressAndPort from, RequestFailureReason failureReaso
...
                         }
+                        maybeRecordRetry(failureReason);
                         finalCallback.onFailure(from, failureReason);
                         return;
                     default:
                         throw new AssertionError("Unknown error handler: " + allowed);
                 }
             }
 
+            private void maybeRecordRetry(@Nullable RequestFailureReason reason)
```
dcapwell: from https://issues.apache.org/jira/browse/CASSANDRA-18952

src/java/org/apache/cassandra/utils/NoSpamLogger.java
```diff
@@ -51,19 +51,18 @@ public enum Level
     }
 
     @VisibleForTesting
-    static interface Clock
+    public interface Clock
```
dcapwell: this was done in https://issues.apache.org/jira/browse/CASSANDRA-18952; simulation detects that global clock is used and rejects it, so we now use the simulated clock here, which means the no spam logic happens often (simulation time moves much faster than wall clock)

test/unit/org/apache/cassandra/repair/ConcurrentIrWithPreviewFuzzTest.java
```diff
@@ -62,7 +64,7 @@ public void concurrentIrWithPreview()
                 // cause a delay in validation to have more failing previews
                 closeables.add(cluster.nodes.get(pickParticipant(rs, previewCoordinator, preview)).doValidation(next -> (cfs, validator) -> {
                     if (validator.desc.parentSessionId.equals(preview.state.id))
-                        cluster.unorderedScheduled.schedule(() -> next.accept(cfs, validator), 1, TimeUnit.HOURS);
+                        delayValidation(cluster, ir, next, cfs, validator);
```
dcapwell, line 67: this was a flakey issue that this patch discovered... with IR retries we can have IR take longer than 1h, so waiting by time is unsafe and leads to flakey builds... now we check IR state and keep retrying until IR completes

test/unit/org/apache/cassandra/repair/FailingRepairFuzzTest.java
```diff
@@ -114,6 +114,7 @@ public void failingRepair()
                 }
 
                 cluster.processAll();
+                Assertions.assertThat(repair.state.isComplete()).describedAs("Repair job did not complete, and no work is pending...").isTrue();
```
dcapwell, line 117: added to produce a more useful error msg when repairs are not complete... These classes all assume that nothing left to process means the repair is complete, but it could also mean a bug in repair that will never get recovered... in that case the repair is not complete and the error message is hard to understand... so this new error message is trying to express this case

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
```diff
@@ -278,6 +282,12 @@ public void setupSchema()
         createSchema();
     }
 
+    protected void cleanupRepairTables()
+    {
+        for (String table : Arrays.asList(SystemKeyspace.REPAIRS))
```
dcapwell, line 287: IR keeps trying to load the system table, which can slow down the tests... so added this as a work around... it would be great if storage could be isolated and wasn't global state =(

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
```diff
@@ -319,25 +329,16 @@ static void enableMessageFaults(Cluster cluster)
             @Override
             public Set<Faults> apply(Cluster.Node node, Message<?> message)
             {
+                if (RepairMessage.ALLOWS_RETRY.contains(message.verb()))
```
dcapwell: this is from https://issues.apache.org/jira/browse/CASSANDRA-18952

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
```diff
@@ -319,25 +329,16 @@ static void enableMessageFaults(Cluster cluster)
...
-                        return Faults.DROPPED;
                     // these messages are not resilent to ephemeral issues
-                    case PREPARE_CONSISTENT_REQ:
-                    case PREPARE_CONSISTENT_RSP:
-                    case FINALIZE_PROPOSE_MSG:
-                    case FINALIZE_PROMISE_MSG:
-                    case FINALIZE_COMMIT_MSG:
-                    case FAILED_SESSION_MSG:
-
+                    case STATUS_REQ:
```
dcapwell, line 340: these messages actually no-op as `StorageService` is not initialized... they actually cause annoying cases as they fail sessions, so didn't try to get working... just didn't want tests to fail if those messages happen

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
```diff
@@ -799,6 +814,8 @@ private <REQ, RSP> void maybeEnqueue(Message<REQ> message, InetAddressAndPort to
                 CallbackContext cb;
                 if (callback != null)
                 {
+                    if (callbacks.containsKey(message.id()))
```
dcapwell, line 815: this is to better match `MessagingService`, this doesn't happen but better to be consistent

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
```diff
@@ -850,7 +867,14 @@ private <REQ, RSP> void maybeEnqueue(Message<REQ> message, InetAddressAndPort to
...
                             {
                                 assert ctx == cb;
-                                ctx.onFailure(to, RequestFailureReason.TIMEOUT);
+                                try
+                                {
+                                    ctx.onFailure(to, RequestFailureReason.TIMEOUT);
+                                }
+                                catch (Throwable t)
+                                {
+                                    failures.add(t);
```
dcapwell, line 874: when IR messages retry they would NPE as they use a null `desc`... this would get handled by the future and ignored, so tests would fail saying the repair didn't complete rather than the NPE... this fix makes sure we fail fast

test/unit/org/apache/cassandra/repair/messages/RepairMessageTest.java
```diff
@@ -0,0 +1,182 @@
...
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mockito;
+
+import static org.apache.cassandra.repair.messages.RepairMessage.always;
+import static org.apache.cassandra.repair.messages.RepairMessage.sendMessageWithRetries;
+import static org.apache.cassandra.test.asserts.ExtendedAssertions.assertThat;
+
+// Tests may use verb / message pairs that do not make sense... that is due to the fact that the message sending logic does not validate this and delegates such validation to messaging, which is mocked within the class...
+// By using messages with simpler state it makes the test easier to read, even though the verb -> message mapping is incorrect.
+public class RepairMessageTest
```
dcapwell: this is from https://issues.apache.org/jira/browse/CASSANDRA-18952

test/unit/org/apache/cassandra/test/asserts/ExtendedAssertions.java
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.test.asserts;
+
+import com.codahale.metrics.Counting;
+import org.assertj.core.api.AbstractObjectAssert;
+
+public class ExtendedAssertions
```
dcapwell: this is from https://issues.apache.org/jira/browse/CASSANDRA-18952

test/unit/org/apache/cassandra/utils/NoSpamLoggerTest.java
```diff
@@ -80,7 +80,7 @@ public boolean equals(Object o)
     @BeforeClass
     public static void setUpClass() throws Exception
     {
-        NoSpamLogger.CLOCK = () -> now;
+        NoSpamLogger.unsafeSetClock(() -> now);
```
dcapwell: this is from https://issues.apache.org/jira/browse/CASSANDRA-18952

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
```diff
@@ -232,4 +298,15 @@ private static ErrorHandling errorHandlingSupported(SharedContext ctx, InetAddre
             return ErrorHandling.TIMEOUT;
         return ErrorHandling.NONE;
     }
+
+    public static void sendFailureResponse(SharedContext ctx, Message<?> respondTo)
```
bdeggleston, line 302: I think we'll need to check the version of the node we're responding to. Otherwise I don't think they'll be expecting a response for this message

bdeggleston (changes requested):

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
dcapwell: thats fine, we drop in the messaging layer... these are *replies* which require we have Callbacks define for the message number, so the below block is what we hit in 4.x

```
RequestCallbacks.CallbackInfo callbackInfo = MessagingService.instance().callbacks.remove(message.id(), message.from());
        if (callbackInfo == null)
        {
            String msg = "Callback already removed for {} (from {})";
            logger.trace(msg, message.id(), message.from());
            Tracing.trace(msg, message.id(), message.from());
            return;
        }
```

bdeggleston: sure, but why reply if it's just going to be dropped? It's just a cleanliness thing. That said, no one's going to notice a few empty messages getting sent into the void, so I'm fine leaving as is if you'd prefer.

bdeggleston (approved):
+1

src/java/org/apache/cassandra/repair/messages/RepairMessage.java
dcapwell: > but why reply if it's just going to be dropped?

in order to know this we need to check gossip, which maybe stale.  So If we think its 4.x but its actually 5.x then we avoid sending the ack/nack and it will retry... 

Now, since we don't 100% support mixed version repairs, by not caring we make this simpler to reason about with the cost of empty dropped messages while in mixed mode (again, we don't support)... so this patch does not break mixed mode, but isn't 100% ideal in mixed mode...

PR #2863, opened 2023-11-03
[CASSANDRA-18968] StartupClusterConnectivityChecker fails on upgrade from 3.X

isaacreath:
Cassandra versions prior to Cassandra 4 do not support the PING_REQ message type which causes the StartupConnectivityChecker::execute method to fail.
This commit changes the behavior of the StartupConnectivityChecker::execute method to only run a connectivity check if there are no nodes which are running a version prior to Cassandra 4.

patch by Isaac Reath ([ireath@bloomberg.net](mailto:ireath@bloomberg.net)); reviewed by Paulo Motta for [CASSANDRA-18968](https://issues.apache.org/jira/browse/CASSANDRA-18968)

This can be reproduced by starting a Cassandra 4 or greater node where at least one peer is running Cassandra 3 or older. In this scenario you should see:
```WARN [main] 2023-10-27 15:58:22,234 StartupClusterConnectivityChecker.java:183 - Timed out after 10002 milliseconds, was waiting for remaining peers to connect: {dc1=[X.Y.Z.W, A.B.C.D]}.```

After applying this patch, instead you should see:
```Skipping startup connectivity check as some nodes may be running Cassandra version 3 or older which does not support connectivity checking.```

### DIFF
--- a/src/java/org/apache/cassandra/net/StartupClusterConnectivityChecker.java
+++ b/src/java/org/apache/cassandra/net/StartupClusterConnectivityChecker.java
@@ -42,6 +42,7 @@ import org.apache.cassandra.gms.Gossiper;
 import org.apache.cassandra.gms.IEndpointStateChangeSubscriber;
 import org.apache.cassandra.gms.VersionedValue;
 import org.apache.cassandra.locator.InetAddressAndPort;
+import org.apache.cassandra.utils.CassandraVersion;
 import org.apache.cassandra.utils.FBUtilities;
 
 import static org.apache.cassandra.net.Verb.PING_REQ;
@@ -54,6 +55,7 @@ public class StartupClusterConnectivityChecker
 
     private final boolean blockForRemoteDcs;
     private final long timeoutNanos;
+    private final Gossiper gossiper;
 
     public static StartupClusterConnectivityChecker create(long timeoutSecs, boolean blockForRemoteDcs)
     {
@@ -61,14 +63,15 @@ public class StartupClusterConnectivityChecker
             logger.warn("setting the block-for-peers timeout (in seconds) to {} might be a bit excessive, but using it nonetheless", timeoutSecs);
         long timeoutNanos = TimeUnit.SECONDS.toNanos(timeoutSecs);
 
-        return new StartupClusterConnectivityChecker(timeoutNanos, blockForRemoteDcs);
+        return new StartupClusterConnectivityChecker(timeoutNanos, blockForRemoteDcs, Gossiper.instance);
     }
 
     @VisibleForTesting
-    StartupClusterConnectivityChecker(long timeoutNanos, boolean blockForRemoteDcs)
+    StartupClusterConnectivityChecker(long timeoutNanos, boolean blockForRemoteDcs, Gossiper gossiper)
     {
         this.blockForRemoteDcs = blockForRemoteDcs;
         this.timeoutNanos = timeoutNanos;
+        this.gossiper = gossiper;
     }
 
     /**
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/java/org/apache/cassandra/net/StartupClusterConnectivityChecker.java [hunk 1]
```diff
@@ -82,6 +85,16 @@ public boolean execute(Set<InetAddressAndPort> peers, Function<InetAddressAndPor
         if (peers == null || this.timeoutNanos < 0)
             return true;
 
+        // Check if there are any nodes which we know are running a version prior to 4.0.
+        // We use this intead of Gossiper::hasMajorVersion3Nodes because in the absence of version information for a peer
+        // we still prefer to run the startup connectivity check.
+        if (gossiper.isUpgradingFromVersionLowerThan(CassandraVersion.CASSANDRA_4_0))
```
smiklosovic: I do not think that you need to propagate Gossiper to constructor. Just use `Gossiper.instance` here. Gossiper is not an interface or similar that you need to access it like that.

isaacreath: The only reason I added it to the constructor is to let me mock it out in the `StartupClusterConnectivityCheckerTest`. Happy to remove, but that would make testing the component a bit trickier.

smiklosovic: @isaacreath  you might still mock it but there would be one additional parameter to `execute` method which would be a function taking `CassandraVersion` and returning `boolean`:  `(version) -> yourMock.isUpgradingFromVersionLowerThan(version)`.

`execute` already takes a function as its second parameter anyway ...

actually, it would be `java.util.function.Predicate`.

isaacreath: @smiklosovic Updated the PR to have `execute` take this as a predicate.

src/java/org/apache/cassandra/net/StartupClusterConnectivityChecker.java
```diff
@@ -77,11 +77,22 @@ public static StartupClusterConnectivityChecker create(long timeoutSecs, boolean
...
         if (peers == null || this.timeoutNanos < 0)
             return true;
 
+        // Check if there are any nodes which we know are running a version prior to 4.0.
+        // We use this intead of Gossiper::hasMajorVersion3Nodes because in the absence of version information for a peer
+        // we still prefer to run the startup connectivity check.
+        if (isUpgradingFromLowerVersionThan.test(CassandraVersion.CASSANDRA_4_0))
+        {
+            logger.debug("Skipping startup connectivity check as some nodes may be running Cassandra version 3 or older " +
+                        "which does not support connectivity checking.");
```
smiklosovic: would you mind to align this with the string above?

test/unit/org/apache/cassandra/net/StartupClusterConnectivityCheckerTest.java
```diff
@@ -64,6 +70,17 @@ else if (peersC.contains(endpoint))
         return null;
     }
 
+    private static class MockIsUpgradingFromVersionLowerThan implements Predicate<CassandraVersion> {
```
smiklosovic: braces on new line in this whole class

test/unit/org/apache/cassandra/net/StartupClusterConnectivityCheckerTest.java
```diff
@@ -64,6 +70,17 @@ else if (peersC.contains(endpoint))
...
+        CassandraVersion clusterVersion;
+        MockIsUpgradingFromVersionLowerThan(CassandraVersion clusterVersion) {
+            this.clusterVersion = clusterVersion;
+        }
+        @Override
+        public boolean test(CassandraVersion other) {
+            return this.clusterVersion.compareTo(other) < 0;
```
smiklosovic: you technically do not need `this` here

test/unit/org/apache/cassandra/net/StartupClusterConnectivityCheckerTest.java
```diff
@@ -36,14 +37,19 @@
 import org.apache.cassandra.gms.Gossiper;
 import org.apache.cassandra.gms.HeartBeatState;
 import org.apache.cassandra.locator.InetAddressAndPort;
+import org.apache.cassandra.utils.CassandraVersion;
 import org.apache.cassandra.utils.FBUtilities;
 
+import static org.mockito.Mockito.mock;
```
smiklosovic: isn't this import unused? Please run `ant check`

PR #2896, opened 2023-11-14
CASSANDRA-18753: Provide cassandra_latest.yaml with suitable defaults for new users

blambov:
The first commit renames cassandra.yaml to cassandra_latest.yaml to make it easier to see the differences, the next commit restores the main yaml, and the third one configures testing.

### DIFF
--- a/.build/run-tests.sh
+++ b/.build/run-tests.sh
@@ -217,8 +217,8 @@ _main() {
     "test-system-keyspace-directory")
       _run_testlist "unit" "testclasslist-system-keyspace-directory" "${split_chunk}" "$(_timeout_for 'test.timeout')"
       ;;
-    "test-trie")
-      _run_testlist "unit" "testclasslist-trie" "${split_chunk}" "$(_timeout_for 'test.timeout')"
+    "test-latest")
+      _run_testlist "unit" "testclasslist-latest" "${split_chunk}" "$(_timeout_for 'test.timeout')"
       ;;
     "test-burn")
       _run_testlist "burn" "testclasslist" "${split_chunk}" "$(_timeout_for 'test.burn.timeout')"
--- a/.circleci/config_template.yml
+++ b/.circleci/config_template.yml
@@ -134,7 +134,7 @@ default_env_vars: &default_env_vars
     # REPEATED_ANT_TEST_TARGET: test-jvm-dtest-some
     # REPEATED_ANT_TEST_TARGET: test-cdc
     # REPEATED_ANT_TEST_TARGET: test-compression
-    # REPEATED_ANT_TEST_TARGET: test-trie
+    # REPEATED_ANT_TEST_TARGET: test-latest
     # REPEATED_ANT_TEST_TARGET: test-oa
     # REPEATED_ANT_TEST_TARGET: test-system-keyspace-directory
     REPEATED_ANT_TEST_TARGET: testsome
@@ -149,7 +149,7 @@ default_env_vars: &default_env_vars
     REPEATED_ANT_TEST_METHODS:
     # Whether the test iteration should use vnodes for JVM dtests (-Dtest.jvm.args='-Dcassandra.dtest.num_tokens=16').
     # This will only be applied as a default to JVM dtests that don't provide their own initial tokens or token count,
-    # in the same way that it's done for *_jvm_dtests_vnode jobs. Ant targets other than JVM dtests will ignore this.
+    # in the same way that it's done for *_jvm_dtests_latest_vnode jobs. Ant targets other than JVM dtests will ignore this.
     REPEATED_ANT_TEST_VNODES: false
     # The number of times that the repeated JUnit test should be run.
     REPEATED_ANT_TEST_COUNT: 500
@@ -277,11 +277,11 @@ j11_separate_jobs: &j11_separate_jobs
         requires:
           - start_j11_jvm_dtests
           - j11_build
-    - start_j11_jvm_dtests_vnode:
+    - start_j11_jvm_dtests_latest_vnode:
         type: approval
-    - j11_jvm_dtests_vnode:
+    - j11_jvm_dtests_latest_vnode:
         requires:
-          - start_j11_jvm_dtests_vnode
+          - start_j11_jvm_dtests_latest_vnode
           - j11_build
     - start_j11_jvm_dtests_repeat:
         type: approval
@@ -289,11 +289,11 @@ j11_separate_jobs: &j11_separate_jobs
         requires:
           - start_j11_jvm_dtests_repeat
           - j11_build
-    - start_j11_jvm_dtests_vnode_repeat:
+    - start_j11_jvm_dtests_latest_vnode_repeat:
         type: approval
-    - j11_jvm_dtests_vnode_repeat:
+    - j11_jvm_dtests_latest_vnode_repeat:
         requires:
-          - start_j11_jvm_dtests_vnode_repeat
+          - start_j11_jvm_dtests_latest_vnode_repeat
           - j11_build
     - start_j17_jvm_dtests:
         type: approval
@@ -301,11 +301,11 @@ j11_separate_jobs: &j11_separate_jobs
         requires:
           - start_j17_jvm_dtests
           - j11_build
-    - start_j17_jvm_dtests_vnode:
+    - start_j17_jvm_dtests_latest_vnode:
         type: approval
-    - j17_jvm_dtests_vnode:
+    - j17_jvm_dtests_latest_vnode:
         requires:
-          - start_j17_jvm_dtests_vnode
+          - start_j17_jvm_dtests_latest_vnode
           - j11_build
     - start_j17_jvm_dtests_repeat:
         type: approval
@@ -313,11 +313,11 @@ j11_separate_jobs: &j11_separate_jobs
         requires:
           - start_j17_jvm_dtests_repeat
           - j11_build
-    - start_j17_jvm_dtests_vnode_repeat:
+    - start_j17_jvm_dtests_latest_vnode_repeat:
         type: approval
-    - j17_jvm_dtests_vnode_repeat:
+    - j17_jvm_dtests_latest_vnode_repeat:
         requires:
-          - start_j17_jvm_dtests_vnode_repeat
+          - start_j17_jvm_dtests_latest_vnode_repeat
           - j11_build
     - start_j11_simulator_dtests:
         type: approval
@@ -465,29 +465,29 @@ j11_separate_jobs: &j11_separate_jobs
         requires:
           - start_j17_utests_compression_repeat
           - j11_build
-    - start_j11_utests_trie:
+    - start_j11_utests_latest:
         type: approval
-    - j11_utests_trie:
+    - j11_utests_latest:
         requires:
-          - start_j11_utests_trie
+          - start_j11_utests_latest
           - j11_build
-    - start_j17_utests_trie:
+    - start_j17_utests_latest:
         type: approval
-    - j17_utests_trie:
+    - j17_utests_latest:
         requires:
-          - start_j17_utests_trie
+          - start_j17_utests_latest
           - j11_build
-    - start_j11_utests_trie_repeat:
+    - start_j11_utests_latest_repeat:
         type: approval
-    - j11_utests_trie_repeat:
+    - j11_utests_latest_repeat:
         requires:
-          - start_j11_utests_trie_repeat
+          - start_j11_utests_latest_repeat
           - j11_build
-    - start_j17_utests_trie_repeat:
+    - start_j17_utests_latest_repeat:
         type: approval
-    - j17_utests_trie_repeat:
+    - j17_utests_latest_repeat:
         requires:
-          - start_j17_utests_trie_repeat
+          - start_j17_utests_latest_repeat
           - j11_build
     - start_j11_utests_stress:
         type: approval
@@ -594,17 +594,17 @@ j11_separate_jobs: &j11_separate_jobs
           - start_j11_dtests_vnode
           - j11_build
     # Java 11 off-heap dtests
-    - start_j11_dtests_offheap:
+    - start_j11_dtests_latest:
         type: approval
-    - j11_dtests_offheap:
+    - j11_dtests_latest:
         requires:
-          - start_j11_dtests_offheap
+          - start_j11_dtests_latest
           - j11_build
-    - start_j11_dtests_offheap_repeat:
+    - start_j11_dtests_latest_repeat:
         type: approval
-    - j11_dtests_offheap_repeat:
+    - j11_dtests_latest_repeat:
         requires:
-          - start_j11_dtests_offheap_repeat
+          - start_j11_dtests_latest_repeat
           - j11_build
     # Java 17 dtests
     - start_j17_dtests:
@@ -620,17 +620,17 @@ j11_separate_jobs: &j11_separate_jobs
           - start_j17_dtests_vnode
           - j11_build
     # Java 17 off-heap dtests
-    - start_j17_dtests_offheap:
+    - start_j17_dtests_latest:
         type: approval
-    - j17_dtests_offheap:
+    - j17_dtests_latest:
         requires:
-          - start_j17_dtests_offheap
+          - start_j17_dtests_latest
           - j11_build
-    - start_j17_dtests_offheap_repeat:
+    - start_j17_dtests_latest_repeat:
         type: approval
-    - j17_dtests_offheap_repeat:
+    - j17_dtests_latest_repeat:
         requires:
-          - start_j17_dtests_offheap_repeat
+          - start_j17_dtests_latest_repeat
           - j11_build
     # Python large DTests
     - start_j11_dtests_large:
@@ -700,16 +700,16 @@ j11_separate_jobs: &j11_separate_jobs
         requires:
         - start_j11_cqlsh_tests
         - j11_build
-    # Java 11 cqlsh offheap dtests offheap
-    - start_j11_cqlsh_tests_offheap:
+    # Java 11 cqlsh latest dtests
+    - start_j11_cqlsh_tests_latest:
         type: approval
-    - j11_cqlsh_dtests_py38_offheap:
+    - j11_cqlsh_dtests_py38_latest:
         requires:
-          - start_j11_cqlsh_tests_offheap
+          - start_j11_cqlsh_tests_latest
           - j11_build
-    - j11_cqlsh_dtests_py311_offheap:
+    - j11_cqlsh_dtests_py311_latest:
         requires:
-          - start_j11_cqlsh_tests_offheap
+          - start_j11_cqlsh_tests_latest
           - j11_build
     # Java 17 cqlsh dtests
     - start_j17_cqlsh_tests:
@@ -731,15 +731,15 @@ j11_separate_jobs: &j11_separate_jobs
           - start_j17_cqlsh_tests
           - j11_build
     # Java 17 cqlsh dtests off-heap
-    - start_j17_cqlsh_tests_offheap:
+    - start_j17_cqlsh_tests_latest:
         type: approval
-    - j17_cqlsh_dtests_py38_offheap:
+    - j17_cqlsh_dtests_py38_latest:
         requires:
-          - start_j17_cqlsh_tests_offheap
+          - start_j17_cqlsh_tests_latest
           - j11_build
-    - j17_cqlsh_dtests_py311_offheap:
+    - j17_cqlsh_dtests_py311_latest:
         requires:
-          - start_j17_cqlsh_tests_offheap
+          - start_j17_cqlsh_tests_latest
           - j11_build
     # Java 11 upgrade tests
     - start_j11_upgrade_dtests:
@@ -819,6 +819,12 @@ j11_pre-commit_jobs: &j11_pre-commit_jobs
     - j11_unit_tests_repeat:
         requires:
           - j11_build
+    - j11_utests_latest:
+        requires:
+          - j11_build
+    - j11_utests_latest_repeat:
+        requires:
+          - j11_build
     - j11_simulator_dtests:
         requires:
           - j11_build
@@ -831,10 +837,10 @@ j11_pre-commit_jobs: &j11_pre-commit_jobs
     - j11_jvm_dtests_repeat:
         requires:
           - j11_build
-    - j11_jvm_dtests_vnode:
+    - j11_jvm_dtests_latest_vnode:
         requires:
           - j11_build
-    - j11_jvm_dtests_vnode_repeat:
+    - j11_jvm_dtests_latest_vnode_repeat:
         requires:
           - j11_build
     - j17_jvm_dtests:
@@ -843,10 +849,10 @@ j11_pre-commit_jobs: &j11_pre-commit_jobs
     - j17_jvm_dtests_repeat:
         requires:
           - j11_build
-    - j17_jvm_dtests_vnode:
+    - j17_jvm_dtests_latest_vnode:
         requires:
           - j11_build
-    - j17_jvm_dtests_vnode_repeat:
+    - j17_jvm_dtests_latest_vnode_repeat:
         requires:
           - j11_build
     - j11_cqlshlib_tests:
@@ -869,6 +875,12 @@ j11_pre-commit_jobs: &j11_pre-commit_jobs
         requires:
           - j11_build
     - j17_unit_tests_repeat:
+        requires:
+          - j11_build
+    - j17_utests_latest:
+        requires:
+          - j11_build
+    - j17_utests_latest_repeat:
         requires:
           - j11_build
     # specialized unit tests (all run on request)
@@ -925,24 +937,6 @@ j11_pre-commit_jobs: &j11_pre-commit_jobs
     - j17_utests_compression_repeat:
         requires:
           - start_utests_compression
-          - j11_build
-    - start_utests_trie:
-        type: approval
-    - j11_utests_trie:
-        requires:
-          - start_utests_trie
-          - j11_build
-    - j17_utests_trie:
-        requires:
-          - start_utests_trie
-          - j11_build
-    - j11_utests_trie_repeat:
-        requires:
-          - start_utests_trie
-          - j11_build
-    - j17_utests_trie_repeat:
-        requires:
-          - start_utests_trie
           - j11_build
     - start_utests_stress:
         type: approval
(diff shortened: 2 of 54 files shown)


conf/cassandra_latest.yaml
```diff
@@ -0,0 +1,2139 @@
...
+# Leave undefined to use a default suitable for the SSTable format
+# in use (64 KiB for BIG, 16KiB for BTI).
+# Min unit: KiB
+column_index_size: 4KiB
+
+# Default compaction strategy, applied when a table's parameters do not
+# specify compaction.
+# The selected compaction strategy will also apply to system tables.
+#
+# The default is to use SizeTieredCompactionStrategy, with its default
```
clohfink: NP: comment says default is SizeTieredCompactionStrategy but setting it to UCS, should update comment too

conf/cassandra_latest.yaml
```diff
@@ -0,0 +1,2139 @@
...
+# new space for cdc-tracked tables has been made available. Default to 250ms
+# Min unit: ms
+# cdc_free_space_check_interval: 250ms
+
+# Whether to, when doing sequential writing, fsync() at intervals in
+# order to force the operating system to flush the dirty
+# buffers. Enable this to avoid sudden dirty buffer flushing from
+# impacting read latencies. Almost always a good idea on SSDs; not
+# necessarily on platters.
+trickle_fsync: false
```
michaelsembwever: time to assume new users are on SSDs ?

bereng:
Given `test-latest` will test tries+oa we could remove all oa test specific jobs to spare CI cycles as they're redundant now?
Also shouldn't this ticket also enable all dtests, both python and jvm, with `latest`?

blambov:
> Given test-latest will test tries+oa we could remove all oa test specific jobs to spare CI cycles as they're redundant now?

Actually, it will not test the the `big-oa` format but the `bti-da` one.

As for dtests, I'm still working on making the relevant changes.

conf/cassandra_latest.yaml
blambov: Done

conf/cassandra_latest.yaml
blambov: Well, the default is indeed STCS. Added "If nothing is specified, " to make a little clearer.

bereng:
Your 'latest' [does already run](https://github.com/blambov/cassandra/blob/CASSANDRA-18753/build.xml#L1281) with SCM `NONE` which is the same as running with `oa` feature-wise. Do we see value in testing `oa` by itself?

What I am trying to explain is that `latest` will already exercise the same code paths as `oa`...

blambov:
`big-oa` is a separate sstable format, and it needs to be tested. Doing only base junit tests with it is likely sufficient; we can skip dtests (python or in-jvm) on it.

bereng:
As `latest` will dtest `oa` current features that sounds like a good compromise to me, agreed.

.jenkins/Jenkinsfile
```diff
@@ -292,7 +292,7 @@ pipeline {
                     sleep(60 * attempt)
                   }
                   attempt = attempt + 1
-                  trie = build job: "${env.JOB_NAME}-test-trie", propagate: false
+                  trie = build job: "${env.JOB_NAME}-test-latest", propagate: false
                   if (trie.result != 'FAILURE') break
               }
               if (trie.result != 'SUCCESS') unstable('trie failures')
```
michaelsembwever:
```suggestion
                  latest = build job: "${env.JOB_NAME}-test-latest", propagate: false
                  if (latest.result != 'FAILURE') break
              }
              if (latest.result != 'SUCCESS') unstable('test-latest failures')
```

.jenkins/Jenkinsfile
```diff
@@ -303,7 +303,7 @@ pipeline {
             always {
                 warnError('missing test xml files') {
                     script {
-                        copyTestResults('test-trie', trie.getNumber())
+                        copyTestResults('test-latest', trie.getNumber())
```
michaelsembwever:
```suggestion
                        copyTestResults('test-latest', latest.getNumber())
```

.jenkins/Jenkinsfile
michaelsembwever: line 286 also needs changing

michaelsembwever:
are we (and how) dtesting this config too ? (java and python dtests), separate ticket ?

bereng:
^ https://github.com/apache/cassandra/pull/2896#issuecomment-1812084239

> As for dtests, I'm still working on making the relevant changes.

test/distributed/org/apache/cassandra/distributed/impl/InstanceConfig.java
```diff
@@ -113,6 +114,45 @@ private InstanceConfig(int num,
                 .set("index_summary_capacity", "50MiB")
                 .set("counter_cache_size", "50MiB")
                 .set("key_cache_size", "50MiB");
+        if (CassandraRelevantProperties.DTEST_JVM_DTESTS_USE_LATEST.getBoolean())
+        {
+            // TODO: make this load latest_diff.yaml or cassandra_latest.yaml
```
bereng, line 120: I would add a note in build.xml or some testing README about this. It's going to be very confusing changing yamls and seeing no effect as verything is hardcoded here.

test/unit/org/apache/cassandra/cql3/validation/operations/SelectTest.java
```diff
@@ -341,8 +344,11 @@ public void testSetContainsWithIndex() throws Throwable
             assertInvalidMessage("Unsupported unset value for column categories",
                                  "SELECT * FROM %s WHERE account = ? AND id = ? AND categories CONTAINS ?", "test", 5, unset());
 
-            assertInvalidMessage(StatementRestrictions.REQUIRES_ALLOW_FILTERING_MESSAGE,
-                                 "SELECT * FROM %s WHERE account = ? AND categories CONTAINS ? AND categories CONTAINS ?", "xyz", "lmn", "notPresent");
+            if (DatabaseDescriptor.getDefaultSecondaryIndex().equals(CassandraIndex.NAME))
```
bereng, line 347: I wonder if it would read better to move this that is repeated so much into a method in CQLTester with some javadoc explanation

test/unit/org/apache/cassandra/db/SchemaCQLHelperTest.java
```diff
@@ -444,13 +446,17 @@ public void testSnapshot() throws Throwable
                          containsString("ALTER TABLE " + keyspace() + "." + tableName + " DROP reg3 USING TIMESTAMP 10000;"),
                          containsString("ALTER TABLE " + keyspace() + "." + tableName + " ADD reg3 int;")));
 
-        assertThat(schema, containsString("CREATE INDEX IF NOT EXISTS " + tableName + "_reg2_idx ON " + keyspace() + '.' + tableName + " (reg2);"));
+        final boolean isIndexLegacy = DatabaseDescriptor.getDefaultSecondaryIndex().equals(CassandraIndex.NAME);
```
bereng, line 449: I wonder if it would read better to move this that is repeated so much into a method in CQLTester with some javadoc explanation

test/unit/org/apache/cassandra/db/compaction/unified/ControllerTest.java
```diff
@@ -270,7 +270,7 @@ public void testGetNumShards_growth_0()
         assertEquals(3 * (int) Controller.MAX_SHARD_SPLIT, controller.getNumShards(Math.scalb(10, 60)));
         assertEquals(3 * (int) Controller.MAX_SHARD_SPLIT, controller.getNumShards(Double.POSITIVE_INFINITY));
         // Check NaN
-        assertEquals(3, controller.getNumShards(Double.NaN));
+        assertEquals(1, controller.getNumShards(Double.NaN));
```
bereng, line 274: Why this change?

src/java/org/apache/cassandra/db/ReadCommand.java
```diff
@@ -658,6 +658,9 @@ public UnfilteredRowIterator applyToPartition(UnfilteredRowIterator iter)
             @Override
             protected Row applyToStatic(Row row)
             {
+                if (row == Rows.EMPTY_STATIC_ROW)
```
blambov, line 661: This was incorrectly counted and brought the size over the limit for `LocalReadSizeWarningTest`.

src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.java
```diff
@@ -95,6 +95,7 @@ public UnifiedCompactionStrategy(ColumnFamilyStore cfs, Map<String, String> opti
         super(cfs, options);
         this.controller = controller;
         estimatedRemainingTasks = 0;
+        lastExpiredCheck = Clock.Global.currentTimeMillis();
```
blambov, line 98: Without this an expiration check was carried out almost immediately, which broke some tests.

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -268,9 +268,10 @@ public int getNumShards(double localDensity)
         if (minSSTableSize > 0)
         {
             double count = localDensity / minSSTableSize;
-            // Minimum size only applies if it is smaller than the base count.
+            // Minimum size only applies if the base count would result in smaller sstables.
+            // We also want to use the min size if we don't yet know the flush size (density is NaN).
             // Note: the minimum size cannot be larger than the target size's minimum.
-            if (count < baseShardCount)
+            if (!(count >= baseShardCount)) // also true for count == NaN
```
blambov, line 274: Tests typically run with unknown flush size and we want them to use just one shard.

It also generally makes better sense to start non-splitting if we don't know what size we should expect.

test/distributed/org/apache/cassandra/distributed/impl/InstanceConfig.java
blambov: Added one in `latest_diff.yaml`.

test/unit/org/apache/cassandra/db/SchemaCQLHelperTest.java
blambov: Created `Util.assumeLegacySecondaryIndex()` to replace most of the repetition.

It doesn't change this particular line, though. Let me know if you insist on that being placed in `Util` as well.

test/unit/org/apache/cassandra/db/compaction/unified/ControllerTest.java
blambov: Because of the switch to go with no sharding for unknown flush size (in that case the flush size is returned as `Double.NaN`), see the change in `Controller.java`.

build.xml
```diff
@@ -1290,22 +1290,22 @@
...
+    <macrodef name="testlist-latest">
         <attribute name="test.file.list" />
         <sequential>
-            <property name="trie_yaml" value="${build.test.dir}/cassandra.trie.yaml"/>
-            <concat destfile="${trie_yaml}">
+            <property name="latest_yaml" value="${build.test.dir}/cassandra.latest.yaml"/>
+            <concat destfile="${latest_yaml}">
                 <fileset file="${test.conf}/cassandra.yaml"/>
-                <fileset file="${test.conf}/trie_memtable.yaml"/>
+                <fileset file="${test.conf}/latest_diff.yaml"/>
```
jacek-lewandowski, line 1299: latest_yaml?

.build/run-tests.sh
```diff
@@ -217,8 +217,8 @@ _main() {
     "test-system-keyspace-directory")
       _run_testlist "unit" "testclasslist-system-keyspace-directory" "${split_chunk}" "$(_timeout_for 'test.timeout')"
       ;;
-    "test-trie")
-      _run_testlist "unit" "testclasslist-trie" "${split_chunk}" "$(_timeout_for 'test.timeout')"
+    "test-latest")
```
bereng, line 220: Do you need a cassandra-builds PR to exercise this path?

.circleci/config_template.yml
```diff
@@ -37,8 +37,8 @@ default_env_vars: &default_env_vars
     CASS_DRIVER_NO_CYTHON: true
     #Skip all syncing to disk to avoid performance issues in flaky CI environments
     CASSANDRA_SKIP_SYNC: true
-    DTEST_REPO: https://github.com/apache/cassandra-dtest.git
-    DTEST_BRANCH: trunk
+    DTEST_REPO: https://github.com/blambov/cassandra-dtest.git
```
bereng: Note to self: These 2 lines are isolated in a circle commit

.build/run-tests.sh
bereng: Note to self: fixed in in-tree jenkins files https://github.com/apache/cassandra/pull/2896/files#diff-917491029f94ac1ce6c114a542d2a8738ce1f2d1e866eaffe7189d2139250200R286

test/distributed/org/apache/cassandra/distributed/impl/InstanceConfig.java
bereng: Do we want to make that param a String instead of a Boolean to open the door to more configs?

test/distributed/org/apache/cassandra/distributed/impl/InstanceConfig.java
```diff
@@ -114,6 +115,47 @@ private InstanceConfig(int num,
                 .set("counter_cache_size", "50MiB")
                 .set("key_cache_size", "50MiB")
                 .set("commitlog_disk_access_mode", "legacy");
+        if (CassandraRelevantProperties.DTEST_JVM_DTESTS_USE_LATEST.getBoolean())
+        {
+            // TODO: make this load latest_diff.yaml or cassandra_latest.yaml
```
bereng, line 120: This TODO seems quite critical. Wouldn't we prefer a crude yaml read for the param values rather than hard-coded?

test/distributed/org/apache/cassandra/distributed/test/CasWriteTest.java
```diff
@@ -253,32 +263,42 @@ public void testWriteUnknownResult()
...
+            Assert.assertEquals("Expecting cause to be " + exceptionClass.getSimpleName(),
+                                exceptionClass.getCanonicalName(), t.getClass().getCanonicalName());
             return;
         }
         Assert.fail("Expecting test to throw a CasWriteUnknownResultException");
     }
 
+    private static boolean isPaxosVariant2()
+    {
+        return "v2".equals(cluster.coordinator(1).instance().config().getString("paxos_variant"));
```
bereng: Could have used `PaxosVariant.v2` enum instead of a harcoded value?

test/distributed/org/apache/cassandra/distributed/test/UpgradeSSTablesTest.java
```diff
@@ -59,8 +59,8 @@ public void upgradeSSTablesInterruptsOngoingCompaction() throws Throwable
             cluster.get(1).acceptsOnInstance((String ks) -> {
                 ColumnFamilyStore cfs = Keyspace.open(ks).getColumnFamilyStore("tbl");
                 cfs.disableAutoCompaction();
-                CompactionManager.instance.setMaximumCompactorThreads(1);
                 CompactionManager.instance.setCoreCompactorThreads(1);
+                CompactionManager.instance.setMaximumCompactorThreads(1);
```
bereng, line 64: Why this reordering?

test/unit/org/apache/cassandra/cql3/statements/DescribeStatementTest.java
```diff
@@ -462,11 +463,18 @@ public void testDescribeCluster() throws Throwable
 
         assertRowsNet(executeDescribeNet(KEYSPACE_PER_TEST, "DESCRIBE CLUSTER"),
                       row("Test Cluster",
-                          "ByteOrderedPartitioner",
+                          trimIfPresent(DatabaseDescriptor.getPartitionerName(), "org.apache.cassandra.dht."),
```
bereng, line 466: `String.replaceFirst()`?

blambov: Works as an implementation of `trimIfPresent` (`src.replaceFirst("^" + Pattern.quote(begin), "")`), but the current code is IMHO a little more obvious.

test/distributed/org/apache/cassandra/distributed/test/UpgradeSSTablesTest.java
blambov: Because the previous order was invalid (if the core is e.g. 4 and you set the max to 1, it's an invalid operation). Reducing the core first is the correct way to do it.

test/distributed/org/apache/cassandra/distributed/impl/InstanceConfig.java
blambov: Not if the non-latest path does not read any configuration file either.

Both need to be fixed to read a config file (possibly applying changes on top of it), but my quick attempt to do it failed and I don't want to delay the introduction and testing of the "latest" config (and 5.0 with it).

test/unit/org/apache/cassandra/cql3/validation/operations/InsertUpdateIfConditionTest.java
```diff
@@ -459,7 +459,7 @@ public void testDropCreateTableIfNotExists() throws Throwable
     @Test
     public void testDropCreateIndexIfNotExists()
     {
-        String tableName = createTable("CREATE TABLE %s (id text PRIMARY KEY, value1 blob, value2 blob)with comment = 'foo'");
+        String tableName = createTable("CREATE TABLE %s (id text PRIMARY KEY, value1 text, value2 blob)with comment = 'foo'");
```
bereng, line 462: Why this change?

test/unit/org/apache/cassandra/cql3/validation/operations/SelectLimitTest.java
```diff
@@ -39,6 +40,7 @@ public static void setUpClass()
 
         StorageService.instance.setPartitionerUnsafe(ByteOrderedPartitioner.instance);
         DatabaseDescriptor.setPartitionerUnsafe(ByteOrderedPartitioner.instance);
+        DatabaseDescriptor.setDefaultSecondaryIndex(CassandraIndex.NAME);
```
bereng: Does this line force the test to run always with that index flavor? If it were so an `assume` would spare some CI cycles.

test/unit/org/apache/cassandra/cql3/validation/operations/InsertUpdateIfConditionTest.java
blambov: SAI does not support indexes on blobs. This tests does not need a blob, text (which SAI supports) suffices.

test/distributed/org/apache/cassandra/distributed/test/CasWriteTest.java
blambov: Done

test/unit/org/apache/cassandra/cql3/validation/operations/SelectLimitTest.java
blambov: BOP is incompatible with SAI.

Only one test uses indexing here, though, so changed only that test to `assumeLegacySecondaryIndex()`.

test/distributed/org/apache/cassandra/distributed/test/SecondaryIndexCompactionTest.java
```diff
@@ -43,6 +44,8 @@ public void test2iCompaction() throws IOException
     {
         try (Cluster cluster = init(Cluster.build(1).start()))
         {
+            Assume.assumeFalse("Test only valid for legacy index",
```
bereng, line 47: Should this be changed by the new assume method added in Util?

test/unit/org/apache/cassandra/config/ParseAndConvertUnitsTest.java
```diff
@@ -104,7 +104,8 @@ public void testConfigurationLoaderParser()
         assertNull(config.file_cache_size);
         assertNull(config.index_summary_capacity);
         assertEquals(new DataStorageSpec.LongMebibytesBound(1), config.prepared_statements_cache_size);
-        assertNull(config.key_cache_size);
+        if (config.key_cache_size != null)  // null in default test config, 0 in latest
```
bereng: `latest what?` is what came to mind when I read that. Add ticket number to the reader can see `latest` refers to a special config set-up. wdyt?

test/unit/org/apache/cassandra/cql3/validation/operations/SelectSingleColumnRelationTest.java
```diff
@@ -456,6 +459,8 @@ public void testLargeClusteringINValues() throws Throwable
     @Test
     public void testMultiplePartitionKeyWithIndex() throws Throwable
     {
+        Assume.assumeFalse("SAI does not allow multi-column slice restrictions",
```
bereng: Could be replaced with the new assume in Util + a comment?

test/unit/org/apache/cassandra/index/CustomIndexTest.java
```diff
@@ -162,6 +165,8 @@ public void indexReceivesWriteTimeDeletionsCorrectly() throws Throwable
     @Test
     public void nonCustomIndexesRequireExactlyOneTargetColumn() throws Throwable
     {
+        Assume.assumeTrue("Test does not work with different default secondary index",
```
bereng: This could be replaced with the new method in Util?

src/java/org/apache/cassandra/db/ReadCommand.java
bereng: +1

test/distributed/org/apache/cassandra/distributed/impl/InstanceConfig.java
bereng: Ok let's create a ticket for this so it doesn't fall through the cracks.

test/unit/org/apache/cassandra/db/SchemaCQLHelperTest.java
bereng: sgtm

test/unit/org/apache/cassandra/index/sai/cql/MixedIndexImplementationsTest.java
```diff
@@ -81,6 +83,8 @@ public void shouldAcceptColumnsWithOtherIndex() throws Throwable
     @Test
     public void shouldRequireAllowFilteringWithOtherIndex() throws Throwable
     {
+        Assume.assumeFalse("Mixed test only makes sense if SAI is not default",
```
bereng: Can be replaced by new method in Util?

bereng:
@blambov I have gone through it and you see we have only minors. I would suggest a rebase + CI runs where I will check the new circle setup. Doing that trying to follow the config.yaml changes, which I did but only cursory, would be eye watering...

test/distributed/org/apache/cassandra/distributed/test/SecondaryIndexCompactionTest.java
blambov: No. The distributed test runner is not the database instance whose configuration we want to check.

test/unit/org/apache/cassandra/config/ParseAndConvertUnitsTest.java
blambov: Done

test/unit/org/apache/cassandra/cql3/validation/operations/SelectSingleColumnRelationTest.java
blambov: Done

test/unit/org/apache/cassandra/index/CustomIndexTest.java
blambov: Done

test/unit/org/apache/cassandra/index/sai/cql/MixedIndexImplementationsTest.java
blambov: Done

bereng:
If [this](https://app.circleci.com/pipelines/github/blambov/cassandra/597/workflows/9f431155-06c0-4333-aa2b-7c45c6893c77) is it's CI it doesn't look pretty :-(

blambov:
Well, the failures appear to be [as expected](https://issues.apache.org/jira/browse/CASSANDRA-18753?focusedCommentId=17799053&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17799053).

We do have a few outstanding problems that we never tested for before.

bereng:
CI workflows LGTM +1 on the overall approach and the PR. Disclaimer I have no deep knowledge of all the options that have been enabled but overall SGTM. Commenting in case you'd like sbdy with deep expertise in some area to doublecheck things.

I suggest sending an email to the ML to find out what you mentioned to me offline: Are we ok merging this and surfacing a number of failures?

conf/cassandra_latest.yaml
```diff
@@ -0,0 +1,2175 @@
...
+
+# When the client triggers a protocol exception or unknown issue (Cassandra bug) we increment
+# a client metric showing this; this logic will exclude specific subnets from updating these
+# metrics
+#client_error_reporting_exclusions:
+#  subnets:
+#    - 127.0.0.1
+#    - 127.0.0.0/31
+
+# Enables read thresholds (warn/fail) across all replicas for reporting back to the client.
```
dcapwell, line 1904: is there a reason guardrails and repair retries are disabled with this config?  I understood the goals for this config to show people what we recommend and the safety/reliably configs are off by default still?

blambov: This is a rebase issue that will be fixed on commit. More precisely, changes since the last rebase [are reflected in `cassandra_latest.yaml`](https://github.com/apache/cassandra/pull/2896/files#diff-85a7a4cf7134a153765a2bda5b64773b9cc7fbf9ad600538742f361cc1760a75R2102) (because of the first renaming commit) which makes it much easier to see and not miss applying them to both configuration files.

blambov: The only reason things that are should be switched on are not is that I do not know about them or how to set them. Could you advise which guardrails should be set, and to what values?

pmcfadin:
Closing as [CASSANDRA-18753](https://issues.apache.org/jira/browse/CASSANDRA-18753) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #2899, opened 2023-11-14
Fix Paxos V2 prepare response serialization

aweisberg:
Patch by Ariel Weisberg; Reviewed by David Capwell for CASANDRA-19023

### DIFF
--- a/src/java/org/apache/cassandra/service/paxos/PaxosPrepare.java
+++ b/src/java/org/apache/cassandra/service/paxos/PaxosPrepare.java
@@ -1218,23 +1218,15 @@ public class PaxosPrepare extends PaxosRequestCallback<PaxosPrepare.Response> im
         }
     }
 
-    private static void serializeRejection(DataOutputPlus out, Ballot supersededBy, ConsensusMigratedAt maybeConsenusMigratedAt, int version) throws IOException
-    {
-        out.writeByte(0);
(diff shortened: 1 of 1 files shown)


dcapwell (changes requested):
this doesn't match trunk, so won't this break paxos v2?

src/java/org/apache/cassandra/service/paxos/PaxosPrepare.java
```diff
@@ -1232,7 +1232,6 @@ public void serialize(Response response, DataOutputPlus out, int version) throws
         {
             if (response.isRejected())
             {
-                out.writeByte(0);
```
bdeggleston, line 1235: Could we just move the contents of `serializeRejection` back here? I don't think it needs to be in a separate method, especially when we're serializing the success response inline. @dcapwell this bug only exists in the cep-15 branch

bdeggleston (changes requested):

bdeggleston (approved):
+1

dcapwell (approved):

aweisberg:
Found a bug in size calculation as well. Removed `serializedRejection` and put the serialization of `maybeConsensusMigratedAt` at the end so it doesn't need to be repeated for accept/reject.

bdeggleston (approved):
+1

PR #2916, opened 2023-11-17
CASSANDRA-18166 Improve code model around IndexContext

mike-tr-adamson:
Improve code model around IndexContext

- Replace IndexContext with IndexTermType and IndexDefinition
- Move index specific managers, factories and metrics to StorageAttachedIndex
- Refactor Expression to explicitly define indexed and unindexed expressions


src/java/org/apache/cassandra/index/sai/utils/IndexTermType.java
```diff
@@ -0,0 +1,858 @@
...
+        return hasProperty(IS_NON_FROZEN_COLLECTION);
+    }
+
+    /**
+     * Returns {@code true} if the index type is a frozen collection. This is the inverse of a non-frozen collection
+     * but this method is here for clarity.
+     */
+    public boolean isFrozenCollection()
+    {
+        return !hasProperty(IS_NON_FROZEN_COLLECTION);
```
maedhroz: If the type isn't a non-frozen collection, that doesn't mean it _is_ a frozen collection, or a collection at all, right? I think this is part of the problem in `IndexTermTypeTest`. Tuples and UDTs are totally not non-frozen collections, but it's because they're not collections.

src/java/org/apache/cassandra/index/sai/utils/IndexTermType.java
```diff
@@ -0,0 +1,858 @@
...
+    private static final int IS_VECTOR                = 1 << 2;
+    private static final int IS_REVERSED              = 1 << 3;
+    private static final int IS_FROZEN                = 1 << 4;
+    private static final int IS_NON_FROZEN_COLLECTION = 1 << 5;
+    private static final int IS_COMPOSITE             = 1 << 6;
+    private static final int IS_INET_ADDRESS          = 1 << 7;
+    private static final int IS_BIG_INTEGER           = 1 << 8;
+    private static final int IS_BIG_DECIMAL           = 1 << 9;
+    private static final int IS_LONG                  = 1 << 10;
+    private static final int IS_COMPOSITE_PARTITION   = 1 << 11;
```
maedhroz: Any reason not to use `EnumSet` here? Something like...

```
enum Flag { IS_LONG, IS_COMPOSITE_PARTITION, ... }
private final EnumSet<Flag> capabilities = EnumSet.of(Flag.IS_LONG, Flag.IS_COMPOSITE_PARTITION);
```

maedhroz: ...and usage...

```
private boolean isCompositePartition()
    {
        return capabilities.contains(Flag.IS_COMPOSITE_PARTITION);
    }
```

src/java/org/apache/cassandra/index/sai/utils/IndexTermType.java
```diff
@@ -0,0 +1,858 @@
...
+    }
+
+    /**
+     * Encode a {@link InetAddress} into a fixed width 16 byte encoded value.
+     * <p>
+     * The encoded value is byte comparable and prefix compressible.
+     * <p>
+     * The encoding is done by converting ipv4 addresses to their ipv6 equivalent.
+     */
+    private ByteBuffer encodeInetAddress(ByteBuffer value)
```
maedhroz: nit: Looks like this, `encodeBigInteger()`, and `encodeDecimal()` can still be `static`?

src/java/org/apache/cassandra/index/sai/utils/IndexIdentifier.java
```diff
@@ -0,0 +1,62 @@
...
+    public String logMessage(String message)
+    {
+        // Index names are unique only within a keyspace.
+        return String.format("[%s.%s.%s] %s", keyspaceName, tableName, indexName, message);
+    }
+
+    @Override
+    public String toString()
+    {
+        return indexName;
```
maedhroz: Would...

```
return String.format("%s.%s", keyspaceName, indexName);
```

...be better?

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
```diff
@@ -709,15 +524,164 @@ public IndexBuildingSupport getBuildTaskSupport()
...
+    {
+        return viewManager.view();
+    }
+
+    public IndexTermType indexTermType()
+    {
+        return this.indexTermType;
+    }
+
+    public IndexIdentifier indexIdentifier()
```
maedhroz:
```suggestion
    public IndexIdentifier identifier()
```
nit: This might look nicer when we have usages like `index.identifier()` rather than `index.indexIdentifier()`...

test/unit/org/apache/cassandra/index/sai/utils/AbstractPrimaryKeyTester.java
```diff
@@ -146,7 +146,7 @@ else if (expected < 0)
     protected DecoratedKey makeKey(TableMetadata table, Object...partitionKeys)
     {
         ByteBuffer key;
-        if (TypeUtil.isComposite(table.partitionKeyType))
+        if (table.partitionKeyType instanceof CompositeType)
```
maedhroz, line 149: The difference is we don't need to worry about getting a base type here now? (Can't reverse a partition key?)

src/java/org/apache/cassandra/index/sai/utils/IndexTermType.java
mike-tr-adamson: Correct, I've fixed on [62970c5](https://github.com/apache/cassandra/pull/2916/commits/62970c5d28560d20cef503bf05299a5a9e6a6c1d). I was trying to optimise the method and completely forgot about tuples and UDTs.

src/java/org/apache/cassandra/index/sai/utils/IndexTermType.java
mike-tr-adamson: I had thought about doing this but (for some reason) stuck with the bitset. I think an `EnumSet` would be far better for debugging, I have been finding the bitset a pain. I'll get that in now.

src/java/org/apache/cassandra/index/sai/utils/IndexIdentifier.java
mike-tr-adamson: I have gone with the option of passing `IndexIdentifier` to string formatting where the index name is needed. I realise that this might be a bad idea so I'll go with this suggestion and change it's usage to be explicitly `indexIdentifier.indexName`.

test/unit/org/apache/cassandra/index/sai/utils/AbstractPrimaryKeyTester.java
mike-tr-adamson: That is my understanding. `ReversedType` is only used for clustering keys that are ordered `DESC`. I have looked for other usages and can't find any.

src/java/org/apache/cassandra/index/sai/disk/SSTableIndex.java
```diff
@@ -61,17 +63,19 @@ public abstract class SSTableIndex implements SegmentOrdering
                                                                         .thenComparing(s -> s.getSSTable().descriptor.id, SSTableIdFactory.COMPARATOR);
 
     protected final SSTableContext sstableContext;
-    protected final IndexContext indexContext;
+    protected final StorageAttachedIndex index;
```
adelapena: I think we don't need the `StorageAttachedIndex` attribute when we are storing references to its `IndexTermType` and `IndexIndentifier`. We can simply remove it and update its only references in `V1SSTableIndex#populateSegmentView` to use `indexIdentifier` and `indexTermType`.

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
```diff
@@ -709,15 +524,164 @@ public IndexBuildingSupport getBuildTaskSupport()
...
+    }
+
+    public View view()
+    {
+        return viewManager.view();
+    }
+
+    public IndexTermType termType()
+    {
+        return this.indexTermType;
```
adelapena:
```suggestion
        return indexTermType;
```

src/java/org/apache/cassandra/index/sai/utils/IndexTermType.java
```diff
@@ -0,0 +1,868 @@
...
+        return Objects.equals(columnMetadata, other.columnMetadata) && (indexTargetType == other.indexTargetType);
+    }
+
+    @Override
+    public int hashCode()
+    {
+        return Objects.hash(columnMetadata, indexTargetType);
+    }
+
+    private EnumSet<Capability> calculateIdentityBitmap(ColumnMetadata columnMetadata, List<ColumnMetadata> partitionKeyColumns, IndexTarget.Type indexTargetType)
```
adelapena: After having changed this to an `EnumSet`, the method could be simply named `calculateCapabilities`.

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
```diff
@@ -267,10 +266,7 @@ public FileHandle createPerSSTableFileHandle(IndexComponent indexComponent, Thro
             final File file = fileFor(indexComponent);
 
             if (logger.isTraceEnabled())
-            {
-                logger.trace(logMessage("Opening {} file handle for {} ({})"),
-                             file, FBUtilities.prettyPrintMemory(file.length()));
-            }
+                logger.trace(logMessage("Opening {} file handle for {} ({})"), file, FBUtilities.prettyPrintMemory(file.length()));
```
adelapena: There are three `{}` placeholders for only two arguments.

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
```diff
@@ -347,36 +345,36 @@ public long sizeOnDiskOfPerSSTableComponents()
...
 
     @SuppressWarnings("BooleanMethodIsAlwaysInverted")
-    public boolean validatePerIndexComponents(IndexContext indexContext, IndexValidation validation)
+    public boolean validatePerIndexComponents(IndexTermType indexTermType, IndexIdentifier indexIdentifier, IndexValidation validation)
     {
         if (validation == IndexValidation.NONE)
             return true;
 
-        logger.info(indexContext.logMessage("Validating per-column index components using mode " + validation));
+        logger.info(logMessage("Validating per-column index components for " + indexIdentifier + " using mode " + validation));
```
adelapena:
```suggestion
        logger.info(logMessage("Validating per-column index components for {} using mode {}"), indexIdentifier, validation);
```

src/java/org/apache/cassandra/index/sai/disk/format/OnDiskFormat.java
```diff
@@ -42,11 +43,11 @@
  * <ul>
  *     <li>Methods taking no parameters. These methods return static information about the
  *     format. This can include static information about the per-sstable components</li>
- *     <li>Methods taking just an {@link IndexContext}. These methods return static information
+ *     <li>Methods taking just a {@link StorageAttachedIndex}. These methods return static information
```
adelapena: I think there aren't such methods.

src/java/org/apache/cassandra/index/sai/disk/format/Version.java
```diff
@@ -127,24 +127,24 @@ public FileNameFormatter fileNameFormatter()
...
     /**
      * SAI default filename formatter. This is the current SAI on-disk filename format
-     *
+     * <p>
      * Format: {@code <sstable descriptor>-SAI+<version>(+<index name>)+<component name>.db}
      * Note: The index name is excluded for per-SSTable index files that are shared
      * across all the per-column indexes for the SSTable.
      */
-    private static String defaultFileNameFormat(IndexComponent indexComponent, IndexContext indexContext, String version)
+    private static String defaultFileNameFormat(IndexComponent indexComponent, IndexIdentifier indexIdentifier, String version)
```
adelapena: Can we mark the `indexIdentifier` parameter as `@Nullable`?

src/java/org/apache/cassandra/index/sai/disk/v1/SSTableIndexWriter.java
```diff
@@ -166,20 +175,20 @@ public void abort(Throwable cause)
...
             long allocated = currentBuilder.totalBytesAllocated();
-            long globalBytesUsed = currentBuilder.release(indexContext);
-            logger.debug(indexContext.logMessage("Aborting index writer for SSTable {} released {}. Global segment memory usage now at {}."),
+            long globalBytesUsed = currentBuilder.release(indexIdentifier);
+            logger.debug(indexIdentifier.logMessage("Aborting index writer for SSTable {} released {}. Global segment memory usage now at {}."),
                          indexDescriptor.sstableDescriptor, FBUtilities.prettyPrintMemory(allocated), FBUtilities.prettyPrintMemory(globalBytesUsed));
         }
 
-        indexDescriptor.deleteColumnIndex(indexContext);
+        indexDescriptor.deleteColumnIndex(index.termType(), indexIdentifier);
```
adelapena:
```suggestion
        indexDescriptor.deleteColumnIndex(indexTermType, indexIdentifier);
```

src/java/org/apache/cassandra/index/sai/disk/v1/SSTableIndexWriter.java
```diff
@@ -62,11 +62,13 @@ public class SSTableIndexWriter implements PerColumnIndexWriter
...
     public static final long MAX_VECTOR_TERM_SIZE = SAI_MAX_VECTOR_TERM_SIZE.getSizeInBytes();
-    public static final String TERM_OVERSIZE_MESSAGE = "Can't add term of column {} to index for key: {}, term size {} " +
+    public static final String TERM_OVERSIZE_MESSAGE = "Can't add term of column {} to index for: {}, term size {} " +
                                                        "max allowed size {}, use analyzed = true (if not yet set) for that column.";
 
     private final IndexDescriptor indexDescriptor;
-    private final IndexContext indexContext;
+    private final StorageAttachedIndex index;
+    private final IndexTermType indexTermType;
+    private final IndexIdentifier indexIdentifier;
```
adelapena: Maybe we should choose between storing a reference to the `StorageAttachedIndex` and storing references to its relevant components (`IndexTermType`, `IndexIdentifier`, `AbstractAnalyzer`, `IndexMetrics` and `IndexWriterConfig`). Having references to both the container and its contents might feel redundant, wdyt?

src/java/org/apache/cassandra/index/sai/memory/TrieMemoryIndex.java
```diff
@@ -65,22 +65,20 @@ public class TrieMemoryIndex extends MemoryIndex
 
     private final InMemoryTrie<PrimaryKeys> data;
     private final PrimaryKeysReducer primaryKeysReducer;
-    private final AbstractAnalyzer.AnalyzerFactory analyzerFactory;
-    private final AbstractType<?> validator;
+    private final IndexTermType indexTermType;
```
adelapena: We already have a reference to `StorageAttachedIndex` in the parent class, and not that many uses of its contained `IndexTermType`, so maybe we can get rid of this attribute and use `index.termType()` instead?

src/java/org/apache/cassandra/index/sai/plan/Expression.java
```diff
@@ -26,16 +26,31 @@
...
-import org.apache.cassandra.db.marshal.AbstractType;
-import org.apache.cassandra.index.sai.IndexContext;
+import org.apache.cassandra.index.sai.StorageAttachedIndex;
 import org.apache.cassandra.index.sai.analyzer.AbstractAnalyzer;
-import org.apache.cassandra.index.sai.utils.TypeUtil;
+import org.apache.cassandra.index.sai.analyzer.NoOpAnalyzer;
+import org.apache.cassandra.index.sai.utils.IndexTermType;
 
-public class Expression
+public interface Expression
```
adelapena: The interface could use some JavaDoc explaining what an `Expression` is, and that we have two subclasses for indexed/not-indexed columns.

src/java/org/apache/cassandra/index/sai/plan/Expression.java
```diff
@@ -106,244 +110,344 @@ public Expression(IndexContext indexContext)
...
+        }
+
+        @Override
+        public AbstractAnalyzer getAnalyzer()
+        {
+            return new NoOpAnalyzer();
         }
+    }
+
+    abstract class AbstractExpression implements Expression
```
adelapena: I'm not sure there is a benefit in having a separate interface and abstract class for this.

src/java/org/apache/cassandra/index/sai/utils/IndexIdentifier.java
```diff
@@ -0,0 +1,62 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.index.sai.utils;
+
+/**
+ * This is a simple wrapper around the index identity. It's primary purpose is to isolate classes that only need
```
adelapena:
```suggestion
 * This is a simple wrapper around the index identity. Its primary purpose is to isolate classes that only need
```

src/java/org/apache/cassandra/index/sai/utils/IndexIdentifier.java
```diff
@@ -0,0 +1,62 @@
...
+ * access to the identity from the main index classes. This is useful in testing but also makes it easier to pass
+ * the log message wrapper {@link #logMessage(String)} to class that don't need any other information about the index.
```
adelapena:
```suggestion
 * the log message wrapper {@link #logMessage(String)} to classes that don't need any other information about the index.
```

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
```diff
@@ -778,6 +800,7 @@ public void truncateWithBuiltIndexes()
     }
 
     @Test
+    @Ignore("Ignored until CASSANDRA-19032")
```
adelapena: We shouldn't ignore flaky tests.

test/unit/org/apache/cassandra/index/sai/SAITester.java
```diff
@@ -250,16 +257,59 @@ public void corrupt(File file) throws IOException
...
+        IndexMetadata indexMetadata = IndexMetadata.fromSchemaMetadata(columnName, IndexMetadata.Kind.CUSTOM, options);
+
+        ColumnFamilyStore cfs = MockSchema.newCFS(table);
+
+        return new StorageAttachedIndex(cfs, indexMetadata);
+    }
+
+    public static IndexTermType createIndexTermType(AbstractType<?> cellType)
+    {
+        return IndexTermType.create(ColumnMetadata.regularColumn("sai", "internal", "val", cellType), Collections.EMPTY_LIST, IndexTarget.Type.SIMPLE);
```
adelapena:
```suggestion
        return IndexTermType.create(ColumnMetadata.regularColumn("sai", "internal", "val", cellType), Collections.emptyList(), IndexTarget.Type.SIMPLE);
```

test/unit/org/apache/cassandra/index/sai/utils/IndexTermTypeTest.java
```diff
@@ -142,25 +172,25 @@ private static void testCollectionType(BiFunction<AbstractType<?>, Boolean, Abst
...
+            assertEquals(elementType.getType(), indexTermType(nonFrozenCollection, IndexTarget.Type.VALUES).indexType());
             nonFrozenCollectionTester.accept(elementType.getType(), nonFrozenCollection);
         }
     }
 
-    private static AbstractType<?> cellValueType(AbstractType<?> type, IndexTarget.Type indexType)
+    private static IndexTermType indexTermType(AbstractType<?> type, IndexTarget.Type indexType)
     {
-        return TypeUtil.cellValueType(column(type), indexType);
+        return IndexTermType.create(column(type), Collections.EMPTY_LIST, indexType);
```
adelapena:
```suggestion
        return IndexTermType.create(column(type), Collections.emptyList(), indexType);
```

test/unit/org/apache/cassandra/index/sai/disk/v1/MetadataTest.java
```diff
@@ -52,21 +51,20 @@ public class MetadataTest extends SAIRandomizedTester
...
-    private IndexContext indexContext;
+    private IndexIdentifier indexIdentifier;
 
     @Before
     public void setup() throws Throwable
     {
         indexDescriptor = newIndexDescriptor();
-        String index = newIndex();
-        indexContext = SAITester.createIndexContext(index, UTF8Type.instance);
+        indexIdentifier = SAITester.createIndexIdentifier("test", "test", newIndex());
```
maedhroz:
```suggestion
        indexIdentifier = createIndexIdentifier("test", "test", newIndex());
```
nit

test/unit/org/apache/cassandra/index/sai/disk/v1/TermsReaderTest.java
```diff
@@ -72,18 +71,17 @@ private void doTestTermsIteration() throws IOException
     {
         final int terms = 70, postings = 2;
         final IndexDescriptor indexDescriptor = newIndexDescriptor();
-        final String index = newIndex();
-        final IndexContext indexContext = SAITester.createIndexContext(index, UTF8Type.instance);
+        final IndexIdentifier indexIdentifier = SAITester.createIndexIdentifier("test", "test", newIndex());
```
maedhroz:
```suggestion
        final IndexIdentifier indexIdentifier = createIndexIdentifier("test", "test", newIndex());
```
nit

test/unit/org/apache/cassandra/index/sai/disk/v1/TermsReaderTest.java
```diff
@@ -99,22 +97,21 @@ private void doTestTermsIteration() throws IOException
     private void testTermQueries(int numTerms, int numPostings) throws IOException
     {
         final IndexDescriptor indexDescriptor = newIndexDescriptor();
-        final String index = newIndex();
-        final IndexContext indexContext = SAITester.createIndexContext(index, UTF8Type.instance);
+        final IndexIdentifier indexIdentifier = SAITester.createIndexIdentifier("test", "test", newIndex());
```
maedhroz:
```suggestion
        final IndexIdentifier indexIdentifier = createIndexIdentifier("test", "test", newIndex());
```
nit

test/unit/org/apache/cassandra/index/sai/disk/v1/bbtree/BlockBalancedTreeQueriesTest.java
```diff
@@ -40,7 +40,7 @@ public class BlockBalancedTreeQueriesTest extends SAIRandomizedTester
     @Test
     public void testMatchesAll()
     {
-        Expression expression = new Expression(SAITester.createIndexContext("meh", Int32Type.instance));
+        Expression expression = Expression.create(SAITester.createMockIndex("meh", Int32Type.instance));
```
maedhroz:
```suggestion
        Expression expression = Expression.create(createMockIndex("meh", Int32Type.instance));
```

test/unit/org/apache/cassandra/index/sai/disk/v1/bbtree/BlockBalancedTreeQueriesTest.java
```diff
@@ -213,7 +213,7 @@ private byte[] toSortableBytes(int value)
 
     private Expression buildExpression(Operator op, int value)
     {
-        Expression expression = new Expression(SAITester.createIndexContext("meh", Int32Type.instance));
+        Expression expression = Expression.create(SAITester.createMockIndex("meh", Int32Type.instance));
```
maedhroz:
```suggestion
        Expression expression = Expression.create(createMockIndex("meh", Int32Type.instance));
```

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
```diff
@@ -21,33 +21,32 @@
...
 import org.apache.lucene.store.IndexInput;
 
 import static net.bytebuddy.matcher.ElementMatchers.named;
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.junit.Assert.assertFalse;
-
 import static org.apache.cassandra.distributed.api.Feature.GOSSIP;
 import static org.apache.cassandra.distributed.api.Feature.NETWORK;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertFalse;
```
maedhroz, line 49: nit: some import reordering here that might not have been strictly necessary

test/unit/org/apache/cassandra/index/sai/cql/UnindexedExpressionsTest.java
```diff
@@ -0,0 +1,105 @@
...
+    {
+        createTable("CREATE TABLE %s (pk int primary key, val1 int, val2 map<int, text>)");
+        createIndex("CREATE INDEX ON %s(val1) USING 'sai'");
+
+        execute("INSERT INTO %s (pk, val1, val2) VALUES (1, 1, {1 : '1', 2 : '2', 3 : '3'})");
+        execute("INSERT INTO %s (pk, val1, val2) VALUES (2, 2, {2 : '2', 3 : '3', 4 : '4'})");
+        execute("INSERT INTO %s (pk, val1, val2) VALUES (3, 3, {3 : '3', 4 : '4', 5 : '5'})");
+        execute("INSERT INTO %s (pk, val1, val2) VALUES (4, 4, {4 : '4', 5 : '5', 6 : '6'})");
+
+        assertInvalidMessage(StatementRestrictions.REQUIRES_ALLOW_FILTERING_MESSAGE, "SELECT pk FROM %s WHERE val1 = 1 AND val2 CONTAINS KEY 1");
```
maedhroz, line 99: nit: Maybe for good measure we can throw in...

```
assertInvalidMessage(StatementRestrictions.REQUIRES_ALLOW_FILTERING_MESSAGE, "SELECT pk FROM %s WHERE val1 = 1 AND val2 CONTAINS '2'");
```

This test seems to hit all the operators and literals/numerics, so coverage is pretty good.

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
```diff
@@ -629,45 +629,53 @@ public void shouldReleaseIndexFilesAfterDroppingLastIndex() throws Throwable
...
         rows = executeNet("SELECT id1 FROM %s WHERE v1>=0");
         assertEquals(2, rows.all().size());
         rows = executeNet("SELECT id1 FROM %s WHERE v2='0'");
         assertEquals(2, rows.all().size());
 
-        dropIndex("DROP INDEX %s." + numericIndexName);
-        verifyIndexFiles(numericIndexContext, literalIndexContext, 0, 2);
-        verifySSTableIndexes(numericIndexName, 2, 0);
-        verifySSTableIndexes(literalIndexName, 2, 2);
+        dropIndex("DROP INDEX %s." + numericIndexIdentifier.indexName);
```
maedhroz: It would actually be nice to just have a method like `dropIndex(numericIndexIdentifier);` given the syntax for DROP is universal and doesn't really need to be constantly repeated in particular tests.

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
maedhroz: If CASSANDRA-19032 is not caused by this PR/patch, it's reasonable to tackle it in a separate issue?

src/java/org/apache/cassandra/index/sai/disk/format/OnDiskFormat.java
mike-tr-adamson: Correct. I have changed this to "an IndexTermType and/or IndexIdentifier". I had been playing around with either passing the `StorageAttachedIndex` around or passing the lighter weight classes and hadn't updated the comment on the final choice.

src/java/org/apache/cassandra/index/sai/disk/v1/SSTableIndexWriter.java
mike-tr-adamson: Agreed. I have removed storing the reference to the index and am storing the `IndexMetrics` and the `IndexWriterConfig` instead. I also removed the null checks on the index metrics in the code because the `StorageAttachedIndex` always creates the `IndexMetrics` now. The null check was a throwback to the `IndexContext`.

src/java/org/apache/cassandra/index/sai/view/IndexViewManager.java
```diff
@@ -28,34 +28,40 @@
...
  * sure per-sstable {@link SSTableContext} are in-sync.
  */
 public class IndexViewManager
 {
     private static final Logger logger = LoggerFactory.getLogger(IndexViewManager.class);
     
-    private final IndexContext context;
+    private final StorageAttachedIndex index;
+    private final IndexTermType indexTermType;
+    private final IndexIdentifier indexIdentifier;
```
maedhroz: nit: Could we get by w/ having just the `index` field, given it contains both the term type and the identifier?

src/java/org/apache/cassandra/index/sai/view/IndexViewManager.java
```diff
@@ -139,11 +145,70 @@ public void drop(Collection<SSTableReader> sstablesToRebuild)
...
         {
             index.markObsolete();
         }
     }
+
+    /**
+     * @return the indexes that are built on the given SSTables on the left and corrupted indexes'
+     * corresponding contexts on the right
+     */
+    private Pair<Collection<SSTableIndex>, Collection<SSTableContext>> getBuiltIndexes(Collection<SSTableContext> sstableContexts, IndexValidation validation)
```
maedhroz, line 154: Yeah, moving this method here made a ton of sense.

src/java/org/apache/cassandra/index/sai/plan/Expression.java
mike-tr-adamson: Agreed, I have made `Expression` an `abstract class`.

src/java/org/apache/cassandra/index/sai/memory/MemtableIndex.java
```diff
@@ -29,26 +29,27 @@
...
 
 public class MemtableIndex implements MemtableOrdering
 {
-    private final MemoryIndex index;
+    private final MemoryIndex memoryIndex;
     private final LongAdder writeCount = new LongAdder();
     private final LongAdder estimatedMemoryUsed = new LongAdder();
 
-    public MemtableIndex(IndexContext indexContext)
+    public MemtableIndex(StorageAttachedIndex memoryIndex)
```
maedhroz:
```suggestion
    public MemtableIndex(StorageAttachedIndex index)
```

src/java/org/apache/cassandra/index/sai/metrics/ColumnQueryMetrics.java
```diff
@@ -21,15 +21,15 @@
...
 import com.codahale.metrics.Timer;
-import org.apache.cassandra.index.sai.IndexContext;
+import org.apache.cassandra.index.sai.StorageAttachedIndex;
 
 import static org.apache.cassandra.metrics.CassandraMetricsRegistry.Metrics;
 
 public abstract class ColumnQueryMetrics extends AbstractMetrics
 {
-    protected ColumnQueryMetrics(IndexContext indexContext)
+    protected ColumnQueryMetrics(StorageAttachedIndex index)
```
maedhroz: Here and in `IndexMetrics` would it make more sense to just use `IndexIdentifier` instead of having to pass the whole `StorageAttachedIndex`?

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
mike-tr-adamson: I didn't want to get bogged down in looking at that issue here. AFAICT the truncate issue has been around since we merged CEP-7 but it is very timing critical and for some reason, we have hit it here. I would prefer not to hold this ticket up with an existing issue.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/NumericIndexSegmentSearcher.java
```diff
@@ -82,25 +82,25 @@ public long indexFileCacheSize()
...
+            throw new IllegalArgumentException(index.identifier().logMessage("Unsupported expression during index query: " + exp));
         }
     }
 
     @Override
     public String toString()
     {
         return MoreObjects.toStringHelper(this)
-                          .add("indexContext", indexContext)
+                          .add("index", index)
```
maedhroz, line 103: The output changes slightly here I think, due to SAI's `toString()` including the table name. Not sure if that matters. Maybe there's a case for just reusing the `IndexIdentifier#toString()` in SAI if we don't care about the table name in a non-logging context.

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
mike-tr-adamson: I'm not sure I follow this one. The import ordering looks correct here.

src/java/org/apache/cassandra/index/sai/disk/v1/MemtableIndexWriter.java
```diff
@@ -55,19 +55,25 @@ public class MemtableIndexWriter implements PerColumnIndexWriter
     private static final Logger logger = LoggerFactory.getLogger(MemtableIndexWriter.class);
 
     private final IndexDescriptor indexDescriptor;
-    private final IndexContext indexContext;
+    private final IndexTermType indexTermType;
+    private final IndexIdentifier indexIdentifier;
+    private final IndexMetrics indexMetrics;
```
maedhroz, line 60: nit: Another one of these cases where I guess we could just pass and store the SAI...if it doesn't make testing harder...

src/java/org/apache/cassandra/index/sai/disk/v1/V1OnDiskFormat.java
```diff
@@ -24,6 +24,9 @@
 import java.util.Set;
 
 import com.google.common.annotations.VisibleForTesting;
+
+import org.apache.cassandra.index.sai.utils.IndexIdentifier;
+import org.apache.cassandra.index.sai.utils.IndexTermType;
```
maedhroz: nit: import order

src/java/org/apache/cassandra/index/sai/disk/SSTableIndex.java
```diff
@@ -238,23 +247,25 @@ public boolean equals(Object o)
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
         SSTableIndex other = (SSTableIndex)o;
-        return Objects.equal(sstableContext, other.sstableContext) && Objects.equal(indexContext, other.indexContext);
+        return Objects.equal(sstableContext, other.sstableContext) &&
+               Objects.equal(indexTermType, other.indexTermType) &&
+               Objects.equal(indexIdentifier, other.indexIdentifier);
```
maedhroz, line 250: Does `IndexIdentifier` need to implement `equals()` and `hashCode()`?

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
maedhroz: I think I meant `import net.bytebuddy.ByteBuddy;` et al. above, not the statics

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
```diff
@@ -133,22 +133,16 @@ public List<DataRange> dataRanges()
...
-                                                    cfs.getPartitioner(),
-                                                    cfs.getComparator(),
-                                                    expression.column(),
-                                                    IndexTarget.Type.VALUES,
-                                                    null)
-                                 : indexes.iterator().next().getIndexContext();
+    public AbstractAnalyzer analyzerFor(RowFilter.Expression expression)
+    {
+        StorageAttachedIndex index = indexFor(expression);
+        return index == null ? new NoOpAnalyzer() : index.analyzer();
```
maedhroz: I know the `NoOpAnalyzer` makes some other parts of the code simpler perhaps, but the object creations are still kind of bothersome when we don't transform the value.

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
```diff
@@ -248,10 +242,11 @@ public void finish()
     public KeyRangeIterator getTopKRows(RowFilter.Expression expression)
     {
         assert expression.operator() == Operator.ANN;
-        var planExpression = new Expression(getContext(expression))
-                             .add(Operator.ANN, expression.getIndexValue().duplicate());
+        StorageAttachedIndex index = indexFor(expression);
+        assert index != null;
```
maedhroz, line 244: nit: worth a quick message?

src/java/org/apache/cassandra/index/sai/metrics/ColumnQueryMetrics.java
mike-tr-adamson: It makes sense here but not in the `IndexMetrics`. That uses a number of methods on the index for metrics calculations.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/NumericIndexSegmentSearcher.java
mike-tr-adamson: Yes, I don't think the table name is important in this case. If we find that we think the table name is important we can always add it to the `IndexIdentifier.toString`

src/java/org/apache/cassandra/index/sai/disk/v1/MemtableIndexWriter.java
mike-tr-adamson: My intent here had always been to only pass references to the classes that were being used. Passing a `StorageAttachedIndex` makes sense in the case of `IndexViewManager` because the index is needed to create the `SSTableIndex`. In this case we don't need an index.

EDIT: The rebase on top of current changes means that we now need the index in this class so I've changed to just using the index and have removed the other references.

src/java/org/apache/cassandra/index/sai/disk/SSTableIndex.java
mike-tr-adamson: Yes, it does. I've added them.

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
mike-tr-adamson: I agree. I've been debating this because I don't like the extra code for the noop analyzer. As it stands all string handling goes through the analyzer path, where a good number probably don't use an analyzer even on indexed strings.

test/unit/org/apache/cassandra/index/sai/cql/StorageAttachedIndexDDLTest.java
adelapena: We sure don't need to fix it here, but we can't commit this patch with the test marked as ignored. We can instead ignore the test failures that might show on CI.

test/distributed/org/apache/cassandra/distributed/test/sai/IndexStreamingFailureTest.java
mike-tr-adamson: They seem to be in the order set by the codestyle. If I reorder them they come out in the order they currently are.

src/java/org/apache/cassandra/index/sai/disk/format/OnDiskFormat.java
adelapena: I think this one hasn't been updated in the last push.

adelapena (approved):
Changes look good to me, except for [this nit](https://github.com/apache/cassandra/pull/2916#discussion_r1400714766) that was missed in the last push.

src/java/org/apache/cassandra/index/sai/disk/format/OnDiskFormat.java
mike-tr-adamson: Sorry, I misinterpreted your comment and didn't remove this section. I have now removed it.

PR #2919, opened 2023-11-19
CASSANDRA-19015 Nodetool 'tablestats' formatting uses inconsistent significant digits

zaaath:
### Before
Nodetool reports milliseconds (ms) with anywhere from 3 to 15 significant digits.  Ratios use five or sixteen decimal places.  Averages use 1 or 13 decimal places.
Example:
```bash
Keyspace: system
        Read Count: 18
        Read Latency: 5.364611111111111 ms
        Write Count: 22
        Write Latency: 0.6970454545454545 ms
        Pending Flushes: 0
                Table: IndexInfo
...
                SSTable Compression Ratio: 1.14583
...
                Local read count: 1
                Local read latency: 17.084 ms
                Local write count: 1
                Local write latency: 1.916 ms
                Local read/write ratio: 1.00000
...
                Bloom filter false ratio: 0.00000
...
                Average live cells per slice (last five minutes): 1.0
                Maximum live cells per slice (last five minutes): 1
                Average tombstones per slice (last five minutes): 1.0
                Maximum tombstones per slice (last five minutes): 1
                Droppable tombstone ratio: 0.00000
```

### After
Milliseconds use 3 decimal places 
Ratios use 3 decimal places (tenths of a percent)
Averages use 2 decimal places
Example:
```bash
Keyspace: system
        Read Count: 18
        Read Latency: 5.084 ms
        Write Count: 22
        Write Latency: 0.566 ms
        Pending Flushes: 0
                Table: IndexInfo
...
                SSTable Compression Ratio: 1.146
...
                Local read count: 1
                Local read latency: 14.237 ms
                Local write count: 1
                Local write latency: 0.770 ms
                Local read/write ratio: 1.000
...
                Bloom filter false ratio: 0.000
...
                Average live cells per slice (last five minutes): 61.43
                Maximum live cells per slice (last five minutes): 72
                Average tombstones per slice (last five minutes): 1.00
                Maximum tombstones per slice (last five minutes): 1
                Droppable tombstone ratio: 0.000
```

### How to reproduce
`nodetool tablestats`

### Tests
`ant testsome -Dtest.name=org.apache.cassandra.utils.FBUtilitiesTest`
`ant testsome -Dtest.name=org.apache.cassandra.tools.nodetool.stats.TableStatsPrinterTest`
`ant testsome -Dtest.name=org.apache.cassandra.tools.nodetool.TableStatsTest`

patch by @zaaath; reviewed by @smiklosovic for [CASSANDRA-19015](https://issues.apache.org/jira/browse/CASSANDRA-19015)

### DIFF
--- a/src/java/org/apache/cassandra/tools/nodetool/stats/TableStatsHolder.java
+++ b/src/java/org/apache/cassandra/tools/nodetool/stats/TableStatsHolder.java
@@ -314,7 +314,7 @@ public class TableStatsHolder implements StatsHolder
                 statsTable.bytesUnrepaired = bytesUnrepaired != null ? bytesUnrepaired : 0;
                 statsTable.bytesPendingRepair = bytesPendingRepair != null ? bytesPendingRepair : 0;
 
-                statsTable.sstableCompressionRatio = probe.getColumnFamilyMetric(keyspaceName, tableName, "CompressionRatio");
+                statsTable.sstableCompressionRatio = toDouble(probe.getColumnFamilyMetric(keyspaceName, tableName, "CompressionRatio"));
                 Object estimatedPartitionCount = probe.getColumnFamilyMetric(keyspaceName, tableName, "EstimatedPartitionCount");
                 if (Long.valueOf(-1L).equals(estimatedPartitionCount))
                 {
@@ -388,6 +388,15 @@ public class TableStatsHolder implements StatsHolder
                 statsKeyspace.tables.add(statsTable);
             }
             keyspaces.add(statsKeyspace);
+        }
+    }
+
+    private double toDouble(Object valueObj) {
+        if (valueObj instanceof Number)
+        {
+            return ((Number) valueObj).doubleValue();
+        } else {
+            return Double.NaN;
         }
     }
 
--- a/src/java/org/apache/cassandra/tools/nodetool/stats/TableStatsHolder.java
+++ b/src/java/org/apache/cassandra/tools/nodetool/stats/TableStatsHolder.java
@@ -274,6 +274,7 @@ public class TableStatsHolder implements StatsHolder
                 Long bytesUnrepaired = null;
                 Long bytesPendingRepair = null;
                 Double sstableCompressionRatio = null;
+                Double bloomFilterFalseRatio = null;
 
                 try
                 {
@@ -287,6 +288,7 @@ public class TableStatsHolder implements StatsHolder
                     bytesUnrepaired = (Long) probe.getColumnFamilyMetric(keyspaceName, tableName, "BytesUnrepaired");
                     bytesPendingRepair = (Long) probe.getColumnFamilyMetric(keyspaceName, tableName, "BytesPendingRepair");
                     sstableCompressionRatio = (Double) probe.getColumnFamilyMetric(keyspaceName, tableName, "CompressionRatio");
+                    bloomFilterFalseRatio = (Double) probe.getColumnFamilyMetric(keyspaceName, tableName, "RecentBloomFilterFalseRatio");
                 }
                 catch (RuntimeException e)
                 {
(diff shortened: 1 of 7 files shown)


src/java/org/apache/cassandra/utils/FBUtilities.java
```diff
@@ -865,6 +865,49 @@ public static String prettyPrintMemory(long size)
...
+    {
+        return String.format("%.3f ms", latency);
+    }
+
+    /**
+     * Formats a ratio value for display, rounds it to three decimal places.
+     * For example, "10.000", "1.000", "0.050", "0.001", "0.000", "NaN".
+     * @param ratioObj   Ratio to print.
+     */
+    public static String prettyPrintRatio(Object ratioObj)
```
zaaath: prettyPrintRatio(Object ratioObj) signature could be changed to prettyPrintRatio(Number ratio), but I can't completely eliminate the ifs. And actually, I'd rather stick with prettyPrintRatio(Object ratioObj).
Here's why: String.format raises an exception when format specifiers don't match the format arguments like in:
```
jshell> String.format("%.3f ms", (Number)123)
|  Exception java.util.IllegalFormatConversionException: f != java.lang.Integer 
```
And then there is also the "null" case.

So because sstableCompressionRatio and bloomFilterFalseRatio are untyped, we have two options:
1. Branch type casting
2. Handle IllegalFormatConversionException
And the third option is to refactor StatsTable.sstableCompressionRatio and StatsTable.bloomFilterFalseRatio to be typed as "double". I assume that the third option is off-the-table since it might result in too many changes.

smiklosovic: @zaaath could you please try the third option?

zaaath: @smiklosovic sure, working on that.

zaaath: @smiklosovic completed, refactored StatsTable.sstableCompressionRatio and StatsTable.bloomFilterFalseRatio to be of the "double" type instead of "Object"

Is there anything else that needs to be handled?

src/java/org/apache/cassandra/tools/nodetool/stats/TableStatsHolder.java
```diff
@@ -391,6 +391,15 @@ private void initializeKeyspaces(NodeProbe probe, boolean ignore, List<String> t
         }
     }
 
+    private double toDouble(Object valueObj) {
```
smiklosovic: I am not too big fan of this method. Don't we know already that these metrics are of type double looking where they are produced? Look at "TableMetrics" class for e.g. "CompressionRatio" there. We know this is a double already and we are not supposed to use nodetool of one version with Cassandra node of another. We see that it was cast in StatsTableComparator to Double because we know it was like that.

For extra safety, once we cast it to Double, it might be `null`, there are null checks in this class if you look around so I would apply same logic here and remove this method. 

You also need to fix `TableStatsPrinterTest` class.

zaaath: @smiklosovic done, removed `toDouble` and fixed `TableStatsPrinterTest`. Let me know how it looks.

As you noted, it’s likely safe to assume that `CompressionRatio` is of the `double` type because that’s how it’s defined in `TableMetrics.compressionRatio`. `RecentBloomFilterFalseRatio` is of the `double` type as well which is defined in `BloomFilterMetrics.BloomFilterFalseRatio`.

If you get a chance, could you explain:
1. Why did you find the `toDouble` approach inappropriate? I thought it’s more robust and better semantically.
2. How did you find out that `TableStatsPrinterTest` fails? Did you have to run the full test suite? Or did the CI pipeline catch this?

smiklosovic: 1. I did not think it was necessary
2. I just know that test exists ...

PR #2924, opened 2023-11-20
CASSANDRA-18839: Catch SSLHandshakeExceptions exceptions

jameshu15869:
This patch prevents `SSLExceptions` from flooding the log with the stack trace. Addresses [CASSANDRA-18839](https://issues.apache.org/jira/browse/CASSANDRA-19015?filter=-3).

### DIFF
--- a/src/java/org/apache/cassandra/transport/ExceptionHandlers.java
+++ b/src/java/org/apache/cassandra/transport/ExceptionHandlers.java
@@ -137,7 +137,7 @@ public class ExceptionHandlers
         }
         else if (Throwables.anyCauseMatches(cause, t -> t instanceof SSLException))
         {
-            // logger.warn("SSLException in client networking with peer {} {}", clientAddress, cause.getMessage());
+            ClientMetrics.instance.markSslHandshakeException();
             NoSpamLogger.log(logger, NoSpamLogger.Level.WARN, 1, TimeUnit.MINUTES, "SSLException in client networking with peer {} {}", clientAddress, cause.getMessage());
         }
         else
--- a/src/java/org/apache/cassandra/transport/ExceptionHandlers.java
+++ b/src/java/org/apache/cassandra/transport/ExceptionHandlers.java
@@ -23,8 +23,6 @@ import java.net.SocketAddress;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
-import javax.net.ssl.SSLException;
-
 import com.google.common.base.Predicate;
 import com.google.common.collect.ImmutableSet;
 
@@ -93,13 +91,6 @@ public class ExceptionHandlers
                     JVMStabilityInspector.inspectThrowable(cause);
                 }
             }
-
-            if (Throwables.anyCauseMatches(cause, t -> t instanceof SSLException))
-            {
-                logger.warn("SSLException in client networking with peer {}", ctx.channel().remoteAddress(), cause);
-                return;
-            }
-            
             if (DatabaseDescriptor.getClientErrorReportingExclusions().contains(ctx.channel().remoteAddress()))
             {
                 // Sometimes it is desirable to ignore exceptions from specific IPs; such as when security scans are
--- a/src/java/org/apache/cassandra/transport/ExceptionHandlers.java
+++ b/src/java/org/apache/cassandra/transport/ExceptionHandlers.java
@@ -91,6 +91,7 @@ public class ExceptionHandlers
                     JVMStabilityInspector.inspectThrowable(cause);
                 }
             }
+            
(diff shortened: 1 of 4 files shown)


src/java/org/apache/cassandra/transport/PreV5Handlers.java
```diff
@@ -327,6 +330,13 @@ public void exceptionCaught(final ChannelHandlerContext ctx, Throwable cause)
                 logger.debug("Excluding client exception for {}; address contained in client_error_reporting_exclusions", ctx.channel().remoteAddress(), cause);
                 return;
             }
+
+            if (Throwables.anyCauseMatches(cause, t -> t instanceof SSLException))
+            {
+                logger.warn("SSLException in client networking with peer {} {}", ctx.channel().remoteAddress(), cause.getMessage());
```
Maxwell-Guo: I think it is better to move this part of code into the method logClientNetworkingExceptions() below.

Maxwell-Guo: And if the log is frequent, we can use NoSpamLogger.log.
Besides, what about add some metric like [protocolException metric](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/metrics/ClientMetrics.java#L51)has done, we can   separate SSL exceptions from unknownException.

smiklosovic: Agree with @Maxwell-Guo . Ping me in the ticket once addressed. Thanks

jameshu15869: That makes sense to me, although moving to `logClientNetworkingExceptions()` looks like we lose the `ctx` variable that provides the host we tried to connect to. Should we still move to `logClientNetworkingExceptions`?

I also added a call to `ClientMetrics.instance.markProtocolException();` since it isn't too much trouble.

smiklosovic: @jameshu15869  just put `ctx.channel().remoteAddress()` into `logClientNetworkingExceptions` as the second parameter of that method and move that `if` there and log it with NoSpamLogger.

Maxwell-Guo: > I also added a call to ClientMetrics.instance.markProtocolException(); since it isn't too much trouble.

My personal suggestion is to create a new metric, refer to protocolexception, but if you are reusing the protocolexception metric, I may think it is better not to add this metric, as it will cause confusion. I think SSLexception is different from ProtocolException

src/java/org/apache/cassandra/transport/ExceptionHandlers.java
```diff
@@ -133,6 +135,11 @@ else if (Throwables.anyCauseMatches(cause, t -> t instanceof Errors.NativeIoExce
             ClientMetrics.instance.markUnknownException();
             logger.trace("Native exception in client networking", cause);
         }
+        else if (Throwables.anyCauseMatches(cause, t -> t instanceof SSLException))
+        {
+            // logger.warn("SSLException in client networking with peer {} {}", clientAddress, cause.getMessage());
```
Maxwell-Guo: delete this line.

src/java/org/apache/cassandra/transport/PreV5Handlers.java
smiklosovic: yes, it feels like a different metric, not protocol exception one

src/java/org/apache/cassandra/transport/ExceptionHandlers.java
jameshu15869: Sorry, I totally overlooked this

src/java/org/apache/cassandra/transport/PreV5Handlers.java
jameshu15869: Makes sense - I added a `SslHandshakeException` metric following the `ProtocolException` model.

Maxwell-Guo: it seems not easy to mock a sslexception for testing. Though i think the code here is ok . But do we need a test for this added metric  ? @smiklosovic @driftx

driftx: This is mostly a cosmetic problem so I don't think metrics or a test are strictly necessary.

smiklosovic: A new metric like this is pretty harmless, why not to add it? As Brandon said, a test is not strictly necessary. I am on the same page there. I think the PR is good as is. As I said on the ticket, I will try to build this in a foreseeable future.

Maxwell-Guo (approved):

PR #2932, opened 2023-11-27
CASSANDRA-19032 Provide truncate task for SAI

mike-tr-adamson:
If a table is truncated during the initial build of an SAI index, there is a chance that the index build will get interrupted, and it won't get marked queryable. This patch provides a truncate task for SAI that marks the index queryable during truncation.

### DIFF
--- a/src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
+++ b/src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
@@ -390,10 +390,15 @@ public class StorageAttachedIndex implements Index
     public Callable<?> getTruncateTask(long truncatedAt)
     {
         /*
-         * index files will be removed as part of base sstable lifecycle in
-         * {@link LogTransaction#delete(java.io.File)} asynchronously.
+         * index files will be removed as part of base sstable lifecycle in {@link LogTransaction#delete(java.io.File)}
+         * asynchronously, but we need to mark the index queryable because if the truncation is during the initial
+         * build of the index it won't get marked queryable by the build.
          */
-        return null;
+        return () -> {
+            logger.info(indexIdentifier.logMessage("Making index queryable during table truncation"));
+            baseCfs.indexManager.makeIndexQueryable(this, Status.BUILD_SUCCEEDED);
+            return null;
+        };
     }
 
     @Override
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ jobs:
     resource_class: medium
     working_directory: ~/
     shell: /bin/bash -eo pipefail -l
-    parallelism: 4
+    parallelism: 25
     steps:
     - attach_workspace:
         at: /home/cassandra
@@ -80,7 +80,7 @@ jobs:
     - CCM_MAX_HEAP_SIZE: 1024M
     - CCM_HEAP_NEWSIZE: 256M
     - REPEATED_TESTS_STOP_ON_FAILURE: false
-    - REPEATED_UTESTS: null
+    - REPEATED_UTESTS: org.apache.cassandra.index.sai.cql.StorageAttachedIndexDDLTest#concurrentTruncateWithIndexBuilding
     - REPEATED_UTESTS_COUNT: 500
     - REPEATED_UTESTS_FQLTOOL: null
     - REPEATED_UTESTS_FQLTOOL_COUNT: 500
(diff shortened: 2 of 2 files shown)


src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
```diff
@@ -390,10 +390,15 @@ public Callable<?> getPreJoinTask(boolean hadBootstrap)
     public Callable<?> getTruncateTask(long truncatedAt)
     {
         /*
-         * index files will be removed as part of base sstable lifecycle in
-         * {@link LogTransaction#delete(java.io.File)} asynchronously.
+         * index files will be removed as part of base sstable lifecycle in {@link LogTransaction#delete(java.io.File)}
+         * asynchronously, but we need to mark the index queryable because if the truncation is during the initial
+         * build of the index it won't get marked queryable by the build.
```
maedhroz, line 395: If the index is currently marked as non-queryable, and we truncate, is this safe because truncation will remove all the on-disk artifacts (or lack of those artifacts) that made the index non-queryable in the first place? (Still thinking through this, but want to make sure there's no way we accidentally make an index queryable when we shouldn't...)

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
```diff
@@ -390,10 +390,15 @@ public Callable<?> getPreJoinTask(boolean hadBootstrap)
...
          */
-        return null;
+        return () -> {
+            logger.info(indexIdentifier.logMessage("Making index queryable during table truncation"));
```
maedhroz, line 398:
```suggestion
            logger.info(indexIdentifier.logMessage("Making index queryable during table truncation..."));
```
nit: Totally don't have to do this, but I like throwing ellipses on there when it's something we're about to do ;)

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
mike-tr-adamson: I am working on the basis that if we truncate we will delete any existing data (sstable and index). As a result, we can safely mark the index as queryable and let further errors mark it not-queryable.

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
mike-tr-adamson: I don't have an issue with that, except that I'd usually leave it for something that was going to take a little time to complete. This is just setting a flag. If you feel it needs it I'm happy for the change.

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
maedhroz: Most clusters I work on don't even allow truncation in production :p

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
maedhroz: Yeah, that makes sense. The other thing I forgot is that `makeIndexQueryable()` has its own logging, maybe adding a new `Index.Status` for truncation would allow us to just piggyback on that...

```
logger.info("Index [{}] became queryable after successful build.", name);
```

...message?

maedhroz (approved):

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
maedhroz: Looked at this a bit more, and I think we're safe. Truncation should clear any existing rationale for "non-queryability" and also should do it after interrupting ongoing builds/compactions.

michaelsembwever (approved):

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
mike-tr-adamson: It starts to get messy if we do that. There is a lot of code that depends on `BUILD_SUCCEEDED`. If we add a new status, we would need to handle that correctly. I'd prefer to leave this as is. As you say, truncation is a bad idea anyway.

michaelsembwever:
Committed https://github.com/apache/cassandra/commit/3928f2992f94aa2b4e56bee0f36a0bdb31087116

PR #2933, opened 2023-11-27
CASSANDRA-19032 Provide truncate task for SAI

mike-tr-adamson:
If a table is truncated during the initial build of an SAI index, there is a chance that the index build will get interrupted, and it won't get marked queryable. This patch provides a truncate task for SAI that marks the index queryable during truncation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ jobs:
     resource_class: medium
     working_directory: ~/
     shell: /bin/bash -eo pipefail -l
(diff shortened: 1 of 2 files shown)


maedhroz (approved):

michaelsembwever (approved):

michaelsembwever:
Committed https://github.com/apache/cassandra/commit/3928f2992f94aa2b4e56bee0f36a0bdb31087116

PR #2935, opened 2023-11-28
CASSANDRA-19018 An SAI-specific mechanism to ensure consistency isn't violated for multi-column AND queries at CL > ONE

maedhroz:


src/java/org/apache/cassandra/index/sai/disk/v1/MemtableIndexWriter.java
```diff
@@ -89,7 +89,12 @@ public void addRow(PrimaryKey key, Row row, long sstableRowId)
     @Override
     public void abort(Throwable cause)
     {
-        logger.warn(indexIdentifier.logMessage("Aborting index memtable flush for {}..."), indexDescriptor.sstableDescriptor, cause);
+        if (cause == null)
+            // This commonly occurs when a Memtable has no rows to flush, and is harmless:
+            logger.debug(indexIdentifier.logMessage("Aborting index memtable flush for {}..."), indexDescriptor.sstableDescriptor);
+        else
+            logger.warn(indexIdentifier.logMessage("Aborting index memtable flush for {}..."), indexDescriptor.sstableDescriptor, cause);
+
```
maedhroz, line 116: @mike-tr-adamson @adelapena This makes some of the tests noisier, given they don't write tons of data or write data to all columns in all cases...so I just fixed it.

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
```diff
@@ -0,0 +1,401 @@
...
+            {
+                // Make sure node 2, which is the only complete replica, can't participate in a query...
+                CLUSTER.get(2).runOnInstance(() -> {
+                    Keyspace keyspace = Objects.requireNonNull(Schema.instance.getKeyspaceInstance(KEYSPACE));
+                    SecondaryIndexManager sim = keyspace.getColumnFamilyStore(TEST_TABLE_NAME).indexManager;
+                    for (String s : Arrays.asList("ck_idx", "s_idx", "a_idx", "b_idx"))
+                        sim.makeIndexNonQueryable(sim.getIndexByName(s), Index.Status.BUILD_FAILED);
+                });
+
+                SAIUtil.waitForIndexNonQueryable(CLUSTER, KEYSPACE, 2);
```
maedhroz: There might be an even better way, but this avoids shutdown/startup overhead and makes it much faster to get through the entire suite.

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
```diff
@@ -0,0 +1,401 @@
...
+        }
+
+        // Introduce a partial update that spans replicas:
+        model.updatePartially();
+
+        if (specification.flushPartials)
+            CLUSTER.stream().forEach(i -> i.flush(KEYSPACE));
+
+        // TODO: Remove, but for now enabling trivially passes all tests:
+        //CLUSTER.get(1).nodetoolResult("repair", KEYSPACE).asserts().success();
```
maedhroz: If you disable RFP in `DataResolver` but repair here, every test trivially passes, of course.

test/distributed/org/apache/cassandra/distributed/test/sai/SAIUtil.java
```diff
@@ -49,8 +51,20 @@ public static void waitForIndexQueryable(Cluster cluster, String keyspace)
...
+
+    /**
+     * Waits until all indexes in the given keyspace become non-queryable on a specific node.
+     */
+    public static void waitForIndexNonQueryable(Cluster cluster, String keyspace, int node)
+    {
+        assertGossipEnabled(cluster);
+        final List<String> indexes = getIndexes(cluster, keyspace);
+        InetAddressAndPort addressAndPort = nodeAddress(cluster.get(node).broadcastAddress());
+        await().atMost(60, TimeUnit.SECONDS).pollInterval(1, TimeUnit.SECONDS)
```
maedhroz: The default polling interval seemed to be creating a lot of logging noise, so I bumped it to 1 second.

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
```diff
@@ -0,0 +1,401 @@
...
+        {
+            for (boolean restrictPartitionKey : new boolean[] { false, true })
+            {
+                for (boolean flushPartials : new boolean[] { false, true })
+                {
+                    for (String[] columns : new String[][] { { "ck", "a" }, { "ck", "s" }, { "s", "a" }, { "a", "b" }, { "a" } })
+                        for (boolean existing : new boolean[] { false, true })
+                            parameters.add(new Object[]{ new Specification(restrictPartitionKey, columns, readCL, existing, StatementType.UPDATE, partitionKey++, flushPartials) });
+
+                    for (String[] columns : new String[][] { { "s", "a" }, { "a", "b" }, { "a" } })
```
maedhroz: An `{ "s" }` scenario might also be useful here.

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
```diff
@@ -0,0 +1,401 @@
...
+        CLUSTER = init(Cluster.build(3).withConfig(config -> config.set("hinted_handoff_enabled", false).with(GOSSIP).with(NETWORK)).start());
+
+        // All parameterized test scenarios share the same table and attached indexes, but write to different partitions
+        // that are deleted after each scenario completes.
+        String createTableDDL = String.format("CREATE TABLE %s.%s (pk int, ck int, s int static, a int, b int, PRIMARY KEY (pk, ck)) WITH read_repair = 'NONE'",
+                                              KEYSPACE, TEST_TABLE_NAME);
+        CLUSTER.schemaChange(createTableDDL);
+        CLUSTER.disableAutoCompaction(KEYSPACE);
+
+        CLUSTER.schemaChange(String.format("CREATE INDEX ck_idx ON %s.%s(ck) USING 'sai'", KEYSPACE, TEST_TABLE_NAME));
```
adelapena, line 100: I think we are missing indexes on composite partition key columns

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
```diff
@@ -0,0 +1,401 @@
...
+    }
+    
+    static class Specification
+    {
+        final boolean restrictPartitionKey;
+        final String[] columns;
+        final ConsistencyLevel readCL;
+        final boolean existing;
+        final StatementType partialUpdateType;
+        final int partitionKey;
```
adelapena, line 114: I think it tests a single partition key per test case. SRP differs for partitions and rows, and I think RFP also groups things per partition. I remember multi-partition queries being problematic in 8272/8273, and [Python dtests for RFP](https://github.com/apache/cassandra-dtest/blob/trunk/replica_side_filtering_test.py) include them. Maybe we should test multi-partition scenarios too?

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
maedhroz: Yeah, I don't think it would be too hard to add this.

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
maedhroz: Yeah, before I wrap this up, I should make sure this new framework subsumes at least those cases from `replica_side_filtering_test.py` that deal w/ partial updates.

src/java/org/apache/cassandra/index/sai/plan/QueryViewBuilder.java
```diff
@@ -105,82 +99,18 @@ private Collection<Pair<Expression, Collection<SSTableIndex>>> getQueryView(Coll
...
-                                                         index.bounds().contains(sstableIndex.bounds().right));
-    }
-
-    // The purpose of this method is to calculate the most selective expression. This is the
-    // expression with the most sstable indexes that match the expression by term and lie
-    // within the key range being queried.
-    //
-    // The result can be null. This indicates that none of the expressions select any
-    // sstable indexes.
-    private Pair<Expression, Collection<SSTableIndex>> calculateMostSelective(Collection<Expression> expressions)
```
maedhroz, line 153: Removal mentioned here: https://issues.apache.org/jira/browse/CASSANDRA-19018?focusedCommentId=17799154&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17799154

src/java/org/apache/cassandra/index/sai/plan/FilterTree.java
```diff
@@ -96,22 +100,56 @@ private boolean localSatisfiedBy(DecoratedKey key, Unfiltered unfiltered, Row st
...
+                ColumnData data = row.getColumnData(column);
+
+                if (data == null)
+                    // Degrade to non-strict filtering if we're missing a value for a filtered column, as it could be
+                    // partially updated on another replica.
+                    return BooleanOperator.OR;
+
+                if (lastTimestamp == null)
+                    lastTimestamp = data.maxTimestamp();
+                else if (lastTimestamp != data.maxTimestamp())
```
maedhroz: The only question I threw around during development here was whether this is really safe for collections/complex types. (Max timestamp is the only timestamp for simple cells, so that's simpler.) I couldn't work through an example where the max timestamp is the same as the other column's timestamps and there is still a problematic partial update possibility, but perhaps I've just not been sufficiently creative...

src/java/org/apache/cassandra/db/filter/RowFilter.java
```diff
@@ -68,19 +68,22 @@ public abstract class RowFilter implements Iterable<RowFilter.Expression>
 
     protected final List<Expression> expressions;
 
-    protected RowFilter(List<Expression> expressions)
+    private final boolean isStrict;
```
maedhroz: This property being a part of the row filter is one of the big decisions made in this patch. One of the reasons it felt appropriate is that `RowFilter` itself is still broken around CASSANDRA-19007. The property being present here might make it easier for the logic below to adjust. I'm not sure if that belongs in this Jira, but if there is a clear path to fixing it, perhaps I can create another one and fast-follow after I commit this patch?

src/java/org/apache/cassandra/db/ReadCommand.java
```diff
@@ -1011,6 +1011,12 @@ public void close() throws Exception
     @VisibleForTesting
     public static class Serializer implements IVersionedSerializer<ReadCommand>
     {
+        private static final int IS_DIGEST = 0x01;
+        private static final int IS_FOR_THRIFT = 0x02;
+        private static final int HAS_INDEX = 0x04;
+        private static final int ACCEPTS_TRIANSIENT = 0x08;
+        private static final int STRICT_FILTERING = 0x10;
```
maedhroz: Not necessary, I know, but if feels nicer to have the constants...

src/java/org/apache/cassandra/cql3/restrictions/StatementRestrictions.java
```diff
@@ -779,7 +779,12 @@ public RowFilter getRowFilter(IndexRegistry indexRegistry, QueryOptions options)
         if (filterRestrictions.isEmpty())
             return RowFilter.none();
 
-        RowFilter filter = RowFilter.create();
+        ConsistencyLevel cl = options.getConsistency();
+        boolean isStrict = cl == ConsistencyLevel.ONE || cl == ConsistencyLevel.LOCAL_ONE || cl == ConsistencyLevel.NODE_LOCAL
+                           // With zero or one restricted mutable columns, strict filtering is safe even at higher consistency levels.
+                           || nonPrimaryKeyRestrictions.getColumnDefinitions().size() <= 1;
```
maedhroz: I think this is all we can do at the coordinator, before we know anything about the local data we need to search. Any other cases I'm missing?

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
```diff
@@ -493,8 +493,7 @@ public long getEstimatedResultRows()
     @Override
     public boolean isQueryable(Status status)
     {
-        // consider unknown status as queryable, because gossip may not be up-to-date for newly joining nodes.
-        return status == Status.BUILD_SUCCEEDED || status == Status.UNKNOWN;
+        return status == Status.BUILD_SUCCEEDED;
```
maedhroz: The main reason I changed this is that it made my more lightweight mechanism for avoiding the middle node in the test matrix work with less effort. However, I'm not entirely sure why we've ever considered UNKNOWN to be a queryable state. If gossip hasn't propagated yet, was the idea to just fire queries at the new node and hope the index is queryable locally? Is doing the simpler thing (only using BUILD_SUCCEEDED) going to cause a major availability problem?

CC @mike-tr-adamson @adelapena

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
```diff
@@ -0,0 +1,515 @@
...
+        CLUSTER.schemaChange(createTableDDL);
+        CLUSTER.disableAutoCompaction(KEYSPACE);
+
+        CLUSTER.schemaChange(String.format("CREATE INDEX pk2_idx ON %s.%s(pk2) USING 'sai'", KEYSPACE, TEST_TABLE_NAME));
+        CLUSTER.schemaChange(String.format("CREATE INDEX ck_idx ON %s.%s(ck) USING 'sai'", KEYSPACE, TEST_TABLE_NAME));
+        CLUSTER.schemaChange(String.format("CREATE INDEX s_idx ON %s.%s(s) USING 'sai'", KEYSPACE, TEST_TABLE_NAME));
+        CLUSTER.schemaChange(String.format("CREATE INDEX a_idx ON %s.%s(a) USING 'sai'", KEYSPACE, TEST_TABLE_NAME));
+        CLUSTER.schemaChange(String.format("CREATE INDEX b_idx ON %s.%s(b) USING 'sai'", KEYSPACE, TEST_TABLE_NAME));
+
+        SAIUtil.waitForIndexQueryable(CLUSTER, KEYSPACE);
```
maedhroz, line 105: Locally, I've actually also tested basic filtering w/ this test matrix. Really all you have to do it eliminate these index creations, comment out the index queryability stuff, and add ALLOW FILTERING to the model validation SELECT, and you fail about 44 of the 184 tests (although it would be more if we forced node 2 to be ignored on QUORUM validations)

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
```diff
@@ -0,0 +1,515 @@
...
+        // used in partition-restricted queries.
+        int partitionKey = 0;
+
+        for (ConsistencyLevel readCL : new ConsistencyLevel[] { ALL, QUORUM })
+        {
+            for (boolean restrictPartitionKey : new boolean[] { false, true })
+            {
+                for (boolean flushPartials : new boolean[] { false, true })
+                {
+                    for (String[] columns : new String[][] { { "ck", "a" }, { "ck", "s" }, { "s", "a" }, { "a", "b" }, { "a" }, { "s" } })
```
maedhroz: I missed two static columns here...that might be good to cover, although it doesn't really matter for SAI as much as it does for regular filtering, where that logic is actually different...

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
maedhroz: See https://github.com/maedhroz/cassandra/pull/14

src/java/org/apache/cassandra/cql3/restrictions/StatementRestrictions.java
```diff
@@ -779,7 +779,12 @@ public RowFilter getRowFilter(IndexRegistry indexRegistry, QueryOptions options)
         if (filterRestrictions.isEmpty())
             return RowFilter.none();
 
-        RowFilter filter = RowFilter.create();
+        ConsistencyLevel cl = options.getConsistency();
+        boolean isStrict = cl == ConsistencyLevel.ONE || cl == ConsistencyLevel.LOCAL_ONE || cl == ConsistencyLevel.NODE_LOCAL
```
adelapena: We have a mostly identical check to emit `SelectStatement.TOPK_CONSISTENCY_LEVEL_WARNING`. Maybe we can put this condition in a method in `ConsistencyLevel`, for example:
```java
private boolean isSingleReplica()
{
    return this == ConsistencyLevel.ONE 
            || this == ConsistencyLevel.LOCAL_ONE 
            || this == ConsistencyLevel.NODE_LOCAL;
}
```
Not sure about the method name, though. Maybe we can use `needsReconciliation` and negate the result?
```java
private boolean needsReconciliation()
{
    return this != ConsistencyLevel.ONE
            && this != ConsistencyLevel.LOCAL_ONE
            && this != ConsistencyLevel.NODE_LOCAL;
}
```

src/java/org/apache/cassandra/db/ReadCommand.java
```diff
@@ -1011,6 +1011,12 @@ public void close() throws Exception
     @VisibleForTesting
     public static class Serializer implements IVersionedSerializer<ReadCommand>
     {
+        private static final int IS_DIGEST = 0x01;
+        private static final int IS_FOR_THRIFT = 0x02;
+        private static final int HAS_INDEX = 0x04;
+        private static final int ACCEPTS_TRIANSIENT = 0x08;
```
adelapena: Nit: typo
```suggestion
        private static final int ACCEPTS_TRANSIENT = 0x08;
```

src/java/org/apache/cassandra/db/ReadCommand.java
adelapena: I'd add some JavaDoc describing the meaning of the flag.

src/java/org/apache/cassandra/db/filter/RowFilter.java
adelapena: I'd add some JavaDoc describing the meaning of the flag, and possibly mentioning this JIRA ticket.

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
```diff
@@ -171,33 +179,70 @@ public UnfilteredRowIterator queryStorage(PrimaryKey key, ReadExecutionControlle
...
+                // This usually means we are making this local index query in the context of a user query that reads 
+                // from a single replica and thus can safely perform local intersections.
+                for (Pair<Expression, Collection<SSTableIndex>> queryViewPair : queryView.view)
+                    builder.add(IndexSearchResultIterator.build(queryViewPair.left, queryViewPair.right, mergeRange, queryContext, true));
+            }
+            else
+            {
+                KeyRangeIterator.Builder repairedBuilder = KeyRangeIntersectionIterator.builder(expressions.size());
+
+                for (Pair<Expression, Collection<SSTableIndex>> queryViewPair : queryView.view)
```
adelapena, line 219: Nit: add braces for the for loop

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
```diff
@@ -171,33 +179,70 @@ public UnfilteredRowIterator queryStorage(PrimaryKey key, ReadExecutionControlle
...
+                                           ? KeyRangeIntersectionIterator.builder(expressions.size())
+                                           : KeyRangeUnionIterator.builder(expressions.size());
+        QueryViewBuilder.QueryView queryView = new QueryViewBuilder(expressions, mergeRange).build();
 
         try
         {
             maybeTriggerGuardrails(queryView);
 
-            for (Pair<Expression, Collection<SSTableIndex>> queryViewPair : queryView.view)
+            if (command.rowFilter().isStrict())
```
adelapena, line 207: I wonder if we should add a system property to disable this protection mechanism optionally. That might be helpful for use cases where users know that they aren't going to make partial updates.

src/java/org/apache/cassandra/index/sai/plan/FilterTree.java
```diff
@@ -96,22 +100,56 @@ private boolean localSatisfiedBy(DecoratedKey key, Unfiltered unfiltered, Row st
...
-                    result = op.apply(result, singletonMatch(value, filter));
+                    result = localOperator.apply(result, singletonMatch(value, filter));
                 }
 
                 // If the operation is an AND then exit early if we get a single false
-                if (op == BooleanOperator.AND && !result)
+                if (localOperator == BooleanOperator.AND && !result)
                     return false;
+
+                // If the operation is an OR then exit early if we get a single true
```
adelapena, line 133:
```suggestion
                // If the operation is a OR then exit early if we get a single true
```

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
```diff
@@ -171,33 +179,70 @@ public UnfilteredRowIterator queryStorage(PrimaryKey key, ReadExecutionControlle
...
      * <p>
-     * The results from each call to {@link IndexSearchResultIterator#build(Expression, Collection, AbstractBounds, QueryContext)}
-     * are added to a {@link KeyRangeIntersectionIterator} and returned.
+     * The results from each call to {@link IndexSearchResultIterator#build(Expression, Collection, AbstractBounds, QueryContext, boolean)}
+     * are added to a {@link KeyRangeIntersectionIterator} and returned if strict filtering is allowed.
+     * <p>
+     * If strict filtering is not allowed, indexes are split into two groups according to the repaired status of their 
+     * backing SSTables. Results from searches over the repaired group are added to a 
+     * {@link KeyRangeIntersectionIterator}, which is then added, along with results from searches on the unrepaired
+     * set, to a top-level {@link KeyRangeUnionIterator}, and returned. This is done to ensure that AND queries do not
```
adelapena, line 188: Maybe mention that the unrepaired set also includes the memtables?

src/java/org/apache/cassandra/index/sai/plan/FilterTree.java
maedhroz: Fun thing...it would be "...a union..." because "union" begins w/ a "y" sound, but usually "an" when the next word starts w/ a vowel. English is weird.

https://www.britannica.com/dictionary/eb/qa/how-do-you-know-whether-to-use-a-or-an#:~:text=Use%20a%20when%20the%20noun,the%20pronunciation%2C%20not%20the%20spelling.

src/java/org/apache/cassandra/cql3/restrictions/StatementRestrictions.java
maedhroz: > Maybe we can use needsReconciliation and negate the result?

That might work.

src/java/org/apache/cassandra/index/sai/plan/FilterTree.java
```diff
@@ -96,22 +100,56 @@ private boolean localSatisfiedBy(DecoratedKey key, Unfiltered unfiltered, Row st
...
+                    return true;
             }
         }
         return result;
     }
 
+    private BooleanOperator getLocalOperator(Row regularRow, Row staticRow)
+    {
+        // This is an AND query, but the coordinator has indicated strict filtering might not be allowed... 
+        if (baseOperator == BooleanOperator.AND && !strict)
```
maedhroz: I think we would also not have to worry about any of this garbage if there were no results of any kind from the unrepaired data...I wonder how difficult it would be to make the FilterTree aware of that...

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
maedhroz: Possibly. It might also be interesting to have some concept of column groups, where we make it illegal to update columns in a group separately.

maedhroz: The system property could be useful, although you'd really have to know what you're doing...

src/java/org/apache/cassandra/index/sai/plan/FilterTree.java
```diff
@@ -96,22 +100,56 @@ private boolean localSatisfiedBy(DecoratedKey key, Unfiltered unfiltered, Row st
...
+        {
+            Long lastTimestamp = null;
+            Iterator<ColumnMetadata> columns = expressions.keySet().iterator();
+
+            while (columns.hasNext())
+            {
+                ColumnMetadata column = columns.next();
```
adelapena: Why not use a `for` loop with an implicit iterator? As in:
```suggestion
            for (ColumnMetadata column : expressions.keySet())
            {
```

src/java/org/apache/cassandra/index/sai/plan/FilterTree.java
```diff
@@ -96,22 +100,56 @@ private boolean localSatisfiedBy(DecoratedKey key, Unfiltered unfiltered, Row st
...
+                Row row = column.kind == Kind.STATIC ? staticRow : regularRow;
+                ColumnData data = row.getColumnData(column);
```
adelapena: I think the bug only affects groups of non-PK columns, because PK columns cannot be partially updated. So we can probably skip PK columns here:
```java
if (column.isPrimaryKeyColumn())
    continue;
```
With this change `PartialUpdateHandlingTest` keeps passing, although I think it doesn't exercise the case where there are PK indexes involved.

src/java/org/apache/cassandra/index/sai/plan/FilterTree.java
adelapena: I was reading `OR` as /ɔːr/ (or /ər/, or /ɔː/, or /ə/), starting with a vowel.

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
```diff
@@ -0,0 +1,541 @@
...
+ *     involved at the implementation level, retrieving enough information to do that filtering is important.)
+ * 8.) The timestamps of partial updates and deletes, especially for single-column queries that might produce
+ *     stale matches if not resolved correctly.
+ */
+@RunWith(Parameterized.class)
+public class PartialUpdateHandlingTest extends TestBaseImpl
+{
+    private static final String TEST_TABLE_NAME = "test_partial_updates";
+    private static final int PARTITIONS_PER_TEST = 10;
+    private static final int NODES = 3;
```
adelapena: Can we not reproduce all the problematic cases with just 2 nodes and save some resources?

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
```diff
@@ -0,0 +1,541 @@
...
+            if (specification.readCL == QUORUM)
+            {
+                node2IndexStatusFrom1 = CLUSTER.get(1).callsOnInstance(() -> IndexStatusManager.instance.peerIndexStatus.remove(node2AddressAndPort)).call();
+                node2IndexStatusFrom3 = CLUSTER.get(3).callsOnInstance(() -> IndexStatusManager.instance.peerIndexStatus.remove(node2AddressAndPort)).call();
+            }
+            else
+            {
+                node2IndexStatusFrom1 = null;
+                node2IndexStatusFrom3 = null;
+            }
```
adelapena: Nit: Can we simply initialize this two vars as `null`?

adelapena: Ah, no, we can't because of the lambdas below. Please ignore.

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
```diff
@@ -0,0 +1,541 @@
...
+
+            Object[][] result = CLUSTER.coordinator(1).execute(select.toString(), specification.readCL);
```
adelapena: There might be some value in also using `executeWithPaging` here. I remember paging being tricky for the RFP/SRP combo during 8272.

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
maedhroz: I think the consensus from some discussion on ASF Slack is that we can still attempt queries against UNKNOWN replicas, but only if there aren't enough BUILD_SUCCEEDED replicas present.

CC @jasonstack @pkolaczk @JeremiahDJordan

src/java/org/apache/cassandra/db/filter/RowFilter.java
maedhroz: I've added the JavaDoc here rather than in `ReadCommand`, given `ReadCommand` is just what carries the flag over the wire.

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
maedhroz: The other difficulty w/ a system property is that whether we're making partial updates or not is a table-specific thing. The value of a system property would, I guess, be disabling this feature altogether. Of course, we didn't even do this with replica-filtering protection.

We can defer this decision a bit, I suppose, but my instinct is to not provide a flag that makes it possible to be globally incorrect. The idea of "column groups" seems like something that might work though...?

maedhroz: https://the-asf.slack.com/archives/C018YGVCHMZ/p1706041297670559

src/java/org/apache/cassandra/index/sai/plan/FilterTree.java
maedhroz: Actually, given the Memtable-attached indexes are implicitly unrepaired, this might be tricky.

src/java/org/apache/cassandra/index/sai/plan/FilterTree.java
maedhroz: That makes sense.

`pk2_idx` should test elements of composite partition keys and `ck_idx` should hit clustering keys in `PartialUpdateHandlingTest`, IIRC

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
maedhroz: Two thoughts for me here:

1.) With the cluster being shared across all tests, one more node seems not to be that problematic.
2.) The QUORUM query cases require a 3rd node, right? Otherwise they're identical to the ALL query cases?

src/java/org/apache/cassandra/db/filter/RowFilter.java
```diff
@@ -68,19 +93,35 @@ public abstract class RowFilter implements Iterable<RowFilter.Expression>
...
+    public static RowFilter create(boolean isStrict)
     {
-        return new CQLFilter(new ArrayList<>());
+        return new CQLFilter(new ArrayList<>(), isStrict);
     }
 
     public static RowFilter create(int capacity)
     {
-        return new CQLFilter(new ArrayList<>(capacity));
+        return new CQLFilter(new ArrayList<>(capacity), true);
```
ifesdjeen: Is there any downside to _not_ use strict filtering and use union by default, and optimise for all other cases?

src/java/org/apache/cassandra/db/filter/RowFilter.java
```diff
@@ -68,19 +93,35 @@ public abstract class RowFilter implements Iterable<RowFilter.Expression>
...
+    private final boolean isStrict;
+
+    protected RowFilter(List<Expression> expressions, boolean isStrict)
     {
         this.expressions = expressions;
+        this.isStrict = isStrict;
     }
 
-    public static RowFilter create()
+    public static RowFilter create(boolean isStrict)
```
ifesdjeen: I know it might be beyond the scope of this patch, but there is only one implementation of `RowFilter` with a removal of `ThriftFilter`. Maybe we should bite a bullet and remove this indirection?

src/java/org/apache/cassandra/index/sai/plan/FilterTree.java
```diff
@@ -43,14 +44,16 @@
  */
 public class FilterTree
 {
-    protected final BooleanOperator op;
+    protected final BooleanOperator baseOperator;
     protected final ListMultimap<ColumnMetadata, Expression> expressions;
     protected final List<FilterTree> children = new ArrayList<>();
+    private final boolean strict;
```
ifesdjeen: we call it `isStrict` throughout the patch, maybe rename it to the same for consistency?

ifesdjeen (approved):

src/java/org/apache/cassandra/db/filter/RowFilter.java
```diff
@@ -60,32 +83,44 @@
...
+     * @see <a href="https://issues.apache.org/jira/browse/CASSANDRA-19018">CASSANDRA-19018</a>
+     */
+    private final boolean isStrict;
 
-    public static RowFilter create()
+    protected RowFilter(List<Expression> expressions, boolean isStrict)
     {
-        return new CQLFilter(new ArrayList<>());
+        this.expressions = expressions;
+        this.isStrict = isStrict;
```
adelapena: Rather than sending `isStrict` serialized to the replicas, we could instead send a more generic `needsReconciliation` attribute, the one we get from the consistency level. Then we could keep the `RowFilter#isStrict()` method but getting the rest from the expressions. For example:
```java
public boolean isStrict()
{
    // With zero or one restricted mutable columns, strict filtering is safe even at higher consistency levels.
    return !needsReconciliation || expressions.stream().filter(e -> !e.column.isPrimaryKeyColumn()).count() <= 1;
}
```
Alternatively, we could memoize that bit in an attribute that isn't included in serialization:
```java
protected RowFilter(List<Expression> expressions, boolean needsReconciliation)
{
    this.expressions = expressions;
    this.needsReconciliation = needsReconciliation;
    this.isStrict = !needsReconciliation || expressions.stream().filter(e -> !e.column.isPrimaryKeyColumn()).count() <= 1;
}
```
The reason to send `needsReconciliation` to the replicas instead of `isStrict` is that it has more potential for reutilization. 

Specifically, it can be used in `ORDER BY` queries, like the ones in the ANN vector index. Those queries read the matching rows in the requested order in the replica, then they order the rows in PK order to prevent problems with reconciliation, and then order them again by the query order in the coordinator. If we send `needsReconciliation` to the replicas, they can consider the lack of reconciliation and send the rows to the coordinator in PK order, saving us two ordering operations. I think there have been some experiments around this and the performance improvements are huge.

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
adelapena: I think all the scenarios of concern can be reproduced with `ONE`/`ALL`. I don't see how `QUORUM`/`QUORUM` can trigger something different. Half of the tested cases use `QUORUM` reads. If we consider `ONE`/`ALL` as a good enough model, we can drop the parameters using `QUORUM` reads and use a 2-node cluster. That would make the test more than twice as fast, saving CI resources.

src/java/org/apache/cassandra/index/sai/plan/FilterTree.java
maedhroz: I found a way to do this by looking at the count of the unrepaired iterator and tracking that in the query context. Will push soon...

maedhroz: Pushed

test/distributed/org/apache/cassandra/distributed/test/sai/PartialUpdateHandlingTest.java
maedhroz: I'll reduce to ONE/ALL for now. I need to separately address the change in `IndexStatusManager` anyway...

See https://github.com/apache/cassandra/pull/2935#discussion_r1442402238

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
maedhroz: Reverted the change along w/ QUORUM/QUORUM test cases, but this needs to be fixed in a separate Jira.

src/java/org/apache/cassandra/index/sai/QueryContext.java
```diff
@@ -87,4 +89,9 @@ public VectorQueryContext vectorContext()
             vectorContext = new VectorQueryContext(readCommand);
         return vectorContext;
     }
+    
+    public boolean useStrictFiltering()
+    {
+        return readCommand.rowFilter().isStrict() || !hasUnrepairedMatches;
```
maedhroz: We arrive here w/ a non-strict row filter and no unrepaired matches in reads in `PartialUpdateHandlingTest` that happen immediately after repair (in `validateCurrent()`).

src/java/org/apache/cassandra/index/sai/plan/FilterTree.java
maedhroz: We just use the `QueryContext` now.

src/java/org/apache/cassandra/db/filter/RowFilter.java
maedhroz: The memoized version doesn't quite work, given expressions are added to the filter after it's created.

src/java/org/apache/cassandra/db/ConsistencyLevel.java
```diff
@@ -258,6 +258,11 @@ public void validateCounterForWrite(TableMetadata metadata) throws InvalidReques
             throw new InvalidRequestException("Counter operations are inherently non-serializable");
     }
 
+    public boolean needsReconciliation()
```
adelapena, line 267: Nit: this could use some JavaDoc

src/java/org/apache/cassandra/db/compaction/CompactionIterator.java
```diff
@@ -269,12 +269,11 @@ public void onMergedPartitionLevelDeletion(DeletionTime mergedDeletion, Deletion
                     {
                     }
 
-                    public Row onMergedRows(Row merged, Row[] versions)
+                    public void onMergedRows(Row merged, Row[] versions)
```
adelapena, line 272: Nit: add `@Override`

src/java/org/apache/cassandra/db/filter/RowFilter.java
```diff
@@ -60,32 +83,31 @@
...
-    public static RowFilter create()
+    protected RowFilter(List<Expression> expressions, boolean needsReconciliation)
     {
-        return new CQLFilter(new ArrayList<>());
+        this.expressions = expressions;
+        this.needsReconciliation = needsReconciliation;
     }
 
-    public static RowFilter create(int capacity)
+    public static RowFilter create(boolean needsReconciliation)
```
adelapena, line 109: Nit: This constructor could use some JavaDoc about `needsReconciliation`.

src/java/org/apache/cassandra/db/filter/RowFilter.java
```diff
@@ -111,14 +133,33 @@ private void add(Expression expression)
...
-        return expressions;
+        return needsReconciliation;
+    }
+
+    /**
+     * If this filter belongs to a read that requires reconciliation at the coordinator, and it contains an intersection
+     * on two or more non-key (and therefore mutable) columns, we cannot strictly apply it to local, unrepaired rows.
+     * When this occurs, we must downgrade the intersection to a union and allow the coordinator to filter strictly 
+     * before sending results to the client.
+     *
```
adelapena:
```suggestion
     * When this occurs, we must downgrade the intersection of expressions to a union and allow the coordinator to 
     * filter strictly before sending results to the client.
     * 
```

src/java/org/apache/cassandra/db/filter/RowFilter.java
```diff
@@ -136,7 +177,66 @@ public boolean hasExpressionOnClusteringOrRegularColumns()
...
+            if (e.column.isStatic() || e.column.isPartitionKey())
+                partitionLevelExpressions.add(e);
+            else
+                rowLevelExpressions.add(e);
+        }
+
+        long numberOfRegularColumnExpressions = rowLevelExpressions.size();
+        final boolean filterNonStaticColumns = numberOfRegularColumnExpressions > 0;
+
+        return new Transformation<BaseRowIterator<?>>()
```
adelapena:
```suggestion
        return new Transformation<>()
```

src/java/org/apache/cassandra/db/filter/RowFilter.java
```diff
@@ -136,7 +177,66 @@ public boolean hasExpressionOnClusteringOrRegularColumns()
         return false;
     }
 
-    protected abstract Transformation<BaseRowIterator<?>> filter(TableMetadata metadata, long nowInSec);
+    protected Transformation<BaseRowIterator<?>> filter(TableMetadata metadata, long nowInSec)
```
adelapena, line 194: It might be worth to comment here and/or in `isStrict()` that the multiple `filter` and `isSatisfiedBy` methods always use conjunction. Otherwise, one might wrongly understand that a non-strict filter would use disjunction in those methods.

src/java/org/apache/cassandra/index/sai/QueryContext.java
```diff
@@ -58,6 +58,8 @@ public class QueryContext
 
     public boolean queryTimedOut = false;
 
+    public boolean hasUnrepairedMatches = false;
```
adelapena, line 71: Nit: add JavaDoc

src/java/org/apache/cassandra/index/sai/QueryContext.java
```diff
@@ -87,4 +89,9 @@ public VectorQueryContext vectorContext()
             vectorContext = new VectorQueryContext(readCommand);
         return vectorContext;
     }
+    
+    public boolean hasUnrepairedMatches()
+    {
+        return hasUnrepairedMatches;
+    }
```
adelapena: This getter seems redundant given that the attribute is `public`.

src/java/org/apache/cassandra/index/sai/disk/v1/MemtableIndexWriter.java
```diff
@@ -83,13 +89,31 @@ public void addRow(PrimaryKey key, Row row, long sstableRowId)
...
+        boolean isStatic = indexTermType.columnMetadata().isStatic();
+
+        // Indexes on static columns should only track static rows, and indexes on non-static columns 
+        // should only track non-static rows. (Within a partition, the row ID for a static row will always
+        // come before any non-static row.) 
+        if (key.kind() == PrimaryKey.Kind.STATIC && isStatic || key.kind() != PrimaryKey.Kind.STATIC && !isStatic)
+        {
+            if (minKey == null)
+                minKey = key;
+            maxKey = key;
```
adelapena, line 102: This way of updating min/max keys is based on the expectation that `PerColumnIndexWriter#addRow` is always called in primary key order. Maybe we can mention that in the JavaDoc for `PerColumnIndexWriter`?

src/java/org/apache/cassandra/index/sai/plan/StorageAttachedIndexQueryPlan.java
```diff
@@ -71,18 +75,24 @@ public static StorageAttachedIndexQueryPlan create(ColumnFamilyStore cfs,
...
+            // QueryPlan#postIndexQueryFilter(). If strict filtering is not allowed, we must reject the query until the
+            // expression(s) in question are compatible with #isSatifiedBy.
             //
             // Note: For both the pre- and post-filters we need to check that the expression exists before removing it
             // because the without method assert if the expression doesn't exist. This can be the case if we are given
             // a duplicate expression - a = 1 and a = 1. The without method removes all instances of the expression.
             if (expression.operator().isIN() || expression.isUserDefined())
             {
+                if (!rowFilter.isStrict())
+                    throw new InvalidRequestException(String.format(UNSUPPORTED_NON_STRICT_OPERATOR, expression.operator()));
```
adelapena, line 89: I think this error could be thrown by `StorageAttachedIndex#validate(ReadCommand)`.

src/java/org/apache/cassandra/service/reads/DataResolver.java
```diff
@@ -149,6 +149,11 @@ private class ResolveContext
         private final DataLimits.Counter mergedResultCounter;
 
         private ResolveContext(E replicas)
+        {
+            this(replicas, true);
+        }
+
+        private ResolveContext(E replicas, boolean enforceLimits)
```
adelapena, line 155: Maybe its better to have a single constructor and be explicit about `enforceLimits` (there are only three usages).

src/java/org/apache/cassandra/service/reads/DataResolver.java
```diff
@@ -311,6 +312,27 @@ private PartitionIterator resolveInternal(ResolveContext context,
         return Transformation.apply(counted, new EmptyPartitionsDiscarder());
     }
 
+    private PartitionIterator resolveWithoutCounter(ResolveContext firstPhaseContext,
```
adelapena: It seems there is a lot of code duplication between `resolveInternal` and `resolveWithoutCounter`. We should probably extract the common part and maybe rename `resolveInternal` to `resolveWithCounter` or something like that. 

Alternatively, we can simply pass a null counter to the old `resolveInternal`:
```java
PartitionIterator counted = preCountFilter == null 
                            ? filtered 
                            : Transformation.apply(preCountFilter.apply(filtered), context.mergedResultCounter);
```

test/unit/org/apache/cassandra/index/sai/cql/intersection/RandomIntersectionTester.java
```diff
@@ -80,7 +80,8 @@ protected void runRestrictedQueries() throws Throwable
         Map<Integer, List<TestRow>> testRowMap = buildAndLoadTestRows();
 
         beforeAndAfterFlush(() -> {
-            for (int queryCount = 0; queryCount < nextInt(10, 100); queryCount++)
+            int queryCount = nextInt(10, 80);
```
adelapena, line 83: Why 10-80 instead of the previous 10-100?

src/java/org/apache/cassandra/db/filter/RowFilter.java
maedhroz: I think that what we'll see in [CASSANDRA-19007](https://issues.apache.org/jira/browse/CASSANDRA-19007) is this method becoming parameterized to handle the situation where we're hitting it from `ReadCommand#executeLocally()` (for normal, non-indexed AF queries). I can add a brief message here mentioning that 19007 will fix it, but that for now it ignores `isStrict()`?

src/java/org/apache/cassandra/index/sai/disk/v1/MemtableIndexWriter.java
maedhroz: Yeah, we should make that explicit in the `PCIW` method doc.

src/java/org/apache/cassandra/index/sai/plan/StorageAttachedIndexQueryPlan.java
maedhroz: I went back and forth on this a bit. It certainly could be thrown there, but here we have both the scaffolding for iterating over the expressions and the logic for what happens when all is well and the filter is in strict mode. It seemed easier to be able to see those things in one place. WDYT?

test/unit/org/apache/cassandra/index/sai/cql/intersection/RandomIntersectionTester.java
maedhroz: So I moved this out of the loop condition because it was re-evaluating on every iteration rather than being set before the loop, as I'm sure we originally intended. I think `10-80` ended up producing a distribution pretty similar to what was there before. (i.e. We very rarely ever got to the upper end of the distribution with re-evaluating on every iteration.)

src/java/org/apache/cassandra/index/sai/plan/StorageAttachedIndexQueryPlan.java
adelapena: Yes, makes sense. Since we build the query plan when creating the command, `Index#validate(ReadCommand)` has lost part of its sense.

src/java/org/apache/cassandra/service/reads/DataResolver.java
```diff
@@ -306,7 +310,10 @@ private PartitionIterator resolveInternal(ResolveContext context,
         UnfilteredPartitionIterator merged = UnfilteredPartitionIterators.merge(results, mergeListener);
         Filter filter = new Filter(command.nowInSec(), command.metadata().enforceStrictLiveness());
         FilteredPartitions filtered = FilteredPartitions.filter(merged, filter);
-        PartitionIterator counted = Transformation.apply(preCountFilter.apply(filtered), context.mergedResultCounter);
+
+        PartitionIterator counted = preCountFilter == null
```
adelapena, line 314: Nit: can we mark the `preCountFilter` parameter as `@Nullable`?

src/java/org/apache/cassandra/service/reads/ReplicaFilteringProtection.java
```diff
@@ -197,6 +198,9 @@ public UnfilteredRowIterators.MergeListener getRowMergeListener(DecoratedKey par
                 for (int i = 0; i < sources.size(); i++)
                     builders.add(i, new PartitionBuilder(partitionKey, sources.get(i), columns, stats));
 
+                boolean[] silentRowAt = new boolean[builders.size()];
+                boolean[] silentColumnAt = new boolean[builders.size()];
```
adelapena, line 200: The class JavaDoc talks about silent rows (`the rows in a replica response that don't have a corresponding row in other replica responses`). We should probably update that also to mention the new silent columns.

maedhroz: Agree. Will add that.

maedhroz:
Committed as https://github.com/apache/cassandra/commit/f7984627e7dd5794dabc099f49ecf80ca3631803

PR #2936, opened 2023-11-28
CASSANDRA-19110 convert deprecated apt-key command to new method on deb systems

zyxep:
Hi,

I've updated the 2 shell scripts that generates the output for Debian based systems.
since `apt-key` is deprecated and getting removed soon then I suggest on moving to the "new" way with `gpg --dearmor` method and pointing the apt source to the new file.

I've pointed the source of the gpg key to `/etc/apt/keyrings` because `/usr/share/keyrings/` is also getting converted.
On Ubuntu 22.04 a default folder `/etc/apt/keyrings` is created but not on `20.04` hence my comment in that shell script.

There are no output of `gpg --dearmor` so I have not updated [add_repo_keys.result](https://github.com/apache/cassandra/blob/trunk/doc/modules/cassandra/examples/RESULTS/add_repo_keys.result)

I've also updated the `deb` line in [get_deb_packages.sh](https://github.com/apache/cassandra/blob/trunk/doc/modules/cassandra/examples/BASH/get_deb_package.sh) to use signed gpg key since `apt` is starting to require keys like that.

### DIFF
--- a/doc/modules/cassandra/examples/BASH/add_repo_keys.sh
+++ b/doc/modules/cassandra/examples/BASH/add_repo_keys.sh
@@ -1,1 +1,2 @@
-$ curl https://downloads.apache.org/cassandra/KEYS | sudo apt-key add -
+# Make sure /etc/apt/keyrings exists on your system.
+$ curl https://downloads.apache.org/cassandra/KEYS | sudo gpg --dearmor -o /etc/apt/keyrings/cassandra-archive-keyring.gpg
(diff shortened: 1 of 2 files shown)


smiklosovic:
Hi @zyxep , I am just checking if this has a corresponding JIRA ticket?

smiklosovic:
Aha I see, this is just about the documentation ...

smiklosovic:
@driftx does this make sense to you? I think you are more fit to review this.

driftx:
This is https://issues.apache.org/jira/browse/CASSANDRA-19110

zyxep:
Ya, only documentation :)
I tried to find `_/download` docs but wasn't able to find it, maybe it just uses the same shell scripts to compile it's data from.

PR #2943, opened 2023-11-29
CASSANDRA-19011 Fix SAI intersection queries

mike-tr-adamson:
- Fix comparison in PostingListRangeIterator for updating skip token
 - Fix binary search in KeyLookup.clusteredSeekToKey
 - Added new on-disk component for storing partition sizes by partition ID


test/unit/org/apache/cassandra/index/sai/cql/RandomItersectionTest.java
```diff
@@ -0,0 +1,209 @@
...
+        }
+    }
+
+    private Map<Integer, List<TestRow>> buildAndLoadTestRows(boolean largeClusters)
+    {
+        Map<Integer, List<TestRow>> testRowMap = new HashMap<>();
+
+        int clusterSize = largeClusters ? nextInt(500, 5000) : nextInt(10, 100);
+        int partition = nextInt(0, numRows);
+        while (testRowMap.containsKey(partition))
```
maedhroz: nit: This is always false, right?

test/unit/org/apache/cassandra/index/sai/cql/RandomItersectionTest.java
```diff
@@ -0,0 +1,209 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.stream.Collectors;
+
+import org.junit.Before;
+import org.junit.Test;
+
+import org.apache.cassandra.index.sai.utils.SAIRandomizedTester;
+
+public class RandomItersectionTest extends SAIRandomizedTester
```
maedhroz:
```suggestion
public class RandomIntersectionTest extends SAIRandomizedTester
```
Although I guess it's kind of a fun combination of iterator and intersection :p

(I have added value to this patch...my work is done...)

test/unit/org/apache/cassandra/index/sai/cql/RandomItersectionTest.java
```diff
@@ -0,0 +1,209 @@
...
+    {
+        createTable("CREATE TABLE %s (pk int, ck int, v1 int, v2 int, PRIMARY KEY(pk, ck))");
+        createIndex("CREATE INDEX ON %s(v1) USING 'sai'");
+        createIndex("CREATE INDEX ON %s(v2) USING 'sai'");
+
+        numRows = nextInt(50000, 200000);
+    }
+
+    @Test
+    public void smallPartitionPartitionRestrictedTest()
```
maedhroz: I guess we could merge `smallPartitionPartitionRestrictedTest`/`largePartitionPartitionRestrictedTest` together and `smallPartitionUnrestrictedTest`/`largePartitionUnrestrictedTest` and just parameterize on partition size.

test/unit/org/apache/cassandra/index/sai/cql/RandomItersectionTest.java
```diff
@@ -0,0 +1,209 @@
...
+    {
+        Map<Integer, List<TestRow>> testRowMap = buildAndLoadTestRows(false);
+
+        for (int queryCount = 0; queryCount < nextInt(10, 100); queryCount++)
+        {
+            int pk = testRowMap.keySet().stream().skip(nextInt(0, testRowMap.size())).findFirst().get();
+            int v1 = nextInt(10, numRows/10);
```
maedhroz: I wonder if we could wrap these up in some `nextV1()` / `nextV2()` methods or something. If we want to tweak the range, it would be easier to do it in one place I guess. I ran through all this a few times, and it does look like we get a reasonable spread of result set sizes, and every once in a while, no matches.

test/unit/org/apache/cassandra/index/sai/cql/RandomItersectionTest.java
```diff
@@ -0,0 +1,209 @@
...
+            int v2 = nextInt(10, numRows/50);
+
+            List<Object[]> expected = testRowMap.get(pk)
+                                                .stream()
+                                                .sorted(Comparator.comparingInt(o -> o.ck))
+                                                .filter(row -> row.v1 > v1 && row.v2 > v2)
+                                                .map(row -> row(row.ck))
+                                                .collect(Collectors.toList());
+
+            assertRowsIgnoringOrder(execute("SELECT ck FROM %s WHERE pk = ? AND v1 > ? AND v2 > ?", pk, v1, v2), expected.toArray(new Object[][]{}));
```
maedhroz: For fun, I changed all these to `assertRows()` and nothing broke. That makes sense for the single-partition cases, but I'm not sure for the non-restricted cases...

(For the single-partition cases, we might as well validate the correct ordering.)

test/unit/org/apache/cassandra/index/sai/cql/RandomItersectionTest.java
maedhroz: This is really a great, compact (after a couple cleanups), powerful test. +100

src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
```diff
@@ -73,7 +73,12 @@ public enum IndexComponent
     TOKEN_VALUES("TokenValues"),
 
     /**
-     * An on-disk block packed index containing the starting and ending rowIds for each partition.
+     * An on-disk block packed index mapping rowIds to partitionIds.
+     */
+    PARTITION_ROWS("PartitionRows"),
+
+    /**
+     * An on-disk block packed index containing the partition size for each partition keyed by partitionId.
```
maedhroz: I would take some care to be clear everywhere in the new code that partition size is the number of rows.

src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
```diff
@@ -73,7 +73,12 @@ public enum IndexComponent
     TOKEN_VALUES("TokenValues"),
 
     /**
-     * An on-disk block packed index containing the starting and ending rowIds for each partition.
+     * An on-disk block packed index mapping rowIds to partitionIds.
+     */
+    PARTITION_ROWS("PartitionRows"),
```
maedhroz: Partly because I'm worried about this being read "rows per partition", I wonder of something like `ROW_TO_PARTITION` would make sense here.

src/java/org/apache/cassandra/index/sai/disk/v1/WidePrimaryKeyMap.java
```diff
@@ -166,10 +177,7 @@ private Clustering<?> readClusteringKey(long sstableRowId)
     // Returns the rowId of the next partition or the number of rows if supplied rowId is in the last partition
     private long startOfNextPartition(long rowId)
     {
-        long partitionId = partitionArray.get(rowId);
-        long nextPartitionRowId = partitionArray.indexOf(++partitionId);
-        if (nextPartitionRowId == -1)
-            nextPartitionRowId = partitionArray.length();
-        return nextPartitionRowId;
+        long partitionSize = partitionSizeArray.get(partitionArray.get(rowId));
```
maedhroz: nit: Would consider being clearer here that this is a number of rows for people skimming.

This is nice though. The idea is to avoid the binary search and just lookup the size directly?

src/java/org/apache/cassandra/index/sai/disk/v1/postings/PostingListRangeIterator.java
```diff
@@ -67,7 +67,7 @@ public class PostingListRangeIterator extends KeyRangeIterator
     private final long rowIdOffset;
 
     private boolean needsSkipping = false;
-    private PrimaryKey skipToToken = null;
+    private PrimaryKey skipToKey = null;
```
maedhroz, line 70: Yeah, naming holdover from a past where this was a token?

src/java/org/apache/cassandra/index/sai/disk/v1/SSTableComponentsWriter.java
```diff
@@ -42,18 +42,21 @@ public class SSTableComponentsWriter implements PerSSTableIndexWriter
     private final IndexDescriptor indexDescriptor;
     private final MetadataWriter metadataWriter;
     private final NumericValuesWriter partitionSizeWriter;
+    private final NumericValuesWriter partitionRowsWriter;
     private final NumericValuesWriter tokenWriter;
     private final KeyStoreWriter partitionKeysWriter;
     private final KeyStoreWriter clusteringKeysWriter;
 
     private long partitionId = -1;
+    private long partitionSizeCount = 0;
```
maedhroz:
```suggestion
    private long partitionRowCount = 0;
```

src/java/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookup.java
```diff
@@ -265,42 +258,39 @@ public void reset() throws IOException
...
         {
-            return pointId >>> blockShift == blockOffsets.length() - 1;
+            currentBlockIndex = pointId >>> blockShift;
         }
 
         // Reset currentPointId and currentKey to be at the start of the block
         // pointed to by currentBlockIndex.
         private void resetToCurrentBlock()
         {
+
```
maedhroz, line 300: nit: revert newline

src/java/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookup.java
```diff
@@ -265,42 +258,39 @@ public void reset() throws IOException
...
+            resetToCurrentBlock();
+            if (compareKeys(key, currentKey) < 0)
+                return -1;
 
-        private boolean notInCurrentBlock(long pointId, BytesRef key)
-        {
-            if (inLastBlock(pointId))
-                return false;
+            if (currentBlockIndex == blockOffsets.length() -1)
+                return 0;
```
maedhroz, line 284: So to check my understanding, so far we've returned -1 if we're less than the first key in the current block (after updating to the block for the point ID) and then if we're at the last block, we assume the key exists there and return zero? (Haven't looked at the rest of the code yet to figure out if we guarantee the key exists somewhere...)

src/java/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookup.java
```diff
@@ -265,42 +258,39 @@ public void reset() throws IOException
...
-            long blockIndex = (pointId >>> blockShift) + 1;
             long currentFp = keysInput.getFilePointer();
-            keysInput.seek(blockOffsets.get(blockIndex) + keysFilePointer);
-            readKey(blockIndex << blockShift, nextBlockKey);
+            keysInput.seek(blockOffsets.get(currentBlockIndex + 1) + keysFilePointer);
+            readKey((currentBlockIndex + 1) << blockShift, nextBlockKey);
             keysInput.seek(currentFp);
-
-            return compareKeys(key, nextBlockKey) >= 0;
+            return compareKeys(key, nextBlockKey) < 0 ? 0 : 1;
```
maedhroz, line 289: ...and as for the rest of the method, we just look at the first key of the next block, returning 0 if the search key is less (so we're in the current block) and 1 otherwise (as we're in some later block).

src/java/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookup.java
```diff
@@ -197,43 +197,34 @@ public long clusteredSeekToKey(ByteComparable key, long startingPointId, long en
             updateCurrentBlockIndex(startingPointId);
             resetToCurrentBlock();
 
-            if (compareKeys(currentKey, skipKey) == 0)
-                return startingPointId;
+            // We can return straightaway if the currentPointId is within the requested partition range and the keys match
+            if (currentPointId >= startingPointId && currentPointId < endingPointId && compareKeys(currentKey, skipKey) == 0)
+                return currentPointId;
```
maedhroz, line 202: So it is possible here that after moving to `startingPointId`'s block, we might have a target/skip key that is the first key in the block, but we've restricted the point ID range to start after that key's point ID?

src/java/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookup.java
```diff
@@ -197,43 +197,34 @@ public long clusteredSeekToKey(ByteComparable key, long startingPointId, long en
...
-                    {
-                        updateCurrentBlockIndex((endingPointId - 1));
-                        resetToCurrentBlock();
-                    }
-
-                    int cmp = compareKeys(currentKey, skipKey);
-
-                    if (cmp == 0)
-                        return currentPointId;
+            while ((splitEndId - splitStartId) >>> blockShift > 0)
```
maedhroz: So this is basically saying the split end is at least the number of points-per-block away, right? (Otherwise we're already at the right block?) Might be worth a quick one-liner comment so casual reader doesn't stumble over the bit shifting, etc.

src/java/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookup.java
```diff
@@ -197,43 +197,34 @@ public long clusteredSeekToKey(ByteComparable key, long startingPointId, long en
...
+            {
+                long midPointId = splitStartId + (splitEndId - splitStartId) / 2;
 
-                    if (cmp < 0)
-                        splitPointId = currentPointId;
+                int position = keyInBlockComparison(midPointId, skipKey);
```
maedhroz: So essentially this `keyInBlockComparison()` method is `moveToBlockAndCompareTo()`? Going back and forth on whether it would be less surprising to just pull up the `updateCurrentBlockIndex()` and `resetToCurrentBlock()` calls to this level to avoid hiding the "move" part in `keyInBlockComparison()`...

test/unit/org/apache/cassandra/index/sai/cql/RandomItersectionTest.java
mike-tr-adamson: Good point, that was duplication from inside the loop.

test/unit/org/apache/cassandra/index/sai/cql/RandomItersectionTest.java
mike-tr-adamson: Dyslexia is a wonderful thing at times. I usually rely on Intellij and the like to pick up these sort of things but it ignores names like this :)

test/unit/org/apache/cassandra/index/sai/cql/RandomItersectionTest.java
mike-tr-adamson: I pretty much borrowed this from another test that wasn't partition restricted. I agree that within a partition they should be ordered.

test/unit/org/apache/cassandra/index/sai/cql/RandomItersectionTest.java
mike-tr-adamson: I've gone a bit further and parameterized the whole test such that we can control partition size, partition restriction and value cardinality. It makes it a bit longer to run but looks to complete in most cases in just over 1 minute.

src/java/org/apache/cassandra/index/sai/disk/v1/postings/PostingListRangeIterator.java
mike-tr-adamson: I thought I'd just clean that up while I was there. It was confusing.

src/java/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookup.java
mike-tr-adamson: I have added a comment to this line to try to clarify it. This is really a deferal of decision making to later because we don't have any further blocks to look at if we are in the last block.

src/java/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookup.java
mike-tr-adamson: Correct. It's just an optimisation on the off chance that this is the case.

src/java/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookup.java
mike-tr-adamson: I like `moveToBlockAndCompareTo`, so have changed it to that because it is closer to the behaviour. I wanted the update and reset block in this method because it shouldn't really be used without it. The name makes the binary search clearer as well.

src/java/org/apache/cassandra/index/sai/disk/v1/keystore/KeyLookup.java
mike-tr-adamson: Correct, I have added some comments to this method to make it clear what is going on. I've also made some small optimisations to avoid unnecessary seeks.

src/java/org/apache/cassandra/index/sai/disk/format/IndexComponent.java
mike-tr-adamson: This was on my list to look at the naming and associated comments of these on-disk components. I have gone with renaming all of them to make it clearer exactly what they contain. I have also made sure the users of these (in the primary key maps) are correctly named and commented.

I have to admit I was getting confused by the naming and what they were meant to be doing.

src/java/org/apache/cassandra/index/sai/disk/v1/WidePrimaryKeyMap.java
mike-tr-adamson: As above, I have changed all the variable names in these classes to make it clearer who is doing what.

As an aside, I found that locally `MonotonicBlockPackedReader.indexOf` is marked as unsupported. I think we missed that when the `indexOf` was put back in. This was causing the final issue we were having, which is why this component was added. I have also now marked that method as unsupported.

mike-tr-adamson:
@maedhroz @michaelsembwever I have pushed a commit with the review changes and am kicking off new CI runs for 5.0 and trunk

test/unit/org/apache/cassandra/index/sai/cql/RandomIntersectionTest.java
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    private int nextV1()
+    {
+        return v1Cardinality ? nextInt(10, numRows/10) : nextInt(10, numRows/1000);
+    }
+
+    private int nextV2()
+    {
+        return v2Cardinality ? nextInt(10, numRows/10) : nextInt(10, numRows/1000);
```
maedhroz, line 185: Should this use `nextInt(10, numRows/50)`, or is it different from `runRestrictedQueries()` (which I'm guessing explicitly uses lower cardinality and not these methods)?

maedhroz (approved):
One little nit on the generation of v2 values in the randomized test, but otherwise, looks good. Thanks!

src/java/org/apache/cassandra/index/sai/disk/v1/WidePrimaryKeyMap.java
```diff
@@ -58,19 +61,24 @@ public static class Factory extends SkinnyPrimaryKeyMap.Factory
...
+            super(indexDescriptor);
 
             this.clusteringKeyBlockOffsetsFile = indexDescriptor.createPerSSTableFileHandle(IndexComponent.CLUSTERING_KEY_BLOCK_OFFSETS, this::close);
             this.clustingingKeyBlocksFile = indexDescriptor.createPerSSTableFileHandle(IndexComponent.CLUSTERING_KEY_BLOCKS, this::close);
+            this.partitionToSizeFile = indexDescriptor.createPerSSTableFileHandle(IndexComponent.PARTITION_TO_SIZE, this::close);
 
             try
             {
                 this.clusteringComparator = indexDescriptor.clusteringComparator;
+                NumericValuesMeta partitionSizeMeta = new NumericValuesMeta(metadataSource.get(indexDescriptor.componentName(IndexComponent.PARTITION_TO_SIZE)));
```
michaelsembwever, line 80: what about when reading index from 5.0-alphaX and this doesn't exist ?

michaelsembwever (approved):

michaelsembwever (approved):

src/java/org/apache/cassandra/index/sai/disk/v1/WidePrimaryKeyMap.java
michaelsembwever: this is not a blocker.  we don't require compat with alphas.  (and there's no data loss, just rebuild/recompact.)

but it would be nice to clarify.

test/unit/org/apache/cassandra/index/sai/cql/RandomIntersectionTest.java
mike-tr-adamson: Sorry, missed that. It should be using `nextV1` and `nextV2` in `runRetrictedQueries`

src/java/org/apache/cassandra/index/sai/disk/v1/WidePrimaryKeyMap.java
mike-tr-adamson: We would rebuild the index. We have introduced a new component as well, so it is always going to need a rebuild.

michaelsembwever:
Committed https://github.com/apache/cassandra/commit/6a7bef12ecdf59e3a67c81b89c13e3c2bf7e19d8

PR #2946, opened 2023-11-29
Test Failure: org.apache.cassandra.distributed.upgrade.MixedModeFrom3…

ifesdjeen:
…LoggedBatchTest.testSimpleStrategy

The problem was mainly the fact that read command was deserialising serialisedAt epoch as null by default and not as Empty, causing NPE.

But a larger problem is that we are introducing Epoch in 5.0, which is not the case anymore

Patch for CASSANDRA-19066; reviewed by TBD


src/java/org/apache/cassandra/db/commitlog/CommitLogDescriptor.java
```diff
@@ -63,13 +63,14 @@ public class CommitLogDescriptor
...
     public static final int VERSION_50 = 8;
+    public static final int VERSION_51 = 9;
 
     /**
      * Increment this number if there is a changes in the commit log disc layout or MessagingVersion changes.
      * Note: make sure to handle {@link #getMessagingVersion()}
      */
     @VisibleForTesting
-    public static final int current_version = DatabaseDescriptor.getStorageCompatibilityMode().isBefore(5) ? VERSION_40 : VERSION_50;
+    public static final int current_version = DatabaseDescriptor.getStorageCompatibilityMode().isBefore(5) ? VERSION_40 : VERSION_51;
```
beobal, line 73: I think that we do need a ` CASSANDRA_5(5)` compatibility mode otherwise mutations in the commitlog  and hints will contain epochs so downgrading to 5.0 will not work unless the 5.1 instance uses SCM `CASSANDRA_4`. It may not be possible to do this if it previously wrote 5.0 format sstables, so a downgrade to 5.0 would only work if the cluster had _always_ been in SCM 4.

Of course, we could punt downgradability & a new SCM mode to a later ticket. This versioning bump is needed now though to prevent 5.1 nodes including epochs in messages to 5.0 nodes.

src/java/org/apache/cassandra/hints/HintsDescriptor.java
```diff
@@ -69,7 +70,7 @@ final class HintsDescriptor
     static final int VERSION_30 = 1;
     static final int VERSION_40 = 2;
     static final int VERSION_50 = 3;
-    static final int CURRENT_VERSION = DatabaseDescriptor.getStorageCompatibilityMode().isBefore(5) ? VERSION_40 : VERSION_50;
+    static final int CURRENT_VERSION = DatabaseDescriptor.getStorageCompatibilityMode().isBefore(5) ? VERSION_40 : VERSION_51;
```
beobal, line 73: Can we add a `HintsDescriptor::VERSION_51` here. There doesn't seem any real point in having these equivalent constants at the moment, but it might be confusing if it isn't consistent.

beobal (changes requested):
Only the comment about HintsDescriptor needs addressing, the downgradability question we can probably leave to a separate JIRA

src/java/org/apache/cassandra/hints/HintsDescriptor.java
ifesdjeen: Oh you are right; we are in fact using a `VERSION_51` constant here, but it refers to MessagingService. Fixed in a follow-up.

src/java/org/apache/cassandra/db/commitlog/CommitLogDescriptor.java
ifesdjeen: I would ask to postpone this to a separate ticket, as we're focusing on fixing tests right now. Would you be OK with that?

beobal: absolutely!

PR #2948, opened 2023-11-29
Fix additional bugs from live migration/interop merge

aweisberg:


src/java/org/apache/cassandra/dht/Range.java
```diff
@@ -942,6 +946,24 @@ public static <T extends RingPosition<T>> boolean equals(Collection<Range<T>> a,
...
+        TokenRange[] tokenRanges = new TokenRange[normalizedRanges.size()];
+        for (int i = 0; i < normalizedRanges.size(); i++)
+        {
+            Range<Token> tokenRange = normalizedRanges.get(i);
+            AccordRoutingKey left = new TokenKey(ks, tokenRange.left);
+            AccordRoutingKey right = new TokenKey(ks, tokenRange.right);
+            if (tokenRange.left.isMinimum())
+                left = AccordRoutingKey.SentinelKey.min(ks);
+            if (tokenRange.right.isMinimum())
+                right = AccordRoutingKey.SentinelKey.max(ks);
```
dcapwell: @bdeggleston mind sanity checking this?  I believe its correct, but just wanted to double check.

bdeggleston: AccordTopologyUtils already has range conversion code, we should just use that

src/java/org/apache/cassandra/repair/AccordRepairJob.java
```diff
@@ -91,30 +94,35 @@ protected void runRepair()
...
 
         TokenRange lastRepaired = null;
         int iteration = 0;
         while (true)
         {
+            if (shouldAbort != null)
+                throw shouldAbort;
             iteration++;
-            if (iteration % 100 == 0)
+            if (iteration % ACCORD_REPAIR_RANGE_STEP_UPDATE_INTERVAL.getInt() == 0)
```
dcapwell: can you cache `ACCORD_REPAIR_RANGE_STEP_UPDATE_INTERVAL` so we don't do this every time in the loop?

src/java/org/apache/cassandra/repair/AccordRepairJob.java
```diff
@@ -19,36 +19,41 @@
...
  * before the Epoch and point in time at which the repair started have their side effects visible to Paxos and regular quorum reads.
  */
 public class AccordRepairJob extends AbstractRepairJob
 {
+    private static final Logger logger = LoggerFactory.getLogger(AccordRepairJob.class);
+
     public static final BigInteger TWO = BigInteger.valueOf(2);
 
+    // Number of barriers to run before doubling the range step size
+    static int RANGE_STEP_UPDATE_INTERVAL = 100;
```
dcapwell: this is dead code now

dcapwell (approved):

aweisberg:
Merged as [128e704ec2500e37a0ee6847d151840976bcbb41](https://github.com/apache/cassandra/tree/128e704ec2500e37a0ee6847d151840976bcbb41)

PR #2957, opened 2023-12-01
CASSANDRA-19118 Add support of vector type to COPY command

szymon-miezal:
[CASSANDRA-19118](https://issues.apache.org/jira/browse/CASSANDRA-19118)

This patch adds a converter that allows parsing vector literals passed via csv files to the COPY command.

### DIFF
--- a/pylib/cqlshlib/copyutil.py
+++ b/pylib/cqlshlib/copyutil.py
@@ -46,7 +46,7 @@ from queue import Queue
 
 from cassandra import OperationTimedOut
 from cassandra.cluster import Cluster, DefaultConnection
-from cassandra.cqltypes import ReversedType, UserType, VarcharType
+from cassandra.cqltypes import ReversedType, UserType, VarcharType, VectorType
 from cassandra.metadata import protect_name, protect_names, protect_value
 from cassandra.policies import RetryPolicy, WhiteListRoundRobinPolicy, DCAwareRoundRobinPolicy, FallthroughRetryPolicy
 from cassandra.query import BatchStatement, BatchType, SimpleStatement, tuple_factory
@@ -2074,6 +2074,12 @@ class ImportConversion(object):
             return ImmutableDict(frozenset((convert_mandatory(ct.subtypes[0], v[0]), convert(ct.subtypes[1], v[1]))
                                  for v in [split(split_format_str % vv, sep=sep) for vv in split(val)]))
 
+        def convert_vector(val, ct=cql_type):
+            string_coordinates = split(val)
+            if len(string_coordinates) != ct.vector_size:
+                raise ParseError("The length of given vector value '%d' is not equal to the vector size from the type definition '%d'" % (len(string_coordinates), ct.vector_size))
+            return [convert_mandatory(ct.subtype, v) for v in string_coordinates]
+
         def convert_user_type(val, ct=cql_type):
             """
             A user type is a dictionary except that we must convert each key into
@@ -2130,6 +2136,7 @@ class ImportConversion(object):
             'map': convert_map,
             'tuple': convert_tuple,
             'frozen': convert_single_subtype,
+            VectorType.typename: convert_vector,
         }
 
         return converters.get(cql_type.typename, convert_unknown)
--- a/test/unit/org/apache/cassandra/tools/cqlsh/CqlshTest.java
+++ b/test/unit/org/apache/cassandra/tools/cqlsh/CqlshTest.java
@@ -18,16 +18,24 @@
 
 package org.apache.cassandra.tools.cqlsh;
 
+import java.io.IOException;
+import java.io.Writer;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
 import org.junit.BeforeClass;
 import org.junit.Test;
 
 import org.apache.cassandra.cql3.CQLTester;
+import org.apache.cassandra.cql3.UntypedResultSet;
 import org.apache.cassandra.tools.ToolRunner;
 import org.apache.cassandra.tools.ToolRunner.ToolResult;
-import org.hamcrest.CoreMatchers;
 
+import static java.lang.String.format;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
 
 public class CqlshTest extends CQLTester
 {
@@ -41,7 +49,119 @@ public class CqlshTest extends CQLTester
     public void testKeyspaceRequired()
     {
         ToolResult tool = ToolRunner.invokeCqlsh("SELECT * FROM test");
-        assertThat(tool.getCleanedStderr(), CoreMatchers.containsStringIgnoringCase("No keyspace has been specified"));
+        tool.asserts().errorContains("No keyspace has been specified");
         assertEquals(2, tool.getExitCode());
+    }
+
+    @Test
+    public void testCopyFloatVector() throws IOException
+    {
+        assertCopyOfVectorTypeSucceeds("float", 6, new Object[][] {
+            row(1, vector(0.1f, 0.2f, 0.3f, 0.4f, 0.5f, 0.6f)),
+            row(2, vector(-0.1f, -0.2f, -0.3f, -0.4f, -0.5f, -0.6f)),
+            row(3, vector(0.9f, 0.8f, 0.7f, 0.6f, 0.5f, 0.4f))
+        });
+
+        assertCopyOfVectorTypeSucceeds("float", 3, new Object[][] {
+            row(1, vector(0.1f, 0.2f, 0.3f)),
+            row(2, vector(-0.4f, -0.5f, -0.6f)),
+            row(3, vector(0.7f, 0.8f, 0.9f))
+        });
+    }
+
+    @Test
+    public void testCopyIntVector() throws IOException
+    {
+        assertCopyOfVectorTypeSucceeds("int", 6, new Object[][] {
+            row(1, vector(1, 2, 3, 4, 5, 6)),
+            row(2, vector(-1, -2, -3, -4, -5, -6)),
+            row(3, vector(9, 8, 7, 6, 5, 4))
+        });
+
+        assertCopyOfVectorTypeSucceeds("int", 3, new Object[][] {
+            row(1, vector(1, 2, 3)),
+            row(2, vector(-4, -5, -6)),
+            row(3, vector(7, 8, 9))
+        });
+    }
+
+    private void assertCopyOfVectorTypeSucceeds(String vectorType, int vectorSize, Object[][] rows) throws IOException
+    {
+        // given a table with a vector column
+        createTable(KEYSPACE, format("CREATE TABLE %%s (id int PRIMARY KEY, embedding_vector vector<%s, %d>)", vectorType, vectorSize));
+        assertTrue("table should be initially empty", execute("SELECT * FROM %s").isEmpty());
+
+        // write the rows into the table
+        for (Object[] row : rows)
+            execute("INSERT INTO %s (id, embedding_vector) VALUES (?, ?)", row);
+
+        // when running COPY TO CSV via cqlsh
+        Path csv = createTempFile("test_copy_to_vector");
+        ToolRunner.ToolResult copyToResult = ToolRunner.invokeCqlsh(format("COPY %s.%s TO '%s'", KEYSPACE, currentTable(), csv.toAbsolutePath()));
+
+        // then all rows should be exported
+        copyToResult.asserts().success();
+        // verify that the exported CSV contains the expected rows
+        assertThat(csv).hasSameTextualContentAs(prepareCSVFile(rows));
+
+        // truncate the table
+        execute("TRUNCATE %s");
+        assertTrue("table should be empty", execute("SELECT * FROM %s").isEmpty());
+
+        // when running COPY FROM via cqlsh
+        ToolRunner.ToolResult copyFromResult = ToolRunner.invokeCqlsh(format("COPY %s.%s FROM '%s'", KEYSPACE, currentTable(), csv.toAbsolutePath()));
+
+        // then all rows should be imported
+        copyFromResult.asserts().success();
+        UntypedResultSet importedRows = execute("SELECT * FROM %s");
+        assertRowsIgnoringOrder(importedRows, rows);
+    }
+
+    @Test
+    public void testCopyOnlyThoseRowsThatMatchVectorTypeSize() throws IOException
+    {
+        // given a table with a vector column and a file containing vector literals
+        createTable(KEYSPACE, "CREATE TABLE %s (id int PRIMARY KEY, embedding_vector vector<int, 6>)");
+        assertTrue("table should be initially empty", execute("SELECT * FROM %s").isEmpty());
+
+        Object[][] rows = {
+            row(1, vector(1, 2, 3, 4, 5, 6)),
+            row(2, vector(1, 2, 3, 4, 5)),
+            row(3, vector(1, 2, 3, 4, 6, 7))
+        };
+
+        Path csv = prepareCSVFile(rows);
+
+        // when running COPY via cqlsh
+        ToolRunner.ToolResult result = ToolRunner.invokeCqlsh(format("COPY %s.%s FROM '%s'", KEYSPACE, currentTable(), csv.toAbsolutePath()));
+
+        // then only rows that match type size should be imported
+        result.asserts().failure();
+        result.asserts().errorContains("The length of given vector value '5' is not equal to the vector size from the type definition '6'");
+        UntypedResultSet importedRows = execute("SELECT * FROM %s");
+        assertRowsIgnoringOrder(importedRows, row(1, vector(1, 2, 3, 4, 5, 6)),
+                                row(3, vector(1, 2, 3, 4, 6, 7)));
+    }
+
+    private static Path prepareCSVFile(Object[][] rows) throws IOException
+    {
+        Path csv = createTempFile("test_copy_from_vector");
+
+        try (Writer out = Files.newBufferedWriter(csv, StandardCharsets.UTF_8))
+        {
+            for (Object[] row : rows)
+            {
+                out.write(String.format("%s,\"%s\"\n", row[0], row[1]));
+            }
+        }
+
+        return csv;
+    }
+
+    private static Path createTempFile(String prefix) throws IOException
+    {
+        Path csv = Files.createTempFile(prefix, ".csv");
+        csv.toFile().deleteOnExit();
+        return csv;
     }
 }


pylib/cqlshlib/copyutil.py
```diff
@@ -2130,6 +2136,7 @@ def convert_unknown(val, ct=cql_type):
             'map': convert_map,
             'tuple': convert_tuple,
             'frozen': convert_single_subtype,
+            VectorType.typename: convert_vector,
```
Maxwell-Guo, line 2139: I would recommend use 'vector' to maintain a consistent coding style

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
```diff
@@ -554,4 +562,97 @@ public void udf() throws Throwable
...
+        });
+
+        assertCopyOfVectorLiteralsFromFileSucceeds("int", 3, new Object[][] {
+            row(1, vector(1, 2, 3)),
+            row(2, vector(-4, -5, -6)),
+            row(3, vector(7, 8, 9))
+        });
+    }
+
+    private void assertCopyOfVectorLiteralsFromFileSucceeds(String vectorType, int vectorSize, Object[][] rows) throws IOException
```
Maxwell-Guo: I would recommend to move this test cases to https://github.com/apache/cassandra/tree/trunk/test/unit/org/apache/cassandra/tools/cqlsh

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
```diff
@@ -554,4 +562,97 @@ public void udf() throws Throwable
...
+        Path csv = prepareCSVFile(rows);
+
+        // when running COPY via cqlsh
+        ToolRunner.ToolResult result = ToolRunner.invokeCqlsh(format("COPY %s.%s FROM '%s'", KEYSPACE, currentTable(), csv.toAbsolutePath()));
+        UntypedResultSet importedRows = execute("SELECT * FROM %s");
+
+        // then only rows that match type size should be imported
+        result.asserts().failure();
+        result.asserts().errorContains("The length of given vector value '5' is not equal to the vector size from the type definition '6'");
+        assertRowsIgnoringOrder(importedRows, row(1, vector(1, 2, 3, 4, 5, 6)),
```
Maxwell-Guo: row should keep aligned with importedRows

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
Maxwell-Guo: If we move this to cases to a single test file, then I think  we can move 
the code "  createTable(KEYSPACE, "CREATE TABLE %s (id int PRIMARY KEY, embedding_vector vector<int, 6>)");" to beforeClass ,no need to create a table per test cases.

smiklosovic:
@szymon-miezal I agree with all what @Maxwell-Guo wrote. Please address and ping me on the ticket.

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
szymon-miezal: I was thinking myself where to put this test, and in the end, I decided to place it among the already existing CQL vector tests as the vector aspect is the common denominator for them.
When I look at it now I think the direct usage of CQLSH is enough of a reason to put it in the `org.apache.cassandra.tools.cqlsh` package. It will also allow to remove `requireNetwork` before each test case.
We could place those tests in a separate file but I think moving the creation of a test table to a `beforeClass` section is not the right motivation to do so. Especially since my test cases use tables with different vector sizes and it's more convenient to create them within the test case - less confusing for the reader as well because you don't have to scroll up and down while reading the test.
Considering all of the above I will move them to https://github.com/apache/cassandra/blob/trunk/test/unit/org/apache/cassandra/tools/cqlsh/CqlshTest.java.

pylib/cqlshlib/copyutil.py
szymon-miezal: I understand the suggestion to use 'vector' for consistency in coding style.
However, I think there is a value in referencing a const that was defined somewhere else - in this case in the type definition - instead of repeating it. This approach makes the connection explicit and easier to track.

From a practical standpoint, it's highly unlikely that the name will change, as it could introduce a breaking change.

While I may not see significant value in maintaining this consistency in this particular instance, I won't strongly oppose it.

Maxwell-Guo: Your concerns are valuable, and I agree with them. However, because vector is a new type and the original type has been implemented in this way, I feel that keeping it consistent is a better choice.

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
Maxwell-Guo: See [VectorFctsTes](https://github.com/apache/cassandra/blob/trunk/test/unit/org/apache/cassandra/cql3/functions/VectorFctsTest.java#L56) , we can do as it has done, create a table with different vector type , so we can make only one table.

pylib/cqlshlib/copyutil.py
szymon-miezal: Alright, I will use a literal then.

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
szymon-miezal: Introducing a table with two types will needlessly complicate the importing code. I don't find it justifiable, especially considering the numerous existing tests that follow a one-table-per-test-case approach (as seen in, for example, [CQLVectorTest](https://github.com/apache/cassandra/blob/3ed09031312a6b32d01ed5a647a65995206bdb75/test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java)). Therefore, I don't believe it's necessary to strive to minimize it, especially that CQLTester already provides a nice framework that abstracts away the table cleanup logic.

Maxwell-Guo: CQLVectorTest only create table that
has vector type of a dimensions 2.

szymon-miezal: Yes, and the same table is created in multiple test cases of this class which supports the claim from my previous comment.

pylib/cqlshlib/copyutil.py
szymon-miezal: One remark though, the way `vector` type is parsed you won't receive the `vector` value at https://github.com/apache/cassandra/pull/2957/files#diff-584645e6e932edd7a17f03c79ae87b1f1f8ed54919a31ce8785af156b89c0b76R2139, instead it will be `org.apache.cassandra.db.marshal.VectorType`, it comes from the python driver - https://github.com/datastax/python-driver/blob/120277da36f880ac6a5508480144fe436bf0d8c1/cassandra/cqltypes.py#L1401. That's another reason to use `VectorType.typename` instead of a string literal.

Maxwell-Guo: So what about fix like https://github.com/datastax/python-driver/blob/120277da36f880ac6a5508480144fe436bf0d8c1/cassandra/cqltypes.py#L946  and https://github.com/apache/cassandra/blob/trunk/pylib/cqlshlib/copyutil.py#L2099, I still think left a  VectorType.typename in a map where keys are characters like 'blob'/'int'  are very strange, and the formats are different

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
Maxwell-Guo: I am ok with what you have said . 
One more , what about change the assertCopyOfVectorLiteralsFromFileSucceeds to validateCopyVectorLiteralsFromFile, not forced, just personal preference

smiklosovic: @szymon-miezal @Maxwell-Guo creating a table in each test case is probably just fine. If we moved it to BeforeClass, then it will be shared by all test methods and that is not isolated when we e.g. want to make some assertions on the content of the table, assert number of rows some select query selects etc etc ... Just move cqlsh stuff to that dedicated test.

Maxwell-Guo: yeah, I' m ok with that

pylib/cqlshlib/copyutil.py
smiklosovic: @Maxwell-Guo  you mean like using `convert_unknown` for vector and then convert vector in it? That would make sense, sure. But on the other hand, are not we abusing the usage of `convert_unknown` ? What is _unknown_ about _vector_? Nothing ... vector is perfectly known. So why would we use `convert_unknown` then ...

smiklosovic: @szymon-miezal I think that using literal for the sake of consistency is just fine. It is not like this is ever going to be renamed to something else. It will be `vector` pretty much for ever.

Another option is to use it like you did here completely everywhere for each entry in this map.

EDIT:

But on the other hand, as you showed, "vector" probably does not play together with this as it is FQCN in the driver which is quite unfortunate.

I dont think we have a luxury of fixing stuff in python driver just to use a literal here ... 

So, putting that to "convert_unknown" does not feel right as I described in the previous comment to Maxwell and if what I wrote here is true that it seems like what is there currently is the best option.

Maxwell-Guo: > @Maxwell-Guo you mean like using `convert_unknown` for vector and then convert vector in it? That would make sense, sure. But on the other hand, are not we abusing the usage of `convert_unknown` ? What is _unknown_ about _vector_? Nothing ... vector is perfectly known. So why would we use `convert_unknown` then ...

I agree with what you have sad and I also think literal should be used. After roughly looking at the relevant code, I found that the type prefixed with “org.apache.cassandra.db.marshal” like UserType, ReserveType are all going into the convert_unknown function, So I just give a suggestion without modifying the  Python driver (I think the best way is modify the driver). Then we can keep  the converters clean and get a right result. 

As for whether it is an unknown type, there is also a way to say it, such as ReserveType is putting into convert_unknown, and see [the debug msg print ](https://github.com/apache/cassandra/tree/trunk/pylib/cqlshlib#L2104) ,it seems that only those that enter convert_unknown and are not UserType and ReserveType are call Unknow type .

Aslo , I tested what I have said of the modification on this patch, it can succeed.

Of course, just some personal opinions.  @smiklosovic

test/unit/org/apache/cassandra/cql3/validation/operations/CQLVectorTest.java
Maxwell-Guo: I actually want to state here that my original intention was that there was no need to test different dimensions for me. A single dimension is enough like testCopyOnlyThoseRowsThatMatchVectorTypeSize has done, The different numbers may be endless. For example, why does match no longer test a 3, and just 6 is left ? And then I can create two tables at once in the beforeclass for testing，and the code will look simpler. That's why I say this, but if you can give more tests, why not? I am happy with that.

szymon-miezal: I did not spot that intention in any previous comments, to me it was from the beginning about moving the table creation to the `beforeClass` section which as stated by @smiklosovic as well could contribute to breaking the test isolation.
Obviously, it does not make sense to test to infinite vector size, I've chosen 2 as a reasonable tradeoff.

pylib/cqlshlib/copyutil.py
szymon-miezal: > Aslo , I tested what I have said of the modification on this patch, it can succeed.

Yes, both approaches will make the tests pass.
 
Note that `UserType` is a class designed for handling User-Defined Types (UDTs). In this context, the exact type is unknown to us, making it more suitable for `convert_unknown` rather than a well-defined and well-known vector type.

In the end, I think using `VectorType.typename` is not that much of an abuse, it points to a `str` value anyway.

Maxwell-Guo: @smiklosovic I have another suggestion. Please help me evaluate it. what about just keep VectorType.typename here, and we can also change other literal to their type's typename ?like :
`'blob' -> BytesType.typename`
`'decimal' -> DecimalType.typename`
I searched all the literals in the driver code , I think this may also be a good choice to ensure that coding style and format can be consistent.

smiklosovic: yeah that is also possible, that is similar to what I wrote in my comment above: _Another option is to use it like you did here completely everywhere for each entry in this map._

szymon-miezal: Is it a blocker for merging this patch?
If we decide to do it here then we will introduce divergence between 5.x and versions < 5.x.

smiklosovic: does that divergence cause any actual problems?

szymon-miezal: I think avoiding divergence when possible is a good habit that's why I am raising it.
Possibly a better way forward would be to merge it as is and change all the possible string keys to `typename` references in the subsequent patch. I would prefer that slightly.

smiklosovic: @adelapena what do you think about all of this? (I see you as a reviewer in the JIRA ticket.)

adelapena: I don't really understand the purpose of having `UserType` and `ReversedType` in the `convert_unknown` function. It seems that all the driver types have had a well known `type_name` attribute for a very long time. So we could use the `VectorType.typename: convert_vector` pattern for all data types:
```python
def convert_unknown(val, ct=cql_type):
    printdebugmsg("Unknown type %s (%s) for val %s" % (ct, ct.typename, val))
    return val

converters = {
    BytesType.type_name: convert_blob,
    DecimalType.type_name: get_convert_decimal_fcn(adapter=Decimal),
    UUIDType.type_name: convert_uuid,
    BooleanType.type_name: convert_bool,
```
I'm fine with doing that either here or in a separate patch.

test/unit/org/apache/cassandra/tools/cqlsh/CqlshTest.java
```diff
@@ -44,4 +53,95 @@ public void testKeyspaceRequired()
...
+        // when running COPY via cqlsh
+        ToolRunner.ToolResult result = ToolRunner.invokeCqlsh(format("COPY %s.%s FROM '%s'", KEYSPACE, currentTable(), csv.toAbsolutePath()));
+        UntypedResultSet importedRows = execute("SELECT * FROM %s");
+
+        // then all rows should be imported
+        result.asserts().success();
+        assertRowsIgnoringOrder(importedRows, rows);
+    }
+
+    private Path prepareCSVFile(Object[][] rows) throws IOException
```
adelapena: Nit: can be `static`. Also, I would probably put this method after `testCopyOnlyThoseRowsThatMatchVectorTypeSize`.

test/unit/org/apache/cassandra/tools/cqlsh/CqlshTest.java
```diff
@@ -44,4 +53,95 @@ public void testKeyspaceRequired()
...
+
+        Path csv = prepareCSVFile(rows);
+
+        // when running COPY via cqlsh
+        ToolRunner.ToolResult result = ToolRunner.invokeCqlsh(format("COPY %s.%s FROM '%s'", KEYSPACE, currentTable(), csv.toAbsolutePath()));
+        UntypedResultSet importedRows = execute("SELECT * FROM %s");
+
+        // then all rows should be imported
+        result.asserts().success();
+        assertRowsIgnoringOrder(importedRows, rows);
```
adelapena: Nit: we can check this immediately after invoking cqlsh, before running the `SELECT` query:
```suggestion
        // when running COPY via cqlsh
        result = ToolRunner.invokeCqlsh(format("COPY %s.%s FROM '%s'", KEYSPACE, currentTable(), csv.toAbsolutePath()));
        result.asserts().success();

        // then all rows should be imported
        UntypedResultSet importedRows = execute("SELECT * FROM %s");
        assertRowsIgnoringOrder(importedRows, rows);
```

test/unit/org/apache/cassandra/tools/cqlsh/CqlshTest.java
```diff
@@ -44,4 +53,95 @@ public void testKeyspaceRequired()
...
+        });
+    }
+
+    private void assertCopyOfVectorLiteralsFromFileSucceeds(String vectorType, int vectorSize, Object[][] rows) throws IOException
+    {
+        // given a table with a vector column and a file containing vector literals
+        createTable(KEYSPACE, format("CREATE TABLE %%s (id int PRIMARY KEY, embedding_vector vector<%s, %d>)", vectorType, vectorSize));
+        assertTrue("table should be initially empty", execute("SELECT * FROM %s").isEmpty());
+
+        Path csv = prepareCSVFile(rows);
```
adelapena: We are testing `COPY FROM` but not `COPY TO`. This could easily be achieved by inserting the rows and copying them to CSV, then verifying that the written file is equals to the expected CSV file synthetically generated by `prepareCSVFile`:
```suggestion
        assertTrue("table should be initially empty", execute("SELECT * FROM %s").isEmpty());
        
        // write the rows into the table
        for (Object[] row : rows)
            execute("INSERT INTO %s (id, embedding_vector) VALUES (?, ?)", row);

        // export the rows to CSV
        Path csv = Files.createTempFile("test_copy_to_vector", ".csv");
        csv.toFile().deleteOnExit();
        ToolRunner.ToolResult result = ToolRunner.invokeCqlsh(format("COPY %s.%s TO '%s'", KEYSPACE, currentTable(), csv.toAbsolutePath()));
        result.asserts().success();

        // verify that the exported CSV contains the expected rows
        Assertions.assertThat(csv).hasSameTextualContentAs(prepareCSVFile(rows));

        // truncate the table
        execute("TRUNCATE %s");
        assertTrue("table should be empty", execute("SELECT * FROM %s").isEmpty());
```

pylib/cqlshlib/copyutil.py
smiklosovic: _I don't really understand the purpose of having UserType and ReversedType in the convert_unknown function._

Yeah, me neither.

OK so we will fix it all in another patch. We go with what is there right that, that is how I understand it.

szymon-miezal: I will create a separate ticket for it and apply this improvement against older branches as well, that will be a valuable refactor.

Maxwell-Guo: > OK so we will fix it all in another patch. We go with what is there right that, that is how I understand it.

+1

test/unit/org/apache/cassandra/tools/cqlsh/CqlshTest.java
szymon-miezal: I have moved the query to the `then` section, it allows us to execute it after cqlsh status assertions and also keep the given/when/then sections clean and separated.

test/unit/org/apache/cassandra/tools/cqlsh/CqlshTest.java
szymon-miezal: That's a good suggestion I have modified the test to incorporate this suggestion, the benefit is that now the testing code has become more generic.

test/unit/org/apache/cassandra/tools/cqlsh/CqlshTest.java
szymon-miezal: Given other methods that use `prepareCSVFile` are not `static` and there is no real value in making it static I decided to keep it as is.

szymon-miezal:
@adelapena I think the PR is ready for a second round, once it's approved I will squash the history and attach patches to the jira ticket for the sake of running tests and merging them further.

test/unit/org/apache/cassandra/tools/cqlsh/CqlshTest.java
adelapena: We usually make methods that don't depend on the instance state `static`, but I wouldn't insist if you have decided to keep it as is.

adelapena:
Looks good to me, running CI for 5.0:
 * [j11](https://app.circleci.com/pipelines/github/adelapena/cassandra/3359/workflows/68cca061-02e3-419c-88f9-0062f42e1b3e)
 * [j17](https://app.circleci.com/pipelines/github/adelapena/cassandra/3359/workflows/f9b3c955-270a-47bc-bfd9-9750164e636d)

I'll run it for the trunk patch too when it's ready.

test/unit/org/apache/cassandra/tools/cqlsh/CqlshTest.java
szymon-miezal: I don't want to break that pattern then, I will update that.

szymon-miezal:
That failure is surprising: https://app.circleci.com/pipelines/github/adelapena/cassandra/3359/workflows/68cca061-02e3-419c-88f9-0062f42e1b3e/jobs/96153

I suspect it's because https://github.com/datastax/python-driver/tree/cassandra-test - a branch used by cqlshlib test - does not support vector type.

adelapena:
> I suspect it's because https://github.com/datastax/python-driver/tree/cassandra-test - a branch used by cqlshlib test - does not support vector type.

Yes, it seems we are going to need to update that branch before we proceed with this, probably in a separate ticket.

szymon-miezal:
We could work around it by using this hated literal instead of a reference to `VectorType` but running tests for https://github.com/apache/cassandra/tree/trunk/pylib directory with a not-compatible driver will hit us again eventually.

I doubt we will be able to afford the luxury of updating [this](https://github.com/datastax/python-driver/tree/cassandra-test) particular branch as that will break old versions that do not support python 3. More likely we will have to create yet another branch or use master.

szymon-miezal:
I have upgraded python-driver to [3.28.0](https://github.com/datastax/python-driver/releases/tag/3.28.0).
When running `cassandra-cqlsh-tests.sh` locally I can see the number of failures is the same on `trunk` and on my branch.

@adelapena I think it would be useful to rerun the CI for this patch.

PR #2961, opened 2023-12-04
CASSANDRA-18796 guardrails for queries without partition-key restrictions for indexes

smiklosovic:
 
```

```

### DIFF
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -2076,6 +2076,15 @@ drop_compact_storage_enabled: false
 # if zero_ttl_on_twcs_enabled is set to false, this property is irrelevant as such statements will fail.
 #zero_ttl_on_twcs_warned: true
 
+# Guardrail enabling secondary index queries that do not restrict on partition key (defaults to true)
+#non_partition_restricted_index_query_enabled: true
+# Maximum number of referenced SAI indexes on a replica when executing a SELECT query
+# before emitting a warning (defaults to 32)
+#sai_sstable_indexes_per_query_warn_threshold: 32
+# Maximum number of referenced SAI indexes on a replica when executing a SELECT query
+# before emitting a failure (defaults to -1 to disable)
+#sai_sstable_indexes_per_query_fail_threshold: -1
+
 # The default secondary index implementation when CREATE INDEX does not specify one via USING.
 # ex. "legacy_local_table" - (default) legacy secondary index, implemented as a hidden table
 # ex. "sai" - "storage-attched" index, implemented via optimized SSTable/Memtable-attached indexes
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -897,6 +897,9 @@ public class Config
     public volatile int maximum_replication_factor_fail_threshold = -1;
     public volatile boolean zero_ttl_on_twcs_warned = true;
     public volatile boolean zero_ttl_on_twcs_enabled = true;
+    public volatile boolean non_partition_restricted_index_query_enabled = true;
+    public volatile int sai_sstable_indexes_per_query_warn_threshold = 32;
+    public volatile int sai_sstable_indexes_per_query_fail_threshold = -1;
 
     public volatile DurationSpec.LongNanosecondsBound streaming_state_expires = new DurationSpec.LongNanosecondsBound("3d");
     public volatile DataStorageSpec.LongBytesBound streaming_state_size = new DataStorageSpec.LongBytesBound("40MiB");
--- a/src/java/org/apache/cassandra/config/GuardrailsOptions.java
+++ b/src/java/org/apache/cassandra/config/GuardrailsOptions.java
@@ -89,6 +89,10 @@ public class GuardrailsOptions implements GuardrailsConfig
         validateMaxRFThreshold(config.maximum_replication_factor_warn_threshold, config.maximum_replication_factor_fail_threshold);
         validateTimestampThreshold(config.maximum_timestamp_warn_threshold, config.maximum_timestamp_fail_threshold, "maximum_timestamp");
         validateTimestampThreshold(config.minimum_timestamp_warn_threshold, config.minimum_timestamp_fail_threshold, "minimum_timestamp");
+        validateMaxLongThreshold(config.sai_sstable_indexes_per_query_warn_threshold,
+                                 config.sai_sstable_indexes_per_query_fail_threshold,
+                                 "sai_sstable_indexes_per_query",
+                                 false);
     }
 
     @Override
@@ -898,6 +902,48 @@ public class GuardrailsOptions implements GuardrailsConfig
                                   fail,
                                   () -> config.minimum_timestamp_fail_threshold,
                                   x -> config.minimum_timestamp_fail_threshold = x);
+    }
+
+    @Override
+    public int getSaiSSTableIndexesPerQueryWarnThreshold()
+    {
+        return config.sai_sstable_indexes_per_query_warn_threshold;
+    }
+
+    @Override
+    public int getSaiSSTableIndexesPerQueryFailThreshold()
+    {
+        return config.sai_sstable_indexes_per_query_fail_threshold;
+    }
+
+    @Override
+    public void setSaiSSTableIndexesPerQueryThreshold(int warn, int fail)
+    {
+        validateMaxIntThreshold(warn, fail, "sai_sstable_indexes_per_query");
+        updatePropertyWithLogging("sai_sstable_indexes_per_query_warn_threshold",
+                                  warn,
+                                  () -> config.sai_sstable_indexes_per_query_warn_threshold,
+                                  x -> config.sai_sstable_indexes_per_query_warn_threshold = x);
+
+        updatePropertyWithLogging("sai_sstable_indexes_per_query_fail_threshold",
+                                  fail,
+                                  () -> config.sai_sstable_indexes_per_query_fail_threshold,
+                                  x -> config.sai_sstable_indexes_per_query_fail_threshold = x);
+    }
+
+    @Override
+    public boolean getNonPartitionRestrictedQueryEnabled()
+    {
+        return config.non_partition_restricted_index_query_enabled;
+    }
+
+    @Override
+    public void setNonPartitionRestrictedQueryEnabled(boolean enabled)
+    {
+        updatePropertyWithLogging("non_partition_restricted_index_query_enabled",
+                                  enabled,
+                                  () -> config.non_partition_restricted_index_query_enabled,
+                                  x -> config.non_partition_restricted_index_query_enabled = x);
     }
 
     private static <T> void updatePropertyWithLogging(String propertyName, T newValue, Supplier<T> getter, Consumer<T> setter)
--- a/src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
+++ b/src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
@@ -404,7 +404,12 @@ public class SelectStatement implements CQLStatement.SingleKeyspaceCqlStatement
         boolean isPartitionRangeQuery = restrictions.isKeyRange() || restrictions.usesSecondaryIndexing();
 
         if (isPartitionRangeQuery)
+        {
+            if (restrictions.isKeyRange() && restrictions.usesSecondaryIndexing() && !SchemaConstants.isLocalSystemKeyspace(table.keyspace))
+                Guardrails.nonPartitionRestrictedIndexQueryEnabled.ensureEnabled(state);
+
             return getRangeCommand(options, state, columnFilter, limit, nowInSec);
+        }
 
         return getSliceCommands(options, state, columnFilter, limit, nowInSec);
     }
--- a/src/java/org/apache/cassandra/db/guardrails/Guardrails.java
+++ b/src/java/org/apache/cassandra/db/guardrails/Guardrails.java
@@ -477,6 +477,21 @@ public final class Guardrails implements GuardrailsMBean
                      format("The modification to table %s has a timestamp %s before the minimum allowable %s threshold %s",
                             what, value, isWarning ? "warning" : "failure", threshold));
 
+    public static final MaxThreshold saiSSTableIndexesPerQuery =
+    new MaxThreshold("sai_sstable_indexes_per_query",
+                     "High number of referenced indexes per query may negatively effect performance",
+                     state -> CONFIG_PROVIDER.getOrCreate(state).getSaiSSTableIndexesPerQueryWarnThreshold(),
+                     state -> CONFIG_PROVIDER.getOrCreate(state).getSaiSSTableIndexesPerQueryFailThreshold(),
+                     ((isWarning, what, value, threshold) ->
+                      format("The number of SSTable indexes queried on index %s violated %s threshold value %s with value %s",
+                             what, isWarning ? "warning" : "failure", threshold, value)));
+
+    public static final EnableFlag nonPartitionRestrictedIndexQueryEnabled =
+    new EnableFlag("non_partition_restricted_index_query_enabled",
+                   "Executing a query on secondary indexes without partition key restriction might degrade performance",
+                   state -> CONFIG_PROVIDER.getOrCreate(state).getNonPartitionRestrictedQueryEnabled(),
+                   "Non-partition key restricted query");
+
     private Guardrails()
     {
         MBeanWrapper.instance.registerMBean(this, MBEAN_NAME);
@@ -1198,6 +1213,36 @@ public final class Guardrails implements GuardrailsMBean
     public void setMinimumTimestampThreshold(String warnSeconds, String failSeconds)
     {
         DEFAULT_CONFIG.setMinimumTimestampThreshold(durationFromString(warnSeconds), durationFromString(failSeconds));
+    }
+
+    @Override
+    public int getSaiSSTableIndexesPerQueryWarnThreshold()
+    {
+        return DEFAULT_CONFIG.getSaiSSTableIndexesPerQueryWarnThreshold();
+    }
+
+    @Override
+    public int getSaiSSTableIndexesPerQueryFailThreshold()
+    {
+        return DEFAULT_CONFIG.getSaiSSTableIndexesPerQueryFailThreshold();
+    }
+
+    @Override
+    public void setSaiSSTableIndexesPerQueryThreshold(int warn, int fail)
+    {
+        DEFAULT_CONFIG.setSaiSSTableIndexesPerQueryThreshold(warn, fail);
+    }
+
+    @Override
+    public boolean getNonPartitionRestrictedQueryEnabled()
+    {
+        return DEFAULT_CONFIG.getNonPartitionRestrictedQueryEnabled();
+    }
+
+    @Override
+    public void setNonPartitionRestrictedQueryEnabled(boolean enabled)
+    {
+        DEFAULT_CONFIG.setNonPartitionRestrictedQueryEnabled(enabled);
     }
 
     private static String toCSV(Set<String> values)
--- a/src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
+++ b/src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
@@ -429,4 +429,38 @@ public interface GuardrailsConfig
      */
     void setMinimumTimestampThreshold(@Nullable DurationSpec.LongMicrosecondsBound warn,
                                       @Nullable DurationSpec.LongMicrosecondsBound fail);
+
+    /**
+     * @return the warning threshold for the number of SAI SSTable indexes searched by a query
+     */
+    int getSaiSSTableIndexesPerQueryWarnThreshold();
+
+    /**
+     * @return the failure threshold for the number of SAI SSTable indexes searched by a query
+     */
+    int getSaiSSTableIndexesPerQueryFailThreshold();
+
+    /**
+     * Sets warning and failure thresholds for the number of SAI SSTable indexes searched by a query
+     *
+     * @param warn value to set for warn threshold
+     * @param fail value to set for fail threshold
+     */
+    void setSaiSSTableIndexesPerQueryThreshold(int warn, int fail);
+
+    /**
+     * Returns whether it is possible to execute a query against secondary indexes without specifying
+     * any partition key restrictions.
+     *
+     * @return true if it is possible to execute a query without a partition key, false otherwise
+     */
+    boolean getNonPartitionRestrictedQueryEnabled();
+
+    /**
+     * Sets whether it is possible to execute a query against indexes (secondary or SAI) without specifying
+     * any partition key restrictions.
+     *
+     * @param enabled {@code true} if a query without partition key is enabled or not
+     */
+    void setNonPartitionRestrictedQueryEnabled(boolean enabled);
 }
--- a/src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
+++ b/src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
@@ -770,4 +770,38 @@ public interface GuardrailsMBean
      *                     {@code 20m}, {@code 30h} or {@code 40d}. A {@code null} value means disabled.
      */
     void setMinimumTimestampThreshold(@Nullable String warnDuration, @Nullable String failDuration);
+
+    /**
+     * @return the warning threshold for the number of SAI SSTable indexes searched by a query
+     */
+    int getSaiSSTableIndexesPerQueryWarnThreshold();
+
+    /**
+     * @return the failure threshold for the number of SAI SSTable indexes searched by a query
+     */
+    int getSaiSSTableIndexesPerQueryFailThreshold();
+
+    /**
+     * Sets warning and failure thresholds for the number of SAI SSTable indexes searched by a query
+     *
+     * @param warn value to set for warn threshold
+     * @param fail value to set for fail threshold
+     */
+    void setSaiSSTableIndexesPerQueryThreshold(int warn, int fail);
+
+    /**
+     * Returns whether it is possible to execute a query against secondary indexes without specifying
+     * any partition key restrictions.
+     *
+     * @return true if it is possible to execute a query without a partition key, false otherwise
+     */
+    boolean getNonPartitionRestrictedQueryEnabled();
+
+    /**
+     * Sets whether it is possible to execute a query against indexes (secondary or SAI) without specifying
+     * any partition key restrictions.
+     *
+     * @param enabled {@code true} if a query without partition key is enabled or not
+     */
+    void setNonPartitionRestrictedQueryEnabled(boolean enabled);
 }
--- a/src/java/org/apache/cassandra/index/sai/plan/QueryController.java
+++ b/src/java/org/apache/cassandra/index/sai/plan/QueryController.java
@@ -32,6 +32,7 @@ import org.apache.cassandra.cql3.Operator;
 import org.apache.cassandra.db.ColumnFamilyStore;
 import org.apache.cassandra.db.DataRange;
 import org.apache.cassandra.db.DecoratedKey;
+import org.apache.cassandra.db.MessageParams;
 import org.apache.cassandra.db.PartitionPosition;
 import org.apache.cassandra.db.PartitionRangeReadCommand;
 import org.apache.cassandra.db.ReadCommand;
@@ -41,6 +42,7 @@ import org.apache.cassandra.db.filter.ClusteringIndexFilter;
 import org.apache.cassandra.db.filter.ClusteringIndexNamesFilter;
 import org.apache.cassandra.db.filter.DataLimits;
 import org.apache.cassandra.db.filter.RowFilter;
+import org.apache.cassandra.db.guardrails.Guardrails;
 import org.apache.cassandra.db.rows.UnfilteredRowIterator;
 import org.apache.cassandra.dht.AbstractBounds;
 import org.apache.cassandra.dht.Range;
@@ -56,7 +58,9 @@ import org.apache.cassandra.index.sai.iterators.KeyRangeOrderingIterator;
 import org.apache.cassandra.index.sai.iterators.KeyRangeUnionIterator;
 import org.apache.cassandra.index.sai.metrics.TableQueryMetrics;
 import org.apache.cassandra.index.sai.utils.PrimaryKey;
+import org.apache.cassandra.net.ParamType;
 import org.apache.cassandra.schema.TableMetadata;
+import org.apache.cassandra.tracing.Tracing;
 import org.apache.cassandra.utils.FBUtilities;
 import org.apache.cassandra.utils.Pair;
 import org.apache.cassandra.utils.Throwables;
@@ -186,6 +190,8 @@ public class QueryController
 
         try
         {
+            maybeTriggerGuardrails(queryView);
+
             for (Pair<Expression, Collection<SSTableIndex>> queryViewPair : queryView.view)
             {
                 KeyRangeIterator indexIterator = IndexSearchResultIterator.build(queryViewPair.left, queryViewPair.right, mergeRange, queryContext);
@@ -201,6 +207,27 @@ public class QueryController
             throw t;
         }
         return builder;
+    }
+
+    private void maybeTriggerGuardrails(QueryViewBuilder.QueryView queryView)
+    {
+        int referencedIndexes = queryView.referencedIndexes.size();
+
+        if (Guardrails.saiSSTableIndexesPerQuery.failsOn(referencedIndexes, null))
+        {
+            String msg = String.format("Query %s attempted to read from too many indexes (%s) but max allowed is %s; " +
+                                       "query aborted (see sai_sstable_indexes_per_query_fail_threshold)",
+                                       command.toCQLString(),
+                                       referencedIndexes,
+                                       Guardrails.CONFIG_PROVIDER.getOrCreate(null).getSaiSSTableIndexesPerQueryFailThreshold());
+            Tracing.trace(msg);
+            MessageParams.add(ParamType.TOO_MANY_REFERENCED_INDEXES_FAIL, referencedIndexes);
+            throw new QueryReferencingTooManyIndexesException(msg);
+        }
+        else if (Guardrails.saiSSTableIndexesPerQuery.warnsOn(referencedIndexes, null))
+        {
+            MessageParams.add(ParamType.TOO_MANY_REFERENCED_INDEXES_WARN, referencedIndexes);
+        }
     }
 
     /**
--- a/src/java/org/apache/cassandra/index/sai/plan/QueryReferencingTooManyIndexesException.java
+++ b/src/java/org/apache/cassandra/index/sai/plan/QueryReferencingTooManyIndexesException.java
@@ -16,24 +16,14 @@
  * limitations under the License.
  */
 
-package org.apache.cassandra.exceptions;
+package org.apache.cassandra.index.sai.plan;
 
-import java.util.Map;
+import org.apache.cassandra.db.RejectException;
 
-import org.apache.cassandra.db.ConsistencyLevel;
-import org.apache.cassandra.locator.InetAddressAndPort;
-
-public class TombstoneAbortException extends ReadAbortException
+public class QueryReferencingTooManyIndexesException extends RejectException
 {
-    public final int nodes;
-    public final long tombstones;
-
-    public TombstoneAbortException(String message, int nodes, long tombstones, boolean dataPresent,
-                                   ConsistencyLevel consistency, int received, int blockFor,
-                                   Map<InetAddressAndPort, RequestFailureReason> failureReasonByEndpoint)
+    public QueryReferencingTooManyIndexesException(String msg)
     {
-        super(message, consistency, received, blockFor, dataPresent, failureReasonByEndpoint);
-        this.nodes = nodes;
-        this.tombstones = tombstones;
+        super(msg);
     }
 }
--- a/src/java/org/apache/cassandra/metrics/KeyspaceMetrics.java
+++ b/src/java/org/apache/cassandra/metrics/KeyspaceMetrics.java
@@ -173,6 +173,9 @@ public class KeyspaceMetrics
     public final Meter rowIndexSizeAborts;
     public final Histogram rowIndexSize;
 
+    public final Meter tooManySSTableIndexesReadWarnings;
+    public final Meter tooManySSTableIndexesReadAborts;
+
     public final ImmutableMap<SSTableFormat<?, ?>, ImmutableMap<String, Gauge<? extends Number>>> formatSpecificGauges;
 
     public final MetricNameFactory factory;
@@ -274,6 +277,9 @@ public class KeyspaceMetrics
         rowIndexSizeWarnings = createKeyspaceMeter("RowIndexSizeWarnings");
         rowIndexSizeAborts = createKeyspaceMeter("RowIndexSizeAborts");
         rowIndexSize = createKeyspaceHistogram("RowIndexSize", false);
+
+        tooManySSTableIndexesReadWarnings = createKeyspaceMeter("TooManySSTableIndexesReadWarnings");
+        tooManySSTableIndexesReadAborts = createKeyspaceMeter("TooManySSTableIndexesReadAborts");
 
         formatSpecificGauges = createFormatSpecificGauges(keyspace);
     }
--- a/src/java/org/apache/cassandra/metrics/TableMetrics.java
+++ b/src/java/org/apache/cassandra/metrics/TableMetrics.java
@@ -277,6 +277,9 @@ public class TableMetrics
     public final TableMeter rowIndexSizeAborts;
     public final TableHistogram rowIndexSize;
 
+    public final TableMeter tooManySSTableIndexesReadWarnings;
+    public final TableMeter tooManySSTableIndexesReadAborts;
+
     public final ImmutableMap<SSTableFormat<?, ?>, ImmutableMap<String, Gauge<? extends Number>>> formatSpecificGauges;
 
     private static Pair<Long, Long> totalNonSystemTablesSize(Predicate<SSTableReader> predicate)
@@ -854,6 +857,9 @@ public class TableMetrics
         rowIndexSizeWarnings = createTableMeter("RowIndexSizeWarnings", cfs.keyspace.metric.rowIndexSizeWarnings);
         rowIndexSizeAborts = createTableMeter("RowIndexSizeAborts", cfs.keyspace.metric.rowIndexSizeAborts);
         rowIndexSize = createTableHistogram("RowIndexSize", cfs.keyspace.metric.rowIndexSize, false);
+
+        tooManySSTableIndexesReadWarnings = createTableMeter("TooManySSTableIndexesReadWarnings", cfs.keyspace.metric.tooManySSTableIndexesReadWarnings);
+        tooManySSTableIndexesReadAborts = createTableMeter("TooManySSTableIndexesReadAborts", cfs.keyspace.metric.tooManySSTableIndexesReadAborts);
 
         formatSpecificGauges = createFormatSpecificGauges(cfs);
     }
--- a/src/java/org/apache/cassandra/service/reads/thresholds/CoordinatorWarnings.java
+++ b/src/java/org/apache/cassandra/service/reads/thresholds/CoordinatorWarnings.java
@@ -96,6 +96,9 @@ public class CoordinatorWarnings
 
             recordAborts(merged.rowIndexReadSize, cql, loggableTokens, cfs.metric.rowIndexSizeAborts, WarningsSnapshot::rowIndexReadSizeAbortMessage);
             recordWarnings(merged.rowIndexReadSize, cql, loggableTokens, cfs.metric.rowIndexSizeWarnings, WarningsSnapshot::rowIndexSizeWarnMessage);
+
+            recordAborts(merged.indexReadSSTablesCount, cql, loggableTokens, cfs.metric.tooManySSTableIndexesReadAborts, WarningsSnapshot::tooManyIndexesReadAbortMessage);
+            recordWarnings(merged.indexReadSSTablesCount, cql, loggableTokens, cfs.metric.tooManySSTableIndexesReadWarnings, WarningsSnapshot::tooManyIndexesReadWarnMessage);
         });
 
         // reset the state to block from double publishing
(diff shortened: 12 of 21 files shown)


src/java/org/apache/cassandra/index/sai/plan/QueryController.java
```diff
@@ -210,6 +215,34 @@ public KeyRangeIterator.Builder getIndexQueryResults(Collection<Expression> expr
...
+        if (!nonPartitionRestricted)
+            return;
+
+        Guardrails.maximumSaiSSTablesWithNonParitionRestrictedQuery.guard(queryView.referencedIndexes.size(),
+                                                                          format("%s on columns %s", command.metadata().toString(),
+                                                                                 queryView.view.stream()
+                                                                                               .map(p -> p.left.getIndexTermType().columnMetadata().name.toString())
+                                                                                               .collect(Collectors.joining(", "))),
+                                                                          false,
+                                                                          null); // ???? how to propagate ClientState here?
```
smiklosovic: this is pretty tricky!

maedhroz: See `SortedTableWriter#guardParititonThreshold()`. It doesn't actually use the client state, but passes null there.

src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
```diff
@@ -404,7 +404,15 @@ public ReadQuery getQuery(QueryOptions options,
         boolean isPartitionRangeQuery = restrictions.isKeyRange() || restrictions.usesSecondaryIndexing();
 
         if (isPartitionRangeQuery)
+        {
+            if (restrictions.usesSecondaryIndexing()
+                && !SchemaConstants.isLocalSystemKeyspace(table.keyspace)
+                && !restrictions.hasPartitionKeyRestrictions()
+                && cfs().isLeveledCompaction())
```
smiklosovic: The deal is that this is an issue on LCS only.

src/java/org/apache/cassandra/db/ReadCommand.java
```diff
@@ -425,11 +425,14 @@ public UnfilteredPartitionIterator executeLocally(ReadExecutionController execut
                                             .collect(Collectors.joining(",")));
             }
 
-            UnfilteredPartitionIterator iterator = (null == searcher) ? queryStorage(cfs, executionController) : searcher.search(executionController);
```
smiklosovic: I think that this should be moved inside `try` because if we get `iterator` and then it will throw in `RTBoundValidator.validate` then such iterator is never closed. Might happen ... right?

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -477,6 +478,21 @@ public final class Guardrails implements GuardrailsMBean
                      format("The modification to table %s has a timestamp %s before the minimum allowable %s threshold %s",
                             what, value, isWarning ? "warning" : "failure", threshold));
 
+    public static final MaxThreshold maximumReferencedIndexesWithNonParitionRestrictedQuery =
```
smiklosovic: I chose to go with guardrail approach even we are not checking guardrail as such (will be visible further in down in the patch). This has a nice property of being update-able via JMX in runtime which would be impossible with a static configuration property.

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
```diff
@@ -210,6 +216,24 @@ public KeyRangeIterator.Builder getIndexQueryResults(Collection<Expression> expr
         return builder;
     }
 
+    private void maybeTriggerGuardrails(QueryViewBuilder.QueryView queryView)
+    {
+        if (!cfs.isLeveledCompaction())
```
smiklosovic: We double check this is happening to LCS-backed table only.

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
```diff
@@ -210,6 +216,24 @@ public KeyRangeIterator.Builder getIndexQueryResults(Collection<Expression> expr
...
+            return;
+
+        int referencedIndexes = queryView.referencedIndexes.size();
+
+        if (Guardrails.maximumReferencedIndexesWithNonParitionRestrictedQuery.failsOn(referencedIndexes, null))
```
smiklosovic: here I am just checking if it is going to fail, we do not need any ClientState for that.

src/java/org/apache/cassandra/index/sai/plan/QueryReferencingTooManyIndexesException.java
```diff
@@ -0,0 +1,29 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.index.sai.plan;
+
+import org.apache.cassandra.db.RejectException;
+
+public class QueryReferencingTooManyIndexesException extends RejectException
```
smiklosovic, line 23: this exception extends `RejectedException`, it is different from `QueryReferencesTooManyIndexesAbortException` which extends `ReadAbortException`. Same pattern is used in tombstone case if you look closely. There are two tombstone-related exceptions, one rejecting the read and another one aborting it.

src/java/org/apache/cassandra/service/reads/thresholds/CoordinatorWarnings.java
```diff
@@ -147,14 +152,14 @@ private static void clearState()
     @FunctionalInterface
     private interface ToString
     {
-        String apply(int count, long value, String cql);
+        String apply(ImmutableSet<InetAddressAndPort> nodes, long value, String cql);
```
smiklosovic: It is cool to know what exact replicas violated the tresholds so an operator can e.g. try to compact them to lower the amount of referenced indexes (or increase the thresholds for them).

src/java/org/apache/cassandra/service/reads/thresholds/WarningsSnapshot.java
```diff
@@ -83,58 +98,79 @@ WarningsSnapshot merge(WarningsSnapshot other)
...
 
     @VisibleForTesting
-    public static String rowIndexSizeWarnMessage(int nodes, long bytes, String cql)
+    public static String rowIndexSizeWarnMessage(ImmutableSet<InetAddressAndPort> nodes, long bytes, String cql)
+    {
+        return String.format("%s nodes loaded over %s bytes in RowIndexEntry and issued warnings for query %s (see row_index_size_warn_threshold)", nodes, bytes, cql);
+    }
+
+    @VisibleForTesting
+    public static String tooManyIndexesReadWarnMessage(ImmutableSet<InetAddressAndPort> nodes, long value, String cql)
```
smiklosovic: it would be cool to know exactly what the violations were, _per violated replica_. The way Warnings and its Counter is done for now is that as it merges responses from replicas, it will always merge it in such a way that we only know _maximum_. So if there are three nodes, one read from 5 referenced indexes, the second one from 10 and the third one from 15, we will see all replicas here but the maximum showed will be 15 - and we do not know what node it is. This would require further refactoring, I think counters would need to be more like a gauge, per node, or still counter, but we can not choose max.

src/java/org/apache/cassandra/service/reads/thresholds/CoordinatorWarnings.java
maedhroz: Possibly, although for LCS, the main source of this kind of problem, compaction will not be a fix. Also, when this problem exists, it will likely exist on all replicas (or almost all replicas). This guardrail, in many ways, is a pre-prod guardrail. If you hit it, it usually means you've gone very wrong at the data modeling stage, and there really isn't an easy fix.

src/java/org/apache/cassandra/service/reads/thresholds/WarningsSnapshot.java
maedhroz: see https://github.com/apache/cassandra/pull/2961#discussion_r1420774359

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailNonPartitionKeyRestrictedQueriesTest.java
```diff
@@ -0,0 +1,202 @@
...
+            else
+                cluster.get(nodesToFlush).forEach((instance) -> instance.flush(KEYSPACE));
+        }
+
+        return value;
+    }
+
+    private void execute(String query, Object... args)
+    {
+        cluster.coordinator(1).execute(format(query), ConsistencyLevel.ALL, args);
```
maedhroz: While I don't think it's harmful to look at the logging side effects when these trigger, that can be brittle, and it looks like `assertLogsOnAction()` is only checking generally for ERROR or WARN, not with any qualifiers that relate to the guardrail?

Either way, the basic assertion we need to make in the warning case is on the warning returned to the client. In the lines above, you can access them this way:

```
SimpleQueryResult result = cluster.coordinator(1).executeWithResult(format(query), ConsistencyLevel.ALL, result.warnings(); // assert on the contents of this
```

In other words, I'm fine w/ keeping the logs checking and tying the assertions directly to specific messages for the features we're testing, but it might make more sense to focus on the actual client warnings and errors.

conf/cassandra.yaml
```diff
@@ -2076,6 +2076,15 @@ drop_compact_storage_enabled: false
 # if zero_ttl_on_twcs_enabled is set to false, this property is irrelevant as such statements will fail.
 #zero_ttl_on_twcs_warned: true
 
+# Guardrail enabling querying a table with indexes without specifying partition keys, defaults to true
```
maedhroz:
```suggestion
# Guardrail enabling secondary index queries that do not restrict on partition key (defaults to true)
```
I'll keep reading, but I'm assuming we're only looking at cases where we actually query an index, rather than just have an index on the table.

conf/cassandra.yaml
```diff
@@ -2076,6 +2076,15 @@ drop_compact_storage_enabled: false
...
+#non_partition_key_restricted_query_enabled: true
+# Maximum number of referenced indexes on a replica when executing a SELECT query without partition key restrictions
```
maedhroz: nit: In our inline comments, we should probably make it clear that this is an SAI-specific thing. SASI could have used it too, in theory, but SAI should be completely replacing it, so nobody cares.

conf/cassandra.yaml
```diff
@@ -2076,6 +2076,15 @@ drop_compact_storage_enabled: false
...
+# before emitting a warning, defaults to 32.
+#maximum_sstables_with_non_paritition_key_restricted_query_warn_threshold: 32
```
maedhroz: How about `sai_sstable_indexes_per_query_warn_threshold` and `sai_sstable_indexes_per_query_fail_threshold `? We communicate this is an SAI-only thing, make it clear that it's per query, etc.

conf/cassandra.yaml
maedhroz: ...and we can remove the bit about the SELECT not being partition key restricted, because that shouldn't matter.

conf/cassandra.yaml
```diff
@@ -2076,6 +2076,15 @@ drop_compact_storage_enabled: false
 # if zero_ttl_on_twcs_enabled is set to false, this property is irrelevant as such statements will fail.
 #zero_ttl_on_twcs_warned: true
 
+# Guardrail enabling querying a table with indexes without specifying partition keys, defaults to true
+#non_partition_key_restricted_query_enabled: true
```
maedhroz: nit: The only other naming option I can think of here, especially since this is related to index queries only, is `non_partition_restricted_index_query_enabled` (the "key" bit we can throw out, since "restricting to a partition" is just as intelligible as "restricting on partition key" or whatever)

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -897,6 +897,9 @@ public static void setClientMode(boolean clientMode)
     public volatile int maximum_replication_factor_fail_threshold = -1;
     public volatile boolean zero_ttl_on_twcs_warned = true;
     public volatile boolean zero_ttl_on_twcs_enabled = true;
+    public volatile boolean non_partition_key_restricted_query_enabled = true;
+    public volatile int maximum_referenced_indexes_for_non_paritition_key_restricted_query_warn_threshold = 32;
+    public volatile int maximum_referenced_indexes_for_non_paritition_key_restricted_query_fail_threshold = -1;
```
maedhroz: Whatever we settle on, naming here needs to agree w/ `cassandra.yaml` and `validateMaxLongThreshold()` argument in `GuardrailsOptions`

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailNonPartitionKeyRestrictedQueriesTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+        assertLogsOnAction(() -> execute("SELECT * from %s WHERE v1 = " + valueToQuery), false, false);
+
+        // create 3 more SSTables on each node, that means each node will have in total 4 SSTables (one from previous run)
+        // this will trigger warn threshold
+        long valueToQuery2 = createSSTables(3);
+        assertNumberOfSSTables(4);
+
+        // we violated warn threshold of 2 so all nodes will warn
+        assertLogsOnAction(() -> execute("SELECT * from %s WHERE v1 = " + valueToQuery2), true, false, 1);
```
maedhroz: Note here that this query will only hit 3 SSTables, because `valueToQuery2` only appears in the range of the second set of 3 SSTables flushed. SAI will filter out the first w/ only the value `valueToQuery`.

src/java/org/apache/cassandra/service/reads/thresholds/CoordinatorWarnings.java
smiklosovic: @maedhroz neverthless, I think this is valuable for other warnings out there. Could this stay?

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
maedhroz: Yeah, the standardization there is nice.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -477,6 +478,21 @@ public final class Guardrails implements GuardrailsMBean
...
+    new MaxThreshold("maximum_referenced_indexes_for_non_paritition_key_restricted_query",
```
maedhroz: nit: Make sure naming is consistent at the end w/ what's in the YAML and `Config`

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -477,6 +478,21 @@ public final class Guardrails implements GuardrailsMBean
...
+                     "High number of referenced indexes per query may negatively effect performance",
+                     state -> CONFIG_PROVIDER.getOrCreate(state).getMaximumReferencedIndexesWithNonPartitionRestrictedQueryWarnThreshold(),
+                     state -> CONFIG_PROVIDER.getOrCreate(state).getMaximumReferencedIndexesWithNonPartitionRestrictedQueryFailThreshold(),
+                     ((isWarning, what, value, threshold) ->
+                      format("The number of referenced indexes on a table with %s on index %s violated %s threshold value %s with value %s",
+                             LeveledCompactionStrategy.class.getName(), what, isWarning ? "warning" : "failure", threshold, value)));
+
+    public static final EnableFlag nonPartitionRestrictedQuery =
+    new EnableFlag("non_partition_restricted_query",
```
maedhroz: nit: Make sure naming is consistent at the end w/ what's in the YAML and `Config`

src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
```diff
@@ -404,7 +404,13 @@ public ReadQuery getQuery(QueryOptions options,
         boolean isPartitionRangeQuery = restrictions.isKeyRange() || restrictions.usesSecondaryIndexing();
 
         if (isPartitionRangeQuery)
+        {
+            if (restrictions.usesSecondaryIndexing()
+                && !SchemaConstants.isLocalSystemKeyspace(table.keyspace))
```
maedhroz: super nit: I'd maybe make the condition a one-liner so it isn't aligned w/ the `Guardrails...` below :D

src/java/org/apache/cassandra/service/reads/thresholds/CoordinatorWarnings.java
maedhroz: I mean, you could leave it, but then you a.) have change the message in `tombstoneAbortMessage()` et al. and b.) might have to deal w/ printing an arbitrarily large set of nodes, right? (You could like, truncate at 3 nodes or something...)

I'm guessing the original intent (CC @dcapwell) was to let users see the problem in aggregate and then hit the logs to see what nodes exactly were having issues. I'd vote for either leaving things as they are or printing a truncated list of problematic nodes in the implementations of `ToString`.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailNonPartitionKeyRestrictedQueriesTest.java
smiklosovic: Sure, I think that checking warnings directly from the response is better. We might get a log while executing some logic on WARN or ERROR level which is not necessarily related to the feature we are testing. It will be more stable as well etc.

conf/cassandra.yaml
smiklosovic: right ... wording ...

conf/cassandra.yaml
smiklosovic: @maedhroz , well ... we plan to cover old / original secondary indexes too, no? Then the message not referencing SAI is probably just fine?

conf/cassandra.yaml
smiklosovic: works for me

conf/cassandra.yaml
maedhroz: No, the concept of SSTable attachment and therefore SSTable indexes is specific to SAI (and SASI, but we don't care). The partition restriction guardrail is the one that applies to all 2i implementations.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -477,6 +478,21 @@ public final class Guardrails implements GuardrailsMBean
                      format("The modification to table %s has a timestamp %s before the minimum allowable %s threshold %s",
                             what, value, isWarning ? "warning" : "failure", threshold));
 
+    public static final MaxThreshold maximumReferencedIndexesWithNonParitionRestrictedQuery =
+    new MaxThreshold("maximum_referenced_indexes_for_non_paritition_key_restricted_query",
+                     "High number of referenced indexes per query may negatively effect performance",
+                     state -> CONFIG_PROVIDER.getOrCreate(state).getMaximumReferencedIndexesWithNonPartitionRestrictedQueryWarnThreshold(),
+                     state -> CONFIG_PROVIDER.getOrCreate(state).getMaximumReferencedIndexesWithNonPartitionRestrictedQueryFailThreshold(),
+                     ((isWarning, what, value, threshold) ->
+                      format("The number of referenced indexes on a table with %s on index %s violated %s threshold value %s with value %s",
```
maedhroz: No need to mention LCS specifically. Something like...

```
The number of SSTable indexes queried on index %s violated...
```

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -770,4 +770,40 @@ public interface GuardrailsMBean
...
+     */
+    boolean getNonPartitionRestrictedQueryEnabled();
+
+    /**
+     * Sets whether it is possible to execute a query against indexes without specifying
+     * any partition key restrictions.
+     *
+     * @param enabled {@code true} if a query without partition key is enabled or not
+     */
+    void setNonPartitionRestrictedQueryEnabled(boolean enabled);
```
maedhroz, line 806: nit: Nothing super urgent here...just come back at the end and make sure things are named in a consistent way, etc.

src/java/org/apache/cassandra/metrics/KeyspaceMetrics.java
```diff
@@ -275,6 +279,10 @@ public KeyspaceMetrics(final Keyspace ks)
         rowIndexSizeAborts = createKeyspaceMeter("RowIndexSizeAborts");
         rowIndexSize = createKeyspaceHistogram("RowIndexSize", false);
 
+        tooManyIndexReadSSTablesWarnings = createKeyspaceMeter("TooManyIndexReadSSTablesWarnings");
```
maedhroz: Does anything update these 3 metrics?

src/java/org/apache/cassandra/metrics/KeyspaceMetrics.java
```diff
@@ -275,6 +279,10 @@ public KeyspaceMetrics(final Keyspace ks)
...
+        tooManyIndexReadSSTablesAborts = createKeyspaceMeter("TooManyIndexReadSSTablesAborts");
+        tooManyIndexReadSSTables = createKeyspaceHistogram("TooManyIndexReadSSTables", false);
```
maedhroz: I noticed we already have this histogram in `TableQueryMetrics.PerQueryMetrics.sstablesHit` BUT (and @mike-tr-adamson can correct me here) it looks like it's only updated for top-k queries? We should probably just make sure update that for all queries and just remove this `TooManyIndexReadSSTables` histogram.

src/java/org/apache/cassandra/metrics/KeyspaceMetrics.java
maedhroz: We might also consider just moving these to `PerQueryMetrics`, but I'd be fine w/ keeping them here if we name appropriately. My proposal would be `TooManySSTableIndexesQueried[Warnings|Aborts]`.

@mike-tr-adamson WDYT?

src/java/org/apache/cassandra/metrics/KeyspaceMetrics.java
smiklosovic: @maedhroz I am happy to remove histogram.

src/java/org/apache/cassandra/metrics/KeyspaceMetrics.java
smiklosovic: @maedhroz  two of them are updated in `CoordinatorWarnings.done()`

src/java/org/apache/cassandra/metrics/TableMetrics.java
```diff
@@ -855,6 +859,10 @@ public Long getValue()
         rowIndexSizeAborts = createTableMeter("RowIndexSizeAborts", cfs.keyspace.metric.rowIndexSizeAborts);
         rowIndexSize = createTableHistogram("RowIndexSize", cfs.keyspace.metric.rowIndexSize, false);
 
+        tooManyIndexReadSSTablesWarnings = createTableMeter("TooManyIndexReadSSTablesWarnings", cfs.keyspace.metric.tooManyIndexReadSSTablesWarnings);
+        tooManyIndexReadSSTablesAborts = createTableMeter("TooManyIndexReadSSTablesAborts", cfs.keyspace.metric.tooManyIndexReadSSTablesAborts);
+        tooManyIndexReadSSTables = createTableHistogram("TooManyIndexReadSSTables", cfs.keyspace.metric.tooManyIndexReadSSTables, false);
```
maedhroz: Same comments as the keyspace metrics in terms of naming and whether we want to move the SSTables hit histogram.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailNonPartitionKeyRestrictedQueriesTest.java
smiklosovic: sure, I dont think this is anything comment-worthy (in the code), is it?

src/java/org/apache/cassandra/service/reads/thresholds/CoordinatorWarnings.java
smiklosovic: Rather to keep it as it was before than truncating it. I don't like to truncate it. BTW I don't think it would be too long anyway ... what RF you have out there at most? 5? 9? And not all replicas are going to violate ... it is not like this list would be practically without a reasonable limit.

maedhroz: This could be hundreds of nodes when the partition key is not restricted, i.e. not RF, but N/RF

Also, when one node violates, it does become much more likely that others will shortly (if they aren't already).

smiklosovic: OK

src/java/org/apache/cassandra/db/ReadCommand.java
maedhroz: I don't think there's actually anything in `RTBoundValidator.validate()` that can throw, and in the previous line, we'll fail before we even get an iterator. I'd revert this, at least as part of this patch. Even if we really wanted to change it, I think we'd just want to a.) move only `validate()` into the `try` and b.) remove the unnecessary `iterator == null` check.

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
maedhroz: We should follow the pattern of `LocalReadSizeTooLargeException` and trace here, using a common error message for the exception, etc.

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -2219,6 +2220,11 @@ protected void runMayThrow()
                     handler.onFailure(FBUtilities.getBroadcastAddressAndPort(), RequestFailureReason.READ_TOO_MANY_TOMBSTONES);
                     logger.error(t.getMessage());
                 }
+                else if (t instanceof QueryReferencingTooManyIndexesException)
+                {
+                    handler.onFailure(FBUtilities.getBroadcastAddressAndPort(), RequestFailureReason.READ_TOO_MANY_INDEXES);
+                    logger.error(t.getMessage());
+                }
```
maedhroz: I *think* we don't have to do this. It was a special case for tombstone thresholds, which existed before the new framework that local read size et al use.

src/java/org/apache/cassandra/metrics/KeyspaceMetrics.java
mike-tr-adamson: `TableQueryMetrics.PerQueryMetrics.sstablesHit` is used on all index search paths. It is updated from `QueryContext.sstablesHit` and that is updated in the `IndexSearchResultIterator`.

maedhroz: I stand corrected. Thanks Mike. Yeah, feels like we can just remove the `tooManyIndexReadSSTables` histogram then.

conf/cassandra.yaml
```diff
@@ -2076,6 +2076,15 @@ drop_compact_storage_enabled: false
 # if zero_ttl_on_twcs_enabled is set to false, this property is irrelevant as such statements will fail.
 #zero_ttl_on_twcs_warned: true
 
+# Guardrail enabling secondary index queries that do not restrict on partition key (defaults to true)
+#non_partition_restricted_index_query_enabled: true
+# Maximum number of referenced SAI indexes on a replica when executing a SELECT query
+# before emitting a warning, defaults to 32.
+#sai_sstable_indexes_per_query_warn_threshold: 32
+# Maximum number of referenced SAI indexes on a replica when executing a SELECT query
+# before emitting a failure, defaults to -1 as it will never fail.
```
maedhroz:
```suggestion
# before emitting a failure (defaults to -1 to disable)
```

conf/cassandra.yaml
```diff
@@ -2076,6 +2076,15 @@ drop_compact_storage_enabled: false
 # if zero_ttl_on_twcs_enabled is set to false, this property is irrelevant as such statements will fail.
 #zero_ttl_on_twcs_warned: true
 
+# Guardrail enabling secondary index queries that do not restrict on partition key (defaults to true)
+#non_partition_restricted_index_query_enabled: true
+# Maximum number of referenced SAI indexes on a replica when executing a SELECT query
+# before emitting a warning, defaults to 32.
```
maedhroz:
```suggestion
# before emitting a warning (defaults to 32)
```

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -89,6 +89,10 @@ public GuardrailsOptions(Config config)
         validateMaxRFThreshold(config.maximum_replication_factor_warn_threshold, config.maximum_replication_factor_fail_threshold);
         validateTimestampThreshold(config.maximum_timestamp_warn_threshold, config.maximum_timestamp_fail_threshold, "maximum_timestamp");
         validateTimestampThreshold(config.minimum_timestamp_warn_threshold, config.minimum_timestamp_fail_threshold, "minimum_timestamp");
+        validateMaxLongThreshold(config.sai_sstable_indexes_per_query_warn_threshold,
+                                 config.sai_sstable_indexes_per_query_fail_threshold,
+                                 "maximum_sstables_with_non_paritition_key_restricted_query",
```
maedhroz:
```suggestion
                                 "sai_sstable_indexes_per_query",
```

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -477,6 +477,21 @@ public final class Guardrails implements GuardrailsMBean
                      format("The modification to table %s has a timestamp %s before the minimum allowable %s threshold %s",
                             what, value, isWarning ? "warning" : "failure", threshold));
 
+    public static final MaxThreshold maximumReferencedIndexesWithNonParitionRestrictedQuery =
```
maedhroz:
```suggestion
    public static final MaxThreshold saiSSTableIndexesPerQuery =
```

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -477,6 +477,21 @@ public final class Guardrails implements GuardrailsMBean
...
+    new MaxThreshold("maximum_referenced_indexes_for_non_paritition_key_restricted_query",
```
maedhroz:
```suggestion
    new MaxThreshold("sai_sstable_indexes_per_query",
```

src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
```diff
@@ -429,4 +429,40 @@ void setMaximumTimestampThreshold(@Nullable DurationSpec.LongMicrosecondsBound w
      */
     void setMinimumTimestampThreshold(@Nullable DurationSpec.LongMicrosecondsBound warn,
                                       @Nullable DurationSpec.LongMicrosecondsBound fail);
+
+    /**
+     * @return The threshold to warn when number of referenced indexes
+     * to be searched for non-partition restricted query is greater than treshold.
+     */
```
maedhroz, line 435: Suggestion:

```
/**
  * @return the warning threshold for the number of SAI SSTable indexes searched by a query
  */
```

src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
```diff
@@ -429,4 +429,40 @@ void setMaximumTimestampThreshold(@Nullable DurationSpec.LongMicrosecondsBound w
...
+    int getSaiSSTableIndexesPerQueryWarnThreshold();
+
+    /**
+     * @return The threshold to fail when number of referenced indexes
+     * to be searched for non-partition restricted query is greater than treshold.
+     */
```
maedhroz, line 440: Suggestion:
```
/**
  * @return the failure threshold for the number of SAI SSTable indexes searched by a query
  */
```

src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
```diff
@@ -429,4 +429,40 @@ void setMaximumTimestampThreshold(@Nullable DurationSpec.LongMicrosecondsBound w
...
+    int getSaiSSTableIndexesPerQueryFailThreshold();
+
+    /**
+     * Sets warn and fail thresholds for a guardrail checking number of referenced indexes
+     * to search for non-partition restricted query.
+     * @param warn value to set for warn threshold
+     * @param fail value to set for fail threshold
+     */
```
maedhroz, line 448: Suggestion:
```
/**
  * Sets warning and failure thresholds for the number of SAI SSTable indexes searched by a query
  * ...
  */
```

maedhroz: For all three of these, big thing is removing the language around partition restrictions, which don't come into play explicitly for this guardrail.

src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
```diff
@@ -429,4 +429,40 @@ void setMaximumTimestampThreshold(@Nullable DurationSpec.LongMicrosecondsBound w
...
+    void setSaiSSTableIndexesPerQueryThreshold(int warn, int fail);
+
+    /**
+     * Returns whether it is possible to execute a query against indexes (secondary or SAI) without specifying
```
maedhroz:
```suggestion
     * Returns whether it is possible to execute a query against secondary indexes without specifying
```
Technically SAI is a secondary index...might as well be general.

src/java/org/apache/cassandra/db/guardrails/GuardrailsConfig.java
```diff
@@ -429,4 +429,40 @@ void setMaximumTimestampThreshold(@Nullable DurationSpec.LongMicrosecondsBound w
...
+     * any partition key restrictions.
+     *
+     * @return true if it is possible to execute a query without a partition key, false otherwise
+     */
+    boolean getNonPartitionRestrictedQueryEnabled();
+
+    /**
+     * Sets whether it is possible to execute a query against indexes (secondary or SAI) without specifying
```
maedhroz, line 460:
```suggestion
     * Sets whether it is possible to execute a query against secondary indexes without specifying
```
Technically SAI is a secondary index...might as well be general.

src/java/org/apache/cassandra/db/guardrails/GuardrailsMBean.java
```diff
@@ -770,4 +770,40 @@ public interface GuardrailsMBean
...
+     * @return true if it is possible to execute a query without a partition key, false otherwise
+     */
+    boolean getNonPartitionRestrictedIndexQueryEnabled();
+
+    /**
+     * Sets whether it is possible to execute a query against indexes without specifying
+     * any partition key restrictions.
+     *
+     * @param enabled {@code true} if a query without partition key is enabled or not
+     */
```
maedhroz, line 805: Basically all the comments from `GuardrailsConfig` above apply here.

src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
```diff
@@ -404,7 +404,12 @@ public ReadQuery getQuery(QueryOptions options,
         boolean isPartitionRangeQuery = restrictions.isKeyRange() || restrictions.usesSecondaryIndexing();
 
         if (isPartitionRangeQuery)
+        {
+            if (restrictions.usesSecondaryIndexing() && !SchemaConstants.isLocalSystemKeyspace(table.keyspace))
+                Guardrails.nonPartitionRestrictedIndexQueryEnabled.ensureEnabled(state);
```
maedhroz, line 409: This is only correct for queries that don't have partition key restrictions. To see this, modify line 289 in your test, in `executeSelectViaDriver()`...

```
return executeViaDriver(String.format("SELECT * from %s.%s WHERE k = 0 AND v1 = 0", KEYSPACE, tableName));
```

The tests will still pass, even though they shouldn't. In other words, we still test the guardrail even when we restrict the partition key. To fix this, all I think we need to do is add a `restrictions.isKeyRange() &&` to the beginning of our new `if` statement and add a test for that case.

maedhroz: Put another way, `restrictions.usesSecondaryIndexing()` can be `true` even when the partition key is restricted.

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailNonPartitionRestrictedQueryTest.java
```diff
@@ -0,0 +1,450 @@
...
+                                 .call().intValue());
+            assertEquals(expectedFail,
+                         instance.callsOnInstance((IIsolatedExecutor.SerializableCallable<Integer>) () -> Guardrails.instance.getSaiSSTableIndexesPerQueryFailThreshold())
+                                 .call().intValue());
+        });
+    }
+
+    private long totalWarnings()
+    {
+        return cluster.stream().mapToLong(i -> i.metrics().getCounter("org.apache.cassandra.metrics.keyspace.TooManyIndexReadSSTablesWarnings.distributed_test_keyspace")).sum();
```
maedhroz: nit: Alternate, more consistent naming

```suggestion
        return cluster.stream().mapToLong(i -> i.metrics().getCounter("org.apache.cassandra.metrics.keyspace. TooManySSTableIndexesQueriedWarnings.distributed_test_keyspace")).sum();
```

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailNonPartitionRestrictedQueryTest.java
```diff
@@ -0,0 +1,450 @@
...
+    }
+
+    private long totalAborts()
+    {
+        return cluster.stream().mapToLong(i -> i.metrics().getCounter("org.apache.cassandra.metrics.keyspace.TooManyIndexReadSSTablesAborts.distributed_test_keyspace")).sum();
```
maedhroz: nit: Alternate, more consistent naming
```suggestion
        return cluster.stream().mapToLong(i -> i.metrics().getCounter("org.apache.cassandra.metrics.keyspace. TooManySSTableIndexesQueriedAborts.distributed_test_keyspace")).sum();
```

src/java/org/apache/cassandra/metrics/KeyspaceMetrics.java
```diff
@@ -275,6 +278,9 @@ public KeyspaceMetrics(final Keyspace ks)
         rowIndexSizeAborts = createKeyspaceMeter("RowIndexSizeAborts");
         rowIndexSize = createKeyspaceHistogram("RowIndexSize", false);
 
+        tooManyIndexReadSSTablesWarnings = createKeyspaceMeter("TooManyIndexReadSSTablesWarnings");
```
maedhroz:
```suggestion
        tooManyIndexReadSSTablesWarnings = createKeyspaceMeter("TooManySSTableIndexesQueriedWarnings");
```

src/java/org/apache/cassandra/metrics/KeyspaceMetrics.java
```diff
@@ -275,6 +278,9 @@ public KeyspaceMetrics(final Keyspace ks)
...
+        tooManyIndexReadSSTablesAborts = createKeyspaceMeter("TooManyIndexReadSSTablesAborts");
```
maedhroz:
```suggestion
        tooManyIndexReadSSTablesAborts = createKeyspaceMeter("TooManySSTableIndexesQueriedAborts");
```

src/java/org/apache/cassandra/config/GuardrailsOptions.java
smiklosovic: damn how could I forget this, shame

src/java/org/apache/cassandra/metrics/TableMetrics.java
```diff
@@ -855,6 +858,9 @@ public Long getValue()
         rowIndexSizeAborts = createTableMeter("RowIndexSizeAborts", cfs.keyspace.metric.rowIndexSizeAborts);
         rowIndexSize = createTableHistogram("RowIndexSize", cfs.keyspace.metric.rowIndexSize, false);
 
+        tooManyIndexReadSSTablesWarnings = createTableMeter("TooManyIndexReadSSTablesWarnings", cfs.keyspace.metric.tooManyIndexReadSSTablesWarnings);
```
maedhroz:
```suggestion
        tooManyIndexReadSSTablesWarnings = createTableMeter("TooManySSTableIndexesQueriedWarnings", cfs.keyspace.metric.tooManyIndexReadSSTablesWarnings);
```

src/java/org/apache/cassandra/metrics/TableMetrics.java
```diff
@@ -855,6 +858,9 @@ public Long getValue()
...
+        tooManyIndexReadSSTablesAborts = createTableMeter("TooManyIndexReadSSTablesAborts", cfs.keyspace.metric.tooManyIndexReadSSTablesAborts);
```
maedhroz:
```suggestion
        tooManyIndexReadSSTablesAborts = createTableMeter("TooManySSTableIndexesQueriedAborts", cfs.keyspace.metric.tooManyIndexReadSSTablesAborts);
```

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
```diff
@@ -184,6 +188,8 @@ public KeyRangeIterator.Builder getIndexQueryResults(Collection<Expression> expr
 
         QueryViewBuilder.QueryView queryView = queryViewBuilder.build();
 
+        maybeTriggerGuardrails(queryView);
```
maedhroz: So I think the reason you're seeing LEAKs in the new tests is that this needs to happen inside the `try` below. If we don't do that, we won't release the `SSTableIndex`es when we hit the failure threshold.

(Just verified locally on my machine that your tests are clean w/ this change.)

maedhroz (changes requested):
Tried to keep the nits to a minimum, but there are a couple things we need to fix.

1.) Make sure the pk restriction guardrail isn't ever activated when the pk is restricted.
2.) Avoiding the `SSTableIndex` leak when the index count fail threshold is triggered.

src/java/org/apache/cassandra/cql3/statements/SelectStatement.java
smiklosovic: ah this is good point

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
smiklosovic: hah great catch!

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -900,6 +904,48 @@ public void setMinimumTimestampThreshold(@Nullable DurationSpec.LongMicroseconds
...
+                                  x -> config.sai_sstable_indexes_per_query_warn_threshold = x);
+
+        updatePropertyWithLogging("sai_sstable_indexes_per_query_fail_threshold",
+                                  fail,
+                                  () -> config.sai_sstable_indexes_per_query_fail_threshold,
+                                  x -> config.sai_sstable_indexes_per_query_fail_threshold = x);
+    }
+
+    @Override
+    public boolean getNonPartitionRestrictedQueryEnabled()
```
maedhroz, line 935:
```suggestion
    public boolean getNonPartitionRestrictedIndexQueryEnabled()
```

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -900,6 +904,48 @@ public void setMinimumTimestampThreshold(@Nullable DurationSpec.LongMicroseconds
...
+    {
+        return config.non_partition_restricted_index_query_enabled;
+    }
+
+    @Override
+    public void setNonPartitionRestrictedQueryEnabled(boolean enabled)
```
maedhroz, line 941:
```suggestion
    public void setNonPartitionRestrictedIndexQueryEnabled(boolean enabled)
```

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -1200,6 +1215,36 @@ public void setMinimumTimestampThreshold(String warnSeconds, String failSeconds)
...
+    }
+
+    @Override
+    public void setSaiSSTableIndexesPerQueryThreshold(int warn, int fail)
+    {
+        DEFAULT_CONFIG.setSaiSSTableIndexesPerQueryThreshold(warn, fail);
+    }
+
+    @Override
+    public boolean getNonPartitionRestrictedQueryEnabled()
```
maedhroz, line 1237:
```suggestion
    public boolean getNonPartitionRestrictedIndexQueryEnabled()
```

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -1200,6 +1215,36 @@ public void setMinimumTimestampThreshold(String warnSeconds, String failSeconds)
...
+    {
+        return DEFAULT_CONFIG.getNonPartitionRestrictedQueryEnabled();
+    }
+
+    @Override
+    public void setNonPartitionRestrictedQueryEnabled(boolean enabled)
```
maedhroz, line 1243:
```suggestion
    public void setNonPartitionRestrictedIndexQueryEnabled(boolean enabled)
```

test/distributed/org/apache/cassandra/distributed/test/guardrails/GuardrailNonPartitionRestrictedQueryTest.java
```diff
@@ -0,0 +1,447 @@
...
+        cluster.forEach(instance -> assertTrue(instance.callsOnInstance((IIsolatedExecutor.SerializableCallable<Boolean>) () -> Guardrails.instance.getNonPartitionRestrictedQueryEnabled()).call()));
+    }
+
+    private void disableGuardrail()
+    {
+        cluster.forEach(instance -> instance.runOnInstance((IIsolatedExecutor.SerializableRunnable) () -> Guardrails.instance.setNonPartitionRestrictedQueryEnabled(false)));
+        cluster.forEach(instance -> assertFalse(instance.callsOnInstance((IIsolatedExecutor.SerializableCallable<Boolean>) () -> Guardrails.instance.getNonPartitionRestrictedQueryEnabled()).call()));
+    }
+
+    private void assertTresholds(int expectedWarn, int expectedFail, int... nodes)
```
maedhroz, line 500:
```suggestion
    private void assertThresholds(int expectedWarn, int expectedFail, int... nodes)
```

maedhroz (approved):

PR #2962, opened 2023-12-05
CASSANDRA-19166: StackOverflowError on ALTER after many previous schema changes

aratno:
Fix re-wrapping of TableMetadataRefCache fields in Collections.UnmodifiableMap, which was causing long reference chains where the underlying map was another instance of an UnmodifiableMap. Calls to `get` had to traverse all the way down the chain, and would eventually overflow the stack (and created a fair bit of extra garbage on the heap).

### DIFF
--- a/src/java/org/apache/cassandra/schema/TableMetadataRefCache.java
+++ b/src/java/org/apache/cassandra/schema/TableMetadataRefCache.java
@@ -48,7 +48,7 @@ class TableMetadataRefCache
     // (keyspace name, index name) -> mutable metadata ref map. We have to update these in place every time an index changes.
     private final Map<Pair<String, String>, TableMetadataRef> indexMetadataRefs;
 
-    public TableMetadataRefCache(Map<TableId, TableMetadataRef> metadataRefs,
+    private TableMetadataRefCache(Map<TableId, TableMetadataRef> metadataRefs,
                                  Map<Pair<String, String>, TableMetadataRef> metadataRefsByName,
                                  Map<Pair<String, String>, TableMetadataRef> indexMetadataRefs)
     {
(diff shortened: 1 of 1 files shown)


jacek-lewandowski (approved):

src/java/org/apache/cassandra/schema/TableMetadataRefCache.java
```diff
@@ -50,9 +52,9 @@ public TableMetadataRefCache(Map<TableId, TableMetadataRef> metadataRefs,
                                  Map<Pair<String, String>, TableMetadataRef> metadataRefsByName,
```
maedhroz, line 52: Let's make the constructor `private` so there's no way we can create a `TableMetadataRefCache` with unprotected maps.

maedhroz (approved):
Left one nit, but LGTM otherwise

src/java/org/apache/cassandra/schema/TableMetadataRefCache.java
aratno: @jacek-lewandowski I'd prefer to make this change as well, but didn't want to change APIs because this class exists for CASSANDRA-17044 (schema modularity). Is it safe to make this constructor private? If not, I'll make a comment expressing the caller's responsibility here.

aratno: I'm going to make this private for now - let me know if I should revert though

jacek-lewandowski: We could use `UnmodifiableMap.decorate` from commons collections and leave the constructor as it is. `decorate` is smarter and it would not wrap already wrapped collections.

jacek-lewandowski:
Committed https://github.com/apache/cassandra/commit/a443990bfa64e239810876121f2877064f2d9ae8

PR #2963, opened 2023-12-05
Additional fixes for CASSANDRA-19023

aweisberg:
The Accord journal entry fix is a bug introduced by interop.

The Accord Commands purger universal durability check fix is for a bug introduced by tighten bolts. `CompactionAccordIteratorsTest` still doesn't pass because reduce deps refactored `AccordKeyspace` pretty extensively and the test hasn't been updated yet. The test does pass if you apply these fixes before reduce deps.


src/java/org/apache/cassandra/db/compaction/CompactionIterator.java
```diff
@@ -805,7 +804,7 @@ protected Row applyToRow(Row row)
 
             // When commands end up being sliced by compaction we need this to discard tombstones and slices
             // without enough information to run the rest of the cleanup logic
-            if (durableBefore.min(txnId) == Universal)
+            if (durableBefore.isUniversal(txnId))
```
dcapwell: so rather than `Universal` we now use `UniversalOrInvalidated`?

dcapwell: shouldn't this be `UniversalOrInvalidated || Universal`?  maybe ask Benedict for clarity?

dcapwell (approved):
Not sure about `src/java/org/apache/cassandra/db/compaction/CompactionIterator.java`, should prob talk to Benedict

Rest LGTM

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -363,6 +363,7 @@ public static Row truncatedApply(SaveStatus newSaveStatus, Row row, long nowInSe
             // If durability is not universal we don't want to delete older versions of the row that might have recorded
             // a higher durability value. maybeDropTruncatedCommandColumns will take care of dropping things even if we don't drop via tombstones.
             // durability should be the only column that could have an older value that is insufficient for propagating forward
+            // TODO (now): with UniversalOrInvalidated should this change?
             boolean doDeletion = durability == Durability.Universal;
```
aweisberg, line 367: @belliottsmith seems like this probably needs to be updated now that `UniversalOrInvalidated` exists? Should this be `>= UniversalOrInvalidted.ordinal()` or equality on both versions of `Universal`?

aweisberg: Maybe the `Durability` enum should have `isSafeToPurge` and we can just ask it?

belliottsmith: I'm not sure such a method would be well-advised, at least in isolation or without some additional parameters. What we are permitted to do in various cases of durability depends on how that durability is implied. If it is recorded against a specific record we are permitted to erase different information.

By the looks of it this, this particular place is very C* specific, as it relates to LSMT compaction.

In this case it appears that we are using the record's own durability data, in which case it should be a bug to even encounter `UniversalOrInvalidated` - we know it cannot be `Invalidated`, because it has been `Applied`. We should perhaps enrich our `Command` validation logic to catch this kind of logic error, if we don't already.

PR #2966, opened 2023-12-06
UCS min_sstable_size should not be lower than target_sstable_size low…

Claudenw:
…er bound

```

```

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/unified/Controller.java
+++ b/src/java/org/apache/cassandra/db/compaction/unified/Controller.java
@@ -600,12 +600,12 @@ public class Controller
                 long sizeInBytes = FBUtilities.parseHumanReadableBytes(s);
                 // zero is a valid option to disable featue
                 if (sizeInBytes < 0)
-                    throw new ConfigurationException(String.format("Invalid configuration, %s should be greater than 0 (zero)",
+                    throw new ConfigurationException(String.format("Invalid configuration, %s should be greater than or equal to 0 (zero)",
                                                                    MIN_SSTABLE_SIZE_OPTION));
                 int limit = (int) Math.ceil(targetSSTableSize * INVERSE_SQRT_2);
                 if (sizeInBytes >= limit )
-                    throw new ConfigurationException(String.format("Invalid configuration, %s should be less than: %sB",
-                                                                   MIN_SSTABLE_SIZE_OPTION,
+                    throw new ConfigurationException(String.format("Invalid configuration, %s (%sB) should be less than: %sB",
+                                                                   MIN_SSTABLE_SIZE_OPTION, sizeInBytes,
                                                                    limit));
             }
             catch (NumberFormatException e)
--- a/src/java/org/apache/cassandra/db/compaction/unified/Controller.java
+++ b/src/java/org/apache/cassandra/db/compaction/unified/Controller.java
@@ -487,12 +487,14 @@ public class Controller
             }
         }
 
+        // preserve the configuration for later use during min_sstable_size.
+        long targetSSTableSize = DEFAULT_TARGET_SSTABLE_SIZE;
         s = options.remove(TARGET_SSTABLE_SIZE_OPTION);
         if (s != null)
         {
             try
             {
-                long targetSSTableSize = FBUtilities.parseHumanReadableBytes(s);
+                targetSSTableSize = FBUtilities.parseHumanReadableBytes(s);
                 if (targetSSTableSize < MIN_TARGET_SSTABLE_SIZE)
                 {
                     throw new ConfigurationException(String.format("%s %s is not acceptable, size must be at least %s",
@@ -596,10 +598,10 @@ public class Controller
             try
             {
                 long sizeInBytes = FBUtilities.parseHumanReadableBytes(s);
-                if (sizeInBytes < 0)
-                    throw new ConfigurationException(String.format("Invalid configuration, %s should be positive: %s",
+                if (sizeInBytes < targetSSTableSize)
+                    throw new ConfigurationException(String.format("Invalid configuration, %s should be greater than or equal to: %sB",
                                                                    MIN_SSTABLE_SIZE_OPTION,
-                                                                   s));
+                                                                   targetSSTableSize));
             }
             catch (NumberFormatException e)
             {
--- a/src/java/org/apache/cassandra/db/compaction/unified/Controller.java
+++ b/src/java/org/apache/cassandra/db/compaction/unified/Controller.java
@@ -598,10 +598,12 @@ public class Controller
             try
             {
                 long sizeInBytes = FBUtilities.parseHumanReadableBytes(s);
-                if (sizeInBytes < targetSSTableSize)
-                    throw new ConfigurationException(String.format("Invalid configuration, %s should be greater than or equal to: %sB",
+                // zero is a valid option to disable featue
+                int limit = (int) Math.ceil(targetSSTableSize * INVERSE_SQRT_2);
+                if (sizeInBytes != 0 && sizeInBytes < limit)
+                    throw new ConfigurationException(String.format("Invalid configuration, %s should be 0 (zero) or greater than or equal to: %sB",
                                                                    MIN_SSTABLE_SIZE_OPTION,
-                                                                   targetSSTableSize));
+                                                                   limit));
             }
             catch (NumberFormatException e)
             {
--- a/src/java/org/apache/cassandra/db/compaction/unified/Controller.java
+++ b/src/java/org/apache/cassandra/db/compaction/unified/Controller.java
@@ -598,13 +598,13 @@ public class Controller
             try
             {
                 long sizeInBytes = FBUtilities.parseHumanReadableBytes(s);
+                // zero is a valid option to disable featue
                 if (sizeInBytes < 0)
                     throw new ConfigurationException(String.format("Invalid configuration, %s should be greater than 0 (zero)",
                                                                    MIN_SSTABLE_SIZE_OPTION));
-                // zero is a valid option to disable featue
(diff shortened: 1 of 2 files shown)


Claudenw:
@ethan-brown2022 is this what you had in mind?

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -596,10 +598,10 @@ public static Map<String, String> validateOptions(Map<String, String> options) t
             try
             {
                 long sizeInBytes = FBUtilities.parseHumanReadableBytes(s);
-                if (sizeInBytes < 0)
-                    throw new ConfigurationException(String.format("Invalid configuration, %s should be positive: %s",
+                if (sizeInBytes < targetSSTableSize)
```
ethan-brown2022: UCS aims to produce sstables with size between $t / \sqrt 2$ and $t \cdot \sqrt 2$ (More information can be found [here](https://github.com/apache/cassandra/blob/cassandra-5.0/src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md)).  We should verify that `sizeInBytes < targetSSTableSize * INVERSE_SQRT_2`

test/unit/org/apache/cassandra/db/compaction/unified/ControllerTest.java
```diff
@@ -207,11 +209,13 @@ public void testGetNumShards()
         Map<String, String> options = new HashMap<>();
         options.putIfAbsent(Controller.BASE_SHARD_COUNT_OPTION, Integer.toString(3));
         options.putIfAbsent(Controller.TARGET_SSTABLE_SIZE_OPTION, FBUtilities.prettyPrintMemory(100 << 20));
-        options.put(Controller.MIN_SSTABLE_SIZE_OPTION, "0B");
         options.put(Controller.SSTABLE_GROWTH_OPTION, "0.0");
         Controller.validateOptions(options);
+        // this is a provably invalid setting but required for this test.
```
ethan-brown2022: This is a valid setting.  With `min_sstable_size = 0`, the feature is disabled.

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
ethan-brown2022: We should also continue to throw a `ConfigurationException` if `sizeInBytes < 0`.

Claudenw: targetSSTableSize must be greater than zero (see see lines 490- 514) so no additional check for greater than zero is necessary.

Claudenw: Should we check that sizeInBytes <  $t \cdot \sqrt 2$ ?

ethan-brown2022: > UCS aims to produce sstables with size between t/2 and t⋅2 (More information can be found [here](https://github.com/apache/cassandra/blob/cassandra-5.0/src/java/org/apache/cassandra/db/compaction/UnifiedCompactionStrategy.md)). We should verify that `sizeInBytes > targetSSTableSize * INVERSE_SQRT_2`

Sorry, I made a mistake in my previous comment.   This is supposed to say `sizeInBytes < targetSSTableSize * INVERSE_SQRT_2`

`min_sstable_size` should be < `targetSSTableSize * INVERSE_SQRT_2`, so if `sizeInBytes > targetSSTableSize * INVERSE_SQRT_2` we should throw an exception.  In this case, we should continue to make sure that `sizeInBytes` is not < 0.  Sorry for the confusion.

test/unit/org/apache/cassandra/db/compaction/unified/ControllerTest.java
Claudenw: should be fixed in the latest code.

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
Claudenw: @ethan-brown2022 I am thinking there should be 3 checks:

( min_sstable_size <  $t \cdot \sqrt 2$  && min_sstable_size > $t / \sqrt 2$ ) || (min_sstable_size = 0)

We have not discussed adding  $t / \sqrt 2$  as a check.

ethan-brown2022: The purpose of the `min_sstable_size` option is to only split sstables over shards if they are at least as large as `min_sstable_size`.  If it is set, fewer than `base_shard_count` shards will be created to ensure that the estimated size of each sstable is at least `min_sstable_size`, but once there is enough data, the number of shards will be such that the size of each sstable is between $t / \sqrt 2$ and $t \cdot \sqrt 2$ (assuming sstable_growth = 0).  

The goal of this ticket is to verify that `min_sstable_size` is less than the lower bound ($t / \sqrt 2$).

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -598,10 +598,13 @@ public static Map<String, String> validateOptions(Map<String, String> options) t
...
             {
                 long sizeInBytes = FBUtilities.parseHumanReadableBytes(s);
+                if (sizeInBytes < 0)
+                    throw new ConfigurationException(String.format("Invalid configuration, %s should be greater than 0 (zero)",
+                                                                   MIN_SSTABLE_SIZE_OPTION));
                 // zero is a valid option to disable featue
                 int limit = (int) Math.ceil(targetSSTableSize * INVERSE_SQRT_2);
-                if (sizeInBytes != 0 && sizeInBytes < limit)
-                    throw new ConfigurationException(String.format("Invalid configuration, %s should be 0 (zero) or greater than or equal to: %sB",
+                if (sizeInBytes != 0 && sizeInBytes >= limit )
```
ethan-brown2022: nit: `limit` will always be > 0 since `MIN_TARGET_SSTABLE_SIZE` is 1MiB, so checking that `sizeInBytes != 0` here is redundant.

ethan-brown2022 (approved):
LGTM!

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
smiklosovic: @Claudenw  I will start the builds once this is addressed. Thank you.

Claudenw: Should be fixed now

Claudenw:
@smiklosovic I think I fixed the test issue.  Please try again.

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -596,10 +598,15 @@ public static Map<String, String> validateOptions(Map<String, String> options) t
             try
             {
                 long sizeInBytes = FBUtilities.parseHumanReadableBytes(s);
+                // zero is a valid option to disable featue
```
blambov, line 601: Nit: feature

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
```diff
@@ -596,10 +598,15 @@ public static Map<String, String> validateOptions(Map<String, String> options) t
...
                 if (sizeInBytes < 0)
-                    throw new ConfigurationException(String.format("Invalid configuration, %s should be positive: %s",
-                                                                   MIN_SSTABLE_SIZE_OPTION,
-                                                                   s));
+                    throw new ConfigurationException(String.format("Invalid configuration, %s should be greater than or equal to 0 (zero)",
+                                                                   MIN_SSTABLE_SIZE_OPTION));
+                int limit = (int) Math.ceil(targetSSTableSize * INVERSE_SQRT_2);
+                if (sizeInBytes >= limit )
+                    throw new ConfigurationException(String.format("Invalid configuration, %s (%sB) should be less than: %sB",
```
blambov, line 607: Nit: Let's add "the target size minimum" after "less than" to make sure the user knows where that number comes from.

Also, make both values human readable, by applying `FBUtilities.prettyPrintMemory()` (see e.g. the exception on line 500).

blambov (approved):

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
smiklosovic: will be taken care of on commit

src/java/org/apache/cassandra/db/compaction/unified/Controller.java
smiklosovic: will be taken care of on commit

Claudenw: Pretty printing the number causes problems in that the rounding of the numbers means that it will, for some values, print an invalid value.  I had this problem in developing the test numbers.

blambov: What kind of invalid value? Could you give an example?

Claudenw: If you take a value for targetSSTableSize and convert it to bytes, multiply that by INVERSE_SQRT_2, subtract 1, and use FBUtilities.prettyPrintMemory() to convert it to the human readable string.  Then take that string and parse it back to a value the value will often be too large for the minimum value.  There are a range of value where  this will occur.  The issue is that FBUtilities.prettyPrintMemory() seems to round up which makes sense for most of the values we use if for, but in this case we have an upper limit that we do not want to exceed and so we run into problems.

blambov: I understand the problem now, but it's not something I personally would mind with the current phrasing of the exception.

When told e.g. "less than 707.106MiB", wouldn't people try "707.105MiB"?

PR #2969, opened 2023-12-06
CASSANDRA-18857 Allow CQL cert authentication without an AUTHENTICATE request

tolbertam:
If we detect that an Authenticator supports certificate authentication,
we should be able to skip sending an AUTHENTICATE message to the
client if we are able to successfully authenticate with a certificate.

Adds a new interface methods
IAuthenticator.supportsEarlyAuthentication and
SaslNegotiator.shouldSendAuthenticateMessage.

If supportsEarlyAuthentication is true on an authenticator, and
SaslNegotiator.shouldSendAuthenticateMessage evaluates to false,
StartupMessage.execute() will attempt to authenticate early.
If authenticated, a READY response is returned, otherwise ERROR.

Both MutualTlsAuthenticator and
MutualTlsWithPasswordFallbackAuthenticator return true for
supportsEarlyAuthentication.

MutualTlsAuthenticator will always return a SaslNegotiator that
returns false for shouldSendAuthenticateMessage.

MutualTlsWithPasswordFallbackAuthenticator will return a SaslNegotiator
that returns false for shouldSendAuthenticateMessage only if the
connection provides a certificate, otherwise it will use
PlainTextSaslNegotiator, which will cause StartupMessage.execute to
defer authentication by sending an AUTHENTICATE response to the client.

This enables a client to connect with a certificate without having
to configure an authenticator (e.g. a username and password).

Also consolidates authentication handling into AuthUtil.handleLogin.

Finally, adds a series of unit tests utilizing SimpleClient to validate
these authentication flows.

patch by Andy Tolbert; reviewed by Francisco Guerrero Hernandez,
Jon Meredith, Stefan Podkowinski, Jyothsna Konisa and Abe Ratnofsky
for [CASSANDRA-18857](https://issues.apache.org/jira/projects/CASSANDRA/issues/CASSANDRA-18857)


tolbertam:
This is currently marked as draft as it is dependent on [CASSANDRA-18811](https://issues.apache.org/jira/browse/CASSANDRA-18811) and will need a few changes (that i'll indicate in comments of code that will need to change after that is merged.

Also note that `testNoClientCertificatePresented` tests currently fail because of dependence on that feature

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
```diff
@@ -79,7 +79,6 @@ public MutualTlsAuthenticator(Map<String, String> parameters)
         }
         certificateValidator = ParameterizedClass.newInstance(new ParameterizedClass(certificateValidatorClassName),
                                                               Arrays.asList("", AuthConfig.class.getPackage().getName()));
-        checkMtlsConfigurationIsValid(DatabaseDescriptor.getRawConfig());
```
tolbertam: This needs to be reasoned about after rebasing with CASSANDRA-18811 fix.

src/java/org/apache/cassandra/transport/SimpleClientSslContextFactory.java
```diff
@@ -0,0 +1,67 @@
...
+public class SimpleClientSslContextFactory extends FileBasedSslContextFactory
+{
+
+    public SimpleClientSslContextFactory(Map<String, Object> parameters)
+    {
+        super(parameters);
+    }
+
+    @Override
+    public SslContext createNettySslContext(boolean verifyPeerCertificate, SocketType socketType,
```
tolbertam: This needs to be reasoned about after rebasing with CASSANDRA-18811 fix (new signature for consuming verifyPeerCertificate mode).

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -533,6 +535,27 @@ public void setup()
...
+    protected static void requireNativeProtocolClientEncryption()
+    {
+        DatabaseDescriptor.updateNativeProtocolEncryptionOptions((encryptionOptions ->
+                encryptionOptions.withEnabled(true)
+                        .withKeyStore("test/conf/cassandra_ssl_test.keystore")
+                        .withKeyStorePassword("cassandra")
+                        .withTrustStore("test/conf/cassandra_ssl_test.truststore")
+                        .withTrustStorePassword("cassandra")
+                        .withRequireEndpointVerification(false)
+                        .withRequireClientAuth(true)));
```
tolbertam: This needs to be reasoned about after rebasing with CASSANDRA-18811 fix

test/unit/org/apache/cassandra/transport/EarlyCertificateAuthenticationTest.java
```diff
@@ -0,0 +1,195 @@
...
+
+    private EncryptionOptions clientEncryptionOptions(boolean presentClientCertificate)
+    {
+        // To regenerate:
+        // 1. generate keystore
+        //    keytool -genkeypair -keystore test/conf/cassandra_ssl_test_spiffe.keystore -validity 100000 -keyalg RSA -dname "CN=Apache Cassandra, OU=ssl_test, O=Unknown, L=Unknown, ST=Unknown, C=Unknown" -keypass cassandra -storepass cassandra -alias spiffecert -ext SAN=URI:spiffe://test.cassandra.apache.org/unitTest/mtls -storetype jks
+        // 2. export cert
+        //    keytool -export -alias spiffecert -file spiffecert.cer -keystore test/conf/cassandra_ssl_test_spiffe.keystore
+        // 3. import cert into truststore
+        //    keytool -import -v -trustcacerts -alias spiffecert -file spiffecert.cer -keystore test/conf/cassandra_ssl_test.truststore
```
aratno: Thanks for including this!

src/java/org/apache/cassandra/transport/messages/StartupMessage.java
```diff
@@ -118,8 +121,29 @@ else if (compression.equals("lz4"))
...
+                    // Attempt to authenticate the user.
+                    return AuthUtil.handleLogin(connection, state, EMPTY_CLIENT_RESPONSE, (negotiationComplete, challenge) ->
+                    {
+                        if (negotiationComplete) {
+                            // Authentication was successful, proceed.
+                            return new ReadyMessage();
+                        } else {
+                            // It's expected that any negotiator that requires a challenge will likely not support early
+                            // certificate authentication, in this case we can just go through the traditional auth flow.
+                            return new AuthenticateMessage(DatabaseDescriptor.getAuthenticator().getClass().getName());
```
aratno, line 146: Just occurred to me - if we're in the "fallback" part of a client-cert-fallback-to-credentials authenticator, shouldn't the client not need to know that it's a custom authenticator, and just receive an AUTHENTICATE with PasswordAuthenticator? Otherwise, we'd have to make sure all clients are aware of a new authenticator.

Doesn't seem like a new issue with this code, just wondering how we handle this.

aratno (approved):
+1

src/java/org/apache/cassandra/transport/messages/StartupMessage.java
tolbertam: Yeah I think that's right - Most client implementations don't event look at the authenticator class name, e.g. I see that the java driver gets passed this value, but its default authenticator doesn't do anything with it:

https://github.com/apache/cassandra-java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/internal/core/auth/PlainTextAuthProvider.java#L69-L70

aratno: Interesting - might cause problems in the future (or with niche clients), good to know. Thanks!

src/java/org/apache/cassandra/auth/IAuthenticator.java
```diff
@@ -145,5 +167,18 @@ public interface SaslNegotiator
...
+
+        /**
+         * Whether its determined if this negotiator requires certificates.  This is used in conjunction with
+         * {@link #supportsEarlyCertificateAuthentication()} in determining if authentication can be done with
+         * client certificates when handling a Startup message.
+         *
+         * If <code>true</code>, will attempt to authenticate while handling a Startup, otherwise will not attempt
+         * there and defer authentication to when handling an AuthResponse message.
+         */
+        default boolean requiresCertificateAuthentication()
```
jyothsnakonisa: Do we need both `requiresCertificateAuthentication` & `supportsEarlyCertificateAuthentication`? can we not have just one method?

src/java/org/apache/cassandra/auth/MutualTlsWithPasswordFallbackAuthenticator.java
```diff
@@ -46,13 +46,21 @@ public void setup()
         mutualTlsAuthenticator.setup();
     }
 
+    @Override
+    public boolean supportsEarlyCertificateAuthentication()
```
jyothsnakonisa: Incase of client making password based authentication this wold be false, Also I am curious about what values will be returned for `supportsEarlyCertificateAuthentication` & `requiresCertificateAuthentication` in case of fallback authenticator.

src/java/org/apache/cassandra/transport/SimpleClientSslContextFactory.java
```diff
@@ -0,0 +1,68 @@
+/*
```
jyothsnakonisa: If this is only used in test cases can we not move it to test package?

src/java/org/apache/cassandra/transport/messages/StartupMessage.java
tolbertam: yeah - although we didn't change anything here in this PR; it's fully compatible with existing behavior, should behave like it did before :+1:

aratno: Yep, this was all adjacent curiosity. PR LGTM.

jyothsnakonisa (commented):
Looks good overall, left couple of questions.

src/java/org/apache/cassandra/transport/messages/AuthUtil.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.cassandra.auth.AuthenticatedUser;
+import org.apache.cassandra.auth.IAuthenticator;
+import org.apache.cassandra.exceptions.AuthenticationException;
+import org.apache.cassandra.metrics.ClientMetrics;
+import org.apache.cassandra.service.QueryState;
+import org.apache.cassandra.transport.Connection;
+import org.apache.cassandra.transport.Message.Response;
+import org.apache.cassandra.transport.ServerConnection;
+
+public class AuthUtil
```
smiklosovic: @tolbertam you might add a private constructor here to not be able to do an instance of this.

src/java/org/apache/cassandra/transport/messages/StartupMessage.java
```diff
@@ -118,8 +121,29 @@ else if (compression.equals("lz4"))
             clientState.setDriverVersion(options.get(DRIVER_VERSION));
         }
 
-        if (DatabaseDescriptor.getAuthenticator().requireAuthentication())
+        IAuthenticator authenticator = DatabaseDescriptor.getAuthenticator();
+        if (authenticator.requireAuthentication()) {
```
smiklosovic: @tolbertam could you put curly braces on new line as done everywhere else?

src/java/org/apache/cassandra/transport/messages/AuthUtil.java
tolbertam: will do!

src/java/org/apache/cassandra/transport/messages/StartupMessage.java
tolbertam: ah good catch, not sure how I missed this;  will fix!

src/java/org/apache/cassandra/auth/IAuthenticator.java
tolbertam: The main reason for the differentiation is if `requiresCertificateAuthentication` is set to true, we can prevent sending an `AUTHENTICATE` response to the client when a certificate is missing, allowing the error to happen earlier in the process.

If we didn't have this when using `MtlSAuthenticator`, If a client had a missing certificate C* would send an `AUTHENTICATE` back to the client, which would cause a different error flow.  e.g. instead of getting a clear error back from Cassandra (`No certificate present on connection`),  the client gets prompted for an authenticator, which it will likely not have configured and the driver will throw a less specific error for.

src/java/org/apache/cassandra/auth/MutualTlsWithPasswordFallbackAuthenticator.java
tolbertam: We set `supportsEarlyCertificateAuthentication` to true here because if the client provides a certificate, Cassandra will try to verify against it.  If not, it will send an `AUTHENTICATE` message back tot he client, and continue with the usual password-based auth flow.

src/java/org/apache/cassandra/auth/IAuthenticator.java
tolbertam: Another point of differentiation:

`supportsEarlyCertificateAuthentication` is a configuration of `IAuthenticator`.

`requiresCertificateAuthentication` is a configuration of `SaslNegotiator`.  This allows us to contextually dictate whether or not certificate authentication is needed based on the chosen `SaslNegotiator` (as in the fallback authenticators `newSaslNegotiator`.

src/java/org/apache/cassandra/transport/SimpleClientSslContextFactory.java
tolbertam: yeah that's a good point.  I think I might have done this because it was adjacent to `SimpleClient`, but I agree that it should live with tests because I don't think we would want it to be used outside of that context.

src/java/org/apache/cassandra/transport/messages/StartupMessage.java
tolbertam: looked around and corrected this and another place I was not following convention.

Was wondering if it's ever been discussed to enforce this through checkstyle?  I suspect maybe it has and was rejected because of possible chaos to reconcile it;  These rules found ~595 violations but a bunch of them come from the same set of files, so maybe manageable.

```diff
❯ git diff
diff --git a/.build/checkstyle.xml b/.build/checkstyle.xml
index 8b81f21281..9bd22dc1ac 100644
--- a/.build/checkstyle.xml
+++ b/.build/checkstyle.xml
@@ -179,6 +179,10 @@
       <message key="matchxpath.match"
                value="'Deprecated annotation must provide 'since' value."/>
     </module>
+    <module name="LeftCurly">
+        <property name="option" value="nl"/>
+        <property name="tokens" value="ANNOTATION_DEF,CLASS_DEF,CTOR_DEF,ENUM_CONSTANT_DEF,INTERFACE_DEF,LITERAL_CASE,LITERAL_CATCH,LITERAL_DEFAULT,LITERAL_DO,LITERAL_ELSE,LITERAL_FINALLY,LITERAL_FOR,LITERAL_IF,LITERAL_SWITCH,LITERAL_SYNCHRONIZED,LITERAL_TRY,LITERAL_WHILE,METHOD_DEF,OBJBLOCK,STATIC_INIT,RECORD_DEF,COMPACT_CTOR_DEF"/>
+    </module>
   </module>

 </module>
```

tolbertam: https://gist.github.com/tolbertam/2b0f6245c3e94a7809f716d99a4ec9c3 if interested in seeing the violations.  I removed `LAMBDA` as that one was violated quite a bit and somewhat reasonable to allow curly braces on same line there for oneliners.

jyothsnakonisa (approved):
Thanks for addressing feedback, looks good to me

src/java/org/apache/cassandra/transport/messages/StartupMessage.java
tolbertam: caught a couple other places I missed..should be good now that I've run checkstyle against my changeset.

src/java/org/apache/cassandra/transport/messages/StartupMessage.java
```diff
@@ -118,8 +121,34 @@ else if (compression.equals("lz4"))
             clientState.setDriverVersion(options.get(DRIVER_VERSION));
         }
 
-        if (DatabaseDescriptor.getAuthenticator().requireAuthentication())
+        IAuthenticator authenticator = DatabaseDescriptor.getAuthenticator();
+        if (authenticator.requireAuthentication())
+        {
+            // If the authenticator supports early certificate authentication, attempt to authenticate with certificates.
+            if (authenticator.supportsEarlyCertificateAuthentication())
```
frankgh: in the mtls with password auth fallback, we could also authenticate here if client certs are presented . I'm thinking the API could be a bit more generic, thinking of supporting other types of early authentication .

```
if (authenticator.supportsEarlyAuthentication(someContextThatHelpsMakingTheDetermination))
```

frankgh (commented):
Overall looks good, but I've left a question about making the API a bit more generic. Let me know what you think

src/java/org/apache/cassandra/transport/messages/StartupMessage.java
tolbertam: If I'm understanding correctly, I think that is the way it should work with the current state of things:

`MutualTlsWithPasswordFallbackAuthenticator.supportsEarlyCertificateAuthentication()` will always return `true` 

The `SaslNegotiator` returned from `MutualTlsWithPasswordFallbackAuthenticator.newSaslNegotiator` will return an implementation that evaluates to `false` for `requiresCertificateAuthentication()` if no certs are present on the connection, and `true` if certificates are present which will then cause it to attempt to authenticate (line 135 below).

--

I do like the idea of more generically naming this though;  Are there other kinds of authentication that you could envision of that we could do eventually without sending an AUTHENTICATE message?  Other things present on the connection that we could validate with?

I think if we renamed:

1. `IAuthenticator.supportsEarlyCertificateAuthentication` to `IAuthenticator.supportsEarlyAuthentication`
2. `SaslNegotiator.requiresCertificateAuthentication` to `SaslNegotiation.requiresEarlyAuthentication`

That could work?  In `AuthUtil.handleLogin` we pass `connection` and `state`, which should be the level of context needed to do other authentication (unless there is something else we can pass through that would be helpful for authentication)

frankgh: > The SaslNegotiator returned from MutualTlsWithPasswordFallbackAuthenticator.newSaslNegotiator will return an implementation that evaluates to false for requiresCertificateAuthentication() if no certs are present on the connection, and true if certificates are present which will then cause it to attempt to authenticate (line 135 below).

Ah yes, I see that now, it is based on whether the certificates are present. This makes sense

> Are there other kinds of authentication that you could envision of that we could do eventually without sending an AUTHENTICATE message?

I was trying to think of other use cases, but I couldn't come up with anything for now. But I think making the method name generic could open it up for future improvements (maybe). We don't really need to _tie_ the implementation to certificates.

+1 to renaming it as proposed above

tolbertam: agreed!  I'll rename the methods and push a commit imminently

frankgh (approved):
+1 thanks for addressing the comments and updating the javadocs!

src/java/org/apache/cassandra/transport/messages/StartupMessage.java
tolbertam: Pushed a change for this [f115a89](https://github.com/apache/cassandra/pull/2969/commits/f115a89605673463ccfaac600057fceb13917f1b)

I made one slight change to my proposal, instead of 

> SaslNegotiator.requiresCertificateAuthentication to SaslNegotiation.requiresEarlyAuthentication

I removed the method and created `SaslNegotiator.shouldSendAuthenticateMessage()` which indicates whether or not a further AUTHENTICATE message should be sent to the client, I felt that was more clear than my earlier proposal.  I am open to changing that though.  Hopefully this looks ok!

frankgh:
Closed via https://github.com/apache/cassandra/commit/c09d0d929baeaa02f3438313c7979ccf6b4b3c5a

PR #2971, opened 2023-12-07
3.0 Updated the OWASP plugin to 9.0.5 and refactored dependency checks

jacek-lewandowski:
Configured OWASP database to use caching and NVD key. Properties can be configured in ~/.ant/build.properties and defined both the key and the storage location

Allow for incremental dependency checking so that for pre-commit we can check only the modified/added dependencies and not fail the build because a new CVE appeared

The user can specify -Dnightly.build.dir=... to provide its own reference distribution against which the dependencies should be compared in incremental mode

Don't check dependencies if build.xml was not changed (to be implemented differently on the newer branches)

### DIFF
--- a/.build/build-owasp.xml
+++ b/.build/build-owasp.xml
@@ -16,30 +16,50 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   -->
-<project basedir="." name="apache-cassandra-owasp-tasks">
-    <property name="dependency-check.version" value="8.3.1"/>
-    <property name="dependency-check.home" value="${build.dir}/dependency-check-ant-${dependency-check.version}"/>
+<project basedir="." name="apache-cassandra-owasp-tasks"
+         xmlns:unless="ant:unless"
+         xmlns:if="ant:if">
+    <property name="dependency-check.version" value="9.0.5"/>
+    <property name="dependency-check.home" value="${tmp.dir}/dependency-check-ant-${dependency-check.version}"/>
+    <property name="dependency-check.archive.dir" value="${local.repository}/org/owasp/dependency-check-ant/${dependency-check.version}"/>
+    <property name="dependency-check.archive.name" value="dependency-check-ant-${dependency-check.version}-release.zip"/>
+    <property name="dependency-check.report.dir" value="${build.dir}/owasp"/>
 
-    <condition property="is.dependency.check.jar">
-        <available file="${dependency-check.home}/dependency-check-ant/dependency-check-ant.jar" type="file" />
+    <condition property="dependency-check-ant.archive.present">
+        <available file="${dependency-check.archive.dir}/${dependency-check.archive.name}" type="file" />
     </condition>
 
-    <target name="dependency-check-download"
-            depends="build"
+    <target name="-dependency-check-download"
             description="Fetch OWASP Dependency checker"
-            unless="is.dependency.check.jar">
+            unless="dependency-check-ant.archive.present">
 
-        <echo>Downloading OWASP Dependency checks ...</echo>
-
-        <mkdir dir="${dependency-check.home}"/>
-
-        <get src="https://github.com/jeremylong/DependencyCheck/releases/download/v${dependency-check.version}/dependency-check-ant-${dependency-check.version}-release.zip"
-             dest="${dependency-check.home}/dependency-check-ant-${dependency-check.version}-release.zip"/>
-
-        <unzip src="${dependency-check.home}/dependency-check-ant-${dependency-check.version}-release.zip" dest="${dependency-check.home}"/>
+        <mkdir dir="${dependency-check.archive.dir}"/>
+        <get src="https://github.com/jeremylong/DependencyCheck/releases/download/v${dependency-check.version}/${dependency-check.archive.name}"
+             dest="${dependency-check.archive.dir}/${dependency-check.archive.name}" retries="3"/>
     </target>
 
-    <target name="dependency-check" description="Dependency-Check Analysis" depends="dependency-check-download,resolver-dist-lib">
+    <target name="-dependency-check-init" depends="-dependency-check-download">
+        <delete dir="${dependency-check.home}" includeemptydirs="true" failonerror="false"/>
+        <mkdir dir="${dependency-check.home}"/>
+        <unzip src="${dependency-check.archive.dir}/${dependency-check.archive.name}" dest="${dependency-check.home}"/>
+    </target>
+
+    <target name="-run-owasp-scan" description="Dependency-Check Analysis"
+            depends="-dependency-check-init,resolver-dist-lib">
+        <fail unless:set="nvd.api.key">
+            Please set the nvd.api.key property to your NVD API key. It is recommended to put that property into your
+            ~/.ant/build.properties file. You can get your API key from https://nvd.nist.gov/developers/request-an-api-key
+        </fail>
+
+        <echo unless:set="nvd.data.dir">
+            Since the NVD database is pretty large, you should consider storing it in some persistent location to reuse
+            it between builds. You can do that by setting the nvd.data.dir property to a directory of your choice,
+            such as ~/.cache/ant/owasp/database for Linux or ~/Library/Caches/Ant/owasp/database for MacOS.
+            Putting that property into your ~/.ant/build.properties file is recommended.
+        </echo>
+
+        <property name="nvd.data.dir" value="${tmp.dir}/owasp/database"/>
+        <property name="nvd.validity.hours" value="4"/>
 
         <path id="dependency-check.path">
             <fileset dir="${dependency-check.home}/dependency-check-ant/lib">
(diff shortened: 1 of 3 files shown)


build.xml
```diff
@@ -30,6 +30,7 @@
     <property environment="env"/>
     <property file="build.properties" />
     <property file="build.properties.default" />
+    <property file="${user.home}/.ant/build.properties"/>
```
michaelsembwever, line 33: do we need this ?

.build/build-resolver.xml
```diff
@@ -46,7 +46,7 @@
 
     <target name="resolver-init" depends="init,_resolver_download" unless="resolver-ant-tasks.initialized" description="Initialize Resolver ANT Tasks">
 
-        <typedef uri="antlib:org.apache.maven.resolver.ant" resource="org/apache/maven/resolver/ant/antlib.xml" classpathref="resolver-ant-tasks.classpath" />
+        <taskdef uri="antlib:org.apache.maven.resolver.ant" resource="org/apache/maven/resolver/ant/antlib.xml" classpathref="resolver-ant-tasks.classpath" />
```
michaelsembwever: what's this about ?

.build/build-resolver.xml
```diff
@@ -20,8 +20,8 @@
          xmlns:unless="ant:unless">
 
     <!-- details of what version of Resolver ANT Tasks to fetch -->
-    <property name="resolver-ant-tasks.version" value="1.3.0" />
-    <property name="resolver-ant-tasks.local" value="${local.repository}/org/apache/maven/resolver/maven-resolver-ant-tasks/${resolver-ant-tasks.version}/maven-resolver-ant-tasks-${resolver-ant-tasks.version}.jar"/>
+    <property name="resolver-ant-tasks.version" value="1.5.0" />
+    <property name="resolver-ant-tasks.local" value="${local.repository}/org/apache/maven/resolver/maven-resolver-ant-tasks/${resolver-ant-tasks.version}/maven-resolver-ant-tasks-${resolver-ant-tasks.version}-uber.jar"/>
```
michaelsembwever: why we updating the resolver version in this PR ?

build.xml
jacek-lewandowski: yes

.build/build-resolver.xml
jacek-lewandowski: I've updated version of this plugin and `taskdef` was in the instruction; I don't know the difference tbh

.build/build-resolver.xml
jacek-lewandowski: without any particular reason, I can revert if you insist, but on the other hand, there are not much PRs which will be applied to all branches starting from 3.0. And this is actually quite painful, so if we are touching the build I thought that we can do such minor changes.

build.xml
jacek-lewandowski: longer answer - yes, it is ported from newer branches (5.0) and it lets the user to define global properties, like tmp / cache directories or nvd api key

.build/build-owasp.xml
```diff
@@ -16,30 +16,49 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   -->
-<project basedir="." name="apache-cassandra-owasp-tasks">
-    <property name="dependency-check.version" value="8.3.1"/>
-    <property name="dependency-check.home" value="${build.dir}/dependency-check-ant-${dependency-check.version}"/>
+<project basedir="." name="apache-cassandra-owasp-tasks"
+         xmlns:unless="ant:unless">
+    <property name="dependency-check.version" value="9.0.2"/>
+    <property name="dependency-check.home" value="${tmp.dir}/dependency-check-ant-${dependency-check.version}"/>
```
smiklosovic, line 22: I am personally putting everything into `/home/user/.cache/cassandra/...`. If I was about to do that here, by modifying `/home/user/.ant/build.properties`, how would that look like? It would be great if `${tmp.dir}` here was actually some other property, like `${dependency-check.home.root}` or similar, because when I override `${dependency-check.home}` in `build.properties`, I do not know what `${dependency-check.version}` is ... it might change over time. Does this make sense?

build.xml
smiklosovic: Jacek has worktrees per branch, I have a worktree per logical branch (I have cassandra-{3.0,3.11,4.0,4.1,5.0,trunk} directories and then I just switch branches for that particular branch inside). `build.properties` are `.gitignor`-ed so that means that these properties would be "lost" in Jacek's case, but not mine. We also do not want to hardcode paths outside of repo except of some dirs in `tmp.dir` nor anywhere in user's home. Somebody is OK with this, somebody is not ... to each their own. This one solves it globally for everybody how they please.

.build/build-resolver.xml
Mmuzaf: +1 to revert unnecessary changes, to avoid this kind of question from anyone looking at the final commit in the future.

.build/dependency-check-suppressions.xml
```diff
@@ -20,13 +20,6 @@
   https://jeremylong.github.io/DependencyCheck/general/suppression.html
 -->
 <suppressions xmlns="https://jeremylong.github.io/DependencyCheck/dependency-suppression.1.3.xsd">
-    <suppress>
```
Mmuzaf, line 23: The build completes successfully without commented cve suppressions, do we still need them?

```
    <suppress>
        <!--  https://issues.apache.org/jira/browse/CASSANDRA-16150 -->
        <packageUrl regex="true">^pkg:maven/org\.yaml/snakeyaml@.*$</packageUrl>
<!--        <cve>CVE-2023-2251</cve>-->
        <cve>CVE-2022-38752</cve>
        <cve>CVE-2022-38751</cve>
        <cve>CVE-2022-38750</cve>
        <cve>CVE-2022-41854</cve>
        <cve>CVE-2022-25857</cve>
        <cve>CVE-2022-38749</cve>
        <cve>CVE-2022-1471</cve>
<!--        <cve>CVE-2022-3064</cve>-->
<!--        <cve>CVE-2021-4235</cve>-->
        <cve>CVE-2017-18640</cve>
    </suppress>
    
        <!-- https://issues.apache.org/jira/browse/CASSANDRA-15417 -->
    <suppress>
        <packageUrl regex="true">^pkg:maven/io\.netty/netty\-all@.*$</packageUrl>
        <cve>CVE-2019-16869</cve>
        <cve>CVE-2019-20444</cve>
        <cve>CVE-2019-20445</cve>
        <cve>CVE-2020-7238</cve>
        <cve>CVE-2021-21290</cve>
        <cve>CVE-2021-21295</cve>
        <cve>CVE-2021-21409</cve>
        <cve>CVE-2021-37136</cve>
        <cve>CVE-2021-37137</cve>
        <cve>CVE-2021-43797</cve>
        <cve>CVE-2022-24823</cve>
        <cve>CVE-2022-41881</cve>
<!--        <cve>CVE-2022-41915</cve>-->
        <cve>CVE-2023-34462</cve>
        <cve>CVE-2023-44487</cve>
    </suppress>
    
    <suppress>
        <packageUrl regex="true">^pkg:maven/com\.thinkaurelius\.thrift/thrift-server@.*$</packageUrl>
        <cve>CVE-2015-3254</cve>
        <cve>CVE-2016-5397</cve>
<!--        <cve>CVE-2018-1320</cve>-->
<!--        <cve>CVE-2018-11798</cve>-->
        <cve>CVE-2019-0205</cve>
    </suppress>
    
        <suppress>
        <packageUrl regex="true">^pkg:maven/org\.codehaus\.jackson/jackson\-mapper\-asl@.*$</packageUrl>
        <cve>CVE-2017-7525</cve>
<!--        <cve>CVE-2017-15095</cve>-->
<!--        <cve>CVE-2017-17485</cve>-->
<!--        <cve>CVE-2018-5968</cve>-->
<!--        <cve>CVE-2018-14718</cve>-->
<!--        <cve>CVE-2018-1000873</cve>-->
<!--        <cve>CVE-2018-7489</cve>-->
        <cve>CVE-2019-10172</cve>
<!--        <cve>CVE-2019-14540</cve>-->
<!--        <cve>CVE-2019-14893</cve>-->
<!--        <cve>CVE-2019-16335</cve>-->
<!--        <cve>CVE-2019-17267</cve>-->
    </suppress>
```

.build/build-owasp.xml
```diff
@@ -52,35 +72,125 @@
...
-                          reportformat="HTML"
+                          nvdApiKey="${nvd.api.key}"
+                          reportoutputdirectory="${dependency-check.report.dir}"
+                          reportformat="ALL"
                           prettyPrint="true"
-                          cveValidForHours="1"
-                          centralAnalyzerUseCache="false"
-                          nodeAuditAnalyzerUseCache="false"
+                          nvdValidForHours="${nvd.validity.hours}"
+                          centralAnalyzerUseCache="true"
```
Mmuzaf, line 94: This is not only about the depencency-check version upgrade, right? We've introduced caching here, so the JIRA issue and its description should be updated accordingly :-)

.build/build-owasp.xml
```diff
@@ -16,30 +16,50 @@
   ~ See the License for the specific language governing permissions and
   ~ limitations under the License.
   -->
-<project basedir="." name="apache-cassandra-owasp-tasks">
-    <property name="dependency-check.version" value="8.3.1"/>
-    <property name="dependency-check.home" value="${build.dir}/dependency-check-ant-${dependency-check.version}"/>
+<project basedir="." name="apache-cassandra-owasp-tasks"
+         xmlns:unless="ant:unless"
+         xmlns:if="ant:if">
+    <property name="dependency-check.version" value="9.0.2"/>
```
Mmuzaf: Should we use the latest one `9.0.4`?

.build/build-owasp.xml
```diff
@@ -16,30 +16,50 @@
...
+    <property name="dependency-check.home" value="${tmp.dir}/dependency-check-ant-${dependency-check.version}"/>
+    <property name="dependency-check.archive.dir" value="${local.repository}/org/owasp/dependency-check-ant/${dependency-check.version}"/>
```
Mmuzaf: Should we use `${local.repository}/org/apache/cassandra/deps/dependency-check-ant` this is a widely used location of other Cassandra dependencies, so I wonder if we should probably have something similar.

.build/build-owasp.xml
```diff
@@ -16,30 +16,50 @@
...
 
-    <target name="dependency-check-download"
-            depends="build"
+    <target name="-dependency-check-download"
             description="Fetch OWASP Dependency checker"
             unless="is.dependency.check.jar">
 
-        <echo>Downloading OWASP Dependency checks ...</echo>
+        <mkdir dir="${dependency-check.archive.dir}"/>
+        <get src="https://github.com/jeremylong/DependencyCheck/releases/download/v${dependency-check.version}/dependency-check-ant-${dependency-check.version}-release.zip"
```
Mmuzaf: As I can see we have the `dependency-check.archive.name` variable defined, why not use it here as well? :-)

.build/build-owasp.xml
```diff
@@ -52,35 +72,125 @@
...
-            (1) https://github.com/jeremylong/DependencyCheck/issues/2166
-            (2) https://nvd.nist.gov/vuln-metrics/cvss
+            (1) https://nvd.nist.gov/vuln-metrics/cvss
         -->
         <dependency-check projectname="Apache Cassandra"
-                          reportoutputdirectory="${basedir}/build"
-                          reportformat="HTML"
+                          nvdApiKey="${nvd.api.key}"
+                          reportoutputdirectory="${dependency-check.report.dir}"
+                          reportformat="ALL"
```
Mmuzaf, line 91: Do we need to generate all formats? Why HTML is not enough now?

.build/build-owasp.xml
```diff
@@ -52,35 +72,125 @@
...
+        <delete file="${nightly.build.dir}/cassandra.tar"/>
+        <exec executable="bash" failifexecutionfails="true" failonerror="true" logerror="true"
+              dir="${nightly.build.dir}">
+            <arg value="-c"/>
+            <arg value="mv apache-cassandra-*/* .; rm -rf apache-cassandra-*"/>
+        </exec>
+    </target>
+
+    <target name="-determine-dependency-changes">
+        <get src="https://raw.githubusercontent.com/apache/cassandra/cassandra-3.0/build.xml" retries="3"
```
Mmuzaf: The branch is hardcoded, is it correct? Should we determine the branch based on the git command output results?

.build/build-owasp.xml
```diff
@@ -16,30 +16,50 @@
...
+         xmlns:if="ant:if">
+    <property name="dependency-check.version" value="9.0.2"/>
+    <property name="dependency-check.home" value="${tmp.dir}/dependency-check-ant-${dependency-check.version}"/>
+    <property name="dependency-check.archive.dir" value="${local.repository}/org/owasp/dependency-check-ant/${dependency-check.version}"/>
+    <property name="dependency-check.archive.name" value="dependency-check-ant-${dependency-check.version}-release.zip"/>
+    <property name="dependency-check.report.dir" value="${build.dir}/owasp"/>
 
     <condition property="is.dependency.check.jar">
-        <available file="${dependency-check.home}/dependency-check-ant/dependency-check-ant.jar" type="file" />
+        <available file="${dependency-check.archive.dir}/${dependency-check.archive.name}" type="file" />
```
Mmuzaf: This is a bit confusing for me, as the property name `is.dependency.check.jar` seems to refer to a jar file, but the `{dependency-check.archive.name}` refers to a zip archive.

.build/build-owasp.xml
```diff
@@ -16,30 +16,50 @@
...
+        <delete dir="${dependency-check.home}" includeemptydirs="true" failonerror="false"/>
         <mkdir dir="${dependency-check.home}"/>
+        <unzip src="${dependency-check.archive.dir}/${dependency-check.archive.name}" dest="${dependency-check.home}"/>
+    </target>
 
-        <get src="https://github.com/jeremylong/DependencyCheck/releases/download/v${dependency-check.version}/dependency-check-ant-${dependency-check.version}-release.zip"
-             dest="${dependency-check.home}/dependency-check-ant-${dependency-check.version}-release.zip"/>
+    <target name="-run-owasp-scan" description="Dependency-Check Analysis"
+            depends="-dependency-check-init,resolver-dist-lib">
+        <echo unless:set="nvd.api.key">
```
Mmuzaf: I think we should fail the build if it's not defined.

.build/build-owasp.xml
jacek-lewandowski: no in this case, we want to compare with the upstream branch thus this will be different in each PR

.build/build-owasp.xml
jacek-lewandowski: HTML is for you to navigate the results easily, JUNIT is for CI to show the results just like other tests results, JSON is for future license scan

.build/build-owasp.xml
jacek-lewandowski: when I implemented that the newest was 9.0.3 and it was broken

.build/dependency-check-suppressions.xml
jacek-lewandowski: how did you find that? I expected the plugin to output those exclusions which are stale but it did not - if you spend time to manually identify those, there is no reason to not apply that :)

.build/build-resolver.xml
jacek-lewandowski: this is discouraging, but ok :)

.build/build-owasp.xml
jacek-lewandowski: It is also widely criticized by me😄 Why don't put dependency-check-ant into the right location of the local repository? Look into Maven central repo: https://search.maven.org/artifact/org.owasp/dependency-check-ant - org/owasp/dependency-check-ant seems the right location for this artifact, isn't it?

.build/build-owasp.xml
jacek-lewandowski: actually, now it is even 9.0.5

.build/build-owasp.xml
jacek-lewandowski: I didn't want to be that strict, but ok

.build/dependency-check-suppressions.xml
Mmuzaf: I've checked all of them locally :-) nice that we can reduce the changes

.build/build-owasp.xml
Mmuzaf: Thanks for the maven link, I should have checked it by myself. +1 then

.build/build-owasp.xml
Mmuzaf: It fails anyway without that key.

Mmuzaf (approved):
+1 as it is

Mmuzaf (approved):

PR #2972, opened 2023-12-07
Cassandra 18902 4.1

jacek-lewandowski:


src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -344,6 +346,11 @@ public enum CassandraRelevantProperties
      */
     GOSSIP_SETTLE_POLL_SUCCESSES_REQUIRED("cassandra.gossip_settle_poll_success_required", "3"),
 
+    CONSISTENT_RANGE_MOVEMENT("cassandra.consistent.rangemovement", "true"),
+    CONSISTENT_SIMULTANEOUS_MOVES_ALLOW("cassandra.consistent.simultaneousmoves.allow"),
+    REPLACE_ADDRESS("cassandra.replace_address"),
+    BROADCAST_INTERVAL_MS("cassandra.broadcast_interval_ms", "60000"),
```
ekaterinadimitrova2: I believe all those are moved to this class already in 5.0+ and they will appear in the VTs. I think we should not move them in older branches now. Also, you would have had to change how they are set/get in more places. Result of a quick grep:
```
grep -r "cassandra.consistent.rangemovement" *
NEWS.txt:     you can set the following property (-Dcassandra.consistent.rangemovement=false)
Binary file build/test/classes/org/apache/cassandra/distributed/test/PartitionDenylistTest.class matches
Binary file build/classes/main/org/apache/cassandra/config/CassandraRelevantProperties.class matches
Binary file build/classes/main/org/apache/cassandra/service/StorageService.class matches
doc/modules/cassandra/pages/operating/topo_changes.adoc:`-Dcassandra.consistent.rangemovement=false`.
src/java/org/apache/cassandra/config/CassandraRelevantProperties.java:    CONSISTENT_RANGE_MOVEMENT("cassandra.consistent.rangemovement", "true"),
src/java/org/apache/cassandra/service/StorageService.java:    public static final boolean useStrictConsistency = Boolean.parseBoolean(System.getProperty("cassandra.consistent.rangemovement", "true"));
src/java/org/apache/cassandra/service/StorageService.java:                    throw new UnsupportedOperationException("Other bootstrapping/leaving/moving nodes detected, cannot bootstrap while cassandra.consistent.rangemovement is true");
src/java/org/apache/cassandra/service/StorageService.java:            throw new UnsupportedOperationException(String.format("Other bootstrapping/leaving/moving nodes detected, cannot bootstrap while cassandra.consistent.rangemovement is true. Nodes detected, bootstrapping: %s; leaving: %s; moving: %s;", bootstrapTokens, leavingTokens, movingTokens));
test/distributed/org/apache/cassandra/distributed/test/PartitionDenylistTest.java:        System.setProperty("cassandra.consistent.rangemovement", "false");

```

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
```diff
@@ -548,8 +551,16 @@ private Future<Void> scheduleSchemaPull(InetAddressAndPort endpoint, VersionInfo
 
         if (shouldPullImmediately(endpoint, info.version))
         {
-            logger.debug("Pulling {} immediately from {}", info, endpoint);
-            submitToMigrationIfNotShutdown(task);
+            if (lastPullFailures.contains(endpoint))
+            {
+                logger.debug("Pulling {} immediately from {} with backoff interval = {}", info, endpoint, PULL_BACKOFF_INTERVAL_MS);
+                ScheduledExecutors.nonPeriodicTasks.schedule(() -> submitToMigrationIfNotShutdown(task), PULL_BACKOFF_INTERVAL_MS, TimeUnit.MILLISECONDS);
```
ekaterinadimitrova2, line 556: And if we happen to run into constant failures for specific endpoint, we will just run into the same issue as we do today, right? I just hope people do not abuse extending this timeout too much

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
```diff
@@ -589,12 +592,20 @@ public void startup(ICluster cluster)
                 // org.apache.cassandra.distributed.impl.AbstractCluster.startup sets the exception handler for the thread
                 // so extract it to populate ExecutorFactory.Global
                 ExecutorFactory.Global.tryUnsafeSet(new ExecutorFactory.Default(Thread.currentThread().getContextClassLoader(), null, Thread.getDefaultUncaughtExceptionHandler()));
+
+                assert !FBUtilities.getReleaseVersionString().equals(FBUtilities.UNKNOWN_RELEASE_VERSION) : "Unknown version";
+                assert !FBUtilities.getReleaseVersionString().isEmpty() : "Empty version";
+                assert FBUtilities.getReleaseVersionString().contains(".") : "Invalid version: " + FBUtilities.getReleaseVersionString();
```
ekaterinadimitrova2, line 596: Isn't this the case when it is `UNKNOWN_RELEASE_VERSION`? (the last assertion)

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
```diff
@@ -589,12 +592,20 @@ public void startup(ICluster cluster)
...
                     // TODO: hacky
-                    System.setProperty("cassandra.ring_delay_ms", "15000");
-                    System.setProperty("cassandra.consistent.rangemovement", "false");
-                    System.setProperty("cassandra.consistent.simultaneousmoves.allow", "true");
+                    if (!RING_DELAY.isPresent())
+                        RING_DELAY.setLong(15000);
+                    if (!CONSISTENT_RANGE_MOVEMENT.isPresent())
+                        CONSISTENT_RANGE_MOVEMENT.setBoolean(false);
+                    if (!CONSISTENT_SIMULTANEOUS_MOVES_ALLOW.isPresent())
+                        CONSISTENT_SIMULTANEOUS_MOVES_ALLOW.setBoolean(true);
```
ekaterinadimitrova2: This is applied after any test configuration is loaded (setUp). Isn't this a bug we need to fix in previous branches? Or at least 4.0, which will still be supported for some time? - I responded to my question - no, those properties are not presented there. I also do not see other options being set on startup, at least in 4.1 that can crash if any test `setUp` method is setting. Please check that nothing has changed in the other newer branches when preparing them.

test/distributed/org/apache/cassandra/distributed/test/MigrationCoordinatorTest.java
```diff
@@ -30,22 +30,26 @@
...
 
 public class MigrationCoordinatorTest extends TestBaseImpl
 {
-
     @Before
     public void setUp()
     {
-        System.clearProperty("cassandra.replace_address");
-        System.clearProperty("cassandra.consistent.rangemovement");
+        REPLACE_ADDRESS.clearValue();
```
ekaterinadimitrova2: Isn't it clearer if we add `@After` for the cleaning?

test/distributed/org/apache/cassandra/distributed/test/MigrationCoordinatorTest.java
```diff
@@ -116,7 +119,6 @@ public void explicitVersionIgnore() throws Throwable
             IInstanceConfig config = cluster.newInstanceConfig();
             config.set("auto_bootstrap", true);
             IGNORED_SCHEMA_CHECK_VERSIONS.setString(initialVersion.toString() + ',' + oldVersion);
-            System.setProperty("cassandra.consistent.rangemovement", "false");
```
ekaterinadimitrova2, line 119: Why was this one removed?

test/unit/org/apache/cassandra/schema/MigrationCoordinatorTest.java
```diff
@@ -96,6 +100,8 @@ public class MigrationCoordinatorTest
             throw new AssertionError(e);
         }
 
+        CassandraRelevantProperties.SCHEMA_PULL_BACKOFF_INTERVAL_MS.setLong(0);
```
ekaterinadimitrova2: So this is optimization not to use the default value. as it does not make sense in these tests?

ekaterinadimitrova2 (commented):
First pass done, thank you for the patch

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
jacek-lewandowski: Yes, all of them are present in 5.0. I just followed what we have today in 4.1 - a bit of mess, variables are defined but not used consistently. I can revert if you like

src/java/org/apache/cassandra/schema/MigrationCoordinator.java
jacek-lewandowski: It is not documented - it is a property just to make tests able to modify - I can add "test" infix to the property name

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
jacek-lewandowski: yes, it is that case. However, we don't expect version to be unknown when starting the node - we could expect that for the version obtained through the Gossiper but in this case we get that version from property file and if it is unknown or missing, something wrong is going on and the node will fail anyway.

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
jacek-lewandowski: yup, will do that (as you can see, those nice methods, `isPresent, setXXX`, are the reasons why I prefer to use `CassandraRelevantProperties`, even if it is inconsistent or enforced

test/distributed/org/apache/cassandra/distributed/test/MigrationCoordinatorTest.java
jacek-lewandowski: Yup

test/distributed/org/apache/cassandra/distributed/test/MigrationCoordinatorTest.java
jacek-lewandowski: Because it is already set in the node initialization, see https://github.com/apache/cassandra/pull/2972/files#diff-321b52220c5bd0aaadf275a845143eb208c889c2696ba0d48a5fc880551131d8L596

test/unit/org/apache/cassandra/schema/MigrationCoordinatorTest.java
jacek-lewandowski: This test is very specific - it investigates different behaviours of `MigrationCoordinator` and sending subsequent requests in particular. We have it under control so we can set it to 0 to run it faster. 

In other tests, we want that backoff because resending a request is a part of normal operation and we don't want to flood the logs

jacek-lewandowski: TBH, this is not a big deal - I can remove this property entirely - it just makes this https://github.com/apache/cassandra/pull/2972/files#diff-019068c4b04d4526506dae4c4a43dfde22aac4d1d74aa317cac755ae519ffe3aR347 wait shorter by one second

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
ekaterinadimitrova2: ` I just followed what we have today in 4.1 - a bit of mess, variables are defined but not used consistently.`
The story is different. The original ticket (CASSANDRA-15876) where `CassandraRelevantProperties` was created added only those system properties already added in the VT. Which other properties will be added to the class and exposed to the VT was a matter of a follow-up ticket (CASSANDRA-17797). The batch of added properties was used around the codebase (in case someone didn't miss anything)

src/java/org/apache/cassandra/utils/FBUtilities.java
```diff
@@ -435,8 +434,14 @@ public static String getReleaseVersionString()
         {
             JVMStabilityInspector.inspectThrowable(e);
             logger.warn("Unable to load version.properties", e);
-            return "debug version";
```
ekaterinadimitrova2, line 438: This is a change of behavior - before, we were returning "`debug version`" here, now we will return null which leads to `UNKNOWN_RELEASE_VERSION`.
Also, this is not a bug fix but an improvement, so we should not commit it to 4.1, but only trunk.

jacek-lewandowski: "debug_version" is nowhere referred. Its history dates back to 2010 and Cassandra 0.7!!! Non-null value was introduced to resolve some problems in IDE (https://issues.apache.org/jira/browse/CASSANDRA-1272). There is no way that we would run Cassandra without version.properties or in case it is unreadable. Also moving this method may be the actual fix because it eliminates to code path which led to the problem.

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1236/workflows/4d8aba5a-1e31-4c92-a5f1-a60c99870089

https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1236/workflows/17ba8442-7e7a-4c8d-a901-9aa90893e5f0

All good except one failure of bootstrap test which is known to be flaky

PR #2976, opened 2023-12-08
CASSANDRA-19182: IR may leak SSTables with pending repair when coming from streaming

dcapwell:


src/java/org/apache/cassandra/db/compaction/CompactionStrategyManager.java
```diff
@@ -749,8 +763,16 @@ else if (i < a.length)
      */
     private void handleFlushNotification(Iterable<SSTableReader> added)
     {
-        for (SSTableReader sstable : added)
-            compactionStrategyFor(sstable).addSSTable(sstable);
+        List<GroupedSSTableContainer> groups = groupSSTables(added);
```
dcapwell: this is the actual bug fix.  I didn't want to add a new method to do `find + add` using the correct locking when we already had... just requires grouping.

If there is a reason to avoid the grouping, more than glad to create a new method to solve this

src/java/org/apache/cassandra/db/compaction/PendingRepairManager.java
```diff
@@ -465,7 +465,7 @@ public boolean hasStrategy(AbstractCompactionStrategy strategy)
 
     public synchronized boolean hasDataForSession(TimeUUID sessionID)
     {
-        return strategies.keySet().contains(sessionID);
+        return strategies.containsKey(sessionID);
```
dcapwell: code cleanup, not correctness issue... just saw while reviewing the code

src/java/org/apache/cassandra/db/compaction/PendingRepairManager.java
```diff
@@ -483,6 +483,14 @@ public Collection<AbstractCompactionTask> createUserDefinedTasks(Collection<SSTa
         return group.entrySet().stream().map(g -> strategies.get(g.getKey()).getUserDefinedTask(g.getValue(), gcBefore)).collect(Collectors.toList());
     }
 
+    public boolean hasPendingRepairSSTable(TimeUUID sessionID, SSTableReader sstable)
```
dcapwell: this isn't thread-safe, but could make it sync to make sure it is.... I only added for a test...

src/java/org/apache/cassandra/db/lifecycle/Tracker.java
```diff
@@ -508,6 +508,8 @@ void notifyAdded(Iterable<SSTableReader> added, boolean isInitialSSTables)
 
     public void notifySSTableRepairedStatusChanged(Collection<SSTableReader> repairStatusesChanged)
     {
+        if (repairStatusesChanged.isEmpty())
```
dcapwell: kept finding we kept triggering notifications constantly with empty state, which caused us to lock a lot but do nothing... so I can remove this, but figured best to not notify when there is nothing to notify of...

src/java/org/apache/cassandra/repair/RepairCoordinator.java
```diff
@@ -209,6 +213,17 @@ private void fail(String reason)
             reason = error != null ? error.toString() : "Some repair failed";
         }
         state.phase.fail(reason);
+        ParticipateState p = ctx.repair().participate(state.id);
```
dcapwell: when debugging the issue I saw the the vtable knew of the repair, but not the result.  I added the plumbing in so we know success/failure in more cases.

`FailSession` gets used even though its only for updating the vtable for non-ir

src/java/org/apache/cassandra/db/compaction/PendingRepairManager.java
bdeggleston: could you add  `@VisibleForTesting` for this and the intermediate method in `PendingRepairHolder`? If it's not threadsafe, I'd also add `unsafe` to the method names, along with an explanation of why it's unsafe

src/java/org/apache/cassandra/db/compaction/CompactionStrategyManager.java
bdeggleston: it seems like the intent here is to add all sstables with a single lock acquisition from PendingRepairManager, but I'm having trouble seeing why that matters and/or how not doing that can put compaction/repair in a bad state. It  might race with a call to `PendingRepairManager#removeSessionIfEmpty`, but the next call to addSStable will add a strategy for the repair id

bdeggleston (changes requested):

src/java/org/apache/cassandra/db/compaction/CompactionStrategyManager.java
bdeggleston: spoke w/ David offline, the previous code fetched the compaction strategy and added to it with 2 lock acquisitions, and the strategy it fetched was getting removed by another operation in the meantime

src/java/org/apache/cassandra/db/compaction/PendingRepairManager.java
dcapwell: added `@VisibleForTesting` and `synchronized`... so its now thread safe (with add/remove sstable), but also document that its for tests

bdeggleston (approved):
+1

bdeggleston (approved):
+1

bdeggleston:
oops, wrong PR!

pmcfadin:
Closing as [CASSANDRA-19182](https://issues.apache.org/jira/browse/CASSANDRA-19182) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #2977, opened 2023-12-09
CASSANDRA-19104 Standardize tablestats formatting and data units

zaaath:
Standardizes space and memory formatting in `nodetool tablestats` human-readable output.

### Before
Nodetool tablestats returns a mix of "bytes", "KiB/MiB/GiB", or no units at all.
Example:
```
        Space used (live): 1463210998523 bytes
        Space used (total): 1463210998523 bytes
        Space used by snapshots (total): 0
        Off heap memory used (total): 607419608 bytes
...
        Memtable data size: 248539982 bytes
        Memtable off heap memory used: 0 bytes
...
        Bytes repaired: 0.000KiB
        Bytes unrepaired: 4315.386GiB
        Bytes pending repair: 0.000KiB
...
        Bloom filter space used: 33590024 bytes
        Bloom filter off heap memory used: 33532560 bytes
        Index summary off heap memory used: 8174024 bytes
        Compression metadata off heap memory used: 565713024 bytes
        Compacted partition minimum bytes: 36
        Compacted partition maximum bytes: 17797419593
        Compacted partition mean bytes: 189740
```

### After
Nodetool tablestats formats space and memory numbers using KiB/MiB/GiB units.
Example:
```
        Space used (live): 1.33 TiB
        Space used (total): 1.33 TiB
        Space used by snapshots (total): 0 bytes
        Off heap memory used (total): 579.28 MiB
...
        Memtable data size: 237.03 MiB
        Memtable off heap memory used: 0 bytes
...
        Bytes repaired: 0 bytes
        Bytes unrepaired: 4.21 TiB
        Bytes pending repair: 0 bytes
...
        Bloom filter space used: 32.03 MiB
        Bloom filter off heap memory used: 31.98 MiB
        Index summary off heap memory used: 7.8 MiB
        Compression metadata off heap memory used: 539.51 MiB
        Compacted partition minimum bytes: 36 bytes
        Compacted partition maximum bytes: 16.58 GiB
        Compacted partition mean bytes: 185.29 KiB
```

### How to reproduce
```bash
nodetool tablestats -H
```

Tests
```bash
ant testsome -Dtest.name=org.apache.cassandra.utils.FBUtilitiesTest
```

patch by @zaaath; reviewed by @smiklosovic for CASSANDRA-19104


src/java/org/apache/cassandra/utils/FBUtilities.java
```diff
@@ -939,6 +939,33 @@ public static String prettyPrintMemory(long size, String separator)
                                  UNIT_PREFIXES.charAt(UNIT_PREFIXES_BASE + prefixIndex));
     }
 
+    /**
+     * Convert the given size in bytes to a human-readable value using binary (i.e. 2^10-based) modifiers
+     * with only three significant digits.
+     * For example, "4.96 KiB", "48.8 KiB", "477 MiB", "0.422 KiB", "0.000 KiB".
+     * @param size      Number to convert.
+     */
+    public static String prettyPrintMemoryShort(long size)
```
smiklosovic: can you just walk us through this method conceptually? It is a little bit hard to follow to be honest.

zaaath: yeah, it is hard to follow. I tried my best to streamline the logic. Let me know if there is more I can do.

This is actually similar to the existing `FBUtilities#prettyPrintMemory` method with some additional logic for keeping only the first three significant digits.

"Line-by-line" explanation:
1. Calculate Prefix Index: `int prefixIndex = (63 - Long.numberOfLeadingZeros(Math.abs(size))) / 10;` determines the appropriate prefix (like Kilo, Mega, Giga, etc.) for the given size. It calculates how many bits are needed to represent the absolute value of `size`, then divides this by 10 to find the index in the `UNIT_PREFIXES` string.
2. Adjust for Smallest Prefix: `if (prefixIndex == 0) prefixIndex = 1;` ensures that for sizes less than 1024 bytes (where `prefixIndex` would be 0), the 'Kilo' prefix is used instead.
3. Get Prefix Character: `char prefix = UNIT_PREFIXES.charAt(UNIT_PREFIXES_BASE + prefixIndex);` retrieves the corresponding unit prefix character from UNIT_PREFIXES string based on the prefixIndex.
4. Scale the Size: `float scaledSize = Math.scalb(size, -prefixIndex * 10);` scales the size down by the factor determined by the prefix index. This is effectively dividing the size by 2^(10×prefixIndex).
5. Format Scaled Size: The block of code after `String scaledSizeStr;` formats the scaled size to have only three significant digits. It checks if the scaled size is between -1 and 1 (not inclusive) and simply calls `String.format` on it; otherwise, it calculates how many decimal places to include to ensure only three significant digits, after which calls `String.format` on it.
6. Merge Scaled Size with Prefix: `return String.format` constructs the final string representing the size in a human-readable format, appending the appropriate unit prefix and the suffix 'iB' for binary unit.

zaaath: Flow chart of the above:
```mermaid
flowchart TD;
    A[Calculate Prefix Index] -->|Determine prefix index based on size| B[Adjust for Smallest Prefix]
    B -->|Set to 'Kilo' for sizes < 1024 bytes| C[Get Prefix Character]
    C -->|Retrieve unit prefix character| D[Scale the Size]
    D -->|Divide size by 2 to the power of 10 * prefixIndex | E[Format Scaled Size]
    E -->|Is scaled size between -1 and 1?| F[Format Small Numbers]
    E -->|Format Larger Numbers| G[Format Large Numbers]
    F & G --> H[Merge Scaled Size with Prefix]
    H -->|Construct final human-readable string| I[End]
```

zaaath: A playground on replit that implements this logic and tests it: https://replit.com/@tytakoff/CASSANDRA-19104#Main.java

smiklosovic: @zaaath can't we just use ` org.apache.cassandra.io.util.FileUtils.stringifyFileSize`?

zaaath: @smiklosovic thanks for the code pointer, I like the output that `org.apache.cassandra.io.util.FileUtils.stringifyFileSize`  produces, but please note that it doesn't quite match what was agreed on in the Jira and in the mailing list:
1. "KiB might make sense as the lowest reportable units."
2. "With a leading zero before the decimal point, there should be three significant digits afterwards"

Comparison:
```
Expected output and FBUtilities.prettyPrintMemoryShort:
0: 0.000 KiB
5: 0.005 KiB
50: 0.049 KiB
500: 0.488 KiB
5000: 4.88 KiB
50000: 48.8 KiB
500000: 488 KiB
5000000: 4.77 MiB
50000000: 47.7 MiB
500000000: 477 MiB

FileUtils.stringifyFileSize:
0: 0 bytes
5: 5 bytes
50: 50 bytes
500: 500 bytes
5000: 4.88 KiB
50000: 48.83 KiB
500000: 488.28 KiB
5000000: 4.77 MiB
50000000: 47.68 MiB
500000000: 476.84 MiB
```

I'll switch the logic to `FileUtils.stringifyFileSize` if you think it's warranted to not fully follow the expected output format. I'm on board with this.

bschoening:
Using the existing stringifyFileSize seems like the right approach.

zaaath:
@bschoening thanks, I totally agree that using the existing stringifyFileSize method is the right approach.

src/java/org/apache/cassandra/utils/FBUtilities.java
zaaath: @smiklosovic I migrated the code to `org.apache.cassandra.io.util.FileUtils.stringifyFileSize` as you suggested. Let me know how it looks.

src/java/org/apache/cassandra/utils/FBUtilities.java
```diff
@@ -921,6 +911,16 @@ public static String prettyPrintAverage(double average)
         return String.format("%.2f", average);
     }
 
+    /**
+     * Convert the given size in bytes to a human-readable value using binary (i.e. 2^10-based) modifiers.
+     * For example, 1.000KiB, 2.100GiB etc., up to 8.000 EiB.
+     * @param size      Number to convert.
+     */
+    public static String prettyPrintMemory(long size)
```
zaaath, line 929: This method has been moved closer to `prettyPrintMemory(long size, String separator)`

test/unit/org/apache/cassandra/io/util/FileUtilsTest.java
```diff
@@ -75,6 +75,35 @@ public void testParseFileSize() throws Exception
             6621259022467L, FileUtils.parseFileSize("6.022 TiB"));
     }
 
+    @Test
+    public void testStringifyFileSize() throws Exception
```
zaaath, line 79: There were no tests for `FileUtils.stringifyFileSize`, so I had to add these.

src/java/org/apache/cassandra/tools/nodetool/stats/TableStatsPrinter.java
```diff
@@ -167,6 +168,11 @@ protected void printStatsTable(StatsTable table, String tableDisplayName, String
             }
             out.println("");
         }
+
+        protected String formatMemory(long bytes, boolean humanReadable)
```
smiklosovic: @zaaath I think this should be called `formatDataSize` or similar, rather than _memory_.

Also, why does this method need to be `protected`? It can be `private`. Whole `DefaultPrinter` is private and that method is not used anywhere but in `printStatsTable`

smiklosovic: Also, could you please take a look at this? (this is an opportunistic fix here)

`out.printf(indent + "  %-" + maxWidth + "s %s%n", "Key", "Size");` 

my IDE tells me that this should be used with all `%s` rather than using concatenating. How is it now does not make a lot of sense if we use `+` together with `%s`. Should be all on `%s`. There is similar case with 

`out.printf(indent + "  %-" + maxWidth + "s %s%n", "Key", "Count");` 

below.

zaaath: @smiklosovic good point, I renamed it to `formatDataSize` and made it private.

Regarding `out.printf(indent...`, it looks weird and convoluted, but it actually makes sense. `maxWidth` sets the width of the "Key" column with left alignment which is what `%-{maxWidth}` is for. For example, let's assume that `index = ""` (empty) and `maxWidth = 10`, then the line becomes:
`out.printf("  %-10s %s%n", "Key", "Size");`
which prints:
`  Key        Size`
where the distance between "K" and "S" is `maxWidth+1=10+1=11` characters.

If instead you were pointing out that `indent +` should be replaced with `%s`, I understand where you are coming from. I think that `indent +` is more readable, but if you want to refactor it, let's do it in another PR since `indent +` is all over `TableStatsPrinter.DefaultPrinter#printStatsTable`

PR #2981, opened 2023-12-10
Fix timeout for blocking read repair when there is a pending node

Runtian:
```
Fix timeout for blocking read repair when there is a pending node

```

The [Cassandra Jira]([CASSANDRA-19120](https://issues.apache.org/jira/browse/CASSANDRA-19120))

### DIFF
--- a/src/java/org/apache/cassandra/locator/ReplicaPlans.java
+++ b/src/java/org/apache/cassandra/locator/ReplicaPlans.java
@@ -19,7 +19,6 @@
 package org.apache.cassandra.locator;
 
 import com.carrotsearch.hppc.ObjectIntHashMap;
-import com.carrotsearch.hppc.cursors.ObjectIntCursor;
 import com.carrotsearch.hppc.cursors.ObjectObjectCursor;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ArrayListMultimap;
@@ -439,6 +438,7 @@ public class ReplicaPlans
     {
         return new Selector()
         {
+
             @Override
             public <E extends Endpoints<E>, L extends ReplicaLayout.ForWrite<E>>
             E select(ConsistencyLevel consistencyLevel, L liveAndDown, L live)
@@ -455,7 +455,8 @@ public class ReplicaPlans
                     int add = consistencyLevel.blockForWrite(liveAndDown.replicationStrategy(), liveAndDown.pending()) - contacts.size();
                     if (add > 0)
                     {
-                        for (Replica replica : filter(live.all(), r -> !contacts.contains(r)))
+                        E all = consistencyLevel.isDatacenterLocal() ? live.all().filter(InOurDcTester.replicas()) : live.all();
+                        for (Replica replica : filter(all, r -> !contacts.contains(r)))
                         {
                             contacts.add(replica);
                             if (--add == 0)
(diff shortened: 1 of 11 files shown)


src/java/org/apache/cassandra/locator/ReplicaPlans.java
```diff
@@ -463,7 +455,7 @@ E select(ConsistencyLevel consistencyLevel, L liveAndDown, L live)
                     int add = consistencyLevel.blockForWrite(liveAndDown.replicationStrategy(), liveAndDown.pending()) - contacts.size();
                     if (add > 0)
                     {
-                        E all = consistencyLevel == ConsistencyLevel.LOCAL_QUORUM ? live.all().sorted(replicasInDcComparator) : live.all();
+                        E all = consistencyLevel.isDatacenterLocal() ? live.all().filter(InOurDcTester.replicas()) : live.all();
```
smiklosovic, line 458: @Runtian 

Blake said "So after the removal of read_repair_chance/dclocal_read_repair_chance in [CASSANDRA-13910](https://issues.apache.org/jira/browse/CASSANDRA-13910), there should never be a case where a DC local read or read repairs are talking to remote DCs."

So I think this is wrong, no? We should just do `live.all().filter(InOurDcTester.replicas())`

src/java/org/apache/cassandra/service/reads/repair/BlockingPartitionRepair.java
```diff
@@ -205,7 +199,9 @@ public void maybeSendAdditionalWrites(long timeout, TimeUnit timeoutUnit)
         if (awaitRepairsUntil(timeout + timeoutUnit.convert(mutationsSentTime, TimeUnit.NANOSECONDS), timeoutUnit))
             return;
 
-        EndpointsForToken newCandidates = writePlan.liveUncontacted();
+        EndpointsForToken newCandidates = writePlan.liveUncontacted().filter(
```
bdeggleston: can you rework this bit as:
`writePlan.consistencyLevel().isDatacenterLocal() ? writePlane.liveUncontacted().filter(InOutDcTester.replicas()) : writePlane.liveUncontacted();` so we're only filtering if we need to?

src/java/org/apache/cassandra/service/reads/repair/BlockingPartitionRepair.java
```diff
@@ -61,31 +60,32 @@ public class BlockingPartitionRepair
...
     {
         this.key = key;
         this.pendingRepairs = new ConcurrentHashMap<>(repairs);
         this.writePlan = writePlan;
-        this.shouldBlockOn = shouldBlockOn;
+
+        // make sure all the read repair targets are contact of the repair write plan
+        if (!all(repairs.keySet(), (r) -> writePlan.contacts().contains(r)))
+        {
+            throw new AssertionError("All repair targets should be part of contacts of read repair write plan.");
```
bdeggleston: Can you use `Preconditions.checkState` here, or throw an `IllegalStateException`?

src/java/org/apache/cassandra/service/reads/repair/BlockingPartitionRepair.java
```diff
@@ -61,31 +60,32 @@ public class BlockingPartitionRepair
...
         {
-            // remote dcs can sometimes get involved in dc-local reads. We want to repair
-            // them if they do, but they shouldn't interfere with blocking the client read.
-            if (!repairs.containsKey(participant) && shouldBlockOn.test(participant.endpoint()))
+            if (!repairs.containsKey(participant))
                 blockFor--;
+
+            // make sure for local consistency, all contacts are local replicas
+            if (writePlan.consistencyLevel().isDatacenterLocal() && !InOurDcTester.replicas().test(participant))
+                throw new AssertionError("Local consistency blocking read repair is trying to contact remote DC node: " + participant.endpoint());
```
bdeggleston: Can you use `Preconditions.checkState` here, or throw an `IllegalStateException`?

bdeggleston (changes requested):

PR #2982, opened 2023-12-11
Pre-requisite changes for CASSANDRA-18888

iamaleksey:
patch by Aleksey Yeschenko; reviewed by Ariel Weisberg for CASSANDRA-18888

### DIFF
--- a/src/java/org/apache/cassandra/journal/AsyncCallbacks.java
+++ b/src/java/org/apache/cassandra/journal/AsyncCallbacks.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.cassandra.journal;
+
+public interface AsyncCallbacks<K, V>
+{
+    /**
+     * Invoked once an entry has been written to the file, and indexes have been updated, but before it
+     * has been flushed to disk. Invoked from the writer thread. Execution order of onWrite() callbacks
+     * with regard to each other is undefined.
+     */
+    void onWrite(long segment, int position, int size, K key, V value, Object writeContext);
+
+    /**
+     * Invoked when anything goes wrong with writing the entry - anywhere from serialization to writing to the file,
+     * to requesting the flush.
+     */
+    void onWriteFailed(K key, V value, Object writeContext, Throwable cause);
+
+    /**
+     * Invoked after {@link Flusher} successfully flushes a segment or multiple segments to disk.
+     * Invocation of this callback implies that any segments older than {@code segment} have been
+     * completed and also flushed.
+     * Invocation of this callback also implies that all {@link #onWrite(long, int, int, Object, Object, Object)}
+     * callbacks for all entries earlier than (segment, position) have finished execution.
+     */
+    void onFlush(long segment, int position);
+
+    void onFlushFailed(Throwable cause);
+}
--- a/src/java/org/apache/cassandra/journal/Flusher.java
+++ b/src/java/org/apache/cassandra/journal/Flusher.java
@@ -32,6 +32,7 @@ import org.apache.cassandra.utils.concurrent.Semaphore;
 import org.apache.cassandra.utils.concurrent.WaitQueue;
 
 import static java.lang.String.format;
+import static java.util.Comparator.comparing;
 import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static org.apache.cassandra.concurrent.ExecutorFactory.Global.executorFactory;
@@ -52,6 +53,7 @@ final class Flusher<K, V>
 
     private final Journal<K, V> journal;
     private final Params params;
+    private final AsyncCallbacks<K, V> callbacks;
 
     private volatile Interruptible flushExecutor;
 
@@ -70,13 +72,14 @@ final class Flusher<K, V>
     private final Semaphore haveWork = newSemaphore(1);
     private volatile boolean flushRequested;
 
-    private final FlushMethod<K> syncFlushMethod;
-    private final FlushMethod<K> asyncFlushMethod;
+    private final FlushMethod<K, V> syncFlushMethod;
+    private final FlushMethod<K, V> asyncFlushMethod;
 
     Flusher(Journal<K, V> journal)
     {
         this.journal = journal;
         this.params = journal.params;
+        this.callbacks = journal.callbacks;
         this.syncFlushMethod = syncFlushMethod(params);
         this.asyncFlushMethod = asyncFlushMethod(params);
     }
@@ -97,7 +100,7 @@ final class Flusher<K, V>
         private final MonotonicClock clock;
         private final NoSpamLogger noSpamLogger;
 
-        private final ArrayList<ActiveSegment<K>> segmentsToFlush = new ArrayList<>();
+        private final ArrayList<ActiveSegment<K, V>> segmentsToFlush = new ArrayList<>();
 
         FlushRunnable(MonotonicClock clock)
         {
@@ -155,38 +158,36 @@ final class Flusher<K, V>
         private void doFlush()
         {
             journal.selectSegmentToFlush(segmentsToFlush);
-            // only schedule onSuccess callbacks for a segment if the preceding segments
-            // have been fully flushed, to preserve 1:1 mapping between record's position
-            // in the journal and onSuccess callback scheduling order
-            boolean scheduleOnSuccessCallbacks = true;
+            segmentsToFlush.sort(comparing(s -> s.descriptor));
+
             try
             {
-                for (ActiveSegment<K> segment : segmentsToFlush)
+                long syncedSegment = -1;
+                int syncedOffset = -1;
+
+                for (ActiveSegment<K, V> segment : segmentsToFlush)
                 {
-                    try
-                    {
-                        scheduleOnSuccessCallbacks = doFlush(segment, scheduleOnSuccessCallbacks) && scheduleOnSuccessCallbacks;
-                    }
-                    catch (Throwable t)
-                    {
-                        segmentsToFlush.forEach(s -> s.scheduleOnFailureCallbacks(t));
-                        throw t;
-                    }
+                    syncedSegment = segment.descriptor.timestamp;
+                    syncedOffset = segment.flush();
+
+                    // if an older segment isn't fully complete + flushed yet, don't attempt to flush any younger ones
+                    if (!segment.isCompletedAndFullyFlushed(syncedOffset))
+                        break;
                 }
+
+                // invoke the onFlush() callback once, covering entire flushed range across all flushed segments
+                if (syncedSegment != -1 && syncedOffset != -1)
+                    callbacks.onFlush(syncedSegment, syncedOffset);
+            }
+            catch (Throwable t)
+            {
+                callbacks.onFlushFailed(t);
+                throw t;
             }
             finally
             {
                 segmentsToFlush.clear();
             }
-        }
-
-        // flush the segment, schedule write callbacks if requested, return whether the segment has been flushed fully
-        private boolean doFlush(ActiveSegment<K> segment, boolean scheduleCallbacks)
-        {
-            int syncedOffset = segment.flush();
-            if (scheduleCallbacks)
-                segment.scheduleOnSuccessCallbacks(syncedOffset);
-            return segment.isFullyFlushed(syncedOffset);
         }
 
         private long firstLaggedAt = Long.MIN_VALUE; // first lag ever or since last logged warning
@@ -231,12 +232,12 @@ final class Flusher<K, V>
     }
 
     @FunctionalInterface
-    private interface FlushMethod<K>
+    private interface FlushMethod<K, V>
     {
-        void flush(ActiveSegment<K>.Allocation allocation);
+        void flush(ActiveSegment<K, V>.Allocation allocation);
     }
 
-    private FlushMethod<K> syncFlushMethod(Params params)
+    private FlushMethod<K, V> syncFlushMethod(Params params)
     {
         switch (params.flushMode())
         {
@@ -247,7 +248,7 @@ final class Flusher<K, V>
         }
     }
 
-    private FlushMethod<K> asyncFlushMethod(Params params)
+    private FlushMethod<K, V> asyncFlushMethod(Params params)
     {
         switch (params.flushMode())
         {
@@ -258,17 +259,17 @@ final class Flusher<K, V>
         }
     }
 
-    void waitForFlush(ActiveSegment<K>.Allocation alloc)
+    void waitForFlush(ActiveSegment<K, V>.Allocation alloc)
     {
         syncFlushMethod.flush(alloc);
     }
 
-    void asyncFlush(ActiveSegment<K>.Allocation alloc)
+    void asyncFlush(ActiveSegment<K, V>.Allocation alloc)
     {
         asyncFlushMethod.flush(alloc);
     }
 
-    private void waitForFlushBatch(ActiveSegment<K>.Allocation alloc)
+    private void waitForFlushBatch(ActiveSegment<K, V>.Allocation alloc)
     {
         pending.incrementAndGet();
         requestExtraFlush();
@@ -277,7 +278,7 @@ final class Flusher<K, V>
         written.incrementAndGet();
     }
 
-    private void asyncFlushBatch(ActiveSegment<K>.Allocation alloc)
+    private void asyncFlushBatch(ActiveSegment<K, V>.Allocation alloc)
     {
         pending.incrementAndGet();
         requestExtraFlush();
@@ -286,7 +287,7 @@ final class Flusher<K, V>
         written.incrementAndGet();
     }
 
-    private void waitForFlushGroup(ActiveSegment<K>.Allocation alloc)
+    private void waitForFlushGroup(ActiveSegment<K, V>.Allocation alloc)
     {
         pending.incrementAndGet();
         alloc.awaitFlush(journal.metrics.waitingOnFlush);
@@ -294,7 +295,7 @@ final class Flusher<K, V>
         written.incrementAndGet();
     }
 
-    private void asyncFlushGroup(ActiveSegment<K>.Allocation alloc)
+    private void asyncFlushGroup(ActiveSegment<K, V>.Allocation alloc)
     {
         pending.incrementAndGet();
         // alloc.awaitFlush(journal.metrics.waitingOnFlush); // TODO (expected): collect async flush metrics
@@ -302,7 +303,7 @@ final class Flusher<K, V>
         written.incrementAndGet();
     }
 
-    private void waitForFlushPeriodic(ActiveSegment<K>.Allocation alloc)
+    private void waitForFlushPeriodic(ActiveSegment<K, V>.Allocation alloc)
     {
         long expectedFlushTime = nanoTime() - periodicFlushLagBlockNanos();
         if (lastFlushedAt < expectedFlushTime)
@@ -314,7 +315,7 @@ final class Flusher<K, V>
         written.incrementAndGet();
     }
 
-    private void asyncFlushPeriodic(ActiveSegment<K>.Allocation ignore)
+    private void asyncFlushPeriodic(ActiveSegment<K, V>.Allocation ignore)
     {
         pending.incrementAndGet();
         // awaitFlushAt(expectedFlushTime, journal.metrics.waitingOnFlush.time()); // TODO (expected): collect async flush metrics
--- a/src/java/org/apache/cassandra/journal/Journal.java
+++ b/src/java/org/apache/cassandra/journal/Journal.java
@@ -24,7 +24,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.Executor;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
@@ -32,8 +31,6 @@ import java.util.function.BooleanSupplier;
 import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.zip.CRC32;
-
-import javax.annotation.Nonnull;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -48,6 +45,7 @@ import org.apache.cassandra.io.util.DataInputBuffer;
 import org.apache.cassandra.io.util.DataOutputBuffer;
 import org.apache.cassandra.io.util.File;
 import org.apache.cassandra.io.util.PathUtils;
+import org.apache.cassandra.journal.Segments.ReferencedSegment;
 import org.apache.cassandra.journal.Segments.ReferencedSegments;
 import org.apache.cassandra.service.StorageService;
 import org.apache.cassandra.utils.Crc;
@@ -86,6 +84,7 @@ public class Journal<K, V> implements Shutdownable
     final String name;
     final File directory;
     final Params params;
+    final AsyncCallbacks<K, V> callbacks;
 
     final KeySupport<K> keySupport;
     final ValueSerializer<K, V> valueSerializer;
@@ -98,12 +97,12 @@ public class Journal<K, V> implements Shutdownable
     volatile long replayLimit;
     final AtomicLong nextSegmentId = new AtomicLong();
 
-    private volatile ActiveSegment<K> currentSegment = null;
+    private volatile ActiveSegment<K, V> currentSegment = null;
 
     // segment that is ready to be used; allocator thread fills this and blocks until consumed
-    private volatile ActiveSegment<K> availableSegment = null;
+    private volatile ActiveSegment<K, V> availableSegment = null;
 
-    private final AtomicReference<Segments<K>> segments = new AtomicReference<>();
+    private final AtomicReference<Segments<K, V>> segments = new AtomicReference<>();
 
     Interruptible allocator;
     private final WaitQueue segmentPrepared = newWaitQueue();
@@ -116,12 +115,14 @@ public class Journal<K, V> implements Shutdownable
     public Journal(String name,
                    File directory,
                    Params params,
+                   AsyncCallbacks<K, V> callbacks,
                    KeySupport<K> keySupport,
                    ValueSerializer<K, V> valueSerializer)
     {
         this.name = name;
         this.directory = directory;
         this.params = params;
+        this.callbacks = callbacks;
 
         this.keySupport = keySupport;
         this.valueSerializer = valueSerializer;
@@ -146,7 +147,7 @@ public class Journal<K, V> implements Shutdownable
                           : descriptors.get(descriptors.size() - 1).timestamp;
         nextSegmentId.set(replayLimit = Math.max(currentTimeMillis(), maxTimestamp + 1));
 
-        segments.set(Segments.ofStatic(StaticSegment.open(descriptors, keySupport)));
+        segments.set(Segments.of(StaticSegment.open(descriptors, keySupport)));
         closer = executorFactory().sequential(name + "-closer");
         allocator = executorFactory().infiniteLoop(name + "-allocator", new AllocateRunnable(), SAFE, NON_DAEMON, SYNCHRONIZED);
         advanceSegment(null);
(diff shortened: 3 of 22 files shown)


src/java/org/apache/cassandra/journal/Flusher.java
```diff
@@ -181,11 +186,11 @@ private void doFlush()
...
 
         // flush the segment, schedule write callbacks if requested, return whether the segment has been flushed fully
-        private boolean doFlush(ActiveSegment<K> segment, boolean scheduleCallbacks)
+        private boolean doFlush(ActiveSegment<K, V> segment, boolean scheduleCallbacks)
         {
             int syncedOffset = segment.flush();
             if (scheduleCallbacks)
-                segment.scheduleOnSuccessCallbacks(syncedOffset);
+                callbacks.onFlush(segment.descriptor.timestamp, syncedOffset);
             return segment.isFullyFlushed(syncedOffset);
```
aweisberg: Add a quick comment with the definition of `FullyFlushed` to the `isFullyFlushed` method since how `endOfBuffer` works.

It seems like `endOfBuffer` is detecting the case where you see the next segment, but don't actually flush all of the entries appended to the previous segment so it's not safe to invoke callbacks for the next segment since you skipped some from the previous segment?

If we don't fully flush, should we flush the next segment if we can't invoke the callbacks for it anyways? Maybe break out of the loop and we can flush everything next time around?

src/java/org/apache/cassandra/journal/Flusher.java
```diff
@@ -155,21 +158,23 @@ public void doRun(Interruptible.State state) throws InterruptedException
...
             // in the journal and onSuccess callback scheduling order
             boolean scheduleOnSuccessCallbacks = true;
             try
             {
-                for (ActiveSegment<K> segment : segmentsToFlush)
+                for (ActiveSegment<K, V> segment : segmentsToFlush)
                 {
                     try
                     {
                         scheduleOnSuccessCallbacks = doFlush(segment, scheduleOnSuccessCallbacks) && scheduleOnSuccessCallbacks;
```
aweisberg: If an earlier segment is not fully flushed (why?) when will we come back later and run the necessary callbacks?

If we didn't fully flush does it imply it will get flushed again in the future? Just checking this is handled.

src/java/org/apache/cassandra/journal/AsyncCallbacks.java
```diff
@@ -0,0 +1,32 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.cassandra.journal;
+
+/**
+ * TODO (expected): document, implement
```
aweisberg: These methods are 90% obvious. Maybe define what flush is, and then contextually at what point each is called.

Could probably be summarized as `onWrite` is called after all in memory and file contents are updated, but before the write is flushed.

`onWriteFailed` could be anywhere from serialization to writing to the file, to requesting the flush.

`onFlush` seems to imply the segment is flushed, but doesn't imply that the success callbacks will or have been run. Not sure it's relevant to mention in this interface.

src/java/org/apache/cassandra/journal/Flusher.java
```diff
@@ -181,11 +186,11 @@ private void doFlush()
         }
 
         // flush the segment, schedule write callbacks if requested, return whether the segment has been flushed fully
-        private boolean doFlush(ActiveSegment<K> segment, boolean scheduleCallbacks)
+        private boolean doFlush(ActiveSegment<K, V> segment, boolean scheduleCallbacks)
```
aweisberg: Maybe be consistent with `scheduleCallbacks` and `scheduleOnSuccessCallbacks`. If `callbacks` contains the `SuccessCallbacks` maybe it should match.

I don't mind the wordiness since we have a lot of callbacks flying around and knowing which one you are looking at makes it more readable.

src/java/org/apache/cassandra/journal/Journal.java
```diff
@@ -194,6 +195,34 @@ public boolean awaitTermination(long timeout, TimeUnit units) throws Interrupted
         return false;
     }
 
+    /**
+     * Read an entry by its address (segment timestamp + offest)
+     *
+     * @return deserialized record if present, null otherwise
+     */
+    public V read(long segmentTimestamp, int offset)
```
aweisberg, line 203: Am I crazy or is this unused?

src/java/org/apache/cassandra/journal/Journal.java
```diff
@@ -665,10 +698,10 @@ void closeActiveSegmentAndOpenAsStatic(ActiveSegment<K> activeSegment)
      */
     public void replayStaticSegments(RecordConsumer<K> consumer)
     {
-        List<StaticSegment<K>> staticSegments = new ArrayList<>(segments().onlyStatic());
-        staticSegments.sort(comparing(segment -> segment.descriptor));
-
-        for (StaticSegment<K> segment : staticSegments)
+        ArrayList<StaticSegment<K, V>> staticSegments = new ArrayList<>();
```
aweisberg: Why switch to `ArrayList`?

src/java/org/apache/cassandra/journal/Segments.java
```diff
@@ -17,123 +17,92 @@
...
 
-    Iterable<Segment<K>> all()
+    Iterable<Segment<K, V>> all()
     {
-        return allSegments;
+        return segments.values();
     }
 
-    Collection<ActiveSegment<K>> onlyActive()
+    void selectActive(long maxTimestamp, Collection<ActiveSegment<K, V>> into)
```
aweisberg, line 94: Could be an iterables transform and filter, but sure this is probably slightly faster.

src/java/org/apache/cassandra/net/Message.java
```diff
@@ -194,6 +197,58 @@ public TraceType traceType()
         return header.traceType();
     }
 
+    /*
+     * minimal response context extraction
+     */
+
+    public ResponseContext extractResponseContext()
```
aweisberg: Unused but I assume that is coming?

src/java/org/apache/cassandra/net/Message.java
```diff
@@ -71,7 +70,7 @@
  *
  * @param <T> The type of the message payload.
  */
-public class Message<T> implements ReplyContext
+public class Message<T> implements ResponseContext
```
aweisberg, line 73: It's getting a little weird that `ResponseContext` and `ReplyContext` are basically the same thing, but one is accord and one is cassandra. Might be better as `CassandraResponseContext` and `AccordResponseContext` or `CassandraResponseContext` and `AccordResponseContext`

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -423,14 +421,14 @@ private Runnable saveCommandsForKey(CommandsForKey before, CommandsForKey after)
     private Runnable saveTimestampsForKey(TimestampsForKey before, TimestampsForKey after)
     {
         Mutation mutation = AccordKeyspace.getTimestampsForKeyMutation(id, before, after, nextSystemTimestampMicros());
-        return null != mutation ? mutation::apply : null;
+        return null != mutation ? mutation::applyUnsafe : null;
```
aweisberg, line 424: We need to remember to eventually also have the mutations applied as part of transactions be applied unsafe as well and then link segments being discarded to flushing.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -211,21 +254,28 @@ public Object shutdownNow()
     @Override
     public boolean awaitTermination(long timeout, TimeUnit units) throws InterruptedException
     {
-        // TODO (expected)
+        // TODO (expected, other)
         return true;
     }
 
-    void appendAuxiliaryRecord(AuxiliaryRecord record)
+    void appendAuxiliaryRecord(AuxiliaryRecord record, Object context)
```
aweisberg, line 267: One sentence description of what an auxiliary record is that is being appended?

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -211,21 +254,28 @@ public Object shutdownNow()
...
     {
         Key key = new Key(record.timestamp, record.type());
-        journal.write(key, record, SENTINEL_HOSTS);
+        journal.asyncWrite(key, record, SENTINEL_HOSTS, context);
     }
 
-    public void appendMessage(Message message, Executor executor, AsyncWriteCallback callback)
+    public void appendRemoteRequest(Request request, ResponseContext context)
```
aweisberg, line 276: Same as above

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -211,21 +254,28 @@ public Object shutdownNow()
...
     {
-        Type type = Type.fromMessageType(message.type());
-        Key key = new Key(type.txnId(message), type);
-        journal.asyncWrite(key, message, SENTINEL_HOSTS, executor, callback);
+        Type type = Type.fromMessageType(request.type());
+        Key key = new Key(type.txnId(request), type);
+        journal.asyncWrite(key, request, SENTINEL_HOSTS, context);
+    }
+
+    public void appendLocalRequest(LocalRequest<?> request)
```
aweisberg, line 286: Same as above

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -237,17 +287,241 @@ public void appendMessageBlocking(Message message)
...
+            long segment = in.readUnsignedVInt();
+            int position = in.readUnsignedVInt32();
+            return new Pointer(segment, position);
+        }
+    }
+
+    private class JournalCallbacks implements AsyncCallbacks<Key, Object>
+    {
+        @Override
+        public void onWrite(long segment, int position, int size, Key key, Object value, Object writeContext)
```
aweisberg, line 410: One sentence comment, route the completion of writes to the next step which is either frame aggregation or frame application if the write to the log was a frame.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -699,6 +1003,246 @@ private static int msVersion(int version)
         }
     }
 
+    /*
+     * Record framing logic
+     */
+
+    private final class FrameAggregator implements Interruptible.Task
```
aweisberg, line 1055: Describe what a frame is, how/why we create them, and what we do with them once created and after they are written, and after they are flushed

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -699,6 +1003,246 @@ private static int msVersion(int version)
...
+             */
+
+            PendingRequest request;
+            while (null != (request = unframedRequests.poll()))
+            {
+                long waitForEpoch = request.context.waitForEpoch;
+                if (!node.topology().hasEpoch(waitForEpoch))
+                {
+                    delayedRequests.computeIfAbsent(waitForEpoch, ignore -> new ArrayList<>()).add(request);
+                    if (!waitForEpochs.containsLong(waitForEpoch))
```
aweisberg, line 1131: A set would resist long searches through large numbers of epochs. If that happens maybe we have bigger problems.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -699,6 +1003,246 @@ private static int msVersion(int version)
...
+
+            for (PendingRequest req : requestBuffer)
+            {
+                pointersBuffer.add(req.pointer);
+                contextsBuffer.add(req.context);
+            }
+
+            if (!requestBuffer.isEmpty())
+            {
+                FrameRecord frame = new FrameRecord(node.uniqueNow(), pointersBuffer.toArray(new Pointer[0]));
```
aweisberg: You could avoid converting to an array which is a copy, and not sure if Java escapes out the new empty array.

When deserializing you can do `Arrays.asList` which is strictly cheaper than always having to do the copy.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -699,6 +1003,246 @@ private static int msVersion(int version)
...
+
+        private volatile Interruptible executor;
+
+        // a signal and flag that callers outside the aggregator thread can use
+        // to signal they want the aggregator to run again
+        private final Semaphore haveWork = newSemaphore(1);
+
+        void onWrite(Pointer pointer, RequestContext context)
+        {
+            unframedRequests.add(new PendingRequest(pointer, context));
```
aweisberg: `RequestContext` could provide storage for `Pointer` instead of allocating a new object. A step further would be to store pointers in a columnar fashion so you don't have pointer objects, but that really depends on how hard you want to go at this.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -699,6 +1003,246 @@ private static int msVersion(int version)
...
+
+            PendingRequest(Pointer pointer, RequestContext context)
+            {
+                this.pointer = pointer;
+                this.context = context;
+            }
+        }
+    }
+
+    private final class FrameApplicator implements Runnable
```
aweisberg, line 1160: Description that can reference `FrameAggregator` for most of the details. This one is pretty straightforward. It gets the aggregated frames containing previously written requests/messages and sorts and "applies" them once that part of the journal is flushed.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -699,6 +1003,246 @@ private static int msVersion(int version)
...
+        private void applyFrame(FrameRecord frame, FrameContext context)
+        {
+            Invariants.checkState(frame.pointers.length == context.requestContexts.length);
+            for (int i = 0; i < frame.pointers.length; i++)
+                applyRequest(frame.pointers[i], context.requestContexts[i]);
+        }
+
+        private void applyRequest(Pointer pointer, RequestContext context)
+        {
+            Request request = (Request) cachedRecords.remove(pointer);
```
aweisberg, line 1225: So right now this cache isn't really a cache, but in the future it will read the message back off of disk if necessary?

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -699,6 +1003,246 @@ private static int msVersion(int version)
...
+            else
+            {
+                Invariants.checkState(type.isLocalRequest());
+                LocalRequestContext ctx = (LocalRequestContext) context;
+                //noinspection unchecked,rawtypes
+                ((LocalRequest) request).process(node, ctx.callback);
+            }
+        }
+
+        private final class PendingFrame
```
aweisberg, line 1248: Comment, frame containing pointers to all the requests in the journal contained by the frame that have been written to the journal, but have not been process by the frame applicaticator yet. Will be processed by the frame applicator once the journal has flushed the frame.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -699,6 +1003,246 @@ private static int msVersion(int version)
...
+            PendingFrame(Pointer start, Pointer end, FrameContext context)
+            {
+                this.start = start;
+                this.end = end;
+                this.context = context;
+            }
+        }
+    }
+
+    static final class FrameContext
```
aweisberg: Could be private

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -699,6 +1003,246 @@ private static int msVersion(int version)
...
+    {
+        final RequestContext[] requestContexts;
+
+        FrameContext(RequestContext[] requestContexts)
+        {
+            this.requestContexts = requestContexts;
+        }
+    }
+
     static abstract class AuxiliaryRecord
```
aweisberg: Could be private

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -711,43 +1255,57 @@ static abstract class AuxiliaryRecord
         abstract Type type();
     }
 
-    /*
-     * Placeholder for future record.
-     */
-    static final class ReplayRecord extends AuxiliaryRecord
+    static final class FrameRecord extends AuxiliaryRecord
```
aweisberg: Ok going to stop mentioning things that could be private :-)

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -711,43 +1255,57 @@ static abstract class AuxiliaryRecord
...
+                    pointers[i] = Pointer.deserialize(in);
+                return new FrameRecord(key.timestamp, pointers);
             }
         };
     }
 
+    /*
+     * Message provider implementation
+     */
+
```
aweisberg, line 1326: Extra line break?

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -765,6 +1323,7 @@ private MessageProvider(TxnId txnId)
         @Override
         public Set<MessageType> test(Set<MessageType> messages)
         {
+            // TODO (expected, correctness): fix other message provider methods to account for synonyms too
```
aweisberg: Which methods need to handle synonyms?

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -482,35 +479,14 @@ public TopologyManager topology()
         }
     }
 
-    private void handleLocalMessage(LocalMessage message, Node node)
+    // TODO (expected): confirm that we don't need to worry about waitForEpoch() here being ahead of known;
+    //                  was not being handled by Accord pre-journal, likely a non-issue.
+    private void handleLocalRequest(LocalRequest<?> request, Node node)
```
aweisberg, line 494: Which `waitForEpoch`?

src/java/org/apache/cassandra/service/accord/AccordVerbHandler.java
```diff
@@ -52,45 +48,34 @@ public void doVerb(Message<T> message) throws IOException
...
 //        ClusterMetadataService.instance().maybeCatchup(message.epoch());
         logger.debug("Receiving {} from {}", message.payload, message.from());
+        Node.Id fromNodeId = endpointMapper.mappedId(message.from());
         T request = message.payload;
+
         long knownEpoch = request.knownEpoch();
         if (!node.topology().hasEpoch(knownEpoch))
         {
             node.configService().fetchTopologyForEpoch(knownEpoch);
             long waitForEpoch = request.waitForEpoch();
```
aweisberg: Request really needs to document the difference between `knownEpoch` and `waitForEpoch`.

src/java/org/apache/cassandra/service/accord/AccordVerbHandler.java
```diff
@@ -52,45 +48,34 @@ public void doVerb(Message<T> message) throws IOException
...
             if (!node.topology().hasEpoch(waitForEpoch))
```
aweisberg: This seems like it's a property independent of `knownEpoch` and we could move it out and not have to copy the journal append part and have less indentation? The journal append is not dependent on having `waitForEpoch` it just depends on side effects. It really just changes whether we process directly or `withEpoch`.

Maybe `request.process` should handle this check. instead of the caller?

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
```diff
@@ -165,35 +166,35 @@ public static class State
...
-                }
-
-                @Override
-                public void onFailure(Throwable error)
-                {
-                    eventsDurable.decrement(); // to make sure we don't block forever
-                    exceptions.add(error);
-                }
-            });
+//            journal.appendMessageTest(request, executor, new AsyncWriteCallback()
```
aweisberg, line 183: Test not passing?

aweisberg (changes requested):
A lot of comment requests, some suggestions that aren't strictly necessary. Will circle back to the Accord part now.

src/java/org/apache/cassandra/journal/Journal.java
iamaleksey: This one is for reading the entry by `Pointer` if it isn't in the cache. Right now unused because message objects are never evicted in this PR.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Yep.

src/java/org/apache/cassandra/journal/Flusher.java
iamaleksey: Agreed. Renamed the method to `isCompletedAndFullyFlushed()` and changed the logic to break out early if an older segment isn't fully complete and flushed. Also to invoke `onFlushed()` callback just once per iteration of the flusher instead of once per segment that'd been flushed.

src/java/org/apache/cassandra/journal/Flusher.java
iamaleksey: See the comment above. Renamed the method to reflect its real intention - to tell if the segment is now fully completed *and* flushed. It will be flushed again in the future, in the next flusher iteration following the completion of the last write to that segment.

src/java/org/apache/cassandra/journal/Flusher.java
iamaleksey: The method is now gone, following your proposed simplification of flushing logic, so I believe this is comment is now obsolete.

src/java/org/apache/cassandra/journal/AsyncCallbacks.java
iamaleksey: Done, thanks.

src/java/org/apache/cassandra/net/Message.java
iamaleksey: Leftover from before a refactoring. This became `RemoteRequestContext` in `AccordJournal`. Cleaned up, cheers.

src/java/org/apache/cassandra/journal/Journal.java
iamaleksey: Just an in the moment declaration stylistic choice. Doesn't matter either way, but rolled it back.

src/java/org/apache/cassandra/journal/Segments.java
iamaleksey: I reuse that list in `Flusher` so as to not generate more garbage than necessary. I believe it's a good pattern in general, though, for methods like this to take an input collection - it's something Matin Thompson tends to advocate for in his talks.

src/java/org/apache/cassandra/net/Message.java
iamaleksey: I have an old branch somewhere that renames all things 'reply' in Accord to 'response', to match Cassandra nomenclature. Should probably do that first, but later, when the project is near completion, to ease rebases.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Done, cheers.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Done, cheers.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Addressed, thanks.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Addressed, thanks.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Addressed, thanks.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Yep, that's what I was thinking too. That this shouldn't happen, and that, if it does, going through a chunky  `LongArrayList`
of non-boxed longs is the least of our problems. I think this would be best kept as is.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Addressed, thanks.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Addressed.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Addressed, and others.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Intended, a comment for the next section of the file, not a particular method or class.

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
iamaleksey: I need to check if things work again, but there was a bug with how interceptors were set up in the initial simulation test implementation that caused things to hang, and the test had to be disabled. I will see what's needed to make it work.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Added, thanks.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Left-over TODO from before I fixed all `read*` method implementations by making `readMessage()` synonym-aware. Cleaned up.

src/java/org/apache/cassandra/service/accord/AccordVerbHandler.java
iamaleksey: Good point. Simplified. Not touching `request.process()` as part of this though.

src/java/org/apache/cassandra/service/accord/AccordService.java
iamaleksey: Updated/clarified.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Updated, cheers.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Good call. Implemented without going too hard.

src/java/org/apache/cassandra/service/accord/AccordVerbHandler.java
aweisberg: To take that a step further... `knownEpoch` has a default implementation that is just `waitForEpoch` and it has no other implementations?

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: I still needed to copy somewhere, as the original buffer was needed for later reuse. But I simplified the logic a little by switching to lists and not keeping additional buffer state between runs of aggregator. 

P.S. As for escaping new empty arrays there, this is from 2016,  regarding `toArray()`: https://shipilev.net/blog/2016/arrays-wisdom-ancients/

src/java/org/apache/cassandra/service/accord/AccordVerbHandler.java
iamaleksey: Correct. Removing its uses from here, and will remove the method from Accord core in the sister PR as well. Simplified some related logic here, reducing nesting and a redundant call to `fetchTopologyForEpoch()`.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -1271,9 +1259,9 @@ private static abstract class AuxiliaryRecord
 
     private static final class FrameRecord extends AuxiliaryRecord
     {
-        final Pointer[] pointers;
+        final List<Pointer> pointers;
 
-        FrameRecord(Timestamp timestamp, Pointer[] pointers)
+        FrameRecord(Timestamp timestamp, List<Pointer> pointers)
```
aweisberg, line 1289: It looks like a frame record is basically just a list of pointers, and the timestamp in the frame record is unused right now?

For the pointers could you use collection serializer instead of hand rolling the serialization?

Would it make sense to only have the list of pointers and no frame record class?

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -1099,7 +1086,8 @@ private void doRun()
                 long waitForEpoch = waitForEpochs.getLong(i);
                 if (!node.topology().hasEpoch(waitForEpoch))
                     break;
-                requestBuffer.addAll(delayedRequests.remove(waitForEpoch));
+                if (null == requests) requests = new ArrayList<>();
```
aweisberg: This could at least be allocated to be the size of the delayed requests in the first epoch.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
```diff
@@ -1123,17 +1111,17 @@ private void doRun()
                 }
                 else
                 {
-                    requestBuffer.add(request);
+                    if (null == requests) requests = new ArrayList<>();
```
aweisberg, line 1139: Hard to say without knowing the actual inputs for this in practice, but `requests` might actually make more sense to have a fixed buffer. One thing you might want to do is artificially constrain frame size to some cache friendly number. I'll bet allocate and copy is plenty fast in that situation, and faster than doing even a few resizes.

But if this is always pretty small then it doesn't help much, but also doesn't hurt?

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: Done, thanks.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: We might indeed. I'll make a note, and possibly revisit in the future, if that's okay.

src/java/org/apache/cassandra/service/accord/AccordJournal.java
iamaleksey: We do need the key later. Did switch to the collection serializer, cheers.

aweisberg (approved):
+1 TY

PR #2988, opened 2023-12-12
CASSANDRA-13855 - Add HTTP/S Seed provider

Claudenw:
 
```

```


conf/cassandra.yaml
```diff
@@ -647,6 +647,17 @@ seed_provider:
       # based on the configured name service on the system. Defaults to "false".
       #  resolve_multiple_ip_addresses_per_dns_record: "false"
 
+
+#  - class_name: org.apache.cassandra.locator.UrlSeedProvider
```
smiklosovic, line 651: will be changed upon commit

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
```diff
@@ -18,122 +18,30 @@
...
-
-import org.apache.cassandra.config.CassandraRelevantProperties;
-import org.apache.cassandra.config.DurationSpec;
-import org.apache.cassandra.exceptions.ConfigurationException;
-
 import static java.lang.String.format;
-import static java.nio.charset.StandardCharsets.UTF_8;
 
-abstract class AbstractCloudMetadataServiceConnector
+abstract class AbstractCloudMetadataServiceConnector extends HttpServiceConnector
```
smiklosovic, line 23: The reason why I chose this approach is that `AbstractCloudMetadataServiceConnector` is meant to be used for _clouds_ in connection with snitches. That means that it has properties like `metadata_url` and `metadata_request_timeout` to reflect the fact that we are querying some _metadata service_ as all these clouds have them.

On the other hand, I am not sure if we are going to use same terminology for seed providers. We could. Currently there is `seeds_url` instead of `metadata_url` as I just think it reads better.

If we decide that using `metadata_*`  terminology is OK for seed providers too, I can just reuse AbstractCloudMetadataServiceConnector in seed providers and get rid of `HttpServiceConnector`.

We would also need to change the error messages a little bit to get rid of mentions about snitches to have it snitch/seed agnostic.

conf/cassandra.yaml
```diff
@@ -647,6 +647,17 @@ seed_provider:
...
+#    parameter:
+        # seeds_url points to config which container one seed per line.
+        # Ex: "http://<seeds-url-address>"
+#      - seeds_url: "http://127.0.0.1"
```
Mmuzaf, line 655: I think it should be a comma separated list of URLs. Having one URL is a single point of failure. Users should be able to set as many points as they like.

conf/cassandra.yaml
```diff
@@ -647,6 +647,17 @@ seed_provider:
...
+        # read timout, 0 = no timeout.
```
Mmuzaf, line 656: It's worth mentioning here, that the default timeout is 30sec.

src/java/org/apache/cassandra/locator/HttpServiceConnector.java
```diff
@@ -0,0 +1,145 @@
...
+
+            try (BufferedReader bufferedReader = new BufferedReader(new InputStreamReader(conn.getInputStream())))
+            {
+                String line;
+                while ((line = bufferedReader.readLine()) != null) {
+                    response.add(line);
+                }
+            }
+
+            return response.stream().collect(Collectors.joining(System.lineSeparator()));
```
Mmuzaf, line 88: Are you sure that the remote server returns the same line separator as the local OS? Look like a breaking change as the snitches rely on the returning string.

src/java/org/apache/cassandra/locator/HttpServiceConnector.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    public static String resolveMetadataUrl(Properties properties, String keyProperty)
+    {
+        String parsedUrl = properties.getProperty(keyProperty);
+
+        try
+        {
+            if (parsedUrl == null)
+                throw new IllegalStateException(String.format("%s was not specified", keyProperty));
```
Mmuzaf, line 104: I guess it should be the `ConfigurationException` type, right? As it might not be handled properly when the node starts

src/java/org/apache/cassandra/locator/HttpSeedProvider.java
```diff
@@ -0,0 +1,135 @@
...
+import org.slf4j.LoggerFactory;
+
+import static java.util.stream.Collectors.toList;
+import static org.apache.cassandra.locator.HttpServiceConnector.resolveMetadataUrl;
+import static org.apache.cassandra.locator.HttpServiceConnector.resolveRequestTimeoutMs;
+
+public class HttpSeedProvider implements SeedProvider
+{
+    private static final Logger logger = LoggerFactory.getLogger(HttpSeedProvider.class);
+    private static final Pattern NEW_LINES_PATTERN = Pattern.compile("\\n");
```
Mmuzaf, line 41: The `System.lineSeparator()` is used in the `apiCall` method and here we expect the `\n` symbol, I think the tests only work because we are running. If someone tried to run it locally on Windows it would fail. I doubt anyone would do that, but the expectations need to be the same.

src/java/org/apache/cassandra/locator/HttpSeedProvider.java
```diff
@@ -0,0 +1,135 @@
...
+import java.util.regex.Pattern;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static java.util.stream.Collectors.toList;
+import static org.apache.cassandra.locator.HttpServiceConnector.resolveMetadataUrl;
+import static org.apache.cassandra.locator.HttpServiceConnector.resolveRequestTimeoutMs;
+
+public class HttpSeedProvider implements SeedProvider
```
Mmuzaf, line 38: Please, add the javadoc to how it can be used and set via configuration.

src/java/org/apache/cassandra/locator/HttpSeedProvider.java
```diff
@@ -0,0 +1,135 @@
...
+        for (String rawHeaderPair : rawRequestHeaders.split(","))
+        {
+            String[] header = rawHeaderPair.trim().split("=");
+            if (header.length != 2)
+                continue;
+
+            String headerKey = header[0].trim();
+            String headerValue = header[1].trim();
+
+            if (headerKey.length() != 0 && headerValue.length() != 0)
```
Mmuzaf, line 119:
```suggestion
            if (!headerKey.isEmpty() && !headerValue.isEmpty())
```

src/java/org/apache/cassandra/locator/HttpSeedProvider.java
```diff
@@ -0,0 +1,135 @@
...
+                                       .distinct()
+                                       .collect(toList());
+
+            List<InetAddressAndPort> seeds = new ArrayList<>(lines.size());
+
+            for (String line : lines)
+            {
+                try
+                {
+                    seeds.add(InetAddressAndPort.getByName(line));
```
Mmuzaf, line 84: If the `line` is specified without a port e.g. `127.0.0.1` the default `7000` is used. Is this correct? I would not expect this from a user point of view for the HTTP seed provider.

src/java/org/apache/cassandra/locator/HttpSeedProvider.java
```diff
@@ -0,0 +1,135 @@
...
+
+    private final HttpServiceConnector serviceConnector;
+    private final Map<String, String> headers;
+
+    public HttpSeedProvider(Map<String, String> parameters)
+    {
+        this(getProperties(parameters));
+    }
+
+    public HttpSeedProvider(Properties properties)
```
Mmuzaf, line 56: This should probably be the package-private visibility.

src/java/org/apache/cassandra/locator/HttpSeedProvider.java
```diff
@@ -0,0 +1,135 @@
...
+    private static Properties getProperties(Map<String, String> args)
+    {
+        if (args == null)
+            return new Properties();
+
+        Properties properties = new Properties();
+        properties.putAll(args);
+        return properties;
+    }
+}
```
Mmuzaf, line 135: Add a nl, please :-)

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
```diff
@@ -18,122 +18,30 @@
...
 {
     static final String METADATA_URL_PROPERTY = "metadata_url";
     static final String METADATA_REQUEST_TIMEOUT_PROPERTY = "metadata_request_timeout";
     static final String DEFAULT_METADATA_REQUEST_TIMEOUT = "30s";
```
Mmuzaf, line 27: Should we remove the constant due to the `HttpSeedProvider#DEFAULT_REQUEST_TIMEOUT` existence?

Mmuzaf: I mean probably the timeout should be a part of the `HttpServiceConnector` configuration.

conf/cassandra.yaml
smiklosovic: well here it seems that seeds_url property is misleading and we should react on metadata_url instead as I was commenting on that already. This is just an http address where that seed provider will find a list of seeds, it is not like this seed provider would contact all the addresses enumerated here.

src/java/org/apache/cassandra/locator/HttpServiceConnector.java
smiklosovic: `response` is populated in `try-with-resources` above. new lines like `\n` `\r` or `\r\n` are treated internally in `bufferReader.readLine()`, here I just get lines, whatever they ended with so I can join them with platform specific "new line" where Cassandra runs.

src/java/org/apache/cassandra/locator/HttpSeedProvider.java
smiklosovic: If port is part of a `line` here, it will be used, if `line` does not contain port, default port of 7000 will be used.

src/java/org/apache/cassandra/locator/HttpSeedProvider.java
Mmuzaf, line 38: Please, update the documentation with a new SeedProvider.

src/java/org/apache/cassandra/locator/AbstractCloudMetadataServiceConnector.java
smiklosovic: Not only that ... I went a little bit wild here for seeds and it is possible to specify headers from them in the configuration for `HttpSeedProvider` too but it would be cool if it was possible to specify headers for snitches as well. That would require to move headers logic in `HttpSeedProvider` to `HttpServiceConnector`.

src/java/org/apache/cassandra/locator/HttpSeedProvider.java
smiklosovic: I think this answers it https://github.com/apache/cassandra/pull/2988/files#r1537872100

We might also use `Pattern.compile("\" + System.lineSeparator())` or similar ...

src/java/org/apache/cassandra/locator/HttpServiceConnector.java
Mmuzaf: The previous version (see below), so I guess it's not a 1:1 replacement. 

```
            byte[] b = new byte[cl];
            try (DataInputStream d = new DataInputStream((InputStream) conn.getContent()))
            {
                d.readFully(b);
            }
            return new String(b, UTF_8);
```

conf/cassandra.yaml
Mmuzaf: Indeed, `metadata_url` or `url` is better. My point is that a URL adds a single point of failure to the cluster, which is not good (without reading the seed it won't start, right?). We can easily support a list of URLs, which should be fine and safe.

src/java/org/apache/cassandra/locator/HttpServiceConnector.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.stream.Collectors;
+
+import com.google.common.collect.ImmutableMap;
+
+import org.apache.cassandra.config.DurationSpec;
+import org.apache.cassandra.exceptions.ConfigurationException;
+
+import static java.lang.String.format;
+
+public class HttpServiceConnector
```
Mmuzaf, line 41: It's worth adding a simple test to the HttpSeedProvider itself, which will be successfully loaded and used in the cassandra.yaml and all seed are successfully loaded from designated http

src/java/org/apache/cassandra/locator/HttpServiceConnector.java
smiklosovic: i think this is just fine

conf/cassandra.yaml
smiklosovic: @Mmuzaf implemented here as the second commit https://github.com/Claudenw/cassandra/pull/11/files

src/java/org/apache/cassandra/locator/HttpServiceConnector.java
smiklosovic: @Mmuzaf 

I think we can add a distributed in-jvm dtest which would spin two nodes where the second one would have a URL seed provider configured which would point to the first node on some non-default storage port to prove the point. There would be also a respective WireMockRule which would mimic / mock a remote metadata service returning the seed as the first node.

smiklosovic: @Mmuzaf  implemented as the third commit here https://github.com/Claudenw/cassandra/pull/11/commits

PR #2989, opened 2023-12-12
CASSANDRA-18598 Simplify segment building in SAI to use single in-memory structure

mike-tr-adamson:
This removes the RAMStringIndexer for literal indexes and replaces
  it with a SegmentRamBuffer that replaces BlockBalancedTreeRamBuffer

patch by Mike Adamson; reviewed by <reviewer> for CASSANDRA-18598

```

```


src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -453,7 +453,7 @@ public enum CassandraRelevantProperties
     SAI_SORTED_TERMS_PARTITION_BLOCK_SHIFT("cassandra.sai.sorted_terms_partition_block_shift", "4"),
 
     SAI_TEST_BALANCED_TREE_DEBUG_ENABLED("cassandra.sai.test.balanced_tree_debug_enabled", "false"),
-    SAI_TEST_DISABLE_TIMEOUT("cassandra.sai.test.disable.timeout", "false"),
+    SAI_TEST_DISABLE_TIMEOUT("cassandra.sai.test.timeout_disabled", "false"),
```
mike-tr-adamson, line 456: I have changed this mainly because it didn't fall into the same naming format as other, similar, properties

maedhroz: I guess the only other option would be to use "enabled" rather than "disabled" verbiage and flip the default.

mike-tr-adamson: I thought about but I don't like the idea of having the default as `true`. Not sure why apart from this is a debug / test setting so I prefer it be explicit what the flag is doing.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
```diff
@@ -78,31 +75,36 @@ public abstract class SegmentBuilder
...
     {
-        private final byte[] scratch;
-        private final BlockBalancedTreeRamBuffer trieBuffer;
+        protected final SegmentRamBuffer segmentRamBuffer;
 
-        public BlockBalancedTreeSegmentBuilder(IndexTermType indexTermType, NamedMemoryLimiter limiter)
+        protected RamBufferSegmentBuilder(IndexTermType indexTermType, NamedMemoryLimiter namedMemoryLimiter)
         {
-            super(indexTermType, limiter);
+            super(indexTermType,namedMemoryLimiter);
```
adelapena:
```suggestion
            super(indexTermType, namedMemoryLimiter);
```

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
```diff
@@ -78,31 +75,36 @@ public abstract class SegmentBuilder
...
 
         @Override
         public boolean isEmpty()
         {
-            return trieBuffer.numRows() == 0;
+            return segmentRamBuffer.numRows() == 0;
+        }
+    }
+
+    public static class BlockBalancedTreeSegmentBuilder extends RamBufferSegmentBuilder
```
adelapena: The same way we have `LiteralSegmentBuilder` and `NumericIndexWriter`, perhaps this could be named `NumericSegmentBuilder`.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
```diff
@@ -78,31 +75,36 @@ public abstract class SegmentBuilder
...
+        {
+            super(indexTermType, limiter);
         }
 
         @Override
         protected long addInternal(ByteBuffer term, int segmentRowId)
         {
-            indexTermType.toComparableBytes(term, scratch);
-            return trieBuffer.add(segmentRowId, scratch);
+            return segmentRamBuffer.add(v -> indexTermType.asComparableBytes(term, v), term.limit(), segmentRowId);
```
adelapena: The only difference between the two implementations of `addInternal` in the children of `RamBufferSegmentBuilder` is how the `ByteSource` term is built. Maybe we could make that part abstract in `RamBufferSegmentBuilder` and provide a common implementation of `addInternal` in `RamBufferSegmentBuilder`. So `RamBufferSegmentBuilder` would contain:
```java
protected abstract ByteSource asBytes(ByteBuffer term, ByteComparable.Version version);

@Override
protected long addInternal(ByteBuffer term, int segmentRowId)
{
    return segmentRamBuffer.add(v -> asBytes(term, v), term.limit(), segmentRowId);
}
```
Then in `BlockBalancedTreeSegmentBuilder`:
```java
@Override
protected ByteSource asBytes(ByteBuffer term, ByteComparable.Version version)
{
    return indexTermType.asComparableBytes(term, version);
}
```
And in `LiteralSegmentBuilder`:
```java
@Override
protected ByteSource asBytes(ByteBuffer term, ByteComparable.Version version)
{
    return ByteSource.fixedLength(term);
}
```

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
```diff
@@ -112,54 +114,31 @@ protected SegmentMetadata.ComponentMetadataMap flushInternal(IndexDescriptor ind
...
-        {
-            return ramIndexer.isEmpty();
         }
 
         @Override
         protected long addInternal(ByteBuffer term, int segmentRowId)
         {
-            copyBufferToBytesRef(term, stringBuffer);
-            return ramIndexer.add(stringBuffer.get(), segmentRowId);
+            return segmentRamBuffer.add(v -> ByteSource.fixedLength(term), term.limit(), segmentRowId);
```
adelapena: Do all literal terms have fixed length?

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentRamBuffer.java
```diff
@@ -55,14 +56,17 @@ public long memoryUsed()
...
         final long initialSizeOnHeap = trie.sizeOnHeap();
         final long reducerHeapSize = postingsAccumulator.heapAllocations();
 
         try
         {
-            trie.putRecursive(v -> ByteSource.fixedLength(value), segmentRowId, postingsAccumulator);
+            if (termLength <= MAX_RECURSIVE_TERM_LENGTH)
+                trie.putRecursive(term, segmentRowId, postingsAccumulator);
+            else
+                trie.apply(Trie.singleton(term, segmentRowId), postingsAccumulator);
```
adelapena: Maybe this can be simplified to:
```java
trie.putSingleton(term, segmentRowId, postingsAccumulator, termLength <= MAX_RECURSIVE_TERM_LENGTH);
```
There is a similar usage pattern in `TrieMemoryIndex#addTerm(PrimaryKey, ByteBuffer)`.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
```diff
@@ -78,31 +75,36 @@ public abstract class SegmentBuilder
...
+        protected RamBufferSegmentBuilder(IndexTermType indexTermType, NamedMemoryLimiter namedMemoryLimiter)
         {
-            super(indexTermType, limiter);
+            super(indexTermType,namedMemoryLimiter);
 
-            scratch = new byte[indexTermType.fixedSizeOf()];
-            trieBuffer = new BlockBalancedTreeRamBuffer(indexTermType.fixedSizeOf());
-            totalBytesAllocated = this.trieBuffer.memoryUsed();
+            segmentRamBuffer = new SegmentRamBuffer();
+            totalBytesAllocated = this.segmentRamBuffer.memoryUsed();
```
adelapena:
```suggestion
            totalBytesAllocated = segmentRamBuffer.memoryUsed();
```

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
mike-tr-adamson: I'm going to have a look at this. I'm not sure why they need a special encoding here and in the in-memory index. Atm, if I change it, tests fail and that doesn't feel right. I just haven't had a chance to have a close look at it.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
mike-tr-adamson: I like this idea and have taken it a little further. I have introduced a `SegmentWriter` interface such that the `NumericSegmentBuilder` and `LiteralSegmentBuilder` provide a `SegmentWriter` and the abstract class writes the `SegmentRamBuffer` to it. This means that the abstract class is now responsible for all the adding and writing logic while the concrete classes are just providers.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
mike-tr-adamson: Some incorrect assumptions were going on here. The `isLiteral` flag was being used as a catch-all for all types here. After looking at it, it is only frozen types that need any special handling, and they only need to use `ByteSource.of`. I have pushed this down to `IndexTermType.asComparableBytes` to handle.

This greatly simplifies things, in the `TrieMemoryIndex` and in the `SegmentBuilder`, since we don't need to handle custom translations of the bytes in multiple places.

src/java/org/apache/cassandra/index/sai/disk/v1/trie/LiteralIndexWriter.java
```diff
@@ -40,72 +40,59 @@
  * Builds an on-disk inverted index structure: terms dictionary and postings lists.
  */
 @NotThreadSafe
-public class LiteralIndexWriter implements Closeable
+public class LiteralIndexWriter implements SegmentWriter
```
mike-tr-adamson, line 42: Remove `Closeable` and make the `LiteralIndexWriter` operate similarly to the `NumericIndexWriter`.  The writers are now opened and closed in the `writeCompleteSegment` method.

src/java/org/apache/cassandra/index/sai/utils/IndexTermType.java
```diff
@@ -525,11 +525,15 @@ public ByteSource asComparableBytes(ByteBuffer value, ByteComparable.Version ver
...
+        else if (isLong())
             // The LongType.asComparableBytes uses variableLengthInteger which doesn't play well with
             // the balanced tree because it is expecting fixed length data. So for SAI we use a optionalSignedFixedLengthNumber
             // to keep all comparable values the same length
-        else if (isLong())
             return ByteSource.optionalSignedFixedLengthNumber(ByteBufferAccessor.instance, value);
+        else if (isFrozen())
+            // We need to override the default frozen implementation here because it will defer to the underlying
+            // type's implementation which will be incorrect, for us, for the case of multi-cell types.
+            return ByteSource.of(value, version);
```
maedhroz, line 536: Trying to clarify here...

If the underlying implementation is incorrect for multi-cell types, why is it the frozen case that we override?

src/java/org/apache/cassandra/index/sai/memory/TrieMemoryIndex.java
```diff
@@ -243,35 +238,7 @@ private void setMinMaxTerm(ByteBuffer term)
...
-                    return END_OF_STREAM;
-                int n = src.next();
-                if (n != END_OF_STREAM)
-                    return n;
-
-                done = true;
-                return ByteSource.TERMINATOR;
-            }
-        };
+        return version -> index.termType().asComparableBytes(input, version);
```
maedhroz, line 241: So the idea here is that this was never actually necessary for single strings, as the purpose of `ByteSource.TERMINATOR` was to make sequences of `ByteSource`s prefix-free?

src/java/org/apache/cassandra/index/sai/disk/v1/bbtree/BlockBalancedTreeWriter.java
```diff
@@ -488,6 +491,7 @@ private class LeafWriter
         private final Callback callback;
         private final ByteBuffersDataOutput leafOrderIndexOutput = new ByteBuffersDataOutput(2 * 1024);
         private final ByteBuffersDataOutput leafBlockOutput = new ByteBuffersDataOutput(32 * 1024);
+        private final byte[] packedValue = new byte[bytesPerValue];
```
maedhroz, line 493: This is basically the old scratch buffer, right?

The only minor thing is that `writeCommonPrefix()` and `valueInBounds()` also use the name `packedValue`, so perhaps we could rename this to avoid possible confusion.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentRamBuffer.java
```diff
@@ -15,34 +15,34 @@
...
 import org.apache.lucene.util.packed.PackedInts;
 import org.apache.lucene.util.packed.PackedLongValues;
 
 /**
- * On-heap buffer for values that provides a sorted view of itself as a {@link BlockBalancedTreeIterator}.
+ * On-heap buffer for values that provides a sorted view of itself as a {@link SegmentRamIterator}.
  */
 @NotThreadSafe
-public class BlockBalancedTreeRamBuffer
+public class SegmentRamBuffer
```
maedhroz:
```suggestion
public class TrieSegmentBuffer
```
We might as well use "trie" to be a little more descriptive. The trie is essential rather than accidental.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentRamBuffer.java
```diff
@@ -73,9 +73,9 @@ public long add(int segmentRowId, byte[] value)
         return (trie.sizeOnHeap() - initialSizeOnHeap) + (postingsAccumulator.heapAllocations() - reducerHeapSize);
     }
 
-    public BlockBalancedTreeIterator iterator()
+    public SegmentRamIterator iterator()
     {
-        return BlockBalancedTreeIterator.fromTrieIterator(trie.entrySet().iterator(), bytesPerValue);
+        return SegmentRamIterator.fromTrieIterator(trie.entrySet().iterator());
```
maedhroz: What if we inlined `fromTrieIterator()` here and moved `fromTermsIterator()` to `TermsIterator`, and then just got rid of the `SegmentRamIterator` in favor of `Iterator<Pair<ByteComparable, PostingList>>`? `SegmentRamBuffer` isn't bloated by any means, and we could even make it `Iterable` in that case, given it already provides `iterator()`. `SegmentRamIterator` is really just half "marker" interface and half utility class for static factories.

If `Iterator<Pair<ByteComparable, PostingList>>` feels to complicated, the solution there has probably aways been to stop using `Pair` and have something like `IndexEntry` or `TermAndPostings` (or something else that communicates well).

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
```diff
@@ -73,103 +67,53 @@ public abstract class SegmentBuilder
...
 
-    public static class BlockBalancedTreeSegmentBuilder extends SegmentBuilder
+    public static class RamBufferSegmentBuilder extends SegmentBuilder
     {
-        private final byte[] scratch;
-        private final BlockBalancedTreeRamBuffer trieBuffer;
+        protected final SegmentRamBuffer segmentRamBuffer;
 
-        public BlockBalancedTreeSegmentBuilder(IndexTermType indexTermType, NamedMemoryLimiter limiter)
+        public RamBufferSegmentBuilder(StorageAttachedIndex index, NamedMemoryLimiter namedMemoryLimiter)
```
maedhroz: nit: It's still `limiter` everywhere else...honestly I think that's enough to know what this is given the context. Maybe just use the shorter name for now?

src/java/org/apache/cassandra/index/sai/utils/IndexTermType.java
mike-tr-adamson: It's really because we don't support proper indexing on frozen indexing, so we don't want the underlying collection as a multi-component `ByteComparable` but as a single-component `ByteComparable`. I'll try and come up with some wording here that is clearer. We could even mention the ticket for full indexing on frozen collections in the comment.

src/java/org/apache/cassandra/index/sai/disk/v1/bbtree/BlockBalancedTreeWriter.java
mike-tr-adamson: I called it `packedValue` because that was what it was called in the method. I don't like `scratchValue` so I'll try and come up with something better.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentRamBuffer.java
mike-tr-adamson: I would prefer `SegmentTrieBuffer`, but I agree that it should have trie in the name somewhere.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentRamBuffer.java
mike-tr-adamson: I agree that the `Pair` usage has got out of hand with the SAI code, so we can start by getting rid of it here. 

I will have a look at the iterator transformations as well. This has been through a number of iterations during this ticket.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentRamBuffer.java
maedhroz: Yeah, my thought was it's a `SegmentBuffer` (imagine a future interface) that uses a trie /shrug

test/unit/org/apache/cassandra/index/sai/disk/v1/InvertedIndexSearcherTest.java
```diff
@@ -220,6 +220,8 @@ private List<Pair<ByteComparable, LongArrayList>> buildTermsEnum(int terms, int
 
     private ByteBuffer wrap(ByteComparable bc)
     {
-        return ByteBuffer.wrap(ByteSourceInverse.readBytes(bc.asComparableBytes(ByteComparable.Version.OSS50)));
+//        return ByteBuffer.wrap(ByteSourceInverse.readBytes(bc.asComparableBytes(ByteComparable.Version.OSS50)));
+
+        return ByteBuffer.wrap(ByteSourceInverse.readBytes(ByteSourceInverse.unescape(ByteSource.peekable(bc.asComparableBytes(ByteComparable.Version.OSS50)))));
```
maedhroz, line 222: nit: clean up commented out bit

We're having to do this now because we've changed `buildStringTermsEnum()` to use `ByteComparable.of()`, which implicitly escapes? I'm trying to remember why we escape at all if we don't index multiple `ByteSource`s...

maedhroz: i.e. I guess this matches what `RamBufferSegmentBuilder` does, given `asComparableBytes()` also escapes, but again, not sure if the escaping matters.

src/java/org/apache/cassandra/index/sai/memory/TrieMemoryIndex.java
mike-tr-adamson: That and we don't do prefix queries for strings, so it isn't necessary. Since we aren't doing prefix queries atm we shouldn't be complicating the code here. When we come to do prefix queries, I suspect things will be very different.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentRamBuffer.java
mike-tr-adamson: I have removed `SegmentRamIterator` and made `TermsIterator` an `Interator<IndexEntry>`. This has simplified things a lot. `SegmentRamBuffer` is now `SegmentTrieBuffer`. `IndexEntry` is preferred because that is used in the internal index to represent the same thing. So we are keeping common terminology.

test/unit/org/apache/cassandra/index/sai/disk/v1/InvertedIndexSearcherTest.java
mike-tr-adamson: The main reason is (as I understand it) is that the `Trie` does not know the difference between a BC terminated with a `0` (escape character) and a BC ending in a `0`.

src/java/org/apache/cassandra/index/sai/utils/TermsIterator.java
```diff
@@ -32,13 +31,8 @@
  * Term enumerations are always ordered by their {@link ByteSource}.
  */
 @NotThreadSafe
-public interface TermsIterator extends Iterator<ByteComparable>, Closeable
+public interface TermsIterator extends Iterator<IndexEntry>, Closeable
 {
-    /**
-     * Get {@link PostingList} for the current term.
-     */
-    PostingList postings();
```
adelapena, line 40: Removing this method makes the first line in the class Javadoc a bit obsolete.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
```diff
@@ -73,103 +67,53 @@ public abstract class SegmentBuilder
...
     private ByteBuffer maxTerm;
 
-    final IndexTermType indexTermType;
+    final StorageAttachedIndex index;
     long totalBytesAllocated;
     int rowCount = 0;
     int maxSegmentRowId = -1;
 
-    public static class BlockBalancedTreeSegmentBuilder extends SegmentBuilder
+    public static class RamBufferSegmentBuilder extends SegmentBuilder
```
adelapena: Wouldn't it make sense to call this `TrieSegmentBuilder`?

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
adelapena: The new `SegmentWriter` interface looks great :)

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentBuilder.java
```diff
@@ -73,103 +67,53 @@ public abstract class SegmentBuilder
...
     {
-        private final byte[] scratch;
-        private final BlockBalancedTreeRamBuffer trieBuffer;
+        protected final SegmentTrieBuffer segmentRamBuffer;
```
adelapena: The var should be named `segmentTrieBuffer`, in line with the new class name. Or maybe just `buffer`?

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentWriter.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.Iterator;
+
+import org.apache.cassandra.index.sai.utils.IndexEntry;
+
+public interface SegmentWriter
+{
+    /**
+     * Appends a set of terms and associated postings to their respective overall SSTable component files.
+     *
+     * @param indexEntryIterator an {@link Iterator} of {@link IndexEntry}s
```
adelapena: It would be good to clarify whether the index entries are expected to be sorted.

src/java/org/apache/cassandra/index/sai/disk/v1/bbtree/BlockBalancedTreeWriter.java
```diff
@@ -130,26 +132,26 @@ public int getMaxPointsInLeafNode()
     }
 
     /**
-     * Write the sorted values from a {@link BlockBalancedTreeIterator}.
+     * Write the sorted values from an {@link Iterator}.
      * <p>
      * @param treeOutput The {@link IndexOutput} to write the balanced tree to
-     * @param values The {@link BlockBalancedTreeIterator} containing the values and rowIDs to be written
+     * @param iterator The {@link Iterator} containing the values and rowIDs to be written
```
adelapena: It would be good to specify whether the index entries are expected to be sorted.

src/java/org/apache/cassandra/index/sai/disk/v1/bbtree/BlockBalancedTreeWriter.java
mike-tr-adamson: I have left this as `packedValue` because it is the only packed value. I have removed other usages of `packedValue` in favour of `packedValues`. In some cases, I have removed method parameters where we were passing class-level fields.

src/java/org/apache/cassandra/index/sai/utils/TermsIterator.java
mike-tr-adamson: Apologies. I had meant to update that javadoc and forgot. I've now updated it.

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentWriter.java
mike-tr-adamson: I have added this to the parameter comment.

src/java/org/apache/cassandra/index/sai/disk/v1/bbtree/BlockBalancedTreeWriter.java
mike-tr-adamson: I have updated this comment.

src/java/org/apache/cassandra/index/sai/disk/v1/bbtree/BlockBalancedTreeWriter.java
```diff
@@ -704,24 +709,24 @@ private int runLen(int start, int end, int byteOffset)
 
         // The following 3 methods are only used when DEBUG is true:
 
-        private void valueInBounds(byte[] packedValue, int packedValueOffset, byte[] minPackedValue, byte[] maxPackedValue)
+        private void valueInBounds(byte[] packedValues, int packedValueOffset, byte[] minPackedValue, byte[] maxPackedValue)
```
maedhroz, line 712: Here and in all these cases below, these are the bytes of one packed value, right? If that's the case, `packedValues` worries me a little?

src/java/org/apache/cassandra/index/sai/disk/v1/segment/SegmentTrieBuffer.java
```diff
@@ -73,9 +77,48 @@ public long add(int segmentRowId, byte[] value)
         return (trie.sizeOnHeap() - initialSizeOnHeap) + (postingsAccumulator.heapAllocations() - reducerHeapSize);
     }
 
-    public BlockBalancedTreeIterator iterator()
+    public Iterator<IndexEntry> iterator()
     {
-        return BlockBalancedTreeIterator.fromTrieIterator(trie.entrySet().iterator(), bytesPerValue);
+        var iterator = trie.entrySet().iterator();
```
maedhroz, line 82: Hey, Java 11 fun!

I have to start using `var` more...

maedhroz (approved):

adelapena (approved):

src/java/org/apache/cassandra/index/sai/disk/v1/bbtree/BlockBalancedTreeWriter.java
mike-tr-adamson: No, these are the bytes of an array of packed values and use the `packedValueOffset` to index to the correct packed value. As such, `packedValues` is the correct name for them. 

While looking into this to see what was going on, I have seen a number of optimisations that can be done to this writer to improve performance. I am going to create a separate ticket to fix these.

PR #2993, opened 2023-12-14
Skip version check if an endpoint is dead state

Runtian:
Skip version check if an endpoint is dead state
If the dead / left node is not skipped, sometimes a such node will cause the upgradeFromVersionSupplier to not able to return memoized data. Which will lead to thread serialization because all of them need to acquire the lock first.

### DIFF
--- a/test/unit/org/apache/cassandra/gms/GossiperTest.java
+++ b/test/unit/org/apache/cassandra/gms/GossiperTest.java
@@ -164,30 +164,20 @@ public class GossiperTest
     @Test
     public void testAssassinatedNodeWillNotContributeToVersionCalculation() throws Exception
     {
-        Gossiper.instance.start(0);
+        int initialNodeCount = 3;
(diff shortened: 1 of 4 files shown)


test/unit/org/apache/cassandra/gms/GossiperTest.java
```diff
@@ -188,7 +188,7 @@ public void testAssassinatedNodeWillNotContributeToVersionCalculation() throws E
         Gossiper.instance.assassinateEndpoint("127.0.0.4");
 
         assertTrue(Gossiper.instance.endpointStateMap.containsKey(InetAddressAndPort.getByName("127.0.0.4")));
-        assertNull(Gossiper.instance.upgradeFromVersionSupplier.get().value());
+        assertTrue(Gossiper.instance.upgradeFromVersionSupplier.get().canMemoize());
```
smiklosovic, line 181: what is that? is not that something different?

Runtian: Just want to make sure the supplier is getting value which can memoize. The null value is the expected memoized value. So they are basically same.

PR #3002, opened 2023-12-18
CASSANDRA-17062 only cache refactoring

smiklosovic:
 
```

```

### DIFF
--- a/doc/modules/cassandra/pages/managing/operating/metrics.adoc
+++ b/doc/modules/cassandra/pages/managing/operating/metrics.adoc
@@ -528,9 +528,16 @@ ViewReplicasSuccess.
 base table and when CL.ONE is achieved on view.
 |===
 
-== Cache Metrics
+== Weighted Cache Metrics
 
-Cassandra caches have metrics to track the effectivness of the caches.
+[NOTE]
+====
+The non-auth caches are weighted (they do care about weight of a single entry) whereas the auth caches are
+non-weighted. Particularly it affects Capacity and Size attributes of the MBeans.
+For weighted caches they represent size of entries in bytes, for non-weighted caches they represent number of entries.
+====
+
+Cassandra caches have metrics to track the effectivness of the weighted caches.
 Though the `Table Metrics` might be more useful.
 
 Reported name format:
@@ -538,7 +545,7 @@ Reported name format:
 *Metric Name*::
   `org.apache.cassandra.metrics.Cache.<MetricName>.<CacheName>`
 *JMX MBean*::
-  `org.apache.cassandra.metrics:type=Cache scope=<CacheName> name=<MetricName>`
+  `org.apache.cassandra.metrics:type=Cache,scope=<CacheName>,name=<MetricName>`
 
 [cols=",,",options="header",]
 |===
@@ -569,8 +576,57 @@ The following caches are covered:
 
 [NOTE]
 ====
-Misses and MissLatency are only defined for the ChunkCache
+* Misses and MissLatency are only defined for the ChunkCache
+* ChunkCache MBean is only available if the cache is enabled
 ====
+
+== Unweighted Cache Metrics
+
+Cassandra caches have metrics to track the effectivness of the unweighted caches.
+
+Reported name format:
+
+*Metric Name*::
+`org.apache.cassandra.metrics.UnweightedCache.<MetricName>.<CacheName>`
+
+*JMX MBean*::
+`org.apache.cassandra.metrics:type=UnweightedCache,scope=<CacheName>,name=<MetricName>`
+
+[cols=",,",options="header",]
+|===
+|Name |Type |Description
+|MaxEntries |Gauge<Integer> |Maximum number of entries.
+|Entries |Gauge<Integer> |Total number of cache entries.
+|FifteenMinuteCacheHitRate |Gauge<Double> |15m cache hit rate.
+|FiveMinuteCacheHitRate |Gauge<Double> |5m cache hit rate.
+|OneMinuteCacheHitRate |Gauge<Double> |1m cache hit rate.
+|HitRate |Gauge<Double> |All time cache hit rate.
+|Hits |Meter |Total number of cache hits.
+|Misses |Meter |Total number of cache misses.
+|MissLatency |Timer |Latency of misses.
+|Requests |Gauge<Long> |Total number of cache requests.
+|===
+
+The following caches are covered:
+
+[cols=",",options="header",]
+|===
+|Name |Description
+|CredentialsCache |Auth cache for credentials.
+|JmxPermissionsCache |Auth cache for JMX permissions.
+|CIDRPermissionsCache | Auth cache for CIDR permissions.
+|IdentityCache| Auth cache for `MutualTlsAuthenticator`.
+|NetworkPermissionsCache |Auth cache for network permissions.
+|PermissionsCache |Auth cache for permissions.
+|RolesCache |Auth cache for roles.
+|===
+
+[NOTE]
+.Note
+====
+AuthCache metrics MBeans are only available if corresponding authorizers and authenticators are enabled.
+====
+
 == CQL Metrics
 
 Metrics specific to CQL prepared statement caching.
--- a/src/java/org/apache/cassandra/auth/AuthCache.java
+++ b/src/java/org/apache/cassandra/auth/AuthCache.java
@@ -33,6 +33,7 @@ import java.util.function.IntConsumer;
 import java.util.function.IntSupplier;
 import java.util.function.Supplier;
 
+import com.google.common.primitives.Ints;
 import com.google.common.util.concurrent.Uninterruptibles;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -40,9 +41,14 @@ import org.slf4j.LoggerFactory;
 import com.github.benmanes.caffeine.cache.Caffeine;
 import com.github.benmanes.caffeine.cache.LoadingCache;
 import com.github.benmanes.caffeine.cache.Policy;
+import com.github.benmanes.caffeine.cache.RemovalCause;
+import org.apache.cassandra.cache.UnweightedCacheSize;
+import com.github.benmanes.caffeine.cache.stats.CacheStats;
+import com.github.benmanes.caffeine.cache.stats.StatsCounter;
 import org.apache.cassandra.concurrent.ExecutorPlus;
 import org.apache.cassandra.concurrent.ScheduledExecutors;
 import org.apache.cassandra.concurrent.Shutdownable;
+import org.apache.cassandra.metrics.UnweightedCacheMetrics;
 import org.apache.cassandra.utils.ExecutorUtils;
 import org.apache.cassandra.utils.MBeanWrapper;
 
@@ -52,13 +58,13 @@ import static org.apache.cassandra.config.CassandraRelevantProperties.AUTH_CACHE
 import static org.apache.cassandra.config.CassandraRelevantProperties.AUTH_CACHE_WARMING_RETRY_INTERVAL_MS;
 import static org.apache.cassandra.config.CassandraRelevantProperties.DISABLE_AUTH_CACHES_REMOTE_CONFIGURATION;
 
-public class AuthCache<K, V> implements AuthCacheMBean, Shutdownable
+public class AuthCache<K, V> implements AuthCacheMBean, UnweightedCacheSize, Shutdownable
 {
     private static final Logger logger = LoggerFactory.getLogger(AuthCache.class);
 
     public static final String MBEAN_NAME_BASE = "org.apache.cassandra.auth:type=";
 
-    private volatile ScheduledFuture cacheRefresher = null;
+    private volatile ScheduledFuture<?> cacheRefresher = null;
 
     // Keep a handle on created instances so their executors can be terminated cleanly
     private static final Set<Shutdownable> REGISTRY = new HashSet<>(4);
@@ -92,6 +98,8 @@ public class AuthCache<K, V> implements AuthCacheMBean, Shutdownable
     // credentials for a role couldn't be loaded without throwing an exception or serving stale
     // values until the natural expiry time.
     private final BiPredicate<K, V> invalidateCondition;
+
+    private final UnweightedCacheMetrics metrics;
 
     /**
      * @param name Used for MBean
@@ -176,6 +184,7 @@ public class AuthCache<K, V> implements AuthCacheMBean, Shutdownable
         this.bulkLoadFunction = checkNotNull(bulkLoadFunction);
         this.enableCache = checkNotNull(cacheEnabledDelegate);
         this.invalidateCondition = checkNotNull(invalidationCondition);
+        this.metrics = new UnweightedCacheMetrics(name, this);
         init();
     }
 
@@ -215,7 +224,7 @@ public class AuthCache<K, V> implements AuthCacheMBean, Shutdownable
     /**
      * Retrieve a value from the cache. Will call {@link LoadingCache#get(Object)} which will
      * "load" the value if it's not present, thus populating the key.
-     * @param k
+     * @param k key
      * @return The current value of {@code K} if cached or loaded.
      *
      * See {@link LoadingCache#get(Object)} for possible exceptions.
@@ -284,7 +293,7 @@ public class AuthCache<K, V> implements AuthCacheMBean, Shutdownable
 
     /**
      * Set maximum number of entries in the cache.
-     * @param maxEntries
+     * @param maxEntries max number of entries
      */
     public synchronized void setMaxEntries(int maxEntries)
     {
@@ -319,6 +328,11 @@ public class AuthCache<K, V> implements AuthCacheMBean, Shutdownable
         return cache == null ? 0L : cache.estimatedSize();
     }
 
+    public UnweightedCacheMetrics getMetrics()
+    {
+        return metrics;
+    }
+
     /**
      * (Re-)initialise the underlying cache. Will update validity, max entries, and update interval if
      * any have changed. The underlying {@link LoadingCache} will be initiated based on the provided {@code loadFunction}.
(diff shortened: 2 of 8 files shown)


src/java/org/apache/cassandra/cache/UnweightedCacheSize.java
```diff
@@ -0,0 +1,42 @@
...
+{
+    /**
+     * Returns the maximum number of entries in this cache.
+     */
+    int maxEntries();
+
+    /**
+     * Specifies the maximum number of entries in this cache.
+     */
+    void setMaxEntries(int maxEntries);
```
Mmuzaf, line 39: This method is not used.

src/java/org/apache/cassandra/metrics/AbstractCacheMetrics.java
```diff
@@ -0,0 +1,113 @@
...
+import com.google.common.annotations.VisibleForTesting;
+
+import com.codahale.metrics.Gauge;
+import com.codahale.metrics.Meter;
+import com.codahale.metrics.RatioGauge;
+
+import static org.apache.cassandra.metrics.CassandraMetricsRegistry.Metrics;
+
+/**
+ * Metrics for an unweighted cache like {@code AuthCache}.
```
Mmuzaf: The comment is not accurate - this class is used to hold both unweighted and weighted cache metrics.

src/java/org/apache/cassandra/metrics/UnweightedCacheMetrics.java
```diff
@@ -0,0 +1,53 @@
...
+
+import com.codahale.metrics.Gauge;
+import org.apache.cassandra.cache.UnweightedCacheSize;
+
+import static org.apache.cassandra.metrics.CassandraMetricsRegistry.Metrics;
+
+/**
+ * Metrics for an unweighted cache like {@code AuthCache}.
+ */
+public class UnweightedCacheMetrics extends AbstractCacheMetrics
```
Mmuzaf, line 28: It took me a while to figure out the difference between weighted and unweighted terms, so I assume the users will have to do the same exercise. At the moment,the `UnweightedCache` fully matches the `AuthCache`, so maybe we should classify these types of caches in this way, wdyt? This should be easier to understand for a user.

src/java/org/apache/cassandra/cache/ICache.java
```diff
@@ -23,6 +23,8 @@
  * This is similar to the Map interface, but requires maintaining a given capacity
  * and does not require put or remove to return values, which lets SerializingCache
  * be more efficient by avoiding deserialize except on get.
+ *
+ * This class represents a weighted cache.
```
Mmuzaf: There is also a cache called `ChunkCache' which also has a weighted type, but it doesn't extend the `ICache' interface. So, we either have to fix javadocs for both of these cases or for none of them. The latter case sounds easier to me as the `CacheSize` already has a description.

src/java/org/apache/cassandra/cache/ChunkCache.java
```diff
@@ -172,7 +171,7 @@ public void onRemoval(Key key, Buffer buffer, RemovalCause cause)
         buffer.release();
     }
 
-    public void close()
+    public void clear()
```
Mmuzaf, line 174: This change is not related to the issue, right?

src/java/org/apache/cassandra/metrics/UnweightedCacheMetrics.java
smiklosovic: @Mmuzaf are AuthCaches the only caches ever which will be unweighted? I think the general distinction between weighted and unweighted makes sense. Maybe we could extra document that? Othewise I think it is just ok as it is ...

src/java/org/apache/cassandra/metrics/UnweightedCacheMetrics.java
```diff
@@ -0,0 +1,53 @@
...
+     * Create metrics for given cache.
+     *
+     * @param type  Type of Cache to identify metrics
+     * @param cache Unweighted Cache to measure metrics
+     */
+    public UnweightedCacheMetrics(String type, UnweightedCacheSize cache)
+    {
+        super(new DefaultNameFactory("UnweightedCache", type), type);
+
+        maxEntries = Metrics.register(factory.createMetricName("Capacity"), cache::maxEntries);
```
Mmuzaf: Should we still keep it synced with the `CacheMetrics` names? I think we can use the "maxEntries" instead of the "Capacity" to avoid confusion.

src/java/org/apache/cassandra/cache/UnweightedCacheSize.java
smiklosovic: is not it used in `AuthCache#setMaxEntries`? It basically implements that method which is call elsewhere, like in `SetAuthCacheConfig`.

smiklosovic: hah so "problem" here is that `AuthCache` implements `AuthCacheMBean` which has a method of the same signature and then it changes the size via `setMaxEntriesDelegate`. I am not going deeper than this. I think this is just fine ... for cases when a cache would not extend MBean with such method name, we would still need a way how to change max entries etc ...

src/java/org/apache/cassandra/metrics/UnweightedCacheMetrics.java
Mmuzaf: In the future, however, weighted and unweighted may not be sufficient for us as well, so I wouldn't want to introduce a new concept at this point, as the `AuthCache` fully covers all our needs. It also sounds clearer without adding pages of documentation, and we can also change it in the future if a new type arises and we will decide to classify it... :-)

src/java/org/apache/cassandra/tools/NodeProbe.java
```diff
@@ -1748,22 +1749,22 @@ public Object getCacheMetric(String cacheType, String metricName)
...
                     return JMX.newMBeanProxy(mbeanServerConn,
-                            new ObjectName("org.apache.cassandra.metrics:type=Cache,scope=" + cacheType + ",name=" + metricName),
-                            CassandraMetricsRegistry.JmxTimerMBean.class).getMean();
+                                             new ObjectName("org.apache.cassandra.metrics:type=" + type + ",scope=" + scope + ",name=" + metricName),
+                                             CassandraMetricsRegistry.JmxTimerMBean.class).getMean();
                 case "MissLatencyUnit":
                     return JMX.newMBeanProxy(mbeanServerConn,
-                            new ObjectName("org.apache.cassandra.metrics:type=Cache,scope=" + cacheType + ",name=MissLatency"),
-                            CassandraMetricsRegistry.JmxTimerMBean.class).getDurationUnit();
+                                             new ObjectName("org.apache.cassandra.metrics:type=" + type + ",scope=" + scope + ",name=MissLatency"),
```
Mmuzaf, line 1766: Should we change `name=MissLatency` to `name=MissLatencyUnit` or am I missing something?

src/java/org/apache/cassandra/cache/ChunkCache.java
smiklosovic: I cant find any usage of that method. You are technically right, yes. But looking at it, why not to rename it to `clear`?
 The impl invalidates the cache which is kind of a clearing ... I think "clear" makes more sense and reflects what that method does than "close". That cache does not implement `(Auto)Closeable` or anything like that, there is nothing to _close_. I think that was the original idea behind this change. Keep in mind I just put it all together after Sam did his improvements.

src/java/org/apache/cassandra/metrics/UnweightedCacheMetrics.java
smiklosovic: Is this change only internally visible? If it is exposed to clients I think we need to keep it like it is to not break anything.

src/java/org/apache/cassandra/tools/NodeProbe.java
```diff
@@ -1734,10 +1734,11 @@ public long getReadRepairRepairedBackground()
     // JMX getters for the o.a.c.metrics API below.
     /**
      * Retrieve cache metrics based on the cache type (KeyCache, RowCache, or CounterCache)
-     * @param cacheType KeyCach, RowCache, or CounterCache
+     * @param type type, like Cache or UnweightedCache
+     * @param scope KeyCach, RowCache, or CounterCache
      * @param metricName Capacity, Entries, HitRate, Size, Requests or Hits.
      */
-    public Object getCacheMetric(String cacheType, String metricName)
+    public Object getCacheMetric(String type, String scope, String metricName)
```
Mmuzaf: Since the `getCacheMetric` method isn't currently used for unweighted caches, would it be better to roll back these changes and do it when the nodetool needs it? 

Another option is to create an overloaded `public Object getCacheMetric` method that takes "Cache" as the first argument, this will help us avoid changes that are not necessary right now. The weighted and unweighted caches are slightly different anyway, so `getCacheMetric` is not quite what we want (e.g. the `size` is different for weighted and unweighted).

smiklosovic: I think this is leftover from times when I was trying to expose unweighted caches as well into nodetool's `Info` so it was not always called with "Cache" but I removed that functionality from `Info` because I just felt in my guts that this is not right thing to do anyway ... Metrics are going to be exposed via CQL and adding more metrics into nodetool just feels wrong. I can revert this whole change.

src/java/org/apache/cassandra/metrics/UnweightedCacheMetrics.java
smiklosovic: I have a hard time to figure out what other metrics we would have so we could not categorize them into weighted or unweighted ones.

src/java/org/apache/cassandra/metrics/UnweightedCacheMetrics.java
Mmuzaf: This will be exposed to an end user, however, as we don't have such metrics in the public API at this moment, we can expose it with the "right" name. Even though auth caches and weighted caches have different MBeans, vtables there is no need to keep the attribute names in sync between them, so I guess we can change it.

src/java/org/apache/cassandra/metrics/UnweightedCacheMetrics.java
Mmuzaf: AuthCache metrics are valuable in their own right, though, so I'm not sure we need an additional 'unweighted' category for them. The simpler it is, the better it will be. 
We can move this question to the issue for visibility. 
@beobal I think it was your original idea to organise classes in this way. wdyt ?

beobal: I tend to agree with @smiklosovic that the general distinction between weighted and unweighted makes sense.

> In the future, however, weighted and unweighted may not be sufficient for us as well, so I wouldn't want to introduce a new concept at this point, as the AuthCache fully covers all our needs.

I don't think any new concept is being introducing here, unweighted caches exist in C* you are just exposing them via the new interface. 

I also don't agree that this requires pages of additional documentation; _some_ docs will be required to direct users to the appropriate metrics, so it seems to me a choice between 

`"For metrics related to auth caches, see the AuthCache metrics 'nameX, nameY, nameZ...'"`
vs
`"For metrics related auth caches, see the UnweightedCache metrics 'nameX, nameY, nameZ...'"`

I don't believe that docs for the metrics themselves don't need to go into great detail about the nature of weighted vs unweighted caches.

That's my $0.02, but I don't feel mega-strongly about it so wouldn't object whichever way you decide.

src/java/org/apache/cassandra/metrics/UnweightedCacheMetrics.java
smiklosovic: @Mmuzaf I would rather stay consistent and rename `maxEntries` gauge to `capacity` to be on par with `CacheMetrics` which were there first which have it `capacity` too. There is no reason to have this naming discrepancy among different types of cache metrics and term "capacity" is just perfectly fine.

smiklosovic: you know what ... you are right. `Capacity` here should be `MaxEntries`. Because `Capacity` and `MaxEntries` are not the same thing.

doc/modules/cassandra/pages/managing/operating/metrics.adoc
```diff
@@ -528,17 +528,17 @@ ViewReplicasSuccess.
 base table and when CL.ONE is achieved on view.
 |===
 
-== Cache Metrics
+== Weighted Cache Metrics
```
Mmuzaf, line 531: Should we also add a few words about 'weighted' and 'unweighted' terms? I think here is the comment with a well-described difference, we can use it:

https://issues.apache.org/jira/browse/CASSANDRA-17062?focusedCommentId=17443143&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-17443143

test/unit/org/apache/cassandra/auth/AuthCacheTest.java
```diff
@@ -322,6 +322,41 @@ public void testCacheLoaderIsNotCalledOnGetAllWhenCacheIsNotEmpty()
         assertEquals(1, loadCounter);
     }
 
+    @Test
+    public void testMetricsOnCacheEnabled()
```
Mmuzaf, line 335: Should we also check the numbers for the `entries` and `maxEntries` method calls? They do not seem to be covered by the tests.

test/unit/org/apache/cassandra/auth/AuthCacheTest.java
```diff
@@ -322,6 +322,41 @@ public void testCacheLoaderIsNotCalledOnGetAllWhenCacheIsNotEmpty()
...
+        authCache.get("11");
+
+        assertThat(authCache.getMetrics().requests.getCount()).isEqualTo(4L);
+        assertThat(authCache.getMetrics().hits.getCount()).isEqualTo(2L);
+        assertThat(authCache.getMetrics().misses.getCount()).isEqualTo(2L);
+        assertEquals(2, loadCounter);
+    }
+
+    @Test
+    public void testMetricsOnCacheDisabled()
```
Mmuzaf, line 376: Should we add a test to check that new metrics are available via the JMX API? I haven't found a good candidate for a test class, but I think it's still worth checking that `getMaxEntries` and `setMaxEntries` match, and the `entries` match the number entries we set to the cache.

src/java/org/apache/cassandra/auth/AuthCache.java
```diff
@@ -225,6 +235,7 @@ public V get(K k)
         if (cache == null)
             return loadFunction.apply(k);
 
+        metrics.requests.mark();
```
Mmuzaf: Aren't we missing the `getAll` path for tracking requests here? The `ChunkCache` itself encapsulates this logic in the `StatsCounter`.  It's worth adding such a test or a verification the metrics value for existing tests in `AuthCacheTest`.

src/java/org/apache/cassandra/cache/UnweightedCacheSize.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Represents an unweighted cache size.
+ */
+public interface UnweightedCacheSize
+{
+    /**
+     * Returns the maximum number of entries in this cache.
+     */
+    int maxEntries();
```
Mmuzaf, line 34: Should this be renamed to `getMaxEntries`?

src/java/org/apache/cassandra/auth/AuthCache.java
```diff
@@ -52,12 +58,13 @@
...
 import static org.apache.cassandra.config.CassandraRelevantProperties.DISABLE_AUTH_CACHES_REMOTE_CONFIGURATION;
 
-public class AuthCache<K, V> implements AuthCacheMBean, Shutdownable
+public class AuthCache<K, V> implements AuthCacheMBean, UnweightedCacheSize, Shutdownable
 {
     private static final Logger logger = LoggerFactory.getLogger(AuthCache.class);
 
     public static final String MBEAN_NAME_BASE = "org.apache.cassandra.auth:type=";
 
+    @SuppressWarnings("rawtypes")
```
Mmuzaf: Maybe we should use `ScheduledFuture<?>` instead?

doc/modules/cassandra/pages/managing/operating/metrics.adoc
smiklosovic: done

src/java/org/apache/cassandra/auth/AuthCache.java
smiklosovic: huh how did this happen ... I don't remember putting it there, sure, can be like suggested.

src/java/org/apache/cassandra/cache/UnweightedCacheSize.java
smiklosovic: The naming of these methods basically copies these in `CacheSize`. I would keep it like it is. In `CacheSize` there is `capacity` / `getCapacity` etc ... We would need to rework everything, `int size()`, `long weightedSize()` to prefix with `get` etc ...

src/java/org/apache/cassandra/auth/AuthCache.java
smiklosovic: Hmm ... good catch.

test/unit/org/apache/cassandra/auth/AuthCacheTest.java
smiklosovic: done

test/unit/org/apache/cassandra/auth/AuthCacheTest.java
smiklosovic: I dont think that involving _JMX_ is necessary. We basically just need to invoke same MBean method programmatically. I have added "testSettingMaxEntries" test for that.

Mmuzaf (approved):
LGTM, +1

PR #3011, opened 2024-01-02
CASSANDRA-18939 5.0 Creating a SASI index after creating an SAI index breaks secondary index queries

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/index/sai/plan/QueryController.java
+++ b/src/java/org/apache/cassandra/index/sai/plan/QueryController.java
@@ -23,8 +23,9 @@ import java.io.UncheckedIOException;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
-import java.util.Set;
 import java.util.stream.Collectors;
+
+import javax.annotation.Nullable;
 
 import com.google.common.collect.Lists;
 
@@ -131,10 +132,10 @@ public class QueryController
         return ranges;
     }
 
+    @Nullable
     public StorageAttachedIndex indexFor(RowFilter.Expression expression)
     {
-        Set<StorageAttachedIndex> indexes = cfs.indexManager.getBestIndexFor(expression, StorageAttachedIndex.class);
-        return indexes.isEmpty() ? null : indexes.iterator().next();
+        return cfs.indexManager.getBestIndexFor(expression, StorageAttachedIndex.class).orElse(null);
     }
 
     public boolean hasAnalyzer(RowFilter.Expression expression)
--- a/src/java/org/apache/cassandra/index/sai/plan/VectorTopKProcessor.java
+++ b/src/java/org/apache/cassandra/index/sai/plan/VectorTopKProcessor.java
@@ -20,7 +20,6 @@ package org.apache.cassandra.index.sai.plan;
 
 import java.nio.ByteBuffer;
 import java.util.Comparator;
-import java.util.Optional;
 import java.util.PriorityQueue;
 import java.util.TreeMap;
 import java.util.TreeSet;
@@ -41,7 +40,6 @@ import org.apache.cassandra.db.partitions.UnfilteredPartitionIterator;
 import org.apache.cassandra.db.rows.BaseRowIterator;
 import org.apache.cassandra.db.rows.Row;
 import org.apache.cassandra.db.rows.Unfiltered;
-import org.apache.cassandra.index.Index;
 import org.apache.cassandra.index.SecondaryIndexManager;
 import org.apache.cassandra.index.sai.StorageAttachedIndex;
 import org.apache.cassandra.index.sai.utils.InMemoryPartitionIterator;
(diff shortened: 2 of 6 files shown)


src/java/org/apache/cassandra/index/sai/plan/VectorTopKProcessor.java
```diff
@@ -187,7 +185,6 @@ private StorageAttachedIndex findVectorIndexFor(SecondaryIndexManager sim, RowFi
         if (e.operator() != Operator.ANN)
             return null;
 
-        Optional<Index> index = sim.getBestIndexFor(e);
-        return (StorageAttachedIndex) index.filter(i -> i instanceof StorageAttachedIndex).orElse(null);
+        return sim.getBestIndexFor(e, StorageAttachedIndex.class).orElse(null);
```
smiklosovic, line 188: @adelapena This seems like a great simplification, I am not sure why the original implementator chose this approach. I think that the version of `getBestIndexFor` accepting class was there already when this was implemented or I am not getting something ...

src/java/org/apache/cassandra/index/sasi/plan/QueryController.java
```diff
@@ -92,11 +91,9 @@ public AbstractType<?> getKeyValidator()
 
     public ColumnIndex getIndex(RowFilter.Expression expression)
     {
-        Optional<Index> index = cfs.indexManager.getBestIndexFor(expression);
-        return index.isPresent() ? ((SASIIndex) index.get()).getIndex() : null;
+        return cfs.indexManager.getBestIndexFor(expression, SASIIndex.class).map(SASIIndex::getIndex).orElse(null);
```
adelapena, line 97: Not related with the changes, but can we annotate the method as `@Nullable`?

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
```diff
@@ -133,8 +132,7 @@ public List<DataRange> dataRanges()
 
     public StorageAttachedIndex indexFor(RowFilter.Expression expression)
     {
-        Set<StorageAttachedIndex> indexes = cfs.indexManager.getBestIndexFor(expression, StorageAttachedIndex.class);
-        return indexes.isEmpty() ? null : indexes.iterator().next();
+        return cfs.indexManager.getBestIndexFor(expression, StorageAttachedIndex.class).orElse(null);
```
adelapena, line 138: Not related with the changes, but can we annotate the method as `@Nullable`?

test/unit/org/apache/cassandra/index/SecondaryIndexManagerTest.java
```diff
@@ -56,6 +57,20 @@ public void after()
...
+    {
+        execute(format("CREATE TABLE %s.abc (id int PRIMARY KEY, val text)", KEYSPACE));
+        execute(format("CREATE INDEX ON %s.abc (val) USING 'sai'", KEYSPACE));
+        waitForTableIndexesQueryable(KEYSPACE, "abc");
+        execute(format("INSERT INTO %s.abc (id, val) VALUES (1, 'a')", KEYSPACE));
+        execute(format("SELECT * FROM %s.abc WHERE val = 'a'", KEYSPACE));
+        flush(KEYSPACE);
+        execute(format("CREATE CUSTOM INDEX on %s.abc (val) USING 'org.apache.cassandra.index.sasi.SASIIndex'", KEYSPACE));
+        waitForTableIndexesQueryable(KEYSPACE, "abc");
+        execute(format("SELECT * FROM %s.abc WHERE val = 'a'", KEYSPACE));
```
adelapena: I think this can be simplified to:
```suggestion
        createTable("CREATE TABLE %s (id int PRIMARY KEY, val text)");
        createIndex("CREATE INDEX ON %s (val) USING 'sai'");
        execute("INSERT INTO %s (id, val) VALUES (1, 'a')");
        execute("SELECT * FROM %s WHERE val = 'a'");
        flush();
        createIndex("CREATE INDEX ON %s (val) USING 'org.apache.cassandra.index.sasi.SASIIndex'");
        execute("SELECT * FROM %s WHERE val = 'a'");
```

adelapena (commented):
Looks good to me, I've just left a couple of nits.

test/unit/org/apache/cassandra/index/SecondaryIndexManagerTest.java
adelapena: It seems that a minor issue with the test is that it doesn't consistently reproduce the bug. It would only reproduce it with the old code if the concurrent map containing the indexes happens to return the SAI index first. This doesn't seem deterministic, and in fact, the test sometimes passes without the fix. This is not a big deal IMO, because the test aims at a very specific bug that doesn't exist anymore.

src/java/org/apache/cassandra/index/sai/plan/VectorTopKProcessor.java
adelapena: I don't know what the history behind this is. Maybe the idea for returning a set of indexes was delegating the index selection to the specific implementation? In any case, this new approach seems simpler and less error-prone for the current usage.

maedhroz (approved):

PR #3013, opened 2024-01-02
CASSANDRA-19245 5.0: Upgrade Python driver to 3.29.0

adelapena:


bin/cqlsh.py
```diff
@@ -21,8 +21,8 @@
 import sys
 from glob import glob
 
-if sys.version_info < (3, 7):
-    sys.exit("\ncqlsh requires Python 3.7+\n")
+if sys.version_info < (3, 8) or sys.version_info >= (3, 12):
```
bereng, line 24: 12 o 11?

adelapena: 12. The max supported version is Python 3.11.x, so we use `>= (3, 12)` to accept minors.

PR #3029, opened 2024-01-05
CASSANDRA-18714 CQLSSTableWriter writes SAI index

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -200,22 +200,24 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean, Memtable.Owner
     are finished. By having flushExecutor size the same size as each of the perDiskflushExecutors we make sure we can
     have that many flushes going at the same time.
     */
-    private static final ExecutorPlus flushExecutor = executorFactory()
-            .withJmxInternal()
-            .pooled("MemtableFlushWriter", getFlushWriters());
+    private static final ExecutorPlus flushExecutor = DatabaseDescriptor.isDaemonInitialized() 
+                                                      ? executorFactory().withJmxInternal().pooled("MemtableFlushWriter", getFlushWriters())
+                                                      : null;
 
     // post-flush executor is single threaded to provide guarantee that any flush Future on a CF will never return until prior flushes have completed
-    private static final ExecutorPlus postFlushExecutor = executorFactory()
-            .withJmxInternal()
-            .sequential("MemtablePostFlush");
+    private static final ExecutorPlus postFlushExecutor = DatabaseDescriptor.isDaemonInitialized()
+                                                          ? executorFactory().withJmxInternal().sequential("MemtablePostFlush")
+                                                          : null;
 
-    private static final ExecutorPlus reclaimExecutor = executorFactory()
-            .withJmxInternal()
-            .sequential("MemtableReclaimMemory");
+    private static final ExecutorPlus reclaimExecutor = DatabaseDescriptor.isDaemonInitialized()
+                                                        ? executorFactory().withJmxInternal().sequential("MemtableReclaimMemory")
+                                                        : null;
 
-    private static final PerDiskFlushExecutors perDiskflushExecutors = new PerDiskFlushExecutors(DatabaseDescriptor.getFlushWriters(),
-                                                                                                 DatabaseDescriptor.getNonLocalSystemKeyspacesDataFileLocations(),
-                                                                                                 DatabaseDescriptor.useSpecificLocationForLocalSystemData());
+    private static final PerDiskFlushExecutors perDiskflushExecutors = DatabaseDescriptor.isDaemonInitialized()
+                                                                       ? new PerDiskFlushExecutors(DatabaseDescriptor.getFlushWriters(),
+                                                                                                  DatabaseDescriptor.getNonLocalSystemKeyspacesDataFileLocations(),
+                                                                                                  DatabaseDescriptor.useSpecificLocationForLocalSystemData())
+                                                                       : null;
 
     /**
      * Reason for initiating a memtable flush.
@@ -394,7 +396,9 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean, Memtable.Owner
         indexManager.reload();
 
         memtableFactory = metadata().params.memtable.factory();
-        switchMemtableOrNotify(FlushReason.SCHEMA_CHANGE, Memtable::metadataUpdated);
+
+        if (DatabaseDescriptor.isDaemonInitialized())
+            switchMemtableOrNotify(FlushReason.SCHEMA_CHANGE, Memtable::metadataUpdated);
     }
 
     public static Runnable getBackgroundCompactionTaskSubmitter()
@@ -855,26 +859,53 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean, Memtable.Owner
         sstableImporter.importNewSSTables(options);
     }
 
-    /**
-     * #{@inheritDoc}
-     */
-    public synchronized List<String> importNewSSTables(Set<String> srcPaths, boolean resetLevel, boolean clearRepaired, boolean verifySSTables, boolean verifyTokens, boolean invalidateCaches, boolean extendedVerify, boolean copyData)
+    @Override
+    public List<String> importNewSSTables(Set<String> srcPaths, boolean resetLevel, boolean clearRepaired,
+                                          boolean verifySSTables, boolean verifyTokens, boolean invalidateCaches,
+                                          boolean extendedVerify, boolean copyData)
     {
-        SSTableImporter.Options options = SSTableImporter.Options.options(srcPaths)
-                                                                 .resetLevel(resetLevel)
-                                                                 .clearRepaired(clearRepaired)
-                                                                 .verifySSTables(verifySSTables)
-                                                                 .verifyTokens(verifyTokens)
-                                                                 .invalidateCaches(invalidateCaches)
-                                                                 .extendedVerify(extendedVerify)
-                                                                 .copyData(copyData).build();
-
-        return sstableImporter.importNewSSTables(options);
+        return sstableImporter.importNewSSTables(SSTableImporter.Options.options(srcPaths)
+                                                                        .resetLevel(resetLevel)
+                                                                        .clearRepaired(clearRepaired)
+                                                                        .verifySSTables(verifySSTables)
+                                                                        .verifyTokens(verifyTokens)
+                                                                        .invalidateCaches(invalidateCaches)
+                                                                        .extendedVerify(extendedVerify)
+                                                                        .copyData(copyData).build());
     }
 
-    public List<String> importNewSSTables(Set<String> srcPaths, boolean resetLevel, boolean clearRepaired, boolean verifySSTables, boolean verifyTokens, boolean invalidateCaches, boolean extendedVerify)
+    @Override
+    public List<String> importNewSSTables(Set<String> srcPaths, boolean resetLevel, boolean clearRepaired,
+                                          boolean verifySSTables, boolean verifyTokens, boolean invalidateCaches,
+                                          boolean extendedVerify)
     {
-        return importNewSSTables(srcPaths, resetLevel, clearRepaired, verifySSTables, verifyTokens, invalidateCaches, extendedVerify, false);
+        return sstableImporter.importNewSSTables(SSTableImporter.Options.options(srcPaths)
+                                                                        .resetLevel(resetLevel)
+                                                                        .clearRepaired(clearRepaired)
+                                                                        .verifySSTables(verifySSTables)
+                                                                        .verifyTokens(verifyTokens)
+                                                                        .invalidateCaches(invalidateCaches)
+                                                                        .extendedVerify(extendedVerify)
+                                                                        .build());
+    }
+
+    @Override
+    public List<String> importNewSSTables(Set<String> srcPaths, boolean resetLevel, boolean clearRepaired,
+                                          boolean verifySSTables, boolean verifyTokens, boolean invalidateCaches,
+                                          boolean extendedVerify, boolean copyData, boolean failOnMissingIndex,
+                                          boolean validateIndexChecksum)
+    {
+        return sstableImporter.importNewSSTables(SSTableImporter.Options.options(srcPaths)
+                                                                        .resetLevel(resetLevel)
+                                                                        .clearRepaired(clearRepaired)
+                                                                        .verifySSTables(verifySSTables)
+                                                                        .verifyTokens(verifyTokens)
+                                                                        .invalidateCaches(invalidateCaches)
+                                                                        .extendedVerify(extendedVerify)
+                                                                        .failOnMissingIndex(failOnMissingIndex)
+                                                                        .validateIndexChecksum(validateIndexChecksum)
+                                                                        .copyData(copyData)
+                                                                        .build());
     }
 
     Descriptor getUniqueDescriptorFor(Descriptor descriptor, File targetDirectory)
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStoreMBean.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStoreMBean.java
@@ -189,12 +189,12 @@ public interface ColumnFamilyStoreMBean
     /** @deprecated See CASSANDRA-16407 */
     @Deprecated(since = "4.0")
     public List<String> importNewSSTables(Set<String> srcPaths,
-                                           boolean resetLevel,
-                                           boolean clearRepaired,
-                                           boolean verifySSTables,
-                                           boolean verifyTokens,
-                                           boolean invalidateCaches,
-                                           boolean extendedVerify);
+                                          boolean resetLevel,
+                                          boolean clearRepaired,
+                                          boolean verifySSTables,
+                                          boolean verifyTokens,
+                                          boolean invalidateCaches,
+                                          boolean extendedVerify);
 
     /**
      * Load new sstables from the given directory
@@ -210,6 +210,8 @@ public interface ColumnFamilyStoreMBean
      *
      * @return list of failed import directories
      */
+    /** @deprecated See CASSANDRA-18714 */
+    @Deprecated(since = "5.0")
     public List<String> importNewSSTables(Set<String> srcPaths,
                                           boolean resetLevel,
                                           boolean clearRepaired,
@@ -218,6 +220,33 @@ public interface ColumnFamilyStoreMBean
                                           boolean invalidateCaches,
                                           boolean extendedVerify,
                                           boolean copyData);
+
+    /**
+     * Load new sstables from the given directory
+     *
+     * @param srcPaths the path to the new sstables - if it is an empty set, the data directories will be scanned
+     * @param resetLevel if the level should be reset to 0 on the new sstables
+     * @param clearRepaired if repaired info should be wiped from the new sstables
+     * @param verifySSTables if the new sstables should be verified that they are not corrupt
+     * @param verifyTokens if the tokens in the new sstables should be verified that they are owned by the current node
+     * @param invalidateCaches if row cache should be invalidated for the keys in the new sstables
+     * @param extendedVerify if we should run an extended verify checking all values in the new sstables
+     * @param copyData if we should copy data from source paths instead of moving them
+     * @param validateIndexChecksum if we should also validate checksum for SAI indexes
+     * @param failOnMissingIndex if loading should fail when SSTables do not contain built SAI indexes too
+     *
+     * @return list of failed import directories
+     */
+    public List<String> importNewSSTables(Set<String> srcPaths,
+                                          boolean resetLevel,
+                                          boolean clearRepaired,
+                                          boolean verifySSTables,
+                                          boolean verifyTokens,
+                                          boolean invalidateCaches,
+                                          boolean extendedVerify,
+                                          boolean copyData,
+                                          boolean failOnMissingIndex,
+                                          boolean validateIndexChecksum);
 
     /** @deprecated See CASSANDRA-6719 */
     @Deprecated(since = "4.0")
--- a/src/java/org/apache/cassandra/db/SSTableImporter.java
+++ b/src/java/org/apache/cassandra/db/SSTableImporter.java
@@ -32,6 +32,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.cassandra.db.lifecycle.SSTableSet;
+import org.apache.cassandra.index.Index;
+import org.apache.cassandra.index.sai.StorageAttachedIndexGroup;
+import org.apache.cassandra.index.sai.disk.format.IndexDescriptor;
+import org.apache.cassandra.index.sai.utils.IndexIdentifier;
 import org.apache.cassandra.io.sstable.Component;
 import org.apache.cassandra.io.sstable.Descriptor;
 import org.apache.cassandra.io.sstable.IVerifier;
@@ -83,7 +87,7 @@ public class SSTableImporter
         List<String> failedDirectories = new ArrayList<>();
 
         // verify first to avoid starting to copy sstables to the data directories and then have to abort.
-        if (options.verifySSTables || options.verifyTokens)
+        if (options.verifySSTables || options.verifyTokens || options.failOnMissingIndex)
         {
             for (Pair<Directories.SSTableLister, String> listerPair : listers)
             {
@@ -97,7 +101,39 @@ public class SSTableImporter
                         try
                         {
                             abortIfDraining();
-                            verifySSTableForImport(descriptor, entry.getValue(), options.verifyTokens, options.verifySSTables, options.extendedVerify);
+
+                            if (options.failOnMissingIndex)
+                            {
+                                Index.Group saiIndexGroup = cfs.indexManager.getIndexGroup(StorageAttachedIndexGroup.GROUP_KEY);
+                                if (saiIndexGroup != null)
+                                {
+                                    IndexDescriptor indexDescriptor = IndexDescriptor.create(descriptor,
+                                                                                             cfs.getPartitioner(),
+                                                                                             cfs.metadata().comparator);
+
+                                    String keyspace = cfs.getKeyspaceName();
+                                    String table = cfs.getTableName();
+
+                                    if (!indexDescriptor.isPerSSTableIndexBuildComplete())
+                                        throw new IllegalStateException(String.format("Missing SAI index to import for SSTable %s on %s.%s",
+                                                                                      indexDescriptor.sstableDescriptor.toString(),
+                                                                                      keyspace,
+                                                                                      table));
+
+                                    for (Index index : saiIndexGroup.getIndexes())
+                                    {
+                                        IndexIdentifier indexIdentifier = new IndexIdentifier(keyspace, table, index.getIndexMetadata().name);
+                                        if (!indexDescriptor.isPerColumnIndexBuildComplete(indexIdentifier))
+                                            throw new IllegalStateException(String.format("Missing SAI index to import for index %s on %s.%s",
+                                                                                          index.getIndexMetadata().name,
+                                                                                          keyspace,
+                                                                                          table));
+                                    }
+                                }
+                            }
+
+                            if (options.verifySSTables || options.verifyTokens)
+                                verifySSTableForImport(descriptor, entry.getValue(), options.verifyTokens, options.verifySSTables, options.extendedVerify);
                         }
                         catch (Throwable t)
                         {
@@ -188,7 +224,7 @@ public class SSTableImporter
             abortIfDraining();
 
             // Validate existing SSTable-attached indexes, and then build any that are missing:
-            if (!cfs.indexManager.validateSSTableAttachedIndexes(newSSTables, false))
+            if (!cfs.indexManager.validateSSTableAttachedIndexes(newSSTables, false, options.validateIndexChecksum))
                 cfs.indexManager.buildSSTableAttachedIndexesBlocking(newSSTables);
 
             cfs.getTracker().addSSTables(newSSTables);
@@ -441,8 +477,13 @@ public class SSTableImporter
         private final boolean invalidateCaches;
         private final boolean extendedVerify;
         private final boolean copyData;
+        private final boolean failOnMissingIndex;
+        public final boolean validateIndexChecksum;
 
-        public Options(Set<String> srcPaths, boolean resetLevel, boolean clearRepaired, boolean verifySSTables, boolean verifyTokens, boolean invalidateCaches, boolean extendedVerify, boolean copyData)
+        public Options(Set<String> srcPaths, boolean resetLevel, boolean clearRepaired,
+                       boolean verifySSTables, boolean verifyTokens, boolean invalidateCaches,
+                       boolean extendedVerify, boolean copyData, boolean failOnMissingIndex,
+                       boolean validateIndexChecksum)
         {
             this.srcPaths = srcPaths;
             this.resetLevel = resetLevel;
@@ -452,6 +493,8 @@ public class SSTableImporter
             this.invalidateCaches = invalidateCaches;
             this.extendedVerify = extendedVerify;
             this.copyData = copyData;
+            this.failOnMissingIndex = failOnMissingIndex;
+            this.validateIndexChecksum = validateIndexChecksum;
         }
 
         public static Builder options(String srcDir)
@@ -481,6 +524,8 @@ public class SSTableImporter
                    ", invalidateCaches=" + invalidateCaches +
                    ", extendedVerify=" + extendedVerify +
                    ", copyData= " + copyData +
+                   ", failOnMissingIndex= " + failOnMissingIndex +
+                   ", validateIndexChecksum= " + validateIndexChecksum +
                    '}';
         }
 
@@ -494,6 +539,8 @@ public class SSTableImporter
             private boolean invalidateCaches = false;
             private boolean extendedVerify = false;
             private boolean copyData = false;
+            private boolean failOnMissingIndex = false;
+            private boolean validateIndexChecksum = true;
 
             private Builder(Set<String> srcPath)
             {
@@ -543,9 +590,24 @@ public class SSTableImporter
                 return this;
             }
 
+            public Builder failOnMissingIndex(boolean value)
+            {
+                failOnMissingIndex = value;
+                return this;
+            }
+
+            public Builder validateIndexChecksum(boolean value)
+            {
+                validateIndexChecksum = value;
+                return this;
+            }
+
             public Options build()
             {
-                return new Options(srcPaths, resetLevel, clearRepaired, verifySSTables, verifyTokens, invalidateCaches, extendedVerify, copyData);
+                return new Options(srcPaths, resetLevel, clearRepaired,
+                                   verifySSTables, verifyTokens, invalidateCaches,
+                                   extendedVerify, copyData, failOnMissingIndex,
+                                   validateIndexChecksum);
             }
         }
     }
--- a/src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
+++ b/src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
@@ -343,7 +343,7 @@ public class StorageAttachedIndexGroup implements Index.Group, INotificationCons
     }
 
     @Override
-    public boolean validateSSTableAttachedIndexes(Collection<SSTableReader> sstables, boolean throwOnIncomplete)
+    public boolean validateSSTableAttachedIndexes(Collection<SSTableReader> sstables, boolean throwOnIncomplete, boolean validateChecksum)
     {
         boolean complete = true;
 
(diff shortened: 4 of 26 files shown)


src/java/org/apache/cassandra/db/SSTableImporter.java
```diff
@@ -97,7 +99,19 @@ synchronized List<String> importNewSSTables(Options options)
...
                             abortIfDraining();
-                            verifySSTableForImport(descriptor, entry.getValue(), options.verifyTokens, options.verifySSTables, options.extendedVerify);
+
+                            if (options.verifySSTables || options.verifyTokens)
+                                verifySSTableForImport(descriptor, entry.getValue(), options.verifyTokens, options.verifySSTables, options.extendedVerify);
+
+                            if (options.failOnMissingIndex)
+                            {
+                                IndexDescriptor indexDescriptor = IndexDescriptor.create(descriptor,
+                                                                                         cfs.getPartitioner(),
```
smiklosovic: keep in mind that for SAI, the only partitioner currently supported is Murmur3Partitioner and it needs to be explicitly set on `CQLSSTableWriter` by `withPartitioner` (visible from tests in `CQLSSTableWriterTest`). 

Anyway, I still think that this is the correct way to do this. We should not hardcode `Murmur3Partitioner.instance` here even it is the only one which makes sense.

src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
```diff
@@ -80,11 +80,11 @@ public final class CreateIndexStatement extends AlterSchemaStatement
     public static final String KEYSPACE_DOES_NOT_MATCH_INDEX = "Keyspace name '%s' doesn't match index name '%s'";
     public static final String MUST_SPECIFY_INDEX_IMPLEMENTATION = "Must specify index implementation via USING";
 
-    private final String indexName;
-    private final String tableName;
-    private final List<IndexTarget.Raw> rawIndexTargets;
-    private final IndexAttributes attrs;
-    private final boolean ifNotExists;
+    public final String indexName;
```
smiklosovic: I need to have this public in `CQLSSTableWriter.createIndexes`

src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
```diff
@@ -618,6 +654,11 @@ public CQLSSTableWriter build()
                     Types types = createTypes(keyspaceName);
                     Schema.instance.transform(SchemaTransformations.addTypes(types, true));
                     tableMetadata = createTable(types);
+
+                    Indexes indexes = createIndexes(tableMetadata);
```
smiklosovic: I need non-null `tableMetadata` first to create `Indexes`. So then I have to unbuild that to set indexes to metadata afterwards.

src/java/org/apache/cassandra/tools/NodeProbe.java
```diff
@@ -1544,9 +1544,15 @@ public void loadNewSSTables(String ksName, String cfName)
         ssProxy.loadNewSSTables(ksName, cfName);
     }
 
-    public List<String> importNewSSTables(String ksName, String cfName, Set<String> srcPaths, boolean resetLevel, boolean clearRepaired, boolean verifySSTables, boolean verifyTokens, boolean invalidateCaches, boolean extendedVerify, boolean copyData)
-    {
-        return getCfsProxy(ksName, cfName).importNewSSTables(srcPaths, resetLevel, clearRepaired, verifySSTables, verifyTokens, invalidateCaches, extendedVerify, copyData);
+    public List<String> importNewSSTables(String ksName, String cfName, Set<String> srcPaths, boolean resetLevel,
```
smiklosovic, line 1547: I created new method instead of modifying the old one to be backward compatible / not break stuff for people using the older version already.

src/java/org/apache/cassandra/tools/nodetool/Import.java
```diff
@@ -80,6 +80,16 @@ public class Import extends NodeToolCmd
             description = "Copy data from source directories instead of moving them")
     private boolean copyData = false;
 
+    @Option(title = "fail_on_missing_index",
+            name = {"-f", "--fail-on-missing-index"},
```
smiklosovic: I am open to different flags here, `-f` and `-i` just makes sense to me.

src/java/org/apache/cassandra/db/SSTableImporter.java
```diff
@@ -97,7 +99,19 @@ synchronized List<String> importNewSSTables(Options options)
                         try
                         {
                             abortIfDraining();
-                            verifySSTableForImport(descriptor, entry.getValue(), options.verifyTokens, options.verifySSTables, options.extendedVerify);
+
+                            if (options.verifySSTables || options.verifyTokens)
+                                verifySSTableForImport(descriptor, entry.getValue(), options.verifyTokens, options.verifySSTables, options.extendedVerify);
+
+                            if (options.failOnMissingIndex)
```
smiklosovic, line 105: I put it here to be still in the "verification" stage of the import so we can fail fast before any sstables are actually imported. Also, I moved it before other verification in order to fail as fast as possible because just checking if index was built is way faster than `verifySSTableForImport` afterwards.

src/java/org/apache/cassandra/db/ColumnFamilyStoreMBean.java
```diff
@@ -219,6 +221,33 @@ public List<String> importNewSSTables(Set<String> srcPaths,
...
+     * @param verifyTokens if the tokens in the new sstables should be verified that they are owned by the current node
+     * @param invalidateCaches if row cache should be invalidated for the keys in the new sstables
+     * @param extendedVerify if we should run an extended verify checking all values in the new sstables
+     * @param copyData if we should copy data from source paths instead of moving them
+     * @param validateIndexChecksum if we should also validate checksum for SAI indexes
+     * @param failOnMissingIndex if loading should fail when SSTables do not contain built SAI indexes too
+     *
+     * @return list of failed import directories
+     */
+    public List<String> importNewSSTables(Set<String> srcPaths,
```
smiklosovic, line 240: I added a new method and deprecated the old one to preserve backward compatibility.

src/java/org/apache/cassandra/db/SSTableImporter.java
```diff
@@ -97,7 +99,19 @@ synchronized List<String> importNewSSTables(Options options)
                         try
                         {
                             abortIfDraining();
-                            verifySSTableForImport(descriptor, entry.getValue(), options.verifyTokens, options.verifySSTables, options.extendedVerify);
+
+                            if (options.failOnMissingIndex && cfs.indexManager.getIndexGroup(StorageAttachedIndexGroup.GROUP_KEY) != null)
```
smiklosovic: we also enter this `if` only if there are some SAI indexes for this table present ... obviously.

test/unit/org/apache/cassandra/db/ImportTest.java
```diff
@@ -710,6 +710,154 @@ public void importExoticTableNamesTest() throws Throwable
...
+                backupDir.deleteRecursive();
+                backupDir.parent().deleteRecursive();
+            }
+
+            execute(String.format("DROP TABLE IF EXISTS %s.%s", KEYSPACE, "sai_test"));
+        }
+    }
+
+    @Test
+    public void skipIndexChecksumOnSAITest() throws Throwable
```
smiklosovic, line 841: It is a little bit tricky this was actually skipped, all I am doing is to just be sure that it does not fail. It is easy to see that it is skipped by debugging this manually.

src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
```diff
@@ -457,6 +467,20 @@ public Builder forTable(String schema)
             return this;
         }
 
+        /**
+         * The schema (CREATE INDEX statement) for index to be created for the table. Only SAI indexes are supported.
+         *
+         * @param indexes CQL statements representing SAI indexes to be created.
+         * @return this builder
+         */
+        public Builder forIndex(String... indexes)
```
JeetKunDoug: NIT: I think this should be named `withIndexes` rather than `forIndex`

src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
JeetKunDoug: with the `applyIndexes` change suggested in CQLSSTableWriter, you don't need these to be public any more.

src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
```diff
@@ -648,6 +696,29 @@ private Types createTypes(String keyspace)
...
+                                                                          .collect(Collectors.toList());
+
+                builder.add(IndexMetadata.fromIndexTargets(indexTargets,
+                                                           statement.indexName,
+                                                           IndexMetadata.Kind.CUSTOM,
+                                                           statement.attrs.getOptions()));
+            }
+
+            return builder.build();
+        }
```
JeetKunDoug: If we in stead do something like `applyIndexes` we can do this in a way I think is a bit cleaner, and will pick up any changes to the index statement logic, by using `CreateIndexStatement#apply`

```suggestion
                /**
         * Applies any provided index definitions to the target table
         * @param ksm the KeyspaceMetadata object that has the table defined
         * @return an updated TableMetadata instance with the indexe create statements applied
         */
        private TableMetadata applyIndexes(KeyspaceMetadata ksm)
        {
            ClientState state = ClientState.forInternalCalls();
            Keyspaces keyspaces = Keyspaces.of(ksm);
            for (CreateIndexStatement.Raw statement: indexStatements)
            {
                keyspaces = statement.prepare(state).apply(keyspaces);
            }
            return keyspaces.get(ksm.name).get()
                   .tables.get(schemaStatement.table()).get();
        }
```

And then, in `#build` we end up with something like this:


``` java
                if (tableMetadata == null)
                {
                    Types types = createTypes(keyspaceName);
                    Schema.instance.transform(SchemaTransformations.addTypes(types, true));
                    tableMetadata = createTable(types);
                    if (!indexStatements.isEmpty())
                        tableMetadata = applyIndexes(ksm.withSwapped(ksm.tables.with(tableMetadata)));
                    Schema.instance.transform(SchemaTransformations.addTable(tableMetadata, true));
                }
```

src/java/org/apache/cassandra/tools/nodetool/Import.java
JeetKunDoug: What do you think about `-ri`/`--require-index-component` and `-siv`/`--skip-index-validation`

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -859,6 +859,22 @@ public void loadNewSSTables()
      * #{@inheritDoc}
      */
     public synchronized List<String> importNewSSTables(Set<String> srcPaths, boolean resetLevel, boolean clearRepaired, boolean verifySSTables, boolean verifyTokens, boolean invalidateCaches, boolean extendedVerify, boolean copyData)
+    {
+        return importNewSSTables(srcPaths, resetLevel, clearRepaired, verifySSTables, verifyTokens, invalidateCaches, extendedVerify, false, false, true);
```
JeetKunDoug: This lost the `copyData` value - I wonder if it wouldn't be better to just duplicate the building of the import options and call to `sstableImporter#importNewSSTables` just to get rid of the huge number of parameters that can be easily mixed up here, since almost all of them are booleans?

If we do this, each method should simply set the Options it needs and leave any option that isn't specified in the method signature alone, so the default from `SSTableImporter#Options` is used.

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -859,6 +859,22 @@ public void loadNewSSTables()
      * #{@inheritDoc}
      */
     public synchronized List<String> importNewSSTables(Set<String> srcPaths, boolean resetLevel, boolean clearRepaired, boolean verifySSTables, boolean verifyTokens, boolean invalidateCaches, boolean extendedVerify, boolean copyData)
```
JeetKunDoug: Does this need to be `synchronized`? The `sstableImporter#importNewSSTables` call already synchronizes on the importer itself, and none of the other methods are synchronized, so I think we can remove this... 

Also, a nit - all of them should have `@Override` annotations per the style guide.

src/java/org/apache/cassandra/db/SSTableImporter.java
JeetKunDoug: Agreed - hard-coding the partitioner here would be bad.

src/java/org/apache/cassandra/tools/NodeProbe.java
JeetKunDoug: So your comment says you added a new one, but it looks like you modified the old one... I don't know if there's anything other than the `Import` command that actually uses it, but wanted to check if it was intentional that this comment doesn't match the code - or did you mean you made a new one in `ColumnFamilyStore/ColumnFamilyStoreMBean` which would also make sense, and which you did... so maybe I just misunderstood this PR comment?

JeetKunDoug (commented):
Thanks so much for picking up this JIRA ticket, and already addressing the review comments I had posted earlier. Generally looks great, but I think there's one area where there's a bug introduced (in ColumnFamilyStore#importNewSSTables, where we lost the value of the `copyData` parameter in one of the overloads).

src/java/org/apache/cassandra/tools/nodetool/Import.java
maedhroz: +1 to `-siv`/`--skip-index-validation`

Still thinking about the other...

src/java/org/apache/cassandra/tools/nodetool/Import.java
```diff
@@ -80,6 +80,16 @@ public class Import extends NodeToolCmd
...
+            description = "Fail import when SSTables do not contain SAI components")
+    private boolean failOnMissingIndex = false;
+
+    @Option(title = "no_validate_index_checksum",
+            name = {"-i", "--no-validate-index-checksum"},
+            description = "Do not validate checksum of indexes")
```
maedhroz:
```suggestion
            description = "Skip SSTable-attached index checksum validation")
```
Works w/ Doug's naming suggestion and is a tiny bit more specific (as I'm guessing this is only ever going to apply to SAI-like indexes that are SSTable-attached)

src/java/org/apache/cassandra/tools/nodetool/Import.java
```diff
@@ -80,6 +80,16 @@ public class Import extends NodeToolCmd
             description = "Copy data from source directories instead of moving them")
     private boolean copyData = false;
 
+    @Option(title = "fail_on_missing_index",
+            name = {"-f", "--fail-on-missing-index"},
+            description = "Fail import when SSTables do not contain SAI components")
```
maedhroz:
```suggestion
            description = "Require existing index components for SSTables w/ attached indexes")
```
Trying something that works w/ Doug's flag name in case we change it.

src/java/org/apache/cassandra/tools/nodetool/Import.java
maedhroz: I'm fine w/ `--require-index-component` too, although I think it should be `--require-index-components`. (An SAI index is really a bag of separate components that enable related functionality, but perhaps I'm too in the weeds...)

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
maedhroz: If we want to keep our behavior in line with the pre-patch code, we would move the `synchronized` to the overload below on line 874. However, as Doug mentions, the importer is already synchronizing on its `importNewSSTables()` method. We can just remove it.

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
maedhroz: Yeah, +100 to just building a `SSTableImporter.Options` as early as possible.

test/unit/org/apache/cassandra/db/ImportTest.java
maedhroz: I guess we could just corrupt one of the column index files and then try to issue a query that matches all 10 rows after import?

test/unit/org/apache/cassandra/db/ImportTest.java
```diff
@@ -710,6 +710,154 @@ public void importExoticTableNamesTest() throws Throwable
...
+            File backupDir = moveToBackupDir(sstables);
+
+            assertEquals(0, execute(String.format("SELECT * FROM %s.%s", KEYSPACE, "sai_test")).size());
+
+            SSTableImporter importer = new SSTableImporter(cfs);
+            SSTableImporter.Options options = SSTableImporter.Options.options(backupDir.toString())
+                                                                     .copyData(true)
+                                                                     .failOnMissingIndex(true)
+                                                                     .build();
+            assertTrue(importer.importNewSSTables(options).isEmpty());
```
maedhroz, line 750: nit: Perhaps throw in an assertion on the actual index?

```
assertEquals(10, execute(String.format("SELECT * FROM %s.%s WHERE d >= 0", KEYSPACE, "sai_test")).size());
```

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
```diff
@@ -31,8 +31,11 @@
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
+import org.apache.cassandra.index.sai.disk.format.IndexDescriptor;
+import org.apache.cassandra.index.sai.utils.IndexIdentifier;
 import org.apache.cassandra.io.sstable.format.big.BigFormat;
 import org.apache.cassandra.io.util.File;
```
maedhroz: nit: I guess we should just move all 4 of these to where they belong below...

src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
```diff
@@ -579,6 +599,18 @@ public Builder sorted()
             return this;
         }
 
+        /**
+         * Whether indexes should be built and serialized to disk along data. Defaults to true.
+         *
+         * @param buildIndexes true if indexes should be built, false otherwise
+         * @return this builder
+         */
+        public Builder withBuiltIndexes(boolean buildIndexes)
```
maedhroz:
```suggestion
        public Builder withBuildIndexes(boolean buildIndexes)
```

test/unit/org/apache/cassandra/db/ImportTest.java
```diff
@@ -710,6 +710,154 @@ public void importExoticTableNamesTest() throws Throwable
...
+
+            File backupDir = moveToBackupDir(sstables);
+
+            assertEquals(0, execute(String.format("SELECT * FROM %s.%s", KEYSPACE, "sai_test")).size());
+
+            SSTableImporter importer = new SSTableImporter(cfs);
+            SSTableImporter.Options options = SSTableImporter.Options.options(backupDir.toString())
+                                                                     .copyData(true)
+                                                                     .failOnMissingIndex(true)
+                                                                     .validateIndexChecksum(false)
```
maedhroz, line 883: This reminded me of the edge case of empty SSTable indexes. We made sure those don't fail streaming/validation in CASSANDRA-19017, but it might be good to at least have a basic sanity check for the importer. Creating an empty index should be pretty easy. (see `StorageAttachedIndexDDLTest#verifyFlushAndCompactEmptyIndex()`)

test/unit/org/apache/cassandra/io/sstable/CQLSSTableWriterTest.java
```diff
@@ -1309,6 +1311,86 @@ public void testWriteWithSortedAndMaxSize() throws Exception
         }
     }
 
+    @Test
+    public void testWriteWithSAI() throws Exception
+    {
+        String schema = "CREATE TABLE " + qualifiedTable + " ("
+                        + "  k int PRIMARY KEY,"
+                        + "  v1 text,"
+                        + "  v2 text )";
```
maedhroz: nit: I'd suggest making v2 an int, just to make sure we're covering all the numeric index component validations.

src/java/org/apache/cassandra/db/SSTableImporter.java
```diff
@@ -97,7 +99,19 @@ synchronized List<String> importNewSSTables(Options options)
...
+                            {
+                                IndexDescriptor indexDescriptor = IndexDescriptor.create(descriptor,
+                                                                                         cfs.getPartitioner(),
+                                                                                         cfs.metadata().comparator);
+
+                                if (!indexDescriptor.isPerSSTableIndexBuildComplete())
+                                    throw new IllegalStateException("Missing SAI indexes to import!");
```
maedhroz: I think we need to check the per-column components here as well.

(See `isPerColumnIndexBuildComplete()`. We should be able to get the indexes and their identifiers from the group above.)

src/java/org/apache/cassandra/db/SSTableImporter.java
```diff
@@ -481,6 +502,8 @@ public String toString()
                    ", invalidateCaches=" + invalidateCaches +
                    ", extendedVerify=" + extendedVerify +
                    ", copyData= " + copyData +
+                   ", failOnMissingIndex= " + failOnMissingIndex +
+                   ", validateIndexChecksum= " + validateIndexChecksum +
```
maedhroz, line 528: I'll come back and do a final review of all this stuff once we settle on naming ;)

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
```diff
@@ -409,9 +409,9 @@ public void checksumPerIndexComponents(IndexTermType indexTermType, IndexIdentif
     }
 
     @SuppressWarnings("BooleanMethodIsAlwaysInverted")
-    public void checksumPerSSTableComponents()
+    public void checksumPerSSTableComponents(boolean validateChecksum)
```
maedhroz: Looking at this, we could probably merge w/ `validatePerSSTableComponents()` above, and we'd get some extra logging for free. We'd have to pass an extra boolean directing the method to throw or not on `UncheckedIOException` though.

Also, looking at `validatePerSSTableComponents()` itself, the log message isn't specific enough. It should probably be:

```
logger.info(logMessage("Validating per-sstable index components for SSTable {} using mode {}..."), sstableDescriptor, validation);
```

Otherwise, the log message will be the same for every SSTable we validate :(

maedhroz: CC @pkolaczk 

(Making sure I'm not missing something.)

maedhroz: Either way, probably can't continue to start the method name w/ `checksum*`...

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
```diff
@@ -353,7 +353,7 @@ public boolean validateSSTableAttachedIndexes(Collection<SSTableReader> sstables
 
             if (indexDescriptor.isPerSSTableIndexBuildComplete())
             {
-                indexDescriptor.checksumPerSSTableComponents();
+                indexDescriptor.checksumPerSSTableComponents(validateChecksum);
 
                 for (StorageAttachedIndex index : indexes)
                 {
```
maedhroz, line 359: We should pass `validateChecksum` to `checksumPerIndexComponents()` (which should probably be renamed or merged w/ the other validation method, see https://github.com/apache/cassandra/pull/3029/files#r1446766349) as well

src/java/org/apache/cassandra/io/sstable/AbstractSSTableSimpleWriter.java
```diff
@@ -49,12 +51,14 @@ abstract class AbstractSSTableSimpleWriter implements Closeable
     protected SSTableFormat<?, ?> format = DatabaseDescriptor.getSelectedSSTableFormat();
     protected static final AtomicReference<SSTableId> id = new AtomicReference<>(SSTableIdFactory.instance.defaultBuilder().generator(Stream.empty()).get());
     protected boolean makeRangeAware = false;
+    protected Collection<Index.Group> indexGroups;
```
maedhroz:
```suggestion
    protected final Collection<Index.Group> indexGroups;
```

maedhroz (changes requested):
Overall structure looks pretty good. A few things to resolve around per-column component validation, some naming to finalize, and maybe a new test case for empty index validation, and we should be set...

src/java/org/apache/cassandra/tools/NodeProbe.java
smiklosovic: @JeetKunDoug correct, I created new one on ColumnFamilyStore/MBean but I changed the one on NodeProbe. That is just fine, because NodeProbe one is "internal" to Cassandra. Nobody depends on _NodeProbe_ across versions ... Does that make sense?

src/java/org/apache/cassandra/tools/nodetool/Import.java
smiklosovic: for `--skip-index-validation`, I do not like this one because if you notice what other flags are, there is prefix `--no` - `--no-verify`, `--no-tokens`, `--no-invalidate-caches` - there is no `--skip...`. We might do `--no-index-validation`, that would work, but not `--skip-index-validation`.

When it comes to `--require-index-component`, I do not like the fact that this does not say what happens when they are required and they are not present. To prepend it with `--fail` is pretty obvious to me. Anyway, one can conclude that if something is required and it is not there, it will probably complain and fail eventually ... yeah, I think this could fly (and the fact it will fail is also mentioned in the description field).

flags are now `-ri` and `-niv` as in `--require-index-components` and `--no-index-validation`.

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
smiklosovic: @maedhroz `logMessage` in `validatePerSSTableComponents()` does print keyspace and table at least if you look into its implementation: 

    public String logMessage(String message)
    {
        // Index names are unique only within a keyspace.
        return String.format("[%s.%s.*] %s",
                             sstableDescriptor.ksname,
                             sstableDescriptor.cfname,
                             message);
    }

This however does not print Descrpitor's

    @Override
    public String toString()
    {
        return baseFile().absolutePath();
    }

We could include it into `logMessage` argument here.

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
smiklosovic: I passed it there too.

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
smiklosovic: I have a hard time to understand what should be done here ...

smiklosovic: I simplified that.

src/java/org/apache/cassandra/db/SSTableImporter.java
```diff
@@ -97,7 +101,39 @@ synchronized List<String> importNewSSTables(Options options)
...
+                                                                                      table));
+
+                                    for (Index index : saiIndexGroup.getIndexes())
+                                    {
+                                        IndexIdentifier indexIdentifier = new IndexIdentifier(keyspace, table, index.getIndexMetadata().name);
+                                        if (!indexDescriptor.isPerColumnIndexBuildComplete(indexIdentifier))
+                                            throw new IllegalStateException(String.format("Missing SAI index to import for index %s on %s.%s",
+                                                                                          index.getIndexMetadata().name,
+                                                                                          keyspace,
+                                                                                          table));
```
maedhroz, line 130: nit: You could use the `logMessage()` helpers from `IndexDescriptor` and `IndexIdentifier` for the exceptions above, but not critical. It would give you the standardized prefix (which we did to make it easier to search for all SAI-related logging) but that's it.

src/java/org/apache/cassandra/db/SSTableImporter.java
```diff
@@ -97,7 +101,39 @@ synchronized List<String> importNewSSTables(Options options)
...
+
+                                    if (!indexDescriptor.isPerSSTableIndexBuildComplete())
+                                        throw new IllegalStateException(String.format("Missing SAI index to import for SSTable %s on %s.%s",
+                                                                                      indexDescriptor.sstableDescriptor.toString(),
+                                                                                      keyspace,
+                                                                                      table));
+
+                                    for (Index index : saiIndexGroup.getIndexes())
+                                    {
+                                        IndexIdentifier indexIdentifier = new IndexIdentifier(keyspace, table, index.getIndexMetadata().name);
```
maedhroz, line 125: nit: `StorageAttachedIndex` has an `IndexIdentifier` already if you want to cast the `Index`, which should be safe.

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
```diff
@@ -384,34 +384,30 @@ public boolean validatePerIndexComponents(IndexTermType indexTermType, IndexIden
...
     }
 
-    public void checksumPerIndexComponents(IndexTermType indexTermType, IndexIdentifier indexIdentifier)
-    {
-        version.onDiskFormat().validatePerColumnIndexComponents(this, indexTermType, indexIdentifier, true);
-    }
-
-    @SuppressWarnings("BooleanMethodIsAlwaysInverted")
-    public void checksumPerSSTableComponents()
+    public void checksumPerIndexComponents(IndexTermType indexTermType, IndexIdentifier indexIdentifier, boolean validateChecksum)
```
maedhroz: We could just remove this method and modify `validatePerIndexComponents()` to take a `rethrow` argument (similar to `validatePerSSTableComponents()`)?

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
```diff
@@ -353,12 +353,12 @@ public boolean validateSSTableAttachedIndexes(Collection<SSTableReader> sstables
...
-                indexDescriptor.checksumPerSSTableComponents();
+                indexDescriptor.validatePerSSTableComponents(IndexValidation.CHECKSUM, true);
 
                 for (StorageAttachedIndex index : indexes)
                 {
                     if (indexDescriptor.isPerColumnIndexBuildComplete(index.identifier()))
-                        indexDescriptor.checksumPerIndexComponents(index.termType(), index.identifier());
+                        indexDescriptor.checksumPerIndexComponents(index.termType(), index.identifier(), validateChecksum);
                     else if (throwOnIncomplete)
                         throw new IllegalStateException(indexDescriptor.logMessage("Incomplete per-column index build"));
```
maedhroz: nit: I know this falls into the category of non-essential clean up, but I'm realizing that the two exception messages created in this method, much like the ones in `IndexDescriptor` before you fixed them, are missing SSTable info. They could probably be fixed in much the same way.

maedhroz (approved):
Left a couple minor notes, but it looks like the important stuff is all taken care of.

LGTM

Thanks for the patch!

src/java/org/apache/cassandra/db/SSTableImporter.java
smiklosovic: I dont like to mix sai specific logging stuff with sstable importer. Seems totally unrelated really ...

src/java/org/apache/cassandra/index/sai/StorageAttachedIndexGroup.java
smiklosovic: yeah I can fix this.

src/java/org/apache/cassandra/db/SSTableImporter.java
smiklosovic: I rather leave this as it is ... casting this etc ... nah.

src/java/org/apache/cassandra/index/sai/disk/format/IndexDescriptor.java
smiklosovic: sure

src/java/org/apache/cassandra/db/SSTableImporter.java
maedhroz: Sure. `IndexIdentifier` itself could actually be an index implementation-agnostic class. There's nothing ASI-specific forcing it to be in `org.apache.cassandra.index.sai.utils`.

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -200,22 +200,36 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean, Memtable.Owner
...
+    // post-flush executor is single threaded to provide guarantee that any flush Future on a CF will never return until prior flushes have completed
+    private static ExecutorPlus postFlushExecutor;
+
+    private static ExecutorPlus reclaimExecutor;
+
+    private static PerDiskFlushExecutors perDiskflushExecutors;
+
+    static
+    {
+        if (!DatabaseDescriptor.isClientOrToolInitialized())
```
smiklosovic: this needs to be completely skipped in the client mode because DatabaseDescriptor is not initialized yet so it would fail to instantiate these executors because invalid values would be passed to that. Interestingly, the fact these fields are left to be null does not hurt anything because they are never accessed.

src/java/org/apache/cassandra/db/memtable/AbstractAllocatorMemtable.java
```diff
@@ -78,9 +79,17 @@ public abstract class AbstractAllocatorMemtable extends AbstractMemtableWithComm
     static MemtablePool createMemtableAllocatorPool()
     {
         Config.MemtableAllocationType allocationType = DatabaseDescriptor.getMemtableAllocationType();
-        long heapLimit = DatabaseDescriptor.getMemtableHeapSpaceInMiB() << 20;
-        long offHeapLimit = DatabaseDescriptor.getMemtableOffheapSpaceInMiB() << 20;
-        float memtableCleanupThreshold = DatabaseDescriptor.getMemtableCleanupThreshold();
+
+        long heapLimit = new DataStorageSpec.IntMebibytesBound((int) (Runtime.getRuntime().maxMemory() / (4 * 1048576))).toMebibytes() << 20;
```
smiklosovic: I took this from the logic when `DatabaseDescriptor` is normally initialized in daemon mode.

src/java/org/apache/cassandra/io/sstable/CQLSSTableWriter.java
```diff
@@ -645,14 +652,31 @@ public CQLSSTableWriter build()
...
-                    if (!indexStatements.isEmpty())
-                        tableMetadata = applyIndexes(ksm.withSwapped(ksm.tables.with(tableMetadata)));
+                    if (buildIndexes && !indexStatements.isEmpty())
+                    {
+                        cfs = new ColumnFamilyStore(Keyspace.mockKS(ksm),
+                                                    tableMetadata.name,
+                                                    () -> new SequenceBasedSSTableId(0),
+                                                    TableMetadataRef.forOfflineTools(tableMetadata),
+                                                    new Directories(tableMetadata, new Directories.DataDirectory[] {
+                                                    new Directories.DataDirectory("/tmp/dirs")
```
smiklosovic: it will create these directories with keyspace and table dirs on disk.

src/java/org/apache/cassandra/locator/AbstractReplicationStrategy.java
```diff
@@ -324,7 +324,8 @@ private static AbstractReplicationStrategy createInternal(String keyspaceName,
         try
         {
             Constructor<? extends AbstractReplicationStrategy> constructor = strategyClass.getConstructor(parameterTypes);
-            strategy = constructor.newInstance(keyspaceName, tokenMetadata, snitch, strategyOptions);
+            IEndpointSnitch endpointSnitch = snitch == null ? new SimpleSnitch() : snitch;
```
smiklosovic: This will propagate `null` snitch in client mode. I just have to put there something non-nullish.

smiklosovic: to be extra safe, we could probably wrap this in DatabaseDescriptor.isClientOrToolInitialized() to set it to SimpleSnitch only in that case, otherwise we would just let this fail ...

maedhroz:
Pushed up an alternate approach to fixing the client mode issues here: https://github.com/apache/cassandra/commit/fecac728ec945e9b376a5fb2c4af54f337d744cf

Tried to minimize the changes to existing code and the need for making up mock values by bypassing certain paths earlier. Also took some inspiration from what `StressCQLSSTableWriter` does in this case.

@smiklosovic Let me know what you think...

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
maedhroz: I kept these final in https://github.com/apache/cassandra/commit/fecac728ec945e9b376a5fb2c4af54f337d744cf. Even if you don't take that patch wholesale, we should keep that part.

src/java/org/apache/cassandra/locator/AbstractReplicationStrategy.java
maedhroz: Agreed. I did this in https://github.com/apache/cassandra/commit/fecac728ec945e9b376a5fb2c4af54f337d744cf.

maedhroz:
This PR is replaced by https://github.com/apache/cassandra/pull/3046, yes?

smiklosovic:
@maedhroz yes, in 3046 there are refactored tests, your commit and one extra commit fixing the issue we talked about

PR #3031, opened 2024-01-07
CASSANDRA-19104/2 Default to human-readable in `nodetool tablestats`

zaaath:
Default to human-readable in `nodetool tablestats`. See a discussion on the topic in the dev mailing list [dev@cassandra.apache.org](https://lists.apache.org/thread/mlp715kxho5b6f1ql9omlzmmnh4qfby9)

### Before
`nodetool tablestats` doesn't use the human-readable formatting by default (with KiB/MiB/GiB units), instead it outputs data size values as is. In order to switch to the human-readable output, `--human-readable` or `--H` CLI flag is required.

### After
`nodetool tablestats` uses the human-readable format by default. For backward compatibility `--no-human-readable` flag can be used to disable it.

### How to reproduce
```bash
nodetool tablestats
```
and
```bash
nodetool tablestats --no-human-readable
```

Tests
```bash
```

patch by @zaaath; reviewed by ??? for [CASSANDRA-19104](https://issues.apache.org/jira/browse/CASSANDRA-19104)

### DIFF
--- a/src/java/org/apache/cassandra/tools/nodetool/TableStats.java
+++ b/src/java/org/apache/cassandra/tools/nodetool/TableStats.java
@@ -36,10 +36,10 @@ public class TableStats extends NodeToolCmd
     @Option(name = "-i", description = "Ignore the list of tables and display the remaining tables")
     private boolean ignore = false;
 
-    @Option(title = "human_readable",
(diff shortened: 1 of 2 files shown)


zaaath:
Moved to #3069

PR #3032, opened 2024-01-08
C19016 accord table config

bdeggleston:


src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1187,6 +1187,47 @@ public static void mutateMV(ByteBuffer dataKey, Collection<Mutation> mutations,
...
+                ConsistencyLevel commitCL = mode.commitCLForStrategy(consistencyLevel);
+                if (result == null || commitCL.compareTo(result) > 0)
+                    result = commitCL;
+            }
+        }
+        Invariants.checkState(result != null);
+        return result;
+    }
+
+    private static boolean writesThroughAccord(List<? extends IMutation> mutations)
```
aweisberg: This isn't range based so the entire table is opted into Accord so it's not possible to limit the blast radius to a single key.

src/java/org/apache/cassandra/cql3/statements/schema/TableAttributes.java
```diff
@@ -165,6 +167,14 @@ private TableParams build(TableParams.Builder builder)
             }
         }
 
+        // TODO (now): fail if attempting to change a system table
+        // TODO (now): set migration to complete if this is a new table
```
aweisberg: A pretty valuable/important to do, do you want to do it in this PR?

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
```diff
@@ -246,7 +246,7 @@ public void finished()
         checkNotNull(minVersion, "Unable to determine minimum cluster version");
         IAccordService accordService = AccordService.instance();
         if (session.streamOperation().requiresBarrierTransaction()
-            && accordService.isAccordManagedTable(cfs.getTableId())
+            && cfs.metadata().isAccordEnabled()
```
aweisberg, line 249: I just realized this probably breaks if we try to run the barrier on a range that is only partially Accord managed. I'll bet Accord generates an error. This should check if the ranges actually intersect with a migrating/migrated range and only run the barrier to limit the blast radius of using Accord.

It will also race with user modifications to the migrating ranges so uh yuck, but should also probably retry a few times. We already have a retrying barrier method in `AccordService` that should probably be modified to handle this error and take a strategy of either bubble up the error or retry with just the intersecting ranges managed by Accord.

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
```diff
@@ -553,6 +554,33 @@ public void validate(ClientState state)
...
+
+            // user is manually updating migration mode, don't interfere
+            if (forceMigrationChange)
+                return next;
+
+            if (!modeChange)
+                return next;
+
+            if (wasMigrating)
+                throw ire(format("Cannot change transactional mode from %s to %s for %s.%s before transactional migration has completed",
```
aweisberg, line 586: They should be able to change between the different interop options if migrating to Accord?

I think the only restriction we want to add here (and via JMX) is that they can be migrating, but there can be no migrating ranges because it's not safe to flip the direction of a range that starts but doesn't compelte migration.

src/java/org/apache/cassandra/cql3/statements/schema/CreateTableStatement.java
```diff
@@ -125,6 +126,10 @@ public Keyspaces apply(ClusterMetadata metadata)
         if (!table.params.compression.isEnabled())
             Guardrails.uncompressedTablesEnabled.ensureEnabled(state);
 
+        if (table.params.transactionalMode.accordIsEnabled && !DatabaseDescriptor.getAccordTransactionsEnabled())
```
aweisberg, line 132: Do we have a way for to default tables into Accord being enabled on creation?

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -706,4 +669,27 @@ public Pair<Int2ObjectHashMap<RedundantBefore>, DurableBefore> getRedundantBefor
         }));
         return Pair.create(redundantBefores, durableBefore.get());
     }
+
+    public static void ensureTableIsEnabledForAccord(TableMetadata metadata)
```
aweisberg: These all assert the table is enabled for Accord they don't ensure it which at least to me implies that they actually enable it.

src/java/org/apache/cassandra/service/consensus/TransactionalMode.java
```diff
@@ -0,0 +1,136 @@
...
+ * This is different from Paxos because Paxos performs consensus on the actual values to be applied so recovery
+ * coordinators will always produce a consistent state when applying a transaction. Accord performs consensus on
+ * the execution order of transaction and different coordinators witnessing different states not managed by Accord
+ * can produce multiple outcomes for a transaction.
+ *
+ * // TODO to safely migrate you would have to route all writes through Accord with the current implementation
+ * // We could do it by range instead in the migration version, but then we need to know when all in flight writes
+ * // are done before marking a range as migrated. Would waiting out the timeout be enough (timeout bugs!)?
+ */
+public enum TransactionalMode
```
aweisberg, line 57: I think this name is problematic.

The current set of settings are kind of idiosynchratic to Accord where mixed is actually still unsafe. Maybe it just shouldn't exist, but I wanted to have it in there so you aren't forced to run all writes through Accord in a high write throughput use cases.

Another concern is that we also support LWT which are transactions even if they are not very good ones.

Interoperability needs to be broken out from enabled/disabled as mentioned in the `AlterTable` comment. It also seems like we want a default so you only have to specify it in unusual situations where you want to trade safety for performance or vice versa from the default.

Defaults are going to be important when you don't own the schema and you want to roll it out widely to existing applications that will be creating and destroying tables outside of our control.

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1187,6 +1187,47 @@ public static void mutateMV(ByteBuffer dataKey, Collection<Mutation> mutations,
         }
     }
 
+    private static ConsistencyLevel consistencyLevelForCommit(Collection<? extends IMutation> mutations, ConsistencyLevel consistencyLevel)
+    {
+        ConsistencyLevel result = null;
+        for (IMutation mutation : mutations)
```
aweisberg, line 1195: `commitCLForStrategy` returns `null` instead of the supplied consistency level if a table is configured to fully use Accord. This would means an earlier iterated table that requires it's consistency level to be honored wouldn't get that.

Additionally it's possible to break an application by enabling migration for one table and not another and we don't support atomically updating multiple tables at once so users can't even do it correctly anyways.

Really what needs to happen is break apart the mutations into ones that need to run through Accord and ones that need to run outside of Accord. This isn't a new deficiency it's just something needs to be added to interop.

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1187,6 +1187,47 @@ public static void mutateMV(ByteBuffer dataKey, Collection<Mutation> mutations,
...
+                    accordWrite = true;
+                if (!normalWrite && !mode.writesThroughAccord)
+                    normalWrite = true;
+            }
+        }
+
+        if (!accordWrite && !normalWrite)
+            throw new IllegalStateException("Mutation is neither a normal write nor an accord write");
+        if (accordWrite && normalWrite)
+            throw new InvalidRequestException("Cannot mix writes to accord and normal tables");
```
aweisberg, line 1236: As mentioned above I don't think we can safely avoid supporting that.

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1220,12 +1261,12 @@ public static void mutateWithTriggers(List<? extends IMutation> mutations,
...
                               .updatesAffectView(mutations, true);
 
+
         long size = IMutation.dataSize(mutations);
         writeMetrics.mutationSize.update(size);
         writeMetricsForLevel(consistencyLevel).mutationSize.update(size);
-        NonSerialWriteStrategy nonSerialWriteStrategy = DatabaseDescriptor.getNonSerialWriteStrategy();
-        if (nonSerialWriteStrategy.writesThroughAccord && !SchemaConstants.getSystemKeyspaces().contains(keyspaceName))
-            mutateWithAccord(augmented != null ? augmented : mutations, consistencyLevel, queryStartNanoTime, nonSerialWriteStrategy);
+        if (writesThroughAccord(mutations) && !SchemaConstants.getSystemKeyspaces().contains(keyspaceName))
```
aweisberg: Maybe the `SchemaConstants` check should now be an assertion. This check could also be in create and alter table.

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1252,11 +1293,11 @@ private static void mutateWithAccord(Collection<? extends IMutation> iMutations,
             }
         }
         // Potentially ignore commit consistency level if the strategy specifies accord and not migration
-        ConsistencyLevel clForCommit = nonSerialWriteStrategy.commitCLForStrategy(consistencyLevel);
-        AccordUpdate update = new TxnUpdate(fragments, TxnCondition.none(), clForCommit);
+        ConsistencyLevel clForCommit = consistencyLevelForCommit(mutations, consistencyLevel);
+        TxnUpdate update = new TxnUpdate(fragments, TxnCondition.none(), clForCommit);
+        logger.info("Mutating with Accord clForCommit " + clForCommit);
```
aweisberg: Remove before commit

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1651,25 +1649,17 @@ public List<Integer> finishConsensusMigration(@Nonnull String keyspace,
...
-    }
-
     @Override
     public String listConsensusMigrations(@Nullable Set<String> keyspaceNames,
                                           @Nullable Set<String> tableNames,
                                           @Nonnull String format)
     {
         ClusterMetadata cm = ClusterMetadata.current();
-        ConsensusMigrationState snapshot = cm.consensusMigrationState;
+        MigrationState snapshot = cm.consensusMigrationState;
```
aweisberg: `MigrationState` is not a great name just because in the lifecycle of Cassandra we might have more migration type things.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1651,25 +1649,17 @@ public List<Integer> finishConsensusMigration(@Nonnull String keyspace,
...
+        // TODO wanted something human and machine readable, but didn't expend a lot of thought
+        // on what human readable conventions should be, also couldn't get snakeyaml
+        // to output YAML containing only primitives without going through the goofy mapping process
+        // it adds tags and references that clutter the output up badly
```
aweisberg: Does this comment need to come back?

src/java/org/apache/cassandra/service/accord/txn/TxnQuery.java
```diff
@@ -154,10 +152,6 @@ public Result compute(TxnId txnId, Timestamp executeAt, Seekables<?, ?> keys, @N
         Epoch epoch = Epoch.create(executeAt.epoch());
         if (transactionIsInMigratingOrMigratedRange(epoch, keys))
         {
-            // Fail fast because we can't be sure where this request should really run or what was intended
```
aweisberg: This check could stay since we still know the tables from the keys and check their metadata. Feels borderline since it means you were able to alter the table unsafely.

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationTarget.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.google.common.primitives.SignedBytes;
+
+import org.apache.cassandra.service.consensus.TransactionalMode;
+
+public enum ConsensusMigrationTarget
+{
+    paxos(0),
+    accord(1),
+    reset(2);  // FIXME: remove
```
aweisberg: This still seems like a pretty valid escape hatch when you want to get out of migration and just pick a specific system.

src/java/org/apache/cassandra/service/consensus/migration/ConsensusRequestRouter.java
```diff
@@ -76,43 +74,56 @@ public static void resetInstance()
...
+        return routeAndMaybeMigrate(cm, metadata, key, consistencyLevel, queryStartNanotime, timeoutNanos, isForWrite);
     }
 
-    protected ConsensusRoutingDecision routeAndMaybeMigrate(@Nonnull  DecoratedKey key, @Nonnull ColumnFamilyStore cfs, ConsistencyLevel consistencyLevel, long queryStartNanoTime, long timeoutNanos, boolean isForWrite)
+    protected ConsensusRoutingDecision routeAndMaybeMigrate(ClusterMetadata cm, @Nonnull TableMetadata tmd, @Nonnull DecoratedKey key, ConsistencyLevel consistencyLevel, long queryStartNanoTime, long timeoutNanos, boolean isForWrite)
     {
-        ClusterMetadata cm = ClusterMetadata.current();
 
-        TableMigrationState tms = cm.consensusMigrationState.tableStates.get(cfs.getTableId());
+        if (tmd.params.transactionalMigrationFrom == TransactionalMigrationFromMode.none)
```
aweisberg: Took me a while to figure out that `transactionalMigrationFrom`  `none` means no migration in progress anymore. Maybe call it `none_in_progress`?

I know it's obvious in retrospect.

src/java/org/apache/cassandra/service/consensus/migration/ConsensusTableMigrationState.java
```diff
@@ -1,907 +0,0 @@
-/*
```
aweisberg: Would be better to use `git mv` to preserve the history.

Were there changes to the other classes pulled of `ConsensusTableMigrationState` that I should check for, or are they just being extracted?

src/java/org/apache/cassandra/service/reads/repair/BlockingReadRepair.java
```diff
@@ -193,9 +192,8 @@ public void awaitWrites()
     @Override
     public void repairPartition(DecoratedKey dk, Map<Replica, Mutation> mutations, ReplicaPlan.ForWrite writePlan)
     {
-        NonSerialWriteStrategy nonSerialWriteStrategy = DatabaseDescriptor.getNonSerialWriteStrategy();
-        if (coordinator.isEventuallyConsistent() && (DatabaseDescriptor.getLWTStrategy() == LWTStrategy.accord
-                                                     || nonSerialWriteStrategy.blockingReadRepairThroughAccord))
+        TransactionalMode transactionalMode = command.metadata().params.transactionalMode;
+        if (coordinator.isEventuallyConsistent() && transactionalMode.blockingReadRepairThroughAccord)
```
aweisberg, line 196: This also should do read repair separately for Accord/non-Accord ranges 😭

src/java/org/apache/cassandra/tcm/transformations/AlterSchema.java
```diff
@@ -249,22 +253,56 @@ private static Map<ReplicationParams, Set<KeyspaceMetadata>> groupByReplication(
...
+                case auto:
+                    Token minToken = DatabaseDescriptor.getPartitioner().getMinimumToken();
+                    ranges = Range.normalize(Collections.singletonList(new Range<>(minToken, minToken)));
+                    break;
+                case explicit:
+                    ranges = Collections.emptyList();
+                    break;
+            }
+
+            migrationState = migrationState.withRangesMigrating(startedIds, ranges, true);
```
aweisberg: When changing directions in a migration we don't want to do `auto` if all ranges weren't already migrated since that repeats work already done. I think we want to invert the migration using `tableMigrationState.withMigrationTarget`.

src/java/org/apache/cassandra/tcm/transformations/AlterSchema.java
```diff
@@ -249,22 +253,56 @@ private static Map<ReplicationParams, Set<KeyspaceMetadata>> groupByReplication(
...
+
+        Set<TableId> droppedIds = Streams.concat(altered.stream().flatMap(diff -> diff.tables.dropped.stream().map(TableMetadata::id)),
+                                                 dropped.stream().flatMap(ks -> ks.tables.stream().map(TableMetadata::id)))
+                                         .collect(toImmutableSet());
+
+        if (!droppedIds.isEmpty())
+            migrationState = migrationState.withMigrationsCompletedFor(droppedIds);
+
+        Set<TableId> completedIds = altered.stream().flatMap(diff -> diff.tables.altered.stream())
+                                           .filter(alt -> alt.before.params.transactionalMigrationFrom != TransactionalMigrationFromMode.none
```
aweisberg: Is this the case where they manually altered `transactionalMigrationFrom` and forced the completion of the migration by just saying "well it's not migrating?"

We should log the ranges that were migrating so when they do this unsafely there is some indication of what might have been impacted and that this unsafe action was performed.

Borderline it shouldn't be allowed unless an explicit option is set on the C* process. We will have real headaches if people change this and then have problems and we lose the logs and don't know what happened.

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
```diff
@@ -553,6 +554,33 @@ public void validate(ClientState state)
...
+                                 prev.transactionalMode, next.transactionalMode,
+                                 keyspaceName, tableName));
+
+            // set table to migrating
+            TransactionalMigrationFromMode migrateFrom = TransactionalMigrationFromMode.fromMode(prev.transactionalMode, next.transactionalMode);
```
aweisberg, line 591: It's very confusing here that users can supply this, but then we replace it.

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
```diff
@@ -553,6 +554,33 @@ public void validate(ClientState state)
             validateDefaultTimeToLive(attrs.asNewTableParams());
         }
 
+        private TableParams validateAndUpdateTransactionalMigration(TableParams prev, TableParams next)
+        {
+            boolean modeChange = prev.transactionalMode != next.transactionalMode;
+            boolean wasMigrating = prev.transactionalMigrationFrom != TransactionalMigrationFromMode.none;
+            boolean forceMigrationChange = prev.transactionalMigrationFrom != next.transactionalMigrationFrom;
```
aweisberg, line 567: Is this meant to be the unsafe way to make a migration complete?

src/java/org/apache/cassandra/service/consensus/migration/TableMigrationState.java
```diff
@@ -0,0 +1,345 @@
...
+    public final String keyspaceName;
+
+    @Nonnull
+    public final String tableName;
+
+    @Nonnull
+    public final TableId tableId;
+
+    @Nonnull
+    public final ConsensusMigrationTarget targetProtocol;
```
aweisberg, line 78: We definitely want to remove this, and get it from the `TableMetadata`. If we store it twice we are not going to keep it in sync.

src/java/org/apache/cassandra/tcm/transformations/AlterSchema.java
```diff
@@ -249,22 +253,56 @@ private static Map<ReplicationParams, Set<KeyspaceMetadata>> groupByReplication(
...
-        }
-        else
+        MigrationState migrationState = prev;
+
+        Set<TableId> droppedIds = Streams.concat(altered.stream().flatMap(diff -> diff.tables.dropped.stream().map(TableMetadata::id)),
+                                                 dropped.stream().flatMap(ks -> ks.tables.stream().map(TableMetadata::id)))
+                                         .collect(toImmutableSet());
+
+        if (!droppedIds.isEmpty())
+            migrationState = migrationState.withMigrationsCompletedFor(droppedIds);
```
aweisberg: This keeps state around for dropped tables. We should remove them entirely the way it used to?

src/java/org/apache/cassandra/service/consensus/migration/MigrationState.java
```diff
@@ -0,0 +1,218 @@
...
+        ConsensusMigrationTarget target = ConsensusMigrationTarget.fromTransactionalMode(metadata.params.transactionalMode);
+        if (!overwrite && current.containsKey(metadata.id))
+        {
+            tableState = current.get(metadata.id).withRangesMigrating(ranges, target);
+        }
+        else
+        {
+            tableState = new TableMigrationState(metadata.keyspace, metadata.name, metadata.id, target, ImmutableSet.of(), ImmutableMap.of(Epoch.EMPTY, ranges));
+        }
+        next.put(metadata.id, tableState);
```
aweisberg: Does this actually work? I thought `ImmutableMap.Builder` throws an exception on duplicate keys when you call `build()`.

There is apparently a `buildKeepingLast` method which should work?

src/java/org/apache/cassandra/service/consensus/migration/MigrationState.java
```diff
@@ -0,0 +1,218 @@
...
+            if (metadata == null)
+                throw new IllegalStateException("Unknown table id: " + id);
+
+            withRangesMigrating(tableStates, updated, metadata, ranges, overwrite);
+        });
+
+        return new MigrationState(lastModified, updated.build());
+    }
+
+    public MigrationState withRangesMigrating(TableMetadata metadata, List<Range<Token>> ranges, boolean overwrite)
```
aweisberg: Unused?

src/java/org/apache/cassandra/tcm/transformations/MaybeFinishConsensusMigrationForTableAndRange.java
```diff
@@ -122,9 +145,15 @@ public Result execute(@Nonnull ClusterMetadata metadata)
         if (!intersects(tms.migratingRanges, normalizedRepairedRanges))
             return new Rejected(INVALID, format("Table %s is migrating ranges %s, which doesn't include repaired ranges %s", ksAndCF, tms.migratingRanges, normalizedRepairedRanges));
 
-        TableMigrationState newTableMigrationState = tms.withRangesRepairedAtEpoch(normalizedRepairedRanges, minEpoch);
+        Transformer next = metadata.transformer();
+        MigrationState migrationState = metadata.consensusMigrationState.withRangesReparedAtEpoch(tbm, normalizedRepairedRanges, minEpoch);
+        next = next.with(migrationState);
+
+        // reset the migration value on the table if the migration has completed
+        if (migrationState.tableStates.containsKey(tbm.id))
```
aweisberg: Why is just having a `migrationState` enough to indicate that migration has completed entirely?

aweisberg (changes requested):
Finished reviewing up to the tests. Found several issues, some were pre-existing problems with interop not being range based, others are unrelated.

It looks like it's mostly there UI wise in terms of getting it into the table schema. and having all the necessary functionality.

I personally had a really hard time figuring out the relationship between the migration from stuff and the transactional mode, but maybe that is just me.

src/java/org/apache/cassandra/tcm/transformations/AlterSchema.java
aweisberg: Also this is a correctness issue because if you don't precisely flip the ranges then you are basically migrating ranges to the new "from" that never actually finished migration.

In fact I think that the current approach for flipping directions is unsafe if we allow it to occur when any range hasn't been fully migrated because the range now consists of a mix of migrated keys and ranges waiting to be migrated and the ranges waiting to be migrated can't be forcibly swapped to the new system because then we will reverse the kind of key migration we are doing without doing a repair.

test/distributed/org/apache/cassandra/distributed/test/accord/AccordCQLTest.java
```diff
@@ -2422,13 +2414,13 @@ private void testSelectUDTReference(String tableDDL) throws Exception
     @Test
     public void testSelectMultiCellUDTFieldReference() throws Exception
     {
-        testSelectUDTFieldReference("CREATE TABLE " + qualifiedTableName + " (k int PRIMARY KEY, customer person)");
+        testSelectUDTFieldReference("CREATE TABLE " + qualifiedTableName + " (k int PRIMARY KEY, customer person) WITH transactional_mode='" + transactionalMode + "'");
```
aweisberg, line 2417: This test is a strong argument in favor of allowing a default to be set. Maybe a `keyspace` level thing? Or `Config`.

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1112,8 +1111,13 @@ public enum PaxosOnLinearizabilityViolation
...
 
-    public LWTStrategy lwt_strategy = LWTStrategy.migration;
-    public NonSerialWriteStrategy non_serial_write_strategy = NonSerialWriteStrategy.normal;
+    /**
+     * Defines the behavior of range migration opt-in when changing transactional settings on a table. In auto,
+     * all ranges are marked as migrating and no additional user action is needed aside from running repairs. In
+     * explicit, no ranges are marked as migrating, and the user needs to explicitly mark ranges as migrating to
+     * the target transactional mode via nodetool.
+     */
+    public volatile TransactionalRangeMigration transactional_range_migration = TransactionalRangeMigration.auto;
```
aweisberg: Should this go in `AccordSpec`?

test/unit/org/apache/cassandra/schema/TransactionalSchemaTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.cql3.QueryProcessor;
+import org.apache.cassandra.db.ConsistencyLevel;
+import org.apache.cassandra.exceptions.InvalidRequestException;
+import org.apache.cassandra.service.consensus.TransactionalMode;
+import org.apache.cassandra.service.consensus.migration.TransactionalMigrationFromMode;
+
+import static java.lang.String.format;
+
+public class TransactionalSchemaTest
```
aweisberg: The test name makes it sound like it is testing the ability to transactionally modify schema when it's really testing test table level transaction config.

The test doesn't validate that it connects to the rest of the request routing stuff correctly. Maybe `MigrationTest` should be parameterized so that it can use `ALTER TABLE` where possible so we cover both JMX and alter table?

test/unit/org/apache/cassandra/schema/TransactionalSchemaTest.java
```diff
@@ -0,0 +1,104 @@
...
+        catch (Throwable t)
+        {
+            Assert.assertTrue(t instanceof InvalidRequestException);
+        }
+        assertTransactionalMode(table, TransactionalMode.full, TransactionalMigrationFromMode.off);
+
+        // explicitly setting the migration mode should work
+        process("ALTER TABLE ks.%s WITH transactional_mode='%s' AND transactional_migration_from='%s'",
+                table, TransactionalMode.off, TransactionalMigrationFromMode.none);
+        assertTransactionalMode(table, TransactionalMode.off, TransactionalMigrationFromMode.none);
```
aweisberg, line 102: A lot of this doesn't really validate that it connects to the rest of the request routing stuff correctly. Maybe `MigrationTest` should be parameterized so that it can use `ALTER TABLE` where possible?

test/unit/org/apache/cassandra/schema/TransactionalSchemaTest.java
```diff
@@ -0,0 +1,104 @@
...
+    public void incompleteMigrationFailure()
+    {
+        String table = "incomplete_table";
+        process("CREATE TABLE ks.%s (k int primary key, v int) WITH transactional_mode='%s' AND transactional_migration_from='%s'",
+                table, TransactionalMode.full, TransactionalMigrationFromMode.off);
+        assertTransactionalMode(table, TransactionalMode.full, TransactionalMigrationFromMode.off);
+
+        try
+        {
+            process("ALTER TABLE ks.%s WITH transactional_mode='%s'", table, TransactionalMode.off);
```
aweisberg: This should work? It's the migration direction reversal request.

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
bdeggleston: yes, or otherwise manually intervene

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
bdeggleston: we only set it if they haven't supplied it, otherwise we return above

src/java/org/apache/cassandra/cql3/statements/schema/CreateTableStatement.java
bdeggleston: there's not, do you think we need a config option?

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationTarget.java
bdeggleston: in that case you'd just manually set the tranaction migration mode to `non` for the table

src/java/org/apache/cassandra/service/consensus/migration/ConsensusTableMigrationState.java
bdeggleston: yeah it was split into multiple files though. I don't _think_ there were any major changes, but I'd give it a once over to be sure

src/java/org/apache/cassandra/cql3/statements/schema/TableAttributes.java
bdeggleston: handled both in create/alter table statements

src/java/org/apache/cassandra/service/accord/txn/TxnQuery.java
bdeggleston: I'd lean towards leaving it out, and I'm not sure what it would look like if we left it. The original check seemed to be more an artifact of LWTs being converted and interop/migration being implemented independently of each other. With the schema config, there isn't really an LWT strategy, your table is either using accord or not, and that, combined with migration state,  informs how LWTs are handled.

src/java/org/apache/cassandra/service/consensus/TransactionalMode.java
bdeggleston: Do you mean the name TransactionalMode? How is it problematic? 

If you mean the individual setting names, I'd agree they could be improved a bit. `mixed` could be called `mixed_unsafe`. I'm not sure what `migration` should be called.

It's not clear how reads are handled for each of these in either case.

Do you have a suggestion for how interop should be broken out? Do you mean as a 3rd schema option?

It sounds like we will want configurable defaults for on/off and/or interop mode that are applied in CreateTableStatement

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
bdeggleston: I think we'd need a migration to occur between the different interop modes as well? Like if we moved from `full` to `migration`, I think we'd want to make sure data written with `full` had been repaired so it would be visible to non-serial reads?

src/java/org/apache/cassandra/service/StorageProxy.java
bdeggleston: We'll have to give this some thought, since that will mean splitting up mutations that write to multiple tables, as well as making write coordination concurrent so that we coordinate normal and accord writes simultaneously

bdeggleston: I think that might be enough of an undertaking to warrant punting it to a follow on ticket. I also wonder if we could get away with just having this restriction for interop.... basically forbidding multi-table / multi-key batches when using accord, or migrating to/from it. It is something we discourage

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
aweisberg: You are right some of them have should require a migration, but you should be able to change interop modes.

From Accord's perspective the only mode that matters is `full` (ignores supplied consistency level, does async commit), and not `full`.

For C* the modes you should be able to switch between unsafe <-> mixed <-> migration. Really `migration` is the only mode that you can safely use on anything. The rest make the assumption the application is not going to mix Accord/non-Accord reads/writes on the same data.

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
aweisberg: Are you going to do anything for this as part of your incremental repair changes?

src/java/org/apache/cassandra/service/consensus/TransactionalMode.java
aweisberg: The name `TransactionalMode` is problematic because this is really hyper focused on interoperability with Accord and we have `LWT` so the word `Transactional` makes it seem like this is how `LWT`s are configured and it's not just that. It's really `AccordInteroperabilityMode`.

I was thinking it might be cleaner to just have an on/off for interoperability, but I have warmed up to the current approach. I just think maybe allowing changing between different interoperability modes should be allowed. More in my comment on alter table.

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
aweisberg: You are right some of them have should require a migration, but you should be able to change interop modes.

From Accord's perspective the only mode that matters is full (ignores supplied consistency level, does async commit), and not full.

For C* the modes you should be able to switch between unsafe <-> mixed <-> migration. Really migration is the only mode that you can safely use on anything. The rest make the assumption the application is not going to mix Accord/non-Accord reads/writes on the same data.

src/java/org/apache/cassandra/cql3/statements/schema/CreateTableStatement.java
aweisberg: I think it's pretty important yeah since otherwise you have to be very careful when creating schema and if you aren't you have to run the migration process. If the migration process were turn key it would be much less of an issue.

src/java/org/apache/cassandra/service/StorageProxy.java
aweisberg: I recall from discussions about some use cases and reviewing their queries that they definitely use unlogged batches and the use case is to split data over multiple partitions so they don't run into the large partition problem.

I would need to resurrect the full query log I have to confirm.

src/java/org/apache/cassandra/service/StorageProxy.java
aweisberg: This check for the system keyspaces is gone now. Should still be kept around as an assertion in case we have test code or otherwise that ends up here?

src/java/org/apache/cassandra/service/accord/txn/TxnQuery.java
aweisberg: The check if whether the key is in a migrating or migrated range from Accord to Paxos, so if the strategy for the table is `TransactionalMode.full` then the key shouldn't be in a migrating or migrated range *from* Accord and still be attempting to execute on Accord.

So is the check useful? It's a bit mid.

This is also another place where we are using the `ConsensusMigrationState` from the current cluster metadata and not the epoch the query executes at! I have a patch set that fixes this in another place I need to update.

src/java/org/apache/cassandra/service/consensus/migration/TransactionalMigrationFromMode.java
```diff
@@ -0,0 +1,74 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.service.consensus.migration;
+
+import org.apache.cassandra.service.consensus.TransactionalMode;
+
+public enum TransactionalMigrationFromMode
```
aweisberg, line 27: Maybe `AccordMigrationFromMode`

and for a comment
```
This tracks the state of a migration either from Paxos -> Accord or Accord -> Paxos. The `TransactionalMode` associated with each transition from a system is how interoperability should be achieved during the migration with various performance/safety tradeoffs.
```

src/java/org/apache/cassandra/service/consensus/migration/TransactionalMigrationFromMode.java
```diff
@@ -0,0 +1,74 @@
...
+{
+    none(null),
```
aweisberg:
```suggestion
    // No migration is in progress. The currently active transaction system could be either Accord or Paxos.
    none(null),
```

aweisberg: Is this comment even technically true? When we finish a migration do we ever actually alter the schema to set this to none or are we basically "always" migrating with the ranges completely migrated and we only ever are in this state when do the very first migration?

test/unit/org/apache/cassandra/schema/TransactionalConfigSchemaTest.java
```diff
@@ -0,0 +1,95 @@
...
+
+    // if an existing table is set to an accord transactional mode, it should be set to migrating
+    @Test
+    public void existingTableMigration()
+    {
+        String table = "existing_table";
+        process("CREATE TABLE ks.%s (k int primary key, v int)", table);
+        assertTransactionalMode(table, TransactionalMode.off, TransactionalMigrationFromMode.none);
+
+        process("ALTER TABLE ks.%s WITH transactional_mode='%s'", table, TransactionalMode.full);
```
aweisberg, line 74: Is this making a table an Accord table with a migration required but without interoperability enabled?

src/java/org/apache/cassandra/service/consensus/migration/TransactionalMigrationFromMode.java
```diff
@@ -0,0 +1,74 @@
...
+    off(TransactionalMode.off),
```
aweisberg, line 30:
```suggestion
    // Unsafe migration from Paxos -> Accord where interoperability is not enabled at all
    off(TransactionalMode.off),
```

src/java/org/apache/cassandra/service/consensus/migration/TransactionalMigrationFromMode.java
```diff
@@ -0,0 +1,74 @@
...
+    unsafe(TransactionalMode.unsafe),
```
aweisberg, line 31:
```suggestion
    // Migrating back to Paxos from unfinished migration to Accord w/unsafe interop mode
    unsafe(TransactionalMode.unsafe),
```

src/java/org/apache/cassandra/service/consensus/migration/TransactionalMigrationFromMode.java
```diff
@@ -0,0 +1,74 @@
...
+    mixed(TransactionalMode.mixed),
```
aweisberg:
```suggestion
    // Migrating back to Paxos from unfinished migration to Accord w/mixed interop mode
    mixed(TransactionalMode.mixed),
```

src/java/org/apache/cassandra/service/consensus/migration/TransactionalMigrationFromMode.java
```diff
@@ -0,0 +1,74 @@
...
+    migration(TransactionalMode.migration),
```
aweisberg:
```suggestion
    // Migrating back to Paxos from unfinished migration to Accord w/migration interop mode
    migration(TransactionalMode.migration),
```

src/java/org/apache/cassandra/service/consensus/migration/TransactionalMigrationFromMode.java
```diff
@@ -0,0 +1,74 @@
...
+    full(TransactionalMode.full);
```
aweisberg, line 34:
```suggestion
    // Migrating from Accord back to Paxos
    full(TransactionalMode.full);
```

src/java/org/apache/cassandra/service/consensus/TransactionalMode.java
```diff
@@ -0,0 +1,136 @@
...
+{
+    off(false, false, false, false),
```
aweisberg, line 60:
```suggestion
    // Running on Paxos V1 or V2 with Accord disabled
    off(false, false, false, false),
```

src/java/org/apache/cassandra/service/consensus/TransactionalMode.java
```diff
@@ -0,0 +1,136 @@
...
+
+    /*
+     * Execute writes through Cassandra via StorageProxy's normal write path. This can lead Accord to compute
+     * multiple outcomes for a transaction that depends on data written by non-SERIAL writes.
+     */
+    unsafe(true, false, false, false),
+
+    /*
+     * Allow mixing of non-SERIAL writes and Accord, but still force BRR through Accord
```
aweisberg:
```suggestion
     * Allow mixing of non-SERIAL writes and Accord, but still force BRR through Accord.
     * This mode makes it safe to perform non-SERIAL or SERIAL reads of Accord data, but unsafe
     * to write data that Accord may attempt to read.
```

src/java/org/apache/cassandra/service/consensus/TransactionalMode.java
```diff
@@ -0,0 +1,136 @@
...
+     */
+    mixed(true, false, false, true),
+
+    /*
+     * Execute writes through Accord skipping StorageProxy's normal write path, but commit
+     * writes at the provided consistency level so they can be read via non-SERIAL consistency levels.
```
aweisberg, line 77:
```suggestion
     * writes at the provided consistency level so they can be read via non-SERIAL consistency levels.
     * This mode makes it safe to read/write data that Accord will read/write.
```

src/java/org/apache/cassandra/service/StorageProxy.java
aweisberg: I confirmed that some use cases definitely do batches to multiple tables and partitions from the full query log I took a while back.

src/java/org/apache/cassandra/service/consensus/migration/TableMigrationState.java
```diff
@@ -0,0 +1,360 @@
...
+        this.migratingAndMigratedRanges = ImmutableList.copyOf(normalize(ImmutableList.<Range<Token>>builder().addAll(migratedRanges).addAll(migratingRanges).build()));
+    }
+
+    public TableMigrationState reverseMigration(ConsensusMigrationTarget target, Epoch epoch)
+    {
+        IPartitioner partitioner = DatabaseDescriptor.getPartitioner();
+        Range<Token> fullRange = new Range<>(partitioner.getMinimumToken(), partitioner.getMinimumToken());
+        List<Range<Token>> allTouched = new ArrayList<>(migratedRanges);
+        allTouched.addAll(migratingRanges);
+        allTouched = Range.deoverlap(allTouched);
```
aweisberg, line 124: As far as I can tell this still has the issue that you can reverse migration for partially migrated ranges. For those ranges some keys are on the old system and some are on the new one so it's not a safe operation to reverse direction.

aweisberg: Actually I think this is safe/fine since all keys already migrated will be key migrated before being used. In fact you could reverse several times and it should be fine? You will migrate on one system before using the other and you are always going to see the latest value for a key because either you picked the correct system, or you did the key migration first.

src/java/org/apache/cassandra/tcm/transformations/AlterSchema.java
aweisberg: Actually I think this is safe/fine since all keys already migrated will be key migrated before being used. In fact you could reverse several times and it should be fine? You will migrate on one system before using the other and you are always going to see the latest value for a key because either you picked the correct system, or you did the key migration first.

aweisberg: Although one random thought is that it's kind of important that the timestamps published by Paxos and Accord have the necessary barriers. IOW Accord should never write a value with a timestamp lower than what Paxos has used.

src/java/org/apache/cassandra/tcm/transformations/AlterSchema.java
```diff
@@ -249,22 +249,73 @@ private static Map<ReplicationParams, Set<KeyspaceMetadata>> groupByReplication(
...
+                .flatMap(diff -> diff.tables.altered.stream())
+                .filter(alt -> alt.before.params.transactionalMigrationFrom != TransactionalMigrationFromMode.none
+                        && alt.after.params.transactionalMigrationFrom == TransactionalMigrationFromMode.none)
+                .map(alt -> alt.after.id)
+                .collect(toImmutableSet());
+
+        if (!completedIds.isEmpty())
+            migrationState = migrationState.withMigrationsCompletedFor(completedIds);
+
+        Map<TableId, TableMetadata> updatedTables = altered.stream()
```
aweisberg: This is unused?

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1221,116 +1218,9 @@ public enum CorruptedTombstoneStrategy
...
-     * coordinators will always produce a consistent state when applying a transaction. Accord performs consensus on
-     * the execution order of transaction and different coordinators witnessing different states not managed by Accord
-     * can produce multiple outcomes for a transaction.
-     *
-     * // TODO (maybe): To safely migrate you would have to route all writes through Accord with the current implementation
-     * // We could do it by range instead in the migration version, but then we need to know when all in flight writes
-     * // are done before marking a range as migrated. Would waiting out the timeout be enough (timeout bugs!)?
-     */
-    public enum NonSerialWriteStrategy
+    public enum TransactionalRangeMigration
```
aweisberg: Maybe this should also get moved into Accord spec. Not really important though.

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
aweisberg: I requested this elsewhere earlier, but can we log any time a forcible migration change happens since it's usually going to be some unsafe combination?

src/java/org/apache/cassandra/service/StorageProxy.java
bdeggleston: deferred to https://issues.apache.org/jira/browse/CASSANDRA-19431

src/java/org/apache/cassandra/service/consensus/TransactionalMode.java
bdeggleston: > It's really AccordInteroperabilityMode.

Its main function is turning accord on or off for a table. Interop modes are included here to avoid having a separate config setting that only makes sense if this one is on

Regarding the naming of the name and/or it's settings, do you have any suggestions?

bdeggleston: for mode names, how about: `off, unsafe, unsafe_writes, mixed_reads, full`? Tthat seem like it more accurately describes the tradeoffs being made w/ each mode

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
bdeggleston: This doesn't prevent you from changing interop modes, but it does prevent you from changing interop mode while there's an unfinished migration. 

Since each of the migration modes provide different consistency guarantees, I think a migration between at least some of them is needed to prevent invalidating programmer assumptions in between them. For instance, you can do normal QUORUM reads during the “migration” (mixed_reads proposed) interop mode. We’ll want to at least run a barrier as a migration so that successfully completed writes don’t become intermittently unreadable since the writes at `full` may have only been persisted on a single node.

When upgrading from an interop mode allowing unsafe writes, don’t think it would be a bad idea to keep the table in a symbolic migrating state for operator visibility and alerting purposes, since the safety of the safer interop mode won’t be available until data is repaired.

As an aside I’d thought the “mixed” interop mode did something like a quorum read for accord operations to prevent non-serial writes from causing multiple outcomes for accord operations, but it looks like I misremembered that (or am not looking hard enough). Assuming I’m not missing something, we should definitely rename mixed to unsafe_writes since non-serial writes can cause multiple outcomes, it’s just a little less dangerous since BRR is routed through accord.

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
bdeggleston: to clarify, have you opened a JIRA that would address this comment? Excluding ranges that aren't migrating wouldn't be too hard here, but am fine with it being deferred to another ticket

src/java/org/apache/cassandra/service/StorageProxy.java
bdeggleston: deferred to https://issues.apache.org/jira/browse/CASSANDRA-19431

src/java/org/apache/cassandra/service/accord/txn/TxnQuery.java
bdeggleston: > The check if whether the key is in a migrating or migrated range from Accord to Paxos, so if the strategy for the table is TransactionalMode.full then the key shouldn't be in a migrating or migrated range from Accord and still be attempting to execute on Accord.

If we do migrations between interop modes then this check would be overly defensive

src/java/org/apache/cassandra/service/consensus/migration/ConsensusRequestRouter.java
bdeggleston: If you don't mind, I'd like to keep the mode name the same, for DDL compactness. I did add an `isMigrating` method to `TransactionalMigrationFromMode` and replaced checks against `none` with it to aid readability. WDYT?

src/java/org/apache/cassandra/service/reads/repair/BlockingReadRepair.java
bdeggleston: assuming this will be done as part of https://issues.apache.org/jira/browse/CASSANDRA-19431

src/java/org/apache/cassandra/service/consensus/migration/TableMigrationState.java
bdeggleston: I think I'd prefer to keep this as is. Since epochs can change during the processing of a query, I think it will be less work to keep this in sync than it will be to ensure you're looking at the migration state and table metadata from the same epoch. I added a sanity check to the ClusterMetadata build method to check that they're in sync

src/java/org/apache/cassandra/service/consensus/migration/TransactionalMigrationFromMode.java
bdeggleston: added comment, will address naming in TransactionalMode thread

test/unit/org/apache/cassandra/schema/TransactionalConfigSchemaTest.java
bdeggleston: yes

src/java/org/apache/cassandra/service/consensus/migration/TransactionalMigrationFromMode.java
bdeggleston: This shouldn't be unsafe. This a transition from accord being off to some accord enabled mode

src/java/org/apache/cassandra/service/consensus/TransactionalMode.java
bdeggleston: Regarding the name TransactionalMode, I see the names Transactional and Accord as having different functions. One refers to the C* feature and and the other refers to the consensus protocol, respectively. Since configuration is user facing, I don’t think naming it accord is appropriate (and think AccordSpec should be renamed to TransactionalSpec). This is similar to how we have a user feature call lwts, and they are implemented w/ paxos.

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
aweisberg: I opened a JIRA. It's actually more complicated because of the need for retries when racing with topology changes.

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
aweisberg: OK, I can sort of see that if you are running an unsafe mode in Cassandra it would require a restart of the migration to actually migrate with the level of safety/correctness of the safe mode (`migration`).

I think the pain point is if you set an unsafe mode and want to switch to a safe mode and restart the migration then you can't. You have to complete the unsafe one which requires a repair to complete and you are exposed to data loss the entire time.

Force migration change still works so I think that is good enough for now.

Accord always respects the consistency levels supplied except in `TransactionMode.full` where everything runs through Accord. It's Cassandra that has all the unsafe ways to operate to try and preserve performance assuming that you won't overlap access in unsafe ways.

src/java/org/apache/cassandra/service/consensus/TransactionalMode.java
aweisberg: OK, that makes sense.

src/java/org/apache/cassandra/service/consensus/migration/ConsensusRequestRouter.java
aweisberg: WFM

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
```diff
@@ -553,6 +557,41 @@ public void validate(ClientState state)
...
+
+            if (modeChange && next.transactionalMode.accordIsEnabled && !DatabaseDescriptor.getAccordTransactionsEnabled())
+                throw ire(format("Cannot change transactional mode to %s for %s.%s with accord_transactions_enabled set to false",
+                                 next.transactionalMode, keyspaceName, tableName));
+
+            // user is manually updating migration mode, don't interfere
+            if (forceMigrationChange)
+            {
+                logger.warn("Forcing unsafe migration change from {} to {}", prev.transactionalMigrationFrom, next.transactionalMigrationFrom);
```
aweisberg: Did we also log `transactionalMode` already? Can we include it in this statement?

src/java/org/apache/cassandra/service/consensus/migration/TransactionalMigrationFromMode.java
aweisberg: OK, can you add a comment to that effect?

src/java/org/apache/cassandra/service/consensus/TransactionalMode.java
aweisberg: I got this slightly wrong. It makes it safe to perform SERIAL writes/reads and non-SERIAL reads. The same is true of the "unsafe" mode.

aweisberg (approved):

bdeggleston:
So I took a second look at your requested changes to the TransactionalMigrationFromMode, and didn’t add them since they don’t make sense in the context of accord <-> accord changes we talked about here: https://github.com/apache/cassandra/pull/3032/files#r1446421908
Without the assumption that a migration is always between paxos and accord, the comments don’t say much, so I just left them out. Let me know if you disagree and we can figure something out, I may be overlooking something

PR #3034, opened 2024-01-08
CASSANDRA-19107 Revert unnecessary read lock acquisition when reading ring version in TokenMetadata introduced in CASSANDRA-16286

maedhroz:
patch by Caleb Rackliffe; reviewed by Francisco Guerrero for CASSANDRA-19107


frankgh (approved):
+1 Makes sense that we don't need the read lock just to get the ringVersion.

maedhroz:
Committed as https://github.com/apache/cassandra/commit/d5cea135c98bb98b16b215d309ead22e86f1329f

PR #3038, opened 2024-01-09
Sort filter operation metrics

pkolaczk:
New SAI metrics:
- `SortThenFilterQueriesCompleted`
- `FilterThenSortQueriesCompleted`

The counters are increased by queries that involve both sorting and filtering.


pkolaczk:
wrong project

PR #3040, opened 2024-01-11
Handle MIN_TOKEN placement correctly.

ifesdjeen:
Patch by Alex Petrov, reviewed by Sam Tunnicliffe for CASSANDRA-19262.

### DIFF
--- a/src/java/org/apache/cassandra/locator/SimpleStrategy.java
+++ b/src/java/org/apache/cassandra/locator/SimpleStrategy.java
@@ -17,12 +17,15 @@
  */
 package org.apache.cassandra.locator;
 
-import java.util.*;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Iterator;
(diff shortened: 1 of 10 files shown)


src/java/org/apache/cassandra/locator/SimpleStrategy.java
```diff
@@ -158,13 +161,4 @@ public Collection<String> recognizedOptions(ClusterMetadata metadata)
     {
         return Collections.singleton(REPLICATION_FACTOR);
     }
-
-    protected static void prepareOptions(Map<String, String> options, Map<String, String> previousOptions)
```
beobal, line 162: This shouldn't be removed. It's only invoked by reflection (in `AbstractReplicationStrategy::prepareReplicationStrategyOptions`) so it appears unused. Restoring it fixes the 6 failing unit tests in the CI results.

PR #3045, opened 2024-01-15
[CEP-21] Improve setup and initialisation of LocalLog

ifesdjeen:
Patch by Alex Petrov; reviewed by Sam Tunnicliffe for CASSANDRA-19217.

### DIFF
--- a/src/java/org/apache/cassandra/tcm/ClusterMetadata.java
+++ b/src/java/org/apache/cassandra/tcm/ClusterMetadata.java
@@ -26,6 +26,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 
 import com.google.common.annotations.VisibleForTesting;
@@ -816,6 +817,15 @@ public class ClusterMetadata
             return null;
         return service.metadata();
     }
+
+    public static Optional<ClusterMetadata> currentOptional()
+    {
+        ClusterMetadataService service = ClusterMetadataService.instance();
+        if (service == null)
+            return Optional.empty();
+        return Optional.of(service.metadata());
+    }
+
 
     public NodeId myNodeId()
     {
--- a/src/java/org/apache/cassandra/tcm/ClusterMetadata.java
+++ b/src/java/org/apache/cassandra/tcm/ClusterMetadata.java
@@ -818,15 +818,6 @@ public class ClusterMetadata
         return service.metadata();
     }
 
-    public static Optional<ClusterMetadata> currentOptional()
-    {
-        ClusterMetadataService service = ClusterMetadataService.instance();
-        if (service == null)
-            return Optional.empty();
-        return Optional.of(service.metadata());
-    }
-
-
     public NodeId myNodeId()
     {
         return directory.peerId(FBUtilities.getBroadcastAddressAndPort());
--- a/src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
+++ b/src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
@@ -159,7 +159,7 @@ public class ClusterMetadataService
         Processor localProcessor;
         if (CassandraRelevantProperties.TCM_USE_ATOMIC_LONG_PROCESSOR.getBoolean())
         {
-            log = logSpec.syncForTests().createLog();
+            log = logSpec.sync().createLog();
             localProcessor = wrapProcessor.apply(new AtomicLongBackedProcessor(log, logSpec.isReset()));
             fetchLogHandler = new FetchCMSLog.Handler((e, ignored) -> logSpec.storage().getLogState(e));
         }
@@ -258,7 +258,7 @@ public class ClusterMetadataService
                                                    // we do not need a post-commit hook for tools
                                                }
                                            })
-                                           .syncForTests()
+                                           .sync()
                                            .withStorage(new AtomicLongBackedProcessor.InMemoryStorage());
         LocalLog log = logSpec.createLog();
         ClusterMetadataService cms = new ClusterMetadataService(new UniformRangePlacement(),
@@ -272,7 +272,7 @@ public class ClusterMetadataService
                                                                 null,
                                                                 new PeerLogFetcher(log));
 
-        log.readyForTests();
+        log.readyUnchecked();
         log.bootstrap(FBUtilities.getBroadcastAddressAndPort());
         ClusterMetadataService.setInstance(cms);
     }
--- a/src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
+++ b/src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
@@ -40,6 +40,7 @@ import org.apache.cassandra.concurrent.ScheduledExecutors;
 import org.apache.cassandra.config.CassandraRelevantProperties;
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.exceptions.ExceptionCode;
+import org.apache.cassandra.exceptions.StartupException;
 import org.apache.cassandra.io.util.FileInputStreamPlus;
 import org.apache.cassandra.io.util.FileOutputStreamPlus;
 import org.apache.cassandra.locator.InetAddressAndPort;
@@ -47,10 +48,10 @@ import org.apache.cassandra.metrics.TCMMetrics;
 import org.apache.cassandra.net.IVerbHandler;
 import org.apache.cassandra.schema.DistributedSchema;
 import org.apache.cassandra.schema.ReplicationParams;
+import org.apache.cassandra.tcm.listeners.SchemaListener;
 import org.apache.cassandra.tcm.log.Entry;
 import org.apache.cassandra.tcm.log.LocalLog;
 import org.apache.cassandra.tcm.log.LogState;
-import org.apache.cassandra.tcm.log.LogStorage;
 import org.apache.cassandra.tcm.log.Replication;
 import org.apache.cassandra.tcm.membership.NodeId;
 import org.apache.cassandra.tcm.membership.NodeVersion;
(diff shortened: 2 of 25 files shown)


src/java/org/apache/cassandra/tcm/ClusterMetadata.java
```diff
@@ -817,6 +818,15 @@ public static ClusterMetadata currentNullable()
         return service.metadata();
     }
 
+    public static Optional<ClusterMetadata> currentOptional()
```
krummas, line 821: should we migrate any `currentNullable()` uses to this instead? Or just use `currentNullable` in `scanLogForPeriod`? (which will soon be gone anyway)

src/java/org/apache/cassandra/tcm/ClusterMetadata.java
```diff
@@ -817,6 +818,15 @@ public static ClusterMetadata currentNullable()
...
+    {
+        ClusterMetadataService service = ClusterMetadataService.instance();
```
krummas: should we migrate any `currentNullable()` uses to this instead? Or just use `currentNullable` in `scanLogForPeriod`? (which will soon be gone anyway)

src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
```diff
@@ -243,13 +243,24 @@ public static void initializeForTools(boolean loadSSTables)
...
-        log.ready();
+        ClusterMetadata emptyFromSystemTables = emptyWithSchemaFromSystemTables(Collections.singleton("DC1"))
+                                                .forceEpoch(Epoch.EMPTY);
+
+        LocalLog.LogSpec logSpec = LocalLog.logSpec()
+                                           .withInitialState(emptyFromSystemTables)
+                                           .loadSSTables(loadSSTables)
+                                           .initializeKeyspaceInstances(false)
+                                           .withDefaultListeners(false)
+                                           .withListener(new SchemaListener(loadSSTables) {
```
krummas, line 254: why do we add this empty listener? Isn't withDefaultListeners(false) above enough?

src/java/org/apache/cassandra/tcm/Startup.java
```diff
@@ -67,6 +50,15 @@
 import org.apache.cassandra.tcm.transformations.UnsafeJoin;
 import org.apache.cassandra.tcm.transformations.cms.Initialize;
 import org.apache.cassandra.utils.FBUtilities;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
```
krummas: nit; not sure why these were moved, but code style says that java.* imports go first

src/java/org/apache/cassandra/tcm/Startup.java
```diff
@@ -298,25 +294,20 @@ public static void reinitializeWithClusterMetadata(String fileName, Function<Pro
 
         if (!metadata.isCMSMember(FBUtilities.getBroadcastAddressAndPort()))
             throw new IllegalStateException("When reinitializing with cluster metadata, we must be in the CMS");
+
         // can use local dc here since we know local host in the cms:
-        ClusterMetadata emptyFromSystemTables = emptyWithSchemaFromSystemTables(Collections.singleton(DatabaseDescriptor.getLocalDataCenter()));
-        metadata.schema.initializeKeyspaceInstances(DistributedSchema.empty());
-        metadata = metadata.forceEpoch(metadata.epoch.nextEpoch());
```
krummas, line 304: Don't we need this anymore?

IIRC the reasoning when adding it was that someone dumps the metadata at epoch X - then, when we force it here, it should always be a new epoch

src/java/org/apache/cassandra/tcm/log/LocalLog.java
```diff
@@ -561,15 +586,65 @@ private void notifyPostCommit(ClusterMetadata before, ClusterMetadata after, boo
...
+        else
+        {
+            logger.info("Adding specified listeners to LocalLog");
+            spec.listeners.forEach(this::addListener);
+            spec.changeListeners.forEach(this::addListener);
+            spec.asyncChangeListeners.forEach(this::addListener);
+        }
+
+        logger.info("Notifying all registered listeners of both pre and post commit event");
+        notifyListeners(spec.initial);
```
krummas: in the LocalLog constructor we set committed to spec.initial - so here I think we always notifyListeners with the same prev/next metadatas - at least in the reset case we should always notify from an empty cluster metadata I think?

looks like we handle this for schema with that new `isFirst` boolean but I think we need it for all listeners (say we change the state of a nodeid from registered to joined - it wouldn't get properly added to gossip (I think?))

test/distributed/org/apache/cassandra/distributed/test/log/ClusterMetadataTestHelper.java
```diff
@@ -119,13 +121,16 @@ public static void setInstanceForTest()
     public static ClusterMetadataService instanceForTest()
     {
```
krummas, line 126: this method is unused

krummas (commented):
first pass, looks like a nice cleanup in general

a few comments inline

src/java/org/apache/cassandra/tcm/ClusterMetadata.java
ifesdjeen: removed in favour of the nullable version.

src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
ifesdjeen: We need to initialize keyspaces (which happens in the pre-commit), but we do not need schema keyspace update, gossip hook, and other things that are happening in the post-commit hook.

src/java/org/apache/cassandra/tcm/Startup.java
ifesdjeen: you're right; my bad, did not do it conciously

src/java/org/apache/cassandra/tcm/Startup.java
ifesdjeen: You are right; i think I have removed this together with `metadata.schema.initializeKeyspaceInstances` by mistake.

src/java/org/apache/cassandra/tcm/ClusterMetadata.java
beobal: > removed in favour of the nullable version.

doesn't look like this is pushed?

src/java/org/apache/cassandra/tcm/log/LocalLog.java
```diff
@@ -96,25 +98,59 @@ public abstract class LocalLog implements Closeable
...
 
         private final Set<LogListener> listeners = new HashSet<>();
         private final Set<ChangeListener> changeListeners = new HashSet<>();
         private final Set<ChangeListener.Async> asyncChangeListeners = new HashSet<>();
 
+        private LogSpec()
+        {
+        }
+
+        public LogSpec syncForTests()
```
beobal: should we rename to `syncForTestsAndTools()` or just to `sync()` and add javadoc to explain intended usages?

src/java/org/apache/cassandra/tcm/log/LocalLog.java
```diff
@@ -561,15 +586,65 @@ private void notifyPostCommit(ClusterMetadata before, ClusterMetadata after, boo
             ScheduledExecutors.optionalTasks.submit(() -> listener.notifyPostCommit(before, after, fromSnapshot));
     }
 
+    /**
+     * Essentially same as `ready` but throws an unchecked exception
+     */
+    @VisibleForTesting
+    public final ClusterMetadata readyForTests()
```
beobal: convention for this is `xxxUnchecked()`

src/java/org/apache/cassandra/tcm/Startup.java
```diff
@@ -141,18 +133,16 @@ public static void initializeAsFirstCMSNode()
 
     public static void initializeAsNonCmsNode(Function<Processor, Processor> wrapProcessor) throws StartupException
     {
-        LocalLog.LogSpec logSpec = new LocalLog.LogSpec().withStorage(LogStorage.SystemKeyspace)
-                                                         .withDefaultListeners();
+        LocalLog.LogSpec logSpec = LocalLog.logSpec()
+                                           .withStorage(LogStorage.SystemKeyspace)
+                                           .afterReplay(Startup::scrubDataDirectories)
```
beobal, line 145: just checking I understood correctly, is the benefit of delegating this to the `LocalLog` that it can be run before any initial listeners are triggered in `ready()`?

src/java/org/apache/cassandra/tcm/log/LocalLog.java
```diff
@@ -96,25 +98,59 @@ public abstract class LocalLog implements Closeable
...
+            this.async = true;
+            return this;
+        }
+
         public LogSpec withDefaultListeners()
         {
             return withDefaultListeners(true);
         }
 
+        public LogSpec initializeKeyspaceInstances(boolean initializeKeyspaceInstances)
```
beobal, line 147: This seems only to be used in conjunction with `withDefaultListeners` - i.e. error if this is true but the default listeners are not added and vice versa. Is it just an extra layer of validation to check the caller's intent or is there something else it provides?

src/java/org/apache/cassandra/tcm/log/LogStorage.java
```diff
@@ -18,12 +18,16 @@
...
+import com.google.common.annotations.VisibleForTesting;
+
 import org.apache.cassandra.tcm.Epoch;
 
 public interface LogStorage extends LogReader
 {
     void append(long period, Entry entry);
     LogState getLogState(Epoch since);
+    @VisibleForTesting
+    void truncate();
```
beobal: The only place which calls this already hardcodes the instance as `LogStorage.SystemKeyspace` so we could just call cast it or even inline the truncation there (`ServerTestUtils::recreateCMS`)

test/unit/org/apache/cassandra/hints/HintsUpgradeTest.java
```diff
@@ -47,6 +48,7 @@
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+@Ignore("TODO: TCM")
```
beobal, line 51: Is this newly failing?

test/unit/org/apache/cassandra/service/ClientWarningsTest.java
```diff
@@ -74,12 +74,12 @@ public void testUnloggedBatch() throws Exception
         {
             client.connect(false);
 
-            QueryMessage query = new QueryMessage(createBatchStatement2(1), QueryOptions.DEFAULT);
-            Message.Response resp = client.execute(query);
+            Message.Response resp = executeWithRetries(client,
```
beobal, line 77: Is this specific to this patch or general improvements?

beobal (changes requested):
Nice, this does definitely improve things quite a bit.

PR #3044, opened 2024-01-12
CASSANDRA-18275 Initial in-tree port of Alex's single-node/single-threaded SAI test

maedhroz:

PR #3054, opened 2024-01-19
CASSANDRA-19278 Avoid over-skipping of key iterators from static column indexes during mixed intersections

maedhroz:
patch by Caleb Rackliffe; reviewed by ? for CASSANDRA-19278


src/java/org/apache/cassandra/index/sai/disk/RowMapping.java
```diff
@@ -173,18 +166,15 @@ public void add(PrimaryKey key, long sstableRowId) throws InMemoryTrie.SpaceExha
...
         count++;
     }
 
     /**
-     * Returns the SSTable row Id for a {@link PrimaryKey}
+     * Returns the SSTable row ID for a {@link PrimaryKey}
      *
      * @param key the {@link PrimaryKey}
-     * @return a valid SSTable row Id for the {@link PrimaryKey} or -1 if the {@link PrimaryKey} doesn't exist
+     * @return a valid SSTable row ID for the {@link PrimaryKey} or -1 if the {@link PrimaryKey} doesn't exist
```
maedhroz, line 197: I'm OCD about using all-caps `ID` in comments :)

src/java/org/apache/cassandra/index/sai/disk/SSTableKeyTracker.java
```diff
@@ -0,0 +1,57 @@
...
+
+    public long maxSSTableRowId = -1;
+
+    protected int count;
+    
+    protected void updateMinMaxKeys(PrimaryKey key)
+    {
+        // Data is written in primary key order. If a schema contains clustering keys, it may also contain static
+        // columns. We track the min and max static keys separately here so that we can pass them to the segment
+        // metadata for indexes on static columns.
```
maedhroz, line 38: I more or less created `SSTableKeyTracker` to avoid duplicating this comment and logic...

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
```diff
@@ -77,7 +77,19 @@ protected PrimaryKey computeNext()
...
+                    // Note that we will either have a data model that produces SKINNY primary keys or a data model
+                    // that produces some combination of WIDE and STATIC prikary keys.
+                    if (nextKey.kind() == PrimaryKey.Kind.WIDE || nextKey.kind() == highestKey.kind())
+                        // We can always skip if the target is of the same kind or this range is non-static. 
+                        nextKey = nextOrNull(range, highestKey);
+                    else if (nextKey.kind() == PrimaryKey.Kind.STATIC && nextKey.compareTo(highestKey) < 0)
+                        // For a range of static keys, only skip if we'e advanced to a new partition, and when we
+                        // do, skip to an actual static key. We may otherwise skip too far, as static row IDs always
+                        // precede non-static ones in on-disk postings lists.
+                        nextKey = nextOrNull(range, highestKey.toStatic());
```
maedhroz: This is the core of the fix. Two problems:

1.) We were advancing streams of STATIC `PrimaryKey` objects from static column indexes before keys from the same partition from non-static column indexes could match. This was discovered by Harry on in-memory indexes.

2.) When we did skip the STATIC key streams, we could skip w/ a non-static `PrimaryKey` object, which would map to a posting row ID that was after the row ID of the static row, and so we could skip past the static `PrimaryKey` we wanted. (This affected on-disk indexes only.)

test/unit/org/apache/cassandra/index/sai/cql/RandomIntersectionTest.java
```diff
@@ -55,23 +59,29 @@ public class RandomIntersectionTest extends SAIRandomizedTester
...
+            parameters.add(new Object[] { "Large partition unrestricted, high, high, " + mode, false, true, true, true, mode });
+            parameters.add(new Object[] { "Large partition unrestricted, low, low, " + mode, false, true, false, false, mode });
+            parameters.add(new Object[] { "Large partition unrestricted, high, low, " + mode, false, true, true, false, mode });
+            parameters.add(new Object[] { "Small partition restricted, high, high, " + mode, true, false, true, true, mode });
+            parameters.add(new Object[] { "Small partition restricted, low, low, " + mode, true, false, false, false, mode });
+            parameters.add(new Object[] { "Small partition restricted, high, low, " + mode, true, false, true, false, mode });
+            parameters.add(new Object[] { "Small partition unrestricted, high, high, " + mode, false, false, true, true, mode });
+            parameters.add(new Object[] { "Small partition unrestricted, low, low, " + mode, false, false, false, false, mode });
+            parameters.add(new Object[] { "Small partition unrestricted, high, low, " + mode, false, false, true, false, mode });
```
maedhroz: In short, the test now looks at `static AND static`, `regular AND regular`, and `regular AND static`.

test/unit/org/apache/cassandra/index/sai/cql/StaticColumnIndexTest.java
```diff
@@ -49,7 +49,26 @@ public void staticIndexAndNonStaticIndex() throws Throwable
         execute("INSERT INTO %s(pk, ck,       val2) VALUES(?, ?,    ?)", 1, 2,     2000);
         execute("INSERT INTO %s(pk, ck, val1, val2) VALUES(?, ?, ?, ?)", 2, 1, 40, 2000);
 
-        beforeAndAfterFlush(() -> assertRows(execute("SELECT pk, ck, val1, val2 FROM %s WHERE val1 = 20 AND val2 = 2000 ALLOW FILTERING"),
+        beforeAndAfterFlush(() -> assertRows(execute("SELECT pk, ck, val1, val2 FROM %s WHERE val1 = 20 AND val2 = 2000"),
                                              row(1, 2, 20, 2000)));
     }
+
+    @Test
+    public void staticAndNonStaticRangeIntersection() throws Throwable
```
maedhroz, line 57: This is a concrete reproduction of the bug Harry found (see https://github.com/apache/cassandra/pull/3044)

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
maedhroz: The thing I'm most worried about here is static/static intersections and whether they terminate in all cases.

pkolaczk: It is not very obvious to me it always terminates. This `if / else if` has a theoretical path where it does *not* advance the range at all. 

As I understood this fix, we don't want to advance a STATIC key until all non-static keys were advanced - is it correct? A static key matches the whole partition, so it must intersect with all the wide keys from another range, and can be advanced only *after* those were used and reported as the result. It is *probably* correct, but very, very non obvious and requires a lot of additional knowledge (like e.g. assuming there can be either only WIDE/STATIC or SKINNY but never mixed).

So maybe instead of doing such complex conditions (I am really staring at it for 30 minutes and trying to understand the logic, constantly switching between "Oh I get it" and "Oh no, I don't") could you split the loop into two loops - one going over non-static keys and trying to advance them, and then another one trying to advance the static ones (only if all non-static agree and the highestkey jumped over the static ones) ?

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
```diff
@@ -77,7 +77,19 @@ protected PrimaryKey computeNext()
...
+
                     if (nextKey == null || nextKey.compareTo(highestKey) > 0)
```
pkolaczk: What if `highestKey` is STATIC? It may mess up the algorithm if there are other `WIDE` keys, because it would basically make this condition false, as STATIC key == any wide keys in the same partition. So it could break intersecting other wide keys within the partition.

I think this would manifest if we had 2 regular columns and intersect it with a static column.
Fortunately the intersection clause limit = 2 would not allow it, but a user can change it and allow more than 2 columns in an intersection.

maedhroz: Would it break end-to-end, or would it just return a `STATIC` `PrimaryKey` from the intersection, which would potentially cause more post-filtering than we want in `ResultRetriever`? In any case, I think in the case where we have a `STATIC` `highestKey`, and `WIDE` `PrimaryKey` can be treated as higher, and we could probably make that change without changing any of the existing `compareTo()` behavior, which seems to open up a can of worms.

In short, I think we could fix this doing something like...

```
if (nextKey == null || nextKey.compareTo(highestKey) > 0 || (nextKey.compareTo(highestKey) == 0 && nextKey.kind() == WIDE && highestKey.kind() == STATIC))
```

...or just add a `compareToStrict()` to `PrimaryKey` that encapsulates this. We **DO** want to avoid a case where we post-filter a whole partition when 2 `WIDE` key iterators don't match anything from each other in the context of a match from a `STATIC` key iterator. Let me work on this a bit...

maedhroz: I've added a new commit with a fix and further testing that should address this. Let me know what you think...

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
maedhroz: While I wasn't quite able to settle on a two-loop solution that I liked, I've tried to a.) further document our assumptions, b.) create some reasonable guardrails (see the new intersection compatibility check in `buildIterator()` and `Kind#isIntersectable()`), and c.) expand the randomized tests even further.

I'm convinced, at this point, that the patch is correct, and that even a rewrite of this intersection logic at a later time  would have sufficient tests ensuring it wouldn't make the same mistakes :)

Let me know what you think...

pkolaczk: I am quite convinced your patch is correct, but I managed to find a way to make the conditions in the main loop simpler (and still leaving only one loop). Now there is exactly one condition to decide whether we bump up the iterator, and another one to decide how we advance (so we advance to the right partition in case of static column):

https://github.com/apache/cassandra/commit/b8b15e2a7c9c0a635ab40bcd9992476f0aa37c25

src/java/org/apache/cassandra/index/sai/disk/SSTableKeyTracker.java
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.index.sai.disk;
+
+import org.apache.cassandra.index.sai.utils.PrimaryKey;
+
+public abstract class SSTableKeyTracker
+{
+    public PrimaryKey minKey;
```
ifesdjeen: Looks like min/max keys are currently only used exclusively (i.e. if there are statics, static will be used, otherwise regular), usages in MemtableIndexFlusher and SegmentBuilder also show mutual exclusivity. Should this logic just be encapsulated in this class, or are there usages that I am missing?

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
```diff
@@ -77,8 +81,23 @@ protected PrimaryKey computeNext()
...
-                    if (nextKey == null || nextKey.compareTo(highestKey) > 0)
+                    PrimaryKey nextKey = range.getCurrent();
+
+                    // Note that we will either have a data model that produces SKINNY primary keys or a data model
+                    // that produces some combination of WIDE and STATIC prikary keys.
+                    if (nextKey.kind() == Kind.WIDE || nextKey.kind() == highestKey.kind())
+                        // We can always skip if the target is of the same kind or this range is non-static. 
+                        nextKey = nextOrNull(range, highestKey);
+                    else if (nextKey.kind() == Kind.STATIC && nextKey.compareTo(highestKey) < 0)
+                        // For a range of static keys, only skip if we'e advanced to a new partition, and when we
```
ifesdjeen: nit: "skip if we've" (`v` is missing)

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
```diff
@@ -77,8 +81,23 @@ protected PrimaryKey computeNext()
...
-                    PrimaryKey nextKey = nextOrNull(range, highestKey);
-                    if (nextKey == null || nextKey.compareTo(highestKey) > 0)
+                    PrimaryKey nextKey = range.getCurrent();
+
+                    // Note that we will either have a data model that produces SKINNY primary keys or a data model
+                    // that produces some combination of WIDE and STATIC prikary keys.
+                    if (nextKey.kind() == Kind.WIDE || nextKey.kind() == highestKey.kind())
+                        // We can always skip if the target is of the same kind or this range is non-static. 
+                        nextKey = nextOrNull(range, highestKey);
+                    else if (nextKey.kind() == Kind.STATIC && nextKey.compareTo(highestKey) < 0)
```
ifesdjeen: here, `nextKey.compareTo(highestKey)` , is the reasoning behind this that nextKey is guaranteed to be static, but highestKey can only be non-static? I can imagine situations where we have two partitions without rows, and only with clusterings. Won’t we trigger `==` in this case?

ifesdjeen (changes requested):
Left a coule of small comments.

Also, implemented a small static torture test (albeit will only work for trunk, so I cherry picked your commits to trunk to test, too).

Commit is here: https://github.com/apache/cassandra/compare/trunk...ifesdjeen:cassandra:CASSANDRA-19278-trunk

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
maedhroz: Fixed in Piotr's version

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
maedhroz: This has changed slightly w/ Piotr's version, but if the `highestKey` and `nextKey` had the same `kind()`, we would have hit the previous `if` condition on line 88.

src/java/org/apache/cassandra/index/sai/disk/SSTableKeyTracker.java
maedhroz: You're right about `SegmentBuilder`. It was never necessary to worry about this there, as instances deal w/ single column indexes. `RowMapping` is global to an SSTable being flushed though, so I'm just going to localize it to that class and get rid of `SSTableKeyTracker`.

pkolaczk (approved):

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
pkolaczk: `compareToStrict` looks good. Also thanks for adding a test that exposes that problem.

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
pkolaczk: Funny how I had exactly same question and realized later the if condition on line 88 would catch that case. 
Yeah, it was a bit non-obvious, but looks correct. Anyway, the final version has fewer conditions. Take a look at that one.

src/java/org/apache/cassandra/index/sai/iterators/KeyRangeIntersectionIterator.java
```diff
@@ -56,52 +62,82 @@ private KeyRangeIntersectionIterator(Builder.Statistics statistics, List<KeyRang
...
     {
-        // Range iterator that has been advanced in the previous cycle of the outer loop.
-        // Initially there hasn't been the previous cycle, so set to null.
-        int alreadyAvanced = -1;
-
-        // The highest primary key seen on any range iterator so far.
+        // Advance one iterator to the next key and remember the key as the highest seen so far.
         // It can become null when we reach the end of the iterator.
-        PrimaryKey highestKey = getCurrent();
+        // If there are both static and non-static keys being iterated here, we advacne a non-static one,
```
ifesdjeen, line 67: Nit: advacne -> advance

ifesdjeen (approved):

maedhroz:
Committed as https://github.com/apache/cassandra/commit/87f2a7b7637a346cdeea23e0b74d9968ccfca619

PR #3058, opened 2024-01-21
Reproduce CASSANDRA-17401

jaydeepkumar1984:
**Reproduce** [Cassandra-17401](https://issues.apache.org/jira/browse/CASSANDRA-17401)

Reproducing the above ticket is extremely tricky because it totally depends on the timing of the multiple threads. In order to reproduce this issue, I've modified the. main code and introduced some delays in scheduling. The test case written in this pull request always fails.

PR #3059, opened 2024-01-21
No need to evict the statements as it creates race condition between …

jaydeepkumar1984:
As part of [CASSANDRA-17248](https://issues.apache.org/jira/browse/CASSANDRA-17248) (in C* 4.0.2), the following code was introduced in [QueryProcessor.java](https://github.com/apache/cassandra/commit/242f7f9b18db77bce36c9bba00b2acda4ff3209e#r137491766)

```
// Make sure the missing one is going to be eventually re-prepared                
evictPrepared(hashWithKeyspace);                
evictPrepared(hashWithoutKeyspace); 
```
This code could very well create a race condition between two calls of [QueryProcessor::prepare](https://github.com/apache/cassandra/blob/cassandra-4.0/src/java/org/apache/cassandra/cql3/QueryProcessor.java#L575) call in which one thread is adding and another thread is silently removing from the cache. Imagine there are thousands of threads calling the API, and then it might be possible for one thread to update the cache and another thread to remove it.

If we look at the code of [Cassandra 3.0.25](https://github.com/apache/cassandra/blob/cassandra-3.0.25/src/java/org/apache/cassandra/cql3/QueryProcessor.java#L391), then such eviction was not present. Hence, this is a regression for the post 4.0.2 version of Cassandra. To fix this issue, we should not evict the cache entries, i.e., the above-mentioned code path introduced since C* 4.0.2 is no longer required on trunk/5.0, as discussed in the [ticket](https://issues.apache.org/jira/browse/CASSANDRA-17401?focusedCommentId=18046549&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-18046549).

For the 4.0 branch, we will keep the eviction logic but optimize it further to address [CASSANDRA-17401](https://issues.apache.org/jira/browse/CASSANDRA-17401), as it is essential for folks on pre-4.0.1 trying to upgrade to the latest 4.0 version.

[Cassandra-17401](https://issues.apache.org/jira/browse/CASSANDRA-17401)

**Reproducing** this is extremely tricky because it totally depends on the timing of the multiple threads, here is the pull request that reproduces it: https://github.com/apache/cassandra/pull/3058

### DIFF
--- a/src/java/org/apache/cassandra/cql3/QueryProcessor.java
+++ b/src/java/org/apache/cassandra/cql3/QueryProcessor.java
@@ -730,13 +730,6 @@ public class QueryProcessor implements QueryHandler
                 return createResultMessage(hashWithKeyspace, cachedWithKeyspace);
             }
         }
-        else
-        {
-            // Make sure the missing one is going to be eventually re-prepared
-            evictPrepared(hashWithKeyspace);
-            evictPrepared(hashWithoutKeyspace);
-        }
-
         Prepared prepared = parseAndPrepare(queryString, clientState, false);
         CQLStatement statement = prepared.statement;
 
--- a/test/unit/org/apache/cassandra/cql3/validation/miscellaneous/PreparedStatementTest.java
+++ b/test/unit/org/apache/cassandra/cql3/validation/miscellaneous/PreparedStatementTest.java
@@ -1,0 +1,85 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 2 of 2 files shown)


ifesdjeen (approved):
+1, assuming 
a) this patch is only committed to trunk
b) 5.0 / 5.1 or whatever version we consider to be an upgrade path for post-15252 still has this logic to preclude incorrect statements being cached during upgrade

jaydeepkumar1984:
Thanks a lot @ifesdjeen for the review! Sure, I will commit this patch to _trunk_ only. 
I will also make sure the upgrade path has this version in a separate PR.

jaydeepkumar1984:
I have just committed this approved PR to trunk. 
https://github.com/apache/cassandra/commit/a06df099f4e1c6265d25c94336c493aded3aa644 & https://github.com/apache/cassandra/commit/aa4da838ec0d05f7c7494536082271a6ed135fa1

Here is the PR for 5.0 https://github.com/jaydeepkumar1984/cassandra/pull/67 - please review.

PR #3060, opened 2024-01-22
Harry model that supports value overrides: an ability to provide spec…

ifesdjeen:
…ific values for clustering, regular, and static columns

Patch by Alex Petrov; reviewed by TBD for CASSANDRA-19284

### DIFF
--- a/test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderIntegrationTest.java
+++ b/test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderIntegrationTest.java
@@ -41,9 +41,9 @@ import org.apache.cassandra.harry.visitors.ReplayingVisitor;
 
 public class HistoryBuilderIntegrationTest extends IntegrationTestBase
 {
-    private final long seed = 1L;
-    private final int STEPS_PER_ITERATION = 1_000;
-    private final int MAX_PARTITIONS = 50;
+    private static final long SEED = 1L;
+    private static final int STEPS_PER_ITERATION = 1_000;
+    private static final int MAX_PARTITIONS = 50;
 
     @Test
     public void simpleDSLTest() throws Throwable
@@ -57,12 +57,12 @@ public class HistoryBuilderIntegrationTest extends IntegrationTestBase
             sut.schemaChange(schema.compile().cql());
 
             ModelChecker<SingleOperationBuilder> modelChecker = new ModelChecker<>();
-            JdkRandomEntropySource rng = new JdkRandomEntropySource(new Random(seed));
+            JdkRandomEntropySource rng = new JdkRandomEntropySource(new Random(SEED));
 
             TokenPlacementModel.ReplicationFactor rf = new TokenPlacementModel.SimpleReplicationFactor(1);
 
             int maxPartitionSize = 100;
-            modelChecker.init(new HistoryBuilder(seed, maxPartitionSize, 10, schema, rf))
+            modelChecker.init(new HistoryBuilder(SEED, maxPartitionSize, 10, schema, rf))
                         .step((history) -> {
                             return history.insert();
                         })
@@ -117,7 +117,7 @@ public class HistoryBuilderIntegrationTest extends IntegrationTestBase
 
                             return true;
                         })
-                        .run(STEPS_PER_ITERATION, seed);
+                        .run(STEPS_PER_ITERATION, SEED);
         }
     }
 
@@ -133,12 +133,12 @@ public class HistoryBuilderIntegrationTest extends IntegrationTestBase
             sut.schemaChange(schema.compile().cql());
 
             ModelChecker<HistoryBuilder> modelChecker = new ModelChecker<>();
-            JdkRandomEntropySource rng = new JdkRandomEntropySource(new Random(seed));
+            JdkRandomEntropySource rng = new JdkRandomEntropySource(new Random(SEED));
 
             TokenPlacementModel.ReplicationFactor rf = new TokenPlacementModel.SimpleReplicationFactor(1);
 
             int maxPartitionSize = 10;
-            modelChecker.init(new HistoryBuilder(seed, maxPartitionSize, 10, schema, rf))
+            modelChecker.init(new HistoryBuilder(SEED, maxPartitionSize, 10, schema, rf))
                         .beforeAll((history) -> {
                             for (int i = 0; i < MAX_PARTITIONS; i++)
                                 history.forPartition(i).ensureClustering(schema.ckGenerator.inflate(rng.nextLong()));
@@ -189,7 +189,7 @@ public class HistoryBuilderIntegrationTest extends IntegrationTestBase
 
                             return true;
                         })
-                        .run(STEPS_PER_ITERATION, seed);
+                        .run(STEPS_PER_ITERATION, SEED);
         }
     }
 }
--- a/test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderIntegrationTest.java
+++ b/test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderIntegrationTest.java
@@ -19,46 +19,31 @@
 package org.apache.cassandra.fuzz.harry.integration.dsl;
 
 import java.util.Random;
-import java.util.function.LongSupplier;
 import java.util.function.Supplier;
 
 import org.junit.Test;
 
-import org.apache.cassandra.fuzz.harry.integration.model.ModelTestBase;
+import org.apache.cassandra.fuzz.harry.integration.model.IntegrationTestBase;
 import org.apache.cassandra.harry.checker.ModelChecker;
-import org.apache.cassandra.harry.core.Configuration;
-import org.apache.cassandra.harry.core.Run;
 import org.apache.cassandra.harry.ddl.SchemaGenerators;
 import org.apache.cassandra.harry.ddl.SchemaSpec;
 import org.apache.cassandra.harry.dsl.BatchVisitBuilder;
 import org.apache.cassandra.harry.dsl.HistoryBuilder;
 import org.apache.cassandra.harry.dsl.SingleOperationBuilder;
-import org.apache.cassandra.harry.dsl.ValueDescriptorIndexGenerator;
 import org.apache.cassandra.harry.gen.rng.JdkRandomEntropySource;
 import org.apache.cassandra.harry.model.Model;
 import org.apache.cassandra.harry.operations.Query;
 import org.apache.cassandra.harry.sut.SystemUnderTest;
 import org.apache.cassandra.harry.sut.TokenPlacementModel;
+import org.apache.cassandra.harry.tracker.DataTracker;
+import org.apache.cassandra.harry.tracker.DefaultDataTracker;
 import org.apache.cassandra.harry.visitors.ReplayingVisitor;
 
-public class HistoryBuilderIntegrationTest extends ModelTestBase
+public class HistoryBuilderIntegrationTest extends IntegrationTestBase
 {
     private final long seed = 1L;
     private final int STEPS_PER_ITERATION = 1_000;
-    private final int MAX_PARTITIONS = 100;
-
-    @Override
-    protected Configuration.ModelConfiguration modelConfiguration()
-    {
-        return new Configuration.QuiescentCheckerConfig();
-    }
-
-    public Configuration.ConfigurationBuilder configuration(long seed, SchemaSpec schema)
-    {
-        return super.configuration(seed, schema)
-                    .setPartitionDescriptorSelector(new Configuration.DefaultPDSelectorConfiguration(2, 2))
-                    .setClusteringDescriptorSelector((builder) -> builder.setOperationsPerLtsDistribution(new Configuration.ConstantDistributionConfig(100_000)));
-    }
+    private final int MAX_PARTITIONS = 50;
 
     @Test
     public void simpleDSLTest() throws Throwable
@@ -67,22 +52,12 @@ public class HistoryBuilderIntegrationTest extends ModelTestBase
         for (int i = 0; i < SchemaGenerators.DEFAULT_RUNS; i++)
         {
             SchemaSpec schema = supplier.get();
-            Configuration config = configuration(i, schema).build();
-
-            Run run = config.createRun();
+            DataTracker tracker = new DefaultDataTracker();
             beforeEach();
-            run.sut.schemaChange(schema.compile().cql());
+            sut.schemaChange(schema.compile().cql());
 
             ModelChecker<SingleOperationBuilder> modelChecker = new ModelChecker<>();
-            JdkRandomEntropySource entropySource = new JdkRandomEntropySource(new Random(seed));
-
-            LongSupplier[] valueGenerators = new LongSupplier[run.schemaSpec.regularColumns.size()];
-            for (int j = 0; j < valueGenerators.length; j++)
-            {
-                valueGenerators[j] = new ValueDescriptorIndexGenerator(run.schemaSpec.regularColumns.get(j),
-                                                                       run.rng)
-                                     .toSupplier(entropySource.derive(), 20, 0.2f);
-            }
+            JdkRandomEntropySource rng = new JdkRandomEntropySource(new Random(seed));
 
             TokenPlacementModel.ReplicationFactor rf = new TokenPlacementModel.SimpleReplicationFactor(1);
 
@@ -92,21 +67,21 @@ public class HistoryBuilderIntegrationTest extends ModelTestBase
                             return history.insert();
                         })
                         .step((history) -> {
-                            return history.insert(entropySource.nextInt(maxPartitionSize));
+                            return history.insert(rng.nextInt(maxPartitionSize));
                         })
                         .step((history) -> {
-                            int row = entropySource.nextInt(maxPartitionSize);
-                            long[] vds = new long[valueGenerators.length];
-                            for (int j = 0; j < valueGenerators.length; j++)
-                                vds[j] = valueGenerators[j].getAsLong();
+                            int row = rng.nextInt(maxPartitionSize);
+                            long[] vIdxs = new long[schema.regularColumns.size()];
+                            for (int j = 0; j < schema.regularColumns.size(); j++)
+                                vIdxs[j] = rng.nextInt(20);
 
-                            return history.insert(row, vds);
+                            return history.insert(row, vIdxs);
                         })
                         .step((history) -> {
                             return history.deleteRow();
                         })
                         .step((history) -> {
-                            return history.deleteRow(entropySource.nextInt(maxPartitionSize));
+                            return history.deleteRow(rng.nextInt(maxPartitionSize));
                         })
                         .step(SingleOperationBuilder::deletePartition)
                         .step(SingleOperationBuilder::deleteColumns)
@@ -115,10 +90,10 @@ public class HistoryBuilderIntegrationTest extends ModelTestBase
                             return history.deleteRowRange();
                         })
                         .step((history) -> {
-                            return history.deleteRowRange(entropySource.nextInt(maxPartitionSize),
-                                                          entropySource.nextInt(maxPartitionSize),
-                                                          entropySource.nextBoolean(),
-                                                          entropySource.nextBoolean());
+                            return history.deleteRowRange(rng.nextInt(maxPartitionSize),
+                                                          rng.nextInt(maxPartitionSize),
+                                                          rng.nextBoolean(),
+                                                          rng.nextBoolean());
                         })
                         .step((history) -> history instanceof HistoryBuilder,
                               (history) -> ((HistoryBuilder) history).beginBatch())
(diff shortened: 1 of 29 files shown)


test/harry/main/org/apache/cassandra/harry/operations/Relation.java
```diff
@@ -97,7 +97,7 @@ public static void addEqRelation(long[] key, List<ColumnSpec<?>> columnSpecs, Li
 
     public static void addRelation(long[] key, List<ColumnSpec<?>> columnSpecs, List<Relation> relations, RelationKind kind)
     {
-        assert key.length == columnSpecs.size() :
+        assert key.length == columnSpecs.size() || key.length > 4 :
```
maedhroz: Do we want to keep the hard-coded "4" here?

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
```diff
@@ -0,0 +1,376 @@
...
+
+                visitor.replayAll();
+                history.validateAll(tracker, sut);
+            }
+        }
+    }
+
+
+    @Test
+    @SuppressWarnings("rawtypes")
```
maedhroz: You could also just throw `<?>` on all the problematic `DataType` and `ColumnSpec` things.

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
```diff
@@ -0,0 +1,376 @@
...
+    public void regularAndStaticOverrideTest() throws Throwable
```
maedhroz:
```suggestion
    public void regularAndStaticOverrideTest()
```

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
```diff
@@ -0,0 +1,376 @@
...
+                    Assert.assertEquals("Should have mutated exactly one CK", found, 1);
+                }
+                history.validateAll(tracker, sut);
+            }
+            Assert.assertEquals(10, foundAt.size());
+        }
+    }
+
+    @Test
+    public void ckOverrideManySortingTest() throws Throwable
```
maedhroz:
```suggestion
    public void ckOverrideManySortingTest()
```

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
```diff
@@ -0,0 +1,376 @@
...
+            if (Arrays.equals(override, Arrays.copyOfRange(row, 4, 9)))
+                found++;
+        }
+        Assert.assertEquals("Should have mutated exactly one CK", found, 1);
+
+        history.validateAll(tracker, sut);
+    }
+
+    @Test
+    public void ckOverrideSortingTest() throws Throwable
```
maedhroz:
```suggestion
    public void ckOverrideSortingTest()
```

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
```diff
@@ -0,0 +1,376 @@
...
+import org.apache.cassandra.harry.tracker.DataTracker;
+import org.apache.cassandra.harry.tracker.DefaultDataTracker;
+import org.apache.cassandra.harry.visitors.ReplayingVisitor;
+
+public class HistoryBuilderOverridesTest extends IntegrationTestBase
+{
+    private final long seed = 1L;
+
+    @Test
+    public void simpleCkOverrideTest() throws Throwable
```
maedhroz:
```suggestion
    public void simpleCkOverrideTest()
```

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
```diff
@@ -0,0 +1,376 @@
...
+        sut.schemaChange(schema.compile().cql());
+
+        JdkRandomEntropySource entropySource = new JdkRandomEntropySource(new Random(seed));
+
+        LongSupplier[] valueGenerators = new LongSupplier[schema.regularColumns.size()];
+        for (int j = 0; j < valueGenerators.length; j++)
+        {
+            valueGenerators[j] = new ValueDescriptorIndexGenerator(schema.regularColumns.get(j), seed)
+                                 .toSupplier(entropySource.derive(), 20, 0.2f);
+        }
```
maedhroz: If I comment out lines 86-93, the test still seems to pass. Dead code from a previous iteration of the test?

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderIntegrationTest.java
```diff
@@ -19,46 +19,31 @@
...
-        return new Configuration.QuiescentCheckerConfig();
-    }
-
-    public Configuration.ConfigurationBuilder configuration(long seed, SchemaSpec schema)
-    {
-        return super.configuration(seed, schema)
-                    .setPartitionDescriptorSelector(new Configuration.DefaultPDSelectorConfiguration(2, 2))
-                    .setClusteringDescriptorSelector((builder) -> builder.setOperationsPerLtsDistribution(new Configuration.ConstantDistributionConfig(100_000)));
-    }
+    private final int MAX_PARTITIONS = 50;
```
maedhroz: nit: I guess `MAX_PARTITIONS` and `STEPS_PER_ITERATION` could be static just to avoid annoying IDE warnings, but not that important...

test/harry/main/org/apache/cassandra/harry/dsl/OverridingCkGenerator.java
```diff
@@ -0,0 +1,187 @@
...
+        return delegate.minValue(idx);
+    }
+
+    @Override
+    public long maxValue(int idx)
+    {
+        return delegate.maxValue(idx);
+    }
+
+    public static class KeyPartOverride<T> implements Bijections.Bijection<T>
```
maedhroz: This shares some code w/ `OverridingBijection`, but I'll let you decide whether there's enough in common to factor something out. Maybe `KeyPartOverride` could be an `OverridingBijection` that just overrides `override()` and `deflate()`...

test/harry/main/org/apache/cassandra/harry/dsl/PartitionVisitStateImpl.java
```diff
@@ -0,0 +1,114 @@
...
+
+
+    /**
+     * Ensures that exactly one of the clustering keys will have given values.
+     */
+    @Override
+    public void ensureClustering(Object[] overrides)
+    {
+        long cd = getCdForOverride(overrides);
+        ckGenerator.override(cd, overrides);
```
maedhroz, line 51: Aside: I'm also kind of wondering now if `OverridingBijection` being an interface would have some organizational usefulness. I guess it only matters if `override()` is going to be used in contexts where we need indirection.

test/harry/main/org/apache/cassandra/harry/dsl/PartitionVisitStateImpl.java
```diff
@@ -0,0 +1,114 @@
...
+            ckGenerator.override(possibleCds[i], overrides[i]);
+    }
+
+    @Override
+    public long pd()
+    {
+        return pd;
+    }
+
+    long getCdForOverride(Object[] ck)
```
maedhroz:
```suggestion
    long findCdForOverride(Object[] ck)
```
Not a big deal, but feels appropriate given the binary search?

test/harry/main/org/apache/cassandra/harry/dsl/PartitionVisitStateImpl.java
```diff
@@ -0,0 +1,114 @@
...
+            else if (cmp > 0)
+                high = mid - 1;
+            else
+                throw new IllegalStateException("This value is already present");
+        }
+
+        return possibleCds[Math.min(possibleCds.length - 1, low)];
+    }
+
+    private int compareCds(Object[] v1, long cd2) {
```
maedhroz: nit: `{` newline

test/harry/main/org/apache/cassandra/harry/dsl/PartitionVisitStateImpl.java
```diff
@@ -0,0 +1,114 @@
...
+        Object[] v2 = schema.ckGenerator.inflate(cd2);
+        return compareCds(v1, v2);
+    }
+
+    private int compareCds(Object[] v1, Object[] v2) {
```
maedhroz: nit: `{` newline

test/harry/main/org/apache/cassandra/harry/dsl/PartitionVisitStateImpl.java
```diff
@@ -0,0 +1,114 @@
...
+        assert v1.length == v2.length : String.format("Values should be of same length: %d != %d\n" +
+                                                      "%s\n" +
+                                                      "%s",
+                                                      v1.length, v2.length, Arrays.toString(v1), Arrays.toString(v2));
```
maedhroz, line 100: nit: I guess if this was a hot path, I'd be worried about the string creation.

test/harry/main/org/apache/cassandra/harry/dsl/SingleOperationBuilder.java
```diff
@@ -36,8 +35,8 @@ public interface SingleOperationBuilder
      * Insert _specific values_ into _specific_ row. Rows are ordered by clustering key and
      * numbered from 0 to maxRows
      */
-    SingleOperationBuilder insert(int rowIdx, long[] vds);
-    SingleOperationBuilder insert(int rowIdx, long[] vds, long[] sds);
+    SingleOperationBuilder insert(int rowIdx, int[] vds);
+    SingleOperationBuilder insert(int rowIdx, int[] vds, int[] sds);
```
maedhroz: Should we be clear here in term of naming that these are indexes and no longer actual ads? `SingleOperationVisitBuilder` is a little clearer, of course...

test/harry/main/org/apache/cassandra/harry/dsl/ValueDescriptorIndexGenerator.java
```diff
@@ -37,10 +37,18 @@
...
 public class ValueDescriptorIndexGenerator implements Surjections.Surjection<Long>
 {
+    public static int UNSET = Integer.MIN_VALUE;
+
     private final OpSelectors.PureRng rng;
     private final long columnHash;
     private final long mask;
 
+    public ValueDescriptorIndexGenerator(ColumnSpec<?> columnSpec,
+                                         long seed)
```
maedhroz: nit: unnecessarily breaking `seed` to a second line?

test/harry/main/org/apache/cassandra/harry/dsl/ValueHelper.java
```diff
@@ -0,0 +1,77 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.cassandra.harry.ddl.ColumnSpec;
+import org.apache.cassandra.harry.ddl.SchemaSpec;
+import org.apache.cassandra.harry.model.OpSelectors;
+
+/**
+ * A class responsible for value
```
maedhroz: I'd probably just get rid of the JavaDoc here, given it doesn't really say much more than the name of the class does. Do we anticipate any other implementations of `ValueOverrides` in the near future? I guess we could just get rid of the interface and call this `ValueOverrides` if not...

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderIntegrationTest.java
```diff
@@ -129,16 +104,88 @@ public void simpleDSLTest() throws Throwable
...
+            ModelChecker<HistoryBuilder> modelChecker = new ModelChecker<>();
+            JdkRandomEntropySource rng = new JdkRandomEntropySource(new Random(seed));
+
+            TokenPlacementModel.ReplicationFactor rf = new TokenPlacementModel.SimpleReplicationFactor(1);
+
+            int maxPartitionSize = 10;
+            modelChecker.init(new HistoryBuilder(seed, maxPartitionSize, 10, schema, rf))
+                        .beforeAll((history) -> {
+                            for (int i = 0; i < MAX_PARTITIONS; i++)
+                                history.forPartition(i).ensureClustering(schema.ckGenerator.inflate(rng.nextLong()));
```
maedhroz, line 144: So this is the crux of the test? i.e. Make sure every partition has the provided clustering key inflated from the underlying `KeyPartOverride`, then do normal validation?

test/harry/main/org/apache/cassandra/harry/dsl/HistoryBuilder.java
```diff
@@ -29,15 +28,18 @@
 import java.util.NavigableSet;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.function.Consumer;
 import java.util.function.LongSupplier;
 
 import org.apache.cassandra.harry.clock.ApproximateClock;
 import org.apache.cassandra.harry.core.Configuration;
 import org.apache.cassandra.harry.core.MetricReporter;
+import org.apache.cassandra.harry.ddl.ColumnSpec;
```
maedhroz: unused `ColumnSpec` import

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
```diff
@@ -0,0 +1,376 @@
...
+                       .insert(i);
+            }
+
+            for (int i = 0; i < partitionSize; i++)
+            {
+                history.visitPartition(1)
+                       .deleteRow(i);
+
+                visitor.replayAll();
+                history.validateAll(tracker, sut);
```
maedhroz: nit: Can we validate after the loop, or is there something we'd miss if we didn't on every iteration?

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
```diff
@@ -0,0 +1,376 @@
...
+            EntropySource rng = new JdkRandomEntropySource(seed);
+
+            Map<String, Set<Object>> perColumnOverrides = new HashMap<>();
+            for (ColumnSpec column : schema.regularColumns)
+            {
+                perColumnOverrides.put(column.name, new HashSet<>());
+                for (int i = 0; i < partitionSize; i++)
+                {
+                    Object override = reGenerators.get(column.name).inflate(rng.next());
+                    history.valueOverrides().override(column.name, i, override);
```
maedhroz: So we're overriding the value at every row of a 100-row partition?

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
```diff
@@ -0,0 +1,376 @@
...
+                    results.put(column.name, new HashSet<>());
+                for (ColumnSpec column : schema.staticColumns)
+                    results.put(column.name, new HashSet<>());
+
+                for (int i = 0; i < partitionSize; i++)
+                {
+                    history.visitPartition(pdIdx)
+                           .insert(i,
+                                   new int[]{ rng.nextInt(100), rng.nextInt(100) },
+                                   new int[]{ rng.nextInt(100), rng.nextInt(100) });
```
maedhroz: I think I made a comment about how we name these in the `SingleOperationBuilder` interface, but to confirm, we're basically asking the history builder to insert our overrides at the random row indexes?

maedhroz (changes requested):
I've left a mix of questions and nits. Overall, this looks reasonable to me, though. I guess the fun part will coming up with problematic values to override with :)

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
```diff
@@ -0,0 +1,376 @@
...
+            ReplayingVisitor visitor = history.visitor(tracker, sut, SystemUnderTest.ConsistencyLevel.ALL);
+            EntropySource rng = new JdkRandomEntropySource(seed);
+
+            Map<String, Set<Object>> perColumnOverrides = new HashMap<>();
+            for (ColumnSpec column : schema.regularColumns)
+            {
+                perColumnOverrides.put(column.name, new HashSet<>());
+                for (int i = 0; i < partitionSize; i++)
+                {
+                    Object override = reGenerators.get(column.name).inflate(rng.next());
```
maedhroz: Say I was testing SAI here and I wanted to try something like an empty string...

```
Object override = column.name.equals("regular1") ? "" : reGenerators.get(column.name).inflate(rng.next());
```

This doesn't seem to quite work when I try it...

test/harry/main/org/apache/cassandra/harry/dsl/SingleOperationBuilder.java
maedhroz: Well, the inserts happen and the SUT matches, but `validateAll()` doesn't like it :)

maedhroz: I see the right `ValueHelper#overrides` bijection getting the mapping, but then nothing accesses the overrides?

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderIntegrationTest.java
ifesdjeen: Exactly, we simply hardcode the clusterings for partition, and then let the test run.

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
ifesdjeen: yes, sorry: this is just a remainder from the previous test

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
ifesdjeen: I thought it might be useful to validate every iteration since some problems may get shadowed by subsequent writes. it should  be relatively cheap still

test/harry/main/org/apache/cassandra/harry/operations/Relation.java
ifesdjeen: Good point. Added a `MAX_UNIQUE_PREFIX_COLUMNS` static field.

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderIntegrationTest.java
ifesdjeen: Did that for some of the fields I have found.

test/harry/main/org/apache/cassandra/harry/dsl/OverridingCkGenerator.java
ifesdjeen: Yeah, good point. Probably the overhead of `valueToDescriptor` is not worth custom class.

test/harry/main/org/apache/cassandra/harry/dsl/PartitionVisitStateImpl.java
ifesdjeen: Since it's likely to be used always in conjuntion with the existing bijection (i.e. overriding it), I think it might be fine to keep it as a class. We can always refactor it if we see more uses.

test/harry/main/org/apache/cassandra/harry/dsl/PartitionVisitStateImpl.java
ifesdjeen: reformatted the whole class

test/harry/main/org/apache/cassandra/harry/dsl/PartitionVisitStateImpl.java
ifesdjeen: To my best knowledge the `:` part is only evaluated on exception. I have removed `+` since they seem to be redundant though.

test/harry/main/org/apache/cassandra/harry/dsl/SingleOperationBuilder.java
ifesdjeen: Good point, made them

```
    SingleOperationBuilder insert(int rowIdx, long[] valueIdxs);
    SingleOperationBuilder insert(int rowIdx, long[] valueIdxs, long[] sValueIdxs);
```

test/harry/main/org/apache/cassandra/harry/dsl/ValueHelper.java
ifesdjeen: I think I was aspiring to write more :)

test/harry/main/org/apache/cassandra/harry/dsl/SingleOperationBuilder.java
ifesdjeen: > Well, the inserts happen and the SUT matches, but validateAll() doesn't like it :)

Could you elaborate a bit? `history.valueOverrides().override(column.name, i, override);` is just used as a convenience method to quickly access override generator.

But we do a swap of `Schema` in `HistoryBuilder` to make sure all child generators are getting the correct (overriden) generators:

```
        this.schema = schema.withCkGenerator(this.ckGenerator, this.ckGenerator.columns)
                            .withColumns(valueHelper.regularColumns, valueHelper.staticColumns);
```

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
ifesdjeen: You are right to note that using `partitionSize` here is somewhat arbitrary. In other words, there's nothing prevents us to use some other `idx` (i.e. larger than 100), and get an actually random value. I've just overriden 100 values per column, but then we're going in the loop and are simply picking one of these 100 values. In the end, there will be only 1 unique static value per partition and ~60 unique values (assuming uniform distribution, so can be more or less) per value column.

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
ifesdjeen: Right: we are simply saying "insert a value number X", where X is generated randomly. There's nothing preventing us from putting `210` as an index, it's just for the purpose of our test we ensure that _all_ the values were overriden. You can still generate arbitrary random values from the indexes you have not overriden.

test/distributed/org/apache/cassandra/fuzz/harry/integration/dsl/HistoryBuilderOverridesTest.java
ifesdjeen: So you simply would do 

```
history.valueOverrides().override("regular1", 12345, "");
```

And then somewhere in your test you do

```
                    history.visitPartition(pdIdx)
                           .insert(i, new long[]{ 12345, 12345 });
```

ifesdjeen: Added some docs!

maedhroz (approved):
Thanks for the new docs!

PR #3068, opened 2024-01-26
Fix timeouts caused by infrequently used Meters

aweisberg:


src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
```diff
@@ -41,15 +57,20 @@
  */
 public class CassandraMetricsRegistry extends MetricRegistry
 {
-    public static final CassandraMetricsRegistry Metrics = new CassandraMetricsRegistry();
+    private static final Logger logger = LoggerFactory.getLogger(CassandraMetricsRegistry.class);
+
+    public static final CassandraMetricsRegistry Metrics = new CassandraMetricsRegistry(Ints.checkedCast(TimeUnit.MICROSECONDS.toNanos(1)));
```
Mmuzaf: It's noted in the JIRA issue itself that it's enough to read the metrics once a day, but `MICROSECONDS` is here. Is it correct?

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
```diff
@@ -41,15 +57,20 @@
...
     private final Map<String, ThreadPoolMetrics> threadPoolMetrics = new ConcurrentHashMap<>();
 
     private final MBeanWrapper mBeanServer = MBeanWrapper.instance;
     public final static TimeUnit DEFAULT_TIMER_UNIT = TimeUnit.MICROSECONDS;
 
-    private CassandraMetricsRegistry()
+    final ScheduledFuture<?> periodicMeterTicker;
+
+    CassandraMetricsRegistry(int tickMetersPeriodMicros)
```
Mmuzaf: I think we don't need to pass the argument to the constructor and can use a constant variable instead. We can also add a javadoc to this constant so that the default value can be indexed by search engines once the javadocs are published, 

e.g. 
`/* Schedule a task to read all the meters once a day so it isn’t done in the request path and we have a more incremental amount to process at a time. Default value {@code 1} day. */ `

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
```diff
@@ -250,6 +271,43 @@ private void removeAlias(MetricName name)
...
+     * that needs to mark the meter to time out. Once a day read every meter to force them to run Meter.tickIfNecessary
+     * so we only ever run at most one day worth of tick work per meter in the request path.
+     *
+     * This can be removed if we ever upgrade and switch the default MovingAverage from EWMA to SlidingWindowTimeAverages
+     */
+    private void tickMeters()
+    {
+        List<Throwable> failures = new ArrayList<>();
+        int droppedFailures = 0;
+        for (Meter meter : getMeters().values())
```
Mmuzaf: I think it is more efficient to filter the metrics by type instead of calling the `getMeters()` method, which creates a new map each time it is called - it is expensive for a large set of metrics.

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
```diff
@@ -250,6 +271,43 @@ private void removeAlias(MetricName name)
...
+            }
+        }
+        if (!failures.isEmpty())
+        {
+            Throwable failure = null;
+            for (Throwable t : failures)
+                failure = Throwables.merge(failure, t);
+            // To avoid the scheduled task being cancelled don't leak exceptions
+            // Runs only once a day so noise is not an issue
+            logger.error(String.format("Had error(s) attempting to tick meter. Dropped %d exceptions.", droppedFailures), failure);
```
Mmuzaf: Should we use the `warn` (or `info`) type instead? I don't think a user can fix this error anyway.

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
aweisberg: Yeah it should be days to micros here.

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
aweisberg: This gets called once a day in a background thread. Assume 100 meters per table and 1000 tables it takes about 26 seconds on my laptop. This is to process a days worth of ticks on every meter as if it were unused.

So maybe `0.00003761574` of daily CPU time on an 8 core server. The time is dominated by actually ticking not iterating the metrics.

I really don't think it's worth the extra lines of code to make iteration faster.

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
aweisberg: It's not something that is ever expected so if it does happen things could be pretty broken not just with this ticking functionality. Since it will only log once a day it seems more appropriate to make it an error so it's more visible and can get fixed.

Warn would be fine, but makes it a little easier for errors to get lost in noise. This is once a day so it wouldn't contribute a lot of noise.

Mmuzaf: I think whatever the exception may be it will be unrelated to Cassandra anyway, so no need for extra visibility. I don't think this nit is critical to the patch, so have it your way :-)

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
aweisberg: This is to make it possible for the unit test to create a metrics registry that runs these ticks frequently. To have it be testable there has to be some extension point, either a method that can be overridden or a parameter to the class to allow for manipulating the task scheduling frequency.

aweisberg:
TY for the review! Pushed an update fixing the once a day ticking. It was also making the tests fail so that's a big help.

test/unit/org/apache/cassandra/metrics/CassandraMetricsRegistryTest.java
```diff
@@ -156,6 +162,32 @@ public void testTimer()
         assertEquals(expectedBucketsWithValues, bucketsWithValues);
     }
 
+    @Test
+    public void testTickMeters() throws InterruptedException
+    {
+        CassandraMetricsRegistry cmr = new CassandraMetricsRegistry(Ints.checkedCast(TimeUnit.SECONDS.toMicros(1)));
```
Mmuzaf:
```suggestion
        CassandraMetricsRegistry cmr = new CassandraMetricsRegistry(TimeUnit.SECONDS.toMicros(1));
```

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
Mmuzaf: This is actually +2 extra lines of code! Anyway, it's not important for the workaround itself, the comment is just based on the assumption that if we can have zero impact, why shouldn't we? So I agree with you that the node can live without it and this is not a place where we should save resources :-)

Mmuzaf (approved):
I've left one nit, +1 to proceed.

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
aweisberg: Which lines are the extra lines?  I would remove 2 extra lines if I could?

Mmuzaf: I would change

```
for (Meter meter : getMeters().values())
{
    ...
}
```

to the following

```
for (Metric metric : getMetrics().values())
{
    if (!Meter.class.isInstance(metric))
        continue;
    ...
   
}
```

aweisberg: Ah, I got what you were saying backwards I thought you meant I had the opportunity to remove 2 lines which I would jump at :-)

Mmuzaf (approved):

PR #3069, opened 2024-01-28
CASSANDRA-19335 Default nodetool tablestats to Human-Readable Output

zaaath:
Default to human-readable in `nodetool tablestats`. See a discussion on the topic in the dev mailing list [dev@cassandra.apache.org](https://lists.apache.org/thread/mlp715kxho5b6f1ql9omlzmmnh4qfby9)

### Before
`nodetool tablestats` doesn't use the human-readable formatting by default (with KiB/MiB/GiB units), instead it outputs data size values as is. In order to switch to the human-readable output, `--human-readable` or `--H` CLI flag is required.

### After
`nodetool tablestats` uses the human-readable format by default. For backward compatibility `--no-human-readable` flag can be used to disable it.

### How to reproduce
```bash
nodetool tablestats
```
and
```bash
nodetool tablestats --no-human-readable
```

Tests
```bash
```

patch by @zaaath; reviewed by ??? for [CASSANDRA-19335](https://issues.apache.org/jira/browse/CASSANDRA-19335)


src/java/org/apache/cassandra/tools/nodetool/TableStats.java
```diff
@@ -36,10 +36,10 @@ public class TableStats extends NodeToolCmd
     @Option(name = "-i", description = "Ignore the list of tables and display the remaining tables")
     private boolean ignore = false;
 
-    @Option(title = "human_readable",
-            name = {"-H", "--human-readable"},
-            description = "Display bytes in human readable form, i.e. KiB, MiB, GiB, TiB")
-    private boolean humanReadable = false;
+    @Option(title = "no_human_readable",
+            name = {"--no-human-readable"},
```
driftx: Can we spare a short flag here too?

zaaath: Sure! How about `-r` (for Raw) or `-nhr` (for No Human Readable)? I'll add `-r` for now.

zaaath: `-r` short flag added (stands for Raw). Is there anything else missing here?

driftx: -r works for me, thanks!

PR #3073, opened 2024-01-30
CASSANDRA-19336 5.0: Ensure that repair doesn't exceed repair_session_space by running repair jobs sequentially

adelapena:


src/java/org/apache/cassandra/repair/Scheduler.java
```diff
@@ -0,0 +1,181 @@
...
+                       ", durations=" + durations.longStream().average().orElse(-1) +
+                       ", score=" + score() +
+                       ", inflight=" + inflight +
+                       ", completed=" + completed +
+                       '}';
+            }
+        }
+    }
+
+    abstract class Task<T> extends AsyncFuture<T> implements Runnable
```
dcapwell: isn't this just `org.apache.cassandra.concurrent.FutureTask`?

src/java/org/apache/cassandra/repair/RepairJob.java
```diff
@@ -189,30 +188,33 @@ public void onFailure(Throwable t)
...
+            {
+                createSyncTasks(paxosRepair, allSnapshotTasks, allEndpoints).addCallback((s, f) -> {
+                    if (f != null)
+                        tryFailure(f);
+                    else
+                        trySuccess(s);
+                });
+            }
+        };
+        session.validationScheduler.schedule(session.getId(), taskExecutor, syncTasks);
```
dcapwell: feel its cleaner to return `Future<T>` from `schedule`

src/java/org/apache/cassandra/repair/Scheduler.java
```diff
@@ -0,0 +1,181 @@
...
+        private final Map<TimeUUID, Group> groups = new HashMap<>();
+
+        LimitedConcurrentScheduler(int concurrentValidations, SharedContext ctx)
+        {
+            this.concurrentValidations = concurrentValidations;
+            this.ctx = ctx;
+        }
+
+        @Override
+        public synchronized void schedule(TimeUUID sessionId, Executor taskExecutor, Task<?> tasks)
```
dcapwell: the usage is a single task, so all sessions will have the same `score`... so a simpler data structure could be used... this class makes the most sense when we have way to compare tasks

dcapwell (approved):

src/java/org/apache/cassandra/repair/Scheduler.java
adelapena: This is slightly different due to how the `run` method sets a callback. I have moved the setting of that callback into the `Task` class to encapsulate it better.

src/java/org/apache/cassandra/repair/RepairJob.java
adelapena: Done.

src/java/org/apache/cassandra/repair/Scheduler.java
adelapena: I have changed it to a single queue of task-executor pairs.

PR #3077, opened 2024-01-31
CASSANDRA-19305: Ephemeral reads

belliottsmith:
Introduce faster per-key linearizable reads (suitable for any single key transactional read, or weaker multi-key reads). These guarantee 1 WAN RT, require minimal state, and offer single replica consistent execution. This is achieved by only calculating dependencies for execution, no execution timestamp.

### DIFF
--- a/src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
+++ b/src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
@@ -79,6 +79,10 @@ import org.apache.cassandra.service.accord.txn.TxnWrite;
 import org.apache.cassandra.transport.messages.ResultMessage;
 import org.apache.cassandra.utils.FBUtilities;
 
+import static accord.primitives.Txn.Kind.EphemeralRead;
+import static accord.primitives.Txn.Kind.Read;
+import static org.apache.cassandra.config.Config.NonSerialWriteStrategy.accord;
+import static org.apache.cassandra.config.DatabaseDescriptor.getNonSerialWriteStrategy;
 import static org.apache.cassandra.cql3.statements.RequestValidations.checkFalse;
 import static org.apache.cassandra.cql3.statements.RequestValidations.checkNotNull;
 import static org.apache.cassandra.cql3.statements.RequestValidations.checkTrue;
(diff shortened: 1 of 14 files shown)


src/java/org/apache/cassandra/service/accord/serializers/GetEphmrlReadDepsSerializers.java
```diff
@@ -0,0 +1,85 @@
...
+import accord.primitives.PartialDeps;
+import accord.primitives.PartialRoute;
+import accord.primitives.Seekables;
+import accord.primitives.TxnId;
+import org.apache.cassandra.db.TypeSizes;
+import org.apache.cassandra.io.IVersionedSerializer;
+import org.apache.cassandra.io.util.DataInputPlus;
+import org.apache.cassandra.io.util.DataOutputPlus;
+
+public class GetEphmrlReadDepsSerializers
```
aweisberg, line 34: Typo in class name

aweisberg (approved):
Just noticed one typo

src/java/org/apache/cassandra/service/accord/serializers/GetEphmrlReadDepsSerializers.java
belliottsmith: Actually it was deliberate to shrink the length - largely to avoid any impact to Verb alignment. Open to either shrinking the name in some other way (even renaming from Ephemeral entirely) or some other approach.

src/java/org/apache/cassandra/cql3/statements/TransactionStatement.java
```diff
@@ -321,7 +326,8 @@ public Txn createTxn(ClientState state, QueryOptions options)
             List<TxnNamedRead> reads = createNamedReads(options, state, ImmutableMap.of(), keySet::add);
             Keys txnKeys = toKeys(keySet);
             TxnRead read = createTxnRead(reads, txnKeys, null);
-            return new Txn.InMemory(txnKeys, read, TxnQuery.ALL);
+            Txn.Kind kind = txnKeys.size() == 1 && getNonSerialWriteStrategy() == accord ? EphemeralRead : Read;
```
aweisberg, line 328: Minor, but this is a per key property that wouldn't require absolutely all writes going through Accord. Would be nice to have the performance benefit during migration, but maybe a TODO.

PR #3083, opened 2024-02-02
CASSANDRA-19018 An SAI-specific mechanism to ensure consistency isn't violated for multi-column AND queries at CL > ONE

maedhroz:


src/java/org/apache/cassandra/service/reads/ReplicaFilteringProtection.java
```diff
@@ -208,34 +211,33 @@ public void onMergedPartitionLevelDeletion(DeletionTime mergedDeletion, Deletion
...
-                        {
-                            Row version = versions[i];
-                            if (version == null || (isStatic && version.isEmpty()))
-                            {
-                                isPotentiallyOutdated = true;
+                            if (versions[i] == null || (merged.isStatic() && versions[i].isEmpty()))
+                                fetchFrom[i] = true;
+
+                        // ...and then if nothing is amiss at the row level, check the individual columns:
+                        // TODO: If all the values are null for a column, do we actually care about fetching those again?
```
maedhroz: @adelapena This might be my biggest outstanding question, as it has both performance and correctness implications. At the row level above, we trigger fetch on any `null` row, but that's because if all versions were `null`, we would never inspect the versions in the first place. Here, it seems like we can actually avoid an RFP fetch if the cells for a column are null across ALL versions...

src/java/org/apache/cassandra/service/reads/ReplicaFilteringProtection.java
```diff
@@ -208,34 +211,33 @@ public void onMergedPartitionLevelDeletion(DeletionTime mergedDeletion, Deletion
...
+                        for (ColumnMetadata column : columns)
+                            for (int i = 0; i < versions.length; i++)
+                                //noinspection DataFlowIssue
```
maedhroz: If `versions[I]` was `null`, we would have marked `fetchFrom[I]`, so I don't think a NPE is possible.

src/java/org/apache/cassandra/service/reads/DataResolver.java
```diff
@@ -226,6 +231,7 @@ private PartitionIterator resolveWithReadRepair(ResolveContext context,
 
     private PartitionIterator resolveWithReplicaFilteringProtection(E replicas, RepairedDataTracker repairedDataTracker)
     {
+        // TODO: Update comments once this has made it through review...
```
maedhroz: Will provide more commentary and adjustments to the existing comments once we're through review of the actual changes...

maedhroz:
Committed as https://github.com/apache/cassandra/commit/58f0e2146b3d7f0bd09c4dbd9d1da5e43f37ed73

PR #3085, opened 2024-02-05
CASSANDRA-19366 trunk - Expose auth mode in system_views.clients, clientstats, metrics

tolbertam:
Adds 'mode' and 'metadata' fields to AuthenticatedUser to add context about how the user was authenticated and updates system_views.clients, nodetool client stats to include this information.

Also adds new metrics to ClientMetrics to help operators identify which authentication modes are being used.

for CASSANDRA-19366

### DIFF
--- a/doc/modules/cassandra/pages/managing/operating/metrics.adoc
+++ b/doc/modules/cassandra/pages/managing/operating/metrics.adoc
@@ -987,7 +987,7 @@ custom mode by implementing `IAuthenticator.getSupportedAuthenticationModes()`.
 [cols=",,",options="header",]
 |===
 |Name |Type |Description
-|AuthenticatedNativeClients |Gauge<Int> |Number of clients connected and
+|ConnectedNativeClients |Gauge<Int> |Number of clients connected and
 authenticated to this nodes native protocol server using this mode
 
 |AuthFailure |Meter | Rate of failed authentications using this mode
--- a/doc/modules/cassandra/pages/managing/operating/metrics.adoc
+++ b/doc/modules/cassandra/pages/managing/operating/metrics.adoc
@@ -977,8 +977,8 @@ An authentication mode is a supported method of authentication. The following
 authentication modes exist for the given supported `IAuthenticators`:
 
 * `PasswordAuthenticator` - Password
-* `MutualTlsAuthenticator` - Mtls
-* `MutualTlsWithPasswordFallbackAuthenticator` - Mtls, Password
+* `MutualTlsAuthenticator` - MutualTls
+* `MutualTlsWithPasswordFallbackAuthenticator` - MutualTls, Password
 * `AllowAllAuthenticator` - Unauthenticated
 
 A custom implementation of `IAuthenticator` may expose metrics on their own
--- a/doc/modules/cassandra/pages/managing/operating/metrics.adoc
+++ b/doc/modules/cassandra/pages/managing/operating/metrics.adoc
@@ -913,14 +913,87 @@ Reported name format:
 [cols=",,",options="header",]
 |===
 |Name |Type |Description
-|connectedNativeClients |Gauge<Integer> |Number of clients connected to
+|AuthFailure |Meter | Rate of failed authentications
+
+|AuthSuccess |Meter | Rate of successful authentications
+
+|ClientsByProtocolVersion |Gauge<List<Map<String, Integer>> | List of
+all connections' protocol version and ip address
+
+|ConnectedNativeClients |Gauge<Integer> |Number of clients connected to
 this nodes native protocol server
 
-|connections |Gauge<List<Map<String, String>> |List of all connections
+|ConnectedNativeClientsByUser |Gauge<Map<String, Integer>> |Number of
+connnective native clients by username
+
+|Connections |Gauge<List<Map<String, String>> |List of all connections
 and their state information
 
-|connectedNativeClientsByUser |Gauge<Map<String, Int> |Number of
-connnective native clients by username
+|PausedConnections|Gauge<Integer>|Number of connections currently
+paused by rate limiter
+
+|ProtocolException|Meter|Rate of requests resulting in a protocol
+exception
+
+|UnknownException|Meter|Rate of requests resulting in an unknown
+exception
+
+|RequestDiscarded|Meter|Rate of requests discarded by rate limiter
+
+|RequestDispatched|Meter|Rate of requests dispatched (not discarded)
+
+|RequestsSizeByIpDistribution|Histogram|Histogram of distribution of
+requests coming from unique IPs
+
+|===
+
+== Client Encryption Metrics
+
+Metrics specific to Client encryption
+
+*Metric Name*::
+`org.apache.cassandra.metrics.Client.<Encrypted|Unencrypted>.<MetricName>`
+*JMX MBean*::
+`org.apache.cassandra.metrics:type=Client scope=<Encrypted|Unencrypted> name=<MetricName>`
+
+[cols=",,",options="header",]
+|===
+|Name |Type |Description
+|ConnectedNativeClients |Gauge<Integer> |Number of clients connected in
+this way to this nodes native protocol server
+
+|===
+
+== Client Authentication Mode-Specific Metrics
+
+Metrics specific to connectivity for a given Authentication 'mode'.
+
+*Metric Name*::
+`org.apache.cassandra.metrics.Client.<Mode>.<MetricName>`
+*JMX MBean*::
+`org.apache.cassandra.metrics:type=Client scope=<Mode> name=<MetricName>`
+
+An authentication mode is a supported method of authentication. The following
+authentication modes exist for the given supported `IAuthenticators`:
+
+* `PasswordAuthenticator` - password
+* `MutualTlsAuthenticator` - mtls
+* `MutualTlsWithPasswordFallbackAuthenticator` - mtls, password
+* `AllowAllAuthenticator` - anonymous
+
+A custom implementation of `IAuthenticator` may expose metrics on their own
+custom mode by implementing `IAuthenticator.getSupportedAuthenticationModes()`.
+
+[cols=",,",options="header",]
+|===
+|Name |Type |Description
+|AuthenticatedNativeClients |Gauge<Int> |Number of clients connected and
+authenticated to this nodes native protocol server using this mode
+
+|AuthFailure |Meter | Rate of failed authentications using this mode
+
+|AuthSuccess |Meter | Rate of successful authentications using this mode
+
 |===
 
 == Batch Metrics
--- a/doc/modules/cassandra/pages/managing/operating/metrics.adoc
+++ b/doc/modules/cassandra/pages/managing/operating/metrics.adoc
@@ -976,10 +976,10 @@ Metrics specific to connectivity for a given Authentication 'mode'.
 An authentication mode is a supported method of authentication. The following
 authentication modes exist for the given supported `IAuthenticators`:
 
-* `PasswordAuthenticator` - password
-* `MutualTlsAuthenticator` - mtls
-* `MutualTlsWithPasswordFallbackAuthenticator` - mtls, password
-* `AllowAllAuthenticator` - anonymous
+* `PasswordAuthenticator` - Password
+* `MutualTlsAuthenticator` - Mtls
+* `MutualTlsWithPasswordFallbackAuthenticator` - Mtls, Password
+* `AllowAllAuthenticator` - Unauthenticated
 
 A custom implementation of `IAuthenticator` may expose metrics on their own
 custom mode by implementing `IAuthenticator.getSupportedAuthenticationModes()`.
--- a/src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
+++ b/src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
@@ -22,12 +22,16 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 
+import com.google.common.collect.Sets;
+
 import org.apache.cassandra.exceptions.AuthenticationException;
 import org.apache.cassandra.exceptions.ConfigurationException;
 
 public class AllowAllAuthenticator implements IAuthenticator
 {
     private static final SaslNegotiator AUTHENTICATOR_INSTANCE = new Negotiator();
+
+    private static final Set<String> MODES = Sets.newHashSet(AuthenticatedUser.ANONYMOUS_USERNAME);
 
     public boolean requireAuthentication()
     {
@@ -50,6 +54,12 @@ public class AllowAllAuthenticator implements IAuthenticator
     public SaslNegotiator newSaslNegotiator(InetAddress clientAddress)
     {
         return AUTHENTICATOR_INSTANCE;
+    }
+
+    @Override
+    public Set<String> getSupportedAuthenticationModes()
+    {
+        return MODES;
     }
 
     public AuthenticatedUser legacyAuthenticate(Map<String, String> credentialsData)
--- a/src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
+++ b/src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
@@ -31,7 +31,7 @@ public class AllowAllAuthenticator implements IAuthenticator
 {
     private static final SaslNegotiator AUTHENTICATOR_INSTANCE = new Negotiator();
 
-    private static final Set<AuthenticationMode> AUTHENTICATION_MODES = Sets.newHashSet(AuthenticationMode.UNAUTHENTICATED);
+    private static final Set<AuthenticationMode> AUTHENTICATION_MODES = Collections.singleton(AuthenticationMode.UNAUTHENTICATED);
 
     public boolean requireAuthentication()
     {
--- a/src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
+++ b/src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
@@ -31,7 +31,7 @@ public class AllowAllAuthenticator implements IAuthenticator
 {
     private static final SaslNegotiator AUTHENTICATOR_INSTANCE = new Negotiator();
 
-    private static final Set<String> MODES = Sets.newHashSet(AuthenticationMode.UNAUTHENTICATED.getDisplayName());
+    private static final Set<AuthenticationMode> AUTHENTICATION_MODES = Sets.newHashSet(AuthenticationMode.UNAUTHENTICATED);
 
     public boolean requireAuthentication()
     {
@@ -57,9 +57,9 @@ public class AllowAllAuthenticator implements IAuthenticator
     }
 
     @Override
-    public Set<String> getSupportedAuthenticationModes()
+    public Set<AuthenticationMode> getSupportedAuthenticationModes()
     {
-        return MODES;
+        return AUTHENTICATION_MODES;
     }
 
     public AuthenticatedUser legacyAuthenticate(Map<String, String> credentialsData)
--- a/src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
+++ b/src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
@@ -31,7 +31,7 @@ public class AllowAllAuthenticator implements IAuthenticator
 {
     private static final SaslNegotiator AUTHENTICATOR_INSTANCE = new Negotiator();
 
-    private static final Set<String> MODES = Sets.newHashSet(AuthenticatedUser.ANONYMOUS_USERNAME);
+    private static final Set<String> MODES = Sets.newHashSet(AuthenticationMode.UNAUTHENTICATED.getDisplayName());
 
     public boolean requireAuthentication()
     {
--- a/src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
+++ b/src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
@@ -22,8 +22,6 @@ import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
 
-import com.google.common.collect.Sets;
-
 import org.apache.cassandra.exceptions.AuthenticationException;
 import org.apache.cassandra.exceptions.ConfigurationException;
 
--- a/src/java/org/apache/cassandra/auth/AuthenticatedUser.java
+++ b/src/java/org/apache/cassandra/auth/AuthenticatedUser.java
@@ -18,6 +18,8 @@
 package org.apache.cassandra.auth;
 
 import java.net.InetSocketAddress;
+import java.util.Collections;
+import java.util.Map;
 import java.util.Set;
 import com.google.common.base.Objects;
 
@@ -33,10 +35,10 @@ import org.apache.cassandra.dht.Datacenters;
 public class AuthenticatedUser
 {
     public static final String SYSTEM_USERNAME = "system";
-    public static final AuthenticatedUser SYSTEM_USER = new AuthenticatedUser(SYSTEM_USERNAME);
+    public static final AuthenticatedUser SYSTEM_USER = new AuthenticatedUser(SYSTEM_USERNAME, SYSTEM_USERNAME);
 
     public static final String ANONYMOUS_USERNAME = "anonymous";
-    public static final AuthenticatedUser ANONYMOUS_USER = new AuthenticatedUser(ANONYMOUS_USERNAME);
+    public static final AuthenticatedUser ANONYMOUS_USER = new AuthenticatedUser(ANONYMOUS_USERNAME, ANONYMOUS_USERNAME);
 
     // User-level permissions cache.
     public static final PermissionsCache permissionsCache = new PermissionsCache(DatabaseDescriptor.getAuthorizer());
@@ -55,13 +57,34 @@ public class AuthenticatedUser
 
     private final String name;
 
-    // Primary Role of the logged in user
+    private final String mode;
+
+    private final Map<String, String> metadata;
+
+    // Primary Role of the logged-in user
     private final RoleResource role;
+
+    /**
+     * Defines a mode that could not be determined (none provided to constructor).
+     */
+    public static final String UNKNOWN_MODE = "UNKNOWN";
 
     public AuthenticatedUser(String name)
     {
+        this(name, UNKNOWN_MODE);
+    }
+
+    public AuthenticatedUser(String name, String mode)
+    {
+        this(name, mode, Collections.emptyMap());
+    }
+
+    public AuthenticatedUser(String name, String mode, Map<String, String> metadata)
+    {
         this.name = name;
         this.role = RoleResource.role(name);
+        this.mode = mode;
+        this.metadata = metadata;
     }
 
     public String getName()
@@ -72,6 +95,22 @@ public class AuthenticatedUser
     public RoleResource getPrimaryRole()
     {
         return role;
+    }
+
+    /**
+     * The mode of authentication used to authenticate this user, e.g. 'password', 'mtls', 'anonymous', etc.
+     */
+    public String getMode()
+    {
+        return mode;
+    }
+
+    /**
+     * {@link IAuthenticator}-contextual metadata about how the user was authenticated.
+     */
+    public Map<String, String> getMetadata()
+    {
+        return metadata;
     }
 
     /**
@@ -179,6 +218,12 @@ public class AuthenticatedUser
     @Override
     public int hashCode()
     {
+        // Note: for reasons of maintaining the invariant that an object that equals maintains the same hashCode,
+        // we do not include mode and metadata in the hashCode calculation.
+        // This is particularly salient as there are cases where AuthenticatedUser is used as a key in
+        // Role/Permissions cache. In effect, we would like to treat all connections sharing the same name as the same
+        // user, where mode and metadata are just additional context about how the user authenticated that
+        // should not factor into 'equivalence' of users.
         return Objects.hashCode(name);
     }
 }
--- a/src/java/org/apache/cassandra/auth/AuthenticatedUser.java
+++ b/src/java/org/apache/cassandra/auth/AuthenticatedUser.java
@@ -96,7 +96,7 @@ public class AuthenticatedUser
     }
 
     /**
-     * The mode of authentication used to authenticate this user.
+     * @returns the mode of authentication used to authenticate this user
      */
     public AuthenticationMode getAuthenticationMode()
     {
--- a/src/java/org/apache/cassandra/auth/AuthenticatedUser.java
+++ b/src/java/org/apache/cassandra/auth/AuthenticatedUser.java
@@ -104,7 +104,7 @@ public class AuthenticatedUser
     }
 
     /**
-     * {@link IAuthenticator}-contextual metadata about how the user was authenticated.
+     * @returns {@link IAuthenticator}-contextual metadata about how the user was authenticated
      */
     public Map<String, Object> getMetadata()
     {
--- a/src/java/org/apache/cassandra/auth/AuthenticatedUser.java
+++ b/src/java/org/apache/cassandra/auth/AuthenticatedUser.java
@@ -60,7 +60,7 @@ public class AuthenticatedUser
 
     private final String name;
 
-    private final String authenticationMode;
+    private final AuthenticationMode authenticationMode;
 
     private final Map<String, Object> metadata;
 
@@ -79,16 +79,6 @@ public class AuthenticatedUser
 
     public AuthenticatedUser(String name, AuthenticationMode authenticationMode, Map<String, Object> metadata)
     {
-        this(name, authenticationMode.getDisplayName(), metadata);
-    }
-
-    public AuthenticatedUser(String name, String authenticationMode)
-    {
-        this(name, authenticationMode, Collections.emptyMap());
-    }
-
-    public AuthenticatedUser(String name, String authenticationMode, Map<String, Object> metadata)
-    {
         this.name = name;
         this.role = RoleResource.role(name);
         this.authenticationMode = authenticationMode;
@@ -106,11 +96,9 @@ public class AuthenticatedUser
     }
 
     /**
-     * The mode of authentication used to authenticate this user, this will usually be
-     * some value of {@link AuthenticationMode#getDisplayName()} unless
-     * an implementor provides their own custom authentication scheme.
+     * The mode of authentication used to authenticate this user.
      */
-    public String getAuthenticationMode()
+    public AuthenticationMode getAuthenticationMode()
     {
         return authenticationMode;
     }
--- a/src/java/org/apache/cassandra/auth/AuthenticatedUser.java
+++ b/src/java/org/apache/cassandra/auth/AuthenticatedUser.java
@@ -77,6 +77,16 @@ public class AuthenticatedUser
         this(name, authenticationMode, Collections.emptyMap());
     }
 
+    /**
+     * Defines authenticated user context established within a client connection.
+     *
+     * @param name The user's role name
+     * @param authenticationMode How the user was authenticated
+     * @param metadata contextual metadata about how the user was authenticated.  Note that this data is exposed
+     *                 through the <code>system_views.clients table</code>, <code>nodetool clientstats</code> and
+     *                 {@link org.apache.cassandra.metrics.ClientMetrics}-based JMX Beans.  Implementors should
+     *                 take care not to store anything sensitive here.
+     */
     public AuthenticatedUser(String name, AuthenticationMode authenticationMode, Map<String, Object> metadata)
     {
         this.name = name;
@@ -104,7 +114,11 @@ public class AuthenticatedUser
     }
 
     /**
-     * @returns {@link IAuthenticator}-contextual metadata about how the user was authenticated
+     * @returns {@link IAuthenticator}-contextual metadata about how the user was authenticated.
+     * <p>
+     * Note that this data is exposed through the <code>system_views.clients table</code>,
+     * <code>nodetool clientstats</code> and {@link org.apache.cassandra.metrics.ClientMetrics}-based JMX Beans.
+     * Implementors should take care not to store anything sensitive here.
      */
     public Map<String, Object> getMetadata()
     {
--- a/src/java/org/apache/cassandra/auth/AuthenticatedUser.java
+++ b/src/java/org/apache/cassandra/auth/AuthenticatedUser.java
@@ -23,8 +23,11 @@ import java.util.Map;
 import java.util.Set;
 import com.google.common.base.Objects;
 
+import org.apache.cassandra.auth.IAuthenticator.AuthenticationMode;
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.dht.Datacenters;
+
+import static org.apache.cassandra.auth.IAuthenticator.AuthenticationMode.UNAUTHENTICATED;
 
 /**
  * Returned from IAuthenticator#authenticate(), represents an authenticated user everywhere internally.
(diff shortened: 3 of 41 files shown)


src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
```diff
@@ -22,13 +22,17 @@
...
+import com.google.common.collect.Sets;
+
 import org.apache.cassandra.exceptions.AuthenticationException;
 import org.apache.cassandra.exceptions.ConfigurationException;
 
 public class AllowAllAuthenticator implements IAuthenticator
 {
     private static final SaslNegotiator AUTHENTICATOR_INSTANCE = new Negotiator();
 
+    private static final Set<String> MODES = Sets.newHashSet(AuthenticatedUser.ANONYMOUS_USERNAME);
```
smiklosovic: it feels strange to me to reuse "anonymous user name" for something like authentication modes. I know that "well it is 'anonymous' and we need that constant' but we should not reuse that even it is of same value.

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
```diff
@@ -33,10 +35,10 @@
 public class AuthenticatedUser
 {
     public static final String SYSTEM_USERNAME = "system";
-    public static final AuthenticatedUser SYSTEM_USER = new AuthenticatedUser(SYSTEM_USERNAME);
+    public static final AuthenticatedUser SYSTEM_USER = new AuthenticatedUser(SYSTEM_USERNAME, SYSTEM_USERNAME);
```
smiklosovic: again, using same constant for "authentication mode" just feels strange. What has _username_ common with _authentication mode_ except it is of same value? We should have an enum where all possible authentication modes are and we should reference it here, not reusing already existing constants for that.

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
```diff
@@ -55,13 +57,34 @@ public static void init()
 
     private final String name;
 
-    // Primary Role of the logged in user
+    private final String mode;
```
smiklosovic: renaming to "authenticationMode" would help. It could be also enum.

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
```diff
@@ -55,13 +57,34 @@ public static void init()
...
+
+    private final Map<String, String> metadata;
+
+    // Primary Role of the logged-in user
     private final RoleResource role;
 
+    /**
+     * Defines a mode that could not be determined (none provided to constructor).
+     */
+    public static final String UNKNOWN_MODE = "UNKNOWN";
```
smiklosovic: extracting this to enum would help with all other modes

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
```diff
@@ -74,6 +97,22 @@ public RoleResource getPrimaryRole()
         return role;
     }
 
+    /**
+     * The mode of authentication used to authenticate this user, e.g. 'password', 'mtls', 'anonymous', etc.
+     */
+    public String getMode()
```
smiklosovic: public AuthenticationMode getAuthenticationMode() { ... }

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
```diff
@@ -74,6 +97,22 @@ public RoleResource getPrimaryRole()
...
+    {
+        return mode;
+    }
+
+    /**
+     * {@link IAuthenticator}-contextual metadata about how the user was authenticated.
```
smiklosovic: how it _was authenticated_ or how _it is going to be authenticated_ ?

src/java/org/apache/cassandra/auth/IAuthenticator.java
```diff
@@ -179,5 +189,13 @@ default boolean shouldSendAuthenticateMessage()
         {
             return true;
         }
+
+        /**
+         * @return The assumed mode of authentication attempted using this negotiator (e.g. 'password', 'mtls')
+         * which is surfaced in metrics.
+         */
+        default String getMode() {
```
smiklosovic: braces

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
```diff
@@ -74,6 +97,22 @@ public RoleResource getPrimaryRole()
...
+     */
+    public Map<String, String> getMetadata()
```
smiklosovic: will a value only be `String` ever?

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
```diff
@@ -69,6 +70,12 @@ public class MutualTlsAuthenticator implements IAuthenticator
     private static final String CACHE_NAME = "IdentitiesCache";
     private final IdentityCache identityCache = new IdentityCache();
     private final MutualTlsCertificateValidator certificateValidator;
+    static final String MODE = "mtls";
```
smiklosovic: should be extracted to that `AuthenticationMode` enum.

src/java/org/apache/cassandra/auth/MutualTlsWithPasswordFallbackAuthenticator.java
```diff
@@ -34,6 +37,9 @@
 public class MutualTlsWithPasswordFallbackAuthenticator extends PasswordAuthenticator
 {
     private final MutualTlsAuthenticator mutualTlsAuthenticator;
+
+    private static final Set<String> MODES = Sets.newHashSet(PasswordAuthenticator.MODE, MutualTlsAuthenticator.MODE);
```
smiklosovic: I am just wondering ... why are we referencing these String constants in either authenticators? I think that the more "clean" approach would be to return in `getSupportedAuthenticationModes` `PasswordAuthenticator.MODES` and `MutualTlsAuthenticator.MODES`. In other words, we should create a new set by combination of `MODES`. That way, when some `MODE` was added to these authenticators, this one would already return them all without any need to add them here.

src/java/org/apache/cassandra/db/virtual/ClientsTable.java
```diff
@@ -86,7 +90,9 @@ public DataSet data()
                   .column(SSL_ENABLED, client.sslEnabled())
                   .column(SSL_PROTOCOL, client.sslProtocol().orElse(null))
                   .column(SSL_CIPHER_SUITE, client.sslCipherSuite().orElse(null))
-                  .column(KEYSPACE_NAME, client.keyspace().orElse(null));
+                  .column(KEYSPACE_NAME, client.keyspace().orElse(null))
+                  .column(MODE, client.mode().orElse(null))
+                  .column(METADATA, client.metadata().orElse(null));
```
smiklosovic: when we return `null` here, does it mean that `public Map<String, String> getMetadata()` in `AuthenticatedUser` might return `null` as well? So maybe putting `Nullable` annotation on top of that method is appropriate? Is not it more comfortable to always return an empty map?

src/java/org/apache/cassandra/metrics/ClientMetrics.java
```diff
@@ -20,12 +20,20 @@
 
 import java.util.*;
```
smiklosovic: could you expand these imports?

src/java/org/apache/cassandra/metrics/ClientMetrics.java
```diff
@@ -20,12 +20,20 @@
...
 
 import com.codahale.metrics.Gauge;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
 import com.codahale.metrics.Reservoir;
+import org.apache.cassandra.auth.AuthenticatedUser;
+import org.apache.cassandra.auth.IAuthenticator;
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.service.ClientState;
 import org.apache.cassandra.transport.*;
```
smiklosovic: same

src/java/org/apache/cassandra/metrics/ClientMetrics.java
```diff
@@ -51,20 +84,62 @@ public final class ClientMetrics
     private Meter protocolException;
     private Meter unknownException;
 
+    private static String AUTH_SUCCESS = "AuthSuccess";
```
smiklosovic: all might be `final` too

src/java/org/apache/cassandra/metrics/ClientMetrics.java
```diff
@@ -51,20 +84,62 @@ public final class ClientMetrics
...
+    public void markAuthFailure(final String mode)
+    {
+        markAuthMeter(authFailure, AUTH_FAILURE, mode);
+    }
+
+    private void markAuthMeter(final Meter meter, final String meterPrefix, final String mode)
+    {
+        meter.mark();
+
+        if (mode != null)
```
smiklosovic: would not it be easier to read if we `return` when `mode == null`? Then you would not put this whole logic into `if` and a reader does not need to hold the context of "ok, mode is not null here" in their head. Just return / break the execution flow as soon as possible.

src/java/org/apache/cassandra/metrics/ClientMetrics.java
```diff
@@ -51,20 +84,62 @@ public final class ClientMetrics
...
+        {
+            Map<String, Meter> meterMap = meterPrefix.equals(AUTH_SUCCESS) ? authSuccessByMode : authFailureByMode;
+
+            // if there is no meter for this mode, the authenticator must have surfaced a mode that it doesn't
+            // list in 'getSupportedAuthenticationModes', in this case, just don't mark.
+            Meter modeMeter = meterMap.get(mode);
+            if (modeMeter != null)
+            {
```
smiklosovic: technically you dont need braces here, it is just a single-statement block. This could be just simplified to more idiomatic one-liner

Optional.ofNullable(meterMap.get(mode)).ifPresent(Meter::mark);

src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
tolbertam: :+1:, makes sense.  Would 'unauthenticated' be ok in this case?  Either that or 'none'?

src/java/org/apache/cassandra/metrics/ClientMetrics.java
```diff
@@ -51,20 +84,62 @@ public final class ClientMetrics
...
     {
         authFailure.mark();
     }
 
+    public void markAuthFailure(final String mode)
+    {
+        markAuthMeter(authFailure, AUTH_FAILURE, mode);
+    }
+
+    private void markAuthMeter(final Meter meter, final String meterPrefix, final String mode)
```
smiklosovic: meterPrefix might be enum, you always pick between `authSuccessByMode` or `authFailureByMode`.

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
tolbertam: I had considered an Enum, but one thing I was mindful of was third party `IAuthenticator` implementations.  If they wanted to introduce their own mode of auth, they would not be able to without making a change to the core codebase since Enums are effectively final.  

Is this something we should support?  If not I'm fine with using an Enum.

src/java/org/apache/cassandra/auth/IAuthenticator.java
tolbertam: argh!  I missed one again, I really need to get going on [CASSANDRA-19276](https://issues.apache.org/jira/browse/CASSANDRA-19276) 😅

src/java/org/apache/cassandra/metrics/ClientMetrics.java
```diff
@@ -118,6 +193,31 @@ public long getCount()
...
 
+        // For each of SSL, non-SSL register a gauge:
+        encryptedConnectedNativeClients = registerGauge(new DefaultNameFactory("Client", "Encrypted"), CONNECTED_NATIVE_CLIENTS, () -> countConnectedClients((ServerConnection::isSSL)));
+        unencryptedConnectedNativeClients = registerGauge(new DefaultNameFactory("Client", "Unencrypted"), CONNECTED_NATIVE_CLIENTS, () -> countConnectedClients(((ServerConnection connection) -> !connection.isSSL())));
+
+        // for each supported authentication mode, register a meter for success and failures.
+        final IAuthenticator authenticator = DatabaseDescriptor.getAuthenticator();
+        for (String mode : authenticator.getSupportedAuthenticationModes())
+        {
+            String capitalizedMode = StringUtils.capitalize(mode);
```
smiklosovic: why can't we save the mode capitalized already? If `mode` was an enum, you would just use its name. 

If there is a 3rd party implementation with some custom mode and they do not capitalize it, why are we making them uppercased? If they want to have it upper-cased, they should capitalize it themselves. If they want to have it lower cased, they will.

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
tolbertam: Primarily for serialization reasons.  I want to surface this in `system_views.clients`;  But I suppose in theory I could convert the map to a `<string,string>` when representing it in the table, so this could feasibly be `<string,object>`

src/java/org/apache/cassandra/metrics/ClientMetrics.java
```diff
@@ -118,6 +193,31 @@ public long getCount()
...
+            MetricNameFactory factory = new DefaultNameFactory("Client", capitalizedMode);
+            authSuccessByMode.put(mode, registerMeter(factory, AUTH_SUCCESS));
+            authFailureByMode.put(mode, registerMeter(factory, AUTH_FAILURE));
+
+            Gauge<Integer> clients = registerGauge(factory, CONNECTED_NATIVE_CLIENTS, () -> countConnectedClients((ServerConnection connection) -> {
+                final ClientState state = connection.getClientState();
```
smiklosovic: Why is `state` final and `user` is not? Why does it have to be final?

smiklosovic: this might be all simplified to

    AuthenticatedUser user = connection.getClientState().getUser();
    return Optional.ofNullable(user).map(u -> mode.equals(u.getMode())).orElse(false);

src/java/org/apache/cassandra/tools/nodetool/ClientStats.java
```diff
@@ -47,6 +50,9 @@ public class ClientStats extends NodeToolCmd
     @Option(title = "list_connections_with_client_options", name = "--client-options", description = "Lists all connections and the client options")
     private boolean clientOptions = false;
 
+    @Option(title = "list_connections_with_metadata", name = "--metadata", description = "Lists all connections and authenticator-specific metadata and mode")
```
smiklosovic: I would probably rename this to `--verbose`. I did something similar here 

https://github.com/apache/cassandra/commit/37acd27f2d0e79fac969e34304bff2d6641728f5#diff-d4d258cb191fc8ee25c5a08c78b56505688d7d6eb2dbea6390f15dc16dfea17d

I think what you also follow by doing this is preserving some kind of a backward compatibility so users are not suddenly put in the position the output of the nodetool would be changed. I think it is a good practice to teach them to use "verbose" flag if they want to know more details.

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
smiklosovic: we could still use enum for our core stuff but we would call .toString() on that to have it formally on String for 3rd parties if it is important for them.

src/java/org/apache/cassandra/transport/Server.java
```diff
@@ -306,6 +315,24 @@ int countConnectedClients()
             return allChannels.size() != 0 ? allChannels.size() - 1 : 0;
         }
 
+        int countConnectedClients(Predicate<ServerConnection> predicate)
+        {
+            int count = 0;
+            for (Channel c : allChannels)
```
smiklosovic, line 321: what happens when connection is closed while you iterate over the channels? what if channels are closed?

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
smiklosovic: stupid question ... of course _how it was authenticated_ as this is in the class called "AuthenticatedUser" :D

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
tolbertam: 🤔 so use an Enum internally, and `String` on the public API methods?  That seems like a good compromise, i'll give that a try!

src/java/org/apache/cassandra/metrics/ClientMetrics.java
```diff
@@ -51,20 +84,62 @@ public final class ClientMetrics
...
+        markAuthMeter(authFailure, AUTH_FAILURE, mode);
+    }
+
+    private void markAuthMeter(final Meter meter, final String meterPrefix, final String mode)
+    {
+        meter.mark();
+
+        if (mode != null)
+        {
+            Map<String, Meter> meterMap = meterPrefix.equals(AUTH_SUCCESS) ? authSuccessByMode : authFailureByMode;
```
smiklosovic: @tolbertam  btw why cant we pass the map in the parameter? you know which map you want based on what method you call it from so you do not need to know prefix at all, no? Do we even need this method at all when it shrinks to meter.mark() and one Optional.... as shown below?

src/java/org/apache/cassandra/db/virtual/ClientsTable.java
tolbertam: Hrmm, yeah thats a good point.  I'll make this non-optional, these values should be NonNull, so they don't really need to be optional, i'll fix that all.

src/java/org/apache/cassandra/metrics/ClientMetrics.java
tolbertam: i'll do that in ClientsTable as well as noticed a wildcard import there as well.

src/java/org/apache/cassandra/metrics/ClientMetrics.java
tolbertam: yeah that would make more sense to just pass in the Map of meters instead of resolving it based on the passed in value, will do :+1:

src/java/org/apache/cassandra/metrics/ClientMetrics.java
tolbertam: :+1: agreed, better to leave it up to the implementors

src/java/org/apache/cassandra/db/virtual/ClientsTable.java
tolbertam: ah, just realized why I did this.  There is code in `ConnectedClient.getUser()` which seems to imply that it is not assured that `ClientState.getUser()` will return a non-null value.  Just to simplify this a bit, I think we can just provide default values (UNAUTHENTICATED and empty map) to not have to reason about optional.

src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
tolbertam: went with 'Unauthenticated' but if have a suggestion for a name would be happy to change it :+1:

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
tolbertam: Just pushed [605d9cb](https://github.com/apache/cassandra/pull/3085/commits/605d9cbaa047c79ba7f822a5c9c4ea915de7f269) which includes `AuthenticationMode` enum in `IAuthenticator`.

It feels a bit weird that i'm converting the enum to a String in a number of places, seems like it could be error prone and I feel like we're not getting the full value of using an enum.

I had a few more possible ideas I'd throw out there to gauge your thoughts:

1. Add an enum value `CUSTOM` to `AuthenticationMode`, that 3rd party implementors can use to indicate that the authentication mode used is something 'custom'.    This would allow us to use `AuthenticationMode` all throughout the API (without having to convert explicitly to String anywhere except maybe in the system table).  The main downside is that it's not very specific, but I think for most implementors this would be sufficient as I don't expect an implementor will support multiple custom modes in their authenticator.
2. Instead of using an enum, introduce a class for `AuthenticationMode` that a third party could implement. I feel this one offers less value, and is only a slight improvement over string.

option 1 isn't perfect but I think could be an improvement over what I just implemented and will make the API cleaner and less error prone.  What do you think?

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
tolbertam: changed it to `<string, object>` :+1:

src/java/org/apache/cassandra/auth/MutualTlsWithPasswordFallbackAuthenticator.java
tolbertam: agreed, changed this :+1:

src/java/org/apache/cassandra/metrics/ClientMetrics.java
tolbertam: :+1: fixed, also used `Optional.ofNullable(` on the mode check

src/java/org/apache/cassandra/tools/nodetool/ClientStats.java
tolbertam: makes sense, replaced `--metadata` with `--verbose` and `--verbose` will simply include all fields of the connection now.

src/java/org/apache/cassandra/transport/Server.java
tolbertam: I'd expect that even if the Channel is closed, the attributes will still be present. If the `ServerConnection` does happen to be null, the instanceof check will fail.  The `ClientState` on the `connection` (which the predicates evaluate) will never be null, so it should generally be safe.

Also, if an exception is thrown here, the calling thread will be a JMX rmi connection thread, which will simply bubble the exception up to the client, so I think it will behave ok even in exceptional states.

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
```diff
@@ -74,6 +105,24 @@ public RoleResource getPrimaryRole()
...
     }
 
+    /**
+     * The mode of authentication used to authenticate this user, this will usually be
+     * some value of {@link AuthenticationMode#getDisplayName()} unless
+     * an implementor provides their own custom authentication scheme.
+     */
+    public String getAuthenticationMode()
+    {
+        return authenticationMode;
```
smiklosovic, line 113: @tolbertam I think that `private final String authenticationMode;` can be of type `AuthenticationMode` and the only place where it needs to be String (here) you would call `displayName()` on that. I would just use AuthenticationMode everywhere I could and converted that to String when necessary.

src/java/org/apache/cassandra/auth/IAuthenticator.java
```diff
@@ -179,5 +194,50 @@ default boolean shouldSendAuthenticateMessage()
...
         }
+
+        /**
+         * @return The assumed mode of authentication attempted using this negotiator, this will usually be some value
+         * of {@link AuthenticationMode#getDisplayName()} unless an implementor provides their own custom authentication
+         * scheme.
+         */
+        default String getAuthenticationMode()
+        {
+            return AuthenticationMode.UNAUTHENTICATED.getDisplayName();
```
smiklosovic, line 205: @tolbertam  see? Like this.

src/java/org/apache/cassandra/auth/IAuthenticator.java
```diff
@@ -179,5 +194,50 @@ default boolean shouldSendAuthenticateMessage()
...
+
+        private final String displayName;
+
+        AuthenticationMode()
+        {
+            this.displayName = StringUtils.capitalize(name().toLowerCase());
+        }
+
+        /**
+         * @return How the mode should be displayed to users. Formally it is the enum name with the first letter
```
smiklosovic: @tolbertam aha, this is interesting ... why have you chose such a naming convention? Why not just `AuthenticationMode.name()` ? Is it because of JMX?

src/java/org/apache/cassandra/metrics/ClientMetrics.java
```diff
@@ -51,20 +93,56 @@ public final class ClientMetrics
...
     {
         authFailure.mark();
     }
 
+    public void markAuthFailure(String mode)
+    {
+        markAuthMeter(authFailure, authFailureByMode, mode);
+    }
+
+    private void markAuthMeter(Meter meter, Map<String, Meter> meterMap, String mode)
```
smiklosovic: I would personally get rid of this method and just put it into methods which call this logic. The only variable there would be `meter` and `meterMap` which would be either `authFailureByMode` or `authSuccessByMode`. This kind of repetition is not a big deal ... I just dont see any "win" by having this method.

src/java/org/apache/cassandra/transport/Server.java
smiklosovic: It would be great to have some basic test around this.

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
smiklosovic: But on the other hand, I think I understand why you did it like this ... because 3rd party implementation with their own mode would not be in AuthenticationMode enum so they still need to find a way how to populate this with something custom.

doc/modules/cassandra/pages/managing/operating/metrics.adoc
```diff
@@ -913,14 +913,87 @@ Reported name format:
...
+
+*Metric Name*::
+`org.apache.cassandra.metrics.Client.<Mode>.<MetricName>`
+*JMX MBean*::
+`org.apache.cassandra.metrics:type=Client scope=<Mode> name=<MetricName>`
+
+An authentication mode is a supported method of authentication. The following
+authentication modes exist for the given supported `IAuthenticators`:
+
+* `PasswordAuthenticator` - password
```
smiklosovic: @tolbertam this should be capitilized? Password, Mtls, Anonymous ... right?

src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
smiklosovic: that's fine

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
smiklosovic: @tolbertam I think the approach 2) is a little bit superior here.  The reason I think so is that, as you said, CUSTOM is quite non-telling and it would also mean that absolutely every other authentication mode, not part of core Cassandra, would be CUSTOM ever. I don't like this. In system_views.clients, it would be all "Custom", for JMX as well ... We can definitely do better. It would make people question what actual authentication mode that node is configured with if they do not see something explicit but just "Custom". Like ... custom what? That is just not helpful.

I wanted to see how it would look like on a class so I implemented it here https://github.com/tolbertam/cassandra/pull/1

It is just a commit on top of your latest stuff here, feel free to cherry-pick that.

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
smiklosovic: this is a non-issue with the PR I provided in the comment above.

src/java/org/apache/cassandra/auth/IAuthenticator.java
tolbertam: I thought `Password` would be more friendly than `PASSWORD` and `name()` returns the all caps representation of the enum (and is not overridable)

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
tolbertam: Thanks for implementing that, that looks very good and does look like a much better solution.  I'll cherry pick it over, thank you!

src/java/org/apache/cassandra/metrics/ClientMetrics.java
tolbertam: :+1: yeah at this point the method has been reduced enough that it doesn't offer much, i'll factor that out

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
tolbertam: > because 3rd party implementation with their own mode would not be in AuthenticationMode enum so they still need to find a way how to populate this with something custom.

Exactly, I think the solution you came up works really well for this.  We get a type with some well-defined values, and third-party implementors can add their own if they'd like.

src/java/org/apache/cassandra/transport/Server.java
tolbertam: :+1: let me think about how I can test this.  The challenge will be that I think to reproduce it the channel will need to be closed.  This approach was borrowed from other methods on the class (`countConnectedClientsByUser`) so maybe there is some existing test precedence I can pull from.

Alternatively I could probably inject an unexpected failure using byteman..will keep thinking on this.

smiklosovic: @tolbertam you could also probably mock that when `c.attr(Connection.attributeKey).get()` is called that get() returns null or throws.

src/java/org/apache/cassandra/auth/AllowAllAuthenticator.java
```diff
@@ -22,13 +22,17 @@
...
+import com.google.common.collect.Sets;
+
 import org.apache.cassandra.exceptions.AuthenticationException;
 import org.apache.cassandra.exceptions.ConfigurationException;
 
 public class AllowAllAuthenticator implements IAuthenticator
 {
     private static final SaslNegotiator AUTHENTICATOR_INSTANCE = new Negotiator();
 
+    private static final Set<AuthenticationMode> AUTHENTICATION_MODES = Sets.newHashSet(AuthenticationMode.UNAUTHENTICATED);
```
frankgh: NIT: prefer JDK APIs when possible:
```suggestion
    private static final Set<AuthenticationMode> AUTHENTICATION_MODES = Collections.singleton(AuthenticationMode.UNAUTHENTICATED);
```

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
```diff
@@ -74,6 +95,22 @@ public RoleResource getPrimaryRole()
         return role;
     }
 
+    /**
+     * The mode of authentication used to authenticate this user.
```
frankgh: NIT:
```suggestion
     * @returns the mode of authentication used to authenticate this user
```

src/java/org/apache/cassandra/auth/AuthenticatedUser.java
```diff
@@ -74,6 +95,22 @@ public RoleResource getPrimaryRole()
...
+     */
+    public AuthenticationMode getAuthenticationMode()
+    {
+        return authenticationMode;
+    }
+
+    /**
+     * {@link IAuthenticator}-contextual metadata about how the user was authenticated.
```
frankgh:
```suggestion
     * @returns {@link IAuthenticator}-contextual metadata about how the user was authenticated
```

src/java/org/apache/cassandra/auth/IAuthenticator.java
```diff
@@ -104,12 +110,23 @@ default SaslNegotiator newSaslNegotiator(InetAddress clientAddress, Certificate[
         return newSaslNegotiator(clientAddress);
     }
 
+    /**
+     * @return The supported authentication 'modes' of this authenticator.
+     * scheme.
```
frankgh:
```suggestion
     * @return The supported authentication 'modes' of this authenticator scheme
```

src/java/org/apache/cassandra/auth/IAuthenticator.java
```diff
@@ -179,5 +195,64 @@ default boolean shouldSendAuthenticateMessage()
...
+
+        /**
+         * User authenticated using a password.
+         */
+        public static final AuthenticationMode PASSWORD = new AuthenticationMode("Password") {};
+
+        /**
+         * User authenticated using a trusted identity in their client certificate.
+         */
+        public static final AuthenticationMode MTLS = new AuthenticationMode("Mtls") {};
```
frankgh: Should we spell out the label?
```suggestion
        public static final AuthenticationMode MTLS = new AuthenticationMode("Mutual TLS") {};
```

src/java/org/apache/cassandra/transport/Server.java
tolbertam: Added `ConnectionTrackerTest` that includes:

* Validate that `Channel` lacking a `Connection.attributeKey` get skipped over.
* Validate that `Channel` having a `Connection.attributeKey` that is not a `ServerConnection` get skipped.
* Validate that if `c.attr(Connection.attributeKey).get()` throws an unhandled exception it gets propagated up to the caller (ultimately preventing the calculation from working, which I think is the behavior we want, but let me know if you think that should be different).
* Also added tests that validate the expected basic functionality of `countConnectedClients(predicate)` and `countConnectedClientsByUser` (which I didn't change in this PR but figured would be nice to have at a unit level).

src/java/org/apache/cassandra/metrics/ClientMetrics.java
```diff
@@ -51,20 +93,50 @@ public final class ClientMetrics
...
+    {
+        markAuthSuccess(null);
+    }
+
+    public void markAuthSuccess(String mode)
     {
         authSuccess.mark();
+        Optional.ofNullable(mode)
+                .flatMap((m) -> Optional.ofNullable(authSuccessByMode.get(m)))
+                .ifPresent(Meter::mark);
```
frankgh: a bit hard to read, but it is the equivalent of:
```
        Meter meterByMode;
        if (mode != null && (meterByMode = authSuccessByMode.get(mode)) != null) {
            meterByMode.mark();
        }
```

src/java/org/apache/cassandra/metrics/ClientMetrics.java
```diff
@@ -51,20 +93,50 @@ public final class ClientMetrics
...
     }
 
+    /**
+     * @deprecated by {@link #markAuthFailure(String)}
+     */
+    @Deprecated(since="5.1", forRemoval = true)
     public void markAuthFailure()
     {
         authFailure.mark();
```
frankgh: should this call `markAuthFailure(null)` just like we do for the success case?

src/java/org/apache/cassandra/tools/nodetool/ClientStats.java
```diff
@@ -86,54 +92,51 @@ public void execute(NodeProbe probe)
...
 
-        if (listConnections)
+        // Note: for compatbility with existing implementation if someone passes --all (listConnections),
+        // --client-options, and --metadata all three will be printed.
+        List<Map<String, String>> clients = (List<Map<String, String>>) probe.getClientMetric("connections");
+        if (!clients.isEmpty() && (listConnections || clientOptions || verbose))
         {
-            List<Map<String, String>> clients = (List<Map<String, String>>) probe.getClientMetric("connections");
-            if (!clients.isEmpty())
+            List<String> tableHeaderBase = Lists.newArrayList("Address", "SSL", "Cipher", "Protocol", "Version",
```
frankgh: NIT: prefer JDK APIs. Since trunk and 5.0 are on Java 11+ 
```suggestion
            List<String> tableHeaderBase = List.of("Address", "SSL", "Cipher", "Protocol", "Version",
```

src/java/org/apache/cassandra/tools/nodetool/ClientStats.java
```diff
@@ -86,54 +92,51 @@ public void execute(NodeProbe probe)
...
+                    "User", "Keyspace", "Requests", "Driver-Name",
+                    "Driver-Version");
+            List<String> tableFieldsBase = Lists.newArrayList(ConnectedClient.ADDRESS, ConnectedClient.SSL,
```
frankgh:
```suggestion
            List<String> tableFieldsBase = List.of(ConnectedClient.ADDRESS, ConnectedClient.SSL,
```

src/java/org/apache/cassandra/transport/messages/AuthUtil.java
```diff
@@ -58,6 +58,7 @@ public final class AuthUtil
     static Response handleLogin(Connection connection, QueryState queryState, byte[] token,
                                 BiFunction<Boolean, byte[], Response> messageToSendBasedOnNegotiation)
     {
+        IAuthenticator.SaslNegotiator negotiator = ((ServerConnection) connection).getSaslNegotiator(queryState);
```
frankgh, line 61: it seems we only need the negotiator in the scope of the catch below, should we move it there instead?

frankgh (commented):
Looks good in general, I've left some minor comments

src/java/org/apache/cassandra/transport/Server.java
smiklosovic: great! that is enough I guess ...

src/java/org/apache/cassandra/auth/IAuthenticator.java
smiklosovic: I think this would break JMX as I am not sure how a metric with a space in its name would behave.

smiklosovic: we should probably check that the display name does not contain space

frankgh: good point, I'm not sure what's the behavior for jmx. But if spaces break JMX, we should add some guards around it

src/java/org/apache/cassandra/metrics/ClientMetrics.java
tolbertam: good catch, this is definitely wrong, will fix it :+1:

src/java/org/apache/cassandra/metrics/ClientMetrics.java
tolbertam: :+1: avoids creating some additional Optional objects as well, will change it.

src/java/org/apache/cassandra/auth/IAuthenticator.java
tolbertam: I'll test whether spaces work w/ JMX, I suspect it does, but isn't desirable as might require clients to escape;  If it doesn't work, I suggest we throw a RuntimeException in the `AuthenticationMode` constructor. 

As far as the name, regardless of whether spaces are allowed, how about `MutualTls` (for consistency with how the authenticator is named)?

src/java/org/apache/cassandra/transport/messages/AuthUtil.java
tolbertam: at line 88 in the catch, we need the `negotiator` to determine the authentication mode used so had to move it outside of the try block.

frankgh: oh I see now, this was moved rather than added.

src/java/org/apache/cassandra/auth/IAuthenticator.java
tolbertam: It looks like it does work:

```
beans
...
org.apache.cassandra.metrics:name=ConnectedNativeClients,scope=Encrypted,type=Client
org.apache.cassandra.metrics:name=ConnectedNativeClients,scope=Mutual Tls,type=Client
org.apache.cassandra.metrics:name=ConnectedNativeClients,scope=Password,type=Client
org.apache.cassandra.metrics:name=ConnectedNativeClients,scope=Unencrypted,type=Client
org.apache.cassandra.metrics:name=ConnectedNativeClients,type=Client
```

But it isn't easy to use, and I imagine will give some tooling trouble.   For it to work with jmxterm for example I had to double `\` escape the space:

```
$>bean org.apache.cassandra.metrics:name=ConnectedNativeClients,scope=Mutual Tls,type=Client
#IllegalArgumentException: Please specify domain using either -d option or domain command
$>bean org.apache.cassandra.metrics:name=ConnectedNativeClients,scope=Mutual\\ Tls,type=Client
#bean is set to org.apache.cassandra.metrics:name=ConnectedNativeClients,scope=Mutual Tls,type=Client
$>get *
#mbean = org.apache.cassandra.metrics:name=ConnectedNativeClients,scope=Mutual Tls,type=Client:
Value = 5;
```

I think we should document that spaces in the mode name aren't preferred as it may not work well with some tooling. I will do that!

tolbertam: Addressed in [8f082a1](https://github.com/apache/cassandra/pull/3085/commits/8f082a189a5ff5f781265b10d7777d47c1af6cc7) and renamed `Mtls` -> `MutualTls`.  I also realized was still using `String` in `ClientMetrics` instead of `AuthenticationMode` so I updated that as well, hopefully that looks good!

src/java/org/apache/cassandra/metrics/ClientMetrics.java
smiklosovic: @tolbertam you dont need the braces :) just super nit but when I see it I just say it.

tolbertam: updated :+1:

frankgh (approved):
+1 I have no further comments. Looks good to me

PR #3088, opened 2024-02-06
CASSANDRA-19189

krummas:
- remove sealed_periods and last_sealed_period tables
- make metadata_snapshots use an ordered partitioner to be able to quickly find the most recent snapshot
- simplify the ways we query snapshots and log table


src/java/org/apache/cassandra/tcm/log/LogReader.java
```diff
@@ -18,16 +18,132 @@
...
+                        // we end up here if `since` is not in currentPeriod or currentPeriod - 1
+                        // so we should return a snapshot - we prefer returning the most recent snapshot + entries in `current`
+                        // but if that doesn't exist we check previous.
+                        if (current.min == null && previous.min == null) // we found no entries >= since in the tables -> since > current
+                            return LogState.EMPTY;
+                        ClusterMetadata snapshot = snapshots().getSnapshot(Epoch.create(current.min.getEpoch() - 1));
+                        if (snapshot != null)
+                            return new LogState(snapshot, ImmutableList.sortedCopyOf(current.entries));
+                    }
+                    ClusterMetadata snapshot = snapshots().getSnapshot(Epoch.create(previous.min.getEpoch() - 1));
```
beobal, line 103: There's a chance we can NPE here if we are missing both a snapshot and entries from the log,  which should come after `since`. This could happen in the local table if we received a snapshot at `since + x`, causing us not to have the individual entries directly following `since`, but then we lost the snapshot. I've pushed a test and fix for this to https://github.com/beobal/cassandra/commit/7d66c2997186a6dda7bc3ba95102207b988cfd4c 
One question there is whether its better to return a `LogState` with a non-contiguous list of entries, or an empty one.

src/java/org/apache/cassandra/tcm/log/LogState.java
```diff
@@ -237,31 +171,122 @@ public Serializer(Version serializationVersion)
...
+        private final LocalLog log;
+
+        public ReplicationHandler(LocalLog log)
+        {
+            this.log = log;
+        }
+
+        public void doVerb(Message<LogState> message) throws IOException
+        {
+            logger.info("Received logstate {} from {}", message.payload, message.from());
```
beobal, line 249: This might be a bit verbose now. Before this, the equivalent log message included only the size & min/max epochs in the Replication. Maybe we could go back to something like that + the epoch of any base state?

src/java/org/apache/cassandra/schema/DistributedMetadataLogKeyspace.java
```diff
@@ -173,105 +171,54 @@ public static boolean tryCommit(Entry.Id entryId,
...
+            this(consistencyLevel, () -> ClusterMetadataService.instance().snapshotManager());
+        }
 
-                assert currentEpoch == lastEpoch;
-                return new Replication(entries.build());
-            }
-            catch (IOException t)
+        public EntryHolder getEntries(long period, Epoch since) throws IOException
+        {
+            UntypedResultSet resultSet = execute(String.format("SELECT current_epoch, period, epoch, kind, transformation, entry_id, sealed FROM %s.%s WHERE period = ? AND epoch >= ?",
```
beobal: `current_epoch` & `period` are selected but not used

beobal (changes requested):
LGTM, only a minor thing in `LogReader::getLogState`

src/java/org/apache/cassandra/tcm/log/LogState.java
krummas: I fixed the toString on LogState

PR #3100, opened 2024-02-13
CASSANDRA-19042: Repair fuzz tests fail with paxos_variant: v2

dcapwell:


src/java/org/apache/cassandra/service/paxos/Paxos.java
```diff
@@ -385,13 +386,18 @@ public EndpointsForToken readCandidates()
         }
 
         static Participants get(TableMetadata table, Token token, ConsistencyLevel consistencyForConsensus)
```
dcapwell, line 388: I left this here as `Paxos` uses it... I updated the stuff that mattered for the existing tests to use the new API.

I did see `PaxosRepair` touches this as well, but I assume it didn't matter as we don't have data in paxos so that logic prob didn't get touched?

src/java/org/apache/cassandra/service/paxos/PaxosRepair.java
```diff
@@ -659,45 +657,24 @@ static boolean validateVersionCompatibility(CassandraVersion version)
         return (version.major == 4 && version.minor > 0) || version.major > 4;
     }
 
-    static String getPeerVersion(InetAddressAndPort peer)
+    static boolean validatePeerCompatibility(IGossiper gossiper, Replica peer)
     {
-        EndpointState epState = Gossiper.instance.getEndpointStateForEndpoint(peer);
```
dcapwell, line 664: this is duplicate logic from `Gossiper`, the only difference was

```
try
        {
            return value.value;
        }
        catch (IllegalArgumentException e)
        {
            return null;
        }
```

Not sure how a field access leads to `IllegalArgumentException`, so removed in favor of the gossip function

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosCleanupSession.java
```diff
@@ -38,23 +36,21 @@
...
 import static org.apache.cassandra.config.CassandraRelevantProperties.PAXOS_CLEANUP_SESSION_TIMEOUT_SECONDS;
 import static org.apache.cassandra.net.Verb.PAXOS2_CLEANUP_REQ;
-import static org.apache.cassandra.utils.Clock.Global.nanoTime;
 
 public class PaxosCleanupSession extends AsyncFuture<Void> implements Runnable,
                                                                       IEndpointStateChangeSubscriber,
                                                                       IFailureDetectionEventListener,
                                                                       RequestCallbackWithFailure<Void>
 {
-    private static final Map<UUID, PaxosCleanupSession> sessions = new ConcurrentHashMap<>();
```
dcapwell, line 55: moved to `org.apache.cassandra.service.paxos.cleanup.PaxosRepairState`

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosCleanupSession.java
```diff
@@ -95,49 +93,40 @@ private static ScheduledFuture<?> schedule(PaxosCleanupSession session)
         }
     }
 
-    private final UUID session = UUID.randomUUID();
+    private final SharedContext ctx;
+    public final UUID session = UUID.randomUUID();
```
dcapwell, line 97: this is non-deterministic, but this problem exists in repair as well... Repair uses random/time uuid outside of the random source/clock from `SharedContext`... since these are keys for a map, I have mostly looked the other way... it makes debugging annoying (as you can not block on the repair id) but its not been an issue so just ignore

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosFinishPrepareCleanup.java
```diff
@@ -77,93 +68,10 @@ public synchronized void onResponse(Message<Void> msg)
             trySuccess(null);
     }
 
-    static class PendingCleanup extends IntrusiveStack<PendingCleanup>
```
dcapwell, line 80: moved to `org.apache.cassandra.service.paxos.cleanup.PaxosRepairState`

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosRepairState.java
```diff
@@ -0,0 +1,205 @@
...
+    public void clearRepairs()
+    {
+        for (PaxosTableRepairs repairs : tableRepairsMap.values())
+            repairs.clear();
+    }
+
+
+    public void setSession(PaxosCleanupSession session)
+    {
+        //TODO (review): I copy/pasted from PaxosCleanupSession... this is a check-then-act pattern which is not thread safe...
```
dcapwell: I can fix if desired, I left alone so I could lower the risk in the refactor

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosRepairState.java
```diff
@@ -0,0 +1,205 @@
...
+        if (session != null)
+            session.finish(from, response);
+    }
+    
+    public void addCleanupHistory(Message<PaxosCleanupHistory> message)
+    {
+        PendingCleanup.add(ctx, pendingCleanup, message);
+    }
+
+    private static class Holder
```
dcapwell, line 116: this is not required, but it helps to see what escapes simulation... can put a break point on `instance()` while running the tests

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosTableRepairs.java
```diff
@@ -211,25 +208,4 @@ AbstractPaxosRepair createRepair(DecoratedKey key, Ballot incompleteBallot, Cons
     {
         return PaxosRepair.create(consistency, key, incompleteBallot, table);
     }
-
-    private static final ConcurrentMap<TableId, PaxosTableRepairs> tableRepairsMap = new ConcurrentHashMap<>();
```
dcapwell, line 215: moved to `org.apache.cassandra.service.paxos.cleanup.PaxosRepairState`

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
```diff
@@ -339,6 +346,12 @@ public Set<Faults> apply(Cluster.Node node, Message<?> message)
                     // these messages are not resilent to ephemeral issues
                     case STATUS_REQ:
                     case STATUS_RSP:
+                    // paxos repair does not support faults and will cause a TIMEOUT error, failing the repair
```
dcapwell, line 349: I tested each verb and we lead to `timeout` in different code paths... some times the user is told its paxos related, some times not... logs are always clear though at least...

This patch does not try to make these messages resilient to dropped messages or concurrent retries, so opts-out so these messages will be happy path only.

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
```diff
@@ -654,6 +667,7 @@ static class Cluster
 
             // We run tests in an isolated JVM per class, so not cleaing up is safe... but if that assumption ever changes, will need to cleanup
             Stage.ANTI_ENTROPY.unsafeSetExecutor(orderedExecutor);
+            Stage.MISC.unsafeSetExecutor(orderedExecutor);
```
dcapwell, line 670: one line of code did this rather than `Stage.ANTI_ENTROPY`... not sure why we do MISC rather than ANTI_ENTROPY given this is repair... but figured best not to change it...  For this test they are the same executor so doesn't matter.

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
```diff
@@ -776,10 +790,46 @@ public void onFailure(InetAddressAndPort from, RequestFailureReason failureReaso
             }
         }
 
+        private static class CallbackKey
```
dcapwell, line 793: repair doesn't reuse messages/callbacks so didn't see that this logic was slightly different than `MessagingService`... so need to have both `(message_id, peer)` as the key and not just `message_id`, else the paxos messages clear the callback and hang forever (since it doesn't enqueue work the test actually fails right away).

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
```diff
@@ -952,6 +1003,18 @@ public EndpointState getEndpointStateForEndpoint(InetAddressAndPort ep)
...
+
+            @Override
+            public void notifyFailureDetector(Map<InetAddressAndPort, EndpointState> remoteEpStateMap)
+            {
+
+            }
+
+            @Override
+            public void applyStateLocally(Map<InetAddressAndPort, EndpointState> epStateMap)
+            {
```
dcapwell, line 1015: if we were testing paxos this would be wrong... I made gossip work with this simulation logic (see https://issues.apache.org/jira/browse/CASSANDRA-18917) so we could always leverage that if we cared... but we would need to figure out a bug found In that JIRA (a node Is added to gossip but a small subset never learns this)...

So for this patch, I went with no-op

src/java/org/apache/cassandra/gms/IGossiper.java
```diff
@@ -30,6 +31,8 @@ public interface IGossiper
 
     @Nullable
     EndpointState getEndpointStateForEndpoint(InetAddressAndPort ep);
+    void notifyFailureDetector(Map<InetAddressAndPort, EndpointState> remoteEpStateMap);
+    void applyStateLocally(Map<InetAddressAndPort, EndpointState> epStateMap);
```
ekaterinadimitrova2, line 35: These two can benefit from some javadoc

src/java/org/apache/cassandra/net/MessageDelivery.java
```diff
@@ -28,4 +29,8 @@ public interface MessageDelivery
     public <REQ, RSP> void sendWithCallback(Message<REQ> message, InetAddressAndPort to, RequestCallback<RSP> cb, ConnectionType specifyConnection);
     public <REQ, RSP> Future<Message<RSP>> sendWithResult(Message<REQ> message, InetAddressAndPort to);
     public <V> void respond(V response, Message<?> message);
+    public default void respondWithFailure(RequestFailureReason reason, Message<?> message)
```
ekaterinadimitrova2, line 32: Can you remove the public modifiers from this interface, please?

src/java/org/apache/cassandra/service/paxos/Paxos.java
ekaterinadimitrova2: Weird, I don't know why my IDE Intellij Idea emits a warning that the Config class is deprecated. Not after your patch, though; it's just observation.

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosRepairState.java
ekaterinadimitrova2: Incremental changes are ok. I guess it is also in 4.1 so not a regression? Let's open a ticket for the community so someone can pick it up

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosRepairState.java
```diff
@@ -0,0 +1,205 @@
...
+import org.apache.cassandra.schema.TableId;
+import org.apache.cassandra.service.paxos.Ballot;
+import org.apache.cassandra.service.paxos.PaxosState;
+import org.apache.cassandra.utils.Throwables;
+import org.apache.cassandra.utils.concurrent.IntrusiveStack;
+
+import static org.apache.cassandra.exceptions.RequestFailureReason.UNKNOWN;
+import static org.apache.cassandra.net.NoPayload.noPayload;
+
+public class PaxosRepairState
```
ekaterinadimitrova2, line 53: Please add Javadoc

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosRepairState.java
```diff
@@ -0,0 +1,205 @@
...
+{
+    private final SharedContext ctx;
+    private final AtomicReference<PendingCleanup> pendingCleanup = new AtomicReference();
```
ekaterinadimitrova2:
```suggestion
    private final AtomicReference<PendingCleanup> pendingCleanup = new AtomicReference<>();
```

src/java/org/apache/cassandra/service/paxos/PaxosRepair.java
ekaterinadimitrova2: I don't know about the IllegalArgumentException, probably some leftover

bdeggleston (approved):
+1

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
```diff
@@ -993,7 +1058,39 @@ private Node(UUID hostId, InetAddressAndPort addressAndPort, Collection<Token> t
                         validator.fail(e);
                     }
                 });
-                this.verbHandler = new RepairMessageVerbHandler(this);
+                this.verbHandler = new IVerbHandler()
```
ekaterinadimitrova2:
```suggestion
                this.verbHandler = new IVerbHandler<>()
```

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
ekaterinadimitrova2: Please add this info as a comment in the codebase

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosRepairState.java
ekaterinadimitrova2: Please add as a comment, it can be useful for someone

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosCleanupSession.java
```diff
@@ -168,13 +157,13 @@ private synchronized void fail(String message)
...
             return;
-        removeSession(this);
+        ctx.paxosRepairState().removeSession(this);
         tryFailure(new PaxosCleanupException(message));
         if (timeout != null)
             timeout.cancel(true);
     }
 
-    private synchronized void finish(InetAddressAndPort from, PaxosCleanupResponse finished)
+    public synchronized void finish(InetAddressAndPort from, PaxosCleanupResponse finished)
```
ekaterinadimitrova2: Isn't `protected `enough?

ekaterinadimitrova2 (approved):
Overall LGTM (though limited knowledge in that part of the codebase, I admit). Some nits left which can be addressed on commit

src/java/org/apache/cassandra/net/MessageDelivery.java
dcapwell: to be consistent with the rest of the file, I left `public` there...

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
dcapwell: guess you really want that change huh?  ^_^

src/java/org/apache/cassandra/service/paxos/cleanup/PaxosRepairState.java
dcapwell: > Tracks the state of paxos repair cleanup work

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
ekaterinadimitrova2: I think I hit a bug actually :D I added it once, GH complained there were new changes after a force push and I cannot add it. I refreshed and added it and then it popped up twice I guess, which I didn't notice.... edited now, sorry :D Not a big deal suggestion really

PR #3101, opened 2024-02-14
CASSANDRA-19393 make all cms-related commands to be subcommands of nodetool cms

nvharikrishna:
Now the commands looks like something like this:

```
nodetool cms describe
nodetool cms reconfigure
nodetool cms initialize
```

Running `nodetool cms` prints the following

```
NAME
        nodetool cms - Manage cluster metadata

SYNOPSIS
        nodetool cms
        nodetool [(-pw <password> | --password <password>)]
                [(-h <host> | --host <host>)] [(-u <username> | --username <username>)]
                [(-pwf <passwordFilePath> | --password-file <passwordFilePath>)]
                [(-pp | --print-port)] [(-p <port> | --port <port>)] cms describe
        nodetool [(-pw <password> | --password <password>)]
                [(-h <host> | --host <host>)] [(-u <username> | --username <username>)]
                [(-pwf <passwordFilePath> | --password-file <passwordFilePath>)]
                [(-pp | --print-port)] [(-p <port> | --port <port>)] cms initialize
                [(-i <ignored endpoints> | --ignore <ignored endpoints>)...]
        nodetool [(-pw <password> | --password <password>)]
                [(-h <host> | --host <host>)] [(-u <username> | --username <username>)]
                [(-pwf <passwordFilePath> | --password-file <passwordFilePath>)]
                [(-pp | --print-port)] [(-p <port> | --port <port>)] cms reconfigure
                [(-c | --cancel)] [(-r | --resume)] [--status]

OPTIONS
        -h <host>, --host <host>
            Node hostname or ip address

        -p <port>, --port <port>
            Remote jmx agent port number

        -pp, --print-port
            Operate in 4.0 mode with hosts disambiguated by port number

        -pw <password>, --password <password>
            Remote jmx agent password

        -pwf <passwordFilePath>, --password-file <passwordFilePath>
            Path to the JMX password file

        -u <username>, --username <username>
            Remote jmx agent username

COMMANDS
        With no arguments, Display help information

        describe
            Describe the current Cluster Metadata Service

        reconfigure
            Reconfigure replication factor of CMS

            With --cancel option, Cancels any in progress CMS reconfiguration

            With --resume option, Whether or not a previously interrupted
            sequence should be resumed

            With --status option, Poll status of the reconfigurecms command. All
            other flags and arguments are ignored when this one is used.

        initialize
            Upgrade from gossip and initialize CMS

            With --ignore option, Hosts to ignore due to them being down
```

The [CASSANDRA-19393](https://issues.apache.org/jira/browse/CASSANDRA-19393)


src/java/org/apache/cassandra/tools/nodetool/clustermetadata/DescribeCMS.java
```diff
@@ -16,29 +16,29 @@
...
-        System.out.printf("Cluster Metadata Service:%n");
-        System.out.printf("Members: %s%n", info.get("MEMBERS"));
-        System.out.printf("Is Member: %s%n", info.get("IS_MEMBER"));
-        System.out.printf("Service State: %s%n", info.get("SERVICE_STATE"));
-        System.out.printf("Is Migrating: %s%n", info.get("IS_MIGRATING"));
-        System.out.printf("Epoch: %s%n", info.get("EPOCH"));
-        System.out.printf("Local Pending Count: %s%n", info.get("LOCAL_PENDING"));
-        System.out.printf("Commits Paused: %s%n", info.get("COMMITS_PAUSED"));
-        System.out.printf("Replication factor: %s%n", info.get("REPLICATION_FACTOR"));
+        output.out.printf("Cluster Metadata Service:%n");
```
smiklosovic: this is a good change, I think this enable us in the future to test the output in unit tests as it is not writing it to System.out but to probe's output. All other nodetool commands avoid printing with System.out altogether as well so we close the gap.

src/java/org/apache/cassandra/tools/nodetool/clustermetadata/InitializeCMS.java
```diff
@@ -26,11 +26,12 @@
 import org.apache.cassandra.tools.NodeProbe;
 import org.apache.cassandra.tools.NodeTool;
 
-@Command(name = "initializecms", description = "Upgrade from gossip and initialize CMS")
+@Command(name = "initialize", description = "Upgrade from gossip and initialize CMS")
 public class InitializeCMS extends NodeTool.NodeToolCmd
 {
-    @Option(title = "ignored endpoints", name = {"-i", "--ignore"}, description = "Hosts to ignore due to them being down", required = false)
```
smiklosovic: for a reviewer - this is marked as redundant (in IDEA, at least) because "required" is false by default.

src/java/org/apache/cassandra/tools/NodeTool.java
```diff
@@ -262,6 +263,13 @@ public int execute(String... args)
                .withCommand(RepairAdmin.SummarizePendingCmd.class)
                .withCommand(RepairAdmin.SummarizeRepairedCmd.class);
 
+        builder.withGroup(CMSCmdGroupHelp.CMS_CMD_GROUP_NAME)
```
smiklosovic: I am not sure if we do not want to put all classes under "CMSAdmin" or something similar as it is done in `RepairAdmin` case but I am overall OK with this as well. I can not estimate if putting it all into one class is better than leaving it like this.

src/java/org/apache/cassandra/tools/nodetool/clustermetadata/CMSCmdGroupHelp.java
```diff
@@ -0,0 +1,42 @@
...
+
+import io.airlift.airline.Help;
+import org.apache.cassandra.tools.INodeProbeFactory;
+import org.apache.cassandra.tools.NodeTool;
+import org.apache.cassandra.tools.Output;
+
+/**
+ * Help class for Cluster Metadata Service nodetool command group.
+ */
+public class CMSCmdGroupHelp extends Help implements NodeTool.NodeToolCmdRunnable
```
smiklosovic: This will print help when `nodetool cms` is executed as it just expects some other sub-command to be specified. This is not the same as for the case of repair_admin, when `nodetool repair_admin` is executed, it will default to `ListCmd`. We might theoretically default to `cms describe` if only `nodetool cms` is run.

I think that having `nodetool cms` to default to `nodetool cms describe` is sensible to do so we might rewrite this as it is done for repair case. Having to write `nodetool cms describe` every single time instead of `nodetool cms` seems like a drag ... 

If we do it like suggested we will get rid of this class and `NodetoolCmdRunnable` in `NodeTool` can stay package protected.

nvharikrishna:
Chagnes merged as part of commit: https://github.com/apache/cassandra/commit/98ca5f8f1aab6659fdfd084bef38495bf50d35d6. Closing the pr.

PR #3102, opened 2024-02-14
CASSANDRA-19365: fix races in DecayingEstimatedHistogramReservoir

jakubzytka:
The patch introduces a DecayingBuckets class that encapsulates decayingBuckets and the now-immutable decayLandmark.
DecayingBuckets are never rescaled in place, so we can now ensure no updates with wrong weight and no snapshots of a half-rescaled reservoir.

The patch retains the original update performance.

The patch contains a disabled (`@Ignore`) test demonstrating the update vs rescale race condition in practice.

### DIFF
--- a/src/java/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoir.java
+++ b/src/java/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoir.java
@@ -150,7 +150,7 @@ public class DecayingEstimatedHistogramReservoir implements SnapshottingReservoi
     private final int distributionPrime;
 
     // decayingBuckets and buckets are one element longer than bucketOffsets -- the last element is values greater than the last offset
-    private final AtomicLongArray decayingBuckets;
+    private volatile DecayingBuckets decayingBuckets;
     private final AtomicLongArray buckets;
 
     public static final long HALF_TIME_IN_S = 60L;
@@ -158,7 +158,6 @@ public class DecayingEstimatedHistogramReservoir implements SnapshottingReservoi
     public static final long LANDMARK_RESET_INTERVAL_IN_NS = TimeUnit.MINUTES.toNanos(30L);
 
     private final AtomicBoolean rescaling = new AtomicBoolean(false);
-    private volatile long decayLandmark;
 
     // Wrapper around System.nanoTime() to simplify unit testing.
     private final MonotonicClock clock;
@@ -222,10 +221,9 @@ public class DecayingEstimatedHistogramReservoir implements SnapshottingReservoi
         }
 
         nStripes = stripes;
-        decayingBuckets = new AtomicLongArray((bucketOffsets.length + 1) * nStripes);
-        buckets = new AtomicLongArray((bucketOffsets.length + 1) * nStripes);
         this.clock = clock;
-        decayLandmark = clock.now();
+        buckets = new AtomicLongArray((bucketOffsets.length + 1) * nStripes);
+        decayingBuckets = new DecayingBuckets(clock.now());
         int distributionPrime = 1;
         for (int prime : DISTRIBUTION_PRIMES)
         {
@@ -250,7 +248,7 @@ public class DecayingEstimatedHistogramReservoir implements SnapshottingReservoi
 
         int index = findIndex(bucketOffsets, value);
 
-        updateBucket(decayingBuckets, index, Math.round(forwardDecayWeight(now)));
+        decayingBuckets.update(index, now);
         updateBucket(buckets, index, 1);
     }
 
@@ -286,11 +284,6 @@ public class DecayingEstimatedHistogramReservoir implements SnapshottingReservoi
 
         int firstCandidate = max(0, min(bucketOffsets.length - 1, ((int) fastLog12(value)) - offset));
         return value <= bucketOffsets[firstCandidate] ? firstCandidate : firstCandidate + 1;
-    }
-
-    private double forwardDecayWeight(long now)
-    {
-        return Math.exp(TimeUnit.NANOSECONDS.toSeconds(now - decayLandmark) / MEAN_LIFETIME_IN_S);
     }
 
     /**
@@ -339,15 +332,14 @@ public class DecayingEstimatedHistogramReservoir implements SnapshottingReservoi
     @VisibleForTesting
     boolean isOverflowed()
     {
-        return bucketValue(bucketOffsets.length, true) > 0;
+        return bucketValue(bucketOffsets.length, decayingBuckets.buckets) > 0;
     }
 
-    private long bucketValue(int index, boolean withDecay)
+    private long bucketValue(int index, AtomicLongArray buckets)
     {
         long val = 0;
-        AtomicLongArray bs = withDecay ? decayingBuckets : buckets;
         for (int stripe = 0; stripe < nStripes; stripe++)
-            val += bs.get(stripedIndex(index, stripe));
+            val += buckets.get(stripedIndex(index, stripe));
 
         return val;
     }
@@ -355,7 +347,7 @@ public class DecayingEstimatedHistogramReservoir implements SnapshottingReservoi
     @VisibleForTesting
     long stripedBucketValue(int i, boolean withDecay)
     {
-        return withDecay ? decayingBuckets.get(i) : buckets.get(i);
+        return withDecay ? decayingBuckets.buckets.get(i) : buckets.get(i);
     }
 
     private void rescaleIfNeeded()
@@ -371,42 +363,50 @@ public class DecayingEstimatedHistogramReservoir implements SnapshottingReservoi
             {
                 try
                 {
-                    rescale(now);
+                    decayingBuckets = rescale(now);
                 }
                 finally
                 {
-                    decayLandmark = now;
                     rescaling.set(false);
                 }
             }
         }
     }
 
-    private void rescale(long now)
+    /**
+     * rescale races with update;
+     * If rescale is called at the same time as an update, and an update falls into an already
+     * processed bucket this update will be lost. We accept that for the sake of update performance.
+     * OTOH, if the update falls into a bucket that is yet to be processed, it will be added
+     * with a weight that is consistent with the rescaleFactor used by rescale, so the new rescaled
+     * buckets will all have values consistent with the new decay landmark.
+     */
+    private DecayingBuckets rescale(long now)
     {
-        // despite striping its safe to rescale each bucket individually
-        final double rescaleFactor = forwardDecayWeight(now);
-        for (int i = 0; i < decayingBuckets.length(); i++)
+        final double rescaleFactor = decayingBuckets.forwardDecayWeight(now);
+        DecayingBuckets newBuckets = new DecayingBuckets(now);
+        for (int i = 0; i < decayingBuckets.buckets.length(); i++)
         {
-            long newValue = Math.round(decayingBuckets.get(i) / rescaleFactor);
-            decayingBuckets.set(i, newValue);
+            long newValue = Math.round(decayingBuckets.buckets.get(i) / rescaleFactor);
+            newBuckets.buckets.set(i, newValue);
         }
+        return newBuckets;
     }
 
     private boolean needRescale(long now)
     {
-        return (now - decayLandmark) > LANDMARK_RESET_INTERVAL_IN_NS;
+        return (now - decayingBuckets.decayLandmark) > LANDMARK_RESET_INTERVAL_IN_NS;
     }
 
     @VisibleForTesting
     public void clear()
     {
-        final int bucketCount = decayingBuckets.length();
+        final int bucketCount = buckets.length();
         for (int i = 0; i < bucketCount; i++)
         {
-            decayingBuckets.set(i, 0L);
             buckets.set(i, 0L);
         }
+        decayingBuckets = new DecayingBuckets(decayingBuckets.decayLandmark);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/java/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoir.java
```diff
@@ -250,7 +248,7 @@ public void update(long value)
 
         int index = findIndex(bucketOffsets, value);
 
-        updateBucket(decayingBuckets, index, Math.round(forwardDecayWeight(now)));
+        decayingBuckets.update(index, now);
```
jakubzytka: I'm on the fence with this one.
On the one hand, I don't want to do a direct `decayingBuckets.buckets` update because the weight may be given incorrectly.
On the other, it looks inconsistent with updating regular buckets, and there is no separation of concerns in DEHR anyway.

pkolaczk: Well, personally I don't like the lack of symmetry here as well, but, well, decayingBuckets does more than regular buckets and it had to be extracted in order to be able to set the reference atomically. Extracting this to a separate method makes the code more readable and less error prone.

pkolaczk (approved):
LGTM. The change doesn't incur performance penalty and should fix the correctness problem. 
Skipping a very tiny fraction of data is much smaller problem than inserting vastly incorrect data into histogram due to a race.

jakubzytka:
> LGTM. The change doesn't incur performance penalty and should fix the correctness problem. Skipping a very tiny fraction of data is much smaller problem than inserting vastly incorrect data into histogram due to a race.

WRT performance. The results of `LatencyTrackingBench` have a lot of uncertainty, but if anything, they show a possible improvement in `update` throughput. It makes sense, as the number of volatile reads is reduced to 1 (using `decayLandmark` is no longer a volatile read)

src/java/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoir.java
```diff
@@ -430,19 +430,64 @@ public void rebase(EstimatedHistogramReservoirSnapshot snapshot)
...
+     * <p/>
+     * DecayingBuckets are not suitable for thread-safe rescaling in place. Instead,
+     * a new instance of DecayingBuckets should be created when rescaling is needed.
+     * <p/>
+     * This class does NOT provide thread safety guarantees for updates vs rescale
+     * As a matter of fact, there is a known race condition between update and rescale
+     * (see {@link #rescale(long)}).
+     * <p/>
+     */
+    private class DecayingBuckets
```
maedhroz, line 471: nit: Could always throw a `@NotThreadSafe` annotation on this.

src/java/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoir.java
```diff
@@ -371,42 +363,50 @@ private void rescaleIfNeeded(long now)
...
     public void clear()
     {
-        final int bucketCount = decayingBuckets.length();
+        final int bucketCount = buckets.length();
         for (int i = 0; i < bucketCount; i++)
         {
-            decayingBuckets.set(i, 0L);
             buckets.set(i, 0L);
         }
+        decayingBuckets = new DecayingBuckets(decayingBuckets.decayLandmark);
```
maedhroz, line 409: I guess you could wrap this up in some crazy to avoid the landmark going backward in time...

```
while (true)
        {
            if (rescaling.compareAndSet(false, true))
            {
                try
                {
                    long landmark = decayingBuckets.decayLandmark;
                    decayingBuckets = new DecayingBuckets(landmark);
                }
                finally
                {
                    rescaling.set(false);
                }
                break;
            }
        }
```
...but it's only used for tests...we don't really care :D

maedhroz (approved):

src/java/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoir.java
Mmuzaf: Let's use the `AtomicReferenceFieldUpdater` to update the `decayingBuckets` (instead of the `rescaling` field) for both the clear and rescaleIfNeeded methods. Even if the former is only used in tests now, the changes should be consistent making any further changes safe without worrying about concurrency.

src/java/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoir.java
```diff
@@ -649,16 +694,20 @@ public EstimatedHistogramReservoirSnapshot(DecayingEstimatedHistogramReservoir r
             super(reservoir);
             
             int length = reservoir.size();
-            double rescaleFactor = reservoir.forwardDecayWeight(reservoir.clock.now());
-
             this.values = new long[length];
-            this.snapshotLandmark = reservoir.decayLandmark;
+
```
Mmuzaf, line 698: Unnecessary empty line?

src/java/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoir.java
```diff
@@ -430,19 +430,64 @@ public void rebase(EstimatedHistogramReservoirSnapshot snapshot)
...
+     * As a matter of fact, there is a known race condition between update and rescale
+     * (see {@link #rescale(long)}).
+     * <p/>
+     */
+    private class DecayingBuckets
+    {
+        private final long decayLandmark;
+        private final AtomicLongArray buckets;
 
+        public DecayingBuckets(long decayLandmark)
```
Mmuzaf, line 476: I would make the `DecayingBuckets` class static and pass `(bucketOffsets.length + 1) * nStripes` as a parameter to it to outline that `buckets` and `DecayingBuckets` have the same length regardless of where they are used. 

In other cases, I must fall into the `DecayingBuckets` class to understand what it does.

src/java/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoir.java
```diff
@@ -430,19 +430,64 @@ public void rebase(EstimatedHistogramReservoirSnapshot snapshot)
...
+            newDecayingBuckets.buckets.set(stripedIndex(i, 0), snapshot.decayingBuckets[i]);
             buckets.set(stripedIndex(i, 0), snapshot.values[i]);
             for (int stripe = 1; stripe < nStripes; stripe++)
             {
-                decayingBuckets.set(stripedIndex(i, stripe), 0);
+                newDecayingBuckets.buckets.set(stripedIndex(i, stripe), 0);
                 buckets.set(stripedIndex(i, stripe), 0);
             }
         }
+        decayingBuckets = newDecayingBuckets;
```
Mmuzaf, line 445: I think it's the same as here: https://github.com/apache/cassandra/pull/3102/files#r1713929010

src/java/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoir.java
```diff
@@ -371,42 +363,50 @@ private void rescaleIfNeeded(long now)
...
+     * OTOH, if the update falls into a bucket that is yet to be processed, it will be added
+     * with a weight that is consistent with the rescaleFactor used by rescale, so the new rescaled
+     * buckets will all have values consistent with the new decay landmark.
+     */
+    private DecayingBuckets rescale(long now)
     {
-        // despite striping its safe to rescale each bucket individually
-        final double rescaleFactor = forwardDecayWeight(now);
-        for (int i = 0; i < decayingBuckets.length(); i++)
+        final double rescaleFactor = decayingBuckets.forwardDecayWeight(now);
```
Mmuzaf, line 386: We are reading the `decayingBuckets` three times here and if the `rebase` method is called, this could potentially cause a problem. So I would save the decayingBuckets to a local ref and read it.

test/unit/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoirTest.java
```diff
@@ -604,6 +614,115 @@ public void testSize()
...
+         * The load has 95% of 42, and 5% of the time it's 1109. Despite that the histogram may be convinced for a long time
+         * that p99 is 42 or that p50 is 1109.
+         * The reason may be seen in the snapshot dump, where after rescale the bucket values may get
+         * very big due to the race condition and too big weight of the inserted samples.
+         * The values were picked to match bucket boundaries, but that's only for aesthetics.
+         * <p>
+         * In production the rescale happens every 30 minutes. In this test time flies roughly 1000 times faster to
+         * hit the condition in a reasonable time.
+         */
+        @Ignore("This test exposes a specific CASSANDRA-19365 race condition; it doesn't make sense to run it in CI")
```
Mmuzaf, line 632: I don't think adding tests that have been ignored by default is good practice. Here is what we can do:
- Add a new test to the `burn` package like ConnectionBurnTest and test that the race between update and rescale doesn't cause problems with the buckets flooding (only some operations are missed)
- Remove the `@Ignore` here and do the same here for a period of 30-60 seconds and check the state at the end as we do now. The test shouldn't be flaky for 10k runs.

We can run on the CircleCI this test thousands of times, to verify that the race doesn't cause any problems.

test/unit/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoirTest.java
```diff
@@ -604,6 +614,115 @@ public void testSize()
...
+            TestClock clock = new TestClock();
+
+            DecayingEstimatedHistogramReservoir histogram = new DecayingEstimatedHistogramReservoir(clock);
+
+            ExecutorService executors = Executors.newFixedThreadPool(2 + NUM_UPDATE_THREADS);
+            // NUM_UPDATE_THREADS threads updating the histogram
+            for (int i = 0; i < NUM_UPDATE_THREADS; i++)
+            {
+                executors.submit(() -> {
+                    while (!stop.get())
```
Mmuzaf, line 649: It would be a good idea to check `isInterrupted` as well.

test/unit/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoirTest.java
```diff
@@ -604,6 +614,115 @@ public void testSize()
...
+                // how many times in a row the percentiles may be invalid before we fail the test
+                int tooManySuspiciousPercentilesThreshold = 5; // 5 translates to 500ms * 1000 speedup = 500s = 8m20s;
+
+                for (int i = 0; i < maxTestDurationMillis / iterationDelayMillis; i++)
+                {
+                    Uninterruptibles.sleepUninterruptibly(iterationDelayMillis, MILLISECONDS);
+                    Snapshot snapshot = histogram.getSnapshot();
+                    double p99 = snapshot.getValue(0.99);
+                    double p50 = snapshot.getValue(0.50);
+                    logger.info("{} p99: {}", clock.now() / 1_000_000, p99);
```
Mmuzaf, line 695: This is unnecessary logging I think, I would print enough just the threads still running every 5-10 sec or so.

test/unit/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoirTest.java
```diff
@@ -604,6 +614,115 @@ public void testSize()
...
+                    }
+                    else
+                    {
+                        consecutiveInvalidPercentiles = 0;
+                    }
+                }
+                stop.set(true);
+            });
+            executors.shutdown();
+            executors.awaitTermination(300, SECONDS);
```
Mmuzaf, line 722: I think it's too high, 1.5 mins should be enough? Let's also move it as constant.

test/unit/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoirTest.java
```diff
@@ -604,6 +614,115 @@ public void testSize()
...
+                    }
+                });
+            }
+            // clock update thread
+            executors.submit(() -> {
+                while (!stop.get())
+                {
+                    Uninterruptibles.sleepUninterruptibly(1, TimeUnit.MILLISECONDS);
+                    // x1000 speedup so that we hit rescale interval every 30 minutes / 1000 = 1.8s
+                    clock.addMillis(1000);
```
Mmuzaf, line 675: It's better to use this value in relation to the `LANDMARK_RESET_INTERVAL_IN_NS` itself.

src/java/org/apache/cassandra/metrics/DecayingEstimatedHistogramReservoir.java
```diff
@@ -371,42 +363,50 @@ private void rescaleIfNeeded(long now)
             {
                 try
                 {
-                    rescale(now);
+                    decayingBuckets = rescale(now);
```
Mmuzaf, line 366: It seems that both `needRescale` and `rescale` should be calculated on the same object reference. Currently it doesn't look like the same condition is being met when the `rescale` method is run after checking `needRescale`, or am I missing something?

The example:
```
private static final AtomicReferenceFieldUpdater<DecayingEstimatedHistogramReservoir, DecayingBuckets> decayingBucketsUpdater =
    AtomicReferenceFieldUpdater.newUpdater(DecayingEstimatedHistogramReservoir.class, DecayingBuckets.class, "decayingBuckets");

private void rescaleIfNeeded(long now)
{
    DecayingBuckets buckets = decayingBuckets;
    while (now - buckets.decayLandmark > LANDMARK_RESET_INTERVAL_IN_NS)
    {
        double rescaleFactor = buckets.forwardDecayWeight(now);
        DecayingBuckets newBuckets = new DecayingBuckets(now);
        for (int i = 0; i < buckets.buckets.length(); i++)
            newBuckets.buckets.set(i, Math.round(buckets.buckets.get(i) / rescaleFactor));

        boolean success = decayingBucketsUpdater.compareAndSet(this, buckets, newBuckets);
        if (success)
            return;
        buckets = decayingBuckets;
    }
}
```

Mmuzaf (changes requested):
@jakubzytka I've left some comments, please take a look

Mmuzaf:
Can we also add a benchmark that emulates rescaling during updates?
e.g. `LatencyTrackingBench`

pmcfadin:
Closing as [CASSANDRA-19365](https://issues.apache.org/jira/browse/CASSANDRA-19365) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #3107, opened 2024-02-15
Fix the duration of speculative read delay

jakubzytka:
Speculative reads will now be triggered after sampleReadLatency past the replica reads are issued instead of past queryStartNanoTime. It's because sampleReadLatency measures the actual execution of read request whereas queryStartNanoTime may, after CASSANDRA-19215, be measured before wait in the NativeTransportRequests queue.

Patch by Jakub Zytka; reviewed by TBD for CASSANDRA-19330

### DIFF
--- a/src/java/org/apache/cassandra/service/reads/ReadCallback.java
+++ b/src/java/org/apache/cassandra/service/reads/ReadCallback.java
@@ -51,10 +51,8 @@ import org.apache.cassandra.tracing.Tracing;
 import org.apache.cassandra.utils.concurrent.UncheckedInterruptedException;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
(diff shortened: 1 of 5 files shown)


src/java/org/apache/cassandra/service/reads/ReadCallback.java
```diff
@@ -95,12 +93,15 @@ protected P replicaPlan()
         return replicaPlan.get();
     }
 
-    public boolean await(long timePastStart, TimeUnit unit)
+    /**
+     * await until the condition is signalled but not longer than timePastStart since startTimeNanos
+     */
+    public boolean awaitFrom(long startTimeNanos, long timePastStart, TimeUnit unit)
```
jakubzytka, line 99: The original `await` name was confusing. It's typically associated with waiting a specified amount of time, which was not the case in the original code.

PR #3108, opened 2024-02-15
CASSANDRA-19405: (Accord) When C* is saving the cache state we can check if we can evict which promotes from Saving to Loaded, allowing removal before the save completes

dcapwell:


src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -229,6 +234,8 @@ static Factory factory(AccordJournal journal, AccordStateCacheMetrics cacheMetri
 
     private void loadRangesToCommands()
     {
+        if (!CassandraRelevantProperties.TEST_ACCORD_STORE_LOAD_RANGES_ENABLED.getBoolean())
```
belliottsmith: static import?

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
```diff
@@ -93,8 +94,12 @@ public class AccordCommandStore extends CommandStore implements CacheSize
 {
     private static final Logger logger = LoggerFactory.getLogger(AccordCommandStore.class);
 
+    private static final boolean CHECK_THREADS = CassandraRelevantProperties.TEST_ACCORD_STORE_THREAD_CHECKS_ENABLED.getBoolean();
```
belliottsmith, line 98: Is it beneficial to optional disable this?

We should probably optimise this to not need to submit a runnable, but that is quite easy.

dcapwell: We have 2 different checks: `inStore` and `notInStore`... so if I have `getThreadId` return the main test thread then `notInStore` fails

src/java/org/apache/cassandra/service/accord/AccordStateCache.java
```diff
@@ -141,36 +141,45 @@ private void maybeEvictSomeNodes()
...
             AccordCachingState<?, ?> node = iter.next();
-            checkState(node.references == 0);
-
-            if (!node.canEvict())
-                continue;
-            /*
-             * TODO (expected, efficiency):
-             *    can this be reworked so we're not skipping unevictable nodes everytime we try to evict?
-             */
-            Status status = node.status(); // status() call completes (if completeable)
```
dcapwell, line 152: this is the bug that the JIRA talks about, this promotes `Saving` to `Loaded`, and since we validate on evict this fails

test/unit/org/apache/cassandra/concurrent/SimulatedExecutorFactory.java
```diff
@@ -153,9 +180,92 @@ public Thread startThread(String name, Runnable runnable, InfiniteLoopExecutor.D
     }
 
     @Override
-    public Interruptible infiniteLoop(String name, Interruptible.Task task, InfiniteLoopExecutor.SimulatorSafe simulatorSafe, InfiniteLoopExecutor.Daemon daemon, InfiniteLoopExecutor.Interrupts interrupts)
+    public Interruptible infiniteLoop(String name,
```
dcapwell, line 183: I think I added this for journal before switching to mocking it out... Think I can remove to limit the scope of change

src/java/org/apache/cassandra/service/accord/AccordCachingState.java
```diff
@@ -605,6 +605,12 @@ public boolean isCompleteable()
             return isDone();
         }
 
+        @Override
+        public V get()
```
dcapwell, line 609: added for tests so we can see what we are saving to see if we should attempt to evict it.  Main eviction logic doesn't touch it while in this state

test/unit/org/apache/cassandra/service/accord/AccordCommandStoreFuzzTest.java
```diff
@@ -0,0 +1,457 @@
...
+    {
+        if (intTbl != null)
+            return;
+        createKeyspace("CREATE KEYSPACE test WITH replication={ 'class' : 'SimpleStrategy', 'replication_factor' : 2 }");
+        createTable("test", "CREATE TABLE test.tbl1 (pk int PRIMARY KEY, value int)");
+        intTbl = Schema.instance.getTableMetadata("test", "tbl1");
+        AddAccordTable.addTable(intTbl.id);
+
+        createTable("test", "CREATE TABLE test.tbl2 (pk vector<bigint, 2> PRIMARY KEY, value int)");
+        reverseTokenTbl = Schema.instance.getTableMetadata("test", "tbl2");
```
dcapwell: ill remove this... this is what the range tests use.  The range tests work with random tokens so can't use `int` as the primary key...

test/unit/org/apache/cassandra/service/accord/AccordCommandStoreFuzzTest.java
```diff
@@ -0,0 +1,457 @@
...
+    }
+
+    private static TableMetadata intTbl, reverseTokenTbl;
+    private static Node.Id nodeId;
+
+    @BeforeClass
+    public static void setUpClass()
+    {
+        CQLTester.setUpClass();
+        DatabaseDescriptor.setIncrementalBackupsEnabled(false);
```
dcapwell, line 110: let me add a code doc explaining... we snapshot `system.info` every time we truncate, which is 95% the cost of this test when this is enabled...

test/unit/org/apache/cassandra/service/accord/AccordCommandStoreFuzzTest.java
```diff
@@ -0,0 +1,457 @@
...
+                                   List<TxnId> rangeConflicts, Ranges ranges,
+                                   PreAccept.PreAcceptOk success)
+    {
+        if (rangeConflicts.isEmpty())
+        {
+            Assertions.assertThat(success.deps.rangeDeps.isEmpty()).describedAs("rangeDeps was not empty").isTrue();
+        }
+        else
+        {
+            Assertions.assertThat(success.deps.rangeDeps.rangeCount()).describedAs("Expected ranges size").isEqualTo(ranges.size());
```
dcapwell, line 247: these asserts are not touched with these tests, they are used in the range tests which are left out of this patch due to range not actually working right now... my range patch gets this all passing.

test/unit/org/apache/cassandra/service/accord/AccordCommandStoreFuzzTest.java
```diff
@@ -0,0 +1,457 @@
...
+                    var now = Timestamp.fromValues(epoch(), now(), nodeId);
+                    if (now.compareTo(atLeast) < 0)
+                        throw new UnsupportedOperationException();
+                    return now;
+                }
+            };
+
+            this.store = new AccordCommandStore(0,
+                                                timeService,
+                                                new TestAgent.RethrowAgent(),
```
dcapwell, line 341: since we catch all exceptions in executors, Rethrowing is fine, don't need `errors` to propagate to the agent here

src/java/org/apache/cassandra/service/accord/AccordCommandStore.java
belliottsmith: I mean that the `Executor` could set it on startup, and expose it as a final field

belliottsmith: Regarding disabling the checks, might be nice to only disable for notInStore if that's what we care about? Or maybe there's some other cleaner way that doesn't affect other paths.

src/java/org/apache/cassandra/service/accord/AccordStateCache.java
belliottsmith: Could you explain this a bit more - we've only affected the `LOADING` state, not the `SAVING` state. We seem here to only be preventing a `LOADING` node from being evicted before being `LOADED`, which is probably a perfectly sensible thing to do. But I'm not sure how to map to your explanation.

I can imagine that by evicting a `LOADING` command we might have a handle to the entry somewhere that is then updated and should be saved but isn't. But that means we've got bad reference counting or something else going on as well?

dcapwell: > we've only affected the LOADING state, not the SAVING state

We are talking about both.  `org.apache.cassandra.service.accord.AccordCachingState#status`

```
public Status status()
    {
        return complete().status();
    }

State<K, V> complete()
    {
        return state.isCompleteable() ? state(state.complete()) : state;
    }
```

`Loading` and `Saving` are the only 2 things completable, so calling this method on either of them promotes them to the next state.


> We seem here to only be preventing a LOADING node from being evicted before being LOADED, which is probably a perfectly sensible thing to do. But I'm not sure how to map to your explanation.

In the `Loading` case we promote to `Loaded` with a `null` value... when we evict `Loaded` we double check that the value in-memory matches disk (see `org.apache.cassandra.service.accord.AccordStateCache#evict`), this check will fail if the data is on disk as `null != data`

> I can imagine that by evicting a LOADING command we might have a handle to the entry somewhere that is then updated and should be saved but isn't. But that means we've got bad reference counting or something else going on as well?

If anything has a reference to this loaded command we have a bad reference count, so a different bug.  If linked and `ref == 0` then nothing should know about this node, so it should be safe to purge.

Now, that doesn't mean we stopped the load, so that will complete and update the `Loading` object, but since nothing holds reference anymore that gets garbage collected away.

dcapwell:
closing as speaking with @belliottsmith turns out that the error I saw was a test issue... We are testing at the store level and the `CommandStores` level is where we update journal...

PR #3109, opened 2024-02-15
CASSANDRA-18951: Add option for MutualTlsAuthenticator to restrict the certificate age

frankgh:
In this commit, we introduce a new parameter option for the `MutualTlsAuthenticator`. The new option is `max_certificate_age`, and it can be configured as a duration configuration parameter as defined by the `DurationSpec` (see CASSANDRA-15234). The resolution for this new property is minutes.

When specified, the certificate validation implementation will take that information and reject certificates that are older than the maximum allowed certificate age, translating into a rejection from the authenticating user.

patch by Francisco Guerrero; reviewed by TBD for CASSANDRA-18951

### DIFF
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -170,11 +170,9 @@ authenticator:
 # MutualTlsAuthenticator can be configured using the following configuration. One can add their own validator
 # which implements MutualTlsCertificateValidator class and provide logic for extracting identity out of certificates
 # and validating certificates.
-# When the max_certificate_age parameter is not set, the age validation will have no effect.
 #  class_name : org.apache.cassandra.auth.MutualTlsAuthenticator
 #  parameters :
 #    validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
-#    max_certificate_age: 365d
 
 # Authorization backend, implementing IAuthorizer; used to limit access/provide permissions
 # Out of the box, Cassandra provides org.apache.cassandra.auth.{AllowAllAuthorizer,
@@ -966,13 +964,11 @@ listen_address: localhost
 # MutualTlsInternodeAuthenticator can be configured using the following configuration.One can add their own validator
 # which implements MutualTlsCertificateValidator class and provide logic for extracting identity out of certificates
 # and validating certificates.
-# When the max_certificate_age parameter is not set, the age validation will have no effect.
 #  class_name : org.apache.cassandra.auth.MutualTlsInternodeAuthenticator
 #  parameters :
 #    validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
 #    trusted_peer_identities: "spiffe1,spiffe2"
 #    node_identity: "spiffe1"
-#    max_certificate_age: 365d
 # Whether to start the native transport server.
 # The address on which the native transport is bound is defined by rpc_address.
 start_native_transport: true
@@ -1600,6 +1596,13 @@ server_encryption_options:
   #   TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA, TLS_RSA_WITH_AES_128_GCM_SHA256, TLS_RSA_WITH_AES_128_CBC_SHA,
   #   TLS_RSA_WITH_AES_256_CBC_SHA
   # ]
+  # Optional setting to define the maximum allowed validity period of the client certificate used for the internode
+  # inbound connections. For example, if the specified max_certificate_validity_period is 30 days and the client
+  # uses a certificate that is issued for more than 30 days, the connection will be rejected.
+  # max_certificate_validity_period: 365d
+  # Optional setting that defines a warning threshold. When the threshold is exceeded for the internode certificate
+  # validity period, warnings with information about the certificate expiration will be logged.
+  # certificate_validity_warn_threshold: 10d
 
 # Configure client-to-server encryption.
 #
@@ -1646,6 +1649,15 @@ client_encryption_options:
   #   TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA, TLS_RSA_WITH_AES_128_GCM_SHA256, TLS_RSA_WITH_AES_128_CBC_SHA,
   #   TLS_RSA_WITH_AES_256_CBC_SHA
   # ]
+  # Optional setting to define the maximum validity period of the client certificate allowed to establish
+  # connections to the server. For example, if max_certificate_validity_period is configured for 10 days,
+  # and a client attempts to authenticate with a certificate with a longer validity period (say 30 days),
+  # then the connection will be rejected.
+  # max_certificate_validity_period: 365d
+  # Optional setting that defines a warning threshold. When the threshold is exceeded for the client certificate
+  # validity, warnings with information about the certificate expiration will be logged. Additionally, client
+  # warnings will be reported during the session establishment.
+  # certificate_validity_warn_threshold: 10d
 
 # internode_compression controls whether traffic between nodes is
 # compressed.
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -173,6 +173,7 @@ authenticator:
 #  class_name : org.apache.cassandra.auth.MutualTlsAuthenticator
 #  parameters :
 #    validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
+#    max_certificate_age: 365d
 
 # Authorization backend, implementing IAuthorizer; used to limit access/provide permissions
 # Out of the box, Cassandra provides org.apache.cassandra.auth.{AllowAllAuthorizer,
@@ -969,6 +970,7 @@ listen_address: localhost
 #    validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
 #    trusted_peer_identities: "spiffe1,spiffe2"
 #    node_identity: "spiffe1"
+#    max_certificate_age: 365d
 # Whether to start the native transport server.
 # The address on which the native transport is bound is defined by rpc_address.
 start_native_transport: true
--- a/conf/cassandra.yaml
+++ b/conf/cassandra.yaml
@@ -170,6 +170,7 @@ authenticator:
 # MutualTlsAuthenticator can be configured using the following configuration. One can add their own validator
 # which implements MutualTlsCertificateValidator class and provide logic for extracting identity out of certificates
 # and validating certificates.
+# When the max_certificate_age parameter is not set, the age validation will have no effect.
 #  class_name : org.apache.cassandra.auth.MutualTlsAuthenticator
 #  parameters :
 #    validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
@@ -965,6 +966,7 @@ listen_address: localhost
 # MutualTlsInternodeAuthenticator can be configured using the following configuration.One can add their own validator
 # which implements MutualTlsCertificateValidator class and provide logic for extracting identity out of certificates
 # and validating certificates.
+# When the max_certificate_age parameter is not set, the age validation will have no effect.
 #  class_name : org.apache.cassandra.auth.MutualTlsInternodeAuthenticator
 #  parameters :
 #    validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
--- a/src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
+++ b/src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
@@ -39,6 +39,7 @@ import org.apache.cassandra.config.DurationSpec;
 import org.apache.cassandra.config.ParameterizedClass;
 import org.apache.cassandra.exceptions.AuthenticationException;
 import org.apache.cassandra.exceptions.ConfigurationException;
+import org.apache.cassandra.metrics.MutualTlsMetrics;
 import org.apache.cassandra.schema.SchemaConstants;
 import org.apache.cassandra.utils.NoSpamLogger;
 
@@ -58,10 +59,10 @@ import static org.apache.cassandra.config.EncryptionOptions.ClientAuth.REQUIRED;
  * validator and configure it in cassandra.yaml.Below is an example on how to configure validator.
  * note that this example uses SPIFFE based validator, It could be any other validator with any defined identifier format.
  *
- * <p>Optionally, the authenticator can be configured to restrict the age of the client certificates. This allows for
- * better server-side controls for authentication. In some cases, clients can provide certificates that expire multiple
- * months/years after the certificate was issued. For those use cases, it is desirable to reject the certificate if
- * the expiration date is too far away in the future.
+ * <p>Optionally, the authenticator can be configured to restrict the validity period of the client certificates.
+ * This allows for better server-side controls for authentication. In some cases, clients can provide certificates
+ * that expire multiple months/years after the certificate was issued. For those use cases, it is desirable to
+ * reject the certificate if the validity period is too big (i.e. certificates issued for 10 years).
  *
  * <p>Example:
  * <pre>
@@ -69,7 +70,6 @@ import static org.apache.cassandra.config.EncryptionOptions.ClientAuth.REQUIRED;
  *   class_name : org.apache.cassandra.auth.MutualTlsAuthenticator
  *   parameters :
  *     validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
- *     max_certificate_age: 5d
  * </pre>
  */
 public class MutualTlsAuthenticator implements IAuthenticator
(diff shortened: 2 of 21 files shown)


src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
```diff
@@ -87,6 +101,18 @@ public MutualTlsAuthenticator(Map<String, String> parameters)
         }
         certificateValidator = ParameterizedClass.newInstance(new ParameterizedClass(certificateValidatorClassName),
                                                               Arrays.asList("", AuthConfig.class.getPackage().getName()));
+
+        String maxCertificateAgeConfig = parameters.get(MAX_CERTIFICATE_AGE_CONFIG);
```
tolbertam: One thing i'm wondering is whether the parameter should be on the authenticator or in the validator.

I noticed that the certificate age itself is evaluated in `SpiffeCertificateValidator`.  Since we use `ParameterizedClass` on `validator_class_name`, technically we should be able to pass parameters to it.

If you pass the parameter to the validator, then you don't need logic in both `MutualTlsAuthenticator` and `MutualTlsInternodeAuthenticator` for parsing out the certificate age.

tolbertam: On the other hand there's an argument to move the evaluation completely out of the validator, so all `MutualTlsCertificateValidator` implementations don't have to implement logic to validate the certificate age.

src/java/org/apache/cassandra/auth/SpiffeCertificateValidator.java
```diff
@@ -31,27 +32,45 @@
...
+     */
     @Override
-    public boolean isValidCertificate(Certificate[] clientCertificateChain)
+    public boolean isValidCertificate(Certificate[] clientCertificateChain, int maxCertificateAgeMinutes) throws AuthenticationException
     {
+        X509Certificate[] castedCerts = castCertsToX509(clientCertificateChain);
+        int certificateAgeInMinutes = getCertificateAgeInMinutes(castedCerts[0]);
+        if (certificateAgeInMinutes > maxCertificateAgeMinutes)
+        {
+            String errorMessage = String.format("The age of the provided certificate exceeds the maximum allowed age of %d minutes",
```
tolbertam: Is there a way to introspect what resolution the given duration is in and report it in that way?  e.g. if I configure `max_certificate_age: 5d` , could we have the message be The age of the provided certificate exceeds the maximum allowed age of 5d?.

It would also be useful to include the age of the certificate in the message.

test/unit/org/apache/cassandra/auth/MutualTlsAuthenticatorTest.java
```diff
@@ -86,7 +90,7 @@ public static void setup()
     @After
     public void after() throws IOException, TimeoutException
     {
-        MBeanWrapper.instance.unregisterMBean("org.apache.cassandra.auth:type=IdentitiesCache");
+        MBeanWrapper.instance.unregisterMBean("org.apache.cassandra.auth:type=IdentitiesCache", MBeanWrapper.OnException.IGNORE);
```
tolbertam: I was wondering why this was needed, is it because `testValidMaxCertificateAgeConfiguration` unregisters after each pair, so at the end of the test the bean is no longer registered?

frankgh: yeah, this prevents the `after` method to fail because the bean is not registered

src/java/org/apache/cassandra/auth/SpiffeCertificateValidator.java
frankgh: Yeah, makes sense, I will make the change

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
frankgh: yeah good point, I think there's more value in moving the validation out of the validator.

src/java/org/apache/cassandra/auth/SpiffeCertificateValidator.java
tolbertam: Looking at `DurationSpec` code, it looks like this isn't easily achievable, and I think there is precedence in C* for logging in the resolved unit rather than the given one, so probably ok to remain this way.

conf/cassandra.yaml
```diff
@@ -158,6 +158,7 @@ authenticator:
 #  class_name : org.apache.cassandra.auth.MutualTlsAuthenticator
 #  parameters :
 #    validator_class_name: org.apache.cassandra.auth.SpiffeCertificateValidator
+#    max_certificate_age: 365d
```
tolbertam: would be good to document that if not explicitly set, this is not validated.

src/java/org/apache/cassandra/auth/MutualTlsCertificateValidator.java
```diff
@@ -35,29 +35,55 @@ public interface MutualTlsCertificateValidator
...
+     *  <li>Validating Issuer information
+     *  <li>Checking organization information etc
+     * </ul>
      *
      * @param clientCertificateChain client certificate chain
-     * @return returns if the certificate is valid or not
+     * @return returns {@code true} if the certificate is valid, {@code false} otherwise
      */
-    boolean isValidCertificate(Certificate[] clientCertificateChain);
+    default boolean isValidCertificate(Certificate[] clientCertificateChain)
```
tolbertam: worth marking deprecated?  From what I can tell its not used anymore. (this may be moot if you end up moving out the age verification from the validator)

src/java/org/apache/cassandra/auth/MutualTlsCertificateValidator.java
```diff
@@ -35,29 +35,55 @@ public interface MutualTlsCertificateValidator
...
+     *  <li>Checking organization information etc
+     * </ul>
+     *
+     * @param clientCertificateChain   client certificate chain
+     * @param maxCertificateAgeMinutes the maximum age allowed for the certificates in the chain
+     * @return returns {@code true} if the certificate is valid, {@code false} otherwise
+     * @throws AuthenticationException when the age of the certificate is greater than the maximum allowed age
+     *                                 for the certificate
+     */
+    boolean isValidCertificate(Certificate[] clientCertificateChain, int maxCertificateAgeMinutes) throws AuthenticationException;
```
tolbertam: maybe `isValidCertificate(Certificate[] clientCertificateChain, int maxCertificateAgeMinutes)` should be `default` and call `isValidCertificate(Certificate[] clientCertificateChain)`, instead of how it currently works.  This way existing implementations don't need to make any implementation changes for this to work.  (this may be moot if you end up moving out the age verification from the validator)

src/java/org/apache/cassandra/auth/MutualTlsCertificateValidator.java
tolbertam: (this may be moot if you end up moving out the age verification from the validator)

conf/cassandra.yaml
frankgh: done, added a comment above

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
```diff
@@ -208,6 +234,16 @@ public AuthenticatedUser getAuthenticatedUser() throws AuthenticationException
                 nospamLogger.error(msg, identity);
                 throw new AuthenticationException(MessageFormatter.format(msg, identity).getMessage());
             }
+
+            // Validates that the certificate age does not exceed the maximum certificate age
+            certificateValidator.certificateAgeConsumer(clientCertificateChain, certificateAgeMinutes -> {
```
tolbertam: Hrmm, I'm wondering why we need both `certificateAgeInMinutes` and `certificateAgeConsumer` on `MutualTlsCertificateValidator`, especiallly if `certificateAgeConsumer` calls `certificateAgeMinutes`?

```java
int certificateAgeMinutes = certificateValidator.certificateAgeInMinutes(clientCertificateChain);

if (certificateAgeMinutes > maxCertificateAgeMinutes)
{
    String errorMessage = String.format("The age of the provided certificate (%d minutes) exceeds the maximum allowed age of %d minutes",  certificateAgeMinutes, maxCertificateAgeMinutes);
     throw new AuthenticationException(errorMessage);
}
```

I'm guessing you are looking to avoid duplicating the logic between the Authenticator and InternodeAuthenticator implementations?   Could't we just make a static utility method instead that both can use? In that case we could potentially go about things in a way where we don't need any new API on `MutualTlsCertificateValidator` (not sure we need validator-specific logic for this).

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
```diff
@@ -208,9 +224,29 @@ public AuthenticatedUser getAuthenticatedUser() throws AuthenticationException
...
+
+            // Report metrics on client certificate expiration
+            MutualTlsMetrics.instance.clientCertificateExpirationDays.update(daysToCertificateExpiration);
+
             return new AuthenticatedUser(role, MTLS, Collections.singletonMap(METADATA_IDENTITY_KEY, identity));
         }
 
+        private int toDays(int minutes)
+        {
+            return (int) TimeUnit.MINUTES.toDays(minutes);
```
tolbertam: just realized, something like 30 minutes will convert to 0 days, which may lead someone to believe that they have a certificate that is expired when they are really close on the margin.  It's probably fine but could see that possibly be misleading.

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
```diff
@@ -208,9 +224,29 @@ public AuthenticatedUser getAuthenticatedUser() throws AuthenticationException
...
             }
+
+            // Validates that the certificate validity period does not exceed the maximum certificate configured validity period
+            int minutesToCertificateExpiration = certificateValidityPeriodValidator.validate(clientCertificateChain);
+            int daysToCertificateExpiration = toDays(minutesToCertificateExpiration);
+
+            if (certificateValidityWarnThreshold != null
+                && minutesToCertificateExpiration < certificateValidityWarnThreshold.toMinutes())
+            {
+                nospamLogger.warn("Certificate with identity '{}' will expire in {} minutes",
```
tolbertam: Since we are logging this, it may be nice to convert it into a duration and then log that so you get something precisely more human readable (e.g. if a certificate expires in 5 hours 4 hours and 37 minutes, this will log '7477' which is not as human readable.

src/java/org/apache/cassandra/auth/MutualTlsCertificateValidityPeriodValidator.java
```diff
@@ -0,0 +1,90 @@
...
+        {
+            return -1;
+        }
+
+        Date notBefore = x509Certificates[0].getNotBefore();
+        Date notAfter = x509Certificates[0].getNotAfter();
+
+        if (notBefore == null || notAfter == null)
+        {
+            throw new AuthenticationException("The provided certificate does not have a notBefore / notAfter field");
```
tolbertam: One nit:  This would throw an exception even if someone didn't have  `max_certificate_validity_period` enabled, which maybe a regression, although I doubt it would be a problem for anyone, but to be on the safe side it may be good to only throw here if the property is configured.

test/unit/org/apache/cassandra/utils/tls/CertificateBuilder.java
```diff
@@ -0,0 +1,234 @@
...
+import org.bouncycastle.cert.X509CertificateHolder;
+import org.bouncycastle.cert.jcajce.JcaX509CertificateConverter;
+import org.bouncycastle.cert.jcajce.JcaX509v3CertificateBuilder;
+import org.bouncycastle.operator.ContentSigner;
+import org.bouncycastle.operator.jcajce.JcaContentSignerBuilder;
+
+/**
+ * A utility class to generate certificates for tests
+ */
+public class CertificateBuilder
```
tolbertam, line 55: Very cool ❤️ assuming this gets in would be fun to do a follow on ticket to see if we can refactor existing tests to use this instead of pre-built keystores.

test/distributed/org/apache/cassandra/distributed/test/auth/MutualTlsCertificateValidityPeriodTest.java
```diff
@@ -0,0 +1,329 @@
...
+    @AfterClass
+    public static void teardown() throws Exception
+    {
+        LOGGER.info("[test step : @AfterClass] teardown");
+        if (CLUSTER != null)
+            CLUSTER.close();
+    }
+
+    @Test
+    public void testExpiringCertificate() throws Exception
```
tolbertam, line 164: Should this be `testValidCertificate`?  From what I can tell this generates a valid certificate and verifies it can connect.

test/distributed/org/apache/cassandra/distributed/test/auth/MutualTlsCertificateValidityPeriodTest.java
```diff
@@ -0,0 +1,329 @@
...
+                                 .with(Feature.NATIVE_PROTOCOL, Feature.GOSSIP));
+        CLUSTER = builder.start();
+
+        configureIdentity();
+    }
+
+    @AfterClass
+    public static void teardown() throws Exception
+    {
+        LOGGER.info("[test step : @AfterClass] teardown");
```
tolbertam: think should remove the diagnostic logs when done :+1:

test/distributed/org/apache/cassandra/distributed/test/auth/MutualTlsCertificateValidityPeriodTest.java
tolbertam: I see, expires within a day, should expect it to log something, did you intend to add logging validation here?

test/distributed/org/apache/cassandra/distributed/test/auth/MutualTlsCertificateValidityPeriodTest.java
```diff
@@ -0,0 +1,329 @@
...
+                      .extractingByKey("identity", as(InstanceOfAssertFactories.STRING)).isEqualTo(IDENTITY);
+            Assertions.assertThat(row.getString("authentication_mode")).isEqualTo("MutualTls");
+        });
+    }
+
+    @Test
+    public void testCertificateReachingMaxValidityPeriod() throws Exception
+    {
+        LOGGER.info("[test step : @Test] testCertificateReachingMaxValidityPeriod");
+        Path clientKeystorePath = generateClientCertificate(b -> b.notBefore(Instant.now().minus(3, ChronoUnit.DAYS))
```
tolbertam: `client_encryption_options.max_certificate_validity_period` is set to 30 days, shouldn't this be a larger range?

test/distributed/org/apache/cassandra/distributed/test/auth/MutualTlsCertificateValidityPeriodTest.java
```diff
@@ -0,0 +1,329 @@
...
+            Assertions.assertThat(exception)
+                      .hasMessageContaining("The validity period of the provided certificate (527040 minutes) exceeds the maximum allowed validity period of 43200 minutes");
+        }
+    }
+
+    @Test
+    public void testFailsWhenCertificateIsExpired() throws Exception
+    {
+        LOGGER.info("[test step : @Test] testFailsWhenCertificateIsExpired");
+        Path clientKeystorePath = generateClientCertificate(b -> b.notBefore(Instant.now().minus(30, ChronoUnit.DAYS))
```
tolbertam, line 244: One thing I learned recently is that expired certificates will successfully handshake if the certificate is self signed and directly in the trust store.  I see you are making certificates signed by another CA so this is failing as expected.  Just wondering if you encountered the same :)

test/distributed/org/apache/cassandra/distributed/test/auth/MutualTlsCertificateValidityPeriodTest.java
```diff
@@ -0,0 +1,329 @@
...
+    private static final char[] KEYSTORE_PASSWORD = "cassandra".toCharArray();
+
+    @ClassRule
+    public static TemporaryFolder tempFolder = new TemporaryFolder();
+
+    static CertificateBundle CA;
+    static Path truststorePath;
+
+    @BeforeClass
+    public static void setupClass() throws Exception
```
tolbertam, line 94: Could also use some additional validation on the MutualTlsMetrics to ensure they are updated.

src/java/org/apache/cassandra/auth/MutualTlsCertificateValidityPeriodValidator.java
frankgh: yeah, I thought about this too, just to err on the safe side I will remove this branch

test/distributed/org/apache/cassandra/distributed/test/auth/MutualTlsCertificateValidityPeriodTest.java
frankgh: yeah, good idea, I'll add the logging validation.

test/distributed/org/apache/cassandra/distributed/test/auth/MutualTlsCertificateValidityPeriodTest.java
frankgh: Not really, I have not seen this in the test. But I guess it's something we could easily add a test for

test/distributed/org/apache/cassandra/distributed/test/auth/MutualTlsCertificateValidityPeriodTest.java
frankgh: done!

test/distributed/org/apache/cassandra/distributed/test/auth/MutualTlsCertificateValidityPeriodTest.java
frankgh: done

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
frankgh: done

src/java/org/apache/cassandra/auth/MutualTlsAuthenticator.java
tolbertam: your solution for this looks great, thank you!

test/distributed/org/apache/cassandra/distributed/test/auth/MutualTlsCertificateValidityPeriodTest.java
tolbertam: :+1: (talked out of band, we don't need a test for this, was just sharing an observation :+1:)

tolbertam (approved):
This looks great, thank you!

frankgh:
force pushing after rebasing

frankgh:
Closed via https://github.com/apache/cassandra/commit/a0af41f666c23a840d9df3f06729ed5fd2c06cd1

PR #3121, opened 2024-02-21
Cassandra 19406 5.0

bereng:
 
```

```

### DIFF
--- a/.circleci/generate.sh
+++ b/.circleci/generate.sh
@@ -34,6 +34,8 @@ print_help()
   echo "   -a Generate the config.yml, config.yml.FREE and config.yml.PAID expanded configuration"
   echo "      files from the main config_template.yml reusable configuration file."
   echo "      Use this for permanent changes in config.yml that will be committed to the main repo."
+  echo "   -d Minimal development checks only. Sanity check during your dev before sending it to review for speed and cost reductions."
+  echo "      Submitting cleaning pre-commit clean CI run is still a requirement when the patch is ready for review"
   echo "   -f Generate config.yml for tests compatible with the CircleCI free tier resources"
   echo "   -p Generate config.yml for tests compatible with the CircleCI paid tier resources"
   echo "   -b Specify the base git branch for comparison when determining changed tests to"
@@ -80,14 +82,17 @@ print_help()
 all=false
 free=false
 paid=false
+dev_min=false
 env_vars=""
 has_env_vars=false
 check_env_vars=true
 detect_changed_tests=true
-while getopts "e:afpib:s" opt; do
+while getopts "e:afpdib:s" opt; do
   case $opt in
       a ) all=true
           detect_changed_tests=false
+          ;;
+      d ) dev_min=true
           ;;
       f ) free=true
           ;;
@@ -253,7 +258,7 @@ if $has_env_vars; then
 fi
 
 # Define function to remove unneeded jobs.
-# The first argument is the file name, and the second arguemnt is the job name.
+# The first argument is the file name, and the second argument is the job name.
 delete_job()
 {
   delete_yaml_block()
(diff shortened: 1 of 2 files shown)


bereng:
Workflow seems to be correct [here](https://app.circleci.com/pipelines/github/bereng/cassandra/1170/workflows/7707b811-6d4e-4cff-92de-9aff85d8ee03)

.circleci/generate.sh
```diff
@@ -34,6 +34,7 @@ print_help()
   echo "   -a Generate the config.yml, config.yml.FREE and config.yml.PAID expanded configuration"
   echo "      files from the main config_template.yml reusable configuration file."
   echo "      Use this for permanent changes in config.yml that will be committed to the main repo."
+  echo "   -d Deduplicate. Removes all config variations for each test groups leaving only the vanilla one to run. Skinny test run."
```
smiklosovic: nit: I would remove the sentence "Skinny test run".

.circleci/generate.sh
```diff
@@ -80,15 +81,18 @@ print_help()
 all=false
 free=false
 paid=false
+dedup=false
```
smiklosovic: all good but ... just semantics, shouldnt we rename this to "development" or something like that? I know it _deduplicates_, I just think it is better if we call it like what it is good for rather than what it does.

I think that all the naming can stay and we would just document "-d" flag as "development", up to you.

.circleci/generate.sh
ekaterinadimitrova2: Instead, we can say this can be used during development for minimum sanity check if we are not ready for full pre-commit CI run.

.circleci/generate.sh
```diff
@@ -332,8 +336,72 @@ delete_repeated_jobs()
   fi
 }
 
+# Define function to leave only a single config run for each test group
+# The first and only argument is the file name.
+dedup_jobs()
+{
+  delete_job "$1" "j11_cqlsh_dtests_py311_offheap"
```
ekaterinadimitrova2, line 354: I have a small preference for replacing all these deletions of particular jobs with something that would delete everything except particular jobs the other way around. Otherwise, I can see how people will start adding new jobs to the workflows without thinking whether they should be in the skinny workflow. Also, the list will be smaller and less prone to merge conflicts, etc. However, I leave it to you, @bereng, to decide whether you want to change this. Not a blocker here.

.circleci/generate.sh
```diff
@@ -332,8 +336,72 @@ delete_repeated_jobs()
   fi
 }
 
+# Define function to leave only a single config run for each test group
+# The first and only argument is the file name.
```
ekaterinadimitrova2, line 351: Maybe we mention in the comment also that this will be used during dev, not pre-commit?

ekaterinadimitrova2 (commented):
Berengue, I think changing the pre-commit workflow to the skinny one in place is confusing. Shouldn't the flag generate a new workflow with a different name? Like during_dev_tests or so? I can see how some people will start cherry-picking config commits from others and use this config for pre-commit because it also uses the same name, and then reviewers will have to make them add the missing jobs. You will be surprised how many people use cherry-pick instead of learning how generate.sh works and using it themselves. If people learn there is a different workflow name, they will be at least cherry-picking a different commit  

I do not see a cqlsh_dtests job? Even as one to be approved?

.circleci/generate.sh
bereng: Ah yes I like 'development' semantics better. Which ties to the comment above about mentioning it in the text +1

.circleci/generate.sh
bereng: +1

.circleci/generate.sh
bereng: If you look at the script details we unfortunately don't support that currently as we don't use any yaml processing tooling but rely on regexps.

bereng:
Review comments added and new workflow example [here](https://app.circleci.com/pipelines/github/bereng/cassandra/1172/workflows/992bd208-91a0-426e-886c-4303f248d297)

Current tooling doesn't support removal of a workflow so I couldn't remove the 'separate' ones. I cleaned them though.

Apologies I just realized I used the wrong base branch name in this PR.

.circleci/generate.sh
ekaterinadimitrova2: Not a blocker, it was just a suggestion, resolving

bereng:
Green CI https://app.circleci.com/pipelines/github/bereng/cassandra/1176/workflows/86564a8e-006a-4b57-9fdb-e43e2ad20666

ekaterinadimitrova2:
I believe this is the right CI link: https://app.circleci.com/pipelines/github/bereng/cassandra/1175/workflows/c10ae06a-6d8f-4cfa-8738-46f2534df952
Still green though :-)

.circleci/generate.sh
```diff
@@ -34,6 +34,7 @@ print_help()
   echo "   -a Generate the config.yml, config.yml.FREE and config.yml.PAID expanded configuration"
   echo "      files from the main config_template.yml reusable configuration file."
   echo "      Use this for permanent changes in config.yml that will be committed to the main repo."
+  echo "   -d Minimal development checks only. Sanity check during your dev before sending it to review for speed and cost reductions."
```
ekaterinadimitrova2, line 37: nit that can be addressed on commit: `We should add: Submitting cleaning pre-commit clean CI run is still a requirement when the patch is ready for review.`
Same for trunk

ekaterinadimitrova2 (approved):

.circleci/generate.sh
bereng: Added to PR

PR #3122, opened 2024-02-22
CASSANDRA-18940 Record latencies for SAI post-filtering reads against local storage

maedhroz:


src/java/org/apache/cassandra/index/sai/plan/QueryController.java
```diff
@@ -73,7 +73,7 @@ public class QueryController
     private final ColumnFamilyStore cfs;
     private final ReadCommand command;
     private final QueryContext queryContext;
-    private final TableQueryMetrics tableQueryMetrics;
+    public final TableQueryMetrics tableQueryMetrics;
```
mike-tr-adamson: This is not a big deal but it seems that we could avoid this change if we stored the `tableQueryMetrics` in the `StorageAttachedIndexSearcher` and make `ResultRetriever` an inner class rather than a static inner class.

If we choose not to do that, can the `tableQueryMetrics`, at least, be separated, so that the `public` and `private` members are separate.

src/java/org/apache/cassandra/index/sai/plan/StorageAttachedIndexSearcher.java
```diff
@@ -370,13 +372,20 @@ private UnfilteredRowIterator queryStorageAndFilter(PrimaryKey key)
...
             {
                 queryContext.partitionsRead++;
                 queryContext.checkpoint();
 
-                return applyIndexFilter(key, partition, filterTree, queryContext);
+                UnfilteredRowIterator filtered = applyIndexFilter(key, partition, filterTree, queryContext);
+
+                // Note that we record the duration of the read after post-filtering, which actually 
+                // materializes the rows from disk.
+                queryController.tableQueryMetrics.postFilteringReadLatency.update(Clock.Global.nanoTime() - startTimeNanos, TimeUnit.NANOSECONDS);
```
mike-tr-adamson: Just to clarify, this is recording the per-partition post-filtering read latency. Do we want to change the metric name to reflect this. Something like `postFilteringPartitionReadLatency`?

maedhroz: I went back and forth on this a little bit. Especially since I'd like to batch row reads into the same partition read sooner rather than later, and even when we read single rows, we're issuing a partition read, it felt unnecessary to qualify the already long name even further. WDYT?

src/java/org/apache/cassandra/index/sai/plan/QueryController.java
maedhroz: It feels like we should have done that anyway. `QueryController` is an odd place for this to be when `ResultRetriever` is the thing that actually actuates recording the metrics...

mike-tr-adamson: When you look at the change for it, it looks so much better and simplifies the code again.

src/java/org/apache/cassandra/index/sai/plan/StorageAttachedIndexSearcher.java
mike-tr-adamson: I'm good with that. We only read partitions right now so it would be excessive.

mike-tr-adamson (approved):

maedhroz:
Committed as https://github.com/apache/cassandra/commit/1d7bae3697b97e64de2c2b958427ef86a1b17731

PR #3133, opened 2024-02-26
CASSANDRA-19429 4.1 remove capacity calls when unnecessary

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -126,6 +126,7 @@ import org.apache.cassandra.io.sstable.SSTableMultiWriter;
 import org.apache.cassandra.io.sstable.format.SSTableFormat;
 import org.apache.cassandra.io.sstable.format.SSTableReader;
 import org.apache.cassandra.io.sstable.format.Version;
+import org.apache.cassandra.io.sstable.format.big.BigFormat;
 import org.apache.cassandra.io.sstable.metadata.MetadataCollector;
 import org.apache.cassandra.io.util.File;
 import org.apache.cassandra.io.util.FileOutputStreamPlus;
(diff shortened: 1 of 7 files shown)


src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -3048,19 +3049,19 @@ public boolean isEmpty()
     public boolean isRowCacheEnabled()
     {
 
-        boolean retval = metadata().params.caching.cacheRows() && CacheService.instance.rowCache.getCapacity() > 0;
+        boolean retval = metadata().params.caching.cacheRows() && DatabaseDescriptor.getRowCacheSizeInMiB() > 0;
```
Mmuzaf, line 3052: `DatabaseDescriptor.getRowCacheSizeInMiB() > 0` -- this condition is unclear to the user/developer facing it for the first time, so my suggestion is to create a new method in the `DatabaseDescriptor` class and reuse it where we need it. I've looked at the latest branch (trunk) and I think we could push a new method there as well.

An example:
```
    public static boolean isRowCacheEnabled()
    {
        return conf.row_cache_size.toMebibytes() > 0;
    }
```

Similar methods should be created for the counter and key caches.

src/java/org/apache/cassandra/io/sstable/format/SSTableReader.java
```diff
@@ -716,9 +716,8 @@ public void setupOnline()
         // under normal operation we can do this at any time, but SSTR is also used outside C* proper,
         // e.g. by BulkLoader, which does not initialize the cache.  As a kludge, we set up the cache
         // here when we know we're being wired into the rest of the server infrastructure.
-        InstrumentingCache<KeyCacheKey, RowIndexEntry> maybeKeyCache = CacheService.instance.keyCache;
-        if (maybeKeyCache.getCapacity() > 0)
-            keyCache = maybeKeyCache;
+        if (DatabaseDescriptor.getKeyCacheSizeInMiB() > 0)
```
Mmuzaf, line 719: I think for the sake of clarity `isRowCacheEnabled()` should be used here.

src/java/org/apache/cassandra/io/sstable/format/SSTableReader.java
```diff
@@ -1333,7 +1332,7 @@ public void cacheKey(DecoratedKey key, RowIndexEntry info)
     {
         CachingParams caching = metadata().params.caching;
 
-        if (!caching.cacheKeys() || keyCache == null || keyCache.getCapacity() == 0)
+        if (!caching.cacheKeys() || keyCache == null || DatabaseDescriptor.getKeyCacheSizeInMiB() == 0)
```
Mmuzaf, line 1335: Similar to the comment above,`isRowCacheEnabled()` should be used here.

src/java/org/apache/cassandra/db/CounterMutation.java
```diff
@@ -211,7 +210,7 @@ private PartitionUpdate processModifications(PartitionUpdate changes)
 
         List<PartitionUpdate.CounterMark> marks = changes.collectCounterMarks();
 
-        if (CacheService.instance.counterCache.getCapacity() != 0)
+        if (DatabaseDescriptor.getCounterCacheSizeInMiB() > 0)
```
Mmuzaf, line 213: This is also relevant for the trunk branch.

Mmuzaf:
I think all the method calls like `getCapacity() != 0` and `getCapacity() == 0` should be replaced by a method in the DatabaseDescriptor class that is clear and easy to use whether a cache is enabled or not. We should avoid using `getCapacity` method to check such a condition. This is true for the trunk, 5.x, 4.x

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
Mmuzaf: I think for the row cache the correct condition is the following:
`getRowCacheClassName() =! null && conf.row_cache_size.toMebibytes() > 0`

As the row cache is pluggable we should also check the pluggable class is not null.

PR #3135, opened 2024-02-26
CASSANDRA-19417 : LIST SUPERUSERS cql command

skoppu22:
Developing a new CQL command LIST SUPERUSERS, to return list of all roles having superuser privilege (this includes roles who have acquired superuser status). Currently LIST ROLES OF <> command tells us is a role superuser (immediate) and membership info of the role. One way to retrieve has a role acquired superuser privilege status by a grant in the hierarchy is by running LIST ROLES OF <> command and check for atleast one row with super=true in the output. While this works for a given role, there is no way to get all roles having superuser privilege. External services working with C* (for example, Sidecar) may need list of superuser privileged roles details.

 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/auth/Roles.java
+++ b/src/java/org/apache/cassandra/auth/Roles.java
@@ -20,6 +20,7 @@ package org.apache.cassandra.auth;
 import java.util.Collections;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
@@ -80,6 +81,13 @@ public class Roles
     public static Set<RoleResource> getAllRoles()
     {
         return cache.getAllRoles();
+    }
+
+    public static Set<RoleResource> getAllRoles(Predicate<RoleResource> predicate)
+    {
+        return getAllRoles().stream()
+                            .filter(predicate)
+                            .collect(Collectors.toSet());
     }
 
     /**
--- a/src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
+++ b/src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
@@ -1,0 +1,102 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.cassandra.cql3.statements;
+
+import java.util.List;
+import java.util.Set;
+
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.commons.lang3.builder.ToStringStyle;
+
+import org.apache.cassandra.audit.AuditLogContext;
+import org.apache.cassandra.audit.AuditLogEntryType;
+import org.apache.cassandra.auth.AuthKeyspace;
+import org.apache.cassandra.auth.Permission;
+import org.apache.cassandra.auth.RoleResource;
+import org.apache.cassandra.auth.Roles;
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.cql3.ColumnIdentifier;
+import org.apache.cassandra.cql3.ColumnSpecification;
+import org.apache.cassandra.cql3.ResultSet;
+import org.apache.cassandra.db.marshal.UTF8Type;
+import org.apache.cassandra.exceptions.InvalidRequestException;
+import org.apache.cassandra.exceptions.RequestExecutionException;
+import org.apache.cassandra.exceptions.RequestValidationException;
+import org.apache.cassandra.exceptions.UnauthorizedException;
+import org.apache.cassandra.schema.SchemaConstants;
+import org.apache.cassandra.service.ClientState;
+import org.apache.cassandra.transport.messages.ResultMessage;
+
+/**
+ * LIST SUPERUSERS cql command returns list of roles with superuser privileges
+ * This includes superusers and all roles who have superuser role granted in the roles hierarchy
+ */
+public class ListSuperUsersStatement extends AuthorizationStatement
+{
+    private static final List<ColumnSpecification> metadata =
+    List.of(new ColumnSpecification(SchemaConstants.AUTH_KEYSPACE_NAME, AuthKeyspace.ROLES,
+                                    new ColumnIdentifier("role", true), UTF8Type.instance));
+
+    public ListSuperUsersStatement()
+    {
+        // nothing to do
+    }
+
+    public void validate(ClientState state) throws UnauthorizedException, InvalidRequestException
+    {
+        state.ensureNotAnonymous();
+    }
+
+    public void authorize(ClientState state) throws InvalidRequestException
+    {
+        // Allow listing superuser privileged users only if the caller has DESCRIBE permission on 'all roles'
+        if (!DatabaseDescriptor.getAuthorizer()
+                               .authorize(state.getUser(), RoleResource.root())
+                               .contains(Permission.DESCRIBE))
+        {
+            throw new UnauthorizedException("You are not authorized to view superuser details");
+        }
+    }
+
+    public ResultMessage execute(ClientState state) throws RequestValidationException, RequestExecutionException
+    {
+        Set<RoleResource> superUsers = Roles.getAllRoles(Roles::hasSuperuserStatus);
+        if (superUsers == null || superUsers.isEmpty())
+            return new ResultMessage.Void();
+
+        ResultSet result = new ResultSet(new ResultSet.ResultMetadata(metadata));
+
+        superUsers.stream()
+                  .sorted(RoleResource::compareTo)
+                  .forEach(role -> result.addColumnValue(UTF8Type.instance.decompose(role.getRoleName())));
+
+        return new ResultMessage.Rows(result);
+    }
+
+    @Override
+    public String toString()
+    {
+        return ToStringBuilder.reflectionToString(this, ToStringStyle.SHORT_PREFIX_STYLE);
+    }
+
+    @Override
+    public AuditLogContext getAuditLogContext()
+    {
+        return new AuditLogContext(AuditLogEntryType.LIST_SUPERUSERS);
+    }
+}
--- a/test/unit/org/apache/cassandra/auth/RolesTest.java
+++ b/test/unit/org/apache/cassandra/auth/RolesTest.java
@@ -18,7 +18,10 @@
 
 package org.apache.cassandra.auth;
 
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import com.google.common.collect.Iterables;
 import org.junit.Assert;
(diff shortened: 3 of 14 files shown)


test/unit/org/apache/cassandra/auth/RolesTest.java
```diff
@@ -29,23 +32,19 @@
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.db.ConsistencyLevel;
 
-import static org.apache.cassandra.auth.AuthTestUtils.ALL_ROLES;
-import static org.apache.cassandra.auth.AuthTestUtils.ROLE_A;
-import static org.apache.cassandra.auth.AuthTestUtils.ROLE_B;
-import static org.apache.cassandra.auth.AuthTestUtils.ROLE_C;
-import static org.apache.cassandra.auth.AuthTestUtils.getRolesReadCount;
-import static org.apache.cassandra.auth.AuthTestUtils.grantRolesTo;
+import static org.apache.cassandra.auth.AuthTestUtils.*;
```
Maxwell-Guo: I would expand the imports.

test/unit/org/apache/cassandra/auth/RolesTest.java
```diff
@@ -29,23 +32,19 @@
...
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 
 public class RolesTest
 {
+    private static IRoleManager roleManager;
```
Maxwell-Guo: I think it is better to keep it as it is, as the roleManager is not used except for  setupClass method.

test/unit/org/apache/cassandra/cql3/statements/ListSuperUsersStatementTest.java
```diff
@@ -0,0 +1,100 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.cql3.statements;
+
+import com.datastax.driver.core.ResultSet;
+import com.datastax.driver.core.exceptions.UnauthorizedException;
+import org.apache.cassandra.auth.*;
```
Maxwell-Guo: I would expand the imports

CHANGES.txt
```diff
@@ -1,4 +1,5 @@
 5.1
+ * Add LIST SUPERUSERS cql command
```
Maxwell-Guo: missing cassandra jira Id

smiklosovic: @skoppu22 please never do this in a patch, it is just a source of conflicts when your patch is older than what is already in trunk so we need to resolve it unnecessarily on rebasing. A committer will take care of this change for sure upon actual commit.

src/java/org/apache/cassandra/auth/Roles.java
```diff
@@ -189,4 +189,11 @@ public static Role fromRoleResource(RoleResource resource, IRoleManager roleMana
                                   .map(RoleResource::getRoleName)
                                   .collect(Collectors.toSet()));
    }
+
+    public static Set<RoleResource> getAllSuperUsers()
+    {
+        return getAllRoles().stream()
```
smiklosovic: maybe it would be cool to have a method `getAllRoles(Predicate<RoleResource> p)` in Roles.java like this:

    public static Set<RoleResource> getAllRoles(Predicate<RoleResource> p)
    {
        return getAllRoles().stream().filter(p).collect(Collectors.toUnmodifiableSet());
    }

That way we would not have a very specific method about `getAllSuperUsers` but developers would have more freedom to filter roles however they please ... e.g. "give me all roles with so and so parents" or applicable permissions or role names etc ...

This means we would remove this method and it would be replaced with 

    Roles.getAllRoles(Roles::hasSuperuserStatus)

everywhere

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
```diff
@@ -0,0 +1,108 @@
...
+            throw new UnauthorizedException("You are not authorized to view superuser details");
+        }
+    }
+
+    public ResultMessage execute(ClientState state) throws RequestValidationException, RequestExecutionException
+    {
+        return resultMessage(Roles.getAllSuperUsers());
+    }
+
+    private ResultMessage resultMessage(Set<RoleResource> roles)
```
smiklosovic: can't be the body of this method just merged with `execute` method? I just dont see the advantage of moving the actual logic into this method when it can be just all done in `execute`.

Something like this:

    public ResultMessage execute(ClientState state) throws RequestValidationException, RequestExecutionException
    {
        Set<RoleResource> allSuperusers = Roles.getAllRoles(Roles::hasSuperuserStatus);

        if (allSuperusers.isEmpty())
            return new ResultMessage.Void();

        ResultSet result = new ResultSet(new ResultSet.ResultMetadata(metadata));

        allSuperusers.stream()
                     .sorted(RoleResource::compareTo)
                     .forEach(role -> result.addColumnValue(UTF8Type.instance.decompose(role.getRoleName())));

        return new ResultMessage.Rows(result);
    }

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
```diff
@@ -0,0 +1,108 @@
...
+    {
+        if (roles.isEmpty())
+            return new ResultMessage.Void();
+
+        ResultSet.ResultMetadata resultMetadata = new ResultSet.ResultMetadata(metadata);
+        ResultSet result = new ResultSet(resultMetadata);
+
+        roles.stream()
+             .sorted(RoleResource::compareTo)
```
smiklosovic: what happens when RoleResource's `name` is null? May it happen?

Also, role.getRoleName() looks like:

    public String getRoleName()
    {
        if (level == Level.ROOT)
            throw new IllegalStateException(String.format("%s role resource has no role name", level));
        return name;
    }

what happens when level is ROOT? Then it will throw? Might that ever happen and if not, what is the proof of that?

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
```diff
@@ -0,0 +1,108 @@
...
+import org.apache.cassandra.transport.messages.ResultMessage;
+
+/**
+ * LIST SUPERUSERS cql command returns list of roles with superuser privileges
+ * This includes superusers and all roles who have superuser role granted in the roles hierarchy
+ */
+public class ListSuperUsersStatement extends AuthorizationStatement
+{
+    private static final List<ColumnSpecification> metadata =
+    ImmutableList.of(new ColumnSpecification(SchemaConstants.AUTH_KEYSPACE_NAME, AuthKeyspace.ROLES,
```
smiklosovic: You can use `java.util.List.of` instead, that is immutable too and we get rid of Guava-isms, was introduced in Java 9. Since we support 11 and 17 and we dropped 8 in 5.0, it will be fine.

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
```diff
@@ -0,0 +1,108 @@
...
+                                             new ColumnIdentifier("role", true), UTF8Type.instance));
+
+    public ListSuperUsersStatement()
+    {
+        // nothing to do
+    }
+
+    public void validate(ClientState state) throws UnauthorizedException, InvalidRequestException
+    {
+        state.ensureNotAnonymous();
```
smiklosovic, line 62: Just saying that we will not be able to list superusers when a user is not able to access this datacenter.

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public void authorize(ClientState state) throws InvalidRequestException
+    {
+        // Allow listing superuser privileged users only if the caller has DESCRIBE permission on 'all roles'
+        if (!DatabaseDescriptor.getAuthorizer()
+                               .authorize(state.getUser(), RoleResource.root())
+                               .contains(Permission.DESCRIBE))
+        {
+            throw new UnauthorizedException("You are not authorized to view superuser details");
```
smiklosovic, line 72: replace "You" with user's name maybe? "You" is pretty-much non-telling as in who tried to list this etc. I guess it will be visible in audit logs but still ... just makes the life easier for everybody.

CHANGES.txt
skoppu22: Do you mean I don't need to add entry to CHANGES.txt? Removing this change

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
skoppu22: Following the same style as LIST ROLES command is doing. This separation usually helps when a command has different clauses supported and formatting differently for different clauses. In this case (as of now), it's quite simple without any clauses, so we can merge

CHANGES.txt
smiklosovic: yeah, committers will take care of that, no worries

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
smiklosovic: yes please merge, this is very straightforward logic

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
skoppu22: Not sure did I understand this comment correctly. ensureNotAnonymous ensures that anonymous user is not running this command and performs auth checks in validateLogin, i.e, verifies the user has DC access and also has CIDR access from that client. So I believe we are good with this here.

After this, authorize() code below ensures the user has DESCRIBE permission on ALL ROLES. 

Please let me know am I missing anything.

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
skoppu22: Here are existing messages we currently return to the user, all of these are referring 'You'. I prefer to be inline with them, because this message goes back to the user performing the operation, for cases likes mtls based auth, the user's identity may be associated with a role that the user may not be aware of (perhaps admins/superusers are associating identities of different teams with different roles internal to C*) and not sure is it OK to expose role associated with the user's identity in this message returned to the end user.

```
   "You have not logged in"
   "You do not have access to this datacenter "
   "You do not have access from this IP "
```

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
smiklosovic: yes correct, just iterating what I see ... no action needed here

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
smiklosovic: ok

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
skoppu22: Here we get roles from getAllRoles which in-turn reads roles details from system_auth.roles table. So results can be empty or roles with a valid name, I tested that we can create a role with name '<space> ', but we cannot create a role with name '' or without name at all. And ROOT role is never returned as part of this. Hence we don't have to worry about null role name.

getAllRoles of anonymous roleManager can return null, so added superUsers == null check at line 79.

I also added unit test testNoRoles() to test for empty roles table.

src/java/org/apache/cassandra/auth/Roles.java
skoppu22: Done

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
skoppu22: Done

src/java/org/apache/cassandra/cql3/statements/ListSuperUsersStatement.java
skoppu22: Done

CHANGES.txt
skoppu22: Removed changes to CHANGES.txt

test/unit/org/apache/cassandra/cql3/statements/ListSuperUsersStatementTest.java
```diff
@@ -0,0 +1,131 @@
...
+        assertRowsNet(executeNet("list superusers"), row("cassandra"), row("role1"), row("role11"));
+
+        useUser("role11", "role11");
+        assertRowsNet(executeNet("list superusers"), row("cassandra"), row("role1"), row("role11"));
+    }
+
+    @Test
+    public void testNoRoles()
+    {
+        try (MockedStatic<Roles> roles = mockStatic(Roles.class))
```
skoppu22, line 80: Modified this to use mock to avoid impacting other test cases if we truncate the roles table

PR #3141, opened 2024-02-27
CASSANDRA-19255 Fixing getRangeTo* operations of StorageService mbea running into NPE for local strategy keyspaces.

nvharikrishna:
getRangeTo* mbean methods are calling `constructRangeToEndpointMap` internally to get endpoints by range for given keyspace and range. This method is internally using CMS to get keyspace metadata. Local strategy keyspaces are not managed by CMS because of which `keyspaceMetadata` is null and then running into NPE. Falling back to `Keyspace.open()` when keyspace metdata is not found in CMS. 

The [Cassandra Jira](https://issues.apache.org/jira/browse/CASSANDRA-19255)


nvharikrishna:
Changes are merged with commit: https://github.com/apache/cassandra/commit/a69c8657d75de627fb1fe518bfe1d657add11740. Closing the pull request.

PR #3157, opened 2024-03-04
CASSANDRA-19355: Accord: PreLoadContext must properly and consistently support ranges

dcapwell:


src/java/org/apache/cassandra/schema/TableId.java
```diff
@@ -50,7 +51,8 @@
 public class TableId implements Comparable<TableId>
 {
     public static final long MAGIC = 1956074401491665062L;
-    // TODO: should this be a TimeUUID?
```
dcapwell, line 46: FYI TCM moved away from time uuid so ids are deterministic now... so this comment doesn't make sense with TCM

src/java/org/apache/cassandra/service/accord/AccordMessageSink.java
```diff
@@ -222,7 +222,7 @@ public void send(Node.Id to, Request request)
         Preconditions.checkNotNull(verb, "Verb is null for type %s", request.type());
         Message<Request> message = Message.out(verb, request);
         InetAddressAndPort endpoint = endpointMapper.mappedEndpoint(to);
-        logger.debug("Sending {} {} to {}", verb, message.payload, endpoint);
```
dcapwell: this is super dense / verbose and slows things down and makes real bugs hard to find in the logs...

src/java/org/apache/cassandra/service/accord/AccordSafeCommandsForRanges.java
```diff
@@ -0,0 +1,128 @@
...
+import java.util.Objects;
+
+import accord.primitives.Range;
+import accord.primitives.Ranges;
+import accord.primitives.TxnId;
+import accord.utils.async.AsyncChains;
+import accord.utils.async.AsyncResult;
+import org.apache.cassandra.utils.Pair;
+
+public class AccordSafeCommandsForRanges implements AccordSafeState<Range, CommandsForRanges>
```
dcapwell, line 31: since `CommandsForRanges` is immutable and we don't support adding to it atm, this class isn't 100% needed... but figure if we did want to think about caching we would need this, so better to plumb it in now

src/java/org/apache/cassandra/service/accord/AccordStateCache.java
```diff
@@ -278,13 +303,36 @@ public Instance(
...
+            if (!listeners.remove(l))
+                throw new AssertionError("Listener was not registered");
+            if (listeners.isEmpty())
+                listeners = null;
+        }
+
         public Stream<AccordCachingState<K, V>> stream()
         {
             return cache.entrySet().stream()
-                        .filter(e -> keyClass.isAssignableFrom(e.getKey().getClass()))
```
dcapwell: we were matching multiple instances due to CFK stuff having the same type key...

src/java/org/apache/cassandra/service/accord/AccordStateCache.java
```diff
@@ -295,6 +343,11 @@ public S acquireOrInitialize(K key, Function<K, V> valueFactory)
                 node = nodeFactory.create(key, index);
                 node.initialize(valueFactory.apply(key));
                 cache.put(key, node);
+                if (listeners != null)
+                {
+                    AccordCachingState<K, V> finalNode = node;
```
dcapwell, line 349: hopefully java 17 removes the requirement to copy this... =(

src/java/org/apache/cassandra/service/accord/serializers/AccordRoutingKeyByteSource.java
```diff
@@ -0,0 +1,244 @@
...
+import org.apache.cassandra.schema.TableId;
+import org.apache.cassandra.service.accord.api.AccordRoutingKey;
+import org.apache.cassandra.utils.ByteArrayUtil;
+import org.apache.cassandra.utils.bytecomparable.ByteComparable;
+import org.apache.cassandra.utils.bytecomparable.ByteSource;
+import org.apache.cassandra.utils.bytecomparable.ByteSourceInverse;
+
+import static org.apache.cassandra.service.accord.api.AccordRoutingKey.RoutingKeyKind.SENTINEL;
+
+public class AccordRoutingKeyByteSource
```
dcapwell, line 40: names are hard...

test/unit/org/apache/cassandra/repair/FuzzTestBase.java
```diff
@@ -540,27 +528,11 @@ private static RepairOption repairOption(RandomSource rs, Cluster.Node coordinat
     {
         RepairType type = repairTypeGen.next(rs);
         PreviewType previewType = previewTypeGen.next(rs);
-        boolean accordRepair = type == RepairType.FULL && previewType == PreviewType.NONE ? rs.nextBoolean() : false;
         List<String> args = new ArrayList<>();
         args.add(ks);
         List<String> tables = tablesGen.next(rs);
         args.addAll(tables);
-        if (accordRepair)
```
dcapwell: this was adding non-deterministic behavior to repair tests.  I spoke with Ariel about this and also pointed out that the new accord logic is blocking (repair expects async)... I didn't want to tackle fixing this logic so just removing this for now to keep CI stable...

we will need to fix the accord repair logic as it can cause perf regressions for repair

test/unit/org/apache/cassandra/service/accord/AccordKeyspaceTest.java
```diff
@@ -97,4 +134,157 @@ public void serde()
...
+        var partitionGen = fromQT(CassandraGenerators.partitioners());
+
+        var sstableFormats = DatabaseDescriptor.getSSTableFormats();
+        List<String> sstableFormatNames = new ArrayList<>(sstableFormats.keySet());
+        sstableFormatNames.sort(Comparator.naturalOrder());
+
+        List<String> memtableFormats = MemtableParams.knownDefinitions().stream()
+                                                     .filter(name -> !name.startsWith("test_") && !name.equals("default"))
+                                                     .sorted()
+                                                     .collect(Collectors.toList());
```
dcapwell, line 151: I honestly added this because I found data loss once we hit SSTables... turns out internal write logic doesn't do enough validation (as we do this in CQL) so we can write to Memtable, but this gets filtered out when we write to SSTables...  Testing the different formats only came in after to help me debug this issue... and figured its nice to do this anyways =)

test/unit/org/apache/cassandra/service/accord/AccordKeyspaceTest.java
```diff
@@ -97,4 +134,157 @@ public void serde()
...
+                    }
+                }
+                if (pk != null)
+                {
+                    try
+                    {
+                        // using Mutation directly (what we do in Accord) can break when user data is too large; leading to data loss
+                        // The memtable will allow the write, but it will be dropped when writing to the SSTable...
+                        //TODO (now, correctness): since we store the user token + user key, if a key is close to the PK limits then we could tip over and loose our CFK
+//                        new Mutation(AccordKeyspace.getCommandsForKeyPartitionUpdate(store, pk, 42, ByteBufferUtil.EMPTY_BYTE_BUFFER)).apply();
```
dcapwell, line 207: this is the logic CFK uses, but this can cause problems down the road.  For this test these keys are filtered out and CQL is used directly to avoid these issues...

There is a task to fix CFK to not have size limits that different than normal CQL

src/java/org/apache/cassandra/service/accord/AccordKeyspace.java
```diff
@@ -191,6 +203,15 @@ public class AccordKeyspace
 
     private static final ClusteringIndexFilter FULL_PARTITION = new ClusteringIndexSliceFilter(Slices.ALL, false);
 
+    //TODO (now, performance): should this be partitioner rather than TableId?  As of this patch distributed tables should only have 1 partitioner...
+    private static final ConcurrentMap<TableId, AccordRoutingKeyByteSource.Serializer> TABLE_SERIALIZERS = new ConcurrentHashMap<>();
```
belliottsmith, line 207: this is primarily a read-cache, so it might be nice to optimise for this and not have the weight and cost of a concurrent hash map on read (e.g. CoW the whole map but use an Object2Object open hash map). 

We should also avoid infinite state growth under schema changes, so it should either periodically be cleared, or have entries removed when tables are dropped, or on update should clear any stale entries (if we're doing CoW this is prob my favoured approach).

PR #3164, opened 2024-03-07
CASSANDRA-19460 towards uuid_sstable_identifiers_enabled: true in -latest.yaml

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/db/Directories.java
+++ b/src/java/org/apache/cassandra/db/Directories.java
@@ -35,6 +35,7 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.Spliterator;
 import java.util.concurrent.ThreadLocalRandom;
@@ -646,6 +647,29 @@ public class Directories
         }
     }
 
+    /**
+     * Returns directory to write a snapshot to. If directory does not exist, then it is NOT created.
+     *
+     * If given {@code location} indicates secondary index, this will return
+     * {@code <cf dir>/snapshots/<snapshot name>/.<index name>}.
+     * Otherwise, this will return {@code <cf dir>/snapshots/<snapshot name>}.
+     *
+     * @param location base directory
+     * @param snapshotName snapshot name
+     * @return directory to write a snapshot
+     */
+    public static Optional<File> getSnapshotDirectoryIfExists(File location, String snapshotName)
+    {
+        if (isSecondaryIndexFolder(location))
+        {
+            return get(location.parent(), SNAPSHOT_SUBDIR, snapshotName, location.name());
+        }
+        else
+        {
+            return get(location, SNAPSHOT_SUBDIR, snapshotName);
+        }
+    }
+
     public File getSnapshotManifestFile(String snapshotName)
     {
         File snapshotDir = getSnapshotDirectory(getDirectoryForNewSSTables(), snapshotName);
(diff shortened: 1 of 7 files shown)


test/conf/latest_diff.yaml
```diff
@@ -59,3 +59,5 @@ default_secondary_index: sai
 default_secondary_index_enabled: true
 
 storage_compatibility_mode: NONE
+
+uuid_sstable_identifiers_enabled: true
```
blambov: InstanceConfig.java also needs to be changed to add this for in-jvm dtests.

smiklosovic: @blambov yeah it already is, in the very first commit.

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -635,14 +636,25 @@ public static File getSnapshotDirectory(Descriptor desc, String snapshotName)
      * @return directory to write snapshot
      */
     public static File getSnapshotDirectory(File location, String snapshotName)
+    {
+        return getSnapshotDirectory(location, snapshotName, true).get();
+    }
+
+    private static Optional<File> getSnapshotDirectory(File location, String snapshotName, boolean create)
```
blambov: Using an extra boolean parameter looks a bit forced and would be very easy to miss. How about leaving the existing methods as they were and adding a new method for the uses where the `create` flag would not be set? Called e.g. `getXXXDirectoryIfPresent`? Actually,  renaming the existing methods to `getOrCreate...` would be even better, but I don't know how much change that would entail.

blambov (approved):

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -685,6 +719,28 @@ public static File getBackupsDirectory(File location)
...
+     * Returns directory to write a backup to. If directory does not exist, then it is NOT created.
+     *
+     * If given {@code location} indicates secondary index, this will return
+     * {@code <cf dir>/backups/.<index name>}.
+     * Otherwise, this will return {@code <cf dir>/backups/}.
+     *
+     * @param location base directory
+     * @return directory to write a backup
+     */
+    public static Optional<File> getBackupsDirectoryIfExists(File location)
```
jacek-lewandowski, line 732: I know it is probably unrelated or too late but I don't really like the fact that the table's directory structure leaks outside the `Descriptor` class.

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -1326,6 +1386,12 @@ else if (!dir.tryCreateDirectories() && !(dir.exists() && dir.isDirectory()))
         return dir;
     }
 
+    public static Optional<File> get(File base, String... subdirs)
+    {
+        File dir = subdirs == null || subdirs.length == 0 ? base : new File(base, join(subdirs));
```
jacek-lewandowski, line 1391: shouldn't `null` be an assertion error there?

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -673,6 +697,16 @@ public static File getBackupsDirectory(Descriptor desc)
...
+     * Returns directory to write a backup to. If directory does not exist, then one is created.
+     *
+     * If given {@code location} indicates secondary index, this will return
+     * {@code <cf dir>/backups/.<index name>}.
+     * Otherwise, this will return {@code <cf dir>/backups/}.
+     *
+     * @param location base directory
+     * @return directory to write a backup
+     */
     public static File getBackupsDirectory(File location)
```
jacek-lewandowski, line 710: Actually, the method name should be changed, it is misleading

jacek-lewandowski (approved):

src/java/org/apache/cassandra/db/Directories.java
smiklosovic: we already talked about this with @blambov , we know it is, but it is too late to do anything about that because it is public and god knows how people use this so we on purpose left it as it was

src/java/org/apache/cassandra/db/Directories.java
smiklosovic: i just took this from getOrCreate()

src/java/org/apache/cassandra/db/Directories.java
smiklosovic: my original patch which i dont have anymore was renaming it but we reverted it

PR #3165, opened 2024-03-07
CASSANDRA-18753 followup: Further test fixes

blambov:
Fix dtest failures caused by an omitted part of the fix to avoid NPE on UCS initialization (CASSANDRA-19169).
Fix BounceIndexRebuildTest for SAI.

### DIFF
--- a/test/distributed/org/apache/cassandra/distributed/test/log/BounceIndexRebuildTest.java
+++ b/test/distributed/org/apache/cassandra/distributed/test/log/BounceIndexRebuildTest.java
@@ -41,10 +41,12 @@ public class BounceIndexRebuildTest extends TestBaseImpl
             cluster.schemaChange(withKeyspace("create index idx on %s.tbl (x)"));
             Object[][] res = cluster.coordinator(1).execute(withKeyspace("select * from %s.tbl where x=5"), ConsistencyLevel.ALL);
             assert res.length > 0;
+            final String pattern = "Index build of idx complete|Index \\[idx\\] became queryable after successful build";
+            int preBounce = cluster.get(1).logs().grep(pattern).getResult().size();
 
             cluster.get(1).shutdown().get();
             cluster.get(1).startup();
-            assertEquals(1, cluster.get(1).logs().grep("Index build of idx complete").getResult().size());
+            assertEquals(preBounce + 1, cluster.get(1).logs().grep(pattern).getResult().size());
             res = cluster.coordinator(1).execute(withKeyspace("select * from %s.tbl where x=5"), ConsistencyLevel.ALL);
             assert res.length > 0;
         }
--- a/test/distributed/org/apache/cassandra/distributed/test/log/BounceIndexRebuildTest.java
+++ b/test/distributed/org/apache/cassandra/distributed/test/log/BounceIndexRebuildTest.java
@@ -18,6 +18,7 @@
 
 package org.apache.cassandra.distributed.test.log;
 
+import org.junit.Assert;
 import org.junit.Test;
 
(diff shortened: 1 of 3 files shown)


blambov:
@beobal could you review these short test fixes?

beobal:
> @beobal could you review these short test fixes?

Sure, will do

test/distributed/org/apache/cassandra/distributed/test/log/BounceIndexRebuildTest.java
```diff
@@ -41,10 +41,12 @@ public void bounceTest() throws Exception
             cluster.schemaChange(withKeyspace("create index idx on %s.tbl (x)"));
             Object[][] res = cluster.coordinator(1).execute(withKeyspace("select * from %s.tbl where x=5"), ConsistencyLevel.ALL);
             assert res.length > 0;
+            final String pattern = "Index build of idx complete|Index \\[idx\\] became queryable after successful build";
```
beobal: It's been a while since I looked at this, so I had to code dive a bit to understand why switching between legacy 2i and SAI would necessitate this change. Would you be ok with adding a comment? Something along the lines of:
```
// Both SAI and legacy 2i indexes will log the "Index [idx] became queryable..." message every time the index is 
// initialised. Only legacy indexes may rebuild a previously built index, in which case they log the "Index
// build of idx complete" message. So this is to verify that happens at most once, implying that legacy
// 2i does not get rebuilt after a restart.
```

beobal (approved):
LGTM. I just made a minor suggestion for an explanatory comment on the index rebuild test, feel free to ignore that if you think it's too obvious.

test/distributed/org/apache/cassandra/distributed/test/log/BounceIndexRebuildTest.java
blambov: From what I can see, though, it does get rebuilt after the restart. With legacy index the pre-bounce count is 0 and it becomes 1 after the restart. With SAI the numbers are 1 and 2, respectively.

I really don't know what the test is supposed to verify; it is introduced with TCM and I went with the understanding that the test is checking that there _is_ a rebuild after the restart.

beobal: > From what I can see, though, it does get rebuilt after the restart. With legacy index the pre-bounce count is 0 and it becomes 1 after the restart. 

Are you sure? What I see on trunk is that the pre-bounce count is 1 and stays at 1 following the restart.
If we change the search to use your pattern that changes to 2 pre-bounce and 3 after, because we always log the `Index [idx] became queryable` message, whether the index was rebuilt or not. 

> With SAI the numbers are 1 and 2, respectively

Yes, because both SAI and legacy 2i indexes will log the `Index [idx] became queryable...` as I said in the comment, but SAI will never log `Index build of idx complete`.

> I really don't know what the test is supposed to verify; it is introduced with TCM and I went with the understanding that the test is checking that there is a rebuild after the restart.

er, no it's the opposite. It's verifying that the legacy 2i index is _not_ rebuilt after a bounce (which would cause it to log `Index build of idx complete`).

blambov: I see. What I needed, then, is to identify the corresponding message for SAI.

There does not appear to be one, though, but it does show "Validating" after restart. Changed test to check that if no legacy build message was reported, then a SAI validation one must be seen post-bounce.

beobal (approved):
+1

PR #2254, opened 2023-03-30
Cassandra 12937 sstable compression rebased to  trunk

Claudenw:
Pull request 2199 rebased on trunk.

```

The [Cassandra Jira](https://issues.apache.org/jira/projects/CASSANDRA/issues/)

Associated pull request: https://github.com/apache/cassandra/pull/2199

PR #3168, opened 2024-03-08
Cassandra 12937 - Default SSTable compression in yaml

Claudenw:
 
```

```


Claudenw:
This pull request adds the ability to define the default table compression in the config.yaml file.

However, doing this led to an issue where System tables should not have the default table compression applied.  In order to support this the `TableParams.builder` needs to know the keyspace name when building so that it can be passed to the `CompressonParams.builder` where the filter for determining which tables get the old DEFAULT and which get the new calculated default Compression params is handled by the `CompressionParams.defaultParams(keyspace)` method.

This change in turn led to changes to several files where the keyspace name needed to be provided for the `TableParams.builder().build()` function call.

Claudenw:
merged changes found in #2282, #2254, #2199

.build/build-rat.xml
michaelsembwever, line 1: why is all this in the PR ?

conf/cassandra.yaml
```diff
@@ -646,6 +646,50 @@ commitlog_disk_access_mode: legacy
 #
 # flush_compression: fast
 
+# Defines the default compression used on tables when none is specified
+# in the CQL command.
+#
+# The class_name is the compressor class name. It may be one of the aliases,
+# the class name of a system ICompressor implementation, or fully qualified
+# name of a class that implements ICompressor and has a public static 'create' method that accepts
+# a Map<String, String> argument and returns an instance of the class.
```
michaelsembwever: nit:
```suggestion
# name of a ICompressor class.
```

more concise.  no need to provide ICompressor implementation details here.

conf/cassandra.yaml
```diff
@@ -646,6 +646,50 @@ commitlog_disk_access_mode: legacy
...
+#  noop     NoopCompressor
+#  snappy   SnappyCompressor
+#  zstd     ZstdCompressor
+#
+# The standard parameters are any required or optional parameter for the instantiation of the
+# specified class, or one of the following standard parameters:
+# Parameter             Usage
+# enabled               Disables compression if set to false. Defaults to true.
+# chunk_length          The length of the compresson chunks, must include KiB, MiB or GiB suffix, defaults to 16KiB.
+# chunk_length_in_kb    Same as above but expects an integer.
```
michaelsembwever: nit: deprecated alias for above in kb an an integer.

conf/cassandra.yaml
```diff
@@ -646,6 +646,50 @@ commitlog_disk_access_mode: legacy
...
+# min_compress_ratio    The minimal acceptable compression, must greater than or equal to 1.0.
+# max_compressed_length The maximum size for a compressed block. Must be less than or equal to chunk_length.
+#                       Must include KiB, MiB or GiB suffix. Defaults to Integer.MAX_VALUE
+#
+# Only one of the min_compress_ratio and max_compressed_length options can be specified.
+# They are mathematically related in that 'chunk_length / max_compressed_length = min_compress_ratio'.
+# If neither option is specified a min_compress_ratio of 0.0 and a max_compressed_length of
+# Integer.MAX_VALUE KB is the default.
+#
+# Only one of chunk_length or chunk_length_in_kb may be specified.
```
michaelsembwever: with the above nit, this line can be removed.

src/java/org/apache/cassandra/cql3/statements/schema/CreateTableStatement.java
```diff
@@ -402,7 +401,7 @@ private void fixupCompactTable(List<ColumnProperties> clusteringTypes,
         if (hasCounters)
             flags.add(TableMetadata.Flag.COUNTER);
 
-        boolean isStaticCompact = !isDense && !isCompound;
+        boolean isStaticCompact = !isDense;
```
michaelsembwever: what changed here ?

src/java/org/apache/cassandra/io/sstable/format/DataComponent.java
```diff
@@ -88,8 +88,12 @@ private static CompressionParams buildCompressionParams(TableMetadata metadata,
                 case fast:
                     if (!compressor.recommendedUses().contains(ICompressor.Uses.FAST_COMPRESSION))
                     {
-                        // The default compressor is generally fast (LZ4 with 16KiB block size)
-                        compressionParams = CompressionParams.DEFAULT;
+                        // The default compressor is generally Fast, but just in case we verify it.
+                        compressionParams = CompressionParams.defaultParams(metadata.keyspace);
+                        compressor = compressionParams.getSstableCompressor();
+                        if (!compressor.recommendedUses().contains(ICompressor.Uses.FAST_COMPRESSION)) {
+                            compressionParams = CompressionParams.FAST;
```
michaelsembwever: might this change of the compressionParams surprise an operator ? 
(does it deserve a warning ? (one that would only be logged so often per table))

test/conf/cassandra-topology.properties
michaelsembwever, line 1: what's happening here ? (why is this in this PR?)

test/unit/org/apache/cassandra/db/guardrails/GuardrailAllowUncompressedTablesTest.java
michaelsembwever, line 1: nit: there's no real change in this file, so formatting changes isn't appropriate.

test/distributed/org/apache/cassandra/distributed/test/SSTableCompressionTest.java
```diff
@@ -0,0 +1,469 @@
...
+    public static void tearDownCluster()
+    {
+        if (CLUSTER != null)
+            CLUSTER.close();
+    }
+
+    @Test
+    public void testCreation() throws IOException
+    {
+        // create clauster with lz4 compression
```
michaelsembwever: typo

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -505,7 +651,6 @@ static class Serializer implements IVersionedSerializer<CompressionParams>
     {
         public void serialize(CompressionParams parameters, DataOutputPlus out, int version) throws IOException
         {
-            assert version >= MessagingService.VERSION_40;
```
michaelsembwever: why remove these ?

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -514,12 +659,15 @@ public void serialize(CompressionParams parameters, DataOutputPlus out, int vers
                 out.writeUTF(entry.getValue());
             }
             out.writeInt(parameters.chunkLength());
-            out.writeInt(parameters.maxCompressedLength);
+            if (version >= MessagingService.VERSION_40)
+                out.writeInt(parameters.maxCompressedLength);
+            else
+                if (parameters.maxCompressedLength != Integer.MAX_VALUE)
+                    throw new UnsupportedOperationException("Cannot stream SSTables with uncompressed chunks to pre-4.0 nodes.");
```
michaelsembwever: we don't stream anything to pre-4.0 nodes ever anyway.
go back to the `assert` approach IMHO.

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -530,7 +678,9 @@ public CompressionParams deserialize(DataInputPlus in, int version) throws IOExc
                 options.put(key, value);
             }
             int chunkLength = in.readInt();
-            int minCompressRatio = in.readInt();
+            int minCompressRatio = Integer.MAX_VALUE;   // Earlier Cassandra cannot use uncompressed chunks.
+            if (version >= MessagingService.VERSION_40)
+                minCompressRatio = in.readInt();
```
michaelsembwever: same as above.  i would think the `assert` is the correct way here.

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -555,7 +704,8 @@ public long serializedSize(CompressionParams parameters, int version)
                 size += TypeSizes.sizeof(entry.getValue());
             }
             size += TypeSizes.sizeof(parameters.chunkLength());
-            size += TypeSizes.sizeof(parameters.maxCompressedLength());
+            if (version >= MessagingService.VERSION_40)
+                size += TypeSizes.sizeof(parameters.maxCompressedLength());
```
michaelsembwever: same as above. i would think the assert is the correct way here.

michaelsembwever (changes requested):
few small changes and questions requested…

src/java/org/apache/cassandra/cql3/statements/schema/CreateTableStatement.java
Claudenw:
```
        boolean isDense = !clusteringTypes.isEmpty();
        boolean isCompound = clusteringTypes.size() > 1;
```
if it is not dense then it is empty,  and therefor it can not be compound.  isCompound is always false if isDense if false.

Removed to make logic more readable.

src/java/org/apache/cassandra/io/sstable/format/DataComponent.java
Claudenw: This ensures that the system works as it did before.  Previously, if the compression associated with the table was not fast, the system would use DEFAULT.  However, we have added a step where the user can define the default compression used for tables so we have to check that.  If it is 'fast' use it, if it is not fast use the default FAST one.

src/java/org/apache/cassandra/schema/CompressionParams.java
Claudenw: because we handle the case where version is < VERSION_40.

Claudenw: @smiklosovic Do you recall why you changed this code?

michaelsembwever: where would we be doing this ?   it cannot happen in an online cluster. (no 5+ node will ever be messaging with a <4 node)

Claudenw: I didn't change this section of code.  It was a change that @smiklosovic made, so I guess we need to wait for his response to understand why it was changed.

smiklosovic: wow I don't remember this at all ... feel free to put it back

src/java/org/apache/cassandra/schema/CompressionParams.java
Claudenw: done.  Updated it latest pull

michaelsembwever (approved):

src/java/org/apache/cassandra/schema/TableParams.java
```diff
@@ -490,6 +490,11 @@ public Builder extensions(Map<String, ByteBuffer> val)
 
     public static class Serializer implements MetadataSerializer<TableParams>
     {
+        private final String keyspace;
```
smiklosovic: I dont think this is a good idea, because then you have `new Serializer....` everywhere. These serializers are meant to be effectively singletons, rather put `keyspace` as parameter of the method instead

Claudenw: `TableParams.Serializer` requires the keyspace for the deserialization method to call the `builder.build(keyspace)` method.  Since `TableParams.Serializer` implements `MetadataSerializer<TableParams>` we would have to change the signature of `deserialize` on all `MetadataSerializer` implementations.  I think this is a far worse idea than having `new TableParams.Serializer(keyspace)` called  in various places.

Claudenw: @michaelsembwever @smiklosovic How do we go about adding the keyspace to a serializer when the `MetadataSerializer` interface is involved?  I don't see a clean way to do this.  Any suggestions would be appreciated.

michaelsembwever: we're pulling in `keyspace` from the context when we shouldn't be (the call to `CompressionParams.defaultParams(keyspace)` needs to be kept in the context).

i'm not sure yet how to do this…

Claudenw: If I read the code correctly `TableParams.Serializer.deserialize()` is only reading something written by `TableParams.Serializer.serialize()`.  That means the compression type was already defined when the data was serialized.  So, we could pass a `null` parameter to the `builder.build()` method and then in the builder catch a `null` keyspace as meaning "just use the old default" then the value that was serialized would be used and the "old" default compression params would be used in any case where the compression was serialized as an empty map.

Would this work?

smiklosovic: @Claudenw  @michaelsembwever 

I tried to fix it here https://github.com/Claudenw/cassandra/pull/12/files

It moved a lot of changes back. They are not necessary at all. Just because we expected keyspace in build method as a param made it more verbose than necessary.

I think it is also not necessary to call it every time, if one reads the context where that code is used, sometimes it is not necessary to deal with compression params again etc ...

Claudenw: The simplest change to fix this would be to have TableParams.Serializer.deserialize() use the DEFAULT if no params were specificed in whatever it is deserializing.  I will make this change shortly.

Claudenw: So how does this deal with the fact that system tables can not use the new default serializer but non-system tables should?

smiklosovic: I don't think that using it just upon deserialization is enough. We want to set this up correctly in whole database context when creating table metadata etc ...

smiklosovic: I dont think we should change that build() method and deal with that in it. It should be as it was. Putting there null is also strange ... Let's @michaelsembwever be an arbiter who decides what is better. Currently we just make more changes than necessary. Putting keyspace into build() is just weird imho.

Claudenw: If we leave the `CompressionParams.defaultParams(keyspace)` method in place then everywhere we need to create compression params we pass the keyspace.  and 99% of the code is correct.

The only place where this causes a problem is in the TableParam serialization, as noted above.  There are 2 cases to consider:

1. Serialization of TableParms to a stream.  In this case the CompressionParams are already set and we don't need to construct them.
2. Deserialization of TableParms from a stream.  In this case the CompressionParams should  have been serialized.  If they were serialized as an empty map then they are the old style default, and we can get that directly from the CompressionParams class.  

I don't see why we need to do a partial build followed by a setting of the params followed by the rest of the build.  This sounds like a recipe for disaster going forward.  Let's keep it as simple as possible, but no simpler.

smiklosovic: we could actually do this https://github.com/Claudenw/cassandra/pull/12/commits/b88cecc712b07c497ccbe55552bfe09aa2b5e91a

so there is not need to have any parameters in build() method at all because we can set that on `params` when building the metadata so de/ser logic stays intact.

smiklosovic: @Claudenw btw you are right that this does not prevent a user to set default custom compressor for system tables. This still needs to be covered. I will think about that ... But in general I think that this is the right direction.

smiklosovic: @Claudenw (@michaelsembwever )

hah I realized that ....


    cqlsh> ALTER TABLE system.local WITH compression = {'class': 'ZstdCompressor'};
    Unauthorized: Error from server: code=2100 [Unauthorized] message="system keyspace is not user-modifiable."

So we are already covered on that front. But if somebody wanted to set that programmatically, we can do this:

https://github.com/Claudenw/cassandra/pull/12/commits/ea26c2e8cf684563dc88d909dd02ba9c2669b62b

smiklosovic: @Claudenw after my changes, that "partiallyBuild" is not necessary, I have removed that  https://github.com/Claudenw/cassandra/pull/12/commits/20c790a255f89da08506b728a8170d36d6ef6865

so, we are not having partially anywhere, we are not putting keyspace anywhere, we are checking that system is always on default and ser/de is not touched at all.

conf/cassandra.yaml
```diff
@@ -1113,6 +1113,46 @@ snapshot_links_per_second: 0
...
+# The class_name is the compressor class name. It may be one of the aliases,
+# the class name of a system ICompressor implementation, or fully qualified
+# name of a ICompressor class.
+#
+# class aliases are:
+#  Alias    System compressor impl.
+#  deflate  DeflateCompressor
+#  lz4      LZ4Compressor
+#  none     (null) -- compresson disabled
+#  noop     NoopCompressor
```
jacek-lewandowski: when a user would want to use noop over none?

conf/cassandra.yaml
```diff
@@ -1113,6 +1113,46 @@ snapshot_links_per_second: 0
...
+#  snappy   SnappyCompressor
+#  zstd     ZstdCompressor
+#
+# The standard parameters are any required or optional parameter for the instantiation of the
+# specified class, or one of the following standard parameters:
+# Parameter             Usage
+# enabled               Disables compression if set to false. Defaults to true.
```
jacek-lewandowski: what does it mean the compressor is enabled or not?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2666,11 +2670,21 @@ public static void setFlushCompression(Config.FlushCompression compression)
...
-     * more, depending on how soon the sync policy stops all writing threads.
-     */
+    public static ParameterizedClass getSSTableCompression()
+    {
+        return sstableCompression;
+    }
+
+    public static void setSSTableCompression(ParameterizedClass compressor)
+    {
+        conf.sstable.default_compression = compressor;
```
jacek-lewandowski: this seems to have no effect for the subsequent uses of `getSSTableCompression`

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2666,11 +2670,21 @@ public static void setFlushCompression(Config.FlushCompression compression)
         conf.flush_compression = compression;
     }
 
-    /**
-     * Maximum number of buffers in the compression pool. The default value is 3, it should not be set lower than that
-     * (one segment in compression, one written to, one in reserve); delays in compression may cause the log to use
-     * more, depending on how soon the sync policy stops all writing threads.
-     */
+    public static ParameterizedClass getSSTableCompression()
```
jacek-lewandowski: those methods should be named `(get|set)DefaultSSTableCompression`, right?

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
```diff
@@ -557,14 +557,14 @@ public void validate(ClientState state)
                 MemtableParams.get(attrs.getString(TableParams.Option.MEMTABLE.toString()));
             Guardrails.tableProperties.guard(attrs.updatedProperties(), attrs::removeProperty, state);
 
-            validateDefaultTimeToLive(attrs.asNewTableParams());
+            validateDefaultTimeToLive(attrs.asNewTableParams(keyspaceName));
```
jacek-lewandowski: Actually I'm not sure why we need to use this - is it just to distinguish whether we should use compression or not for the system tables? why don't you just make the system table explicitly use no compression?

test/conf/cassandra-with-sstable-compressor.yaml
```diff
@@ -0,0 +1,111 @@
+
```
jacek-lewandowski: we already have a yaml for compression

jacek-lewandowski:
> However, doing this led to an issue where System tables should not have the default table compression applied. In order to support this the TableParams.builder needs to know the keyspace name when building so that it can be passed to the CompressonParams.builder where the filter for determining which tables get the old DEFAULT and which get the new calculated default Compression params is handled by the CompressionParams.defaultParams(keyspace) method.

> This change in turn led to changes to several files where the keyspace name needed to be provided for the TableParams.builder().build() function call.

Actually why do we need all of those changes? System tables are not something which magically appear in the system - they are created explicitly and we can explicitly set compression to none when creating them?

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -41,163 +47,316 @@
...
+        {
+            throw new ConfigurationException("Access forbiden", e);
+        }
+        catch (IllegalAccessException e)
+        {
+            throw new ConfigurationException("Cannot access method create in " + className.getName(), e);
+        }
+        catch (InvocationTargetException e)
+        {
+            if (e.getTargetException() instanceof ConfigurationException)
```
jacek-lewandowski: we have `Throwables.unwrap` method for this

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -41,163 +47,316 @@
...
+
+        try
+        {
+            Method method = className.getMethod("create", Map.class);
+            ICompressor compressor = (ICompressor) method.invoke(null, parameters);
+            // Check for unknown options
+            CompressionParams.checkCompressorOptions(compressor, parameters.keySet());
+            return compressor;
+        }
+        catch (NoSuchMethodException e)
```
jacek-lewandowski: I don't see a reason to catch each of those exception in its distinct block - to me it would be enough to just catch all the exception in one block, unwrap and rethrow as `ConfigurationException` if it isn't already

jacek-lewandowski (changes requested):

conf/cassandra.yaml
Claudenw: I don't know, but it was defined in the system before I made changes.

conf/cassandra.yaml
Claudenw: This is a flag that was in the system before I started.  It indicates that the compression has been turned off.  It seems that it can be defined but not used.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
Claudenw: Should now be fixed.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
Claudenw: should now be fixed

src/java/org/apache/cassandra/schema/CompressionParams.java
Claudenw: changed to use Throwables.unwrap, thanks for the pointer.

src/java/org/apache/cassandra/schema/CompressionParams.java
Claudenw: changed

src/java/org/apache/cassandra/cql3/statements/schema/AlterTableStatement.java
Claudenw: They currently use the DEFAULT compression.  I believe that Stefan discovered during testing that setting the system tables to anything except the DEFAULT definition caused errors.

conf/cassandra.yaml
jacek-lewandowski: I think that things like this should be cleared out as a part of this ticket.

smiklosovic: I think we should copy what is currently there as closely as possible. If somebody wants to create a table with default compression which is disabled, they should be able to do that.

conf/cassandra.yaml
smiklosovic: Noop: 

    A Compressor which doesn't actually compress any data. This is useful for either non-compressible data
    (typically already compressed or encrypted) that you still want block checksums for or for fast writing.
    Some relevant tickets:
    <p>
        <ul>
          <li>CASSANDRA-12682: Non compressed SSTables can silently corrupt data</li>
          <li>CASSANDRA-9264: Non compressed SSTables are written without checksums</li>
      </ul>
  </p>

src/java/org/apache/cassandra/schema/TableParams.java
smiklosovic: I think that this whole conversation is not relevant anymore by CASSANDRA-19592

test/conf/cassandra-with-sstable-compressor.yaml
smiklosovic: i dont understand this comment.

PR #3173, opened 2024-03-12
Cassandra 19458 5.0

bereng:
 
```

```

### DIFF
--- a/.circleci/generate.sh
+++ b/.circleci/generate.sh
@@ -157,6 +157,10 @@ if $has_env_vars && $check_env_vars; then
   done
 fi
 
+if $dev_min && $detect_changed_tests; then
+  die "-d doesn't support repeated tests. Use -s to skip it."
+fi
+
 if $free; then
   ($all || $paid) && die "Cannot use option -f with options -a or -p"
   echo "Generating new config.yml file for free tier from config_template.yml"
@@ -282,8 +286,8 @@ delete_repeated_jobs()
     delete_job "$1" "j17_utests_cdc_repeat"
     delete_job "$1" "j11_utests_compression_repeat"
     delete_job "$1" "j17_utests_compression_repeat"
-    delete_job "$1" "j11_utests_trie_repeat"
-    delete_job "$1" "j17_utests_trie_repeat"
+    delete_job "$1" "j11_utests_latest_repeat"
+    delete_job "$1" "j17_utests_latest_repeat"
     delete_job "$1" "j11_utests_oa_repeat"
     delete_job "$1" "j17_utests_oa_repeat"
     delete_job "$1" "j11_utests_system_keyspace_directory_repeat"
@@ -306,9 +310,9 @@ delete_repeated_jobs()
   fi
   if (! (echo "$env_vars" | grep -q "REPEATED_JVM_DTESTS=")); then
     delete_job "$1" "j11_jvm_dtests_repeat"
-    delete_job "$1" "j11_jvm_dtests_vnode_repeat"
+    delete_job "$1" "j11_jvm_dtests_latest_vnode_repeat"
     delete_job "$1" "j17_jvm_dtests_repeat"
-    delete_job "$1" "j17_jvm_dtests_vnode_repeat"
+    delete_job "$1" "j17_jvm_dtests_latest_vnode_repeat"
   fi
   if (! (echo "$env_vars" | grep -q "REPEATED_JVM_UPGRADE_DTESTS=")); then
     delete_job "$1" "start_jvm_upgrade_dtests_repeat"
@@ -317,10 +321,10 @@ delete_repeated_jobs()
   if (! (echo "$env_vars" | grep -q "REPEATED_DTESTS=")); then
     delete_job "$1" "j11_dtests_repeat"
     delete_job "$1" "j11_dtests_vnode_repeat"
-    delete_job "$1" "j11_dtests_offheap_repeat"
+    delete_job "$1" "j11_dtests_latest_repeat"
     delete_job "$1" "j17_dtests_repeat"
     delete_job "$1" "j17_dtests_vnode_repeat"
-    delete_job "$1" "j17_dtests_offheap_repeat"
+    delete_job "$1" "j17_dtests_latest_repeat"
   fi
   if (! (echo "$env_vars" | grep -q "REPEATED_LARGE_DTESTS=")); then
     delete_job "$1" "j11_dtests_large_repeat"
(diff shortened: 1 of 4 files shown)


bereng:
CI workflow [LGTM](https://app.circleci.com/pipelines/github/bereng/cassandra/1185/workflows/280a09fd-75f7-447e-b94d-9224df3e5045). Opened CASSANDRA-19466 for the flaky.

.circleci/generate.sh
```diff
@@ -391,22 +396,20 @@ build_dev_min_jobs()
...
+  delete_job "$1" "j17_utests_latest"
   delete_job "$1" "j17_utests_system_keyspace_directory"
-  delete_job "$1" "start_utests_trie"
   delete_job "$1" "start_utests_system_keyspace_directory"
   delete_job "$1" "start_utests_stress"
   delete_job "$1" "start_utests_long"
   delete_job "$1" "start_utests_fqltool"
   delete_job "$1" "start_utests_compression"
   delete_job "$1" "start_utests_cdc"
-  delete_job "$1" "start_utests_trie"
```
ekaterinadimitrova2, line 404: Shouldn't this be replaced by `start_utests_latest`?

ekaterinadimitrova2:
Your CI link points to github.
Also,  you need to run .generate.sh -a to update the default files. 
Plus:

ekaterinadimitrova2 (commented):
Your CI link points to github.
Also, you need to run .generate.sh -a to update the default files.
Plus:
https://github.com/apache/cassandra/pull/3173/files#r1521467653

bereng:
- Oh I posted the wrong link apologies. Amended now.
- `generate.sh -a` is unnecessary iiuc. It's noop. We're changing nothing there correct?
- `start` jobs get automatically deleted when you delete a job if a regexp matches.

.circleci/generate.sh
bereng: No iiuc. Some `start` counterparts get removed automatically depending some regexp magic.

ekaterinadimitrova2: We should either do regex intentionally or remove jobs intentionally for all the start jobs. I am an engineer, not a magician :-)  Otherwise, it is brittle and unclear what is going on.

ekaterinadimitrova2:
`generate.sh -a is unnecessary iiuc. It's noop. We're changing nothing there correct?`
I tested yesterday that it is not noop. The deletion lines in the script trigger the removal of the correct repeated jobs from the default files which makes sense to me.
I pulled your branch, removed the `do not merge` commit and ran `./generate.sh -a`

``start` jobs get automatically deleted when you delete a job if a regexp matches.`
That is unclear and brittle. Why some jobs are removed by regex and some are deleted by us? What is this regex meant for originally? We should have either the regex deal with all the `start_` jobs or rely on our manual deletion. Otherwise it becomes prone to bugs and messy.

.circleci/generate.sh
bereng: I'm not sure I am following. This is what it has always [been](https://github.com/apache/cassandra/pull/3173/commits/9351754c928505696bd39d7ba587b0f870f11cf4#diff-9405c8896ec20e5e91399a71f450251c8da83d336df2d3a00c66d6c914ca05b4R275), known behavior and we even recently reviewed that regexp match in another ticket.

If we want to change that to make it specific it's going to be _very_ verbose to remove each `start_` job individually. Food for another ticket imo, specially since it's going to hit _all_ branches. Not convinced about the value at this point imo either. Let's merge this one to fix the problem in this ticket imo and discuss heavy refactorings elsewhere, wdyt?

bereng:
> I tested yesterday that it is not noop

Right you are! I tested on the wrong folder apologies. Good catch. I pushed with the regenerated files.

> That is unclear and brittle

Yes that's why getting the dev workflow hit regexp bugs we fixed together etc., it was a bit more involved, etc. But changing that implies a multi-branch _big_ change refactor unrelated to the problem here. I wouldn't like to block this ticket on that.

.circleci/generate.sh
ekaterinadimitrova2: If the problem was really with the regex - I agree we need to fix it in all branches in another ticket. 
I just verified the exact issue and it is actually with the naming in `java_11_dev_tests` and `java_17_dev_tests` and it exists only in 5.0 and trunk.
There is start_utests_long but the actual job names are prefixed with j11_ and j17_.
And we also created `start_jX_cqlsh-dtests-latest` to rule all the Python versions. 
The mentioned regex was created for the repeated jobs where the naming convention is followed, and everything is fine. 
I am not blocking anyone, but the issue is with the dev workflow, which was supposed to be fixed in this ticket, and it covers only 5.0 and trunk, where the dev workflow was committed.  Probably you did not realize it is the naming and not the regex. We need to follow the naming conventions and to be consistent. This is something we also got as feedback from the community.

bereng: Let's go bit by bit:
- That a regexp is used is known and there was a bug fixed in CASSANDRA-19410. In this [comment](https://issues.apache.org/jira/browse/CASSANDRA-19410?focusedCommentId=17819957&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17819957) the problem about some job names not having a common prefix is identified.
- The `dev` workflow is just the std workflow renamed (CASSANDRA-19414). So the problem with jobs not having a common prefix is in the std workflows and known.
- There isn't a 'dev' circle workflow introduced in 5.0 and trunk. It's just the std one but trimmed and renamed. it's the same yaml.
- Are you suggesting changing all the job names to follow the convention but only in 5.0 and trunk? Given the `dev` workflow _is_ the std's workflow yaml then it does also imply multi-branch, big refactor, little value to this ticket, etc imo rename. You don't want to rename the std workflow's jobs only in 5.0 and trunk.

Iiuc I don't want to do a multi-branch renaming of jobs in this ticket. We can do that in the future one where we'll revisit the pre-commit workflow i.e. We can decide there if we like the regexp technique or not, rename any jobs to a convention, etc.

ekaterinadimitrova2: `Are you suggesting changing all the job names to follow the convention but only in 5.0 and trunk? `
Not at all :-) 
But I should apologize to you, I was looking at config.yml workflow and I thought we change the start buttons names only... I thought I am looking into the config_template.yml.

Yes, that definitely will not work. 

To make it clear what is going on here I suggest we modify the delete_job something like this and call it a day until we have time for proper fix. Then it will be at least clear what and why we do it when we need to get back to it next time:

Here's a refactored version with two functions:

```
# Function to delete a specific YAML block.
delete_specific_yaml_block()
{
  sed -Ei.bak "/^    - ${2}/,/^    [^[:space:]]+|^  [^[:space:]]+/{//!d;}" "$1"
  sed -Ei.bak "/^    - ${2}/d" "$1"
}

# Function to delete a job with the option to delete additional related blocks.
delete_job()
{
  file="$BASEDIR/$1"
  
  # Delete the specific block.
  delete_specific_yaml_block "$file" "${2}:"
  
  # Delete the additional related block.
  delete_specific_yaml_block "$file" "start_${2}:"
}

# Now, you have the flexibility to directly call delete_specific_yaml_block if you want to delete only a specific block.
# Example usage to delete just one block:
# delete_specific_yaml_block "$BASEDIR/filename.yaml" "job_name:"

# And use delete_job to perform the full original behavior.
# Example usage to delete both related blocks:
# delete_job filename.yaml job_name
```

I haven't tested it, if you are fine with the idea, I can spin the patch

ekaterinadimitrova2: Then in the cases where we will delete the one job or start_job and the previous delete_job did not work, we can use the other function and make it clear. As I am not sure when or whether anyone will have the time to fix naming, etc in all branches

bereng: Icwym. You want to split that function into 2 for better readability. Why wouldn't you want to do that to all branches? The script is the same in that area in all branches so it would be cleaner forward merges when the script changes. So we can copy-paste this suggestion to CASSANDRA-19407 i.e. which I intend to tackle once 5.0 is done. Does it work for you?

bereng: Agreed with @ekaterinadimitrova2 offline those can be done in another ticket and we're ready to commit.

bereng:
Workflow lgtm and executes https://app.circleci.com/pipelines/github/bereng/cassandra/1189/workflows/875aff91-3548-4ed1-a6df-23c20a0e6964

PR #3174, opened 2024-03-14
CASSANDRA-19472 - CEP-15 (C*) Integrate accord with repair

bdeggleston:
Wires accord repair into the normal repair path and removes the distinction between an accord repair job and a cassandra repair job, this makes both repair types coexist


src/java/org/apache/cassandra/repair/RepairJob.java
```diff
@@ -91,14 +99,24 @@ public class CassandraRepairJob extends AbstractRepairJob
...
         this.taskExecutor = session.taskExecutor;
         this.parallelismDegree = session.parallelismDegree;
         this.desc = new RepairJobDesc(session.state.parentRepairSession, session.getId(), session.state.keyspace, columnFamily, session.state.commonRange.ranges);
+        this.ks = Keyspace.open(desc.keyspace);
+        this.cfs = ks.getColumnFamilyStore(columnFamily);
+        this.state = new JobState(ctx.clock(), desc, session.state.commonRange.endpoints);
+
+        TableMetadata metadata = this.cfs.metadata();
+        if (session.paxosOnly && !metadata.supportsPaxosOperations())
+            throw new IllegalArgumentException(String.format("Cannot run paxos only repair on %s.%s, which isn't configured for paxos operations", cfs.keyspace.getName(), cfs.name));
```
dcapwell, line 115: would be nice to check this in coordinator validation, if we are repairing multiple tables and only one hits this condition we have a lot of wasted work

src/java/org/apache/cassandra/repair/RepairJob.java
```diff
@@ -91,14 +99,24 @@ public class CassandraRepairJob extends AbstractRepairJob
...
+
+        if (session.accordOnly && !metadata.requiresAccordSupport())
+            throw new IllegalArgumentException(String.format("Cannot run accord only repair on %s.%s, which isn't configured for accord operations", cfs.keyspace.getName(), cfs.name));
```
dcapwell, line 118: same comment, we should check this at the start of repair validation

src/java/org/apache/cassandra/repair/RepairJob.java
```diff
@@ -114,26 +132,41 @@ public long getNowInSeconds()
...
         allEndpoints.add(ctx.broadcastAddressAndPort());
 
+        TableMetadata metadata = cfs.metadata();
+
         Future<List<TreeResponse>> treeResponses;
         Future<Void> paxosRepair;
         Epoch repairStartingEpoch = ClusterMetadata.current().epoch;
-        boolean doPaxosRepair = paxosRepairEnabled() && (((useV2() || isMetadataKeyspace()) && session.repairPaxos) || session.paxosOnly);
+
+        Preconditions.checkArgument(!session.paxosOnly || !session.accordOnly);
```
dcapwell, line 158: can we move this to the repair coordinator?  Also can we fail with a useful error message for operators?

src/java/org/apache/cassandra/repair/RepairJob.java
```diff
@@ -164,19 +198,59 @@ public void onFailure(Throwable t)
             return;
         }
 
+        Future<Void> accordRepair;
+        if (doAccordRepair)
+        {
+            accordRepair = paxosRepair.flatMap(unused -> {
+                logger.info("{} {}.{} starting accord repair", session.previewKind.logPrefix(session.getId()), desc.keyspace, desc.columnFamily);
+                IPartitioner partitioner = desc.ranges.iterator().next().left.getPartitioner();
```
dcapwell:
```suggestion
                IPartitioner partitioner = metadata.getPartitioner();
```

rather than walking ranges, we know the table why not pull there?

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -265,7 +265,7 @@ private AccordService(Id localId)
     {
         Invariants.checkState(localId != null, "static localId must be set before instantiating AccordService");
         logger.info("Starting accord with nodeId {}", localId);
-        AccordAgent agent = new AccordAgent();
+        AccordAgent agent = FBUtilities.construct(System.getProperty("cassandra.accord.agent", AccordAgent.class.getName()), "AccordAgent");
```
dcapwell: this will no longer build as we don't allow random system properties in the code anymore; check style *should* block this.  Need to migrate to the related properties enum.

dcapwell: Also, if this is for testing and not for pluggability, maybe `cassandra.test` prefix?

src/java/org/apache/cassandra/service/accord/repair/AccordRepair.java
```diff
@@ -31,65 +32,93 @@
...
     private volatile Throwable shouldAbort = null;
 
-    public AccordRepairJob(RepairSession repairSession, String cfname)
+    public AccordRepair(Listener listener, IPartitioner partitioner, String keyspace, Collection<Range<Token>> ranges)
     {
-        super(repairSession, cfname);
-        IPartitioner partitioner = desc.ranges.iterator().next().left.getPartitioner();
-        this.ranges = AccordTopology.toAccordRanges(desc.keyspace, desc.ranges);
-        this.splitter = partitioner.accordSplitter().apply(ranges);
+        this.listener = listener != null ? listener : Listener.NOOP;
```
dcapwell: `listener` is currently null, we do the following

```
AccordRepair repair = new AccordRepair(null, partitioner, desc.keyspace, desc.ranges);
```

This logic dropped metrics in favor of listeners, but doesn't have the listeners do anything yet...

src/java/org/apache/cassandra/service/accord/repair/AccordRepair.java
```diff
@@ -150,20 +179,17 @@ private void repairRange(TokenRange range) throws Throwable
             }
             catch (RuntimeException e)
             {
-                // TODO Placeholder for dependency limit overflow
-//                dependencyOverflow = true;
-                cfs.metric.rangeMigrationDependencyLimitFailures.mark();
```
dcapwell, line 155: why drop metric?

src/java/org/apache/cassandra/service/accord/repair/AccordRepair.java
```diff
@@ -150,20 +179,17 @@ private void repairRange(TokenRange range) throws Throwable
...
+                listener.onBarrierException(e);
                 throw e;
             }
             catch (Throwable t)
             {
-                // unexpected error
-                cfs.metric.rangeMigrationUnexpectedFailures.mark();
```
dcapwell, line 161: why drop metric?

src/java/org/apache/cassandra/service/accord/repair/AccordRepair.java
```diff
@@ -150,20 +179,17 @@ private void repairRange(TokenRange range) throws Throwable
...
+                listener.onBarrierException(t);
                 throw new RuntimeException(t);
             }
             finally
             {
-                cfs.metric.rangeMigration.addNano(start);
```
dcapwell, line 166: why drop metric?  Also I think it forgot to do `end - start`?

dcapwell (changes requested):

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
```diff
@@ -246,7 +246,7 @@ public void finished()
         checkNotNull(minVersion, "Unable to determine minimum cluster version");
         IAccordService accordService = AccordService.instance();
         if (session.streamOperation().requiresBarrierTransaction()
-            && cfs.metadata().isAccordEnabled()
+            && cfs.metadata().requiresAccordSupport()
```
aweisberg, line 249: Need to puzzle through here if there is a race similar to mutations where the mutation is supposed to write through Accord, but doesn't and so we don't do a barrier here when need to.

src/java/org/apache/cassandra/repair/RepairJob.java
```diff
@@ -163,19 +196,59 @@ public void onFailure(Throwable t)
...
+            accordRepair.addCallback(new FutureCallback<Void>()
+            {
+                public void onSuccess(Void ignored)
+                {
+                    logger.info("{} {}.{} accord repair completed", session.previewKind.logPrefix(session.getId()), desc.keyspace, desc.columnFamily);
+                    trySuccess(new RepairResult(desc, Collections.emptyList(), ConsensusMigrationRepairResult.fromAccordOnlyRepair(repairStartingEpoch, session.excludedDeadNodes)));
+                }
+
+                /**
+                 * Snapshot, validation and sync failures are all handled here
```
aweisberg: Comment is wrong I think

src/java/org/apache/cassandra/repair/RepairJob.java
```diff
@@ -163,19 +196,59 @@ public void onFailure(Throwable t)
...
         }
 
+        Future<Void> accordRepair;
+        if (doAccordRepair)
+        {
+            accordRepair = paxosRepair.flatMap(unused -> {
+                logger.info("{} {}.{} starting accord repair", session.previewKind.logPrefix(session.getId()), desc.keyspace, desc.columnFamily);
+                IPartitioner partitioner = desc.ranges.iterator().next().left.getPartitioner();
+                AccordRepair repair = new AccordRepair(null, partitioner, desc.keyspace, desc.ranges);
+                return repair.repair(taskExecutor);
```
aweisberg, line 203: To confirm my understanding, the Accord repair will complete before the regular repair so everything it puts in memtables will be included in the regular repair?

src/java/org/apache/cassandra/repair/messages/RepairOption.java
```diff
@@ -199,21 +198,16 @@ public static RepairOption parse(Map<String, String> options, IPartitioner parti
...
             Preconditions.checkArgument(!repairPaxos, "repairPaxos must be set to false for preview repairs");
             Preconditions.checkArgument(!paxosOnly, "paxosOnly must be set to false for preview repairs");
-            Preconditions.checkArgument(!accordRepair, "accordRepair must be set to false for preview repairs");
-        }
-
-        if (accordRepair)
-        {
-            Preconditions.checkArgument(!paxosOnly, "paxosOnly must be set to false for Accord repairs");
-            Preconditions.checkArgument(previewKind == PreviewKind.NONE, "Can't perform preview repair with an Accord repair");
-            Preconditions.checkArgument(!force, "Accord repair only requires a quorum to work so force is not supported");
```
aweisberg, line 215: Is the force check done anywhere else?

We should still be checking for dead nodes so technically yes we can run with force and only fail if it actually does something, but that might be less useful then just failing immediately.

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationRepairResult.java
```diff
@@ -33,18 +34,30 @@ private ConsensusMigrationRepairResult(ConsensusMigrationRepairType type, Epoch
...
+        if (deadNodesExcluded) return INELIGIBLE;
+        if (paxosRepaired && accordRepaired) return new ConsensusMigrationRepairResult(ConsensusMigrationRepairType.either, minEpoch);
+        if (paxosRepaired) return new ConsensusMigrationRepairResult(ConsensusMigrationRepairType.paxos, minEpoch);
+        if (accordRepaired) return new ConsensusMigrationRepairResult(ConsensusMigrationRepairType.accord, minEpoch);
+        return INELIGIBLE;
+    }
+
+    public static ConsensusMigrationRepairResult fromPaxosOnlyRepair(Epoch minEpoch, boolean deadNodesExcluded)
+    {
+        return fromRepair(minEpoch, true, false, deadNodesExcluded);
```
aweisberg: I think this is mixing up how Paxos repair needs to work unless I am wrong. Paxos repair doesn't finish at `ALL` and the requirement for Accord for migration is that it be done at `ALL` so that eventually Accord can safely switch to single replica reads.

That is why previously the Paxos only repair was not marked as migration eligible.

If Paxos repair actually applies everything at all then this would work, but I recall that it doesn't because that wouldn't be highly available.

src/java/org/apache/cassandra/tools/nodetool/Repair.java
```diff
@@ -105,6 +105,9 @@ public class Repair extends NodeToolCmd
     @Option(title = "paxos-only", name = {"-paxos-only", "--paxos-only"}, description = "If the --paxos-only flag is included, no table data is repaired, only paxos operations..")
     private boolean paxosOnly = false;
 
+    @Option(title = "accord-only", name = {"-accord-only", "--accord-only"}, description = "If the --accord-only flag is included, no table data is repaired, only accord operations..")
+    private boolean accordOnly = false;
```
aweisberg, line 109: Last we discussed this you actually pushed to not expose this to users in repair and I still agree with that. An accord-only repair isn't something useful to end users and we added a separate finish migration step to the JMX interface to handle invoking the Accord only repair.

We definitely should as part of repair do the Accord repairs to help maintain the contract people rely on where the repaired set is safe to read at CL=1.

test/distributed/org/apache/cassandra/distributed/test/accord/AccordMigrationTest.java
```diff
@@ -570,7 +570,7 @@ public void testAccordToPaxos() throws Exception
                  assertTargetPaxosWrite(runCasNoApply, 1, paxosMigratingKeys.next(), 2, 1, 1, 1, 1);
 
                  // Repair the currently migrating range from when targets were switched, but it's not an Accord repair, this is to make sure the wrong repair type doesn't trigger progress
-                 nodetool(coordinator, "repair", "-st", upperMidToken.toString(), "-et", maxAlignedWithLocalRanges.toString());
+                 nodetool(coordinator, "repair", "-st", upperMidToken.toString(), "-et", maxAlignedWithLocalRanges.toString(), "--paxos-only");
```
aweisberg, line 573: Paxos only is not sufficient here, we need to repair both Paxos and the base table for migration. We are also trying to address partially applied mutations that aren't from Paxos.

aweisberg (changes requested):
Big improvement having the Accord repair sequenced before the regular repair so that regular repair can ensure the accord stuff is applied at CL.ALL since it will only do it at quorum.

Some correctness issues primarily around Paxos repair alone not being sufficient for migration.

src/java/org/apache/cassandra/service/accord/repair/AccordRepair.java
bdeggleston: I'd initially intended to keep accord migration and normal accord repairs separate, which didn't end up making sense. I'll rename and restore the metrics

src/java/org/apache/cassandra/repair/RepairJob.java
bdeggleston: Added this check (and the others) to RepairCoordinator, though I'll leave them here just in case

src/java/org/apache/cassandra/db/streaming/CassandraStreamReceiver.java
bdeggleston: This probably counts as a separate migration/interop jira, but we could probably pass along the epoch in the StreamInitMessage and wait to be caught up locally before responding. That should make schema change -> migration repair safe for this case. There may still be a race cause be changing the schema or completing a migration while an unrelated streaming session is in progress though

src/java/org/apache/cassandra/repair/RepairJob.java
bdeggleston: yeah looks like that was copied incorrectly when I did the initial paxos repair work and kept copying it here :)
removed from both

src/java/org/apache/cassandra/repair/RepairJob.java
bdeggleston: yes that's right

src/java/org/apache/cassandra/service/accord/repair/AccordRepair.java
```diff
@@ -120,7 +135,7 @@ private void repairRange(TokenRange range) throws Throwable
             BigInteger remaining = rangeSize.subtract(offset);
             BigInteger length = remaining.min(rangeStep);
 
-            long start = ctx.clock().nanoTime();
+            long start = Clock.Global.nanoTime();
```
dcapwell:
```suggestion
            long start = ctx.clock().nanoTime();
```

src/java/org/apache/cassandra/service/accord/repair/AccordRepair.java
```diff
@@ -150,20 +165,18 @@ private void repairRange(TokenRange range) throws Throwable
...
             {
-                // unexpected error
-                cfs.metric.rangeMigrationUnexpectedFailures.mark();
+                cfs.metric.accordRepairUnexpectedFailures.mark();
                 throw new RuntimeException(t);
             }
             finally
             {
-                cfs.metric.rangeMigration.addNano(start);
+                long end = Clock.Global.nanoTime();
```
dcapwell:
```suggestion
                long end = ctx.clock().nanoTime();
```

dcapwell (commented):
finished review, only blocking comment is the usage of `Clock.Global`

dcapwell (approved):

src/java/org/apache/cassandra/repair/messages/RepairOption.java
bdeggleston: Force would still be useful for accord tables if some nodes were down. The barrier that's run and is considered an accord repair only requires a quorum. So running a force repair on an accord table with a node down would still apply the barrier and stream between available nodes. It would fail pretty quickly if there weren't enough live nodes to complete the barrier though

src/java/org/apache/cassandra/tools/nodetool/Repair.java
bdeggleston: I think my objection was exposing the initial iteration of accord repair as a repair, since it didn't sync data on disk between all nodes. An accord _only_ repair is a little different, it's basically the accord version of a paxos-only repair. Both paxos and accord only repairs only complete pending operations and do not sync data.

I'm also don't agree that accord-only repairs wouldn't be useful to end users. For instance, one of the use cases of paxos repairs is to establish a lower bound for applied paxos operations so that data can be exported to an ldap system for processing. Having said that, the accord repair is not on par with paxos repair since it issues a barrier, but does not wait for it to be applied on all nodes, and I think that's something I should add before committing

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationRepairResult.java
bdeggleston: good catch, thanks. Fixed

src/java/org/apache/cassandra/repair/messages/RepairOption.java
aweisberg: Yeah that's true the repair succeeding implies quorum was achieved.

src/java/org/apache/cassandra/repair/RepairCoordinator.java
```diff
@@ -282,12 +283,29 @@ public void run()
...
+        if (options.paxosOnly() && options.accordOnly())
+            throw new IllegalArgumentException("Cannot specify a repair as both paxos only and accord only");
+
+        for (ColumnFamilyStore cfs : columnFamilies)
+        {
+            TableMetadata metadata = cfs.metadata();
+            if (options.paxosOnly() && !metadata.supportsPaxosOperations())
+                throw new IllegalArgumentException(String.format("Cannot run paxos only repair on %s.%s, which isn't configured for paxos operations", cfs.keyspace.getName(), cfs.name));
+
+            if (options.accordOnly() && !metadata.requiresAccordSupport())
```
aweisberg, line 297: As an operator this might be painful because it requires you to know which kind you should invoke when in reality you just want to run the active consensus system's repair and you don't care which. I think we should split the ranges and then run the appropriate repair for each.

If we can make existing external tooling continue to "just work" we definitely should.

src/java/org/apache/cassandra/repair/RepairSession.java
```diff
@@ -166,13 +170,14 @@ public RepairSession(SharedContext ctx,
                          boolean optimiseStreams,
                          boolean repairPaxos,
                          boolean paxosOnly,
-                         boolean accordRepair,
+                         boolean accordOnly, boolean isConsensusMigration,
```
aweisberg: Comment parameter list should probably be updated

src/java/org/apache/cassandra/repair/messages/RepairOption.java
```diff
@@ -313,21 +307,23 @@ else if (ranges.isEmpty())
     private final boolean repairPaxos;
     private final boolean paxosOnly;
 
-    private final boolean accordRepair;
+    private final boolean accordOnly;
+    private final boolean isConsensusMigration;
```
aweisberg, line 311: `isConsensusMigration` might be too declarative at the level of specifying what a repair should do. Maybe it should just indicate that this is all vs quorum for Accord.

src/java/org/apache/cassandra/service/accord/repair/RepairSyncPointAdapter.java
```diff
@@ -0,0 +1,69 @@
...
+import accord.primitives.FullRoute;
+import accord.primitives.Seekables;
+import accord.primitives.SyncPoint;
+import accord.primitives.Timestamp;
+import accord.primitives.Txn;
+import accord.primitives.TxnId;
+import accord.primitives.Writes;
+import accord.topology.Topologies;
+
+public class RepairSyncPointAdapter<S extends Seekables<?, ?>> extends CoordinationAdapter.Adapters.AbstractSyncPointAdapter<S>
```
aweisberg, line 51: Needs a class level summary about what it is trying to do. Looks like this is how we are adding `ALL` to barriers? 

I am not sure this is 100% safe since the required list of endpoints could have an endpoint that will never be included in the actual transaction because of a topology change.

Will Accord eventually finish the transaction?

src/java/org/apache/cassandra/tools/nodetool/Repair.java
aweisberg: That makes sense, but just to make sure we are on the same page, Barriers used for repair should wait for application, they should call `ApplyThenWaitUntilApplied` after gathering dependencies. There is the exclusive sync points which don't wait, and instead populate that map that rejects transactions that are too old.

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -341,16 +357,17 @@ public IVerbHandler<? extends Request> verbHandler()
...
         AccordClientRequestMetrics metrics = isForWrite ? accordWriteMetrics : accordReadMetrics;
         TxnId txnId = null;
         try
         {
             logger.debug("Starting barrier key: {} epoch: {} barrierType: {} isForWrite {}", keysOrRanges, epoch, barrierType, isForWrite);
             txnId = node.nextTxnId(Kind.SyncPoint, keysOrRanges.domain());
-            AsyncResult<Timestamp> asyncResult = node.barrier(keysOrRanges, epoch, barrierType);
+            AsyncResult<Timestamp> asyncResult = syncPoint == null
+                                                 ? node.barrier(keysOrRanges, epoch, barrierType)
+                                                 : Barrier.barrier(node, keysOrRanges, epoch, barrierType, syncPoint);
```
aweisberg, line 370: One version launches straight into `Barrier.barrier` the other goes through `node.barrier`. Should we pick a single level of interfacing with Accord here?

aweisberg (changes requested):
Looks good. I am basically +1 I just have one concern.

`RequiredResponseTracker` might not have the same nodes passed in from repair that Accord ends up contacting. It really seems like the node list should come from Accord based on what it actually decides to contact?

The other thing is just that requiring operators to precisely know which consensus system is managing which table/range at any given point in time in order to invoke repair is a big burden. We should probably handle it for them.

src/java/org/apache/cassandra/repair/RepairCoordinator.java
aweisberg: Thinking a little harder about this being the `X only` case which is unusual, it still seems like we shouldn't expose the need to determine which ranges belong to which system outside of C*.

If you want to make this a TODO and file a JIRA we can kick the can down the road on this one.

src/java/org/apache/cassandra/repair/messages/RepairOption.java
bdeggleston: I don't really agree. Understanding what consistency repair should perform a given operation at seems is a bit specialized for someone reading the code for repair or consensus migration tool.

src/java/org/apache/cassandra/service/accord/repair/RepairSyncPointAdapter.java
bdeggleston: It should finish the transaction. WRT blocking, each shard tracker only blocks on the intersection of it's electorate and the endpoints provided. So if you, hypothetically, created a RepairSyncPointAdapter with a totally disjoint set of endpoints, it wouldn't block on anything

src/java/org/apache/cassandra/repair/messages/RepairOption.java
aweisberg: Totally reasonable. I think a comment somewhere explaining that consensus migration wants to be highly available so it can occur at QUORUM while repair wants to ensure visibility of data before the repair started at all replicas.

Really migration being highly available is just icing on the cake since we can theoretically stay migrating indefinitely and it is fine. It just works that way because it's additional work to add all which you have now done.

I think it's worth keeping just because we don't really know the downsides of being in migration yet.

aweisberg (approved):

src/java/org/apache/cassandra/repair/RepairCoordinator.java
bdeggleston: opened https://issues.apache.org/jira/browse/CASSANDRA-19588

PR #3180, opened 2024-03-16
CASSANDRA-19477 trunk dont go to disk for hints sizes

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/hints/HintsDescriptor.java
+++ b/src/java/org/apache/cassandra/hints/HintsDescriptor.java
@@ -81,6 +81,8 @@ final class HintsDescriptor
     final UUID hostId;
     final int version;
     final long timestamp;
+    final String hintsFileName;
+    final String crc32FileName;
 
     final ImmutableMap<String, Object> parameters;
     final ParameterizedClass compressionConfig;
@@ -93,6 +95,8 @@ final class HintsDescriptor
         this.hostId = hostId;
         this.version = version;
         this.timestamp = timestamp;
+        hintsFileName = hostId + "-" + timestamp + '-' + version + ".hints";
+        crc32FileName = hostId + "-" + timestamp + '-' + version + ".crc32";
         compressionConfig = createCompressionConfig(parameters);
 
         EncryptionData encryption = createEncryption(parameters);
@@ -205,12 +209,12 @@ final class HintsDescriptor
 
     String fileName()
     {
-        return String.format("%s-%s-%s.hints", hostId, timestamp, version);
+        return hintsFileName;
     }
 
     String checksumFileName()
     {
-        return String.format("%s-%s-%s.crc32", hostId, timestamp, version);
+        return crc32FileName;
     }
 
     File file(File hintsDirectory)
@@ -221,6 +225,22 @@ final class HintsDescriptor
     File checksumFile(File hintsDirectory)
     {
         return new File(hintsDirectory, checksumFileName());
+    }
+
+    /** cached size of the represented hints file */
+    private transient volatile long hintsFileSize = -1L;
+
+    long hintsFileSize(File hintsDirectory)
+    {
+        long size = hintsFileSize;
+        if (size == -1L) // we may race and duplicate lookup the first time the size is being queried, but that is fine
+            hintsFileSize = size = file(hintsDirectory).length();
+        return size;
+    }
+
+    void hintsFileSize(long value)
+    {
+        hintsFileSize = value;
     }
 
     int messagingVersion()
--- a/src/java/org/apache/cassandra/hints/HintsStore.java
+++ b/src/java/org/apache/cassandra/hints/HintsStore.java
@@ -194,8 +194,8 @@ final class HintsStore
                 if (predicate.test(descriptor))
                 {
                     cleanUp(descriptor);
-                    delete(descriptor);
                     removeSet.add(descriptor);
+                    delete(descriptor);
                 }
             }
         }
@@ -235,7 +235,6 @@ final class HintsStore
         dispatchPositions.put(descriptor, inputPosition);
     }
 
-
     /**
      * @return the total size of all files belonging to the hints store, in bytes.
      */
@@ -243,9 +242,12 @@ final class HintsStore
     {
         long total = 0;
         for (HintsDescriptor descriptor : Iterables.concat(dispatchDequeue, corruptedFiles))
-        {
-            total += descriptor.file(hintsDirectory).length();
-        }
+            total += descriptor.hintsFileSize(hintsDirectory);
+
+        HintsWriter currentWriter = getWriter();
+        if (null != currentWriter)
+            total += currentWriter.descriptor().hintsFileSize(hintsDirectory);
+
         return total;
     }
 
--- a/src/java/org/apache/cassandra/hints/HintsWriter.java
+++ b/src/java/org/apache/cassandra/hints/HintsWriter.java
@@ -81,6 +81,7 @@ class HintsWriter implements AutoCloseable
             ByteBuffer descriptorBytes = dob.unsafeGetBufferAndFlip();
             updateChecksum(crc, descriptorBytes);
             channel.write(descriptorBytes);
+            descriptor.hintsFileSize(channel.position());
 
             if (descriptor.isEncrypted())
                 return new EncryptedHintsWriter(directory, descriptor, file, channel, fd, crc);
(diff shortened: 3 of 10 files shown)


clohfink:
+1 thanks for this!

src/java/org/apache/cassandra/hints/HintsStore.java
```diff
@@ -76,9 +77,13 @@ private HintsStore(UUID hostId, File hintsDirectory, ImmutableMap<String, Object
         dispatchDequeue = new ConcurrentLinkedDeque<>(descriptors);
         corruptedFiles = new ConcurrentLinkedQueue<>();
         hintsExpirations = new ConcurrentHashMap<>();
+        hintsSizes = new ConcurrentHashMap<>();
 
         //noinspection resource
         lastUsedTimestamp = descriptors.stream().mapToLong(d -> d.timestamp).max().orElse(0L);
+
+        for (HintsDescriptor descriptor : Iterables.concat(dispatchDequeue, corruptedFiles))
```
smiklosovic: it is not technically necessary to iterate over corrputedFiles because they are empty here

src/java/org/apache/cassandra/hints/HintsStore.java
```diff
@@ -61,6 +61,7 @@ final class HintsStore
     private final Deque<HintsDescriptor> dispatchDequeue;
     private final Queue<HintsDescriptor> corruptedFiles;
     private final Map<HintsDescriptor, Long> hintsExpirations;
+    private final Map<String, Long> hintsSizes;
```
iamaleksey: Why not `HintsDescriptor` as key?

src/java/org/apache/cassandra/hints/HintsWriteExecutor.java
```diff
@@ -260,19 +261,33 @@ private void flushInternal(Iterator<ByteBuffer> iterator, HintsStore store)
...
+                        break;
+                }
+            }
+            finally
             {
-                session.append(iterator.next());
-                if (session.position() >= maxHintsFileSize)
-                    break;
+                Map<String, Long> hintsSizes = store.getHintsSizes();
+                hintsSizes.compute(descriptor.hintsFileName, (d, currentSize) ->
```
iamaleksey: You have `session.position()` ~~and `computeIfAbsent()`~~.

src/java/org/apache/cassandra/hints/HintsStore.java
```diff
@@ -282,7 +297,11 @@ boolean isWriting()
     HintsWriter getOrOpenWriter()
     {
         if (hintsWriter == null)
+        {
             hintsWriter = openWriter();
+            hintsSizes.put(hintsWriter.descriptor().fileName(), hintsWriter.descriptor().file(hintsDirectory).length());
```
iamaleksey: This seems pretty unnecessary or at the very least unjustified.

src/java/org/apache/cassandra/hints/HintsWriteExecutor.java
iamaleksey: But really just `put()` is sufficient.

src/java/org/apache/cassandra/hints/HintsStore.java
```diff
@@ -203,14 +213,21 @@ private void deleteHints(Predicate<HintsDescriptor> predicate)
...
+                hintsSizes.remove(descriptor.fileName());
         }
     }
 
     void delete(HintsDescriptor descriptor)
     {
         File hintsFile = descriptor.file(hintsDirectory);
         if (hintsFile.tryDelete())
+        {
+            // this method is also called standalone (without deleteHints)
```
iamaleksey: You are calling this twice on the normal path. Just leave the one call in this method, in `finally` if need be.

src/java/org/apache/cassandra/hints/HintsWriteExecutor.java
iamaleksey: (This is a capturing lambda fyi)

iamaleksey:
I would just cache the size value in a transient field in `HintsDescriptor`. A CHM is an overkill here.

src/java/org/apache/cassandra/hints/HintsStore.java
smiklosovic: It was like that before, then I changed it to String. Is not this just better? I don't mind to revert this.

src/java/org/apache/cassandra/hints/HintsStore.java
smiklosovic: we do not include .crc32 files too btw ... I can remove this though. There is a test at the bottom which thinks that by mere writing of a descriptor it should has some size. I can rewrite that test accordingly.

iamaleksey: CRC files don't matter, an entirely different order of magnitude where total limits are involved.

src/java/org/apache/cassandra/hints/HintsStore.java
```diff
@@ -282,7 +293,11 @@ boolean isWriting()
     HintsWriter getOrOpenWriter()
     {
         if (hintsWriter == null)
+        {
             hintsWriter = openWriter();
+            hintsSizes.put(hintsWriter.descriptor(), hintsWriter.descriptor().file(hintsDirectory).length());
```
iamaleksey: I don't think we need to do this.

src/java/org/apache/cassandra/hints/HintsWriteExecutor.java
```diff
@@ -260,19 +261,30 @@ private void flushInternal(Iterator<ByteBuffer> iterator, HintsStore store)
...
+                    if (session.position() >= maxHintsFileSize)
+                        break;
+                }
+            }
+            finally
             {
-                session.append(iterator.next());
-                if (session.position() >= maxHintsFileSize)
-                    break;
+                store.getHintsSizes().put(descriptor, session.position());
```
iamaleksey: No need to put this in a finally block I'd say.

src/java/org/apache/cassandra/hints/HintsWriter.java
```diff
@@ -259,7 +261,7 @@ void append(Hint hint) throws IOException
             if (hintBuffer == buffer)
                 bytesWritten += totalSize;
             else
-                append((ByteBuffer) hintBuffer.flip());
```
smiklosovic, line 262: not necessary

src/java/org/apache/cassandra/hints/HintsWriter.java
```diff
@@ -224,7 +226,7 @@ void append(ByteBuffer hint) throws IOException
          * Serializes and appends the hint (with CRC included) to this session's aggregation buffer,
          * writes to the underlying channel when the buffer is overflown.
          *
-         * Used mainly by tests and {@link LegacyHintsMigrator}
```
smiklosovic, line 227: does not exist anymore

src/java/org/apache/cassandra/hints/HintsWriter.java
```diff
@@ -271,6 +273,7 @@ public void close() throws IOException
             flushBuffer();
             maybeFsync();
             maybeSkipCache();
+            descriptor.size = position();
```
smiklosovic: why not to just update it in close() ? Then when this is used in try-with-resources in HintsWriteExecutor, we do not need to change there anything at all and not mess with finally etc ...

smiklosovic: this should be probably in finally though, not sure ... as all above can throw

src/java/org/apache/cassandra/hints/HintsWriter.java
iamaleksey: Not anymore in Java 11, was needed in Java 8 :)

src/java/org/apache/cassandra/hints/HintsDescriptor.java
```diff
@@ -81,18 +81,26 @@ final class HintsDescriptor
...
 
     final ImmutableMap<String, Object> parameters;
     final ParameterizedClass compressionConfig;
 
     private final Cipher cipher;
     private final ICompressor compressor;
 
+    // size of underlying hint file (without descriptor header and crc file sizes)
+    // the value is populated in Session's close method in HintsWriter
+    volatile long size = 0;
```
iamaleksey: `size` -> `fileSize` slightly more descriptive. May also mark it as `transient`, to borrow the Accord convention (for the reader, not for the JVM).

smiklosovic: why not

PR #3185, opened 2024-03-20
CASSANDRA-19461 Ensure SAI indexes empty byte buffers for types that allow them as a valid input

maedhroz:
patch by Caleb Rackliffe; reviewed by ? for CASSANDRA-19461

### DIFF
--- a/src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
+++ b/src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
@@ -242,7 +242,7 @@ public final class CreateIndexStatement extends AlterSchemaStatement
             throw ire(ONLY_PARTITION_KEY, column);
 
         if (column.type.isFrozenCollection() && target.type != Type.FULL)
-            throw ire(CREATE_ON_FROZEN_COLUMN, target.type, column, column);
+            throw ire(CREATE_ON_FROZEN_COLUMN, target.type, column, column.name.toCQLString());
 
         if (!column.type.isFrozenCollection() && target.type == Type.FULL)
             throw ire(FULL_ON_FROZEN_COLLECTIONS);
--- a/src/java/org/apache/cassandra/index/sai/disk/v1/SSTableIndexWriter.java
+++ b/src/java/org/apache/cassandra/index/sai/disk/v1/SSTableIndexWriter.java
@@ -200,7 +200,8 @@ public class SSTableIndexWriter implements PerColumnIndexWriter
             currentBuilder = newSegmentBuilder();
         }
 
-        if (term.remaining() == 0) return;
+        // Some types support empty byte buffers:
+        if (term.remaining() == 0 && !index.termType().indexType().allowsEmpty()) return;
 
         if (analyzer == null || !index.termType().isLiteral())
         {
--- a/src/java/org/apache/cassandra/index/sai/memory/MemtableIndex.java
+++ b/src/java/org/apache/cassandra/index/sai/memory/MemtableIndex.java
@@ -28,6 +28,7 @@ import java.util.function.Function;
 import org.apache.cassandra.db.Clustering;
 import org.apache.cassandra.db.DecoratedKey;
 import org.apache.cassandra.db.PartitionPosition;
+import org.apache.cassandra.db.marshal.AbstractType;
 import org.apache.cassandra.dht.AbstractBounds;
 import org.apache.cassandra.index.sai.QueryContext;
 import org.apache.cassandra.index.sai.StorageAttachedIndex;
@@ -46,10 +47,12 @@ public class MemtableIndex implements MemtableOrdering
     private final MemoryIndex memoryIndex;
     private final LongAdder writeCount = new LongAdder();
     private final LongAdder estimatedMemoryUsed = new LongAdder();
+    private final AbstractType<?> type;
 
     public MemtableIndex(StorageAttachedIndex index)
     {
         this.memoryIndex = index.termType().isVector() ? new VectorMemoryIndex(index) : new TrieMemoryIndex(index);
+        this.type = index.termType().indexType();
     }
 
     public long writeCount()
@@ -79,7 +82,7 @@ public class MemtableIndex implements MemtableOrdering
 
     public long index(DecoratedKey key, Clustering<?> clustering, ByteBuffer value)
     {
-        if (value == null || value.remaining() == 0)
+        if (value == null || (value.remaining() == 0 && !type.allowsEmpty()))
             return 0;
 
         long ram = memoryIndex.add(key, clustering, value);
--- a/test/unit/accord/utils/Gens.java
+++ b/test/unit/accord/utils/Gens.java
@@ -372,7 +372,12 @@ public class Gens {
         @Override
         public ListDSL<T> unique()
         {
-            return new ListDSL<>(new GenReset<>(fn));
+            return new ListDSL<>(new GenReset<>(fn, false));
+        }
+
+        public ListDSL<T> uniqueBestEffort()
+        {
+            return new ListDSL<>(new GenReset<>(fn, true));
         }
 
         @Override
@@ -384,7 +389,16 @@ public class Gens {
                 int size = sizeGen.nextInt(r);
                 List<T> list = new ArrayList<>(size);
                 for (int i = 0; i < size; i++)
-                    list.add(fn.next(r));
+                {
+                    try
+                    {
+                        list.add(fn.next(r));
+                    }
+                    catch (IgnoreGenResult e)
+                    {
+                        // ignore
+                    }
+                }
                 return list;
             };
         }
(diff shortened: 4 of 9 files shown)


src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
```diff
@@ -242,7 +242,7 @@ private void validateIndexTarget(TableMetadata table, IndexMetadata.Kind kind, I
             throw ire(ONLY_PARTITION_KEY, column);
 
         if (column.type.isFrozenCollection() && target.type != Type.FULL)
-            throw ire(CREATE_ON_FROZEN_COLUMN, target.type, column, column);
+            throw ire(CREATE_ON_FROZEN_COLUMN, target.type, column, column.name.toCQLString());
```
dcapwell, line 245: not for this patch: honestly, when do we want toString to not be valid CQL?  this is a common issue...

src/java/org/apache/cassandra/index/sai/disk/v1/SSTableIndexWriter.java
```diff
@@ -200,7 +200,8 @@ else if (shouldFlush(sstableRowId))
             currentBuilder = newSegmentBuilder();
         }
 
-        if (term.remaining() == 0) return;
+        // Some types support empty byte buffers:
+        if (term.remaining() == 0 && !index.termType().indexType().allowsEmpty()) return;
```
dcapwell, line 204: yay, a non-test code path using this new method!  Glad I added it ^_^... semantics are hard...

src/java/org/apache/cassandra/index/sai/memory/MemtableIndex.java
```diff
@@ -79,7 +82,7 @@ public ByteBuffer getMaxTerm()
 
     public long index(DecoratedKey key, Clustering<?> clustering, ByteBuffer value)
     {
-        if (value == null || value.remaining() == 0)
+        if (value == null || (!type.allowsEmpty() && value.remaining() == 0))
```
dcapwell:
```suggestion
        if (value == null || (value.remaining() == 0 && !type.allowsEmpty()))
```

nit: we only want to call `allowsEmpty` when its empty, else it adds a virtual dispatch that is `false` in most cases... so slightly less overhead in the common case (non-empty)

test/unit/accord/utils/Gens.java
```diff
@@ -372,7 +372,12 @@ public ListDSL(Gen<T> fn) {
         @Override
         public ListDSL<T> unique()
         {
-            return new ListDSL<>(new GenReset<>(fn));
+            return new ListDSL<>(new GenReset<>(fn, false));
+        }
+
+        public ListDSL<T> uniqueBestEffort()
```
dcapwell, line 378: not for this patch: going to have to do the same in accord... before we merge accord to trunk we are going to have to account for the drift between the 2 repos....

test/unit/org/apache/cassandra/index/sai/cql/AbstractSimpleEqTest.java
```diff
@@ -0,0 +1,89 @@
...
+import org.assertj.core.api.Assertions;
+
+import static accord.utils.Property.qt;
+
+public abstract class AbstractSimpleEqTest extends SAITester
+{
+    static
+    {
+        // The info table gets updated and force-flushed every time we truncate, so disable that and avoid the overhead:
+        CassandraRelevantProperties.UNSAFE_SYSTEM.setBoolean(true);
```
dcapwell:
```suggestion
        CassandraRelevantProperties.UNSAFE_SYSTEM.setBoolean(true);
        // ignore SAI timeouts as this is just validating the read/write logic of the index, will help make the test more stable
        CassandraRelevantProperties.SAI_TEST_DISABLE_TIMEOUT.setBoolean(true);
```

just in case we are slow, disabling sai timeouts will make this less flakey...

test/unit/org/apache/cassandra/index/sai/cql/AbstractSimpleEqTest.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    protected void test(AbstractType<?> type, @Nullable Long seed, int examples, Gen<Gen<ByteBuffer>> distribution)
+    {
+        createTable("CREATE TABLE %s (pk int PRIMARY KEY, value " + type.asCQL3Type() + ')');
```
dcapwell: this tests the index correctness, but it might be good to also change the index column away from regular to static/cluster/partition columns based off the randomness...

Rather than having 1 test for the type, we can have 1 test for each type of column?

this feedback is optional, just thinking it could improve coverage

test/unit/org/apache/cassandra/index/sai/cql/AbstractSimpleEqTest.java
```diff
@@ -0,0 +1,89 @@
...
+            Map<ByteBuffer, IntArrayList> termIndex = new TreeMap<>();
+
+            for (int i = 0; i < 1000; i++)
+            {
+                ByteBuffer term = support.next(rs);
+                execute("INSERT INTO %s (pk, value) VALUES (?, ?)", i, term);
+                termIndex.computeIfAbsent(term, ignore -> new IntArrayList()).addInt(i);
+            }
+            
+            flush();
```
dcapwell: to make sure we test memtable index as well, we should loop over the for loop 2 times: first is memtable, second is disk...

```
for (int I = 0; I < 2; I++)
{
  if (I == 1) flush();
...
}
```

test/unit/org/apache/cassandra/index/sai/cql/AllTypesSimpleEqTest.java
```diff
@@ -0,0 +1,95 @@
...
+        this.type = type;
+    }
+
+    @Parameterized.Parameters(name = "{0}")
+    public static Collection<Object[]> data()
+    {
+        return StorageAttachedIndex.SUPPORTED_TYPES.stream()
+                                                   .map(CQL3Type::getType)
+                                                   // TODO: Track down unicode character edge cases...
+                                                   .filter(t -> t != UTF8Type.instance)
```
dcapwell, line 63: don't you need to filter ascii as well?

test/unit/org/apache/cassandra/index/sai/cql/EmptyStringLifecycleTest.java
```diff
@@ -0,0 +1,79 @@
...
+    {
+        createTable("CREATE TABLE %s (k int PRIMARY KEY, v text)");
+        disableCompaction(KEYSPACE);
+        createIndex(String.format(CREATE_INDEX_TEMPLATE, 'v'));
+
+        execute("INSERT INTO %s (k, v) VALUES (0, '')");
+        execute("INSERT INTO %s (k) VALUES (1)");
+        flush();
+
+        execute("INSERT INTO %s (k, v) VALUES (0, '')");
```
dcapwell: I think it would be nice here to have this be `1, ''`, and have the test validate it read from both values...

test/unit/org/apache/cassandra/index/sai/cql/EmptyStringLifecycleTest.java
```diff
@@ -0,0 +1,79 @@
...
+        flush();
+        
+        compact();
```
dcapwell, line 74: this doesn't validate we actually compacted, so would be good to validate we had 2 files now have 1... simple test bug can mean we don't actually compact (you disable compaction, so we actually no op right now...)

dcapwell (changes requested):

src/java/org/apache/cassandra/index/sai/disk/v1/SSTableIndexWriter.java
maedhroz: Having prescient visions over there?

test/unit/org/apache/cassandra/index/sai/cql/AllTypesSimpleEqTest.java
maedhroz: Ascii should pass without issue, since it doesn't have the problematic unicode characters.

test/unit/org/apache/cassandra/index/sai/cql/EmptyStringLifecycleTest.java
maedhroz: Talked about this offline. Disabling auto-compaction doesn't disable forced major compactions, which this is.

src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
maedhroz: It's just a matter of the quotes, right?

test/unit/org/apache/cassandra/index/sai/cql/AbstractSimpleEqTest.java
maedhroz: Didn't do that here since it's more or less the point of `testBeforeAndAfterFlush()`

test/unit/org/apache/cassandra/index/sai/cql/EmptyStringLifecycleTest.java
maedhroz: Fair enough...let's do it.

src/java/org/apache/cassandra/cql3/statements/schema/CreateIndexStatement.java
dcapwell: for the most part yah..

dcapwell (approved):

maedhroz:
Committed as https://github.com/apache/cassandra/commit/46acaf22e688e7a2e707ac61fd88c96ed33b60d7

PR #3196, opened 2024-03-22
CASSANDRA-19344 Range movements involving transient replicas must safely enact changes to read and write replica sets

beobal:
When an instance moves from a transient to a full replica for a given range, it must begin acting as a full replica for writes before it does so for reads. Otherwise, consistency can be violated as data streamed to the instance early in the operation can be removed by cleanup if it occurs before the instance assumes responsibility for full writes. Also, coordinators will route read requests to instances which may not have received all preceding writes, causing unnecessary read repair or potentially inconsistent results. 

The root cause of the flaky test failure which originally produced this issue was that in `TransientRangeMovementTest::testRemoveNode` cleanup happened to run on one instance before it had enacted the final step of the removal operation, leading it to remove more data than it should.

### DIFF
--- a/src/java/org/apache/cassandra/tcm/ownership/PlacementForRange.java
+++ b/src/java/org/apache/cassandra/tcm/ownership/PlacementForRange.java
@@ -336,7 +336,7 @@ public class PlacementForRange
          *
          * Where an endpoint is present in both groups, prefer the proposed iff it is a FULL replica. During a
          * multi-step operation (join/leave/move), we want any change from transient to full to happen as early
-         * as possible so that a replica being whose ownership is modified in this way becomes FULL for writes before it
+         * as possible so that a replica whose ownership is modified in this way becomes FULL for writes before it
          * becomes FULL for reads. This works as additions to write replica groups are applied before any other
          * placement changes (i.e. in START_[JOIN|LEAVE|MOVE]).
          *
--- a/src/java/org/apache/cassandra/tcm/ownership/PlacementForRange.java
+++ b/src/java/org/apache/cassandra/tcm/ownership/PlacementForRange.java
@@ -321,15 +321,49 @@ public class PlacementForRange
                                                (t, v) -> {
                                                    EndpointsForRange old = v.get();
                                                    return VersionedEndpoints.forRange(Epoch.max(v.lastModified(), replicas.lastModified()),
-                                                                                      replicas.get()
-                                                                                              .newBuilder(replicas.size() + old.size())
-                                                                                              .addAll(old)
-                                                                                              .addAll(replicas.get(), ReplicaCollection.Builder.Conflict.ALL)
-                                                                                              .build());
+                                                                                      combine(old, replicas.get()));
                                                });
             if (group == null)
                 replicaGroups.put(replicas.range(), replicas);
             return this;
+        }
+
+        /**
+         * Combine two replica groups, assuming one is the current group and the other is the proposed.
+         * During range movements this is used when calculating the maximal placement, which combines the current and
+         * future replica groups. This special cases the merging of two replica groups to make sure that when a replica
+         * moves from transient to full, it starts to act as a FULL write replica as early as possible.
+         *
+         * Where an endpoint is present in both groups, prefer the proposed iff it is a FULL replica. During a
+         * multi-step operation (join/leave/move), we want any change from transient to full to happen as early
+         * as possible so that a replica being whose ownership is modified in this way becomes FULL for writes before it
+         * becomes FULL for reads. This works as additions to write replica groups are applied before any other
+         * placement changes (i.e. in START_[JOIN|LEAVE|MOVE]).
+         *
+         * @param prev Initial set of replicas for a given range
+         * @param next Proposed set of replicas for the same range.
+         * @return The union of the two groups
+         */
+        private EndpointsForRange combine(EndpointsForRange prev, EndpointsForRange next)
+        {
+            Map<InetAddressAndPort, Replica> e1 = prev.byEndpoint();
+            Map<InetAddressAndPort, Replica> e2 = next.byEndpoint();
+            EndpointsForRange.Builder combined = prev.newBuilder(prev.size() + next.size());
+            e1.forEach((e, r1) -> {
+                Replica r2 = e2.get(e);
+                if (null == r2)          // not present in next
+                    combined.add(r1);
+                else if (r2.isFull())    // prefer replica from next, if it is moving from transient to full
+                    combined.add(r2);
+                else
+                    combined.add(r1);    // replica is moving from full to transient, or staying the same
+            });
+            // any new replicas not in prev
+            e2.forEach((e, r2) -> {
+                if (!combined.contains(e))
+                    combined.add(r2);
+            });
+            return combined.build();
         }
 
         public Builder withReplicaGroups(Iterable<VersionedEndpoints.ForRange> replicas)
--- a/src/java/org/apache/cassandra/tcm/ownership/PlacementTransitionPlan.java
+++ b/src/java/org/apache/cassandra/tcm/ownership/PlacementTransitionPlan.java
@@ -18,6 +18,23 @@
 
 package org.apache.cassandra.tcm.ownership;
 
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+import javax.annotation.Nullable;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.apache.cassandra.dht.Range;
+import org.apache.cassandra.dht.Token;
+import org.apache.cassandra.exceptions.ExceptionCode;
+import org.apache.cassandra.locator.InetAddressAndPort;
+import org.apache.cassandra.locator.RangesAtEndpoint;
+import org.apache.cassandra.locator.Replica;
+import org.apache.cassandra.schema.ReplicationParams;
+import org.apache.cassandra.tcm.Epoch;
+import org.apache.cassandra.tcm.Transformation;
 import org.apache.cassandra.tcm.sequences.LockedRanges;
 
 /**
@@ -32,6 +49,8 @@ import org.apache.cassandra.tcm.sequences.LockedRanges;
  */
 public class PlacementTransitionPlan
 {
+    private static final Logger logger = LoggerFactory.getLogger(MovementMap.class);
+
     public final PlacementDeltas toSplit;
     public final PlacementDeltas toMaximal;
     public final PlacementDeltas toFinal;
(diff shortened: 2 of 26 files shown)


src/java/org/apache/cassandra/tcm/ownership/PlacementForRange.java
```diff
@@ -307,17 +307,51 @@ public Builder withReplicaGroup(VersionedEndpoints.ForRange replicas)
...
 
+        /**
+         * Combine two replica groups, assuming one is the current group and the other is the proposed.
+         * During range movements this is used when calculating the maximal placement, which combines the current and
+         * future replica groups. This special cases the merging of two replica groups to make sure that when a replica
+         * moves from transient to full, it starts to act as a FULL write replica as early as possible.
+         *
+         * Where an endpoint is present in both groups, prefer the proposed iff it is a FULL replica. During a
+         * multi-step operation (join/leave/move), we want any change from transient to full to happen as early
+         * as possible so that a replica being whose ownership is modified in this way becomes FULL for writes before it
```
ifesdjeen: nit: Could you rephrase this part a bit?

> a replica being whose ownership is modified in this way becomes FULL for writes before it

I am a bit thrown off by replica being part.

src/java/org/apache/cassandra/tcm/ownership/PlacementForRange.java
```diff
@@ -307,17 +307,51 @@ public Builder withReplicaGroup(VersionedEndpoints.ForRange replicas)
...
+                Replica r2 = e2.get(e);
+                if (null == r2)          // not present in next
+                    combined.add(r1);
+                else if (r2.isFull())    // prefer replica from next, if it is moving from transient to full
+                    combined.add(r2);
+                else
+                    combined.add(r1);    // replica is moving from full to transient, or staying the same
+            });
+            // any new replicas not in prev
+            e2.forEach((e, r2) -> {
```
ifesdjeen, line 362: Would it make sense to construct a set to iterate over in the above loop?

src/java/org/apache/cassandra/tcm/ownership/PlacementForRange.java
```diff
@@ -307,17 +307,51 @@ public Builder withReplicaGroup(VersionedEndpoints.ForRange replicas)
...
+        private EndpointsForRange combine(EndpointsForRange prev, EndpointsForRange next)
+        {
+            Map<InetAddressAndPort, Replica> e1 = prev.byEndpoint();
+            Map<InetAddressAndPort, Replica> e2 = next.byEndpoint();
+            EndpointsForRange.Builder combined = prev.newBuilder(prev.size() + next.size());
+            e1.forEach((e, r1) -> {
+                Replica r2 = e2.get(e);
+                if (null == r2)          // not present in next
+                    combined.add(r1);
+                else if (r2.isFull())    // prefer replica from next, if it is moving from transient to full
```
ifesdjeen, line 356: nit: could probably short-circuit this with 

```
if (r2 != null && r2.isFull)
  combined.add(r2);
else
  combined.add(r1);
```

If you deliberately have split these, I do nit mind leaving as-is.

src/java/org/apache/cassandra/tcm/ownership/PlacementTransitionPlan.java
```diff
@@ -125,4 +144,78 @@ public String toString()
...
+                    // BUT we split and merge ranges so we need to check containment both ways
+                    for (Replica newReadReplica : addedReadReplicas)
+                    {
+                        if (existingWriteReplicas.contains(newReadReplica))
+                            continue;
+                        boolean contained = false;
+                        Set<Range<Token>> intersectingRanges = new HashSet<>();
+                        for (Replica writeReplica : existingWriteReplicas)
+                        {
+                            if (writeReplica.isFull() == newReadReplica.isFull() || writeReplica.isFull() && newReadReplica.isTransient())
```
ifesdjeen: nit: would you mind adding brackets here to make operator order more explicit here?

test/harry/main/org/apache/cassandra/harry/sut/TokenPlacementModel.java
```diff
@@ -199,41 +228,28 @@ public static List<Node> peerStateToNodes(Object[][] resultset)
 
     public static class NtsReplicationFactor extends ReplicationFactor
     {
-        private final Lookup lookup = new DefaultLookup();
```
ifesdjeen, line 202: oh thank you for removing this! This was an artifact of integrating with a simulator, but it never was actually used in this context.

ifesdjeen (approved):

src/java/org/apache/cassandra/tcm/sequences/RemoveNodeStreams.java
```diff
@@ -131,33 +131,12 @@ private static MovementMap movementMap(InetAddressAndPort leaving, ClusterMetada
...
                                            candidateBuilder.add(replica, ReplicaCollection.Builder.Conflict.NONE);
                                    });
-                                   movements.putAll(newReplica, candidateBuilder.build(), ReplicaCollection.Builder.Conflict.NONE);
+                                   EndpointsForRange sources = candidateBuilder.build();
+                                   // log if newReplica is an existing transient replica moving to a full replica
+                                   if (startWriteRemovals.get(newReplica.endpoint()).contains(newReplica.range(), false))
+                                       logger.debug("Streaming transient -> full conversion to {} from {}", newReplica.endpoint(), sources);
+                                   movements.putAll(newReplica, sources, ReplicaCollection.Builder.Conflict.NONE);
                                });
-            // and check if any replicas went from transient -> full:
```
ifesdjeen, line 136: it also happens that we do not need mid and finish delta for this method anymore. Should we remove them?

src/java/org/apache/cassandra/tcm/sequences/RemoveNodeStreams.java
```diff
@@ -131,33 +131,12 @@ private static MovementMap movementMap(InetAddressAndPort leaving, ClusterMetada
                                        if (!replica.endpoint().equals(leaving) && !replica.endpoint().equals(newReplica.endpoint()))
                                            candidateBuilder.add(replica, ReplicaCollection.Builder.Conflict.NONE);
                                    });
-                                   movements.putAll(newReplica, candidateBuilder.build(), ReplicaCollection.Builder.Conflict.NONE);
+                                   EndpointsForRange sources = candidateBuilder.build();
+                                   // log if newReplica is an existing transient replica moving to a full replica
```
ifesdjeen, line 133: We can make a similar change in `UnbootstrapStreams` calculation of the movement map, change the mid check to 

``` 
startWriteAdditions.get(removal.endpoint()).contains(removal.range(), true);
```

src/java/org/apache/cassandra/tcm/transformations/PrepareJoin.java
```diff
@@ -156,6 +153,12 @@ public Result execute(ClusterMetadata prev)
                                                              transitionPlan.toSplit,
                                                              startJoin, midJoin, finishJoin,
                                                              joinTokenRing, streamData);
+        if (!(this instanceof UnsafeJoin) && !prev.tokenMap.isEmpty())
```
ifesdjeen: Would you be OK to move this logic to an instance method, and have implemenation in PrepareJoin, and override it in UnsafeJoin? It would be great if we could keep it polymorphic.

src/java/org/apache/cassandra/tcm/transformations/PrepareMove.java
```diff
@@ -103,7 +104,9 @@ public Result execute(ClusterMetadata prev)
         StartMove startMove = new StartMove(nodeId, transitionPlan.addToWrites(), lockKey);
         MidMove midMove = new MidMove(nodeId, transitionPlan.moveReads(), lockKey);
         FinishMove finishMove = new FinishMove(nodeId, tokens, transitionPlan.removeFromWrites(), lockKey);
-
+        Result res = PlacementTransitionPlan.assertPreExistingWriteReplica(prev.placements, transitionPlan);
```
ifesdjeen: Do you folks think maybe having an Option here could be useful? Or some form of polymorphic result that would visually explain the significance of `null` equivalent?

test/distributed/org/apache/cassandra/distributed/test/TransientRangeMovement2Test.java
```diff
@@ -51,9 +59,30 @@ public void testMoveBackward() throws IOException, ExecutionException, Interrupt
...
-            cluster.get(3).nodetoolResult("move", "25").asserts().success();
+            // At the start, node1 is a TRANSIENT replica for (20,30] and FULL for (30, 40], (40,] & (,10]. When moving
+            // node3 to token 25, node1 becomes a FULL replica for (25, 40], effectively going from TRANSIENT to FULL
+            // for (25,30]. A T->F transition will always cause data for that range to be streamed to the transitioning
+            // node, which happens after StartMove and before MidMove. Running cleanup before node1 considers itself a
+            // FULL replica would remove any of the newly streamed data which is marked repaired. To avoid this, we
+            // ensure that any T->F transition is applied for writes as part of the StartMove. Have the CMS node (node1)
+            // pause before the MidMove step is committed, at which point we know that streaming has completed.
+            Callable<Epoch> pending = pauseBeforeCommit(cluster.get(1), (e) -> e instanceof PrepareMove.MidMove);
+            new Thread(() -> cluster.get(3).nodetoolResult("move", "25").asserts().success()).start();
```
ifesdjeen: Would you mind to add a `join` after we have finished with this thread below?

test/distributed/org/apache/cassandra/distributed/test/TransientRangeMovement2Test.java
```diff
@@ -51,9 +59,30 @@ public void testMoveBackward() throws IOException, ExecutionException, Interrupt
...
+
+            // Before allowing node1 to proceed with committing the MidMove, instruct it to pause node1 before it
```
ifesdjeen: nit: This line reads a bit tautological, would maybe something like "To gate/prevent node1 from proceeding with committing MidMove, instruct it to pause before enacting it" ?

test/harry/main/org/apache/cassandra/harry/sut/TokenPlacementModel.java
```diff
@@ -62,18 +75,34 @@ public ReplicatedRanges replicate(List<Node> nodes)
         public abstract ReplicatedRanges replicate(Range[] ranges, List<Node> nodes);
     }
 
+    public static class DCReplicas
+    {
+        public final int totalCount;
```
ifesdjeen, line 80: Nit: should we try keeping API here consistent with RF?

```
    public final int allReplicas;
    public final int fullReplicas;
```

test/distributed/org/apache/cassandra/distributed/test/log/PlacementSimulator.java
```diff
@@ -985,20 +977,84 @@ public Diff<T> invert()
...
+        List<Replica> removals = new ArrayList<>(unfiltered.additions.size());
+        for (Replica added : unfiltered.additions)
+        {
+            // Include any new FULL replicas here. If there exists the removal of a corresponding Transient
+            // replica (i.e. a switch from T -> F), add that too. We want T -> F transitions to happen early
+            // in a multi-step operation, at the same time as new write replicas are added.
+            if (added.isFull())
+            {
+                additions.add(added);
+                for (Replica removed : unfiltered.removals)
```
ifesdjeen: Nit: should use same style for filtering / checking removals in both branches? 

I _think_ we can also break out from the loop here early if we did add a new entry to `removals`.

test/distributed/org/apache/cassandra/distributed/test/log/PlacementSimulator.java
```diff
@@ -985,20 +977,84 @@ public Diff<T> invert()
...
+        List<Replica> removals = new ArrayList<>(unfiltered.removals.size());
+        for (Replica removed : unfiltered.removals)
+        {
+            // Include any new FULL replicas here. If there exists the removal of a corresponding Transient
+            // replica (i.e. a switch from T -> F), add that too. We want T -> F transitions to happen early
+            // in a multi-step operation, at the same time as new write replicas are added.
+            if (removed.isFull())
+            {
+                removals.add(removed);
+                for (Replica added : unfiltered.additions)
```
ifesdjeen: Nit: I think same comment applies then to the above `additionsAndTransientToFull` code.

ifesdjeen (approved):

src/java/org/apache/cassandra/tcm/ownership/PlacementForRange.java
krummas: think its just that extra "being" which makes this confusing, fixed

src/java/org/apache/cassandra/tcm/ownership/PlacementForRange.java
krummas: I think I prefer the split version for the comments per-case

src/java/org/apache/cassandra/tcm/sequences/RemoveNodeStreams.java
krummas: yep, almost - this became
```
if (startWriteRemovals.get(newReplica.endpoint()).contains(newReplica.range(), false))
    logger.debug("Streaming transient -> full conversion to {} from {}", newReplica, oldReplicas.get(newReplica.range()));
movements.put(oldReplicas.get(newReplica.range()), newReplica);
```

src/java/org/apache/cassandra/tcm/transformations/PrepareMove.java
krummas: made `assertPreExistingWriteReplica` throw `RejectedTransformationException` instead

test/harry/main/org/apache/cassandra/harry/sut/TokenPlacementModel.java
krummas: think the naming is clear enough that this won't introduce any issues

test/distributed/org/apache/cassandra/distributed/test/log/PlacementSimulator.java
krummas: what did you have in mind here?

ifesdjeen: I thought maybe something like this:

```
            if (added.isFull())
            {
                additions.add(added);
                Optional<Replica> removed = unfiltered.removals.stream()
                                                               .filter(r -> r.isTransient() && r.node().equals(added.node()))
                                                               .findFirst();

                if (removed.isPresent())
                    removals.add(removed.get());
            }
            else
            {
                // Conversely, when a replica transitions from F -> T, it's enacted late in a multi-step operation.
                // So only include TRANSIENT additions if there is no removal of a corresponding FULL replica.
                boolean include = unfiltered.removals.stream()
                                                     .noneMatch(removed -> removed.node().equals(added.node())
                                                                           && removed.isFull());
                if (include)
                    additions.add(added);
            }
```

PR #3197, opened 2024-03-22
CASSANDRA-19482 Revisit metadata log schema to remove period field

beobal:
The distributed metadata log table can be simplified by leveraging the fact that replicas are all responsible for the entire token range. Given this assumption, we can then use ReversedLongLocalPartitioner introduced in [CASSANDRA-19391](https://issues.apache.org/jira/browse/CASSANDRA-19391) to make it much easier to append to/read from the tail of the log, effectively removing the need for the Period construct. This will also improve apply to the local metadata log used at startup.


src/java/org/apache/cassandra/locator/CMSPlacementStrategy.java
```diff
@@ -115,8 +117,11 @@ public Set<NodeId> reconfigure(Set<NodeId> currentCms, ClusterMetadata metadata)
                 }
             }
 
-            EndpointsForRange endpoints = NetworkTopologyStrategy.calculateNaturalReplicas(EntireRange.entireRange.left,
-                                                                                           EntireRange.entireRange,
+            // Although MetaStrategy has its own entireRange, it uses a custom partitioner which isn't compatible with
+            // NTS. For that reason, we select replicas here using tokens provided by the configured partitioner.
```
ifesdjeen: nit: i understand this is technically correct to say NTS here, but maybe it is worth saying "regular placements" or non-CMS placements?

src/java/org/apache/cassandra/locator/LocalStrategy.java
```diff
@@ -53,4 +55,17 @@ public ReplicationFactor getReplicationFactor()
...
         return RF;
     }
+
+    /**
+     * For lazy initialisation. In some circumstances, we may want to instantiate LocalStrategy without initialising
+     * DatabaseDescriptor; FQL replay is one such usage as we initialise the KeyspaceMetadata objects, which now eagerly
+     * creates the replication strategy.
+     */
+    static class EntireRange
+    {
```
ifesdjeen, line 65: Should we have these per partitioner maybe?

src/java/org/apache/cassandra/locator/ReplicaLayout.java
```diff
@@ -383,17 +383,32 @@ static ReplicaLayout.ForRangeRead forRangeReadLiveSorted(ClusterMetadata metadat
 
     static EndpointsForRange forNonLocalStategyRangeRead(ClusterMetadata metadata, KeyspaceMetadata keyspace, AbstractBounds<PartitionPosition> range)
     {
-        return metadata.placements.get(keyspace.params.replication).reads.forRange(range.right.getToken()).get();
+        // MetaStrategy distributes the entire keyspace to all replicas. In addition, its tables (currently only
+        // the dist log table) don't use the globally configured partitioner. For these reasons we don't lookup the
+        // replicas using the supplied token as this can actually be of the incorrect type (for example when
+        // performing Paxos repair).
+        Token searchToken = keyspace.params.replication.isMeta() ? MetaStrategy.entireRange.right : range.right.getToken();
```
ifesdjeen: When looking at things like that I am wondering if we should add a jira for actually making it possible to run with different partitioners. 

With paxos repair I can see the point: there the paxos table has a different token, but token reads/writes, or for range reads - do we really need to have this special-casing? Here, we should be using the bounds consistent with the keyspace. Or at least removing this special casing does not break `ConsistentBootstrapTest` or `ReconfigureCMSTest`.

src/java/org/apache/cassandra/repair/messages/PrepareMessage.java
```diff
@@ -106,10 +106,7 @@ public void serialize(PrepareMessage message, DataOutputPlus out, int version) t
             message.parentRepairSession.serialize(out);
             out.writeInt(message.ranges.size());
             for (Range<Token> r : message.ranges)
-            {
```
ifesdjeen, line 109: I am not opposed to this, but could you add a comment for posterity about why you have decided to remove `validate`?

src/java/org/apache/cassandra/tcm/ownership/DataPlacement.java
```diff
@@ -25,19 +25,28 @@
...
 import org.apache.cassandra.locator.Replica;
+import org.apache.cassandra.schema.ReplicationParams;
 import org.apache.cassandra.tcm.Epoch;
 import org.apache.cassandra.tcm.serialization.MetadataSerializer;
 import org.apache.cassandra.tcm.serialization.Version;
 
 public class DataPlacement
 {
-    public static final Serializer serializer = new Serializer();
+    private static final Serializer serializer = new Serializer(DatabaseDescriptor.getPartitioner());
```
ifesdjeen: Should we attempt to standartize and use `global` here? Also, should we rename this private field to something like global serialiser?

src/java/org/apache/cassandra/tcm/ownership/MovementMap.java
```diff
@@ -144,7 +146,8 @@ public MovementMap deserialize(DataInputPlus in, int version) throws IOException
             for (int i = 0; i < size; i++)
             {
                 ReplicationParams params = ReplicationParams.messageSerializer.deserialize(in, version);
-                EndpointsByReplica endpointsByReplica = EndpointsByReplica.serializer.deserialize(in, version);
+                IPartitioner partitioner = params.isMeta() ? MetaStrategy.partitioner : IPartitioner.global();
```
ifesdjeen, line 149: I have only briefly checked but it looks like we are not using movement maps for meta keyspace, only diffs. Do we need this?

src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
```diff
@@ -754,14 +752,9 @@ public MetadataSnapshots snapshotManager()
...
-    public ClusterMetadata sealPeriod()
+    public ClusterMetadata triggerSnapshot()
     {
-        return ClusterMetadataService.instance.commit(SealPeriod.instance,
-                                                      (ClusterMetadata metadata) -> metadata,
-                                                      (code, reason) -> {
-                                                          // If the transformation got rejected, someone else has beat us to seal this period
-                                                          return ClusterMetadata.current();
-                                                      });
+        return ClusterMetadataService.instance.commit(TriggerSnapshot.instance);
```
ifesdjeen, line 757: Ah, interesting, looks like there's no way for the snapshot to get rejected now. I can't see how this can cause anything bad though.

ifesdjeen (approved):

src/java/org/apache/cassandra/locator/CMSPlacementStrategy.java
beobal: fixed in fb1f796de1c6acb23688a421e71e1b4ea76e5359

src/java/org/apache/cassandra/locator/LocalStrategy.java
beobal: That's a great idea and it is doable, but it will involve jumping through some hoops around static initialisation. I'd rather not start pulling on that thread here, so if it's ok we can file a follow up JIRA for that.

src/java/org/apache/cassandra/locator/ReplicaLayout.java
beobal: Like you say, paxos repair (and to an extent paxos in general) is complicated by the fact that the keys and ranges in the system paxos tables actually represent keys in the tables which are the subjects of the paxos transactions, but now these may be using different partitioners. 
This special casing *did* turn out to not be entirely necessary though (it was a bit of a remnant from earlier on in development). I've reworked it in 43ba1cd83833bde270ea949016556baa635d6bf9

src/java/org/apache/cassandra/tcm/ownership/MovementMap.java
beobal: We do need it, as `MovementMap` is used for the meta keyspace when reconfiguring the CMS.

src/java/org/apache/cassandra/tcm/ownership/DataPlacement.java
beobal: Done in f16defd9791106a87098e6e1d58ad32c06dc59e9

PR #3203, opened 2024-03-25
CASSANDRA-19486 enrich hints vtable with hints sizes

smiklosovic:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/hints/HintsStore.java
+++ b/src/java/org/apache/cassandra/hints/HintsStore.java
@@ -18,23 +18,29 @@
 package org.apache.cassandra.hints;
 
 import java.io.IOException;
-import java.util.*;
+import java.util.Deque;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Queue;
+import java.util.Set;
+import java.util.UUID;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.function.Predicate;
-
 import javax.annotation.Nullable;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
-import org.apache.cassandra.io.util.File;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.cassandra.gms.FailureDetector;
 import org.apache.cassandra.io.FSWriteError;
+import org.apache.cassandra.io.util.File;
 import org.apache.cassandra.locator.InetAddressAndPort;
 import org.apache.cassandra.schema.Schema;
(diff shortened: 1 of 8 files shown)


src/java/org/apache/cassandra/hints/HintsServiceMBean.java
```diff
@@ -50,4 +50,17 @@ public interface HintsServiceMBean
      * @return a list of endpoints with relevant hint information - total number of files, newest and oldest timestamps.
      */
     List<Map<String, String>> getPendingHints();
+
+    /**g
+     * Gets total size of all hints for a respective node
+     * @param address address of a node to get size of hints for
+     * @return size of hints for a given address
+     */
+    long getHintsSizeForEndpoint(String address);
```
smiklosovic: When one checks this more closely, there is only resolution of nodes on address level, if there are two nodes behind same address (but on different ports), then this does not work. HintsCatalog and HintsStore are (host)-UUID based but these methods are not.

This does not work already for `deleteAllHintsForEndpoint`. 

Ideally, there should be these methods added too:

`deleteAllHintsForNodeId`
`getHintsSizeForNodeId`

We might leave getHintsSize(String address)

src/java/org/apache/cassandra/hints/HintsStore.java
```diff
@@ -111,17 +117,37 @@ PendingHintsInfo getPendingHintsInfo()
...
+        Iterator<HintsDescriptor> corruptedDescriptors = corruptedFiles.iterator();
+        while (corruptedDescriptors.hasNext())
+        {
+            HintsDescriptor corruptedDescriptor = corruptedDescriptors.next();
+            try
+            {
+                corruptedFilesSize += corruptedDescriptor.hintsFileSize(hintsDirectory);
+            } catch (Exception ex)
+            {
+                // ignore
```
smiklosovic: the logic behind this is that if a descriptor was added among corrupted, it was done so in a catch, so it is probable that if we ask its size it would throw again, just to be super sure we do not ruin whole query, lets just wrap it in a try-catch

bereng: Why not add that explanation to the java comment?

bereng:
In HintsReaderTest and HintsDescriptorTest there is code to generate corrupt hints files and then you could test that. Iiuc your patch doesn't test that it catches corrupt files?

smiklosovic:
@bereng ah thanks, making corrupted hints files is definitely useful

smiklosovic:
@bereng I choose different approach, it is not so easy to mark some hints descriptor corrupted because it happens asynchronously in `HintsDispatchExecutor.DispatchHintsTask`. I just marked one of descriptors as corrupted calling respective `markCorrupted` method in `HintsStore`.

bereng:
Icwym. I think it's nice we test the corruption though I didn't expect it to be so involved. I though throwing a new test method would have been enough :shrug:

There are a ton of `Hints*Test.java` tests. Would you like to rename your `PendingHintsTableTest.java` to `HintsPendingTableTest.java`?

Otherwise code lgtm. CI is missing?

PR #3210, opened 2024-03-28
Implement types compatibility tests

jacek-lewandowski:


build.xml
```diff
@@ -710,9 +710,9 @@
           <dependency groupId="com.google.j2objc" artifactId="j2objc-annotations" version="1.3"/>
           <!-- adding this dependency is necessary for assertj. When updating assertj, need to also update the version of
              this that the new assertj's `assertj-parent-pom` depends on. -->
-          <dependency groupId="org.junit" artifactId="junit-bom" version="5.6.0" type="pom" scope="test"/>
```
smiklosovic, line 713: I would be more comfortable with bumping this in a separate ticket rather than burying it in a patch like this, for better visibility. Is it also absolutely necessary to bump that or is this mostly an opportunistic "why not" bump?

jacek-lewandowski: `SoftAssertions` is the reason. If it was a separate ticket it would be "why not bump" 😄

jacek-lewandowski: sorry - you asking specifically about "bom" or assertj?

smiklosovic: @jacek-lewandowski basically both ... so what is the reason for bom?

smiklosovic: I guess you just followed this, correct?

jacek-lewandowski: yes

smiklosovic: ok, so I think it is all justified.

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1324/workflows/4ab8e461-b506-4868-aeb4-fcb4ebee89e4
Known failures:
`BatchTest`: https://issues.apache.org/jira/browse/CASSANDRA-18974
`MemtableSizeTest`: https://issues.apache.org/jira/browse/CASSANDRA-17298

+ repeated tests of failed case (repeated tests looks failed because they produced too long output, but the results are ok)
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1327/workflows/0f3d9e83-0ccb-4ebd-b59a-cca82db8c162

j11 build: https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1327/workflows/9e93735c-f85c-4030-9aca-e37294841ef5

PR #3228, opened 2024-04-03
CASSANDRA-19515: unhandled exception handling in jvm dtest may see errors from other clusters

dcapwell:


test/distributed/org/apache/cassandra/distributed/impl/AbstractCluster.java
```diff
@@ -1066,11 +1092,23 @@ private void uncaughtExceptions(Thread thread, Throwable error)
...
+        if (!ownsThread(thread))
+        {
+            // another cluster hit us... which one?
+            for (AbstractCluster<?> cluster : ALL_CLUSTERS)
+            {
+                if (cluster == this || !cluster.ownsThread(thread)) continue;
+                cluster.uncaughtExceptions(thread, error);
+                break;
+            }
+            return;
```
bdeggleston: we should handle the case where no clusters think they own the exception and log it so that we don't swallow exceptions

bdeggleston (changes requested):

bdeggleston (approved):
+1

PR #3234, opened 2024-04-08
CASSANDRA-19128

ifesdjeen:


src/java/org/apache/cassandra/tcm/transformations/PrepareMove.java
```diff
@@ -67,6 +67,11 @@ public PrepareMove(NodeId nodeId,
         this.streamData = streamData;
     }
 
+    public NodeId nodeId()
```
krummas, line 71: maybe mark VisibleForTesting

src/java/org/apache/cassandra/tcm/MetadataKeys.java
```diff
@@ -52,4 +60,44 @@ public static MetadataKey make(String...parts)
...
     }
 
+    public static ImmutableSet<MetadataKey> diffKeys(ClusterMetadata before, ClusterMetadata after)
+    {
+        ImmutableSet.Builder<MetadataKey> builder = new ImmutableSet.Builder<>();
+        diffKeys(before, after, builder::add);
+        return builder.build();
+    }
+
+    private static void diffKeys(ClusterMetadata before, ClusterMetadata after, Consumer<MetadataKey> consumer)
```
krummas: this `Consumer` is always `builder::add`, could use that directly above instead (and move the logic in this helper method there)

krummas (commented):
2 nits inline

src/java/org/apache/cassandra/tcm/MetadataKeys.java
ifesdjeen: I could yeah; i just had an follow-up patch that would also add checks for cluster metadata equality / epoch bump, but I guess i can separate the two without pre-committing any code

PR #3240, opened 2024-04-09
CASSANDRA-19532: Allow operators to disable the execution of triggers

aratno:
This PR adds a new TriggersPolicy configuration to allow operators to disable the execution of triggers. It does not prevent the creation of triggers (with CQL's CREATE TRIGGER) because Config can change between initial creation and metadata replay, and we don't want to break replay if triggers are disabled but still present in replay. So triggers can still be created with `TriggersPolicy.{disabled,forbidden}`, but care has been taken to prevent the trigger creation from executing class initialization (and static blocks).

This has already been reviewed by @beobal.

### DIFF
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -1311,5 +1311,5 @@ public class Config
         forbidden
     }
 
-    public TriggersPolicy triggers_enabled = TriggersPolicy.enabled;
+    public TriggersPolicy triggers_policy = TriggersPolicy.enabled;
 }
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -1311,5 +1311,5 @@ public class Config
         forbidden
     }
 
-    public volatile TriggersPolicy triggers_policy = TriggersPolicy.enabled;
+    public TriggersPolicy triggers_policy = TriggersPolicy.enabled;
 }
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -1300,4 +1300,16 @@ public class Config
     public volatile DurationSpec.LongMillisecondsBound progress_barrier_backoff = new DurationSpec.LongMillisecondsBound("1000ms");
     public volatile DurationSpec.LongSecondsBound discovery_timeout = new DurationSpec.LongSecondsBound("30s");
     public boolean unsafe_tcm_mode = false;
+
+    public enum TriggersPolicy
+    {
+        // Execute triggers
+        enabled,
+        // Don't execute triggers when executing queries
+        disabled,
+        // Throw an exception when attempting to execute a trigger
+        forbidden
+    }
+
+    public volatile TriggersPolicy triggers_policy = TriggersPolicy.enabled;
 }
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -1311,5 +1311,5 @@ public class Config
         forbidden
     }
 
-    public TriggersPolicy triggers_policy = TriggersPolicy.enabled;
+    public TriggersPolicy triggers_enabled = TriggersPolicy.enabled;
 }
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -5153,4 +5153,15 @@ public class DatabaseDescriptor
     {
         return conf.sai_sstable_indexes_per_query_fail_threshold;
     }
+
+    public static void setTriggersPolicy(Config.TriggersPolicy policy)
+    {
+        logger.info("triggers_policy set to {}", policy);
+        conf.triggers_policy = policy;
+    }
+
+    public static Config.TriggersPolicy getTriggersPolicy()
+    {
+        return conf.triggers_policy;
+    }
 }
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -5154,6 +5154,7 @@ public class DatabaseDescriptor
         return conf.sai_sstable_indexes_per_query_fail_threshold;
     }
 
+    @VisibleForTesting
     public static void setTriggersPolicy(Config.TriggersPolicy policy)
     {
         logger.info("triggers_policy set to {}", policy);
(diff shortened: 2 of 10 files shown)


src/java/org/apache/cassandra/triggers/TriggerExecutor.java
```diff
@@ -252,8 +272,23 @@ private List<Mutation> executeInternal(PartitionUpdate update)
...
+        // CreateTriggerStatement#apply.
+        // Check that triggerClass is available on the classpath, but do not initialize the class since that would
+        // execute static blocks.
+        customClassLoader.loadClass(triggerClass).getConstructor();
+    }
+
     public synchronized ITrigger loadTriggerInstance(String triggerClass) throws Exception
     {
+        Config.TriggersPolicy policy = DatabaseDescriptor.getTriggersPolicy();
+        if (policy == Config.TriggersPolicy.disabled || policy == Config.TriggersPolicy.forbidden)
```
smiklosovic, line 287: @aratno can't be this just simplitied to `if (policy != enabled)` ?

test/unit/org/apache/cassandra/triggers/TriggersTest.java
```diff
@@ -80,6 +97,57 @@ public void setup() throws Exception
...
+        DatabaseDescriptor.setTriggersPolicy(TriggersPolicy.disabled);
+        QueryProcessor.process(String.format("INSERT INTO %s.%s (k, v1) VALUES (0, 0)", ksName, cfName), ConsistencyLevel.ONE);
+
+        UntypedResultSet rs = QueryProcessor.process(String.format("SELECT * FROM %s.%s WHERE k=%s", ksName, cfName, 0), ConsistencyLevel.ONE);
+        assertRowValue(rs.one(), 0, "v1", 0); // from original update
+        assertEquals(-1, rs.one().getInt("v2", -1)); // from trigger
+        QueryProcessor.process(String.format("DELETE FROM %s.%s WHERE k = 0", ksName, cfName), ConsistencyLevel.ONE);
+
+        DatabaseDescriptor.setTriggersPolicy(TriggersPolicy.forbidden);
+        try
```
smiklosovic: this whole try-catch might be replaced with more idiomatic `Assertions.assertThatThrownBy`

src/java/org/apache/cassandra/triggers/TriggerExecutor.java
```diff
@@ -220,6 +230,16 @@ private List<Mutation> executeInternal(PartitionUpdate update)
         Triggers triggers = update.metadata().triggers;
         if (triggers.isEmpty())
             return null;
+        Config.TriggersPolicy policy = DatabaseDescriptor.getTriggersPolicy();
```
smiklosovic, line 233: @aratno can't be this whole logic extracted to some method returning true / false and a boolean field would be populated in this class so we do not need to do this all over again every single time execute() method is called? That is given if we give up on a guardrail.

at least DatabaseDescriptor.getTriggersPolicy() might be "cached" so we do not need to call it all over again.

I would also move it to `execute` method, outside of the for loop where `executeInternal` is executed, because we would unnecessarily go through this logic all over again calling warn on logger (even it is non-spamming one).

smiklosovic: I just feel like we are burying this logic in some executeInternal method instead of failing as the very first thing we have a chance to do that.

src/java/org/apache/cassandra/triggers/TriggerExecutor.java
aratno: I could, it's a personal style preference to have enum checks be exhaustive / explicit. I find that it makes usage of each enum value easier to understand. I'd do `!=` if there were so many fields it felt wrong to list.

test/unit/org/apache/cassandra/triggers/TriggersTest.java
aratno: That is better, thanks!

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1300,4 +1300,16 @@ public static void log(Config config)
...
+    {
+        // Execute triggers
+        enabled,
+        // Don't execute triggers when executing queries
+        disabled,
+        // Throw an exception when attempting to execute a trigger
+        forbidden
+    }
+
+    public volatile TriggersPolicy triggers_policy = TriggersPolicy.enabled;
```
beobal: Probably doesn't need to be volatile

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -5153,4 +5153,15 @@ public static int getSaiSSTableIndexesPerQueryFailThreshold()
     {
         return conf.sai_sstable_indexes_per_query_fail_threshold;
     }
+
+    public static void setTriggersPolicy(Config.TriggersPolicy policy)
```
beobal, line 5158: can we annotate this `@VisibleForTesting` please?

src/java/org/apache/cassandra/triggers/TriggerExecutor.java
beobal: > at least DatabaseDescriptor.getTriggersPolicy() might be "cached" so we do not need to call it all over again.

As the policy is effectively immutable, this is almost equivalent to a constant lookup. Caching it isn't going to give us much in performance terms, and it won't do anything for code clarity. If it were memoized on `TriggersExecutor`, tests would need a way to set/override it, so for the sake of clarity I'd suggest leaving it as is. 

> I just feel like we are burying this logic in some executeInternal method instead of failing as the very first thing we have a chance to do that.

This is the first point at which we inspect the update to check whether the update actually has any triggers attached, so it seems a reasonable place to do this validation. It's intrinsic to the execution of any triggers, so I'd say the method that encapsulates that (i.e. `executeInternal`) is the ideal place for it.

> I would also move it to execute method, outside of the for loop where executeInternal is executed, because we would unnecessarily go through this logic all over again calling warn on logger (even it is non-spamming one).

Triggers are associated with a specific `PartitionUpdate` so in the version of `execute` which takes a collection, this has to be checked for each mutation.

beobal (approved):
Couple of minor nits, otherwise LGTM

src/java/org/apache/cassandra/triggers/TriggerExecutor.java
smiklosovic: yeah it would be a little bit hairy the other way ... we also do not have access to `triggers` from `update.metadata().triggers` unless we loop over `PartitionUpdate`s for the sake of using them in logger or exception upon message construction. 

Just to be explicit, so this would throw an exception on the first `Mutation` and on the first `PartitionUpdate` of that mutation.

beobal: It would throw on the first `PartitionUpdate` which has a trigger. It may iterate over other updates which don't before it gets to one, but that's fine too.

PR #3241, opened 2024-04-09
CASSANDRA-19493 Optionally fail writes when SAI refuses to index a term value exceeding a configured maximum size

maedhroz:
patch by Caleb Rackliffe; reviewed by ? for CASSANDRA-19493

### DIFF
--- a/src/java/org/apache/cassandra/config/Config.java
+++ b/src/java/org/apache/cassandra/config/Config.java
@@ -903,10 +903,17 @@ public class Config
     public volatile boolean zero_ttl_on_twcs_warned = true;
     public volatile boolean zero_ttl_on_twcs_enabled = true;
     public volatile boolean non_partition_restricted_index_query_enabled = true;
-    public volatile int sai_sstable_indexes_per_query_warn_threshold = 32;
-    public volatile int sai_sstable_indexes_per_query_fail_threshold = -1;
     public volatile boolean intersect_filtering_query_warned = true;
     public volatile boolean intersect_filtering_query_enabled = true;
+
+    public volatile int sai_sstable_indexes_per_query_warn_threshold = 32;
+    public volatile int sai_sstable_indexes_per_query_fail_threshold = -1;
+    public volatile DataStorageSpec.LongBytesBound sai_string_term_size_warn_threshold = new DataStorageSpec.LongBytesBound("1KiB");
+    public volatile DataStorageSpec.LongBytesBound sai_string_term_size_fail_threshold = new DataStorageSpec.LongBytesBound("8KiB");
+    public volatile DataStorageSpec.LongBytesBound sai_frozen_term_size_warn_threshold = new DataStorageSpec.LongBytesBound("1KiB");
+    public volatile DataStorageSpec.LongBytesBound sai_frozen_term_size_fail_threshold = new DataStorageSpec.LongBytesBound("8KiB");
+    public volatile DataStorageSpec.LongBytesBound sai_vector_term_size_warn_threshold = new DataStorageSpec.LongBytesBound("16KiB");
+    public volatile DataStorageSpec.LongBytesBound sai_vector_term_size_fail_threshold = new DataStorageSpec.LongBytesBound("32KiB");
 
     public volatile DurationSpec.LongNanosecondsBound streaming_state_expires = new DurationSpec.LongNanosecondsBound("3d");
     public volatile DataStorageSpec.LongBytesBound streaming_state_size = new DataStorageSpec.LongBytesBound("40MiB");
--- a/src/java/org/apache/cassandra/config/GuardrailsOptions.java
+++ b/src/java/org/apache/cassandra/config/GuardrailsOptions.java
@@ -961,6 +961,90 @@ public class GuardrailsOptions implements GuardrailsConfig
     }
 
     @Override
+    @Nullable
+    public DataStorageSpec.LongBytesBound getSaiStringTermSizeWarnThreshold()
+    {
+        return config.sai_string_term_size_warn_threshold;
+    }
+
+    @Override
+    @Nullable
+    public DataStorageSpec.LongBytesBound getSaiStringTermSizeFailThreshold()
+    {
+        return config.sai_string_term_size_fail_threshold;
+    }
+
+    @Override
+    public void setSaiStringTermSizeThreshold(@Nullable DataStorageSpec.LongBytesBound warn, @Nullable DataStorageSpec.LongBytesBound fail)
+    {
+        validateSizeThreshold(warn, fail, false, "sai_string_term_size");
+        updatePropertyWithLogging("sai_string_term_size_warn_threshold",
+                                  warn,
+                                  () -> config.sai_string_term_size_warn_threshold,
+                                  x -> config.sai_string_term_size_warn_threshold = x);
+        updatePropertyWithLogging("sai_string_term_size_fail_threshold",
+                                  fail,
+                                  () -> config.sai_string_term_size_fail_threshold,
+                                  x -> config.sai_string_term_size_fail_threshold = x);
+    }
+
+    @Override
+    @Nullable
+    public DataStorageSpec.LongBytesBound getSaiFrozenTermSizeWarnThreshold()
+    {
+        return config.sai_frozen_term_size_warn_threshold;
+    }
+
+    @Override
+    @Nullable
+    public DataStorageSpec.LongBytesBound getSaiFrozenTermSizeFailThreshold()
+    {
+        return config.sai_frozen_term_size_fail_threshold;
+    }
+
+    @Override
+    public void setSaiFrozenTermSizeThreshold(@Nullable DataStorageSpec.LongBytesBound warn, @Nullable DataStorageSpec.LongBytesBound fail)
+    {
+        validateSizeThreshold(warn, fail, false, "sai_frozen_term_size");
+        updatePropertyWithLogging("sai_frozen_term_size_warn_threshold",
+                                  warn,
+                                  () -> config.sai_frozen_term_size_warn_threshold,
+                                  x -> config.sai_frozen_term_size_warn_threshold = x);
+        updatePropertyWithLogging("sai_frozen_term_size_fail_threshold",
+                                  fail,
+                                  () -> config.sai_frozen_term_size_fail_threshold,
+                                  x -> config.sai_frozen_term_size_fail_threshold = x);
+    }
+
+    @Override
+    @Nullable
+    public DataStorageSpec.LongBytesBound getSaiVectorTermSizeWarnThreshold()
+    {
+        return config.sai_vector_term_size_warn_threshold;
+    }
+
+    @Override
+    @Nullable
+    public DataStorageSpec.LongBytesBound getSaiVectorTermSizeFailThreshold()
+    {
+        return config.sai_vector_term_size_fail_threshold;
+    }
+
+    @Override
+    public void setSaiVectorTermSizeThreshold(@Nullable DataStorageSpec.LongBytesBound warn, @Nullable DataStorageSpec.LongBytesBound fail)
+    {
+        validateSizeThreshold(warn, fail, false, "sai_vector_term_size");
+        updatePropertyWithLogging("sai_vector_term_size_warn_threshold",
+                                  warn,
+                                  () -> config.sai_vector_term_size_warn_threshold,
+                                  x -> config.sai_vector_term_size_warn_threshold = x);
+        updatePropertyWithLogging("sai_vector_term_size_fail_threshold",
+                                  fail,
+                                  () -> config.sai_vector_term_size_fail_threshold,
+                                  x -> config.sai_vector_term_size_fail_threshold = x);
+    }
+
+    @Override
     public boolean getNonPartitionRestrictedQueryEnabled()
     {
         return config.non_partition_restricted_index_query_enabled;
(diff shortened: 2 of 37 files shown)


test/anttasks/org/apache/cassandra/anttasks/TestNameCheckTask.java
```diff
@@ -158,14 +161,21 @@ private Class<?> normalize(Class<?> klass)
...
+        {
+            Class<?> type = typeStack.pop();
+
+            if (!Modifier.isAbstract(type.getModifiers()))
+                concreteTypes.add(type);
+
+            reflections.getSubTypesOf(type).forEach(typeStack::push);
+        }
+
+        return concreteTypes.stream();
```
maedhroz, line 178: @jacek-lewandowski @smiklosovic This was causing issues for `ValueThresholdTester`, which is a subclass of `ThresholdTester` but is also `abstract`. Basically we can handle more than one level of inheritance now. WDYT?

maedhroz: CC @bereng

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -1080,8 +1164,8 @@ private static void validateWarnLowerThanFail(long warn, long fail, String name)
             return;
 
         if (fail < warn)
-            throw new IllegalArgumentException(format("The warn threshold %d for %s_warn_threshold should be lower " +
-                                                      "than the fail threshold %d", warn, name, fail));
+            throw new IllegalArgumentException(format("The warn threshold %d for %s_warn_threshold should not be " +
```
bereng: That should be ok. I doubt anybody is parsing these messages. Just mentioning it to be on the safe side.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -359,7 +359,7 @@ public final class Guardrails implements GuardrailsMBean
                      state -> sizeToBytes(CONFIG_PROVIDER.getOrCreate(state).getColumnValueSizeWarnThreshold()),
                      state -> sizeToBytes(CONFIG_PROVIDER.getOrCreate(state).getColumnValueSizeFailThreshold()),
                      (isWarning, what, value, threshold) ->
-                     format("Value of column %s has size %s, this exceeds the %s threshold of %s.",
+                     format("Value of column '%s' has size %s, this exceeds the %s threshold of %s.",
```
bereng, line 362: Again change should be ok.

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
```diff
@@ -497,11 +497,15 @@ public boolean isQueryable(Status status)
...
 
     @Override
-    public void validate(PartitionUpdate update) throws InvalidRequestException
+    public void validate(PartitionUpdate update, ClientState state) throws InvalidRequestException
     {
         DecoratedKey key = update.partitionKey();
-        for (Row row : update)
-            validateMaxTermSizeForRow(key, row, true);
+
+        if (indexTermType.columnMetadata().isStatic())
```
bereng, line 504: Do iiuc static rows were being missed before?

test/anttasks/org/apache/cassandra/anttasks/TestNameCheckTask.java
bereng: Sgtm, you're just walking the whole tree.

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -2549,6 +2549,14 @@ protected Vector<Float> vector(float[] v)
         return new Vector<>(v2);
     }
 
+    protected List<Float> vectorAsList(float[] v)
```
bereng: Guava `Floats.asList()`?

test/unit/org/apache/cassandra/db/guardrails/GuardrailSaiFrozenTermSizeTest.java
```diff
@@ -0,0 +1,139 @@
...
+        ByteBuffer largeTuple = ByteBuffer.allocate(warnThreshold() + 1);
+        ByteBuffer smallTuple = ByteBuffer.allocate(1);
+
+        createTable(KEYSPACE, "CREATE TABLE %s (k int PRIMARY KEY, t tuple<text>)");
+        execute("INSERT INTO %s (k, t) VALUES (0, (?))", largeTuple);
+        execute("INSERT INTO %s (k, t) VALUES (1, (?))", smallTuple);
+        createIndex("CREATE INDEX ON %s(t) USING 'sai'");
+
+        // verify that the large tuple is written on initial index build
+        assertEquals(((ResultMessage.Rows) execute("SELECT * FROM %s WHERE t = (?)", largeTuple)).result.size(), 1);
```
bereng, line 120: No usage of `assertThresholdWarns`?

src/java/org/apache/cassandra/cql3/statements/BatchUpdatesCollector.java
```diff
@@ -124,15 +125,15 @@ private IMutationBuilder makeMutationBuilder(TableMetadata metadata, DecoratedKe
      * Returns a collection containing all the mutations.
      * @return a collection containing all the mutations.
      */
-    public List<IMutation> toMutations()
+    public List<IMutation> toMutations(ClientState state)
```
smiklosovic, line 140: nit @maedhroz what about updating the javadoc too?

src/java/org/apache/cassandra/db/IMutation.java
```diff
@@ -24,24 +24,25 @@
...
 import org.apache.cassandra.db.partitions.PartitionUpdate;
 import org.apache.cassandra.schema.TableId;
+import org.apache.cassandra.service.ClientState;
 
 public interface IMutation
 {
-    public long MAX_MUTATION_SIZE = DatabaseDescriptor.getMaxMutationSize();
+    long MAX_MUTATION_SIZE = DatabaseDescriptor.getMaxMutationSize();
 
-    public void apply();
```
smiklosovic, line 32: if you are removing this, we should be consistent and remove it for `void validateSize` and `static long dataSize` methods at the end of this file as well

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -903,11 +903,18 @@ public static void setClientMode(boolean clientMode)
...
     public volatile boolean zero_ttl_on_twcs_enabled = true;
     public volatile boolean non_partition_restricted_index_query_enabled = true;
-    public volatile int sai_sstable_indexes_per_query_warn_threshold = 32;
-    public volatile int sai_sstable_indexes_per_query_fail_threshold = -1;
     public volatile boolean intersect_filtering_query_warned = true;
     public volatile boolean intersect_filtering_query_enabled = true;
 
+    public volatile int sai_sstable_indexes_per_query_warn_threshold = 32;
+    public volatile int sai_sstable_indexes_per_query_fail_threshold = -1;
+    public volatile DataStorageSpec.LongBytesBound sai_string_term_size_warn_threshold = new DataStorageSpec.LongBytesBound("1KiB");
```
smiklosovic, line 911: @maedhroz you should also update `cassandra.yaml` with all these guardrails otherwise people wont have a clue they are available.

smiklosovic: by the way ... where are all these values for defaults coming from? I am not necessarily arguing the sizes as such, I am curious based on what we put these sizes there? Is that some kind of an empirical observation or something else? Shouldn't we just put everywhere "0" so people will make their minds about the sizes? We can default to zero (as not triggered) and then in cassandra.yaml we would provide the recommended sizes if people choose to turn it on.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
```diff
@@ -1164,8 +1164,8 @@ private static void validateWarnLowerThanFail(long warn, long fail, String name)
             return;
 
         if (fail < warn)
-            throw new IllegalArgumentException(format("The warn threshold %d for %s_warn_threshold should not be " +
-                                                      "higher than the fail threshold %d", warn, name, fail));
+            throw new IllegalArgumentException(format("The warn threshold %d for %s_warn_threshold should be lower " +
+                                                      "than the fail threshold %d", warn, name, fail));
```
maedhroz: @adelapena @bereng @smiklosovic I don't really want to fix it in this patch, but the check and the warning message don't strictly agree. Either it should be `if (fail <= warn)` or the error message needs to change.

src/java/org/apache/cassandra/config/Config.java
maedhroz: So you'll notice the frozen and vector fail thresholds are just the previous system property maxes: https://github.com/apache/cassandra/pull/3241/files#diff-32a8b4fa02cda1ef5a7216fe68d70d23a9784a4bb33b1230d2ec776140dae3d7L428

For strings, I bumped the fail threshold to a number @JeremiahDJordan had mentioned in Slack that Astra was using as a fail threshold (and left the previous max as the new warning threshold).

JeremiahDJordan: I would set frozen and string to be the same at 8k.
I think 32k still works for vectors.  4096 float32 is still the largest embedding I have seen truly used.

maedhroz: So how does this look?

string: 1KiB warn/8KiB fail
frozen: 1KiB warn/8KiB fail
vector: 16KiB warn/32KiB fail

test/anttasks/org/apache/cassandra/anttasks/TestNameCheckTask.java
maedhroz: Bonus: Getting to use a `Deque` for something :D

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
maedhroz: Yeah, basically just did it so I didn't have to special case things in the tests.

src/java/org/apache/cassandra/index/sai/StorageAttachedIndex.java
maedhroz: As far as I can tell, they absolutely were missed.

maedhroz: The static row isn't in the tree and so isn't returned from the iterator. Gotta love static rows, huh?

test/unit/org/apache/cassandra/cql3/CQLTester.java
maedhroz: Perfect. Was looking for something like that...

test/unit/org/apache/cassandra/db/guardrails/GuardrailSaiFrozenTermSizeTest.java
maedhroz: In this case, there is no client warning, because the INSERT happens before the index exists. There would be a log message, however.

src/java/org/apache/cassandra/db/IMutation.java
maedhroz: Done

maedhroz:
Pushed a follow-up commit that I think should address the comments so far...

test/anttasks/org/apache/cassandra/anttasks/TestNameCheckTask.java
bereng: Yeah I noticed lol!

test/unit/org/apache/cassandra/db/guardrails/GuardrailSaiFrozenTermSizeTest.java
bereng: Right apologies.

src/java/org/apache/cassandra/config/Config.java
bereng: I don't have an opinion here. I don't have production numbers to refer to neither I know SAI in that regard, so whatever you guys think it's reasonable.

src/java/org/apache/cassandra/config/GuardrailsOptions.java
bereng: If you find a bug you fix it... I don't see a problem with that.

src/java/org/apache/cassandra/config/Config.java
smiklosovic: @maedhroz  I dont know if we should go with the defaults in the first place. That is my biggest concern. I would just keep it disabled and people might opt-in. You can bring this to ML and ask. Maybe I am a little bit paranoid here but we should triple check that these figures make sense to as broad audience as possible.

maedhroz: I think @JeremiahDJordan's input counts for something here, given he's the only one of us (maybe @bereng too?) running SAI in production (in Astra). Also, if you look at this patch from the perspective of what was here before, all I've done is take the same (or even more lenient) max term sizes, and make sure writes fail visibly for the client. This seems strictly better than silently compromising the index with not so much as a warning. (I didn't even change the behavior on initial index build or compaction, other than making the warning threshold a bit more lenient.)

The limits on term size before my patch here were derived at some point during development at DS to protect the index, and I'm -100 on them being opt-in, especially because they are brand new in 5.0 (as is SAI itself). I'm more or less willing to die on this hill.

CC @mike-tr-adamson

smiklosovic: Alright. Let's keep it like it is now.

maedhroz:
Committed as https://github.com/apache/cassandra/commit/9bfaee91c4fd7a269e3ff924e8a504bad5d6514a

PR #3249, opened 2024-04-14
CASSANDRA-19483 return null on invalid legacy protocol magic

ilyazr:
Changed handling of invalid legacy protocol magic. We used to throw an `InvalidLegacyProtocolMagic`. With this PR `null` will be returned instead and that case will be logged at `WARN` with `NoSpamLogger`

```
return null on invalid legacy protocol magic

patch by ilyazr; reviewed by <Reviewers> for CASSANDRA-19483 

```

The [Cassandra Jira](https://issues.apache.org/jira/browse/CASSANDRA-19483)

### DIFF
--- a/src/java/org/apache/cassandra/net/InboundMessageHandler.java
+++ b/src/java/org/apache/cassandra/net/InboundMessageHandler.java
@@ -302,10 +302,7 @@ public class InboundMessageHandler extends AbstractMessageHandler
 
         JVMStabilityInspector.inspectThrowable(cause);
 
-        if (cause instanceof Message.InvalidLegacyProtocolMagic)
-            logger.error("{} invalid, unrecoverable CRC mismatch detected while reading messages - closing the connection", id());
-        else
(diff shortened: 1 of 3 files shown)


bschoening:
@ilyazr Can this be removed as well?

public static final class InvalidLegacyProtocolMagic extends IOException
....

ilyazr:
@bschoening Done. The class has been removed

src/java/org/apache/cassandra/net/InboundMessageHandler.java
```diff
@@ -302,10 +302,7 @@ protected void fatalExceptionCaught(Throwable cause)
 
         JVMStabilityInspector.inspectThrowable(cause);
 
-        if (cause instanceof Message.InvalidLegacyProtocolMagic)
-            logger.error("{} invalid, unrecoverable CRC mismatch detected while reading messages - closing the connection", id());
```
smiklosovic, line 306: @ilyazr what is wrong with keeping everyhing as is and just use "noSpam1m" here?

PR #3263, opened 2024-04-17
Fix peers v2 system table behaviour when 2 nodes swap their IP Addresses

ifesdjeen:
Throw if node id has been changed and does not match directory. If, however the _ip_ address has changed, issue Startup and correct the IP address. Disallow picking over identity of other nodes via hijacking their IPs or via overriding local node id with theirs.

### DIFF
--- a/src/java/org/apache/cassandra/db/virtual/PeersTable.java
+++ b/src/java/org/apache/cassandra/db/virtual/PeersTable.java
@@ -113,15 +113,6 @@ public class PeersTable extends AbstractVirtualTable
         return result;
     }
 
-    public static void initializeLegacyPeerTables(ClusterMetadata prev, ClusterMetadata next)
-    {
-        QueryProcessor.executeInternal(String.format("TRUNCATE %s.%s", SYSTEM_KEYSPACE_NAME, PEERS_V2));
-        QueryProcessor.executeInternal(String.format("TRUNCATE %s.%s", SYSTEM_KEYSPACE_NAME, LEGACY_PEERS));
-
-        for (NodeId nodeId : next.directory.peerIds())
-            updateLegacyPeerTable(nodeId, prev, next);
-    }
-
     private static String peers_v2_query = "INSERT INTO %s.%s ("
                                             + "peer, peer_port, "
                                             + "preferred_ip, preferred_port, "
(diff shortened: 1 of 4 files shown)


src/java/org/apache/cassandra/db/virtual/PeersTable.java
```diff
@@ -156,9 +156,7 @@ public static void updateLegacyPeerTable(NodeId nodeId, ClusterMetadata prev, Cl
         if (next.directory.peerState(nodeId) == null || next.directory.peerState(nodeId) == NodeState.LEFT)
         {
             NodeAddresses addresses = prev.directory.getNodeAddresses(nodeId);
-            logger.debug("Purging {} from system.peers_v2 table", addresses);
-            QueryProcessor.executeInternal(String.format(peers_delete_query, SYSTEM_KEYSPACE_NAME, PEERS_V2), addresses.broadcastAddress.getAddress(), addresses.broadcastAddress.getPort());
-            QueryProcessor.executeInternal(String.format(legacy_peers_delete_query, SYSTEM_KEYSPACE_NAME, LEGACY_PEERS), addresses.broadcastAddress.getAddress());
+            removeFromLegacyPeerTable(addresses.broadcastAddress);
```
beobal: can also replace the calls on lines 171-173 below

src/java/org/apache/cassandra/db/virtual/PeersTable.java
```diff
@@ -197,4 +195,11 @@ else if (NodeState.isPreJoin(next.directory.peerState(nodeId)))
                                            tokens);
         }
     }
+
+    public static void removeFromLegacyPeerTable(InetAddressAndPort addr)
```
beobal: nit: IMO this name is  slightly confusing because we execute not only `legacy_peers_delete_query` but also `peers_delete_query`. I know that *both* those tables are "legacy" in the sense that `ClusterMetadata.directory` is the primary datasource now, but I'd either rename the method to `removeFromSystemPeersTables` or rename the queries to `peers_delete_query / peers_v2_delete_query`

beobal (approved):
just nits

src/java/org/apache/cassandra/db/virtual/PeersTable.java
ifesdjeen: Good point. Renamed to `removeFromLegacyPeerTable`.

PR #3267, opened 2024-04-19
CASSANDRA-19567 Reduce heap usage by metrics

Mmuzaf:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -90,14 +90,14 @@ import org.apache.cassandra.db.compaction.CompactionStrategyManager;
 import org.apache.cassandra.db.compaction.OperationType;
 import org.apache.cassandra.db.filter.ClusteringIndexFilter;
 import org.apache.cassandra.db.filter.DataLimits;
-import org.apache.cassandra.db.memtable.Flushing;
-import org.apache.cassandra.db.memtable.Memtable;
-import org.apache.cassandra.db.memtable.ShardBoundaries;
 import org.apache.cassandra.db.lifecycle.LifecycleNewTracker;
 import org.apache.cassandra.db.lifecycle.LifecycleTransaction;
 import org.apache.cassandra.db.lifecycle.SSTableSet;
 import org.apache.cassandra.db.lifecycle.Tracker;
 import org.apache.cassandra.db.lifecycle.View;
+import org.apache.cassandra.db.memtable.Flushing;
+import org.apache.cassandra.db.memtable.Memtable;
+import org.apache.cassandra.db.memtable.ShardBoundaries;
 import org.apache.cassandra.db.partitions.CachedPartition;
 import org.apache.cassandra.db.partitions.PartitionUpdate;
 import org.apache.cassandra.db.repair.CassandraTableRepairManager;
@@ -320,6 +320,7 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean, Memtable.Owner
     private final Directories directories;
 
     public final TableMetrics metric;
+    private final Runnable memtableMetricsReleaser;
     public volatile long sampleReadLatencyMicros;
     public volatile long additionalWriteLatencyMicros;
 
@@ -506,14 +507,11 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean, Memtable.Owner
 
         logger.info("Initializing {}.{}", getKeyspaceName(), name);
 
-        // Create Memtable and its metrics object only on online
-        Memtable initialMemtable = null;
-        TableMetrics.ReleasableMetric memtableMetrics = null;
-        if (DatabaseDescriptor.isDaemonInitialized())
-        {
-            initialMemtable = createMemtable(new AtomicReference<>(CommitLog.instance.getCurrentPosition()));
-            memtableMetrics = memtableFactory.createMemtableMetrics(metadata);
-        }
+        Memtable initialMemtable = DatabaseDescriptor.isDaemonInitialized() ?
+                                   createMemtable(new AtomicReference<>(CommitLog.instance.getCurrentPosition())) :
+                                   null;
+        memtableMetricsReleaser = memtableFactory.createMemtableMetricsReleaser(metadata);
+
         data = new Tracker(this, initialMemtable, loadSSTables);
 
         // Note that this needs to happen before we load the first sstables, or the global sstable tracker will not
@@ -545,7 +543,9 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean, Memtable.Owner
             indexManager.addIndex(info, true);
         }
 
-        metric = new TableMetrics(this, memtableMetrics);
+        // See CASSANDRA-16228. We need to ensure that metrics are exposed after the CFS is initialized,
+        // so the order of the following line is important and should not be moved.
+        metric = new TableMetrics(this);
 
         if (data.loadsstables)
         {
@@ -749,6 +749,7 @@ public class ColumnFamilyStore implements ColumnFamilyStoreMBean, Memtable.Owner
         }
 
         // unregister metrics
+        memtableMetricsReleaser.run();
         metric.release();
     }
 
--- a/src/java/org/apache/cassandra/db/memtable/Memtable.java
+++ b/src/java/org/apache/cassandra/db/memtable/Memtable.java
@@ -19,7 +19,6 @@
 package org.apache.cassandra.db.memtable;
 
 import java.util.concurrent.atomic.AtomicReference;
-
 import javax.annotation.concurrent.NotThreadSafe;
 
 import org.apache.cassandra.db.ColumnFamilyStore;
@@ -33,7 +32,6 @@ import org.apache.cassandra.db.rows.EncodingStats;
 import org.apache.cassandra.db.rows.UnfilteredSource;
 import org.apache.cassandra.index.transactions.UpdateTransaction;
 import org.apache.cassandra.io.sstable.format.SSTableWriter;
-import org.apache.cassandra.metrics.TableMetrics;
 import org.apache.cassandra.schema.TableMetadata;
 import org.apache.cassandra.schema.TableMetadataRef;
 import org.apache.cassandra.utils.FBUtilities;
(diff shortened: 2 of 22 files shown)


src/java/org/apache/cassandra/db/memtable/Memtable.java
```diff
@@ -144,7 +143,7 @@ default boolean streamFromMemtable()
          * Memtable metrics lifecycle matches table lifecycle. It is the table that owns the metrics and
          * decides when to release them.
          */
-        default TableMetrics.ReleasableMetric createMemtableMetrics(TableMetadataRef metadataRef)
+        default TrieMemtableMetricsView createMemtableMetrics(TableMetadataRef metadataRef)
```
maedhroz: It seems like we'd want to keep something like the `ReleasableMetric` interface to keep details about particular implementations out of the `Memtable` interface. What am I missing?

maedhroz: Perhaps the idea is trying to avoid creating a new lambda in `TrieMemtable`, but we could just have `TrieMemtableMetricsView` implement `ReleasableMetric` right?

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -290,6 +291,7 @@ public enum FlushReason
     private volatile boolean valid = true;
 
     private volatile Memtable.Factory memtableFactory;
+    private final TrieMemtableMetricsView memtableMetrics;
```
maedhroz: see https://github.com/apache/cassandra/pull/3267/files#r1579775946

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -508,12 +510,15 @@ public ColumnFamilyStore(Keyspace keyspace,
...
         if (DatabaseDescriptor.isDaemonInitialized())
         {
             initialMemtable = createMemtable(new AtomicReference<>(CommitLog.instance.getCurrentPosition()));
-            memtableMetrics = memtableFactory.createMemtableMetrics(metadata);
+            this.memtableMetrics = memtableFactory.createMemtableMetrics(metadata);
         }
+        else
+            this.memtableMetrics = null;
+
+        metric = new TableMetrics(this);
```
maedhroz: This appears to reintroduce [CASSANDRA-16228](https://issues.apache.org/jira/browse/CASSANDRA-16228)

src/java/org/apache/cassandra/index/sai/metrics/AbstractMetrics.java
```diff
@@ -57,28 +55,28 @@ public abstract class AbstractMetrics
...
-                                                                                                 TYPE, name, metricScope, createMBeanName(name, scope));
-        tracked.add(metricName.getMetricName());
-        return metricName;
+    private CassandraMetricsRegistry.MetricName createMetricName(String name, String scope)
+    {
+        assert name.indexOf('.') == -1 : String.format("Metric name '%s' should not contain '.'", name);
+        return new CassandraMetricsRegistry.MetricName(DefaultNameFactory.GROUP_NAME,
+                                                       TYPE,
+                                                       name,
+                                                       MetricRegistry.name(keyspace, table, index, scope, name),
```
maedhroz, line 73: It feels like something around this change is causing problems w/ `index` handling that pop up in `GuardrailNonPartitionRestrictedQueryTest`, `IndexAvailabilityTest`, and a few other places.

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
```diff
@@ -444,30 +410,71 @@ void remove(ThreadPoolMetrics metrics)
...
+        String prefix = name(group, type);
+        if (fullName.startsWith(prefix))
+            return fullName.substring(prefix.length() + 1, fullName.indexOf('.', prefix.length() + 1));
+
+        return null;
+    }
+
     public void remove(MetricName name)
     {
         // Aliases are removed in onMetricRemoved by metrics listener.
```
maedhroz: nit: This comment is no longer true?

maedhroz: The way this works does make more sense now, though. The table metrics were the only place we cared about the aliases, and they handle removal specifically now.

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
```diff
@@ -444,30 +410,71 @@ void remove(ThreadPoolMetrics metrics)
...
-        remove(name.getMetricName());
+        boolean success = remove(name.getMetricName());
+        if (success)
```
maedhroz, line 483: nit: Should we ever call `remove()` here and have `!success`? It seems like that would be a programmer error...and something we should perhaps make more visible w/ logging, `assert`, etc.?

maedhroz (changes requested):
Looks like a few test failures to resolve, and some abstraction-related items, but overall I like this a lot. Using `removeMatching()` makes the removal plumbing much nicer.

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
Mmuzaf: Yes, I've added an extra comment for that. The initialization of metrics is critical, depending on which line they are created from.

src/java/org/apache/cassandra/index/sai/metrics/AbstractMetrics.java
Mmuzaf: Some of the metrics are created with `scope = null`, so the `resolveShortMetricName` method won't work in this case. This is an implementation bug. Fixed.

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
Mmuzaf: Comment removed.

src/java/org/apache/cassandra/db/memtable/Memtable.java
Mmuzaf: I've looked at it again and I think we can remove the `createMemtableMetrics` method, as it doesn't directly instantiate metrics for memtables and the metric handling is a bit confusing so far.

The metric class instance created by the `createMemtableMetrics` method differs from the metrics created for the memtables themselves via a constructor (see [new TrieMemtable(...)](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/db/memtable/TrieMemtable.java#L126)). So far they only match and released just because the metric names are exactly the same. All `TrieMemtableMetricsView` share the same `Metric` instances, but if we add a memtable-specific metric it will immediately lead to a memory leak as this metric won't be released properly.

I suggest explicitly releasing all metrics assigned to the memtables (e.g. by the `TrieMemtableMetricsView`), even if it ends up being released only once. I've pushed the changes into the PR. It seems to be much cleaner now. WDYT?

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
Mmuzaf: `!success` should never happen in general, but I've noticed that because of the hierarchical structure of some metrics and the lifecycle we have, we can try to release a metric more than once.

The `mBeanWrapper.unregisterMBean(name, MBeanWrapper.OnException.IGNORE);` indicates that some of the exceptions about missing requested metrics are simply ignored due to the fact that the requested metric is already released. 

So I'd keep it as is for safety, but we can change it and see how many tests will fail.

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -320,6 +320,7 @@ public enum FlushReason
     private final Directories directories;
 
     public final TableMetrics metric;
+    private final Runnable metricsReleaser;
```
maedhroz: nit: `memtableMetricsReleaser`?

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -545,7 +543,9 @@ public ColumnFamilyStore(Keyspace keyspace,
             indexManager.addIndex(info, true);
         }
 
-        metric = new TableMetrics(this, memtableMetrics);
+        // See CASSANDRA-16228. We need to ensure that metrics are exposed after the CFS is initialized,
+        // so the order of the following line is important and should not be moved.
+        metric = new TableMetrics(this);
```
maedhroz, line 548: Thanks for adding the comment ;)

src/java/org/apache/cassandra/index/sai/metrics/ColumnQueryMetrics.java
```diff
@@ -97,6 +104,14 @@ public BalancedTreeIndexMetrics(IndexIdentifier indexIdentifier)
             postingsListener = new PostingListEventsMetrics(postingDecodes);
         }
 
+        @Override
+        public void release()
+        {
+            super.release();
+            Metrics.remove(createMetricName("NumPostings", BALANCED_TREE_POSTINGS_TYPE));
+            Metrics.remove(createMetricName("PostingDecodes", BALANCED_TREE_POSTINGS_TYPE));
```
maedhroz: nit: We could make constants out of "PostingDecodes" and "NumPostings" now, I guess, since we use them multiple places, but not a big deal either way.

maedhroz (approved):

src/java/org/apache/cassandra/metrics/CassandraMetricsRegistry.java
```diff
@@ -93,34 +89,14 @@ public class CassandraMetricsRegistry extends MetricRegistry
 {
     public static final UnaryOperator<String> METRICS_GROUP_POSTFIX = name -> name + "_group";
 
-    /**
-     * A map of metric name constructed by {@link com.codahale.metrics.MetricRegistry#name(String, String...)} and
-     * its full name in the way how it is represented in JMX. The map is used by {@link CassandraJmxMetricsExporter}
-     * to export metrics to JMX.
-     */
-    private static final ConcurrentMap<String, Deque<MetricName>> ALIASES = new ConcurrentHashMap<>();
```
maedhroz, line 101: @krummas FYI, this should take care of [CASSANDRA-19571](https://issues.apache.org/jira/browse/CASSANDRA-19571)

Mmuzaf:
Committed:
https://github.com/apache/cassandra/commit/8619010cdca8bce471754e4fbeb861f036535007

PR #3268, opened 2024-04-19
CASSANDRA-19577 Make queries visible to the "system_views.queries" virtual table at the coordinator level

maedhroz:
patch by Caleb Rackliffe; reviewed by Chris Lohfink for CASSANDRA-19577

### DIFF
--- a/src/java/org/apache/cassandra/concurrent/TaskFactory.java
+++ b/src/java/org/apache/cassandra/concurrent/TaskFactory.java
@@ -32,7 +32,7 @@ import static org.apache.cassandra.utils.Shared.Scope.SIMULATION;
 /**
  * A simple mechanism to impose our desired semantics on the execution of a task without requiring a specialised
  * executor service. We wrap tasks in a suitable {@link FutureTask} or encapsulating {@link Runnable}.
- *
+ * <p>
  * The encapsulations handle any exceptions in our standard way, as well as ensuring {@link ExecutorLocals} are
  * propagated in the case of {@link #localAware()}
  */
@@ -52,7 +52,7 @@ public interface TaskFactory
     static TaskFactory standard() { return Standard.INSTANCE; }
     static TaskFactory localAware() { return LocalAware.INSTANCE; }
 
-    public class Standard implements TaskFactory
+    class Standard implements TaskFactory
     {
         static final Standard INSTANCE = new Standard();
         protected Standard() {}
@@ -90,8 +90,8 @@ public interface TaskFactory
         @Override
         public <T> RunnableFuture<T> toSubmit(WithResources withResources, Runnable runnable)
         {
-            return withResources.isNoOp() ? newTask(callable(runnable))
-                                          : newTask(withResources, callable(runnable));
+            return withResources.isNoOp() ? newTask(runnable)
+                                          : newTask(withResources, runnable);
         }
 
         @Override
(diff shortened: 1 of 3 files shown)


src/java/org/apache/cassandra/concurrent/TaskFactory.java
```diff
@@ -90,6 +90,10 @@ public Runnable toExecute(WithResources withResources, Runnable runnable)
         @Override
         public <T> RunnableFuture<T> toSubmit(WithResources withResources, Runnable runnable)
         {
+            if (runnable instanceof RunnableDebuggableTask)
+                return withResources.isNoOp() ? newTask(runnable)
+                                              : newTask(withResources, runnable);
+
             return withResources.isNoOp() ? newTask(callable(runnable))
                                           : newTask(withResources, callable(runnable));
         }
```
dcapwell: is there a reason to keep this logic and guard the above logic with `if (runnable instanceof RunnableDebuggableTask)`?  As far as I can tell there isn't a good reason... we convert the runnable to `callable` for the execution logic, and the debug logic deals with the raw type...

test/distributed/org/apache/cassandra/distributed/test/QueriesTableTest.java
```diff
@@ -18,72 +18,152 @@
...
 
+    private static Cluster SHARED_CLUSTER;
+    private static com.datastax.driver.core.Cluster DRIVER_CLUSTER;
+    private static Session SESSION;
+
+    @BeforeClass
+    public static void createCluster() throws IOException
+    {
+        SHARED_CLUSTER = init(Cluster.build(1).withConfig(c -> c.with(Feature.NATIVE_PROTOCOL)).start());
+        DRIVER_CLUSTER = com.datastax.driver.core.Cluster.builder().addContactPoint("127.0.0.1").build();
```
dcapwell: FYI `org.apache.cassandra.distributed.test.JavaDriverUtils#create(org.apache.cassandra.distributed.api.ICluster<? extends org.apache.cassandra.distributed.api.IInstance>)`

does the same thing in this case.

test/distributed/org/apache/cassandra/distributed/test/QueriesTableTest.java
```diff
@@ -18,72 +18,152 @@
...
+        {
+            for (int i = 0; i < ITERATIONS; i++)
+            {
+                SESSION.executeAsync("INSERT INTO " + KEYSPACE + ".tbl (k, v) VALUES (" + i + ", 0)");
+                SESSION.executeAsync("SELECT * FROM " + KEYSPACE + ".tbl WHERE k = " + (i - 1));
+            }
+
+            totalOperations += ITERATIONS;
+
+            SimpleQueryResult result = SHARED_CLUSTER.get(1).executeInternalWithResult("SELECT * FROM system_views.queries");
```
dcapwell: this is super racy and leaves you only validating that you saw reads/writes and not that they were "correct".  Might want to use BB to make this test stable and avoid timing out (keep running over and over trying to find the last query)

dcapwell (approved):

src/java/org/apache/cassandra/concurrent/TaskFactory.java
maedhroz: It certainly doesn't fail any tests to simplify this...

maedhroz:
Committed as https://github.com/apache/cassandra/commit/428fa1f4453735ef43d508621e5e7d9e0a054415

PR #3270, opened 2024-04-23
[CASSANDRA-15439] Add option to override the FatClient timeout for Bootstrapping nodes

rhuffy:
This implements the suggestion by @driftx in [CASSANDRA-15439](https://issues.apache.org/jira/browse/CASSANDRA-15439). See https://the-asf.slack.com/archives/CK23JSY2K/p1713820180414349 for further discussion.

If a bootstrapping node experiences a GC pause such that it fails to gossip for 30s, it may be removed from another node's Gossip state as it has exceeded the FatClient timeout, which by default is equal to RING_DELAY=30s.

This violates Cassandra's consistency guarantees. When there are pending ranges, the number of nodes that must ACK a write at quorum is increased. For example, with RF=3, quorum is 2. If there are pending ranges, quorum is 3. If a node is actually bootstrapping, but another node has removed it and believes there are no pending ranges, requests that it coordinates may only reach two other nodes. If the nodes reached are a node sending a stream (A) and one other node (B), then once the bootstrapping node joins the ring and takes ownership of the pending range, that write will only exist on node that owns the data (node B).

For operators who are particularly concerned with the durability of writes during expansion, this option can be set to a higher value - ideally longer than the longest expected GC pause. If this value is increased significantly, like to multiple hours, or set to -1, operators will need to take manual action when a bootstrap has failed, like assassinating the failed node.

Example usage:
```
-Dcassandra.bootstrapping_fat_client_timeout_ms=300000
```

### DIFF
--- a/src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
+++ b/src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
@@ -155,6 +155,8 @@ public enum CassandraRelevantProperties
 
     RING_DELAY("cassandra.ring_delay_ms"),
 
+    FAILED_BOOTSTRAP_TIMEOUT("cassandra.failed_bootstrap_timeout_ms"),
+
     /**
      * When bootstraping we wait for all schema versions found in gossip to be seen, and if not seen in time we fail
      * the bootstrap; this property will avoid failing and allow bootstrap to continue if set to true.
--- a/src/java/org/apache/cassandra/gms/Gossiper.java
+++ b/src/java/org/apache/cassandra/gms/Gossiper.java
@@ -293,7 +293,7 @@ public class Gossiper implements IFailureDetectionEventListener, GossiperMBean
         }
         else
         {
-            return FAT_CLIENT_TIMEOUT;
+            return FAT_CLIENT_TIMEOUT * 2;
         }
     }
 
--- a/src/java/org/apache/cassandra/gms/Gossiper.java
+++ b/src/java/org/apache/cassandra/gms/Gossiper.java
@@ -123,6 +123,9 @@ public class Gossiper implements IFailureDetectionEventListener, GossiperMBean
         SILENT_SHUTDOWN_STATES.add(VersionedValue.STATUS_BOOTSTRAPPING);
         SILENT_SHUTDOWN_STATES.add(VersionedValue.STATUS_BOOTSTRAPPING_REPLACE);
     }
+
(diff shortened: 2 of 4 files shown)


src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -155,6 +155,8 @@ public enum CassandraRelevantProperties
 
     RING_DELAY("cassandra.ring_delay_ms"),
 
+    BOOTSTRAPPING_FAT_CLIENT_TIMEOUT("cassandra.bootstrapping_fat_client_timeout_ms"),
+
```
driftx, line 159: I think we should name this something like FAILED_BOOTSTRAP_TIMEOUT, I don't want to expose 'fat client' to users since that is not immediately understandable and we have moved to calling them 'gossip only clients'

src/java/org/apache/cassandra/gms/Gossiper.java
```diff
@@ -123,6 +123,9 @@ public static class Props
         SILENT_SHUTDOWN_STATES.add(VersionedValue.STATUS_BOOTSTRAPPING);
         SILENT_SHUTDOWN_STATES.add(VersionedValue.STATUS_BOOTSTRAPPING_REPLACE);
     }
+
+    static final List<String> BOOTSTRAPPING_STATES = Arrays.asList(VersionedValue.STATUS_BOOTSTRAPPING,
```
dcapwell: This duplicates `org.apache.cassandra.gms.VersionedValue#BOOTSTRAPPING_STATUS`

src/java/org/apache/cassandra/gms/Gossiper.java
```diff
@@ -272,6 +278,25 @@ private static long getVeryLongTime()
         return 259200 * 1000; // 3 days
     }
 
+    private static long getFailedBootstrapTimeout()
+    {
+        String newtimeout = CassandraRelevantProperties.FAILED_BOOTSTRAP_TIMEOUT.getString();
+        if (newtimeout != null)
+        {
+            long longValue = Long.parseLong(newtimeout);
```
dcapwell, line 286: why not just call `getLong(-1)`?  These properties expose the parsing in a generic way

src/java/org/apache/cassandra/gms/Gossiper.java
rhuffy: 👍  That was added in 5.0. I can add it to this patch to minimize the diff between versions

PR #3274, opened 2024-04-29
CASSANDRA-19534 (5.0 patch): Unbounded queues in native transport requests lead to node instability

ifesdjeen:


src/java/org/apache/cassandra/transport/Dispatcher.java
```diff
@@ -99,8 +100,23 @@ public Dispatcher(boolean useLegacyFlusher)
...
+            // they will defunct the connection. Without a protocol version bump that introduces an "I am going away message",
+            // we have to stick to an existing error code.
+            Message.Response response = ErrorMessage.fromException(new OverloadedException("Server is shutting down"));
+            response.setStreamId(request.getStreamId());
+            response.setWarnings(ClientWarn.instance.getWarnings());
+            response.attach(request.connection);
+            FlushItem<?> toFlush = forFlusher.toFlushItem(channel, request, response);
+            flush(toFlush);
+            return;
+        }
```
maedhroz, line 119: Trying to find a test that hits this block/codepath...

src/java/org/apache/cassandra/concurrent/FutureTask.java
```diff
@@ -89,6 +89,34 @@ protected boolean tryFailure(Throwable t)
...
+                {
+                    return task.startTimeNanos();
+                }
+
+                public String description()
+                {
+                    return task.description();
+                }
+            };
+        }
```
maedhroz, line 118: Aside: I think there's actually a chance the changes I made to `TaskFactory` in CASSANDRA-19577 overlap w/ this. In other words, I think `callable()` doing this might actually make my `TaskFactory` changes unnecessary. I don't think our changes will break each other though...

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1283,4 +1283,21 @@ public static void log(Config config)
...
+    public volatile CQLStartTime cql_start_time = CQLStartTime.REQUEST;
+
+    public boolean native_transport_throw_on_overload = false;
+    public double native_transport_queue_max_item_age_threshold = Double.MAX_VALUE;
+    public DurationSpec.LongMillisecondsBound native_transport_min_backoff_on_queue_overload = new DurationSpec.LongMillisecondsBound("10ms");
+    public DurationSpec.LongMillisecondsBound native_transport_max_backoff_on_queue_overload = new DurationSpec.LongMillisecondsBound("200ms");
+
+    // 3.0 Cassandra Driver has its "read" timeout set to 12 seconds. Our recommendation is match this. Default is 100 seconds, which is extremely high
+    // but still not completely unbounded.
+    public DurationSpec.LongMillisecondsBound native_transport_timeout = new DurationSpec.LongMillisecondsBound("100000ms");
```
maedhroz: This default is likely moving to something lower, at least from what I'm reading in the Jira?  Is there any reference value in anything else on the classpath, or do we kind of just have to hard-code something w/ out-of-band knowledge. Do we want to use something like 20, 30...or go all the way to 12.

Also, which of these, if any, do we want to document in `cassandra.yaml`? I'm okay w/ leaving things out that we really don't think operators should change...

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1283,4 +1283,21 @@ public static void log(Config config)
     public double severity_during_decommission = 0;
 
     public StorageCompatibilityMode storage_compatibility_mode;
+
+    public enum CQLStartTime
+    {
+        REQUEST, // uses a timestamp that represent the start of processing of the request
+        QUEUE    // uses a timestamp that represents when the request was enqueued
+    }
+    public volatile CQLStartTime cql_start_time = CQLStartTime.REQUEST;
```
maedhroz, line 1296: nit: For this one, I guess we should clarify that it only applies to the replica timeout and `native_transport_timeout` still governs the overall coordinator-level request.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -990,6 +990,15 @@ else if (conf.max_value_size.toMebibytes() >= 2048)
...
         conf.sai_options.validate();
+
+        if (conf.native_transport_min_backoff_on_queue_overload.toMilliseconds() <= 0)
+            throw new IllegalArgumentException("Min backoff on queue overload should be positive");
+
+        if (conf.native_transport_min_backoff_on_queue_overload.toMilliseconds() >= conf.native_transport_max_backoff_on_queue_overload.toMilliseconds())
+            throw new IllegalArgumentException(String.format("Min backoff on queue overload should be strictly less than max backoff, but %s >= %s",
+                                                             conf.native_transport_min_backoff_on_queue_overload,
+                                                             conf.native_transport_max_backoff_on_queue_overload));
+
```
maedhroz, line 1001: nit: Should these be `ConfigurationException`?

nit: Should we use the actual YAML names in the error message? (ex. `native_transport_min_backoff_on_queue_overload`)

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2238,6 +2247,91 @@ public static long getMinRpcTimeout(TimeUnit unit)
...
+    {
+        if (minBackoffMillis <= 0)
+            throw new IllegalArgumentException("Min backoff on queue overload should be positive");
+
+        if (minBackoffMillis >= maxBackoffMillis)
+            throw new IllegalArgumentException(String.format("Max backoff on queue overload should be greater than min backoff, but %s >= %s", minBackoffMillis, maxBackoffMillis));
+
+
+        conf.native_transport_min_backoff_on_queue_overload = new DurationSpec.LongMillisecondsBound(minBackoffMillis, timeUnit);
+        conf.native_transport_max_backoff_on_queue_overload = new DurationSpec.LongMillisecondsBound(maxBackoffMillis, timeUnit);
```
maedhroz, line 2302: nit: I'm still not entirely sure if these need to be `volatile`?

src/java/org/apache/cassandra/config/Config.java
maedhroz:
```suggestion
    public DurationSpec.LongMillisecondsBound native_transport_timeout = new DurationSpec.LongMillisecondsBound("12s");
```
...or whatever we decide.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -2238,6 +2247,91 @@ public static long getMinRpcTimeout(TimeUnit unit)
...
+        return conf.cql_start_time;
+    }
+
+    public static void setCQLStartTime(Config.CQLStartTime value)
+    {
+        conf.cql_start_time = value;
+    }
+
+    /**
+     * How much time the item is allowed to spend in (currently only Native) queue, compared to {@link #nativeTransportIdleTimeout()},
```
maedhroz, line 2261: Compared to `nativeTransportIdleTimeout()` or to `native_transport_timeout`? It seems like in `Dispatcher` we have:

```
return requestExecutor.oldestTaskQueueTime() < (DatabaseDescriptor.getNativeTransportTimeout(TimeUnit.NANOSECONDS) * threshold);
```

Also, should the default threshold be `1.0` or `Double.MAX_VALUE`? Seems like the latter would blow up the value of `DatabaseDescriptor.getNativeTransportTimeout(TimeUnit.NANOSECONDS) * threshold` and almost guarantee `hasQueueCapacity()` returns `true`?

src/java/org/apache/cassandra/cql3/CQLStatement.java
```diff
@@ -75,9 +76,9 @@ default Iterable<Function> getFunctions()
      *
      * @param state the current query state
      * @param options options for this query (consistency, variables, pageSize, ...)
-     * @param queryStartNanoTime the timestamp returned by System.nanoTime() when this statement was received
+     * @param requestTime request enqueue / and start times
```
maedhroz: nit: Worth mentioning it's dependent on `cql_start_time`?

src/java/org/apache/cassandra/db/MutationVerbHandler.java
```diff
@@ -41,6 +43,13 @@ private void failed()
 
     public void doVerb(Message<Mutation> message)
     {
+        if (approxTime.now() > message.expiresAtNanos())
```
maedhroz, line 46: So the approximate times are safe to compare across nodes? (Looks like they're corrected w/ epoch time, but just making sure...)

src/java/org/apache/cassandra/db/ReadQuery.java
```diff
@@ -143,7 +144,7 @@ public ColumnFilter columnFilter()
      * @param state client state
      * @return the result of the query.
      */
-    public PartitionIterator execute(ConsistencyLevel consistency, ClientState state, long queryStartNanoTime) throws RequestExecutionException;
+    public PartitionIterator execute(ConsistencyLevel consistency, ClientState state, Dispatcher.RequestTime requestTime) throws RequestExecutionException;
```
maedhroz, line 148: nit: JavaDoc update while we're here?

src/java/org/apache/cassandra/net/Message.java
```diff
@@ -862,9 +879,9 @@ private Header deserializeHeader(DataInputPlus in, InetAddressAndPort peer, int
             long currentTimeNanos = approxTime.now();
             MonotonicClockTranslation timeSnapshot = approxTime.translate();
             long creationTimeNanos = calculateCreationTimeNanos(in.readInt(), timeSnapshot, currentTimeNanos);
-            long expiresAtNanos = getExpiresAtNanos(creationTimeNanos, currentTimeNanos, TimeUnit.MILLISECONDS.toNanos(in.readUnsignedVInt()));
-            Verb verb = Verb.fromId(in.readUnsignedVInt32());
-            int flags = in.readUnsignedVInt32();
+            long expiresAtNanos = getExpiresAtNanos(creationTimeNanos, TimeUnit.MILLISECONDS.toNanos(in.readUnsignedVInt()));
+            Verb verb = Verb.fromId(Ints.checkedCast(in.readUnsignedVInt()));
+            int flags = Ints.checkedCast(in.readUnsignedVInt());
```
maedhroz: nit: Is `in.readUnsignedVInt32()` not equivalent to `Ints.checkedCast(in.readUnsignedVInt())`?

src/java/org/apache/cassandra/service/paxos/v1/AbstractPaxosCallback.java
```diff
@@ -17,6 +17,8 @@
  */
 package org.apache.cassandra.service.paxos.v1;
 
+import org.apache.cassandra.transport.Dispatcher;
+import org.apache.cassandra.utils.Clock;
 import org.apache.cassandra.utils.concurrent.CountDownLatch;
```
maedhroz: nit: All 3 of these imports should move to their places below?

src/java/org/apache/cassandra/service/reads/AbstractReadExecutor.java
```diff
@@ -225,17 +229,27 @@ public boolean hasLocalRead()
...
-        if (cfs.sampleReadLatencyMicros > command.getTimeout(MICROSECONDS))
+        if (sampleLatencyNanos > command.getTimeout(NANOSECONDS))
         {
-            if (logger.isTraceEnabled())
-                logger.trace("Decided not to speculate as {} > {}", cfs.sampleReadLatencyMicros, command.getTimeout(MICROSECONDS));
+            logger.trace("Decided not to speculate as {}ns > {}ns", sampleLatencyNanos, command.getTimeout(NANOSECONDS));
             return false;
         }
 
+        if (now + sampleLatencyNanos > requestTime.clientDeadline())
```
maedhroz, line 241: Is the `enqueuedAtNanos` that informs `clientDeadline()` also a precise time?

src/java/org/apache/cassandra/service/reads/ReadCallback.java
```diff
@@ -27,6 +27,7 @@
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.db.MessageParams;
 import org.apache.cassandra.locator.ReplicaPlan;
+import org.apache.cassandra.transport.Dispatcher;
 import org.apache.cassandra.utils.concurrent.Condition;
```
maedhroz: nit: These imports are all out of place?

src/java/org/apache/cassandra/service/reads/ReadCallback.java
```diff
@@ -96,12 +95,28 @@ protected P replicaPlan()
...
+
+    /**
+     * In case of speculation, we want to time out the request immediately if we have _also_ hit a deadline.
+     *
+     * For example, we have a read timeout of 10s, 99% latency of 5 seconds, and time base is QUEUE:
+     *   * Request has spent 3 seconds in the queue. Here, we will wait for 2 seconds and try to speculate
+     *   * Request has spent 10 seconds in the queue. Here, we will wait for 0 seconds and try to speculate
+     *
+     *  If the time base is REQUEST:
+     *   * Request has spent 10 seconds in the queue. Here, we will only wait 2 seconds and then try to speculate
```
maedhroz, line 112: Assuming a `native_transport_timeout` of 12s? If so, good to mention?

src/java/org/apache/cassandra/service/AbstractWriteResponseHandler.java
```diff
@@ -34,6 +34,7 @@
 import org.apache.cassandra.locator.EndpointsForToken;
 import org.apache.cassandra.locator.ReplicaPlan;
 import org.apache.cassandra.locator.ReplicaPlan.ForWrite;
+import org.apache.cassandra.transport.Dispatcher;
```
maedhroz, line 37: nit: A bunch of imports out of place?

src/java/org/apache/cassandra/config/Config.java
```diff
@@ -1283,4 +1283,21 @@ public static void log(Config config)
...
+
+    public boolean native_transport_throw_on_overload = false;
+    public double native_transport_queue_max_item_age_threshold = Double.MAX_VALUE;
```
maedhroz, line 1299: nit: I want to try to work the token "ratio" into the name of this one, but haven't been able to come up w/ something concrete :D

src/java/org/apache/cassandra/tools/nodetool/DisableBinary.java
```diff
@@ -19,15 +19,19 @@
 
 import io.airlift.airline.Command;
 
+import io.airlift.airline.Option;
```
maedhroz, line 21: nit: Move up to be next to `Command` import

src/java/org/apache/cassandra/transport/CQLMessageHandler.java
```diff
@@ -178,7 +182,8 @@ protected boolean processOneContainedMessage(ShareableBytes bytes, Limit endpoin
 
         // max CQL message size defaults to 256mb, so should be safe to downcast
         int messageSize = Ints.checkedCast(header.bodySizeInBytes);
```
maedhroz, line 185: Not related to your patch, but I think I forgot to throw an `@Override` annotation on `processOneContainedMessage()`

src/java/org/apache/cassandra/transport/CQLMessageHandler.java
```diff
@@ -188,56 +193,67 @@ protected boolean processOneContainedMessage(ShareableBytes bytes, Limit endpoin
...
+
+                // If we triggered backpressure, make sure the caller stops processing frames after the request completes.
                 return false;
+            }
         }
 
         return processRequestAndUpdateMetrics(bytes, header, messageSize, Overload.NONE);
     }
 
+
```
maedhroz: nit: unnecessary newline?

src/java/org/apache/cassandra/transport/CQLMessageHandler.java
```diff
@@ -263,13 +279,24 @@ private void discardAndThrow(Limit endpointReserve, Limit globalReserve,
     }
 
     private OverloadedException buildOverloadedException(Limit endpointReserve, Limit globalReserve, Overload overload) {
```
maedhroz: nit: my fault, but let's move the `{` to the next line

src/java/org/apache/cassandra/transport/PreV5Handlers.java
```diff
@@ -172,20 +183,33 @@ private void checkLimits(ChannelHandlerContext ctx, Message.Request request)
...
+                {
+                    delay = queueBackpressure.markAndGetDelay(RATE_LIMITER_DELAY_UNIT);
+
+                    if (delay > 0)
+                        backpressure = Overload.QUEUE_TIME;
+                }
+
+                if (delay > 0)
+                {
+                    assert backpressure != Overload.NONE;
```
maedhroz: nit: Should this be just asserting that backpressure isn't `NONE` or should it be asserting that it's `REQUESTS` or `QUEUE_TIME`, which are the only things that would have an associated delay?

src/java/org/apache/cassandra/transport/PreV5Handlers.java
```diff
@@ -219,15 +243,30 @@ private void discardAndThrow(Message.Request request, long requestSize, Overload
...
+                                                                      endpointPayloadTracker));
+                    break;
+                case QUEUE_TIME:
+                    exception = new OverloadedException(String.format("Request has spent over %s time of the maximum timeout %dms in the queue",
+                                                                      DatabaseDescriptor.getNativeTransportQueueMaxItemAgeThreshold(),
+                                                                      DatabaseDescriptor.getNativeTransportTimeout(TimeUnit.MILLISECONDS)));
+                    break;
+                default:
+                    throw new IllegalArgumentException(String.format("Can't create an exception from %s", overload));
+            }
```
maedhroz: I wonder if we could combine this w/ `buildOverloadedException()` in `CQLMessageHandler`? (Pass the limier as an argument, etc.)

src/java/org/apache/cassandra/config/Config.java
maedhroz: Can we just do this?

```
public DurationSpec.LongMillisecondsBound native_transport_timeout = new DurationSpec.LongMillisecondsBound(SocketOptions.DEFAULT_READ_TIMEOUT_MILLIS, TimeUnit.MILLISECONDS);
```

Even if we can't, could we reference `{@link SocketOptions.DEFAULT_READ_TIMEOUT_MILLIS}` in the JavaDoc around the patch for posterity?

src/java/org/apache/cassandra/transport/Dispatcher.java
```diff
@@ -154,14 +323,49 @@ public String description()
         {
             return request.toString();
         }
+
+        @Override
+        public String toString()
+        {
+            return "RequestProcessor{" +
+                   "request=" + request +
```
maedhroz, line 333: nit: `Message` has a `toString()`, but `Request` does not, so we'll miss `Request.createdAtNanos`...if that matters.

src/java/org/apache/cassandra/transport/Dispatcher.java
```diff
@@ -154,14 +323,49 @@ public String description()
...
+    private static Message.Response processRequest(ServerConnection connection, Message.Request request, Overload backpressure, RequestTime requestTime)
     {
+        long queueTime = requestTime.timeSpentInQueueNanos();
+
+        // If we have already crossed the max timeout for all possible RPCs, we time out the query immediately.
+        // We do not differentiate between query types here, since if we got into a situation when, say, we have a PREPARE
+        // query that is stuck behind the EXECUTE query, we would rather time it out and catch up with a backlog, expecting
+        // that the bursts are going to be short-lived.
+        ClientMetrics.instance.queueTime(queueTime, TimeUnit.NANOSECONDS);
+        if (queueTime > DatabaseDescriptor.getNativeTransportTimeout(TimeUnit.NANOSECONDS))
```
maedhroz, line 365: nit: This is a hot-ish (?) path. Would it make sene to memoize the native transport timeout so we don't have to call `TimeUnit#covert()` so much?

src/java/org/apache/cassandra/transport/QueueBackpressure.java
```diff
@@ -0,0 +1,148 @@
...
+        {
+            // Last time we have applied backpressure was over a second ago, consider this a new incident
+            if (appliedAt > 0 && now - appliedAt >= TimeUnit.SECONDS.toNanos(1))
+            {
+                return new Impl(minDelayNanos, maxDelayNanos, now, 1, 1);
+            }
+            // Continuing incident: apply backpressure but do not bump severity level yet
+            else if (appliedTimes < 10)
+            {
+                return new Impl(minDelayNanos, maxDelayNanos, now, severityLevel == 0 ? 1 : severityLevel, appliedTimes + 1);
```
maedhroz: nit: Can we just start the `severityLevel` at 1?

src/java/org/apache/cassandra/transport/QueueBackpressure.java
```diff
@@ -0,0 +1,148 @@
...
+
+
+    static Impl noBackpressure(LongSupplier minDelayNanos, LongSupplier maxDelayNanos)
+    {
+        return new Impl(minDelayNanos, maxDelayNanos,
+                        -1, 0, 0);
+    }
+
+    @VisibleForTesting
+    class Impl
```
maedhroz:
```suggestion
    class Incident
```
WDYT?

src/java/org/apache/cassandra/transport/QueueBackpressure.java
maedhroz: nit: It seems like there's a lot of `Impl` creation going on here during a spike. Is there any way we could perhaps moderate that a tiny bit by perhaps making `appliedTimes` an `AtomicInteger`?

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
ifesdjeen: Right; the idea was that for starters we begin with this feature effectively disabled. Keep in mind I did not know how this feature is going to be perceived since it does change many notions simultaneously, but since both Jon and Brendon have tested it and found it useful, we can go with 1.0 here and 12 seconds of deadline as defaults.

src/java/org/apache/cassandra/transport/Dispatcher.java
ifesdjeen: Sure: this is tested in `DisableBinaryTest#testDisallowsNewRequests`. I did try to separate this into the going away commit, but have missed separating tests. Apologies!

src/java/org/apache/cassandra/concurrent/FutureTask.java
ifesdjeen: I think I have first tried to do it very similarly to how you did it, but this has ported better to other branches. Please let me know how you want to proceed.

src/java/org/apache/cassandra/config/Config.java
ifesdjeen: The problem is that some people expect that if their read timeout is 5 seconds, and request has spent 5 seconds in the queue, it will still be executed. Native request timeout is based off when we have admitted the request to the queue, which will make it time out after the 5 seconds in the queue. 

 I have changed the default to 12 seconds for now, which based on discussion with @driftx on Jira seems to be a consensus.

src/java/org/apache/cassandra/config/Config.java
ifesdjeen: Good point; this one was created before native transport timeout existed so i have not thought about it!

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
ifesdjeen: Good point; changed it here and above

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
ifesdjeen: They do not need to be volatile. We had a discussion about this with Aleksey and Benedict. I did not dare to remove `volatile` everywhere in the `Config` to avoid clutter, but in general we should do this.

src/java/org/apache/cassandra/config/Config.java
ifesdjeen: 👍 switched to 12s

src/java/org/apache/cassandra/cql3/CQLStatement.java
ifesdjeen: Mentioned this in `baseTimeNanos` if you are OK with this

src/java/org/apache/cassandra/db/MutationVerbHandler.java
ifesdjeen: My logic here was that:
a) approx time is _really_ bad only for metrics (but even then, only for idle/very fast clusters)
b) relative time between replicas is controlled by something like NTP anyways, and operators should know their drift
c) diff between approx and exact time is up to 2 micros, which means that for cross-node it'll be up to 4 micros; if we time out request 4 micros too early, I think it is more than fine, we should not operate that close to timeouts anyways

src/java/org/apache/cassandra/net/Message.java
ifesdjeen: Oh, I did not know it existed. Thank you!

src/java/org/apache/cassandra/service/reads/AbstractReadExecutor.java
ifesdjeen: Yes. `enqueuedAtNanos` is precise: 

in `Request` constructor:
```
createdAtNanos = MonotonicClock.Global.preciseTime.now();
```

and then

```
processRequest(channel, request, forFlusher, backpressure, new RequestTime(request.createdAtNanos, startTimeNanos));
```

Also, all metrics are using precise time.

src/java/org/apache/cassandra/service/paxos/v1/AbstractPaxosCallback.java
ifesdjeen: interesting; not sure why my intellij has moved them

src/java/org/apache/cassandra/service/reads/ReadCallback.java
ifesdjeen: Must have been merge. Thank you for spotting!

src/java/org/apache/cassandra/service/reads/ReadCallback.java
ifesdjeen: Good point!

src/java/org/apache/cassandra/transport/Dispatcher.java
ifesdjeen: Did this: 
```
    private static long native_transport_timeout_nanos_cached = -1;

    public static long getNativeTransportTimeout(TimeUnit timeUnit)
    {
        if (timeUnit == TimeUnit.NANOSECONDS)
        {
            if (native_transport_timeout_nanos_cached == -1)
                native_transport_timeout_nanos_cached = conf.native_transport_timeout.to(TimeUnit.NANOSECONDS);

            return native_transport_timeout_nanos_cached;
        }
        return conf.native_transport_timeout.to(timeUnit);
    }
```

But arguably we should have a more generic pattern for these things. Maybe even always precompute millis nanos and micros. Should be extremely cheap, and constant time, if we use a tiny array.

src/java/org/apache/cassandra/transport/PreV5Handlers.java
ifesdjeen: Right, so `backpressure != Overload.NONE` is tauthological at that spot. Changed to 

```
assert backpressure == Overload.REQUESTS || backpressure == Overload.QUEUE_TIME : backpressure;
```

src/java/org/apache/cassandra/transport/PreV5Handlers.java
ifesdjeen: Sure, great idea.

src/java/org/apache/cassandra/transport/QueueBackpressure.java
ifesdjeen: Unfortunately because we have at least two variables that we need to update atomically, now and appliedTimes, we will either have to create some sort of object, or do some binary math (but then we lose precision). I am afraid I could not find a quick and easy way to make this more lightweight.

I would also like to highlight that as soon as we have applied timeout, we will have the client back-off for the given amount of time, so this might be less of a problem: we do this only if there is no capacity in the queue.

Also, incident does start at 1, could you elaborate?

src/java/org/apache/cassandra/concurrent/FutureTask.java
maedhroz: As long as the tests I wrote in 19577 still pass, I'm 100% fine if you want to revert those bits from my patch, unless doing it that way entails more object creation.

src/java/org/apache/cassandra/transport/QueueBackpressure.java
maedhroz: I was just wondering why we had to do `severityLevel == 0 ? 1 : severityLevel` rather than just starting incidents at 1, but that means you have to check `appliedAt` or something in `delay()` to make sure you get zero before an incident has actually started.

src/java/org/apache/cassandra/transport/Dispatcher.java
```diff
@@ -99,8 +100,24 @@ public Dispatcher(boolean useLegacyFlusher)
...
+            // We can not respond with a custom, transport, or server exceptions since, given current implementation of clients,
+            // they will defunct the connection. Without a protocol version bump that introduces an "I am going away message",
+            // we have to stick to an existing error code.
+            Message.Response response = ErrorMessage.fromException(new OverloadedException("Server is shutting down"));
+            response.setStreamId(request.getStreamId());
+            response.setWarnings(ClientWarn.instance.getWarnings());
+            response.attach(request.connection);
+            FlushItem<?> toFlush = forFlusher.toFlushItem(channel, request, response);
+            flush(toFlush);
+            System.out.println(123123);
```
maedhroz, line 117: TODO: remove `println`

src/java/org/apache/cassandra/transport/PreV5Handlers.java
```diff
@@ -19,9 +19,11 @@
 package org.apache.cassandra.transport;
 
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 
 import com.google.common.base.Predicate;
 
+import com.sun.jna.platform.win32.GL;
```
maedhroz, line 26: nit: unused import

maedhroz (approved):

src/java/org/apache/cassandra/transport/Dispatcher.java
ifesdjeen: Whoops!

maedhroz:
Committed as https://github.com/apache/cassandra/commit/617a75843c9bfaf241249514f9604466f6c8ccab

PR #3280, opened 2024-04-30
Use per CFS queue to track pending flushes

aweisberg:


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -3993,7 +3993,7 @@ protected synchronized void drain(boolean isFinalShutdown) throws IOException, I
             CommitLog.instance.shutdownBlocking();
 
             // wait for miscellaneous tasks like sstable and commitlog segment deletion
-            ColumnFamilyStore.shutdownPostFlushExecutor();
+            ColumnFamilyStore.awaitPendingFlushes();
```
aweisberg, line 3996: I am concerned that this might get stuck waiting pending flushes and then never actually finish because the executors or components that might finish them never get to run. Need to check this when I get back to working on this.

PR #3288, opened 2024-05-03
CASSANDRA-19158

ifesdjeen:


src/java/org/apache/cassandra/tcm/EpochAwareDebounce.java
```diff
@@ -62,66 +45,46 @@ private EpochAwareDebounce()
...
-                    }
-                    catch (Throwable t)
-                    {
-                        fetchResult.cancel(true);
-                        inflightRequests.remove(fetchResult);
-                        promise.setFailure(t);
-                    }
-                });
-                return promise;
+                EpochAwareFuture promise = new EpochAwareFuture(epoch, fetchFunction.get());;
```
beobal, line 71: nit: `;;`

src/java/org/apache/cassandra/tcm/EpochAwareDebounce.java
```diff
@@ -62,66 +45,46 @@ private EpochAwareDebounce()
      * calls to peers, retrieving a LogState which can be applied locally to produce the necessary {@code
      * ClusterMetadata}. The function takes a {@code Promise<LogState>} as input, with the expectation that this
      * specific instance will be used to provide blocking behaviour when making the rpc calls that fetch the {@code
```
beobal, line 48: should remove this too: `The function takes a {@code Promise<LogState>} as input, with the expectation that this specific instance will be used to provide blocking behaviour when making the rpc calls that fetch the {@code
LogState}.`

src/java/org/apache/cassandra/tcm/EpochAwareDebounce.java
```diff
@@ -62,66 +45,46 @@ private EpochAwareDebounce()
...
-     * LogState}. These promises are memoized in order to cancel them when {@link #shutdownAndWait(long, TimeUnit)} is
-     * called. This causes the fetch function to stop waiting on any in flight {@code LogState} requests and prevents
-     * shutdown from being blocked.
+     * LogState}.
      *
      * @param  fetchFunction executes the request for LogState. It's expected that this popluates fetchResult with the
```
beobal, line 51: And probably reword this too

src/java/org/apache/cassandra/tcm/PeerLogFetcher.java
```diff
@@ -74,40 +74,54 @@ public ClusterMetadata fetchLogEntriesAndWait(InetAddressAndPort remote, Epoch a
...
+                log.append(logState);
+                ClusterMetadata fetched = log.waitForHighestConsecutive();
+                if (fetched.epoch.isEqualOrAfter(awaitAtleast))
+                {
+                    TCMMetrics.instance.peerLogEntriesFetched(before, logState.latestEpoch());
+                    return fetched;
+                }
+                else
+                {
+                    throw new IllegalStateException(String.format("Queried for epoch %s, but could not catch up", awaitAtleast));
```
beobal, line 111: do we want to throw here, or let the caller decide how to handle a failure to catch up? In some cases, this is purely a background task, so perhaps we should log but allow the caller to proceed? (I know we are certain that `remote` should be able to satisfy the request as we call this in response to seeing `awaitAtLeast` in some message from it).

src/java/org/apache/cassandra/tcm/RemoteProcessor.java
```diff
@@ -152,34 +150,37 @@ public ClusterMetadata fetchLogAndWait(Epoch waitFor, Retry.Deadline retryPolicy
 
     public static ClusterMetadata fetchLogAndWait(CandidateIterator candidateIterator, LocalLog log)
     {
-        Promise<LogState> remoteRequest = new AsyncPromise<>();
-        return fetchLogAndWaitInternal(remoteRequest, candidateIterator, log);
+        try
+        {
+            return fetchLogAndWaitInternal(candidateIterator, log).awaitUninterruptibly().get();
```
beobal, line 155: Should we `awaitThrowUncheckedOnInterrupt`, or even just `await` here? I'm thinking of tests really where the origin implementation (before CASSANDRA-19514) could cause occasional timeouts.

beobal (changes requested):
LGTM, just a handful of minor comments. There are some test failures in relevant places too that we should investigate.

src/java/org/apache/cassandra/tcm/PeerLogFetcher.java
ifesdjeen: Right now, we will log this (rare) exception, which ji think is a reasonable behaviour. 

On the other hand, if we have a listener, it will have to deal with exception one way or the other: this seems ok, too.

src/java/org/apache/cassandra/tcm/RemoteProcessor.java
ifesdjeen: Do you mean `await` with a timeout, or without? My logic for doing it this way was that we will rely on messaging service timeouts now to time this one out, so it should never be indefinite.

src/java/org/apache/cassandra/tcm/PeerLogFetcher.java
beobal: ok, seems reasonable

src/java/org/apache/cassandra/tcm/RemoteProcessor.java
beobal: fair enough, I was concerned about in-flight requests preventing a timely shutdown, but this will now be bounded by `new Retry.Backoff(TCMMetrics.instance.fetchLogRetries));`, right?

PR #3289, opened 2024-05-03
CASSANDRA-19617: Refresh stale paxos commit

belliottsmith:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/db/SystemKeyspace.java
+++ b/src/java/org/apache/cassandra/db/SystemKeyspace.java
@@ -125,6 +125,7 @@ import org.apache.cassandra.utils.concurrent.Future;
 import static java.lang.String.format;
 import static java.util.Collections.emptyMap;
 import static java.util.Collections.singletonMap;
+import static java.util.concurrent.TimeUnit.MICROSECONDS;
 import static org.apache.cassandra.config.Config.PaxosStatePurging.legacy;
 import static org.apache.cassandra.config.DatabaseDescriptor.paxosStatePurging;
 import static org.apache.cassandra.cql3.QueryProcessor.executeInternal;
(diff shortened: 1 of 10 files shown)


src/java/org/apache/cassandra/db/SystemKeyspace.java
```diff
@@ -1373,14 +1376,38 @@ public static PaxosState.Snapshot loadPaxosState(DecoratedKey partitionKey, Tabl
...
         }
 
+        long purgeBefore = 0;
+        long overrideTtlSeconds = 0;
+        switch (paxosStatePurging())
+        {
+            default: throw new AssertionError();
+            case legacy:
+            case gc_grace:
+                overrideTtlSeconds = metadata.params.gcGraceSeconds;
```
bdeggleston, line 1386: I think this creates another window for state resurrection if you switch back to repaired without having run a paxos repair in the meantime though, since you'd no longer be applying a synthetic ttl to data that had previously been synthetically ttl’d. The timeline would be: you have incomplete operation A happen with purging mode `repaired`, you switch to `gc_grace` and the ttl window elapses, you run operation B, which deletes the data that would have been written by operation A, the ttl window for that operation passes and it’s removed from disk, you switch back to `repaired` purging without having run a repair in the meantime, operation A becomes visible again and is applied.

Re-reading the PaxosStatePurging docs, we only state that it’s unsafe to go back to legacy once migrating from it, which implies that it’s ok to return to gc_grace if you’re having repair problems and a growing paxos state table, which I think is a useful escape hatch to keep around.

Switching back to gc_grace would also prevent the operations written while in repaired mode from being purged properly. While we’re on the topic, increasing gc grace when using the gc grace purge mode could also resurrect older operations by the same general mechanism described in the JIRA.

I think we need to track a local purgeable low bound across config and schema changes.

bdeggleston (changes requested):

pmcfadin:
Closing as [CASSANDRA-19617](https://issues.apache.org/jira/browse/CASSANDRA-19617) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #3293, opened 2024-05-03
Cassandra 19618

dcapwell:


src/java/org/apache/cassandra/config/AccordSpec.java
```diff
@@ -71,4 +73,62 @@ public enum TransactionalRangeMigration
     public TransactionalMode default_transactional_mode = TransactionalMode.off;
     public boolean ephemeralReadEnabled = false;
     public boolean state_cache_listener_jfr_enabled = true;
+    public final JournalSpec journal = new JournalSpec();
+
+    public static class JournalSpec implements Params
+    {
+        public int segmentSize = 32 << 20;
```
dcapwell, line 80: we don't actually need to have our variables as `segment_size` as the yaml layer converts camel case to the expected naming...   this is *only* `segment_size` in yaml, same for every field

src/java/org/apache/cassandra/config/AccordSpec.java
```diff
@@ -71,4 +73,62 @@ public enum TransactionalRangeMigration
...
+
+        @JsonIgnore
+        @Override
+        public int flushPeriodMillis()
+        {
+            return flushPeriod == null ? DatabaseDescriptor.getCommitLogSyncPeriod()
+                                       : flushPeriod.toMilliseconds();
+        }
+
+        @JsonIgnore
```
dcapwell, line 112: I think this isn't needed as `get` is needed for it to be picked up... I was just trying to be explicit...

dcapwell:
merged

PR #3296, opened 2024-05-08
CASSANDRA-19624: ModificationStatement#casInternal leaks RowIterator

michaeljmarshall:
Fixes https://issues.apache.org/jira/browse/CASSANDRA-19624

This patch closes the RowIterator created in the ModificationStatement#casInternal method, preventing the leak of resources initialized for the RowIterator.

This is my first C* patch, please let me know if I need to change anything.

### DIFF
--- a/src/java/org/apache/cassandra/cql3/statements/ModificationStatement.java
+++ b/src/java/org/apache/cassandra/cql3/statements/ModificationStatement.java
@@ -652,8 +652,7 @@ public abstract class ModificationStatement implements CQLStatement.SingleKeyspa
         SinglePartitionReadQuery readCommand = request.readCommand(nowInSeconds);
         FilteredPartition current;
         try (ReadExecutionController executionController = readCommand.executionController();
-             PartitionIterator iter = readCommand.executeInternal(executionController);
-             RowIterator rowIterator = PartitionIterators.getOnlyElement(iter, readCommand))
+             RowIterator rowIterator = PartitionIterators.getOnlyElement(readCommand.executeInternal(executionController), readCommand))
         {
             // FilteredPartition consumes the row but does not close the iterator
             current = FilteredPartition.create(rowIterator);
--- a/src/java/org/apache/cassandra/cql3/statements/ModificationStatement.java
+++ b/src/java/org/apache/cassandra/cql3/statements/ModificationStatement.java
@@ -652,9 +652,11 @@ public abstract class ModificationStatement implements CQLStatement.SingleKeyspa
         SinglePartitionReadQuery readCommand = request.readCommand(nowInSeconds);
         FilteredPartition current;
         try (ReadExecutionController executionController = readCommand.executionController();
-             PartitionIterator iter = readCommand.executeInternal(executionController))
+             PartitionIterator iter = readCommand.executeInternal(executionController);
+             RowIterator rowIterator = PartitionIterators.getOnlyElement(iter, readCommand))
         {
-            current = FilteredPartition.create(PartitionIterators.getOnlyElement(iter, readCommand));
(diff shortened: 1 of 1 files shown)


jacek-lewandowski (approved):

src/java/org/apache/cassandra/cql3/statements/ModificationStatement.java
```diff
@@ -652,9 +652,11 @@ static RowIterator casInternal(CQL3CasRequest request, long timestamp, int nowIn
         SinglePartitionReadQuery readCommand = request.readCommand(nowInSeconds);
         FilteredPartition current;
         try (ReadExecutionController executionController = readCommand.executionController();
-             PartitionIterator iter = readCommand.executeInternal(executionController))
+             PartitionIterator iter = readCommand.executeInternal(executionController);
+             RowIterator rowIterator = PartitionIterators.getOnlyElement(iter, readCommand))
```
jacek-lewandowski: Actually, one thing I'm wondering - is it possible that we may cause a double close of the `iter` (`PartitionIterator`) as the returned row iterator includes a "close transformation"? If so, is it a problem?

michaeljmarshall: I wrote a bit on that point here https://github.com/datastax/cassandra/pull/1103#issue-2278426122.

> Note that it doesn't look like the `Close` class defined in the `PartitionIterators#getOnlyElement` method is ever called. I'm not sure if that's an issue or not. However, I added some basic logging to the chunk cache during my testing, and I can see that this PR's change fixes a reference leak.

In my quick testing, I found that the `PartitionIterator` wasn't closed via the close transformation. It is possible that my debugging wasn't representative, as I am somewhat new to this code.

michaeljmarshall: Just tested again, and I was wrong, that method is called. Sorry for the confusion.

michaeljmarshall: I removed the `iter` from the try-with-resources block, so it won't be double closed now. I can see with code coverage that it is getting run via the `PreparedStatementsTest`. Resolving this for now, let me know if it needs to change.

michaeljmarshall: Though, we're at risk of leaking the `PartitionIterator` in the event of an exception in the `PartitionIterators.getOnlyElement`.

jacek-lewandowski: I know, all in all your patch fixes a leak which was occurring all the time, leaving a leak occurring only in case of a failure :-) I suppose this is not the only place vulnerable to that problem, in other words, perhaps it can be a subject of another ticket.

jacek-lewandowski (approved):

jacek-lewandowski:
https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1355/workflows/c258dce2-1174-4347-9d54-9ea0b5ea0ca1 (j8) 🆗

jacek-lewandowski:
for 4.1 https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1357/workflows/165f3829-ca68-4700-a33d-4a34184a9846 🆗 

for 5.0 https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1358/workflows/7cac1e4f-93a7-4995-9f87-84cdd6b80cf9 🆗 

for 5.1 https://app.circleci.com/pipelines/github/jacek-lewandowski/cassandra/1359/workflows/ac4d59cb-e068-407b-9216-e2b08135b4c8 🆗

PR #3297, opened 2024-05-08
cep-15-accord 5/7/2024 trunk rebase

maedhroz:


src/java/org/apache/cassandra/dht/IPartitionerDependentSerializer.java
```diff
@@ -51,11 +41,8 @@ public interface IPartitionerDependentSerializer<T>
...
-    /**
-     * Calculate serialized size of object without actually serializing.
-     * @param t object to calculate serialized size
-     * @param version protocol version
-     * @return serialized size of object t
-     */
-    public long serializedSize(T t, int version);
+    default T deserialize(DataInputPlus in, int version) throws IOException
+    {
+        return deserialize(in, null, version);
```
beobal, line 46: Would it make more sense to default to `IPartitioner.global`, rather than `null`?

beobal (commented):
I think it would make more sense to revert `StreamSummarySerializer` to `IVersionedSerializer` and always get the cfs using the tableid. `SyncResponse.serializer::deserialize` is its only caller that doesn't explicitly pass a null partitioner and even that will always pass null as its only non-test callsite is `Message::deserialize` which uses the `default` impl. So `SyncResponse.serializer` needn't really implement `IPartitionerDependentSerializer` either.

maedhroz:
@beobal I think the only reason StreamSummarySerializer doesn’t always get the partitioner from the TableId/CFS is SerializationsTest et al. Perhaps there’s a less janky way to do that…

beobal:
How about https://github.com/beobal/cassandra/commit/e8a6c58662cb51bbd42b1c848292a122acb2eec7

beobal:
Disclaimer: I didn't do a full CI run on that yet, so I don't know which other tests are affected

maedhroz:
`cep-15-accord` is now rebased to `trunk` at 534da538d229a5b4fc1d9ef7cc8b37ead455191c

Pre-rebase backup exists at https://github.com/apache/cassandra/tree/cep-15-accord-pre-5-7-trunk-rebase

PR #3305, opened 2024-05-13
CASSANDRA-18732 Baseline Diagnostic vtables for Accord

maedhroz:
Accord PR: https://github.com/apache/cassandra-accord/pull/89

### DIFF
--- a/src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
+++ b/src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
@@ -18,57 +18,230 @@
 
 package org.apache.cassandra.db.virtual;
 
-import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
-import com.google.common.annotations.VisibleForTesting;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import accord.local.CommandStores;
+import accord.primitives.TxnId;
+import accord.utils.async.AsyncChain;
+import accord.utils.async.AsyncChains;
 import org.apache.cassandra.config.DatabaseDescriptor;
 import org.apache.cassandra.cql3.statements.schema.CreateTableStatement;
+import org.apache.cassandra.db.ColumnFamilyStore;
+import org.apache.cassandra.db.DecoratedKey;
+import org.apache.cassandra.db.Keyspace;
+import org.apache.cassandra.db.marshal.UTF8Type;
+import org.apache.cassandra.dht.Range;
+import org.apache.cassandra.dht.Token;
+import org.apache.cassandra.exceptions.InvalidRequestException;
+import org.apache.cassandra.schema.Schema;
+import org.apache.cassandra.schema.TableId;
 import org.apache.cassandra.schema.TableMetadata;
+import org.apache.cassandra.service.accord.AccordCommandStore;
+import org.apache.cassandra.service.accord.AccordKeyspace;
 import org.apache.cassandra.service.accord.AccordService;
-import org.apache.cassandra.service.accord.IAccordService;
+import org.apache.cassandra.service.accord.AccordStateCache;
+import org.apache.cassandra.service.consensus.migration.ConsensusMigrationState;
+import org.apache.cassandra.service.consensus.migration.TableMigrationState;
+import org.apache.cassandra.tcm.ClusterMetadata;
+import org.apache.cassandra.utils.Clock;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
 
 public class AccordVirtualTables
 {
-    private AccordVirtualTables()
-    {
-
-    }
+    private AccordVirtualTables() {}
 
     public static Collection<VirtualTable> getAll(String keyspace)
     {
         if (!DatabaseDescriptor.getAccordTransactionsEnabled())
             return Collections.emptyList();
 
-        return Arrays.asList(
-        new Epoch(keyspace)
+        return List.of(
+            new CommandStoreCache(keyspace),
+            new MigrationState(keyspace),
+            new CoordinationStatus(keyspace)
         );
     }
 
-    @VisibleForTesting
-    public static final class Epoch extends AbstractVirtualTable
+    public static final class CommandStoreCache extends AbstractVirtualTable
     {
-
-        protected Epoch(String keyspace)
+        private CommandStoreCache(String keyspace)
         {
-            super(parse(keyspace, "Accord Epochs",
-                        "CREATE TABLE accord_epochs(\n" +
-                        "  epoch bigint,\n" +
-                        "  PRIMARY KEY ( (epoch) )" +
-                        ")"));
+            super(parse(keyspace,
+                        "Accord Command Store Cache Metrics",
+                        "CREATE TABLE accord_command_store_cache(\n" +
+                        "  id int,\n" +
+                        "  scope text,\n" +
+                        "  queries bigint,\n" +
+                        "  hits bigint,\n" +
+                        "  misses bigint,\n" +
+                        "  PRIMARY KEY (id, scope)" +
+                        ')'));
         }
 
         @Override
         public DataSet data()
         {
-            IAccordService accord = AccordService.instance();
+            CommandStores stores = ((AccordService) AccordService.instance()).node().commandStores();
 
-            long epoch = accord.currentEpoch();
+            AsyncChain<List<Map<String, AccordStateCache.ImmutableStats>>> statsByStoreChain = stores.map(store -> {
+                Map<String, AccordStateCache.ImmutableStats> snapshots = new HashMap<>(3);
+                AccordCommandStore accordStore = (AccordCommandStore) store.commandStore();
+                snapshots.put(AccordKeyspace.COMMANDS, accordStore.commandCache().statsSnapshot());
+                snapshots.put(AccordKeyspace.COMMANDS_FOR_KEY, accordStore.commandsForKeyCache().statsSnapshot());
+                snapshots.put(AccordKeyspace.TIMESTAMPS_FOR_KEY, accordStore.timestampsForKeyCache().statsSnapshot());
+                return snapshots;
+            });
 
+            List<Map<String, AccordStateCache.ImmutableStats>> statsByStore = AsyncChains.getBlockingAndRethrow(statsByStoreChain);
             SimpleDataSet result = new SimpleDataSet(metadata());
-            result.row(epoch);
+
+            for (int storeID : stores.ids())
+            {
+                Map<String, AccordStateCache.ImmutableStats> storeStats = statsByStore.get(storeID);
+                addRow(storeStats.get(AccordKeyspace.COMMANDS), result, storeID, AccordKeyspace.COMMANDS);
+                addRow(storeStats.get(AccordKeyspace.COMMANDS_FOR_KEY), result, storeID, AccordKeyspace.COMMANDS_FOR_KEY);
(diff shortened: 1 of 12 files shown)


maedhroz:
Accord PR: https://github.com/apache/cassandra-accord/pull/89

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
```diff
@@ -42,22 +46,24 @@ public static Collection<VirtualTable> getAll(String keyspace)
...
-        return Arrays.asList(
-        new Epoch(keyspace)
+        return List.of(
+            new Epoch(keyspace),
+            new CommandStoreCache(keyspace)
         );
     }
 
     @VisibleForTesting
     public static final class Epoch extends AbstractVirtualTable
```
dcapwell: not for this patch... but wondering if we should drop this table... I added it as a hack when we didn't have TCM...

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
```diff
@@ -73,6 +79,43 @@ public DataSet data()
...
 
+    public static final class CommandStoreCache extends AbstractVirtualTable
+    {
+        private CommandStoreCache(String keyspace)
+        {
+            super(parse(keyspace,
+                        "Accord Command Store Cache Metrics",
+                        "CREATE TABLE accord_command_store_cache(\n" +
+                        "  id int,\n" +
+                        "  queries bigint,\n" +
```
dcapwell: one issue with this table is it uses the global stats and not the instance stats... I kinda feel we should expose both?

What I see in practice is that `commands` and `commandsForKeys` have very different cache hit stats and the global is showing it for all...  If you look at the patch I did to add JFR support I show the stats per instance with a label so we know where its coming from... that helps show if commands are getting flushed out constantly in favor of CFK

dcapwell (changes requested):

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
```diff
@@ -73,6 +90,88 @@ public DataSet data()
...
+            ConsensusMigrationState snapshot = cm.consensusMigrationState;
+
+            for (TableMigrationState state : snapshot.tableStates())
+            {
+                result.row(state.keyspaceName, state.tableName);
+                result.column("table_id", state.tableId.asUUID());
+                result.column("target_protocol", state.targetProtocol.toString());
+
+                List<String> primitiveMigratedRanges = state.migratedRanges.stream().map(Objects::toString).collect(toImmutableList());
+                result.column("migrated_ranges", PojoToString.pojoToString(primitiveMigratedRanges, "MINIFIED-JSON"));
```
dcapwell: makes me wonder if `frozen<list<text>>` is better...  we wouldn't need to convert it at all and it would be easier for users to access in CQL?

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
```diff
@@ -73,6 +90,88 @@ public DataSet data()
...
+
+                Map<Long, List<String>> primitiveRangesByEpoch = new LinkedHashMap<>();
+                for (Map.Entry<org.apache.cassandra.tcm.Epoch, List<Range<Token>>> entry : state.migratingRangesByEpoch.entrySet())
+                    primitiveRangesByEpoch.put(entry.getKey().getEpoch(), entry.getValue().stream().map(Objects::toString).collect(toImmutableList()));
+
+                result.column("migrating_ranges_by_epoch", PojoToString.pojoToString(primitiveRangesByEpoch, "MINIFIED-JSON"));
```
dcapwell: similar comment, `frozen<map<int64, list<text>>>`?

src/java/org/apache/cassandra/utils/PojoToString.java
```diff
@@ -120,7 +120,7 @@ public static String pojoMapToString(Map<String, Object> map, String formatStrin
         return pojoToString(map, formatString);
     }
 
-    private static String pojoToString(Object obj, String formatString)
+    public static String pojoToString(Object obj, String formatString)
```
dcapwell: not for this patch: why isn't the second argument `Format`?

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
```diff
@@ -73,6 +90,88 @@ public DataSet data()
...
+
+            return result;
+        }
+    }
+
+    public static final class MigrationStates extends AbstractVirtualTable
+    {
+        private MigrationStates(String keyspace)
+        {
+            super(parse(keyspace,
```
dcapwell, line 136: how do I know the migration is *done*?

dcapwell (commented):
for the migration states table, one of the reasons for using the list/map collections is also better documentation.  `text` is flexible for us but once people start parsing (tools) then we can't actually change things... so we are stuck with the format...

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
maedhroz: Yeah, I think I'll just clean it up/remove it while I'm here.

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
maedhroz: I think we've now discussed this bit offline. For posterity, mentioned in https://issues.apache.org/jira/browse/CASSANDRA-18732?focusedCommentId=17848391&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17848391

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
maedhroz: Does a range move from "migrating" -> "migrated"?

src/java/org/apache/cassandra/utils/PojoToString.java
maedhroz: I'm guessing it's because it starts as a string in `listConsensusMigrations()` and we just didn't convert earlier. For my patch, it looks like I may actually be able to avoid `PojoToString` altogether, looking at the comments above...

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
maedhroz: Yeah, this and the comment for `migrating_ranges_by_epoch` make sense. Let's try frozen collections instead...

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
```diff
@@ -18,57 +18,162 @@
...
+            new MigrationStates(keyspace),
+            new CoordinationStatus(keyspace)
         );
     }
 
-    @VisibleForTesting
-    public static final class Epoch extends AbstractVirtualTable
+    public static final class CommandStoreCache extends AbstractVirtualTable
     {
+        private CommandStoreCache(String keyspace)
```
dcapwell, line 79: it would be useful to include instance as well, either as its own table or a column here... having a low hit rate doesn't let you know what's going on, need to see which instances are contributing to it.

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
dcapwell: @aweisberg?

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
maedhroz: Is there even any reason to populate this w/ the global `Stats`? Would it make more sense to expose the `Stats` from `commandCache`, `timestampsForKeyCache`, and `commandsForKeyCache` by store ID instead?

dcapwell: the global one is def nice to have, but if you are trying to debug an issue you kinda need to know which "instance" is having the issues... if command cache is 99% cache hit and the global miss rate is 50%, you know something is going on with the CFK caches (but which one?!?!)

The `system_metrics` vtables also expose the JMX metrics, which is the JVM level (includes each store)

TL;DR - I personally feel that cmd/cfk cache level is more useful than global and a global is nice but not required (slight duplicate with JMX table)

test/distributed/org/apache/cassandra/distributed/test/accord/AccordMigrationTest.java
```diff
@@ -696,6 +705,27 @@ private static void assertMigrationState(String tableName, ConsensusMigrationTar
         });
     }
 
+    private static void assertVtableMigrationState(String tableName, List<Range<Token>> migratedRanges, List<Range<Token>> migratingRanges) throws JsonProcessingException
```
maedhroz: nit: need to get rid of `JsonProcessingException`

test/distributed/org/apache/cassandra/distributed/test/accord/AccordMigrationTest.java
```diff
@@ -696,6 +705,27 @@ private static void assertMigrationState(String tableName, ConsensusMigrationTar
...
+    {
+        SimpleQueryResult vtableResult =
+            SHARED_CLUSTER.get(1).executeInternalWithResult("SELECT * FROM system_views.accord_migration_state WHERE table_name = ?", tableName);
+        assertTrue(vtableResult.hasNext());
+        Row tableState = vtableResult.next();
+        List<String> vtableMigratedRanges = tableState.getList("migrated_ranges");
+        assertEquals(migratedRanges, vtableMigratedRanges.stream().map(Range::fromString).collect(Collectors.toList()));
```
maedhroz: @aweisberg It looks like this assertion is a bit flakey. I haven't been able to reproduce locally, but a couple times in CI...

```
junit.framework.AssertionFailedError: expected:<[]>  but was:<[(-1,9223372036854775807]]>
	at org.apache.cassandra.distributed.test.accord.AccordMigrationTest.assertVtableMigrationState(AccordMigrationTest.java:715)
	at org.apache.cassandra.distributed.test.accord.AccordMigrationTest.assertMigrationState(AccordMigrationTest.java:633)
	at org.apache.cassandra.distributed.test.accord.AccordMigrationTest.lambda$testPaxosToAccordCAS$13(AccordMigrationTest.java:404)
```

maedhroz: I'm not sure how that range has moved to "migrated", unless I'm looking at the wrong keyspace/table...or there was something I'm waiting for that I shouldn't (as this looks a lot like I'm getting a later state than expected)

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
```diff
@@ -18,57 +18,205 @@
...
+
+            List<AccordStateCache.ImmutableStats> commandsForKeyStats =
+                AsyncChains.getBlockingAndRethrow(stores.map(store -> ((AccordCommandStore) store.commandStore()).commandsForKeyCache().statsSnapshot()));
+            for (int storeID : stores.ids())
+                addRow(commandsForKeyStats.get(storeID), result, storeID, "commands_for_key");
+
+            List<AccordStateCache.ImmutableStats> timestampsForKeyStats =
+                AsyncChains.getBlockingAndRethrow(stores.map(store -> ((AccordCommandStore) store.commandStore()).timestampsForKeyCache().statsSnapshot()));
+            for (int storeID : stores.ids())
+                addRow(timestampsForKeyStats.get(storeID), result, storeID, "timestamps_for_key");
```
dcapwell: nit: why schedule on the same stores 3 times when you could do this once?  this would be 3x faster to do

dcapwell (approved):

src/java/org/apache/cassandra/db/virtual/AccordVirtualTables.java
aweisberg: You know it's done when `migrated_ranges` is the whole ring and should have a single entry. Right now once a migration state has been created for a table it never goes away.

PR #3306, opened 2024-05-13
autorepair v2 framework

jaydeepkumar1984:
- AutoRepir v2 framework that covers both FR & IR 
- An interface with a default implementation for TokenRange calculation
- Enable/Disable flag as a CQL Table property


src/java/org/apache/cassandra/repair/autorepair/AutoRepair.java
```diff
@@ -0,0 +1,401 @@
...
+                    }
+
+                    for (String tableName : tablesToBeRepaired)
+                    {
+                        try
+                        {
+                            ColumnFamilyStore columnFamilyStore = keyspace.getColumnFamilyStore(tableName);
+                            if (columnFamilyStore.metadata().params.disableAutomatedRepair)
+                            {
+                                logger.info("Repair is disabled for keyspace {} for tables: {}", keyspaceName, tableName);
```
kzalys, line 229:
```suggestion
                                logger.info("Repair is disabled for keyspace {} for table: {}", keyspaceName, tableName);
```

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
```diff
@@ -0,0 +1,354 @@
...
+    }
+
+    public int getRepairMinIntervalInHours(RepairType repairType)
+    {
+        return applyOverrides(repairType, opt -> opt.min_repair_interval_in_hours);
+    }
+
+    public void setRepairMinIntervalInHours(RepairType repairType, int repairMinFrequencyInHours)
+    {
+        repair_type_overrides.get(repairType).min_repair_interval_in_hours = repairMinFrequencyInHours;
```
kzalys, line 145: JFYI, there's a bug in these setters. If you provide a config in `cassandra.yaml` such as the following

```
auto_repair:
  repair_type_overrides:
    full:
      enabled: true
```

`repair_type_overrides.get("incremental")`  will be null and we'll get an NPE if we call a setter with incremental repair type

kzalys: Though this only affects nodetool functionality

PR #3307, opened 2024-05-15
Using Capture to save results in CQLSH with Tracing on incorrectly saves the trace details

GautamGottipati:
…ves the trace details

Tried to make initial changes.
Some of the loopholes after these changes :
1) When we enable capture the self.color becomes false (line 1649 in qlshmain.py), due to this the trace is not colored when printed in shell( sysout) 2) There is some issue with Tracing session when tried to print in file. The Tracing session: does n't appear in the first run, it appears in the next runs.

```
Using Capture to save results in CQLSH with Tracing on incorrectly saves the trace details


```

The [Cassandra Jira](https://issues.apache.org/jira/browse/CASSANDRA-19105)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,5 @@
 5.1
+ * Consolidate logging on trace level (CASSANDRA-19632)
  * Expand DDL statements on coordinator before submission to the CMS (CASSANDRA-19592)
  * Fix number of arguments of String.format() in various classes (CASSANDRA-19645)
  * Remove unused fields from config (CASSANDRA-19599)
--- a/src/java/org/apache/cassandra/auth/jmx/AuthorizationProxy.java
+++ b/src/java/org/apache/cassandra/auth/jmx/AuthorizationProxy.java
@@ -191,9 +191,10 @@ public class AuthorizationProxy implements InvocationHandler
     @VisibleForTesting
     public boolean authorize(Subject subject, String methodName, Object[] args)
     {
-        logger.trace("Authorizing JMX method invocation {} for {}",
-                     methodName,
-                     subject == null ? "" :subject.toString().replaceAll("\\n", " "));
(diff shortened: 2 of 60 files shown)


bschoening (changes requested):
1) In tracing.py:

It seems for output we should use **self.shunted_query_out** if its not None, otherwise query_out.  This is similar (but opposite) to how print_result() uses it.

Shouldn't all the shell.writeresults() go to self.shunted_query_out, which is the original stdout?

2) There is no color because its saved in shunted_color.  

3) Maybe if print_formatted_result had a boolean arg for useShunted, it could use the shunted_query_out and shunted_color

3) If possible, there should be a unit test for this.  For example, run CAPTURE "/dev/null" with tracing and make sure the stdout has tracing msgs.  This is primarily so it doesn't get broken in the future.

GautamGottipati:
Hi @bschoening,

Here are some of the observations:
Initially when there is no Capture.
shunted_query_out is `None`
query_out is `<_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>` which is stdout.

When we start Capturing (example `Capture 'demo_capture.txt'`):
shunted_query_out is `<_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>` which is stdout
query_out is `<_io.TextIOWrapper name='demo_capture.txt' mode='a' encoding='UTF-8'>` which is the file demo_capture.txt.

According to the observations 
We are not able to get the file information from shunted_query_out. I feel we need to send query_out to shell.writeresults() .
In Tracing.py the  shell.print_formatted_result should be`shell.print_formatted_result(formatted_names, formatted_values, with_header=True, tty=shell.tty, out=shell.shunted_query_out)` in case `out` is None then output should be `query_out`.

Still figuring it out. Open for suggestions.

PR #3313, opened 2024-05-18
CASSANDRA-19647 implemented try-with-resources for Autocloseable

dk2k:
CASSANDRA-19647 implemented try-with-resources for Autocloseable

I'm not introducing this approach. Scroll down class Ballots to line 210 and you'll see preexistng try-with-resources
As for nullability. The Java Language Specification specifies that Autocloseable is closed only if non-null, in section 14.20.3. try-with-resources: A resource is closed only if it initialized to a non-null value. This can actually be useful, when a resource might present sometimes, and absent others.


dk2k:
Please review

PR #3330, opened 2024-05-23
CASSANDRA-12937 Stefan's squashed

smiklosovic:
 
```

```

### DIFF
--- a/doc/modules/cassandra/pages/developing/cql/cql_singlefile.adoc
+++ b/doc/modules/cassandra/pages/developing/cql/cql_singlefile.adoc
@@ -688,11 +688,10 @@ provided by specifying the full class name as a link:#constants[string
 constant].
 |`enabled` |true |By default compression is enabled. To disable it, set
 `enabled` to `false`
-|`chunk_length_in_kb` |64 |On disk SSTables are
+|`chunk_length_in_kb` |16 |On disk SSTables are
 compressed by block (to allow random reads). This defines the size (in
 KB) of said block. Bigger values may improve the compression rate, but
 increases the minimum size of data to be read from disk for a read
-| `chunk_length` | 64KiB| alias to `chunk_length_in_kb` where size is specified with unit, available from CASSANDRA-12973.
 |===
 
 [[cachingOptions]]
--- a/doc/modules/cassandra/pages/developing/cql/cql_singlefile.adoc
+++ b/doc/modules/cassandra/pages/developing/cql/cql_singlefile.adoc
@@ -678,20 +678,21 @@ to the risk the jvm must also be started with
 
 For the `compression` property, the following sub-options are available:
 
-[cols=",,,,,",options="header",]
+[cols=",,",options="header",]
 |===
-|option |default |description | | |
+|option |default |description
 |`class` |LZ4Compressor |The compression algorithm to use. Default
 compressor are: LZ4Compressor, SnappyCompressor and DeflateCompressor.
 Use `'enabled' : false` to disable compression. Custom compressor can be
 provided by specifying the full class name as a link:#constants[string
-constant]. | | |
-
+constant].
 |`enabled` |true |By default compression is enabled. To disable it, set
-`enabled` to `false` |`chunk_length_in_kb` |64KB |On disk SSTables are
+`enabled` to `false`
+|`chunk_length_in_kb` |64 |On disk SSTables are
 compressed by block (to allow random reads). This defines the size (in
 KB) of said block. Bigger values may improve the compression rate, but
 increases the minimum size of data to be read from disk for a read
+| `chunk_length` | 64KiB| alias to `chunk_length_in_kb` where size is specified with unit, available from CASSANDRA-12973.
 |===
 
 [[cachingOptions]]
--- a/doc/modules/cassandra/pages/developing/cql/ddl.adoc
+++ b/doc/modules/cassandra/pages/developing/cql/ddl.adoc
@@ -628,9 +628,8 @@ Custom compressor can be provided by specifying the full class name as a xref:ca
 | `enabled` | true | Enable/disable sstable compression.
 If the `enabled` option is set to `false`, no other options must be specified.
 
-| `chunk_length_in_kb` | 64 | On disk SSTables are compressed by block (to allow random reads).
+| `chunk_length_in_kb` | 16 | On disk SSTables are compressed by block (to allow random reads).
 This option defines the size (in KB) of said block. See xref:cassandra:developing/cql/ddl.adoc#chunk_note[note] for further information.
-| `chunk_length` | 64KiB| alias to `chunk_length_in_kb` where size is specified with unit, available from CASSANDRA-12973.
 
 | `compression_level` | 3 | Compression level. Only applicable for `ZstdCompressor`.
 Accepts values between `-131072` and `22`.
--- a/doc/modules/cassandra/pages/developing/cql/ddl.adoc
+++ b/doc/modules/cassandra/pages/developing/cql/ddl.adoc
@@ -630,6 +630,7 @@ If the `enabled` option is set to `false`, no other options must be specified.
 
 | `chunk_length_in_kb` | 64 | On disk SSTables are compressed by block (to allow random reads).
 This option defines the size (in KB) of said block. See xref:cassandra:developing/cql/ddl.adoc#chunk_note[note] for further information.
+| `chunk_length` | 64KiB| alias to `chunk_length_in_kb` where size is specified with unit, available from CASSANDRA-12973.
 
 | `compression_level` | 3 | Compression level. Only applicable for `ZstdCompressor`.
 Accepts values between `-131072` and `22`.
--- a/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
+++ b/src/java/org/apache/cassandra/config/DatabaseDescriptor.java
@@ -107,6 +107,7 @@ import org.apache.cassandra.locator.IEndpointSnitch;
 import org.apache.cassandra.locator.InetAddressAndPort;
 import org.apache.cassandra.locator.Replica;
 import org.apache.cassandra.locator.SeedProvider;
+import org.apache.cassandra.schema.CompressionParams;
 import org.apache.cassandra.security.AbstractCryptoProvider;
 import org.apache.cassandra.security.EncryptionContext;
 import org.apache.cassandra.security.JREProvider;
@@ -342,7 +343,7 @@ public class DatabaseDescriptor
 
         applyCompatibilityMode();
 
-        applySSTableFormats();
+        applySSTableConfig();
 
         applySimpleConfig();
 
@@ -397,7 +398,7 @@ public class DatabaseDescriptor
         conf = configSupplier.get();
         applyCompatibilityMode();
         diskOptimizationStrategy = new SpinningDiskOptimizationStrategy();
-        applySSTableFormats();
+        applySSTableConfig();
     }
 
     private static void assertNotDaemonInitialized()
@@ -504,7 +505,7 @@ public class DatabaseDescriptor
     {
         applyCompatibilityMode();
 
-        applySSTableFormats();
+        applySSTableConfig();
 
         applyCryptoProvider();
 
@@ -1630,22 +1631,47 @@ public class DatabaseDescriptor
             storageCompatibilityMode = conf.storage_compatibility_mode;
     }
 
-    private static void applySSTableFormats()
+    private static void applySSTableConfig()
     {
         ServiceLoader<SSTableFormat.Factory> loader = ServiceLoader.load(SSTableFormat.Factory.class, DatabaseDescriptor.class.getClassLoader());
         List<SSTableFormat.Factory> factories = Iterables.toList(loader);
         if (factories.isEmpty())
             factories = ImmutableList.of(new BigFormat.BigFormatFactory());
-        applySSTableFormats(factories, conf.sstable);
+        applySSTableConfig(factories, conf.sstable);
     }
 
-    private static void applySSTableFormats(Iterable<SSTableFormat.Factory> factories, Config.SSTableConfig sstableFormatsConfig)
+    private static void applySSTableConfig(Iterable<SSTableFormat.Factory> factories, Config.SSTableConfig sstableConfig)
     {
+        String defaultCompression = sstableConfig.default_compression;
+        if (defaultCompression != null)
+        {
+            CompressionParams.CompressorType compressorType = CompressionParams.CompressorType.fromName(defaultCompression);
+            try
+            {
+                if (compressorType != null)
+                {
+                    CompressionParams.fromParameterizedClass(new ParameterizedClass(compressorType.name(),
+                                                                                    sstableConfig.compression.getOrDefault(compressorType.name(),
+                                                                                                                           new HashMap<>())));
+                }
+                else
+                {
+                    CompressionParams.fromParameterizedClass(new ParameterizedClass(defaultCompression,
+                                                                                    sstableConfig.compression.getOrDefault(defaultCompression,
+                                                                                                                           new HashMap<>())));
+                }
+            }
+            catch (ConfigurationException ex)
+            {
+                throw new ConfigurationException(String.format("Invalid configuration of sstable default compression: %s", ex.getMessage()));
+            }
+        }
+
         if (sstableFormats != null)
             return;
 
         validateSSTableFormatFactories(factories);
-        ImmutableMap<String, Supplier<SSTableFormat<?, ?>>> providers = validateAndMatchSSTableFormatOptions(factories, sstableFormatsConfig.format);
+        ImmutableMap<String, Supplier<SSTableFormat<?, ?>>> providers = validateAndMatchSSTableFormatOptions(factories, sstableConfig.format);
 
         ImmutableMap.Builder<String, SSTableFormat<?, ?>> sstableFormatsBuilder = ImmutableMap.builder();
         providers.forEach((name, provider) -> {
@@ -1660,7 +1686,7 @@ public class DatabaseDescriptor
         });
         sstableFormats = sstableFormatsBuilder.build();
 
-        selectedSSTableFormat = getAndValidateWriteFormat(sstableFormats, sstableFormatsConfig.selected_format);
+        selectedSSTableFormat = getAndValidateWriteFormat(sstableFormats, sstableConfig.selected_format);
 
         sstableFormats.values().forEach(SSTableFormat::allComponents); // make sure to reach all supported components for a type so that we know all of them are registered
         logger.info("Supported sstable formats are: {}", sstableFormats.values().stream().map(f -> f.name() + " -> " + f.getClass().getName() + " with singleton components: " + f.allComponents()).collect(Collectors.joining(", ")));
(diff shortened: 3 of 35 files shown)


Claudenw (changes requested):
LGTM except that there is a change to the CQL parameters and I don't see any user documentation for it.

smiklosovic:
@Claudenw what exact parameters do you mean? `chunk_length` ?

test/unit/org/apache/cassandra/cql3/validation/operations/AlterTest.java
```diff
@@ -697,10 +698,10 @@ public void testAlterTableWithCompression() throws Throwable
         createTable("CREATE TABLE %s (a text, b int, c int, primary key (a, b))");
         assertSchemaOption("compression", map("chunk_length_in_kb", "16", "class", "org.apache.cassandra.io.compress.LZ4Compressor"));
 
-        alterTable("ALTER TABLE %s WITH compression = { 'class' : 'SnappyCompressor', 'chunk_length_in_kb' : 32 };");
+        alterTable("ALTER TABLE %s WITH compression = { 'class' : 'SnappyCompressor', 'chunk_length' : '32KiB' };");
         assertSchemaOption("compression", map("chunk_length_in_kb", "32", "class", "org.apache.cassandra.io.compress.SnappyCompressor"));
 
-        alterTable("ALTER TABLE %s WITH compression = { 'class' : 'LZ4Compressor', 'chunk_length_in_kb' : 64 };");
+        alterTable("ALTER TABLE %s WITH compression = { 'class' : 'LZ4Compressor', 'chunk_length' : '64KiB' };");
```
Claudenw: @smiklosovic  Here is an example of `chunk_length` as in CQL. It looks like you can use "chunk_length" : 35KiB much the same as you can use chunk_length_in_kb : 35 but chunk_length is not documented as a CQL option in the user documentation.

Claudenw (changes requested):

test/unit/org/apache/cassandra/cql3/validation/operations/AlterTest.java
smiklosovic: I added it.

doc/modules/cassandra/pages/developing/cql/cql_singlefile.adoc
```diff
@@ -678,20 +678,21 @@ to the risk the jvm must also be started with
...
 compressor are: LZ4Compressor, SnappyCompressor and DeflateCompressor.
 Use `'enabled' : false` to disable compression. Custom compressor can be
 provided by specifying the full class name as a link:#constants[string
-constant]. | | |
-
+constant].
 |`enabled` |true |By default compression is enabled. To disable it, set
-`enabled` to `false` |`chunk_length_in_kb` |64KB |On disk SSTables are
+`enabled` to `false`
+|`chunk_length_in_kb` |64 |On disk SSTables are
```
Mmuzaf: The default value is 16, right? According to the `DEFAULT_CHUNK_LENGTH` constant and `cassandra.yaml`. Why 64?

doc/modules/cassandra/pages/developing/cql/ddl.adoc
```diff
@@ -630,6 +630,7 @@ If the `enabled` option is set to `false`, no other options must be specified.
 
 | `chunk_length_in_kb` | 64 | On disk SSTables are compressed by block (to allow random reads).
 This option defines the size (in KB) of said block. See xref:cassandra:developing/cql/ddl.adoc#chunk_note[note] for further information.
+| `chunk_length` | 64KiB| alias to `chunk_length_in_kb` where size is specified with unit, available from CASSANDRA-12973.
```
Mmuzaf: I don't think it's a good idea to add an alias for a configuration parameter, it would lead us eventually to a messy configuration and backwards compatibility maintenance problems. 

We already did some housekeeping here:
https://issues.apache.org/jira/browse/CASSANDRA-18742

I would say we have a few options:
- don't add `chunk_length` at all and use only `chunk_length_in_kb`;
- create another issue and deprecate the `chunk_length_in_kb` in favor of `chunk_length`

The issue we are working on is - "Move defaults to yaml" we shouldn't mix it with refactoring/deprecation:
https://issues.apache.org/jira/browse/CASSANDRA-12937

doc/modules/cassandra/pages/developing/cql/cql_singlefile.adoc
Mmuzaf: Yes, the defaults has changed:
https://issues.apache.org/jira/browse/CASSANDRA-13241

doc/modules/cassandra/pages/developing/cql/ddl.adoc
smiklosovic: @Mmuzaf ok I removed chunk_length, I do not have any strong feelings about that. If we want to introduce it and deprecate chunk_length_in_kb, we might just dedicate a separate ticket for that and do it as a standalone change as you suggested.

test/unit/org/apache/cassandra/schema/TestCompressionParamsFactory.java
```diff
@@ -0,0 +1,170 @@
...
+
+import static java.util.Collections.emptyMap;
+import static org.apache.cassandra.schema.CompressionParams.DEFAULT_CHUNK_LENGTH;
+import static org.apache.cassandra.schema.CompressionParams.DEFAULT_MIN_COMPRESS_RATIO;
+
+/**
+ * Contains simple constructors for various Compression implementations.
+ * They are a little inconsistent in their choice of parameters -- this is done on purpose to test out various compression parameter combinations.
+ */
+public class TestCompressionParamsFactory
```
Mmuzaf, line 37: I'm going to be a bit of a nerd, but this class doesn't seem to have anything to do with the changes we're going to be making. 
Concerns are the following:
- Looks like refactoring doesn't need to be done here, as the changes are not directly related to the issue
-the `CompressionParams` and default constructors might be used for dependent projects, so moving these methods to a test class might break the compilation there;
- Renaming to the `TestCompressionParamsFactory` generates a lot of code changes;

So I would suggest rolling back those changes for now and creating another LHF issue to do that.

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -252,6 +254,8 @@ public class DatabaseDescriptor
                                                                                                            ? new CommitLogSegmentManagerCDC(c, DatabaseDescriptor.getCommitLogLocation())
                                                                                                            : new CommitLogSegmentManagerStandard(c, DatabaseDescriptor.getCommitLogLocation());
 
+    private static Map<String, ParameterizedClass> defaultSSTableCompressionConfigs = new HashMap<>();
```
Mmuzaf, line 257: As far as I understand, if the configuration is not set in the yaml file we'll initialize this property with a default configuration and leave `Config.SSTableConfig` empty.

When we traverse the `Config` class to expose the running configuration to a user e.g. via the settings virtual table it won't see any compression configuration that has been initialized by default, and this is a bad thing. The Config class must be always consistent in terms of which configuration we are running. 

This is an example of the issue we previously had to deal with:
https://issues.apache.org/jira/browse/CASSANDRA-17738

doc/modules/cassandra/pages/developing/cql/cql_singlefile.adoc
```diff
@@ -678,17 +678,17 @@ to the risk the jvm must also be started with
 
 For the `compression` property, the following sub-options are available:
 
-[cols=",,,,,",options="header",]
+[cols=",,",options="header",]
 |===
-|option |default |description | | |
+|option |default |description
```
Mmuzaf, line 683: Should we also update these docs with all the compression classes we support?

doc/modules/cassandra/pages/developing/cql/cql_singlefile.adoc
```diff
@@ -678,17 +678,17 @@ to the risk the jvm must also be started with
...
 compressor are: LZ4Compressor, SnappyCompressor and DeflateCompressor.
 Use `'enabled' : false` to disable compression. Custom compressor can be
 provided by specifying the full class name as a link:#constants[string
-constant]. | | |
-
+constant].
 |`enabled` |true |By default compression is enabled. To disable it, set
-`enabled` to `false` |`chunk_length_in_kb` |64KB |On disk SSTables are
+`enabled` to `false`
+|`chunk_length_in_kb` |16 |On disk SSTables are
```
Mmuzaf, line 691: It looks like we can update the default in the `compress-subproperties.adoc` as well. 

https://github.com/instaclustr/cassandra/blob/CASSANDRA-12937-squashed/doc/modules/cassandra/partials/compress-subproperties.adoc?plain=1#L59

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -20,184 +20,355 @@
...
+     * @return The compresson parameters for the keyspace.
+     */
+    public static CompressionParams defaultParams(String keyspace)
+    {
+        if (keyspace != null && SchemaConstants.getLocalAndReplicatedSystemKeyspaceNames().contains(keyspace))
+            return DEFAULT;
 
-        int chunkLength = removeChunkLength(options);
-        double minCompressRatio = removeMinCompressRatio(options);
+        CompressionParams result = CALCULATED_DEFAULT;
```
Mmuzaf, line 132: Saving the calculated default from the config doesn't make sense to me either. From my point of view the process should look like this:

1. We parse the yaml file
2. If the yaml contains compression params, _then_ set them to the `Config.SSTableConfig`
3. If the yaml doesn't have compression params, _then_ initialize the `Config.SSTableConfig` with the defaults from the `CompressionParams DEFAULT`.
4. When we want create a `CompressonParams` for a keyspace e.g. by calling `defaultParams(String keyspace)` we always read the configuration stored in `Config.SSTableConfig` and convert it to CompressonParams inside the `defaultParams` method.

It seems that such a technique could remove a lot of null-checks in this method and all calculated defaults stored out of the Config class. It also seems that will simplify the code a lot. WDYT?

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -20,184 +20,355 @@
...
+        if (result == null)
+        {
+            if (DatabaseDescriptor.getRawConfig() == null)
```
Mmuzaf, line 135: Should we throw an exception in this case if the config is not initialized when this method is called?

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -20,184 +20,355 @@
...
+            else
+                result = CALCULATED_DEFAULT = fromParameterizedClass(defaultSSTableCompression);
+        }
 
-        return cp;
+        return result;
     }
 
-    public Class<? extends ICompressor> klass()
+    public static CompressionParams fromParameterizedClass(ParameterizedClass options)
```
Mmuzaf, line 151: I might be missing something, but this method probably should have the following signature:
`public static CompressionParams fromParameterizedClass(SSTableCompressionConfig config)`

src/java/org/apache/cassandra/config/DatabaseDescriptor.java
```diff
@@ -1630,22 +1634,44 @@ else if (conf != null && conf.storage_compatibility_mode != null)
...
+        applySSTableConfig(factories, conf.sstable);
     }
 
-    private static void applySSTableFormats(Iterable<SSTableFormat.Factory> factories, Config.SSTableConfig sstableFormatsConfig)
+    private static void applySSTableConfig(Iterable<SSTableFormat.Factory> factories, Config.SSTableConfig sstableConfig)
     {
+        InheritingClass defaultParameterizedClass = new InheritingClass(null, CompressionParams.DEFAULT.klass().getName(), CompressionParams.DEFAULT.getOtherOptions());
+
+        if (sstableConfig.compression == null || sstableConfig.compression.configurations == null)
+            setDefaultSSTableCompressionConfigs(ImmutableMap.of(InheritingClass.DEFAULT_CONFIGURATION_KEY, defaultParameterizedClass));
```
Mmuzaf, line 1651: The `applyConfig*` methods from my point of view mean that we apply the defaults to the Config class instance itself, and the use this initialized defaults, specifically the `SSTableCompressionConfig` throughout the code. I don't think we should store the intermediate calculations in local variables or anywhere outside of the Config.

test/conf/cassandra-with-sstable-compressor.yaml
```diff
@@ -0,0 +1,113 @@
...
+            parameters:
+                invalid: throw
+        test_invalid_factory_method:
+            class_name: org.apache.cassandra.cql3.validation.operations.CreateTest$InvalidMemtableFactoryMethod
+        test_invalid_factory_field:
+            class_name: org.apache.cassandra.cql3.validation.operations.CreateTest$InvalidMemtableFactoryField
+
+sstable:
+    compression:
+        configurations:
```
Mmuzaf, line 106: This is probably not related to this issue and we should create a follow-up issue for the following question: Should we disallow the case where we list customized compression configurations and yet have no 'default' configuration?

test/distributed/org/apache/cassandra/distributed/test/SSTableCompressionTest.java
```diff
@@ -0,0 +1,465 @@
...
+import static org.apache.cassandra.schema.SchemaConstants.LOCAL_SYSTEM_KEYSPACE_NAMES;
+import static org.apache.cassandra.schema.SchemaConstants.REPLICATED_SYSTEM_KEYSPACE_NAMES;
+import static org.apache.commons.io.FileUtils.readFileToString;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertTrue;
+
+public class SSTableCompressionTest
```
Mmuzaf, line 68: We should also add some tests for the `system_views.settings' table, which will return the current configuration used after initializing a node, e.g. the default compression configuration if the config is omitted in the yaml file, and the custom configuration read from the yaml.

The following query currently returns `null` config for the default config:
`SELECT * FROM vts.settings WHERE name = 'sstable.compression.configurations`

test/distributed/org/apache/cassandra/distributed/test/SSTableCompressionTest.java
```diff
@@ -0,0 +1,465 @@
...
+            cluster = null;
+        }
+    }
+
+    @Test
+    public void testCreation() throws IOException
+    {
+        setupCluster(DEFAULT_CONFIG, null);
+        cluster.schemaChange(format("CREATE KEYSPACE IF NOT EXISTS %s WITH replication = {'class': 'SimpleStrategy', 'replication_factor': '1'}", KEYSPACE));
+        testDisabled();
```
Mmuzaf, line 156: Minor: These methods are used only once, it's easier and shorter to inline them just to avoid falling into the methods to figure out what exactly they are testing.

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -20,184 +20,355 @@
...
+                                                                          DEFAULT_MIN_COMPRESS_RATIO,
+                                                                          emptyMap());
 
-    public static CompressionParams fromMap(Map<String, String> opts)
-    {
-        Map<String, String> options = copyOptions(opts);
+    /**
+     * A guaranteed FAST compressor
+     */
+    public static final CompressionParams FAST = DEFAULT;
```
Mmuzaf, line 91: I think we should either remove the `FAST` and use the `DEFAULT` everywhere or rename it to `LZ4_FAST_COMPRESSOR` to make it more clear.

src/java/org/apache/cassandra/io/sstable/format/DataComponent.java
```diff
@@ -86,12 +86,17 @@ private static CompressionParams buildCompressionParams(TableMetadata metadata,
...
+                    ICompressor maybeFastCompressor = defaultCompressionParams.getSstableCompressor();
+                    if (maybeFastCompressor.recommendedUses().contains(ICompressor.Uses.FAST_COMPRESSION)) {
+                        compressionParams = defaultCompressionParams;
+                    }
+                    else
                     {
-                        // The default compressor is generally fast (LZ4 with 16KiB block size)
-                        compressionParams = CompressionParams.DEFAULT;
-                        break;
+                        compressionParams = CompressionParams.FAST;
```
Mmuzaf, line 97: Please add a test to the `CompressionParamsTest` that checks if the `CompressionParams.DEFAULT` uses the `LZ4Compressor` "fast" by default or an assertion here. The current assumption that it's "fast" is fragile. The latter is better.

src/java/org/apache/cassandra/io/sstable/format/DataComponent.java
```diff
@@ -86,12 +86,17 @@ private static CompressionParams buildCompressionParams(TableMetadata metadata,
                     compressionParams = CompressionParams.NOOP;
                     break;
                 case fast:
-                    if (!compressor.recommendedUses().contains(ICompressor.Uses.FAST_COMPRESSION))
+                    // The default compressor is generally Fast, but just in case we verify it.
+                    CompressionParams defaultCompressionParams = CompressionParams.defaultParams(metadata.keyspace);
+                    ICompressor maybeFastCompressor = defaultCompressionParams.getSstableCompressor();
+                    if (maybeFastCompressor.recommendedUses().contains(ICompressor.Uses.FAST_COMPRESSION)) {
```
Mmuzaf, line 92: I would fix this logic. For the system keyspaces, we should not change the default compression at all as the defaults might change, regardless of whether it is "fast" for the system keyspaces or not.

src/java/org/apache/cassandra/schema/CompressionParams.java
```diff
@@ -20,184 +20,355 @@
...
+    /**
+     * Gets the default compression params for the keyspace.
+     * This method accounts for issues when the compression params change for system keyspaces.
+     *
+     * @param keyspace the name of the keyspace to get params for. (may be null for non-System keyspaces)
+     * @return The compresson parameters for the keyspace.
+     */
+    public static CompressionParams defaultParams(String keyspace)
+    {
+        if (keyspace != null && SchemaConstants.getLocalAndReplicatedSystemKeyspaceNames().contains(keyspace))
```
Mmuzaf, line 129: If `keyspace == null` why should we return the `DEFAULT` instead of the defaults, that were set in the yaml?

test/unit/org/apache/cassandra/config/DatabaseDescriptorRefTest.java
```diff
@@ -286,7 +286,14 @@ public class DatabaseDescriptorRefTest
     "org.apache.cassandra.utils.concurrent.SelfRefCounted",
     "org.apache.cassandra.utils.concurrent.Transactional",
     "org.apache.cassandra.utils.concurrent.UncheckedInterruptedException",
-    "org.apache.cassandra.utils.StorageCompatibilityMode"
+    "org.apache.cassandra.utils.StorageCompatibilityMode",
```
Mmuzaf, line 289: Please, sort the list `validClasses`

src/java/org/apache/cassandra/io/sstable/format/DataComponent.java
Mmuzaf: Found the following test, I think assertion would be better. 

`org.apache.cassandra.io.compress.CQLCompressionTest#lz4FlushTest`

Mmuzaf (commented):
@smiklosovic Hi, I've left some comments.

PR #3340, opened 2024-05-29
CASSANDRA-19669: Audit Log entries are missing identity for mTLS connections

frankgh:
This patch adds the missing metadata from the authenticated user, which includes the identity for mTLS connections. Additionally, it notifies authentication failures during TLS handshake, for example when a client is attempting to authenticate with an expired certificate. This now will be recorded as a failed authentication attempt.

Patch by Francisco Guerrero; Reviewed by TBD for CASSANDRA-19669

### DIFF
--- a/src/java/org/apache/cassandra/audit/AuditLogEntry.java
+++ b/src/java/org/apache/cassandra/audit/AuditLogEntry.java
@@ -242,7 +242,7 @@ public class AuditLogEntry
             operation = entry.operation;
             options = entry.options;
             state = entry.state;
-            metadata = entry.metadata;
+            metadata = Map.copyOf(entry.metadata);
         }
 
         public Builder setType(AuditLogEntryType type)
--- a/src/java/org/apache/cassandra/audit/AuditLogEntry.java
+++ b/src/java/org/apache/cassandra/audit/AuditLogEntry.java
@@ -19,9 +19,11 @@ package org.apache.cassandra.audit;
 
 import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
+import java.util.Map;
 import java.util.UUID;
 import javax.annotation.Nullable;
 
+import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.lang3.StringUtils;
 
 import org.apache.cassandra.auth.AuthenticatedUser;
@@ -47,39 +49,37 @@ public class AuditLogEntry
     private final String operation;
     private final QueryOptions options;
     private final QueryState state;
+    private final Map<String, Object> metadata;
 
-    private AuditLogEntry(AuditLogEntryType type,
-                          InetAddressAndPort source,
-                          String user,
-                          long timestamp,
-                          UUID batch,
-                          String keyspace,
-                          String scope,
-                          String operation,
-                          QueryOptions options,
-                          QueryState state)
+    private AuditLogEntry(Builder builder)
     {
-        this.type = type;
-        this.source = source;
-        this.user = user;
-        this.timestamp = timestamp;
-        this.batch = batch;
-        this.keyspace = keyspace;
-        this.scope = scope;
-        this.operation = operation;
-        this.options = options;
-        this.state = state;
+        this.type = builder.type;
+        this.source = builder.source;
+        this.user = builder.user;
+        this.timestamp = builder.timestamp;
+        this.batch = builder.batch;
+        this.keyspace = builder.keyspace;
+        this.scope = builder.scope;
+        this.operation = builder.operation;
+        this.options = builder.options;
+        this.state = builder.state;
+        this.metadata = builder.metadata;
     }
 
-    String getLogString()
+    @VisibleForTesting
+    public String getLogString()
     {
         StringBuilder builder = new StringBuilder(100);
         builder.append("user:").append(user)
-               .append("|host:").append(host)
-               .append("|source:").append(source.getAddress());
-        if (source.getPort() > 0)
+               .append("|host:").append(host);
+
+        if (source != null)
         {
-            builder.append("|port:").append(source.getPort());
+            builder.append("|source:").append(source.getAddress());
+            if (source.getPort() > 0)
+            {
+                builder.append("|port:").append(source.getPort());
+            }
         }
 
         builder.append("|timestamp:").append(timestamp)
@@ -101,6 +101,10 @@ public class AuditLogEntry
         if (StringUtils.isNotBlank(operation))
         {
             builder.append("|operation:").append(operation);
+        }
+        if (metadata != null && !metadata.isEmpty())
+        {
+            metadata.forEach((key, value) -> builder.append('|').append(key).append(':').append(value));
         }
         return builder.toString();
     }
@@ -189,6 +193,7 @@ public class AuditLogEntry
         private String operation;
         private QueryOptions options;
         private QueryState state;
+        private Map<String, Object> metadata;
 
         public Builder(QueryState queryState)
         {
@@ -204,9 +209,15 @@ public class AuditLogEntry
                     source = InetAddressAndPort.getByAddressOverrideDefaults(addr.getAddress(), addr.getPort());
                 }
 
-                if (clientState.getUser() != null)
+                AuthenticatedUser authenticatedUser = clientState.getUser();
+                if (authenticatedUser != null)
                 {
-                    user = clientState.getUser().getName();
+                    user = authenticatedUser.getName();
+
+                    if (authenticatedUser.getMetadata() != null)
+                    {
+                        metadata = Map.copyOf(authenticatedUser.getMetadata());
+                    }
                 }
                 keyspace = clientState.getRawKeyspace();
             }
@@ -231,6 +242,7 @@ public class AuditLogEntry
             operation = entry.operation;
             options = entry.options;
             state = entry.state;
+            metadata = entry.metadata;
         }
 
         public Builder setType(AuditLogEntryType type)
@@ -312,10 +324,16 @@ public class AuditLogEntry
             return this;
         }
 
+        public Builder setMetadata(Map<String, Object> metadata)
+        {
+            this.metadata = Map.copyOf(metadata);
+            return this;
+        }
+
         public AuditLogEntry build()
         {
             timestamp = timestamp > 0 ? timestamp : currentTimeMillis();
-            return new AuditLogEntry(type, source, user, timestamp, batch, keyspace, scope, operation, options, state);
+            return new AuditLogEntry(this);
         }
     }
 }
--- a/src/java/org/apache/cassandra/audit/AuditLogEntry.java
+++ b/src/java/org/apache/cassandra/audit/AuditLogEntry.java
@@ -242,7 +242,7 @@ public class AuditLogEntry
             operation = entry.operation;
             options = entry.options;
             state = entry.state;
-            metadata = Map.copyOf(entry.metadata);
+            metadata = entry.metadata != null ? Map.copyOf(entry.metadata) : null;
         }
 
         public Builder setType(AuditLogEntryType type)
@@ -326,7 +326,7 @@ public class AuditLogEntry
 
         public Builder setMetadata(Map<String, Object> metadata)
         {
-            this.metadata = Map.copyOf(metadata);
+            this.metadata = metadata != null ? Map.copyOf(metadata) : null;
             return this;
         }
 
--- a/src/java/org/apache/cassandra/audit/AuditLogManager.java
+++ b/src/java/org/apache/cassandra/audit/AuditLogManager.java
@@ -23,13 +23,11 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.UUID;
-
 import javax.annotation.Nullable;
 import javax.management.openmbean.CompositeData;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -79,7 +77,7 @@ public class AuditLogManager implements QueryEvents.Listener, AuthEvents.Listene
         }
         else
         {
-            logger.debug("Audit logging is disabled.");
+            logger.info("Audit logging is disabled.");
             auditLogger = new NoOpAuditLogger(Collections.emptyMap());
         }
 
--- a/src/java/org/apache/cassandra/transport/PreV5Handlers.java
+++ b/src/java/org/apache/cassandra/transport/PreV5Handlers.java
@@ -18,9 +18,13 @@
 
 package org.apache.cassandra.transport;
 
+import java.net.SocketAddress;
+import java.security.cert.CertificateException;
 import java.util.List;
+import javax.net.ssl.SSLException;
 
 import com.google.common.base.Predicate;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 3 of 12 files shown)


src/java/org/apache/cassandra/audit/AuditLogEntry.java
```diff
@@ -231,6 +242,7 @@ public Builder(AuditLogEntry entry)
             operation = entry.operation;
             options = entry.options;
             state = entry.state;
+            metadata = entry.metadata;
```
bbotella: Should we use the setMetadata method that is defined below as part of this PR?

src/java/org/apache/cassandra/audit/AuditLogManager.java
```diff
@@ -79,7 +77,7 @@ private AuditLogManager()
         }
         else
         {
-            logger.debug("Audit logging is disabled.");
+            logger.info("Audit logging is disabled.");
```
bbotella: Why the change?

frankgh: From an auditing point of view I think it's important to know whether audit logging is enabled or disabled when you start the instance.

src/java/org/apache/cassandra/audit/AuditLogEntry.java
frankgh: Just making this consistent with the rest of the code block. But I think we may want to copy the map. Let me address that

bbotella:
LGTM

src/java/org/apache/cassandra/audit/AuditLogEntry.java
```diff
@@ -47,39 +49,37 @@ public class AuditLogEntry
...
+    public String getLogString()
     {
         StringBuilder builder = new StringBuilder(100);
         builder.append("user:").append(user)
-               .append("|host:").append(host)
-               .append("|source:").append(source.getAddress());
-        if (source.getPort() > 0)
+               .append("|host:").append(host);
+
+        if (source != null)
```
tolbertam: just out of curiosity, why would source be null?

src/java/org/apache/cassandra/audit/AuditLogEntry.java
```diff
@@ -204,9 +209,15 @@ public Builder(QueryState queryState)
...
-                if (clientState.getUser() != null)
+                AuthenticatedUser authenticatedUser = clientState.getUser();
+                if (authenticatedUser != null)
                 {
-                    user = clientState.getUser().getName();
+                    user = authenticatedUser.getName();
+
+                    if (authenticatedUser.getMetadata() != null)
+                    {
+                        metadata = Map.copyOf(authenticatedUser.getMetadata());
```
tolbertam: probably just picking nits, but why copy the map?

src/java/org/apache/cassandra/audit/AuditLogEntry.java
```diff
@@ -312,10 +324,16 @@ public Builder setOptions(QueryOptions options)
...
+        {
+            this.metadata = Map.copyOf(metadata);
+            return this;
+        }
+
         public AuditLogEntry build()
         {
             timestamp = timestamp > 0 ? timestamp : currentTimeMillis();
-            return new AuditLogEntry(type, source, user, timestamp, batch, keyspace, scope, operation, options, state);
+            return new AuditLogEntry(this);
```
tolbertam: :+1: i'm a fan of using Builders in constructors, makes it less difficult to pass the wrong fields in the wrong position.

src/java/org/apache/cassandra/transport/PreV5Handlers.java
```diff
@@ -318,13 +325,22 @@ public void exceptionCaught(final ChannelHandlerContext ctx, Throwable cause)
                 if (isFatal(cause))
                     future.addListener((ChannelFutureListener) f -> ctx.close());
             }
-            
-            if (DatabaseDescriptor.getClientErrorReportingExclusions().contains(ctx.channel().remoteAddress()))
+
+            SocketAddress remoteAddress = ctx.channel().remoteAddress();
+            AuthenticationException authenticationException = maybeExtractAndWrapAuthenticationException(cause);
```
tolbertam: We'll need similar code in `ExceptionHandlers.PostV5ExceptionHandler` as well too right?

src/java/org/apache/cassandra/transport/PreV5Handlers.java
```diff
@@ -318,13 +325,22 @@ public void exceptionCaught(final ChannelHandlerContext ctx, Throwable cause)
                 if (isFatal(cause))
                     future.addListener((ChannelFutureListener) f -> ctx.close());
             }
-            
-            if (DatabaseDescriptor.getClientErrorReportingExclusions().contains(ctx.channel().remoteAddress()))
+
```
tolbertam: Far out of scope of this change, but I can imagine this being useful for rejected internode connections as well; but I believe audit logging only applies to user queries over the native protocol, so it probably doesn't fit.

src/java/org/apache/cassandra/transport/PreV5Handlers.java
```diff
@@ -318,13 +325,22 @@ public void exceptionCaught(final ChannelHandlerContext ctx, Throwable cause)
...
+            SocketAddress remoteAddress = ctx.channel().remoteAddress();
+            AuthenticationException authenticationException = maybeExtractAndWrapAuthenticationException(cause);
+            if (authenticationException != null)
+            {
+                QueryState queryState = new QueryState(ClientState.forExternalCalls(remoteAddress));
+                AuthEvents.instance.notifyAuthFailure(queryState, authenticationException);
+            }
+
+            if (remoteAddress != null && DatabaseDescriptor.getClientErrorReportingExclusions().contains(remoteAddress))
```
tolbertam: When might remoteAddress be null?  Guessing you ran into this?

test/distributed/org/apache/cassandra/distributed/test/auth/AuthAuditLoggingTest.java
```diff
@@ -0,0 +1,391 @@
...
+import org.apache.cassandra.distributed.test.JavaDriverUtils;
+import org.apache.cassandra.distributed.test.TestBaseImpl;
+import org.apache.cassandra.utils.tls.CertificateBuilder;
+import org.apache.cassandra.utils.tls.CertificateBundle;
+
+import static org.apache.cassandra.audit.AuditLogEntryType.LOGIN_ERROR;
+import static org.apache.cassandra.audit.AuditLogEntryType.LOGIN_SUCCESS;
+import static org.apache.cassandra.auth.CassandraRoleManager.DEFAULT_SUPERUSER_NAME;
+import static org.apache.cassandra.auth.CassandraRoleManager.DEFAULT_SUPERUSER_PASSWORD;
+import static org.apache.cassandra.distributed.test.auth.MutualTlsCertificateValidityPeriodTest.configureIdentity;
```
tolbertam: Would suggest moving these methods to a shared utils

test/distributed/org/apache/cassandra/distributed/test/auth/AuthAuditLoggingTest.java
```diff
@@ -0,0 +1,391 @@
...
+import static org.apache.cassandra.distributed.test.auth.MutualTlsCertificateValidityPeriodTest.withAuthenticatedSession;
+import static org.apache.cassandra.transport.TlsTestUtils.SERVER_KEYSTORE_PASSWORD;
+import static org.apache.cassandra.transport.TlsTestUtils.SERVER_TRUSTSTORE_PASSWORD;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
+/**
+ * Tests authentication audit logging events
+ */
+public class AuthAuditLoggingTest extends TestBaseImpl
```
tolbertam: It looks like there is already an `AuditLoggerAuthTest` and `AuditLoggerTest`, and there are also tests in `CQLUserAuditTest` around authing with invalid passwords.  Wondering if these can be merged somehow or maybe rename this in a way that it makes it clear how they are differentiated?  I think the key difference here is that this configures the fallback authenticator and exercises MTLS authentication as well.

tolbertam: ah - this also uses the dtest framework which is a big differentiator with the other test I suppose!

tolbertam (commented):
review still in flight, just submitting now to surface my comments earlier.

test/distributed/org/apache/cassandra/distributed/test/auth/AuthAuditLoggingTest.java
```diff
@@ -0,0 +1,391 @@
...
+                assertThat(entry2.getLogString()).matches(expectedLogStringRegex);
+            });
+        }
+    }
+
+    @Test
+    public void testMutualTlsAuthenticationFailedWithUntrustedCertificate() throws Exception
+    {
+        configureMutualTlsAuthenticator();
+        CharSequence expectedLogStringRegex = "^user:null\\|host:.*/127.0.0.1:\\d+\\|timestamp:\\d+" +
```
tolbertam: This fails for me locally because source is present for me (also on `testMutualTlsAuthenticationFailedWithExpiredCertificate`):

```
java.lang.AssertionError: 
Expecting actual:
  "user:null|host:localhost/127.0.0.1:42421|source:/127.0.0.1|port:49366|timestamp:1717027629823|type:LOGIN_ERROR|category:AUTH|operation:LOGIN FAILURE; Empty client certificate chain"
to match pattern:
  "^user:null\|host:.*/127.0.0.1:\d+\|timestamp:\d+\|type:LOGIN_ERROR\|category:AUTH\|operation:LOGIN FAILURE; Empty client certificate chain.*$"
```

test/distributed/org/apache/cassandra/distributed/test/auth/AuthAuditLoggingTest.java
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    @Test
+    public void testMutualTlsAuthenticationFailedWithInvalidSpiffeCertificate() throws Exception
+    {
+        CharSequence expectedLogStringRegex = "^user:null\\|host:.*/127.0.0.1:\\d+\\|source:/127.0.0.1" +
+                                              "\\|port:\\d+\\|timestamp:\\d+\\|type:LOGIN_ERROR\\|category:AUTH" +
+                                              "\\|operation:LOGIN FAILURE; Unable to extract Spiffe from the certificate.*$";
+
+        Path expiredClientKeystorePath = generateClientCertificate(b -> b.clearSubjectAlternativeNames()
```
tolbertam: nit: Should rename variable to `invalidSpiffeClientKeyStorePath`

test/distributed/org/apache/cassandra/distributed/test/auth/AuthAuditLoggingTest.java
```diff
@@ -0,0 +1,391 @@
...
+    }
+
+    @Test
+    public void testMutualTlsAuthenticationFailedWithIdentityThatDoesNotMapToARole() throws Exception
+    {
+        CharSequence expectedLogStringRegex = "^user:null\\|host:.*/127.0.0.1:\\d+\\|source:/127.0.0.1" +
+                                              "\\|port:\\d+\\|timestamp:\\d+\\|type:LOGIN_ERROR\\|category:AUTH" +
+                                              "\\|operation:LOGIN FAILURE; Certificate identity 'spiffe://test.cassandra.apache.org/dTest/notMapped' not authorized.*$";
+
+        Path expiredClientKeystorePath = generateClientCertificate(b -> b.clearSubjectAlternativeNames()
```
tolbertam: nit: Should rename variable to `nonMappedIdentityClientKeyStorePath`

test/distributed/org/apache/cassandra/distributed/test/auth/AuthAuditLoggingTest.java
```diff
@@ -0,0 +1,391 @@
...
+    @Test
+    public void testMutualTlsAuthenticationFailedWithUntrustedCertificate() throws Exception
+    {
+        configureMutualTlsAuthenticator();
+        CharSequence expectedLogStringRegex = "^user:null\\|host:.*/127.0.0.1:\\d+\\|timestamp:\\d+" +
+                                              "\\|type:LOGIN_ERROR\\|category:AUTH" +
+                                              "\\|operation:LOGIN FAILURE; Empty client certificate chain.*$";
+        Path clientKeystorePath = generateSelfSignedCertificate(null, tempFolder.getRoot());
+
+        try (com.datastax.driver.core.Cluster c = JavaDriverUtils.create(CLUSTER, null, b -> b.withSSL(getSSLOptions(clientKeystorePath, truststorePath)));
```
tolbertam: nit: Noticed a pattern in the tests that fail auth is that the code is the same except for the message passed to `fail` and the `expectedLogStringRegex` , you could probably refactor into a common method that these 4 methods call.

src/java/org/apache/cassandra/transport/PreV5Handlers.java
frankgh: yeah, I actually saw remote address being null in the in-jvm dtest. It was causing an NPE coming from the `contains` method.

src/java/org/apache/cassandra/audit/AuditLogEntry.java
frankgh: It was `null` here: https://github.com/apache/cassandra/pull/3340/files#diff-bd4e9e38614c46ae498bef5d2a344766cd14f21e0272961a9e26fe549b6af061R329 . Basically, the channel's remote address was `null` in the `ExceptionHandler`. Maybe that's a problem in itself

src/java/org/apache/cassandra/audit/AuditLogEntry.java
frankgh: It's a protective copy since the `getMetadata` cannot be guaranteed to be non-mutable. I think we can move the creation of the `AuthenticatedUser` to use the `Map.of` method (`new AuthenticatedUser(role, MTLS, Map.of(METADATA_IDENTITY_KEY, identity))`. This in turn would give us a map that extends from type `AbstractImmutableMap`. `Map.copyOf` will avoid creating a new object and copying the map entries if the type is already an `AbstractImmutableMap`. So this should address the issue of increasing the memory usage.

src/java/org/apache/cassandra/transport/PreV5Handlers.java
frankgh: hmm, I don't know actually. Let me look into that

test/distributed/org/apache/cassandra/distributed/test/auth/AuthAuditLoggingTest.java
frankgh: yeah, good point, let me do that

src/java/org/apache/cassandra/transport/PreV5Handlers.java
tolbertam: We were looking at this in 1:1 chat and we saw that `PipelineConfigurator.configureInitialPipeline` https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/transport/PipelineConfigurator.java#L249-L255 wires up `PreV5Handlers.ExceptionHandler.instance` initially and then it gets swapped out in response to a Startup message handling https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/transport/InitialConnectionHandler.java#L124.

Because this is handling failures happening during a handshake, the PostV5 handler would never be in place to hit this.

src/java/org/apache/cassandra/transport/PreV5Handlers.java
frankgh: It might only be happening in my Mac dev environment

test/distributed/org/apache/cassandra/distributed/test/auth/AuthAuditLoggingTest.java
frankgh: ah interesting . let me dig into this a bit more

frankgh: addressed

tolbertam (approved):
👍 looks great!

frankgh:
Rebased on top of latest trunk

pmcfadin:
Closing as [CASSANDRA-19669](https://issues.apache.org/jira/browse/CASSANDRA-19669) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #3342, opened 2024-05-31
CASSANDRA-19674: CEP-15: (Accord) Bootstraps LocalOnly txn is not written to journal which causes issues once evicted from the cache

dcapwell:


src/java/org/apache/cassandra/service/accord/AccordSafeCommandsForRanges.java
```diff
@@ -21,61 +21,25 @@
...
 
-import accord.primitives.Range;
 import accord.primitives.Ranges;
 import accord.primitives.TxnId;
 import accord.utils.async.AsyncChains;
 import accord.utils.async.AsyncResult;
 import org.apache.cassandra.utils.Pair;
 
-public class AccordSafeCommandsForRanges implements AccordSafeState<Range, CommandsForRanges>
+public class AccordSafeCommandsForRanges extends ImmutableAccordSafeState<Ranges, CommandsForRanges>
```
dcapwell, line 30: the reason for this was that `revert` failed for ranges; found by the improved test cases.  I could just impl `revert` properly in this class but figured a `Immutable*` class would make things more clear

src/java/org/apache/cassandra/service/accord/api/PartitionKey.java
```diff
@@ -147,8 +148,8 @@ public <V> int serialize(PartitionKey key, V dst, ValueAccessor<V> accessor, int
         public PartitionKey deserialize(DataInputPlus in, int version) throws IOException
         {
             TableId tableId = TableId.deserialize(in);
-            TableMetadata metadata = Schema.instance.getExistingTableMetadata(tableId);
-            DecoratedKey key = metadata.partitioner.decorateKey(ByteBufferUtil.readWithShortLength(in));
+            IPartitioner partitioner = Schema.instance.getExistingTablePartitioner(tableId);
```
dcapwell, line 151: using this new method lets test focus only on the partitioner and not define a random table just to access a partitioner

src/java/org/apache/cassandra/service/accord/async/AsyncLoader.java
```diff
@@ -196,7 +195,11 @@ public void onAdd(AccordCachingState<Key, CommandsForKey> state)
 
     private AsyncChain<Set<? extends Key>> findOverlappingKeys(Ranges ranges)
     {
-        Invariants.checkArgument(!ranges.isEmpty());
+        if (ranges.isEmpty())
```
dcapwell, line 198: failed during improved testing.  so now allow and documented the edge case seen

src/java/org/apache/cassandra/service/accord/AccordSafeCommandsForRanges.java
belliottsmith: I think most safe states are immutable, aren't they? Should they all be migrated to this new interface?

belliottsmith (approved):

src/java/org/apache/cassandra/service/accord/AccordSafeCommandsForRanges.java
dcapwell: most allow `set` to update.  The data is immutable but the references may be updated; but CommandsForRanges doesn't allow `set`.

PR #3344, opened 2024-06-02
CASSANDRA-19676 - Replace Stream iteration with for-loop for StorageProxy

samueldlightfoot:
On profiling a write-heavy workload (90% writes) using easy-cass-stress, it became very clear StorageProxy::updateCoordinatorWriteLatencyTableMetric was a hot path that ~15% of the CPU cycles of ModificationStatement::executeWithoutCondition were taken up by (see attached async-profiler image).

We should convert this stream to a simple for loop, as has been discussed recently on the mail list.

Before

image

After

image


cscotta:
Great find, thanks for this profile and patch.

Do you have an “after” profile; or can you share before/after perf stats from easy-cass-stress?

samueldlightfoot:
> Great find, thanks for this profile and patch.
> 
> Do you have an “after” profile; or can you share before/after perf stats from easy-cass-stress?

Updated the PR desc with after. This seems rather significant...

PR #3348, opened 2024-06-05
CASSANDRA-19679 - Replace Stream iteration with for-loop for SimpleRestriction::bindAndGetClusteringElements

samueldlightfoot:
Part 2 (of 2) of low-hanging fruit Stream performance improvements.

The second main Stream contributor to allocations and CPU was SimpleRestriction::bindAndGetClusteringElements, which contributes to 4.58% of all allocations for a 50/50 workload. The image attached shows allocation profiling with trunk (see purple highlighted sections).

The [Cassandra Jira](https://issues.apache.org/jira/browse/CASSANDRA-19679)

See Jira for full HTML flame graphs.

Before (4.58% allocations)

image

After (1.37% allocations)

image

### DIFF
--- a/src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
+++ b/src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
@@ -256,7 +256,8 @@ public final class SimpleRestriction implements SingleRestriction
     private List<ClusteringElements> bindAndGetMultiTermClusteringElements(QueryOptions options)
     {
         List<List<ByteBuffer>> values = bindAndGetElements(options);
-        if (values.isEmpty()) {
+        if (values.isEmpty())
+        {
             return Collections.emptyList();
         }
 
--- a/src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
+++ b/src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
@@ -243,7 +243,7 @@ public final class SimpleRestriction implements SingleRestriction
     {
         List<ByteBuffer> values = bindAndGet(options);
         if (!values.isEmpty()) {
-            List<ClusteringElements> elements = new ArrayList<>();
+            List<ClusteringElements> elements = new ArrayList<>(values.size());
             for (ByteBuffer value : values) {
                 ClusteringElements byteBuffers = ClusteringElements.of(columnsExpression.columnSpecification(), value);
                 elements.add(byteBuffers);
@@ -257,7 +257,7 @@ public final class SimpleRestriction implements SingleRestriction
     {
         List<List<ByteBuffer>> values = bindAndGetElements(options);
         if (!values.isEmpty()) {
-            List<ClusteringElements> elements = new ArrayList<>();
+            List<ClusteringElements> elements = new ArrayList<>(values.size());
             for (List<ByteBuffer> value : values) {
                 ClusteringElements byteBuffers = ClusteringElements.of(columnsExpression.columns(), value);
                 elements.add(byteBuffers);
--- a/src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
+++ b/src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
@@ -248,7 +248,7 @@ public final class SimpleRestriction implements SingleRestriction
                 ClusteringElements byteBuffers = ClusteringElements.of(columnsExpression.columnSpecification(), value);
                 elements.add(byteBuffers);
             }
-            return elements;
+            return Collections.unmodifiableList(elements);
         }
         return Collections.emptyList();
     }
@@ -262,7 +262,7 @@ public final class SimpleRestriction implements SingleRestriction
                 ClusteringElements byteBuffers = ClusteringElements.of(columnsExpression.columns(), value);
                 elements.add(byteBuffers);
             }
-            return elements;
+            return Collections.unmodifiableList(elements);
         }
         return Collections.emptyList();
     }
--- a/src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
+++ b/src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
@@ -19,6 +19,7 @@
 package org.apache.cassandra.cql3.restrictions;
 
 import java.nio.ByteBuffer;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
```diff
@@ -225,17 +226,22 @@ public void restrict(RangeSet<ClusteringElements> rangeSet, QueryOptions options
 
     private List<ClusteringElements> bindAndGetClusteringElements(QueryOptions options)
     {
+        List<ClusteringElements> elements = new ArrayList<>();
```
smiklosovic: @samueldlightfoot if you think about that ... if columnsExpression.kind() goes to default case which throws, then you have created an object of an empty list unnecessarily. I think what you could do is to make that list directly in either TOKEN or MULTI_COLUMN case. That way we would save an object creation when it goes to default. 

It would be also interesting to investigate if it is possible (and how frequent) that bindAndGet(options) is empty. Then I would probably check if it is empty and only if it is not I would create that list. Presumably, if it is empty more often than not then checking its size makes sense as I guess that is faster / cheaper than object creation.

edit: checking if an array list is empty or not takes just one method invocation of isEmpty() and comparing size property of int to 0.

samueldlightfoot: A good point. I had assumed the exception creation cost would dwarf list creation. I have tidied this up.

Interesting idea on the empty element checks -- I'd have to do some digging on this. We'll have to ensure an immutable list can be returned, as currently it is a mutable ArrayList returned.

I have included the list element check, but I would be keen to hear from anyone familiar if this is a valid case to handle.

src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
```diff
@@ -229,18 +231,42 @@ private List<ClusteringElements> bindAndGetClusteringElements(QueryOptions optio
...
+    private List<ClusteringElements> bindAndGetSingleTermClusteringElements(QueryOptions options)
+    {
+        List<ByteBuffer> values = bindAndGet(options);
+        if (!values.isEmpty()) {
+            List<ClusteringElements> elements = new ArrayList<>();
+            for (ByteBuffer value : values) {
+                ClusteringElements byteBuffers = ClusteringElements.of(columnsExpression.columnSpecification(), value);
+                elements.add(byteBuffers);
+            }
+            return elements;
```
smiklosovic: @samueldlightfoot just wrap `elements` in `Collections.unmodifiableList(elements)` if you need it to be unmodifiable

samueldlightfoot: The stream based approach did not return an immutable list, but given this collection is exposed as part of the public interface via _values_ I believe this is a good idea.

src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
```diff
@@ -229,18 +231,42 @@ private List<ClusteringElements> bindAndGetClusteringElements(QueryOptions optio
...
+                return bindAndGetMultiTermClusteringElements(options);
             default:
                 throw new UnsupportedOperationException();
         }
     }
 
+    private List<ClusteringElements> bindAndGetSingleTermClusteringElements(QueryOptions options)
+    {
+        List<ByteBuffer> values = bindAndGet(options);
+        if (!values.isEmpty()) {
```
smiklosovic: @samueldlightfoot could you please put the opening braces on a new line everywhere in this patch in order to follow the code style?

src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
```diff
@@ -229,18 +231,42 @@ private List<ClusteringElements> bindAndGetClusteringElements(QueryOptions optio
...
+            List<ClusteringElements> elements = new ArrayList<>(values.size());
+            for (ByteBuffer value : values) {
+                ClusteringElements byteBuffers = ClusteringElements.of(columnsExpression.columnSpecification(), value);
+                elements.add(byteBuffers);
```
smiklosovic: You could omit `byteBuffers` variable by putting the result of `ClusteringElements.of` directly to `elements.add()`

By doing so, you will have a `for` loop consisting of one line of the code in its body., In that case you can omit braces for `for` entirely.

src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
```diff
@@ -229,18 +231,40 @@ private List<ClusteringElements> bindAndGetClusteringElements(QueryOptions optio
...
         }
     }
 
+    private List<ClusteringElements> bindAndGetSingleTermClusteringElements(QueryOptions options)
+    {
+        List<ByteBuffer> values = bindAndGet(options);
+        if (!values.isEmpty())
+        {
+            List<ClusteringElements> elements = new ArrayList<>(values.size());
+            for (ByteBuffer value : values)
```
blerer: If we are concerned about allocations at this place, we should use a `for (int i = , m = values.size(); i < m; i++)` loop. We replaced those loops in the past already for the same reason.

src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
```diff
@@ -229,18 +231,40 @@ private List<ClusteringElements> bindAndGetClusteringElements(QueryOptions optio
...
+                elements.add(ClusteringElements.of(columnsExpression.columnSpecification(), value));
+            return Collections.unmodifiableList(elements);
```
blerer: There is no need to wrap the list into an unmodifiable list.

src/java/org/apache/cassandra/cql3/restrictions/SimpleRestriction.java
```diff
@@ -229,18 +231,40 @@ private List<ClusteringElements> bindAndGetClusteringElements(QueryOptions optio
...
+                return bindAndGetMultiTermClusteringElements(options);
             default:
                 throw new UnsupportedOperationException();
         }
     }
 
+    private List<ClusteringElements> bindAndGetSingleTermClusteringElements(QueryOptions options)
+    {
+        List<ByteBuffer> values = bindAndGet(options);
+        if (!values.isEmpty())
```
blerer: The code will be easier to read if we handle the empty case right from the start:
```
if (values.isEmpty())
    return Collections.emptyList();
```

blerer (approved):

PR #3350, opened 2024-06-06
CASSANDRA-19400 IndexStatusManager should prioritize replicas with BUILD_SUCCEEDED indexes over UNKNOWN ones

arkn98:
Patch by Arun Ganesh; reviewed by TBD for CASSANDRA-19400

### DIFF
--- a/src/java/org/apache/cassandra/index/IndexStatusManager.java
+++ b/src/java/org/apache/cassandra/index/IndexStatusManager.java
@@ -19,6 +19,7 @@
 package org.apache.cassandra.index;
 
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -41,6 +42,7 @@ import org.apache.cassandra.gms.Gossiper;
 import org.apache.cassandra.gms.VersionedValue;
 import org.apache.cassandra.locator.Endpoints;
 import org.apache.cassandra.locator.InetAddressAndPort;
+import org.apache.cassandra.locator.Replica;
 import org.apache.cassandra.serializers.MarshalException;
 import org.apache.cassandra.service.StorageService;
 import org.apache.cassandra.utils.FBUtilities;
@@ -84,17 +86,32 @@ public class IndexStatusManager
      */
     public <E extends Endpoints<E>> E filterForQuery(E liveEndpoints, Keyspace keyspace, Index.QueryPlan indexQueryPlan, ConsistencyLevel level)
     {
+        // UNKNOWN states are transient/rare; only a few replicas should have this state at any time. See CASSANDRA-19400
+        Set<Replica> queryableNonSucceeded = new HashSet<>(4);
+
         E queryableEndpoints = liveEndpoints.filter(replica -> {
 
+            boolean allBuilt = true;
             for (Index index : indexQueryPlan.getIndexes())
             {
                 Index.Status status = getIndexStatus(replica.endpoint(), keyspace.getName(), index.getIndexMetadata().name);
                 if (!index.isQueryable(status))
                     return false;
+
+                if (status != Index.Status.BUILD_SUCCEEDED)
+                    allBuilt = false;
             }
+
+            if (!allBuilt)
+                queryableNonSucceeded.add(replica);
 
             return true;
         });
+
+        // deprioritize replicas with queryable but non-succeeded indexes
+        if (!queryableNonSucceeded.isEmpty() && queryableNonSucceeded.size() != queryableEndpoints.size())
+            queryableEndpoints = queryableEndpoints.sorted(
+                    Comparator.comparingInt(e -> queryableNonSucceeded.contains(e) ? 1 : -1));
 
         int initial = liveEndpoints.size();
         int filtered = queryableEndpoints.size();
--- a/test/unit/org/apache/cassandra/index/IndexStatusManagerTest.java
+++ b/test/unit/org/apache/cassandra/index/IndexStatusManagerTest.java
@@ -1,0 +1,423 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.index;
+
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
+
+import org.junit.Test;
+import org.mockito.Mockito;
+
+import org.apache.cassandra.db.ConsistencyLevel;
+import org.apache.cassandra.db.Keyspace;
+import org.apache.cassandra.exceptions.ReadFailureException;
+import org.apache.cassandra.gms.VersionedValue;
+import org.apache.cassandra.locator.AbstractReplicationStrategy;
+import org.apache.cassandra.locator.EndpointsForRange;
+import org.apache.cassandra.locator.InetAddressAndPort;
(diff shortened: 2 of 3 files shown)


src/java/org/apache/cassandra/index/IndexStatusManager.java
```diff
@@ -84,26 +85,39 @@ private IndexStatusManager()
      */
     public <E extends Endpoints<E>> E filterForQuery(E liveEndpoints, Keyspace keyspace, Index.QueryPlan indexQueryPlan, ConsistencyLevel level)
     {
+        int required = level.blockFor(keyspace.getReplicationStrategy());
+        ReplicaCollection.Builder<E> buildSucceededEndpoints = liveEndpoints.newBuilder(0);
```
maedhroz:
```suggestion
    ReplicaCollection.Builder<E> buildSucceededEndpoints = liveEndpoints.newBuilder(liveEndpoints.size());
```
The only other option would be to use `required` as an initial size, but I think we'll expect most nodes to be in the `BUILD_SUCCEEDED` state most of the time, so that would force more array creations. For instance, if we have an RF = 3, and CL = QUORUM, we probably want to use `liveEndpoints.size()` which would be 3, instead of using 2 and having to do an extra array copy if all three replicas are `BUILD_SUCCEEDED`.

WDYT?

arkn98: I'm no expert, but what you suggested makes sense. I'll use `liveEndpoints.size()`.

test/unit/org/apache/cassandra/index/IndexStatusManagerTest.java
```diff
@@ -0,0 +1,393 @@
...
+
+        static class Builder
+        {
+            String keyspace;
+            int minRequiredNodes;
+            AbstractReplicationStrategy replicationStrategy;
+            Map<InetAddressAndPort, Map<String, Index.Status>> indexStatus;
+            EndpointsForRange expected;
+
+            Builder keyspace(String kp)
```
maedhroz:
```suggestion
            Builder keyspace(String ks)
```
super nit: We usually use "ks" as an abbreviation for keyspace, but no important

test/unit/org/apache/cassandra/index/IndexStatusManagerTest.java
```diff
@@ -0,0 +1,393 @@
...
+            }
+        }
+    }
+
+    // we have enough BUILD_SUCCEEDED endpoints: filterForQuery should prioritize them, even if the other endpoints'
+    // indexes are queryable but not built (like UNKNOWN)
+    @Test
+    public void filterForQuery_ShouldPrioritizeSuccessfulEndpoints()
+    {
+        runTest("exactly enough endpoints with BUILD_SUCCEEDED indexes",
```
maedhroz: I'd make this its own top-level test, called perhaps `shouldPrioritizeSingleSuccessfulEndpoint()`

test/unit/org/apache/cassandra/index/IndexStatusManagerTest.java
```diff
@@ -0,0 +1,393 @@
...
+                                )
+                        ))
+                        .minRequiredNodes(1)
+                        .expected(EndpointsForRange.of(
+                                full(InetAddressAndPort.getByNameUnchecked("127.0.0.254"))
+                        ))
+                        .build()
+        );
+
+        runTest("more than enough endpoints with BUILD_SUCCEEDED indexes",
```
maedhroz: I'd make this its own top-level test, called perhaps `shouldPrioritizeMultipleSuccessfulEndpoints()`

test/unit/org/apache/cassandra/index/IndexStatusManagerTest.java
```diff
@@ -0,0 +1,393 @@
...
+                                full(InetAddressAndPort.getByNameUnchecked("127.0.0.252")),
+                                full(InetAddressAndPort.getByNameUnchecked("127.0.0.251"))
+                        ))
+                        .build()
+        );
+    }
+
+    // we don't have enough BUILD_SUCCEEDED endpoints: filterForQuery should return all queryable endpoints
+    @Test
+    public void filterForQuery_ReturnAllEndpointsWhenNotEnoughSuccessfulEndpoints()
```
maedhroz:
```suggestion
    public void shouldIncludeUnknownEndpointsWhenNotEnoughSucceeded()
```

test/unit/org/apache/cassandra/index/IndexStatusManagerTest.java
```diff
@@ -0,0 +1,393 @@
...
+                        ))
+                        .build()
+        );
+    }
+
+    // we don't have enough queryable endpoints: filterForQuery should throw an exception
+    @Test
+    public void filterForQuery_ThrowExceptionWhenNotEnoughQueryableEndpoints()
+    {
+        assertThatThrownBy(() ->
```
maedhroz: I'd make this its own top-level test, called perhaps `shouldThrowWhenNotEnoughQueryableEndpoints()`

test/unit/org/apache/cassandra/index/IndexStatusManagerTest.java
```diff
@@ -0,0 +1,393 @@
...
+                                        )
+                                ))
+                                .minRequiredNodes(4)
+                                .build()))
+                .isInstanceOf(ReadFailureException.class)
+                .hasMessageStartingWith("Operation failed")
+                .hasMessageContaining("INDEX_NOT_AVAILABLE from /127.0.0.254:7000")
+                .hasMessageContaining("INDEX_NOT_AVAILABLE from /127.0.0.252:7000");
+
+        assertThatThrownBy(() ->
```
maedhroz: I'd make this its own top-level test, called perhaps `shouldThrowWithNoQueryableEndpoints()`

maedhroz (approved):
The new tests look good! Just left a couple minor comments on naming and making sure we break them up into a single `@Test` per `Testcase`.

src/java/org/apache/cassandra/index/IndexStatusManager.java
```diff
@@ -84,26 +85,39 @@ private IndexStatusManager()
      */
     public <E extends Endpoints<E>> E filterForQuery(E liveEndpoints, Keyspace keyspace, Index.QueryPlan indexQueryPlan, ConsistencyLevel level)
```
jasonstack, line 87: The next call after `filterForQuery` is to filter for EACH_QUOUM in `ReplicaPlans#contactForRead`. If 1st DC has 3 succeeded replicas and 2nd DC has 3 unknown repliacs. `filterForQuery` will return 3 succeeded + 1 unknown and will fail the consistency requirement for `EACH_QUOUM`. Not sure if we care about EACH_QUOUM for indexes.

The safest approach here is to sort replicas by query status. How about following?

```
        Set<Replica> queryableNonSucceeded = new HashSet<>(liveEndpoints.size());
        E queryableEndpoints = liveEndpoints.filter(replica -> {

            boolean allBuilt = true;
            for (Index index : indexQueryPlan.getIndexes())
            {
                Index.Status status = getIndexStatus(replica.endpoint(), keyspace.getName(), index.getIndexMetadata().name);
                if (!index.isQueryable(status))
                    return false;

                if (status != Index.Status.BUILD_SUCCEEDED)
                    allBuilt = false;
            }

            if (!allBuilt)
                queryableNonSucceeded.add(replica);

            return true;
        });

        // deprioritize replica with queryable but non-succeeded state
        if (!queryableNonSucceeded.isEmpty())
            queryableEndpoints = queryableEndpoints.sorted(Comparator.comparingInt(r -> queryableNonSucceeded.contains(r) ? 1 : -1));
```

maedhroz: `contactForRead()` just generates a sub-list from left to right, so that should work. In most cases, the sort shouldn't even happen, as `UNKNOWN` states should be transient/rare. Since the number of nodes in this state should very small, and we may have a large cluster, I'd also make the initial size of `queryableNonSucceeded` something reasonably small, like 4 (w/ a comment), or even use the default (16). With the `ReplicaCollection.Builder` it made sense to do the opposite, but that, in retrospect, would not have been very nice either in a large cluster with most things built/succeeded (aside from avoiding resizing).

maedhroz: For partition-restricted index queries, the `liveEndpoints` might be small, of course, but I'd still size `queryableNonSucceeded` explicitly at something like 4.

arkn98: Addressed in the latest commit. Thanks for the review!

jasonstack (approved):

maedhroz:
Committed as https://github.com/apache/cassandra/commit/b8d69d02614973d79d2c70d65aea4612dad6e36c

PR #3355, opened 2024-06-07
CASSANDRA-19677: Adds new size Guardrails

bbotella:
CASSANDRA-19677: Adds new size Guardrails

This commit adds these column size guardrails:
```
column_ascii_value_size_warn_threshold
column_ascii_value_size_fail_threshold
column_blob_value_size_warn_threshold
column_blob_value_size_fail_threshold
column_text_value_size_warn_threshold
column_text_value_size_fail_threshold
column_varchar_value_size_warn_threshold
column_varchar_value_size_fail_threshold
```

And these collection size guardrails:
```
collection_map_size_warn_threshold
collection_map_size_fail_threshold
collection_set_size_warn_threshold
collection_set_size_fail_threshold
collection_list_size_warn_threshold
collection_list_size_fail_threshold
```

patch by Bernardo Botella Corbi; reviewed by <Reviewers> for CASSANDRA-19677


conf/cassandra.yaml
```diff
@@ -2077,6 +2077,42 @@ drop_compact_storage_enabled: false
 # column_value_size_warn_threshold:
 # column_value_size_fail_threshold:
 #
+# Guardrail to warn or fail when writing ascii column values larger than threshold.
+# This guardrail is only applied to the values of regular columns because both the serialized partitions keys and the
+# values of the components of the clustering key already have a fixed, relatively small size limit of 65535 bytes, which
+# is probably lesser than the thresholds defined here.
+# The two thresholds default to null to disable.
```
yifan-c, line 2323: Can you add the clarification when both `column_value_size_warn_threshold` and `column_ascii_value_size_warn_threshold` are enabled? 
For example, `column_value_size_warn_threshold=100B` and `column_ascii_value_size_warn_threshold=200B`. If the incoming write has the `ascii` field size of 123B, should it warn or not?

src/java/org/apache/cassandra/cql3/UpdateParameters.java
```diff
@@ -163,8 +166,21 @@ public Cell<?> addCell(ColumnMetadata column, ByteBuffer value) throws InvalidRe
 
     public Cell<?> addCell(ColumnMetadata column, CellPath path, ByteBuffer value) throws InvalidRequestException
     {
+        // General column value size
         Guardrails.columnValueSize.guard(value.remaining(), column.name.toString(), false, clientState);
 
+        if (column.type instanceof AsciiType) // Ascii size specific guardrail
+        {
+            Guardrails.columnAsciiValueSize.guard(value.remaining(), column.name.toString(), false, clientState);
+        } else if (column.type instanceof BytesType) // Blob size specific guardrail
```
yifan-c: move to new line. There are a few other places that the `else if` is placed wrongly. Please correct them

src/java/org/apache/cassandra/cql3/UpdateParameters.java
```diff
@@ -163,8 +166,21 @@ public Cell<?> addCell(ColumnMetadata column, ByteBuffer value) throws InvalidRe
...
+        {
+            Guardrails.columnBlobValueSize.guard(value.remaining(), column.name.toString(), false, clientState);
+        } else if (column.type instanceof UTF8Type) // text and varchar size specific guardrails
+        {
+            Guardrails.columnTextValueSize.guard(value.remaining(), column.name.toString(), false, clientState);
+            Guardrails.columnVarcharValueSize.guard(value.remaining(), column.name.toString(), false, clientState);
+        }
```
yifan-c: It is confusing (sometimes surprising) to end-users to have 2 guardrails to check on the same value. Since `text` and `varchar` are identical internally, can you combine those 2 guardrails?

src/java/org/apache/cassandra/cql3/UpdateParameters.java
```diff
@@ -163,8 +166,21 @@ public Cell<?> addCell(ColumnMetadata column, ByteBuffer value) throws InvalidRe
 
     public Cell<?> addCell(ColumnMetadata column, CellPath path, ByteBuffer value) throws InvalidRequestException
     {
+        // General column value size
         Guardrails.columnValueSize.guard(value.remaining(), column.name.toString(), false, clientState);
 
+        if (column.type instanceof AsciiType) // Ascii size specific guardrail
```
yifan-c: nit: prefer comparing the enum type, instead of `instanceof`, e.g.

```suggestion
        if (column.type.asCQL3Type() == CQL3Type.Native.ASCII)
```

src/java/org/apache/cassandra/cql3/UpdateParameters.java
```diff
@@ -163,8 +163,28 @@ public Cell<?> addCell(ColumnMetadata column, ByteBuffer value) throws InvalidRe
 
     public Cell<?> addCell(ColumnMetadata column, CellPath path, ByteBuffer value) throws InvalidRequestException
     {
+        // General column value size
         Guardrails.columnValueSize.guard(value.remaining(), column.name.toString(), false, clientState);
 
+        if (column.type.asCQL3Type() == CQL3Type.Native.ASCII) // Ascii size specific guardrail
```
jrwest: Why not a switch statement here? if not, no need for the else's with the nested ifs, it makes it more verbose -- use else if instead. Also think moving this to a `validateSize` function (or something similarly named) would make it more clear

src/java/org/apache/cassandra/cql3/terms/Lists.java
```diff
@@ -414,12 +414,14 @@ static void doAppend(Term.Terminal value, ColumnMetadata column, UpdateParameter
                     dataSize += cell.dataSize();
                 }
                 Guardrails.collectionSize.guard(dataSize, column.name.toString(), false, params.clientState);
+                Guardrails.collectionListSize.guard(dataSize, column.name.toString(), false, params.clientState);
```
yifan-c: I am wondering if you can sort of extending the `collectionListSize` from `collectionSize` and only call `collectionListSize` here, to avoid the "repetition". Similarly, it can be applied for map and set. 

Basically, something like this. Note the `warnThreshold` and `failThreshold` functions. 
```java
    public static final MaxThreshold collectionListSize =
    new MaxThreshold("collection_list_size",
                     null,
                     state -> {
                         Long listSizeWarnThreshold = sizeToBytes(CONFIG_PROVIDER.getOrCreate(state).getCollectionListSizeWarnThreshold());
                         return Objects.requireNonNullElseGet(listSizeWarnThreshold, () -> collectionSize.warnValue(state));
                     },
                     state -> {
                         Long listSizeWarnThreshold = sizeToBytes(CONFIG_PROVIDER.getOrCreate(state).getCollectionListSizeFailThreshold());
                         return Objects.requireNonNullElseGet(listSizeWarnThreshold, () -> collectionSize.failValue(state));
                     },
                     format("Detected collection %s of size %s, this exceeds the %s threshold of %s.",
```

bbotella: I hear you, and it makes sense. But I kind of like the verbosity of having the two separate guardrails explicitly called. I think from a readability standpoint, it makes it easier to see "both" guardrails being checked in order, instead of hiding the check of the former as part of the check of the latter. Again, on the fence on this one, so, if you feel this is the right way to go, I can definitely change it.

yifan-c: Let's go a bit deeper and compare the difference. (Yes. They have difference behavior). 

The curried function (my suggestion) defines the evaluation priority and check once. That is, as long as collectionListSize is defined, it takes the precedence, regardless of their threshold values. It makes sense, since this patch brings a set of more finer-grained guardrails regarding collection sizes. 

The handling in this patch, check the size _twice_, if they are both defined/enabled. In this case, it is the lowest threshold value that enforce the guardrail. The value could come from `collectionListSize` or `collectionSize`. Imagining from users' perspective, if the size guardrail on List is defined, it could cause confusion if a list size that satisfies the list size guardrail fails (due to the lower threshold from collection size). 

In the end, it needs documentation to elaborate the behaviors of the overlapping guardrails.

conf/cassandra.yaml
```diff
@@ -2365,8 +2363,8 @@ drop_compact_storage_enabled: false
 # of non-frozen collections there could be unaccounted parts of the collection on the sstables. This is done this way to
 # prevent read-before-write. The guardrail is also checked at sstable write time to detect large non-frozen collections,
 # although in that case exceeding the fail threshold will only log an error message, without interrupting the operation.
-# Also, this guardrail can be used with the generic collection_size_fail_threshold. The lowest of the two guardrail will
-# be the one enforced.
+# When collection_map_size_warn_threshold is defined, it takes the precedence over generic collection_size_fail_threshold
+# guardrail.
```
yifan-c: This still seems wrong to me. 

```suggestion
# When collection_map_size_warn_threshold and/or collection_map_size_fail_threshold are defined, they take the precedence
# over the cooresponding collection_size_warn_threshold and/or collection_size_fail_threshold.
```

pmcfadin:
Closing as [CASSANDRA-19677](https://issues.apache.org/jira/browse/CASSANDRA-19677) has been resolved as Fixed — if this PR contains changes not captured in the merged fix, please reopen.

PR #3365, opened 2024-06-10
CASSANDRA-19693 Relax slow_query_log_timeout for MultiNodeSAITest

maedhroz:
patch by Caleb Rackliffe; reviewed by Alex Petrov for CASSANDRA-19693

### DIFF
--- a/test/distributed/org/apache/cassandra/fuzz/sai/MultiNodeSAITest.java
+++ b/test/distributed/org/apache/cassandra/fuzz/sai/MultiNodeSAITest.java
@@ -45,7 +45,10 @@ public class MultiNodeSAITest extends SingleNodeSAITest
         cluster = Cluster.build()
                          .withNodes(2)
                          // At lower fetch sizes, queries w/ hundreds or thousands of matches can take a very long time. 
-                         .withConfig(InJvmSutBase.defaultConfig().andThen(c -> c.set("range_request_timeout", "180s").set("read_request_timeout", "180s")
+                         .withConfig(InJvmSutBase.defaultConfig().andThen(c -> c.set("range_request_timeout", "180s")
+                                                                                .set("read_request_timeout", "180s")
+                                                                                .set("native_transport_timeout", "180s")
+                                                                                .set("slow_query_log_timeout", "180s")
                                                                                 .with(GOSSIP).with(NETWORK)))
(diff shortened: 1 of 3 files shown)


test/distributed/org/apache/cassandra/fuzz/sai/MultiNodeSAITest.java
```diff
@@ -45,7 +45,10 @@ public static void before() throws Throwable
         cluster = Cluster.build()
                          .withNodes(2)
                          // At lower fetch sizes, queries w/ hundreds or thousands of matches can take a very long time. 
-                         .withConfig(InJvmSutBase.defaultConfig().andThen(c -> c.set("range_request_timeout", "180s").set("read_request_timeout", "180s")
+                         .withConfig(InJvmSutBase.defaultConfig().andThen(c -> c.set("range_request_timeout", "180s")
```
bbotella, line 48: Nit: Maybe worth extracting that `180s` to a constant?

test/harry/main/org/apache/cassandra/harry/sut/injvm/InJvmSutBase.java
```diff
@@ -144,38 +144,41 @@ public Object[][] execute(String statement, ConsistencyLevel cl, int coordinator
...
-                       .coordinator(coordinator)
-                       .execute(statement, toApiCl(cl), bindings);
-            }
-        }
-        catch (Throwable t)
-        {
-            if (retryStrategy.apply(t))
-                return execute(statement, cl, bindings);
+                if (retryStrategy.apply(t))
+                    continue;
```
bbotella, line 175: Just a question for my education. Are we not applying some backoff or delay between retries?

ifesdjeen: We could apply backoff, but this is an in-jvm test, so the initial thinking was that it's a very controlled environment.

test/distributed/org/apache/cassandra/fuzz/sai/MultiNodeSAITest.java
maedhroz: Want to leave some room to change these in the future, as they've been chosen somewhat arbitrarily.

PR #3367, opened 2024-06-11
autorepair v2 framework (against 4.1.6)

jaydeepkumar1984:
CEP:
https://cwiki.apache.org/confluence/display/CASSANDRA/CEP-37+Apache+Cassandra+Unified+Repair+Solution

Design doc:
https://docs.google.com/document/d/1CJWxjEi-mBABPMZ3VWJ9w5KavWfJETAGxfUpsViPcPo/edit?tab=t.0#heading=h.r112r46toau0


src/java/org/apache/cassandra/schema/SchemaKeyspace.java
```diff
@@ -552,7 +554,8 @@ private static void addTableParamsToRowBuilder(TableParams params, Row.SimpleBui
                .add("compaction", params.compaction.asMap())
                .add("compression", params.compression.asMap())
                .add("read_repair", params.readRepair.toString())
-               .add("extensions", params.extensions);
+               .add("extensions", params.extensions)
+               .add("disabled_automated_repair", params.disableAutomatedRepair);
```
tolbertam: :+1: thanks for doing this.   I imagine it could be useful to be able to configure things more granularly.   I have some cases where we want to run full repair on a table, but never incremental (the main use case is analytics based where etl processes are loading a bunch of SSTables every few hours).   Will come back to this as I review and experiment more with the PR, if there is a bunch of things that we would want configurable at table level maybe we could consider adding something like a RepairParams, otherwise maybe this is fine as is.

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
```diff
@@ -0,0 +1,390 @@
...
+
+    public boolean isAutoRepairEnabled(RepairType repairType)
+    {
+        return enabled && applyOverrides(repairType, opt -> opt.enabled);
+    }
+
+    public void setAutoRepairEnabled(RepairType repairType, boolean enabled)
+    {
+        if (enabled && repairType == RepairType.incremental &&
+            (DatabaseDescriptor.getMaterializedViewsEnabled() || DatabaseDescriptor.isCDCEnabled()))
```
tolbertam: I think for a first pass this is ok, but eventually would like to see this apply more at a table label;  Tables that have/are views would be skipped for incremental, and also those that have CDC enabled.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
tolbertam: This may be overkill and some of the options may not be applicable, but wonder if we could just have a single field that functions similarly to `AutoRepairConfig.repair_type_overrides`:

So one could do something like:

```
ALTER TABLE x.y WITH automated_repair={'incremental': {'enabled': false}};
```

Most of the `repair_type_overrides` aren't really applicable at a table level, so maybe we'd just have a separate class like `AutoRepairParams` that functioned similarly to `CompactionParams`

src/java/org/apache/cassandra/cql3/UntypedResultSet.java
```diff
@@ -432,6 +432,15 @@ public long getLong(String column)
...
         }
 
+        // this function will return the default value if the row doesn't have that column or the column data is null
+        // This function is used to avoid the nullpointerexception
+        public long getLongOrDefault(String column, long defaultValue) {
+            if (!has(column)){
+                return defaultValue;
+            }
+            return getLong(column);
+        }
```
tolbertam: Nit: there are some existing methods (`getInt`, `getUUID`) that are overloaded with an additional `ifNull` parameter, for consistency we should do the same for getLong.

```suggestion
        public long getLong(String column, long ifNull) {
            ByteBuffer bytes = data.get(column);
            return bytes == null ? ifNull : LongType.instance.compose(bytes);
        }
```

src/java/org/apache/cassandra/repair/autorepair/AutoRepairUtils.java
```diff
@@ -0,0 +1,875 @@
...
+    "SELECT * FROM %s.%s WHERE %s = ? AND %s = ?", SchemaConstants.AUTO_REPAIR_KEYSPACE_NAME,
+    AutoRepairKeyspace.AUTO_REPAIR_PRIORITY, COL_REPAIR_TYPE, COL_PID);
+    final static String DEL_REPAIR_PRIORITY = String.format(
+    "DELETE %s[?] FROM %s.%s WHERE %s = ? AND %s = ?", COL_REPAIR_PRIORITY, SchemaConstants.AUTO_REPAIR_KEYSPACE_NAME,
+    AutoRepairKeyspace.AUTO_REPAIR_PRIORITY, COL_REPAIR_TYPE, COL_PID);
+    final static String ADD_PRIORITY_HOST = String.format(
+    "UPDATE %s.%s SET %s = %s + ?  WHERE %s = ? AND %s = ?", SchemaConstants.AUTO_REPAIR_KEYSPACE_NAME,
+    AutoRepairKeyspace.AUTO_REPAIR_PRIORITY, COL_REPAIR_PRIORITY, COL_REPAIR_PRIORITY, COL_REPAIR_TYPE, COL_PID);
+
+    final static String INSERT_NEW_REPAIR_HISTORY = String.format(
```
tolbertam, line 141: Is there precedence for using Lightweight transactions in internal cassandra stuff?   Looking at how this is used, I see we check to see if the CAS insert was applied, but that just effects logging, so I think there is little harm in removing `IF NOT EXISTS` and just let it overwrite.

src/java/org/apache/cassandra/repair/autorepair/AutoRepairUtils.java
```diff
@@ -0,0 +1,875 @@
...
+    "INSERT INTO %s.%s (%s, %s, %s, %s, %s, %s, %s) values (?, ?, ? ,?, ?, {}, ?) IF NOT EXISTS",
+    SchemaConstants.AUTO_REPAIR_KEYSPACE_NAME, AutoRepairKeyspace.AUTO_REPAIR_HISTORY, COL_REPAIR_TYPE, COL_PID,
+    COL_HOST_ID, COL_REPAIR_START_TS, COL_REPAIR_FINISH_TS, COL_DELETE_HOSTS, COL_DELETE_HOSTS_UPDATE_TIME);
+
+    final static String ADD_HOST_ID_TO_DELETE_HOSTS = String.format(
+    "UPDATE %s.%s SET %s = %s + ?, %s = ? WHERE %s = ? AND %s = ? AND %s = ? IF EXISTS"
```
tolbertam: Similar to `INSERT_NEW_REPAIR_HISTORY`, we can just do a blind delete here (no need for IF EXISTS), if it doesn't exist it'll just create a row tombstone.  Given how this is only executed when we detect the host isn't in the ring seems fine to just do that.

src/java/org/apache/cassandra/metrics/AutoRepairMetrics.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.cassandra.repair.autorepair.AutoRepair;
+
+import static org.apache.cassandra.metrics.CassandraMetricsRegistry.Metrics;
+
+/**
+ * Metrics related to AutoRepair.
+ */
+public class AutoRepairMetrics
+{
+    public Gauge<Integer> repairsInProgress;
```
tolbertam: Fantastic to have these out of the gate :+1:.  We should add some docs to [`metrics.adoc`](https://github.com/apache/cassandra/blob/trunk/doc/modules/cassandra/pages/managing/operating/metrics.adoc) to document these.

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
```diff
@@ -0,0 +1,390 @@
...
+        public Options()
+        {
+        }
+
+        @VisibleForTesting
+        protected static Options getDefaultOptions()
+        {
+            Options opts = new Options();
+
+            opts.enabled = false;
```
tolbertam, line 464: One thing worth discussing;  by default I noticed that if I set:

```yaml
auto_repair:
  enabled: true
```

That it doesn't actually run repairs because by default both full and incremental repair are disabled.  Is this how we want this to work?  I think it's a bit subjective and I don't have a good answer, but felt it was worth discussing.

I think two possible good defaults would be:

1. Keep it as-is, if you want to enable repair, you should have to choose which of the repair options you would want enabled by default.
2. Full repair is enabled with the above config.

I think maybe what we should just do is go with option one and add a commented section in cassandra.yaml that looks something like this:

```yaml
# Documentation about auto repair goes here
# auto_repair:
#  # Enables auto repair, defaults to false (disabled)
#  enabled: true
#  repair_type_overrides:
#    full:
#      # Enables auto full repairs, defaults to false
#      enabled: true
#     # ... all other various property defaults here
#    incremental:
#      enabled: false
```

src/java/org/apache/cassandra/repair/autorepair/AutoRepairStateFactory.java
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.cassandra.repair.autorepair;
+
+import org.apache.cassandra.repair.autorepair.AutoRepairConfig.RepairType;
+
+public class AutoRepairStateFactory
+{
+    public static AutoRepairState getAutoRepairState(RepairType repairType)
```
tolbertam: Small preference:  It could be better to just put this on `RepairType`, can just define a `AutoRepairState getAutoRepairState()` on the `RepairType` enum.

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
```diff
@@ -0,0 +1,390 @@
...
+    public final Boolean enabled;
+    // the interval in seconds between checks for eligible repair operations. Cannot be modified dynamically.
+    public final Integer repair_check_interval_in_sec = 300; // 5 minutes
+    // configures how long repair history is kept for a replaced node
+    public volatile Integer history_clear_delete_hosts_buffer_in_sec = 60 * 60 * 2;  // two hours
+    // global_settings overides Options.defaultOptions for all repair types
+    public volatile Options global_settings;
+
+    public enum RepairType
+    {full, incremental}
```
tolbertam: I like the way you did this.  Seems like if I wanted to in theory add a new repair type to be scheduled (e.g. 'Preview repairs on repaired set', 'Preview repairs on unrepaired set', 'Paxos repairs').  I'd just have to add a new enum here, and then implement a `AutoRepairState` for it and add it to the `AutoRepairStateFactory`? Nice!

tolbertam: Seems to be the case!  That was incredibly easy!

```diff
git diff
diff --git a/src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java b/src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
index 731d95a8dd..03d3b240c5 100644
--- a/src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
+++ b/src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
@@ -46,7 +46,7 @@ public class AutoRepairConfig implements Serializable
     public volatile Options global_settings;
 
     public enum RepairType
-    {full, incremental}
+    {full, incremental, preview_repaired}
 
     // repair_type_overrides overrides the global_settings for a specific repair type
     public volatile Map<RepairType, Options> repair_type_overrides = new EnumMap<>(RepairType.class);
diff --git a/src/java/org/apache/cassandra/repair/autorepair/AutoRepairState.java b/src/java/org/apache/cassandra/repair/autorepair/AutoRepairState.java
index fc0281a7f0..ebda85c52a 100644
--- a/src/java/org/apache/cassandra/repair/autorepair/AutoRepairState.java
+++ b/src/java/org/apache/cassandra/repair/autorepair/AutoRepairState.java
@@ -283,6 +283,27 @@ public abstract class AutoRepairState implements ProgressListener
     }
 }
 
+class PreviewRepairedState extends AutoRepairState
+{
+    public PreviewRepairedState()
+    {
+        super(RepairType.preview_repaired);
+    }
+
+
+    @Override
+    public RepairRunnable getRepairRunnable(String keyspace, List<String> tables, Set<Range<Token>> ranges, boolean primaryRangeOnly)
+    {
+        RepairOption option = new RepairOption(RepairParallelism.PARALLEL, primaryRangeOnly, false, false,
+                                               AutoRepairService.instance.getAutoRepairConfig().getRepairThreads(repairType), ranges,
+                                               !ranges.isEmpty(), false, false, PreviewKind.REPAIRED, false, true, false, false);
+
+        option.getColumnFamilies().addAll(tables);
+
+        return getRepairRunnable(keyspace, option);
+    }
+}
+
 class IncrementalRepairState extends AutoRepairState
 {
     public IncrementalRepairState()
diff --git a/src/java/org/apache/cassandra/repair/autorepair/AutoRepairStateFactory.java b/src/java/org/apache/cassandra/repair/autorepair/AutoRepairStateFactory.java
index 3812ed6e4d..8c3ba373db 100644
--- a/src/java/org/apache/cassandra/repair/autorepair/AutoRepairStateFactory.java
+++ b/src/java/org/apache/cassandra/repair/autorepair/AutoRepairStateFactory.java
@@ -30,6 +30,8 @@ public class AutoRepairStateFactory
                 return new FullRepairState();
             case incremental:
                 return new IncrementalRepairState();
+            case preview_repaired:
+                return new PreviewRepairedState();
         }
 
         throw new IllegalArgumentException("Invalid repair type: " + repairType);
```

```
INFO  [Thread-12] 2024-06-20 14:02:32,047 RepairRunnable.java:324 - Starting repair command #11 (a5a3fffa-2f37-11ef-a8c2-e106f7a22c03), repairing keyspace easy_cass_stress with repair options (parallelism: parallel, primary range: true, incremental: false, job threads: 1, ColumnFamilies: [keyvalue], dataCenters: [], hosts: [], previewKind: REPAIRED, # of ranges: 1, pull r
epair: false, force repair: false, optimise streams: false, ignore unreplicated keyspaces: true, repairPaxos: false, paxosOnly: false)
```

src/java/org/apache/cassandra/tools/nodetool/SetAutoRepairConfig.java
```diff
@@ -0,0 +1,156 @@
...
+            case "enabled":
+                probe.setAutoRepairEnabled(repairType, Boolean.parseBoolean(paramVal));
+                break;
+            case "threads":
+                probe.setRepairThreads(repairType, Integer.parseInt(paramVal));
+                break;
+            case "subranges":
+                probe.setRepairSubRangeNum(repairType, Integer.parseInt(paramVal));
+                break;
+            case "minrepairintervalinhours":
```
tolbertam: One thing I just stumbled into is that the formatting of multi-word config here is different than the yaml configuration, e.g. `minrepairinternvalinhours` in the yaml is `min_repair_internval_in_hours`.  Can we update the nodetool config to match the yaml one?

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
```diff
@@ -0,0 +1,390 @@
...
+    public static class Options implements Serializable
+    {
+        // The separator separating different DCs in repair_dc_groups
+        public static final String DC_GROUP_SEPARATOR = "\\|";
+
+        // defaultOptions defines the default auto-repair behavior when no overrides are defined
+        @VisibleForTesting
+        protected static final Options defaultOptions = getDefaultOptions();
+
+        public Options()
```
tolbertam, line 455: Suggestion: implement `toString()` to return all the options like a map, otherwise `system_views.settings` can't read these:

> auto_repair.repair_type_overrides | {incremental=org.apache.cassandra.repair.autorepair.AutoRepairConfig$Options@64af73a1, preview_repaired=org.apache.cassandra.repair.autorepair.AutoRepairConfig$Options@2ae56cae}

src/java/org/apache/cassandra/repair/autorepair/AutoRepairKeyspace.java
```diff
@@ -0,0 +1,86 @@
...
+                                   .id(TableId.forSystemTable(SchemaConstants.AUTO_REPAIR_KEYSPACE_NAME, name))
+                                   .comment(description)
+                                   .gcGraceSeconds((int) TimeUnit.DAYS.toSeconds(90))
+                                   .build();
+    }
+
+    public static KeyspaceMetadata metadata()
+    {
+        Tables tables = Tables.of(AutoRepairHistory, AutoRepairPriority);
+        return KeyspaceMetadata.create(SchemaConstants.AUTO_REPAIR_KEYSPACE_NAME, KeyspaceParams.simple(1), tables);
```
tolbertam: Should do something similar we did for other system keyspaces in `CassandraRelevantProperties`:

```java
    SYSTEM_AUTH_DEFAULT_RF("cassandra.system_auth.default_rf", "1"),
    SYSTEM_TRACES_DEFAULT_RF("cassandra.system_traces.default_rf", "2"),
    SYSTEM_DISTRIBUTED_DEFAULT_RF("cassandra.system_distributed.default_rf", "3"),
```

which is used like:

```java
KeyspaceParams.simple(Math.max(DEFAULT_RF, DatabaseDescriptor.getDefaultKeyspaceRF()))
```

tolbertam: although, i wonder if we should just leverage `system_distributed` instead of creating a separate keyspace? This may reduce some operator burden.  e.g. in our cluster bring up one of the first things we do is alter these keyspaces to network topology strategy and we also have allowlist/blocklist for system keyspaces that we'd have to update.

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
jaydeepkumar1984: > 👍 thanks for doing this. I imagine it could be useful to be able to configure things more granularly. I have some cases where we want to run full repair on a table, but never incremental (the main use case is analytics based where etl processes are loading a bunch of SSTables every few hours). Will come back to this as I review and experiment more with the PR, if there is a bunch of things that we would want configurable at table level maybe we could consider adding something like a RepairParams, otherwise maybe this is fine as is.

Sure, makes sense. We can finalize all the configuration granularity that we want at a table level and then change it accordingly.

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
jaydeepkumar1984: Sure, we can provide a more granular table-level configuration with multiple configuration knobs.

src/java/org/apache/cassandra/cql3/UntypedResultSet.java
jaydeepkumar1984: > Nit: there are some existing methods (`getInt`, `getUUID`) that are overloaded with an additional `ifNull` parameter, for consistency we should do the same for getLong.

ACK - will update it

src/java/org/apache/cassandra/metrics/AutoRepairMetrics.java
jaydeepkumar1984: > Fantastic to have these out of the gate 👍. We should add some docs to [`metrics.adoc`](https://github.com/apache/cassandra/blob/trunk/doc/modules/cassandra/pages/managing/operating/metrics.adoc) to document these.

ACK. Will extend the documentation

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
jaydeepkumar1984: IMO, by default, we should just enable full repair. For the IR, we should let an operator decide whether they want to enable or not.

src/java/org/apache/cassandra/tools/nodetool/SetAutoRepairConfig.java
jaydeepkumar1984: Sure, will do

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
tolbertam: that seems sensible to me :+1:.  I'm not sure whether or not this would get too tricky, but I could see that the 'defaults' for overrides could be contextual based on the kind of repair, e.g. I would want a more frequent `min_repair_interval_in_hours` for Incremental Repair than for Full Repair.    I think that may be too involved though, I think it's ok to just special case full repair to be enabled by default if you have `auto_repair.enabled: true`

src/java/org/apache/cassandra/repair/autorepair/AutoRepairState.java
```diff
@@ -0,0 +1,349 @@
...
+{
+    public IncrementalRepairState()
+    {
+        super(RepairType.incremental);
+    }
+
+    @Override
+    public RepairRunnable getRepairRunnable(String keyspace, List<String> tables, Set<Range<Token>> ranges, boolean primaryRangeOnly)
+    {
+        RepairOption option = new RepairOption(RepairParallelism.PARALLEL, primaryRangeOnly, true, false,
```
tolbertam, line 333: Discussion point:  We are currently passing `false` for `optimiseStreams`.  This is a nice enhancement in the form of [`CASSANDRA-3200`](https://issues.apache.org/jira/browse/CASSANDRA-3200) which cuts down the amount of overstreaming when inconsistencies are found.

I'm not sure it has a lot of precedence for use, but it's been in tree for a long time.  We've only started to look into rolling it out ourselves.

I think we should consider setting it to true here, or making it configurable somehow.

src/java/org/apache/cassandra/schema/TableParams.java
```diff
@@ -57,7 +57,8 @@ public enum Option
         ADDITIONAL_WRITE_POLICY,
         CRC_CHECK_CHANCE,
         CDC,
-        READ_REPAIR;
+        READ_REPAIR,
+        DISABLED_AUTOMATED_REPAIR;
```
clohfink: NIT: from a naming, having `AUTOMATED_REPAIR: true` vs `DISABLED_AUTOMATED_REPAIR: false` is easier for people to reason. Avoiding the double negatives and extra boolean logic for what people expect.

jaydeepkumar1984: ACK - will update it accordingly

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
```diff
@@ -0,0 +1,390 @@
...
+        public volatile Set<String> ignore_dcs;
+        // Set this 'true' if AutoRepair should repair only the primary ranges owned by this node; else, 'false'
+        // It is the same as -pr in nodetool repair options.
+        public volatile Boolean repair_primary_token_range_only;
+        // By default, the value is empty, which means all nodes are in one ring and the repair in that ring should be done node by node.
+        // Adding this config for special cases where certain keyspace data is only stored in
+        // certain data centers. We can run multiple node repairs at the same time.
+        // For example, if the value is “dc1,dc2|dc3|dc4|dc5,dc6,dc7”, there will be 4 groups {dc1, dc2}, {dc3}, {dc4} and {dc5, dc6, dc7}.
+        // This means we can run repair parallely on 4 nodes, each in one group.
+        public volatile Set<String> repair_dc_groups;
```
clohfink: Since this is a global setting, I am not sure this is safe or necessary. Other keyspaces not in an island configuration to a single DC will not get repaired correctly (e.g., system_distributed, system_auth). For single keyspaces that are not replicated, there are already checks on a node if it is not replicated for a keyspace in checkNodeContainsKeyspaceReplica of AutoRepairUtils. Could simplify things a little bit and get rid of the pid in schema?

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
```diff
@@ -0,0 +1,390 @@
...
+
+import org.apache.cassandra.config.DatabaseDescriptor;
+import org.apache.cassandra.exceptions.ConfigurationException;
+
+public class AutoRepairConfig implements Serializable
+{
+    // enable/disable auto repair globally, overrides all other settings. Cannot be modified dynamically.
+    public final Boolean enabled;
+    // the interval in seconds between checks for eligible repair operations. Cannot be modified dynamically.
+    public final Integer repair_check_interval_in_sec = 300; // 5 minutes
```
clohfink: this is kinda across all of these settings, but if look at Config.java . Can see now after 4.x that we for time settings instead of having _in_sec just pass a duration type. Same goes for using the mib options intead of _mb etc. Should probably stick with this kinda thing for most of the config options

src/java/org/apache/cassandra/metrics/AutoRepairMetrics.java
jaydeepkumar1984: Incorporated

src/java/org/apache/cassandra/repair/autorepair/AutoRepairState.java
jaydeepkumar1984: Ok, changed it to `true`

src/java/org/apache/cassandra/schema/SchemaKeyspace.java
jaydeepkumar1984: I incorporated this comment. Now, we have more granular configuration support as follows:
```
    AND automated_repair_full = {'enabled': 'true'}
    AND automated_repair_incremental = {'enabled': 'true'};

```
Please note: I had to use two separate configurations instead of merging due to the limitation of the CQL grammar, which does not allow nested configuration for DDL statements.
```
    AND automated_repair = {'full': {'enabled': 'true'}, 'incremental': {'enabled': 'true'}}

```

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
jaydeepkumar1984: Done - please see above.

src/java/org/apache/cassandra/cql3/UntypedResultSet.java
jaydeepkumar1984: Done

src/java/org/apache/cassandra/metrics/AutoRepairMetrics.java
jaydeepkumar1984: Done

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
jaydeepkumar1984: Done. I changed the `min_repair_interval_in_hours` to `DurationSpec.IntSecondsBound` so we can now go up to the second level of granularity.

src/java/org/apache/cassandra/repair/autorepair/AutoRepairStateFactory.java
jaydeepkumar1984: Done

src/java/org/apache/cassandra/tools/nodetool/SetAutoRepairConfig.java
jaydeepkumar1984: Done

src/java/org/apache/cassandra/repair/autorepair/AutoRepairKeyspace.java
jaydeepkumar1984: Done. Moved these two tables under `system_distributed` itself.

src/java/org/apache/cassandra/schema/TableParams.java
jaydeepkumar1984: Done

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
jaydeepkumar1984: Done

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
jaydeepkumar1984: Yeah, it is not necessary. I have removed it.

src/java/org/apache/cassandra/repair/autorepair/AutoRepairConfig.java
jaydeepkumar1984: Updated the configs to match with the 4.x by using the `DurationSpec.*`

jaydeepkumar1984:
@tolbertam @clohfink I've incorporated all the review comments; please review the updated PR. Thanks!

src/java/org/apache/cassandra/schema/TableParams.java
```diff
@@ -105,6 +112,13 @@ private TableParams(Builder builder)
         extensions = builder.extensions;
         cdc = builder.cdc;
         readRepair = builder.readRepair;
+        automatedRepair = new HashMap<>()
```
clohfink: this breaks java8 builds, cant use <> on anonymous inner classes

src/java/org/apache/cassandra/schema/TableParams.java
```diff
@@ -84,12 +89,14 @@ public String toString()
     public final boolean cdc;
     public final ReadRepairStrategy readRepair;
 
+    public final Map<AutoRepairConfig.RepairType, AutoRepairParams> automatedRepair;
```
clohfink: Can change to EnumMap

src/java/org/apache/cassandra/schema/TableParams.java
clohfink:
```
        automatedRepair = new EnumMap<>(AutoRepairConfig.RepairType.class);
        automatedRepair.put(AutoRepairConfig.RepairType.full, builder.automatedRepairFull);
        automatedRepair.put(AutoRepairConfig.RepairType.incremental, builder.automatedRepairIncremental);
 ```

File: src/java/org/apache/cassandra/repair/autorepair/AutoRepairUtils.java [hunk 19]
```diff
@@ -0,0 +1,806 @@
...
+                                                                        Lists.newArrayList(ByteBufferUtil.bytes(repairType.toString()),
+                                                                                           ByteBufferUtil.bytes(hostId))), Dispatcher.RequestTime.forImmediateExecution());
+    }
+
+    public static void setForceRepairNewNode(RepairType repairType)
+    {
+        // this function will be called when a node bootstrap finished
+        UUID hostId = Gossiper.instance.getHostId(FBUtilities.getBroadcastAddressAndPort());
+        // insert the data first
+        insertNewRepairHistory(repairType, System.currentTimeMillis(), System.currentTimeMillis());
```
clohfink: fails checkstyles,  use org.apache.cassandra.utils.Clock.Global or org.apache.cassandra.utils.Clock interface

src/java/org/apache/cassandra/repair/autorepair/AutoRepair.java
```diff
@@ -0,0 +1,401 @@
...
+                                ranges.add(new Range<>(childStartToken, childEndToken));
+                                totalProcessedSubRanges++;
+                                if ((totalProcessedSubRanges % config.getRepairThreads(repairType) == 0) ||
+                                    (totalProcessedSubRanges == totalSubRanges))
+                                {
+                                    RepairRunnable task = repairState.getRepairRunnable(keyspaceName,
+                                                                                        config.getRepairByKeyspace(repairType) ? tablesToBeRepaired : ImmutableList.of(tableName),
+                                                                                        ranges, primaryRangeOnly);
+                                    repairState.resetWaitCondition();
+                                    new Thread(NamedThreadFactory.createAnonymousThread(new FutureTask<>(task, null))).start();
```
clohfink: This fails checkstyles requirements. Should create a single thread pool for this or reuse one of existing ones

src/java/org/apache/cassandra/repair/autorepair/AutoRepair.java
```diff
@@ -0,0 +1,401 @@
...
+    }
+
+    // repairAsync runs a repair session of the given type asynchronously.
+    public void repairAsync(AutoRepairConfig.RepairType repairType, long millisToWait)
+    {
+        repairExecutors.get(repairType).submit(() -> repair(repairType, millisToWait));
+    }
+
+    // repair runs a repair session of the given type synchronously.
+    public void repair(AutoRepairConfig.RepairType repairType, long millisToWait)
```
clohfink: NP: maybe simplify this method a little breaking it up, its very long and deep

test/unit/org/apache/cassandra/repair/autorepair/AutoRepairParameterizedTest.java
```diff
@@ -0,0 +1,581 @@
...
+import org.apache.cassandra.schema.TableMetadata;
+import org.apache.cassandra.service.AutoRepairService;
+import org.apache.cassandra.utils.FBUtilities;
+import org.apache.cassandra.utils.progress.ProgressEvent;
+import org.mockito.Mock;
+import org.mockito.Mockito;
+import org.mockito.MockitoAnnotations;
+
+import static org.apache.cassandra.Util.setAutoRepairEnabled;
+import static org.apache.cassandra.Util.token;
```
clohfink: unused import breaks build

test/unit/org/apache/cassandra/repair/autorepair/AutoRepairParameterizedTest.java
```diff
@@ -0,0 +1,581 @@
...
+import org.apache.cassandra.service.StorageService;
+import org.apache.cassandra.utils.Pair;
+
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+import junit.framework.Assert;
```
clohfink: change to org.junit.Assert

clohfink (changes requested):

test/unit/org/apache/cassandra/repair/autorepair/AutoRepairParameterizedTest.java
clohfink: also in AutoRepairTest

src/java/org/apache/cassandra/schema/TableParams.java
jaydeepkumar1984: Incorporated

src/java/org/apache/cassandra/schema/TableParams.java
jaydeepkumar1984: Incorporated

src/java/org/apache/cassandra/repair/autorepair/AutoRepairUtils.java
jaydeepkumar1984: Incorporated

src/java/org/apache/cassandra/repair/autorepair/AutoRepair.java
jaydeepkumar1984: Incorporated

src/java/org/apache/cassandra/repair/autorepair/AutoRepair.java
jaydeepkumar1984: Chopped it a bit - see if this looks better now

test/unit/org/apache/cassandra/repair/autorepair/AutoRepairParameterizedTest.java
jaydeepkumar1984: Incorporated

test/unit/org/apache/cassandra/repair/autorepair/AutoRepairParameterizedTest.java
jaydeepkumar1984: Incorporated

test/unit/org/apache/cassandra/repair/autorepair/AutoRepairParameterizedTest.java
```diff
@@ -0,0 +1,681 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Sets;
+
+import org.apache.cassandra.config.DurationSpec;
```
clohfink, line 36: unused import breaks build with checkstyles

jaydeepkumar1984: Sorry, I missed this again. I forgot to run this against Java8. I've just fixed everything and pushed the changes - it now works with Java8. Could you please retry?
Next time, I will make sure to compile it with Java8 before pushing the newer changes.

src/java/org/apache/cassandra/repair/autorepair/AutoRepair.java
```diff
@@ -0,0 +1,457 @@
...
+                                totalProcessedSubRanges++;
+                                if ((totalProcessedSubRanges % config.getRepairThreads(repairType) == 0) ||
+                                    (totalProcessedSubRanges == totalSubRanges))
+                                {
+                                    int retryCount = 0;
+                                    Future<?> f = null;
+                                    while (retryCount <= config.getRepairMaxRetries())
+                                    {
+                                        RepairRunnable task = repairState.getRepairRunnable(keyspaceName,
+                                                                                            config.getRepairByKeyspace(repairType) ? tablesToBeRepaired : ImmutableList.of(tableName),
```
tolbertam: Since this is being evaluated for each table, won't this mean that for each Table in a keyspace, it will issue a full keyspace repair?

tolbertam: Ah nevermind, I see there is a break at line 340 that breaks out after evaluating over the first iteration.

src/java/org/apache/cassandra/repair/autorepair/AutoRepairKeyspace.java
```diff
@@ -0,0 +1,84 @@
...
+    private AutoRepairKeyspace()
+    {
+    }
+
+    /**
+     * Generation is used as a timestamp for automatic table creation on startup.
+     * If you make any changes to the tables below, make sure to increment the
+     * generation and document your change here.
+     */
+    public static final long GENERATION = 0;
```
clohfink: I think this should just be folded into SystemDistributedKeyspace, since the metadata() returns that but missing the correct defaults from the system distributed it might cause some weirdness, also having the GENERATION number 0 when being merged into the system_distributed. I think would be better to hve these tables in the keyspace that owns them

src/java/org/apache/cassandra/repair/autorepair/AutoRepairState.java
```diff
@@ -0,0 +1,351 @@
...
+import static org.apache.cassandra.utils.Clock.Global.currentTimeMillis;
+import static org.apache.cassandra.utils.concurrent.Condition.newOneTimeCondition;
+
+// AutoRepairState represents the state of automated repair for a given repair type.
+public abstract class AutoRepairState implements ProgressListener
+{
+    protected static final Logger logger = LoggerFactory.getLogger(AutoRepairState.class);
+    private final SimpleDateFormat format = new SimpleDateFormat("yyyy-MM-dd HH:mm:ss,SSS");
+    @VisibleForTesting
+    protected static Supplier<Long> timeFunc = System::currentTimeMillis;
```
clohfink: org.apache.cassandra.utils.Clock was made to handle this kinda thing - theres normally a checkstyles thing that would fail on this method call but because its a lambda reference it misses it

src/java/org/apache/cassandra/tools/nodetool/GetAutoRepairConfig.java
```diff
@@ -0,0 +1,81 @@
...
+        sb.append("\n\tpriority hosts: " + Joiner.on(',').skipNulls().join(probe.getRepairPriorityForHosts(repairType)));
+        sb.append("\n\tsstable count higher threshold: " + config.getRepairSSTableCountHigherThreshold(repairType));
+        sb.append("\n\ttable max repair time in sec: " + config.getAutoRepairTableMaxRepairTime(repairType));
+        sb.append("\n\tignore datacenters: " + Joiner.on(',').skipNulls().join(config.getIgnoreDCs(repairType)));
+        sb.append("\n\trepair primary token-range: " + config.getRepairPrimaryTokenRangeOnly(repairType));
+        sb.append("\n\tnumber of parallel repairs within group: " + config.getParallelRepairCount(repairType));
+        sb.append("\n\tpercentage of parallel repairs within group: " + config.getParallelRepairPercentage(repairType));
+        sb.append("\n\tmv repair enabled: " + config.getMVRepairEnabled(repairType));
+        sb.append("\n\tinitial scheduler delay: " + config.getInitialSchedulerDelay(repairType));
+        sb.append("\n\trepair setssion timeout: " + config.getRepairSessionTimeout(repairType));
```
clohfink: spelling setssion -> session

test/unit/org/apache/cassandra/cql3/statements/DescribeStatementTest.java
```diff
@@ -1004,7 +1004,9 @@ private static String tableParametersCql()
                "    AND memtable_flush_period_in_ms = 0\n" +
                "    AND min_index_interval = 128\n" +
                "    AND read_repair = 'BLOCKING'\n" +
-               "    AND speculative_retry = '99p';";
+               "    AND speculative_retry = '99p'\n" +
+               "    AND automated_repair_full = {'enabled': 'true'}\n" +
```
clohfink: np: should we maybe nix the "automated" prefix?

src/java/org/apache/cassandra/repair/autorepair/AutoRepairKeyspace.java
jaydeepkumar1984: Incorporated

src/java/org/apache/cassandra/repair/autorepair/AutoRepairState.java
jaydeepkumar1984: Incorporated

src/java/org/apache/cassandra/tools/nodetool/GetAutoRepairConfig.java
jaydeepkumar1984: Incorporated

test/unit/org/apache/cassandra/cql3/statements/DescribeStatementTest.java
jaydeepkumar1984: Incorporated

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -346,6 +346,17 @@ public enum CassandraRelevantProperties
      */
     GOSSIP_SETTLE_POLL_SUCCESSES_REQUIRED("cassandra.gossip_settle_poll_success_required", "3"),
 
+    /**
+     * If set to true, mutations streamed during anti-entropy repair will be replayed via the regular write path for associated views.
+     */
+    STREAMING_REQUIRES_VIEW_BUILD_DURING_REPAIR("cassandra.streaming.requires_view_build_during_repair", "true"),
```
tolbertam: This currently prevents startup if trying to configure an incremental repair schedule:

```
ERROR [main] 2024-11-03 12:48:38,892 CassandraDaemon.java:900 - Fatal configuration error
org.apache.cassandra.exceptions.ConfigurationException: Cannot run incremental repair while materialized view replay is enabled.
        at org.apache.cassandra.service.AutoRepairService.checkCanRun(AutoRepairService.java:69)
        at org.apache.cassandra.repair.autorepair.AutoRepair.setup(AutoRepair.java:128)
        at org.apache.cassandra.service.StorageService.doAutoRepairSetup(StorageService.java:1355)
        at org.apache.cassandra.service.StorageService.finishJoiningRing(StorageService.java:1326)
        at org.apache.cassandra.service.StorageService.joinTokenRing(StorageService.java:1214)
        at org.apache.cassandra.service.StorageService.joinTokenRing(StorageService.java:1150)
        at org.apache.cassandra.service.StorageService.initServer(StorageService.java:941)
        at org.apache.cassandra.service.StorageService.initServer(StorageService.java:859)
        at org.apache.cassandra.service.CassandraDaemon.setup(CassandraDaemon.java:421)
        at org.apache.cassandra.service.CassandraDaemon.activate(CassandraDaemon.java:744)
        at org.apache.cassandra.service.CassandraDaemon.main(CassandraDaemon.java:878)
```

It seems a bit unnecessary to have to set some system properties to run an incremental repair schedule if we aren't using mvs or cdc, was wondering if we could make this smarter.  Currently both materialized views and cdc have guardrail properties, `materialized_views_enabled` and `cdc_enabled` which both default to false.  Can we involve these properties somehow?  If they are not set, then we don't need to consider either during streaming.

jaydeepkumar1984: Ok, I have made the following two changes:
1. Backported https://issues.apache.org/jira/browse/CASSANDRA-17666
2. Implemented a similar flag to https://issues.apache.org/jira/browse/CASSANDRA-17666 for MV

tl;dr - I've removed these properties instead it is now controlled by two yaml flags

```
cdc_on_repair_enabled: false
materialized_views_on_repair_enabled: false
```

The above two flags are `true` by default, so one needs to disable them. The changes have been pushed to the same PR; please take a look at it.

PR #3374, opened 2024-06-13
CASSANDRA-18111 snapshotting centralisation

smiklosovic:
 
```

```


src/java/org/apache/cassandra/service/snapshot/SnapshotWatcher.java
```diff
@@ -0,0 +1,186 @@
...
+                    for (WatchEvent<?> event : key.pollEvents())
+                    {
+                        if (event.kind() == ENTRY_DELETE)
+                        {
+                            Path dir = watchKeyPathMap.get(key);
+
+                            if (dir == null)
+                                continue;
+
+                            snapshotManager.findSnapshot(dir).ifPresent(snapshotManager::clearSnapshot);
```
smiklosovic: This should not block and clearing of a snapshot should be done asynchronously to process events as fast as possible as doing it by a blocking manner might cause (in theory) the overflowing of deletion events under absolutely extreme conditions but the probability this happens in the environment we use this API is basically equal to zero.

Anyway, to be extra sure, we should just populate a queue of snapshots to be cleared and then a thread in SnapshotManager should just go over it, similarly as expired snapshots are cleared.

tools/stress/src/org/apache/cassandra/stress/CompactionStress.java
```diff
@@ -340,7 +340,7 @@ public void run()
             }
 
             double currentSizeGiB;
-            while ((currentSizeGiB = directories.getRawDiretoriesSize() / BYTES_IN_GIB) < totalSizeGiB)
+            while ((currentSizeGiB = directories.getRawDirectoriesSize() / BYTES_IN_GIB) < totalSizeGiB)
```
frankgh, line 343: good catch

src/java/org/apache/cassandra/service/SnapshotVerbHandler.java
```diff
@@ -44,7 +45,7 @@ public void doVerb(Message<SnapshotCommand> message)
         SnapshotCommand command = message.payload;
         if (command.clear_snapshot)
         {
-            StorageService.instance.clearSnapshot(command.snapshot_name, command.keyspace);
+            SnapshotManager.instance.clearSnapshot(command.keyspace, command.column_family, command.snapshot_name);
```
frankgh: This is a behavior change, not sure if the original intent is correct or not. But the previous behavior deletes all snapshots with the given `snapshot_name` in the `keyspace` for __all__ tables, whereas now you only clear snapshots for a single table.

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -2368,19 +2289,25 @@ public boolean snapshotExists(String snapshotName)
      */
     public void clearSnapshot(String snapshotName)
     {
-        RateLimiter clearSnapshotRateLimiter = DatabaseDescriptor.getSnapshotRateLimiter();
-
-        List<File> snapshotDirs = getDirectories().getCFDirectories();
-        Directories.clearSnapshot(snapshotName, snapshotDirs, clearSnapshotRateLimiter);
+        SnapshotManager.instance.clearSnapshots(snapshotName,
```
frankgh: hmm, before the change calling clear snapshots here is restricted to the specific CFS table directories. With the change, it seems you may end up clearing snapshots for other tables other than the current table.

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -3302,7 +3229,7 @@ public double getDroppableTombstoneRatio()
     @Override
     public long trueSnapshotsSize()
     {
-        return getDirectories().trueSnapshotsSize();
+        return SnapshotManager.instance.trueSnapshotsSize(metadata(), directories);
```
frankgh: NIT: we are using `getDirectories()` elsewhere
```suggestion
        return SnapshotManager.instance.trueSnapshotsSize(metadata(), getDirectories());
```

frankgh: I'm trying this out, and it is returning true size of 0 for me. We build a `SSTableSizeSummer` like this:

```
        SSTableSizeSummer visitor = new SSTableSizeSummer(metadata, directories
                                                                    .sstableLister(Directories.OnTxnErr.THROW)
                                                                    .listFiles());
```

the second parameter is a list of files to skip. Basically we end up skipping every sstable component file from the summer (adder is a better name?) . And we end up with no files to consider for the true size. Is this behavior desired/correct?

Maybe some unit tests are missing here to validate the expected behavior at the CFS level?

frankgh (commented):
This is a work in progress review. I will add more comments in the next couple of days if anything else catches my eye

src/java/org/apache/cassandra/service/SnapshotVerbHandler.java
smiklosovic: Aha interesting, yeah I can see the difference, we should copy that.

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
smiklosovic: correct.

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
smiklosovic: I ll get back to you with the feedback, not sure what is going on here yet as I look at this.

test/unit/org/apache/cassandra/service/snapshot/TableSnapshotTest.java
```diff
@@ -298,7 +298,7 @@ public void testShouldClearSnapshot() throws Exception
                 // 1. snapshot to clear is not ephemeral
                 // 2. tag to clear is null, empty, or it is equal to snapshot tag
                 // 3. byTimestamp is true
-                if (TableSnapshot.shouldClearSnapshot(testingTag, olderThanTimestamp).test(snapshot))
+                if (SnapshotManager.shouldClearSnapshot(testingTag, Set.of(keyspace), olderThanTimestamp, false).test(snapshot))
```
frankgh: this is another behavior change in the test and I'm not sure if it's desired. Should we avoid including the keyspace? The comments above seem to indicate we don't want to include the keyspace here
```suggestion
                if (SnapshotManager.shouldClearSnapshot(testingTag, Set.of(), olderThanTimestamp, false).test(snapshot))
```

test/distributed/org/apache/cassandra/distributed/impl/Instance.java
```diff
@@ -632,7 +632,10 @@ public void startup(ICluster cluster)
                 {
                     partialStartup(cluster);
                 }
-                StorageService.instance.startSnapshotManager();
+                SnapshotManager.instance.start();
```
frankgh: should we mirror what CassandraDaemon is doing here and start the snapshot manager right after the commit log starts? it should be `org.apache.cassandra.service.CassandraDaemon#setup`

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
```diff
@@ -138,11 +152,21 @@ public boolean isExpiring()
 
     public long computeSizeOnDiskBytes()
     {
-        return snapshotDirs.stream().mapToLong(FileUtils::folderSize).sum();
+        long sum = sizeOnDisk;
+        if (sum == 0)
+        {
+            for (File snapshotDir : snapshotDirs)
+                sizeOnDisk = sum += FileUtils.folderSize(snapshotDir);
+        }
```
frankgh: I'm worried about races here. What if we have two concurrent requests to the `Table Snapshot` virtual table. 2 threads will see sum == 0 and compute the sizeOnDisk twice. Should `sizeOnDisk` potentially be volatile as well? Maybe a way to prevent two threads increasing this value independently is to refactor this code as this:

```suggestion
        long sum = sizeOnDisk;
        if (sum == 0)
        if (sum == 0)
        {
            for (File snapshotDir : snapshotDirs)
                sum += FileUtils.folderSize(snapshotDir);
            sizeOnDisk = sum;
        }
```

With the only potential side effect being that two threads do the same work.

frankgh: Another potential issue is that 1 thread starts updating the `sizeOnDisk`. The second threads sees a partial, but incorrect value of the sizeOnDisk

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2969,77 +2958,49 @@ private Keyspace getValidKeyspace(String keyspaceName)
...
-        {
-            for (String keyspaceDir : new File(dataDir).tryListNames())
-            {
-                // Only add a ks if it has been specified as a param, assuming params were actually provided.
-                if (keyspaceNames.length > 0 && !Arrays.asList(keyspaceNames).contains(keyspaceDir))
-                    continue;
-                keyspaces.add(keyspaceDir);
-            }
-        }
+        Set<String> keyspaces = new HashSet<>(Arrays.asList(keyspaceNames));
```
frankgh: I think you can do this now in java 11? Are we bringing this to 4.1 or 4.0 and are trying to make it java 8 compatible?
```suggestion
        Set<String> keyspaces = Set.of(keyspaceNames);
```

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2969,77 +2958,49 @@ private Keyspace getValidKeyspace(String keyspaceName)
...
+        long maxCreatedAt = getMaxSnapshotCreatedAt(options);
+
+        SnapshotManager.instance.clearSnapshots(tag, keyspaces, maxCreatedAt);
 
+        if (logger.isDebugEnabled())
+            logger.debug("Cleared out snapshot directories");
```
frankgh: maybe an opportunity to improve this log message:
```suggestion
            logger.debug("Cleared out snapshot directories tag={} keyspaces={} maxCreatedAt={}", tag, keyspaceNames, maxCreatedAt);
```

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2969,77 +2958,49 @@ private Keyspace getValidKeyspace(String keyspaceName)
...
+    }
+
+    private static long getMaxSnapshotCreatedAt(Map<String, Object> options)
+    {
         Object olderThan = options.get("older_than");
         Object olderThanTimestamp = options.get("older_than_timestamp");
 
-        final long clearOlderThanTimestamp;
+        long maxCreatedAt = Clock.Global.currentTimeMillis();
```
frankgh: This also changes the behavior. If no option is provided we are using `0` before. Now, we are using `Clock.Global.currentTimeMillis()`

test/unit/org/apache/cassandra/db/DirectoriesTest.java
```diff
@@ -350,40 +364,13 @@ public void testListSnapshots() throws Exception {
...
-        snapshots = directories.listSnapshots();
+        snapshots = listSnapshots(directories);
         assertThat(snapshots.keySet()).isEqualTo(Sets.newHashSet(SNAPSHOT2, SNAPSHOT3));
         assertThat(snapshots.get(SNAPSHOT2)).isEqualTo(snapshot2.asTableSnapshot());
         assertThat(snapshots.get(SNAPSHOT3)).isEqualTo(snapshot3.asTableSnapshot());
         assertThat(snapshots.get(SNAPSHOT3).isEphemeral()).isTrue();
     }
 
-    @Test
-    public void testListSnapshotDirsByTag() throws Exception {
```
frankgh, line 361: do we lose this test?

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -1169,37 +1169,6 @@ private BiPredicate<File, FileType> getFilter(boolean includeForeignTables)
         }
     }
 
-    public Map<String, TableSnapshot> listSnapshots()
```
frankgh, line 1174: this looks like it's only used in test code

src/java/org/apache/cassandra/db/Directories.java
```diff
@@ -1228,100 +1197,10 @@ protected static SnapshotManifest maybeLoadManifest(String keyspace, String tabl
         return null;
     }
 
-    @VisibleForTesting
-    protected Map<String, Set<File>> listSnapshotDirsByTag()
```
frankgh, line 1234: also only used in test code from what I can tell

test/unit/org/apache/cassandra/db/DirectoriesTest.java
frankgh: it looks like `listSnapshotDirsByTag()` was not used in production code, so makes sense to remove it

src/java/org/apache/cassandra/service/snapshot/SnapshotWatcher.java
```diff
@@ -0,0 +1,216 @@
...
+            return;
+
+        Collection<File> directories = snapshot.getDirectories();
+        if (directories == null)
+            return;
+
+        for (File snapshotDir : directories)
+        {
+            Path rootSnapshotsDir = snapshotDir.parent().toPath();
+            if (!watchKeyPathMap.containsValue(rootSnapshotsDir))
```
frankgh: contains value is expensive, can we do something else rather?

frankgh (commented):
Added some additional comments

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
smiklosovic: @frankgh 

Yes, this is intended behavior. The way it works is that there are two "sizes" of a snapshot. "true allocated size" of a snapshot in a snapshot directory is the size of all files which are located in the snapshot dir BUT THEY ARE NOT IN HOT DATA DIR.

In other words, because a snapshot is just a collection of hardlinks, then if you take a snapshot then its size is ... 0. Why? Because every file in a snapshot has its counterpart in a hot / live data dir.

But when some SSTables in hot data dir are compacted away, for example, then the hardlink / file in snapshot dir suddenly becomes "real" and it starts to be counted towards true snapshot size ...

If I have a data dir with 10 GiB of data and I take a snapshot, the data don't occupy 20GiB, they still occupy 10GiB. So "true snapshot size" is effectively 0.

test/unit/org/apache/cassandra/service/snapshot/TableSnapshotTest.java
smiklosovic: I think this is just fine. `shouldClearSnapshot` method indeed accepts keyspace now, but the logic in it takes into account the situation when a keyspaces set is empty, preserving the original behaviour.

src/java/org/apache/cassandra/service/StorageService.java
smiklosovic: To be honest I think this is what Paulo did and I just took his stuff he had in a branch and applied it here. I need to check this ...

smiklosovic: @frankgh 

So ... this is all fine. Because if you check how it was in TableSnapshot (1), then we are filtering on time only when it is bigger than 0.

Now the logic is that we put there the current time, indeed, but we do this:

                Instant createdAt = ts.getCreatedAt();
                if (createdAt != null)
                    byTimestamp = createdAt.isBefore(Instant.ofEpochMilli(olderThanTimestamp));


But here, `byTimestamp` will be `true`, because, presumably, `createdAt` will be always before the current time. We cant create a snapshot in the future so `byTimestamp` would be evaluated to false. So logically the behaviour is same. 

I would keep the check on olderThanTimestamp > 0 in place though anyway. We do not want to trigger this when somebody forcibly puts there `olderThanTimestamp` to be 0 or negative ... doesnt make sense.

(1) https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java#L341-L346

src/java/org/apache/cassandra/db/Directories.java
smiklosovic: yeah that's why I deleted them.

src/java/org/apache/cassandra/db/Directories.java
smiklosovic: we should be fine to delete it then, no?

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
smiklosovic: How can the race happen querying the vtable when the version of `getSnapshots` which is eventually called in SnapshotManager is `synchronized`?

smiklosovic: And yes, we can make it `volatile`.

smiklosovic: BTW we hit this recently in HintsDescriptor too. That variable is volatile but there we do not have any loop so the fact that two threads compute it twice does not matter ... The probability is also super low when it is just effectively "cached".

This makes the most sense to me, when we make `SizeOnDisk` to be `volatile`.

    public long computeSizeOnDiskBytes()
    {
        long sum = sizeOnDisk;
        if (sum == 0)
        {
            for (File snapshotDir : snapshotDirs)
                sum += FileUtils.folderSize(snapshotDir);

            sizeOnDisk = sum;
        }

        return sum;
    }

(1) https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/hints/HintsDescriptor.java#L233-L239

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
frankgh: makes sense now, thanks for the clarification.

frankgh (approved):
Took a second look. Looks good to me

src/java/org/apache/cassandra/service/snapshot/SnapshotWatcher.java
smiklosovic: this is just fine, the map contains as many entries as you have data dirs. This is by default 1, if you have three data dirs, that map contains three entries.

src/java/org/apache/cassandra/service/snapshot/SnapshotManager.java
```diff
@@ -17,156 +17,830 @@
...
+            }
+        }
+
+        Instant creationTime = now();
+
+        for (Map.Entry<Keyspace, Set<ColumnFamilyStore>> entry : entitiesForSnapshot.entrySet())
+        {
+            Keyspace keyspace = entry.getKey();
+            for (ColumnFamilyStore table : entry.getValue())
+                keyspace.snapshot(tag, table.getTableName(), skipFlush, ttl, snapshotRateLimiter, creationTime);
```
pauloricardomg: We already have all the `ColumnFamilyStore` objects we want to snapshot here. Why can't we bypass `Keyspace` altogether and just call `ColumnFamilyStore.snapshot` for each snapshot in the `entitiesForSnapshot` map ?

I think this would allow removing or simplifying most snapshot logic from [Keyspace](https://github.com/apache/cassandra/blob/fe025c7f79e76d99e0db347518a7872fd4a114bc/src/java/org/apache/cassandra/db/Keyspace.java#L239) and  ColumnFamilyStore which is ultimately calling [SnapshotManager.instance.createSnapshot](https://github.com/instaclustr/cassandra/blob/CASSANDRA-18111/src/java/org/apache/cassandra/db/ColumnFamilyStore.java#L2138).

Lmk if this makes sense.

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -2368,19 +2224,63 @@ public boolean snapshotExists(String snapshotName)
...
+        Set<TableSnapshot> snapshots = new SnapshotLoader(getDirectories()).loadSnapshots();
+        Map<String, TableSnapshot> tagSnapshotsMap = new HashMap<>();
+
+        for (TableSnapshot snapshot : snapshots)
+            tagSnapshotsMap.put(snapshot.getTag(), snapshot);
+
+        return tagSnapshotsMap;
+    }
+
+    public Refs<SSTableReader> getSnapshotSSTableReaders(String tag) throws IOException
```
pauloricardomg: Can this be moved to `TableSnapshot` ? I don't think `ColumnFamilyStore` needs to be involved in how snapshot sstables are read.

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
```diff
@@ -490,6 +490,12 @@ public enum CassandraRelevantProperties
     SNAPSHOT_CLEANUP_INITIAL_DELAY_SECONDS("cassandra.snapshot.ttl_cleanup_initial_delay_seconds", "5"),
     /** snapshots ttl cleanup period in seconds */
     SNAPSHOT_CLEANUP_PERIOD_SECONDS("cassandra.snapshot.ttl_cleanup_period_seconds", "60"),
+    /**
+     * When there is a snapshot with old / basic format (basically pre-CASSANDRA-16789),
+     * it will enrich it with more metadata upon snapshot's loading at startup.
+     * Defaults to true, when set to false, no enriching will be done.
+     * */
+    SNAPSHOT_MANIFEST_ENRICH_ENABLED("cassandra.snapshot.enrich.enabled", "true"),
```
pauloricardomg: Since this is an invasive refactoring/cleanup patch I'd prefer to leave this behavior change to another ticket, unless this is directly related to caching and centralizing snapshots in SnapshotManager (CASSANDRA-18111).

src/java/org/apache/cassandra/service/snapshot/SnapshotManager.java
```diff
@@ -17,156 +17,830 @@
...
+     * @param cfs          column family to create a snapshot for
+     * @param tag          name of snapshot
+     * @param ephemeral    true if the snapshot is ephemeral, false otherwise
+     * @param ttl          time after the created snapshot will be removed
+     * @param creationTime time the snapshot was created
+     * @param rateLimiter  limiter for hard-links creation, if null, limiter from DatabaseDescriptor will be used
+     * @return logical representation of a snapshot
      */
-    public synchronized void clearSnapshot(TableSnapshot snapshot)
+    public TableSnapshot createSnapshot(ColumnFamilyStore cfs,
```
pauloricardomg: One of the goals of centralizing snapshot logic into `SnapshotManager` was to simplify things but this snapshot creation logic is getting complex enough that probably deserves its own class (ie. `CreateSnapshotTask`) that receives a set of snapshot creation options (ie. tag/ttl/etc) and returns a `Set<TableSnapshot>` that is then added to the set of `liveSnapshots`. 

This would remove the coupling of `SnapshotManager` with `ColumnFamilyStore` since all the handling of snapshot creation would be encapsulated in the `CreateSnapshotTask`.

This would make `SnapshotManager` be responsible for the management of snapshot lifecycle (similar to `CompactionManager`) while the actual snapshot creation task would be encapsulated in its own `CreateSnapshotTask` class (similar to `AbstractCompactionTask`). This is analogous to encapsulating snapshot listing logic into its own `SnapshotLoader` class to make `SnapshotManager` leaner.

With this the implementation of `SnapshotManagerMBean.takeSnapshot` would be something like:
```java
    synchronized takeSnapshot(String tag, Map<String, String> options, String... entities)
   {
          CreateSnapshotTask task = new CreateSnapshotTask(tag, entitites, snapshotRateLimiter);
          Set<TableSnapshot> snapshot = task.executeBlocking()
          liveSnapshots.addAll(snapshot)
   } 
```

Do you think this makes sense ?

src/java/org/apache/cassandra/db/ColumnFamilyStore.java
smiklosovic: I think this is just shuffling stuff around but if you insist ... putting that to TableSnapshot would need to be a static method.

src/java/org/apache/cassandra/config/CassandraRelevantProperties.java
smiklosovic: changed default to false

src/java/org/apache/cassandra/service/snapshot/SnapshotManager.java
smiklosovic: it is better to base it on a builder pattern because I hit a situation when I had "telescopic constructors" and creating a builder for that was just more comfortable.

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,880 @@
...
+        stateful()
+        .withExamples(1)
+        .withSteps(1000)
+        .check(commands(() -> State::new, state -> new SnapshotsSut(state).start())
+               .add(10, new CreateKeyspaceCommand())
+               .add(5, new DropKeyspaceCommand())
+               .add(40, new CreateTableCommand())
+               .add(15, new DropTableCommand())
+               .add(500, new TakeSnapshotCommand())
+               .add(50, new ClearSnapshotCommand())
```
dcapwell: personally I would drop the weights, let every time the test run we try different weights; far more likely to see bugs in areas you didn't think about than the ones you did

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,880 @@
...
+public class SnapshotsTest
+{
+    @Test
+    public void fuzzySnapshotsTest()
+    {
+        stateful()
+        .withExamples(1)
+        .withSteps(1000)
+        .check(commands(() -> State::new, state -> new SnapshotsSut(state).start())
+               .add(10, new CreateKeyspaceCommand())
```
dcapwell: there is `addIf` that would let you block creating the keyspace if there are too many

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,880 @@
...
+               .add(5, new DropKeyspaceCommand())
```
dcapwell: same, do you want to do drop keyspace when there are no keyspaces?  `addIf` would let you ignore that case

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,880 @@
...
+               .add(40, new CreateTableCommand())
+               .add(15, new DropTableCommand())
+               .add(500, new TakeSnapshotCommand())
+               .add(50, new ClearSnapshotCommand())
+               .destroyState(State::destroy)
+               .destroySut(SnapshotsSut::stop)
+               .build());
+    }
+
+    public static class State implements Serializable
```
dcapwell: why `implements Serializable`?  you shouldn't need

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,880 @@
...
+                   Objects.equals(snapshots, state.snapshots) &&
+                   Objects.equals(droppedSnapshots, state.droppedSnapshots) &&
+                   Objects.equals(truncatedSnapshots, state.truncatedSnapshots);
+        }
+
+        @Override
+        public int hashCode()
+        {
+            return Objects.hash(schema, snapshots, droppedSnapshots, truncatedSnapshots);
+        }
```
dcapwell: don't think you need, so this is dead code?

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,880 @@
...
+        {
+            return "DropTableCommand{" +
+                   "state=" + state +
+                   '}';
+        }
+    }
+
+    public static class CreateKeyspaceCommand implements Command<State, SnapshotsSut, State>
+    {
+        private State state;
```
dcapwell: this is a dangerous pattern... the API doesn't give `Sut` the `State` as we should be able to execute directly on the Sut...

The common pattern is to define the state needed for this work, which looks like `state.addRandomKeyspace`.  In your case it looks like you should pass in to the constructor a `String ks` and use that to create the ks in the `State` and the `Sut`

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,880 @@
...
+
+            return maybeTooLongString.substring(0, 48);
+        }
+    }
+
+    public static class SnapshotsSut implements Serializable
+    {
+        private final Cluster.Builder builder = Cluster.build(1).withConfig(c -> c.with(Feature.values()));
+        private Cluster cluster;
+        public State state;
```
dcapwell: this is a bad pattern, as `Sut` should be independent from `State`.  For your test I would honestly just drop `Sut` and merge this class into `State`.

If you look at `TopologyMixupTestBase` we do that there as well

```
statefulBuilder.check(commands(this::stateGen)
```

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,880 @@
...
+        private String trimIfNecessary(String maybeTooLongString)
+        {
+            if (maybeTooLongString.length() <= 48)
+                return maybeTooLongString;
+
+            return maybeTooLongString.substring(0, 48);
+        }
+    }
+
+    public static class SnapshotsSut implements Serializable
```
dcapwell: I would merge this class with `State` and simplify this test to only have a `State` and no `Sut`.  this would simplify a lot of the test logic, and it's more the pattern you have anyways

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,880 @@
...
+
+        public Optional<String> pickRandomKeyspace()
+        {
+            if (schema.keySet().isEmpty())
+                return Optional.empty();
+            else
+                return Optional.of(rs.pick(schema.keySet()));
+        }
+
+        public Pair<String, String> pickOrCreateRandomTable()
```
dcapwell: I feel that this is a bad pattern.  If there are no tables you can't take a snapshot, but your history you publish won't show you did this so the error showing what happened becomes partially meaningless and hides details.

You should *select* what table you are going to take a snapshot of before creating `TakeSnapshotCommand`, and you shouldn't be taking a snapshot when there are no tables

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,880 @@
...
+
+public class SnapshotsTest
+{
+    @Test
+    public void fuzzySnapshotsTest()
+    {
+        stateful()
+        .withExamples(1)
+        .withSteps(1000)
+        .check(commands(() -> State::new, state -> new SnapshotsSut(state).start())
```
dcapwell: said in other places, but the code would be simplifier if you merge `State` and `SnapshotsSut`.  The 2 classes reach into each other and know about it each other, but this just bloated the code size... merging will shrink the amount of code without changing any of the logic.


`State` vs `Sut` more boils down to when you have 2 different objects that have state you want to track.  So the example of a `Map` and a `RangeSearchMap` come up as the `Map` is the model of what *should* be in the `RangeSearchMap` and the `Sut` is the `RangeSearchMap`.  When you are dealing with `Cluster` or `CQLTester` you really don't have a good `Sut` so it makes sense to just drop that in favor of `State` only.

If you look at `TopologyMixupTestBase` we don't have a `Sut` and only have a `State`, this test is more in-line with that class

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
dcapwell:
```suggestion
               .add((rs, state) -> new CreateKeyspaceCommand(state.keyspaceName.next(rs), state.nodeGen.nextInt(rs)))
```

its best practice to define your command state *before* the command is created, this makes the name easier, and also makes the `apply` logic cleaner.  

When you merge everything into the run method then you have to start to save state into this object and you *hope* that the framework calls you in the same order (its free to call `run` before `apply`).

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,880 @@
...
+        @Override
+        public String toString()
+        {
+            return "DropTableCommand{" +
+                   "state=" + state +
+                   '}';
+        }
+    }
+
+    public static class CreateKeyspaceCommand implements Command<State, SnapshotsSut, State>
```
dcapwell: spoke in slack, but this should be rewritten as

```
    public static class CreateKeyspaceCommand implements Property.StateOnlyCommand<State>
    {
        private static final String CQL = "CREATE KEYSPACE %s WITH replication = { 'class' : 'SimpleStrategy', 'replication_factor' : '1' }";
        private final String ks;
        private final int node;

        public CreateKeyspaceCommand(String ks, int node)
        {
            this.ks = ks;
            this.node = node;
        }

        @Override
        public String detailed(State state)
        {
            return "Create Keyspace " + ks + " on node" + node;
        }

        @Override
        public void applyUnit(State state)
        {
            state.addKeyspace(ks);
            state.cluster.get(node).executeInternal(String.format(CQL, ks));
        }
    }
```

The reason is as follows

1) figuring out what the command state should be done *before* creating the command, this is to simplify the `details` message as you can say what is "going" to happen... we create the details string *before* we call `apply`
2) it makes the `apply` logic far simpler as its only action on one thing
3) when you actually have a real `Sut` it makes it easier to define logic to work with the `Sut` that doesn't even know about the `State`

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,1145 @@
...
+        {
+            Set<String> existingKeyspaces = state.getKeyspaces();
+            Set<String> keyspacesToBe = state.schema.keySet();
+
+            Set<String> difference = difference(keyspacesToBe, existingKeyspaces);
+
+            for (String keyspaceToCreate : difference)
+                state.getNode().executeInternal(format(CREATE_KEYSPACE_QUERY, keyspaceToCreate));
+
+            return difference;
```
dcapwell: why all this complexity?  you know what keyspace to create, why do we need set diff to get involved?

in the constrictor we can just save `String ks = state.addRandomKeyspace()` then apply just does 

```
state.getNode().executeInternal(format(CREATE_KEYSPACE_QUERY, ks));
```

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,1145 @@
...
+        {
+            super(rs, state);
+            keyspaceToCreate.add(state.addRandomKeyspace());
+        }
+
+        @Override
+        public void applyUnit(State state)
+        {
+            createKeypaces(state);
+            assertEquals(state.schema.keySet(), state.getKeyspaces());
```
dcapwell, line 292: why this assert?  you are checking if your model has your model?  seems redundant?

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
```diff
@@ -0,0 +1,1145 @@
...
+
+            // dropping of a keyspace will create dropped tables which create "dropped" snapshots
+            Optional<Pair<String, List<String>>> maybeDroppedKeyspaceWithTables = state.removeRandomKeyspace();
+            if (maybeDroppedKeyspaceWithTables.isPresent())
+            {
+                String keyspace = maybeDroppedKeyspaceWithTables.get().left;
+                List<String> tables = maybeDroppedKeyspaceWithTables.get().right;
+
+                for (String table : tables)
+                    state.addDroppedSnapshot(keyspace, table);
```
dcapwell: nit: its best to avoid mutating the `state` before `apply` is called.  There are different code paths that can choose to reorder commands or drop them, so by mutating the state *before* `apply` you set yourself to get corrupted.

In your specific test no reordering should be done, but pointing out its not a safe pattern.

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: dropped

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: when doing equals i am doing hashCode too ... ah you mean ... everything? I will check.

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: The logic I have is too complicated to put the state outside / before command creation. I'll keep it like it is for now.

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: i will try to do something with this.

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: yeah

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: not applicable anymore

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: not applicable anymore / resolved

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: merged

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: I am OK with dropping this, we would just have `State::hasTables` in `addIf` so we just skip the command calling this altogether.

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: Yes, I may simplify, it is true that we always have just one keyspace to create. I probably just followed what the code was giving me and then it ended up in more "generic" way with multiple keyspaces ... 

EDIT:

well, I know that `difference` method will give me a set of one keyspace, yes. That's because what I am doing - one keyspace at a time. So `for` loop is not necessary. But I like that I am making difference like that. It is just `difference` set will contain always just one entry.

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: yes :D I am checking that ... I don't know. I dont believe it, better to check. It is virtually for free ...

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: I may reshuffle a little bit, yes.

test/distributed/org/apache/cassandra/fuzz/snapshots/SnapshotsTest.java
smiklosovic: not necessary

src/java/org/apache/cassandra/service/snapshot/SnapshotManager.java
```diff
@@ -17,156 +17,541 @@
...
         {
-            logger.info("Scheduling expired snapshot cleanup with initialDelaySeconds={} and cleanupPeriodSeconds={}",
-                        initialDelaySeconds, cleanupPeriodSeconds);
-            cleanupTaskFuture = executor.scheduleWithFixedDelay(this::clearExpiredSnapshots, initialDelaySeconds,
-                                                                cleanupPeriodSeconds, TimeUnit.SECONDS);
+            TruncationNotification truncationNotification = (TruncationNotification) notification;
+            ColumnFamilyStore cfs = truncationNotification.cfs;
+            if (!truncationNotification.disableSnapshot && cfs.isAutoSnapshotEnabled())
+            {
+                SnapshotOptions opts = SnapshotOptions.systemSnapshot(cfs.name, SnapshotType.TRUNCATE, cfs.getKeyspaceTableName())
```
frankgh, line 496: I think we are missing the cfs here.
```suggestion
                SnapshotOptions opts = SnapshotOptions.systemSnapshot(cfs.name, SnapshotType.TRUNCATE, cfs.getKeyspaceTableName()).cfs(cfs)
```

src/java/org/apache/cassandra/io/util/PathUtils.java
```diff
@@ -401,6 +401,18 @@ public static void deleteRecursive(Path path)
         delete(path);
     }
 
+    /**
+     * Empties everything in directory of "path" but keeps the directory itself.
+     *
+     * @param path directory to be emptied
+     * @throws FSWriteError if any part of the tree cannot be deleted
+     */
+    public static void clearDirectory(Path path)
```
frankgh: it feels weird to have production code only for a test. I would move this out of the prod code into test code. Especially concerning because of what this method can do.

src/java/org/apache/cassandra/repair/PreviewRepairTask.java
```diff
@@ -128,7 +129,7 @@ private void maybeSnapshotReplicas(TimeUUID parentSession, String keyspace, List
             for (String table : mismatchingTables)
             {
                 // we can just check snapshot existence locally since the repair coordinator is always a replica (unlike in the read case)
-                if (!Keyspace.open(keyspace).getColumnFamilyStore(table).snapshotExists(snapshotName))
+                if (SnapshotManager.instance.getSnapshot(keyspace, table, snapshotName).isEmpty())
```
frankgh: I see a lot of calls checking for existence of a snapshot in the form of getSnapshot -> isEmpty. I wonder if we should add a dedicated method to check for snapshot existence?

src/java/org/apache/cassandra/service/CassandraDaemon.java
```diff
@@ -291,7 +300,7 @@ protected void setup()
         {
             SystemKeyspace.snapshotOnVersionChange();
         }
-        catch (IOException e)
+        catch (Throwable e)
```
frankgh, line 303: why the change here?

src/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -308,7 +309,9 @@ public interface StorageServiceMBean extends NotificationEmitter
      * @param options map of options for cleanup operation, consult nodetool's ClearSnapshot
      * @param tag name of snapshot to clear, if null or empty string, all snapshots of given keyspace will be cleared
      * @param keyspaceNames name of keyspaces to clear snapshots for
+     * @deprecated See CASSANDRA-18111
      */
+    @Deprecated(since = "5.1")
```
frankgh, line 314: ah interesting, these move to its own MBean, something to keep in mind for Cassandra ANalytics

src/java/org/apache/cassandra/service/snapshot/ClearSnapshotTask.java
```diff
@@ -0,0 +1,196 @@
...
+
+    /**
+     * Returns predicate which will pass the test when arguments match.
+     *
+     * @param tag name of snapshot
+     * @param options options for filtering
+     * @param keyspaceNames names of keyspaces a snapshot is supposed to be from
+     * @return predicate which will pass the test when arguments match.
+     */
+    public static Predicate<TableSnapshot> getPredicateForCleanedSnapshots(String tag, Map<String, Object> options, String... keyspaceNames)
```
frankgh: do we need this as public? maybe make it package-private instead?

frankgh (commented):
I've added some additional comments

src/java/org/apache/cassandra/service/snapshot/SnapshotManager.java
smiklosovic: @frankgh 

Actually, no ... it is good as it is, for dropping we need to add it, but for truncation we do not. 

The reason we need it in case of dropping a table is that if we have not passed it there like that, then later on when we are getting CFS in `TakeSnapshotTask`, that CFS would not be found. Look into `dropCf` method in `Keyspace`. There we are removing cfs from `columnFamilyStores` by `tableId`. But later on, when we get CFS in `TakeSnapshotTask`, we would try to get that from the very same map  - which would return null to us. So just for dropping case, we pass just removed cfs into `TableDroppedNotification`. 

On the other hand, I think we do not need it in `TablePreScrubNotification` case.

src/java/org/apache/cassandra/io/util/PathUtils.java
smiklosovic: moved to where it is used.

src/java/org/apache/cassandra/service/CassandraDaemon.java
smiklosovic: because what it internally does is that it takes a snapshot and `executeTask` in snapshot manager will ultimately throw `SnapshotException` which is `RuntimeException`, not `IOException`. I do not want `SnapshotManager` to throw checked exceptions, except cases when we moved the snapshotting methods from `StorageService(MBean)` to `SnapshotManagerMBean` as we need to still honor that contract for whoever is using it externally. 

`IOException` was caught here because `snapshotOnVersionChange` was throwing it while snapshotting. We may catch `SnapshotException` here instead `Throwable` but I like this more, it is actually safer, because whatever happens in `snapshotOnVersionChange` will be evaluated as errorneous. I do know under what circumstance we would be OK with a node throwing exception in that method and we would act like nothing wrong happened and just carried on.

src/java/org/apache/cassandra/repair/PreviewRepairTask.java
smiklosovic: added two `exists` methods, one taking a predicate, another just ks name, cf name and snapshot name

src/java/org/apache/cassandra/service/snapshot/ClearSnapshotTask.java
smiklosovic: made package protected

test/unit/org/apache/cassandra/index/sasi/SASIIndexTest.java
```diff
@@ -199,7 +200,7 @@ public void testSASIComponentsAddedToSnapshot() throws Throwable
 
         try
         {
-            store.snapshot(snapshotName);
+            SnapshotManager.instance.takeSnapshot(snapshotName, store.getKeyspaceTableName());
```
frankgh, line 203: not that it matters much, but the equivalent of this is actually:
```suggestion
            SnapshotManager.instance.takeSnapshot(snapshotName, store.getKeyspaceTableName(), store.getTableName());
```

and in fact you clear the snapshot after the test for the specific table

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
```diff
@@ -126,8 +127,8 @@ protected void runMayThrow() throws Exception
 
         if (DatabaseDescriptor.isSnapshotBeforeCompaction())
         {
-            Instant creationTime = now();
-            cfs.snapshotWithoutMemtable(creationTime.toEpochMilli() + "-compact-" + cfs.name, creationTime);
+            SnapshotOptions options = SnapshotOptions.systemSnapshot(cfs.name, SnapshotType.COMPACT, cfs.getKeyspaceTableName()).skipFlush().build();
```
frankgh, line 130: I don't think we were specifying a single ks/table previously. So essentially we will only be taking a snapshot of the keyspace/table during the compaction task. Before we used to take a snapshot of all keyspaces and tables. I think that's fine for the compaction task, but want to make sure you are aware of this change and make sure this is intended.

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
```diff
@@ -287,66 +473,79 @@ protected static String buildSnapshotId(String keyspaceName, String tableName, U
...
         {
-            // Snapshot file structure format is {data_dir}/snapshots/{snapshot_name}/.{index}/{sstable-component}.db
-            liveDir = File.getPath(liveDir.getParent().toString(), snapshotFilePath.getParent().getFileName().toString());
+            throw Throwables.unchecked(e);
         }
-        return new File(liveDir.toString(), snapshotFilePath.getFileName().toString());
     }
 
-    public static Predicate<TableSnapshot> shouldClearSnapshot(String tag, long olderThanTimestamp)
+    public static Refs<SSTableReader> getSnapshotSSTableReaders(String keyspace, String table, String tag) throws IOException
```
frankgh, line 496: I'd like to have a signature here that takes a CFS, any reason we don't have it? it can simplify the call sites
```suggestion
    public static Refs<SSTableReader> getSnapshotSSTableReaders(String keyspace, String table, String tag) throws IOException
    {
        return getSnapshotSSTableReaders(Keyspace.open(keyspace).getColumnFamilyStore(table), tag);
    }

    public static Refs<SSTableReader> getSnapshotSSTableReaders(ColumnFamilyStore cfs, String tag)
```

src/java/org/apache/cassandra/db/repair/CassandraValidationIterator.java
```diff
@@ -270,7 +272,7 @@ public void close()
         {
             // we can only clear the snapshot if we are not doing a global snapshot validation (we then clear it once anticompaction
             // is done).
-            cfs.clearSnapshot(snapshotName);
+            SnapshotManager.instance.clearSnapshot(cfs.keyspace.getName(), cfs.name, snapshotName);
```
frankgh: for consistency with the other changes in this file:
```suggestion
            SnapshotManager.instance.clearSnapshot(cfs.getKeyspaceName(), cfs.getTableName(), snapshotName);
```

frankgh (commented):
small final comments

test/unit/org/apache/cassandra/index/sasi/SASIIndexTest.java
smiklosovic: not applicable

src/java/org/apache/cassandra/db/compaction/CompactionTask.java
smiklosovic: not applicable, we indeed take snapshot on one table and before we operated on `cfs` which was one table too.

src/java/org/apache/cassandra/service/snapshot/TableSnapshot.java
smiklosovic: will be done

src/java/org/apache/cassandra/db/repair/CassandraValidationIterator.java
smiklosovic: OK

PR #3375, opened 2024-06-14
CASSANDRA-19705: Reconfigure CMS after replacement, bootstrap and move operations

krummas:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
+++ b/src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
@@ -373,6 +373,20 @@ public class ClusterMetadataService
         InProgressSequences.finishInProgressSequences(ReconfigureCMS.SequenceKey.instance);
     }
 
+    public void maybeReconfigureCMS(ClusterMetadata metadata)
+    {
+        try
+        {
+            if (PrepareCMSReconfiguration.needsReconfiguration(metadata))
+                reconfigureCMS(ReplicationParams.meta(metadata));
+        }
+        catch (Throwable t)
+        {
+            JVMStabilityInspector.inspectThrowable(t);
+            logger.error("Could not reconfigure CMS, operator should run `nodetool cms reconfigure` to make sure CMS placement is correct", t);
+        }
+    }
+
     public boolean applyFromGossip(ClusterMetadata expected, ClusterMetadata updated)
     {
         logger.debug("Applying from gossip, current={} new={}", expected, updated);
--- a/src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
+++ b/src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
@@ -362,7 +362,6 @@ public class ClusterMetadataService
         }
     }
 
-    // This method is to be used _only_ for interactive purposes (i.e. nodetool), since it assumes no retries are going to be attempted on reject.
     public void reconfigureCMS(ReplicationParams replicationParams)
     {
         Transformation transformation = new PrepareCMSReconfiguration.Complex(replicationParams);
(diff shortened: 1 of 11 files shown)


src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
```diff
@@ -373,6 +373,20 @@ public void reconfigureCMS(ReplicationParams replicationParams)
         InProgressSequences.finishInProgressSequences(ReconfigureCMS.SequenceKey.instance);
     }
 
+    public void maybeReconfigureCMS(ClusterMetadata metadata)
+    {
+        try
+        {
+            if (PrepareCMSReconfiguration.needsReconfiguration(metadata))
+                reconfigureCMS(ReplicationParams.meta(metadata));
```
beobal: There's a comment on `reconfigureCMS` saying it's only to be used for interactive purposes due to its lack of retries. I _think_ this comment is simply outdated as retries is baked into the processor implementations now.

src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
```diff
@@ -373,6 +373,20 @@ public void reconfigureCMS(ReplicationParams replicationParams)
         InProgressSequences.finishInProgressSequences(ReconfigureCMS.SequenceKey.instance);
     }
 
+    public void maybeReconfigureCMS(ClusterMetadata metadata)
```
beobal: I think it might actually be better to unconditionally reconfigure as it removes the window between checking if we need to and submitting. It's currently safe to do so anyway as a no-op will add 2 entries to the log, something like:
```
20 |  72058140398682507 | 2024-06-20 12:40:49.681000+0000 |         ADVANCE_CMS_RECONFIGURATION | AdvanceCMSReconfiguration{idx=0, current=FinishReconfiguration(), diff=Diff{additions=[], removals=[]}, activeTransition=null}
19 |  72058140398682506 | 2024-06-20 12:40:49.631000+0000 | PREPARE_COMPLEX_CMS_RECONFIGURATION | PrepareCMSReconfiguration#Complex{replicationParams=ReplicationParams{class=org.apache.cassandra.locator.MetaStrategy, datacenter1=3}}
```
but we could reduce this to just the prepare entry if we move the check to  `PrepareCMSReconfiguration::execute`
```
if (newCms.equals(currentCms))
{
    logger.info("Proposed CMS reconfiguration resulted in no required modifications at epoch {}", prev.epoch.getEpoch());
    return Transformation.success(prev.transformer(), LockedRanges.AffectedRanges.EMPTY);
}
```

src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
```diff
@@ -373,6 +373,20 @@ public void reconfigureCMS(ReplicationParams replicationParams)
...
+    {
+        try
+        {
+            if (PrepareCMSReconfiguration.needsReconfiguration(metadata))
+                reconfigureCMS(ReplicationParams.meta(metadata));
+        }
+        catch (Throwable t)
+        {
+            JVMStabilityInspector.inspectThrowable(t);
+            logger.error("Could not reconfigure CMS, operator should run `nodetool cms reconfigure` to make sure CMS placement is correct", t);
```
beobal: if we do make a change to submit the reconfig unconditionally, it would be good to still log something like this in the event that the submission fails, maybe at `WARN`?

beobal (changes requested):

src/java/org/apache/cassandra/tcm/ClusterMetadataService.java
krummas: yeah either way I think we're good here - if the reconfiguration fails we'll alert the operator via metrics and `nodetool cms`

PR #3377, opened 2024-06-14
[CASSANDRA-19706] Enable colored output of nodetool status

rtib:
This improves the readability of nodetool status' output by enabling colored status lines for nodes joining, leaving or moving printed yellow and nodes not up printed red.


Spy156:
Some environments might not support ANSI color codes, so it would be good to add detection for environments that don't support it, and either default to non-colored output or provide a clear error/warning.

rtib:
> Some environments might not support ANSI color codes, so it would be good to add detection for environments that don't support it, and either default to non-colored output or provide a clear error/warning.

Well, that was the idea with the environment variables to activate the feature. They're set in the environments which support ANSI colors. But, as this is going to be part of the reengineering work ongoing with [CASSANDRA-17445](https://issues.apache.org/jira/browse/CASSANDRA-17445), the coloring will be implemented with PicoCLI, which is auto-detecting terminal features.

PR #3378, opened 2024-06-18
CEP-15 (C*): Don't run incremental repairs for accord only repairs

bdeggleston:


dcapwell (approved):
looks good, but would be good to add `--paxos-only` to `org.apache.cassandra.repair.FuzzTestBase#repairOption`

PR #3379, opened 2024-06-18
ExecuteSyncPoint and Fetch now throw Timeout if all sub-errors are Ti…

dcapwell:


src/java/org/apache/cassandra/net/Verb.java
```diff
@@ -335,8 +335,8 @@ public enum Verb
     ACCORD_GET_EPHMRL_READ_DEPS_REQ (162, P2, writeTimeout, IMMEDIATE,          () -> GetEphmrlReadDepsSerializers.request, AccordService::verbHandlerOrNoop, ACCORD_GET_EPHMRL_READ_DEPS_RSP),
     ACCORD_GET_MAX_CONFLICT_RSP     (163, P2, writeTimeout, IMMEDIATE,          () -> GetMaxConflictSerializers.reply,      RESPONSE_HANDLER                                                            ),
     ACCORD_GET_MAX_CONFLICT_REQ     (164, P2, writeTimeout, IMMEDIATE,          () -> GetMaxConflictSerializers.request,    AccordService::verbHandlerOrNoop, ACCORD_GET_MAX_CONFLICT_RSP),
-    ACCORD_FETCH_DATA_RSP           (145, P2, repairTimeout,IMMEDIATE,          () -> FetchSerializers.reply,               RESPONSE_HANDLER                                                            ),
-    ACCORD_FETCH_DATA_REQ           (146, P2, repairTimeout,IMMEDIATE,          () -> FetchSerializers.request,             AccordService::verbHandlerOrNoop, ACCORD_FETCH_DATA_RSP                     ),
+    ACCORD_FETCH_DATA_RSP           (145, P2, writeTimeout,IMMEDIATE,           () -> FetchSerializers.reply,               RESPONSE_HANDLER                                                            ),
```
dcapwell, line 338: spoke with @bdeggleston about this in slack, this timeout is too large and using an unrelated config (bootstrap and repair are not the same thing).  This isn't 100% related to this patch as its from my topology fix patch, but trying to pull out bug fixes to make reviews easier

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -430,57 +442,76 @@ public long repair(@Nonnull Seekables keysOrRanges, long epoch, long queryStartN
...
-        RuntimeException existingFailures = null;
+        Throwable existingFailures = null;
         Long success = null;
-        long backoffMillis = 0;
+        long backoffMillis = initialBackoffMillis;
         for (int attempt = 0; attempt < retryAttempts; attempt++)
         {
-            try
-            {
-                Thread.sleep(backoffMillis);
```
dcapwell, line 453: Mostly refactored to make the tests easier (as they count calls to sleep) and make it clear *when* we backoff; which is only when the type of error is expected.

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -430,57 +442,76 @@ public long repair(@Nonnull Seekables keysOrRanges, long epoch, long queryStartN
...
+                backoffMillis = Math.min(backoffMillis * 2, maxBackoffMillis);
             }
+            // if an unknown/unexpected error happens retry stops right away
         }
         if (success == null)
         {
             checkState(existingFailures != null, "Didn't have success, but also didn't have failures");
-            throw existingFailures;
+            Throwables.throwIfUnchecked(existingFailures);
+            throw new RuntimeException(existingFailures);
```
dcapwell, line 513: this is possible on the JVM, but super tacky to hit this case... you either do this *outside* of javac or trick javac... In the common case (most likely 100%) we will return from the previous line

test/distributed/org/apache/cassandra/distributed/test/accord/AccordBootstrapTest.java
```diff
@@ -267,7 +267,7 @@ public void bootstrapTest() throws Throwable
                     StreamListener.listener.forSession(session -> {
                         Assert.assertEquals(node3Addr, session.peer.getAddress());
                         Assert.assertEquals(0, session.getNumRequests());
-                        Assert.assertTrue(session.getNumTransfers() > 0);
+                        Assert.assertTrue(session.getNumKeyspaceTransfers() > 0);
```
dcapwell, line 270: I punted on trying to fix the test... we actually don't stream anything....

test/unit/org/apache/cassandra/repair/LocalSyncTaskTest.java
```diff
@@ -171,7 +171,7 @@ private static void assertNumInOut(StreamPlan plan, int expectedIncoming, int ex
         StreamCoordinator coordinator = plan.getCoordinator();
         StreamSession session = Iterables.getOnlyElement(coordinator.getAllStreamSessions());
         assertEquals(expectedIncoming, session.getNumRequests());
-        assertEquals(expectedOutgoing, session.getNumTransfers());
+        assertEquals(expectedOutgoing, session.getNumKeyspaceTransfers());
```
dcapwell, line 174: I punted on trying to fix the test... we actually don't stream anything....

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -430,57 +442,83 @@ public long repair(@Nonnull Seekables keysOrRanges, long epoch, long queryStartN
...
+    static ReadExhaustedException newBarrierExhausted(TxnId txnId, boolean global)
+    {
+        //TODO (usability): not being able to show the txn is a bad UX, this becomes harder to trace back in logs
+        return new ReadExhaustedException(global ? ConsistencyLevel.ANY : ConsistencyLevel.QUORUM, 0, 0, false, ImmutableMap.of());
+    }
+
+    @VisibleForTesting
+    static boolean isTimeout(Throwable t)
+    {
+        return t instanceof Timeout || t instanceof ReadTimeoutException || t instanceof Preempted || t instanceof ReadPreemptedException;
```
aweisberg, line 467: I literally just encountered the Preempted == timeout issue in my testing today.

src/java/org/apache/cassandra/service/accord/exceptions/ReadExhaustedException.java
```diff
@@ -0,0 +1,39 @@
...
+package org.apache.cassandra.service.accord.exceptions;
+
+import java.util.Map;
+
+import org.apache.cassandra.db.ConsistencyLevel;
+import org.apache.cassandra.exceptions.ReadFailureException;
+import org.apache.cassandra.exceptions.RequestFailureReason;
+import org.apache.cassandra.locator.InetAddressAndPort;
+
+public class ReadExhaustedException extends ReadFailureException
```
aweisberg, line 28: We can only hit exhausted for reads and specifically barriers?

aweisberg (approved):
+1

src/java/org/apache/cassandra/service/accord/AccordService.java
dcapwell: if its jvm-dtest there is a function to retry read/writes (assuming they are idempotent, else we fail)... both are timeouts from a users point of view...

src/java/org/apache/cassandra/service/accord/exceptions/ReadExhaustedException.java
dcapwell: I saw other code paths that use it as well, but they are not user facing; `CoordinateGloballyDurable` (background task), and `FetchCoordinator` (bootstrap / streaming of ranges)

dcapwell: I didn't want to touch those as the coverage is mostly via `BurnTest` so subtle changes can take a long time to detect issues.

PR #3382, opened 2024-06-20
Cassandra 19719

dcapwell:


src/java/org/apache/cassandra/service/accord/AccordSyncPropagator.java
```diff
@@ -304,7 +344,22 @@ public boolean invokeOnFailure()
             {
                 return true;
             }
-        });
+        };
+        if (!failureDetector.isAlive(toEp))
```
dcapwell, line 348: sadly this will still log the error that the address isn't known...

src/java/org/apache/cassandra/service/accord/AccordSyncPropagator.java
```diff
@@ -304,7 +344,22 @@ public boolean invokeOnFailure()
...
+        {
+            // was the endpoint removed from membership?
+            ClusterMetadata metadata = ClusterMetadata.current();
+            if (Gossiper.instance.getEndpointStateForEndpoint(toEp) == null && !metadata.directory.allJoinedEndpoints().contains(toEp) && !metadata.fullCMSMembers().contains(toEp))
```
dcapwell, line 352: was thinking it would be good to refactor as this was copy/pasted from `FailureDetector` but wasn't sure where to put it...

test/unit/org/apache/cassandra/net/SimulatedMessageDelivery.java
```diff
@@ -0,0 +1,282 @@
...
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import javax.annotation.Nullable;
+
+import org.apache.cassandra.exceptions.RequestFailure;
+import org.apache.cassandra.locator.InetAddressAndPort;
+import org.apache.cassandra.utils.concurrent.AsyncPromise;
+import org.apache.cassandra.utils.concurrent.Future;
+
+public class SimulatedMessageDelivery implements MessageDelivery
```
dcapwell, line 35: this was pulled out of the repair simulation tests

test/unit/org/apache/cassandra/service/accord/EpochSyncTest.java
```diff
@@ -0,0 +1,673 @@
...
+        {
+            current = current.forceEpoch(Epoch.create(current.epoch.getEpoch() + 1));
+            notify(current);
+        }
+
+        private void notify(ClusterMetadata current)
+        {
+            Ranges ranges = AccordTopology.createAccordTopology(current).ranges().mergeTouching();
+            if (!current.directory.isEmpty())
+                Assertions.assertThat(ranges).hasSize(1);
```
dcapwell, line 546: there was a test bug early on where there were gaps in the token ring... this isn't *needed* anymore, but left just to make sure nothing breaks again

build.xml
```diff
@@ -1278,13 +1278,12 @@
       <condition property="maxMemory" value="8G" else="4G">
           <equals arg1="${test.classlistprefix}" arg2="distributed"/>
       </condition>
-      <testmacrohelper inputdir="${test.dir}/${test.classlistprefix}" filelist="@{test.file.list}" exclude="**/*.java" timeout="${test.timeout}">
+      <testmacrohelper inputdir="${test.dir}/${test.classlistprefix}" filelist="@{test.file.list}" exclude="**/*.java" timeout="${test.timeout}" maxmemory="${maxMemory}">
```
dcapwell, line 1281: the new epoch sync test was failing and turns out we try to set Xmx below, but that is override by `testmacrohelper` who uses `1024m`.  Need to pass `maxMemory` along to avoid that...

there is a thread about this in https://the-asf.slack.com/archives/CK23JSY2K/p1718991597495909

Accord branch has the `maxmemory` as a property we can override, but not on any other branch...

test/unit/org/apache/cassandra/net/SimulatedMessageDelivery.java
```diff
@@ -0,0 +1,333 @@
...
+{
+    public enum Action { DELIVER, DELIVER_WITH_FAILURE, DROP, DROP_PARTITIONED, FAILURE }
```
dcapwell, line 37: added more actions than repair normally supports, this is more in-line with BurnTests' `NodeSink`

src/java/org/apache/cassandra/service/accord/AccordConfigurationService.java
```diff
@@ -114,15 +120,226 @@ protected EpochState createEpochState(long epoch)
...
+
+        @Override
+        public EpochDiskState truncateTopologyUntil(long epoch, EpochDiskState diskState)
+        {
+            return AccordKeyspace.truncateTopologyUntil(epoch, diskState);
+        }
+    }
+
+    enum MockDiskStateManager implements DiskStateManager
+    {
```
bdeggleston: Can we move this and the `LoggingDiskStateManager` into the equivalent test package?

src/java/org/apache/cassandra/service/accord/AccordSyncPropagator.java
```diff
@@ -210,6 +232,28 @@ public String toString()
                '}';
     }
 
+    public synchronized void onNodesRemoved(Set<Node.Id> removed)
+    {
+        for (Node.Id node : removed)
+        {
+            PendingEpochs epochs = pending.get(node.id);
```
bdeggleston: could you rename this to `pending` or `pendingEpochs`?

src/java/org/apache/cassandra/service/accord/AccordSyncPropagator.java
```diff
@@ -210,6 +232,28 @@ public String toString()
...
+            if (epochs == null) continue;
+            long[] complete = new long[epochs.size()];
```
bdeggleston: could you rename this to `toComplete`?

bdeggleston (changes requested):
just some minor change, looks good otherwise

src/java/org/apache/cassandra/service/accord/AccordSyncPropagator.java
dcapwell: went with `pendingEpochs`; `pending` is already taken

bdeggleston (approved):
+1

PR #3384, opened 2024-06-25
CASSANDRA-17457 CEP-24 Password validation / generation

smiklosovic:
 
```

```


.build/parent-pom-template.xml
```diff
@@ -1239,6 +1239,11 @@
         <artifactId>semver4j</artifactId>
         <version>3.1.0</version>
       </dependency>
+      <dependency>
+        <groupId>org.passay</groupId>
```
frankgh, line 1256: I think it's worth checking with legal about the dual licensing for passay. I'm always a bit nervous when I see LGPL.

conf/cassandra.yaml
```diff
@@ -2148,6 +2148,56 @@ drop_compact_storage_enabled: false
...
+#  # If a password does not contain at least n digits, the validator will emit a failure.
+#  min_digits_chars_fail: 1
+#  # If a password does not contain at least n special characters, the validator will emit a warning.
+#  min_special_chars_warn: 2
+#  # If a password does not contain at least n special characters, the validator will emit a failure.
+#  min_special_chars_fail: 1
+#  # illegal sequences that will fail when they are >= 5 characters long
+#  # illegal sequences might be either alphabetical (form 'abcde'),
+#  # numerical (form '34567'), or US qwery (form 'asdfg').
+#  illegal_sequence_length: 5
```
frankgh, line 2202: I am able to configure this value to -1. I was expecting Cassandra refuse to startup because of misconfiguration. What really happened is that I can't create a password, validation happens only when trying to create a role.

```cql
cassandra@cqlsh> CREATE ROLE francisco2 WITH  PASSWORD = '1234567' AND LOGIN = true;
ConfigurationException: Unable to create instance of validator of class org.apache.cassandra.db.guardrails.CassandraPasswordValidator: Illegal sequence length can not be lower than 5.
```

Should we validate during startup instead?

conf/cassandra.yaml
```diff
@@ -2148,6 +2148,56 @@ drop_compact_storage_enabled: false
...
+#  # tested against this validator. When not in form of FQCN, the
+#  # package name org.apache.cassandra.db.guardrails.generators is prepended.
+#  generator_class_name: CassandraPasswordGenerator
+#  # If set to true, alter role statements which are changing passwords
+#  # will be checking if that password was not already used in the past.
+#  # Password validation fails if it was used already. Defaults to false.
+#  validate_against_historical_values: true
+#  # Maximum previous passwords to validate a password to be altered against
+#  # If a password to validate is found among the last 5, the validation will fail.
+#  max_historical_values: 5
```
frankgh: I don't think these properties are used anywhere

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -2375,6 +2375,12 @@ public interface CheckedFunction
         void apply() throws Throwable;
     }
 
+    @FunctionalInterface
+    public interface CheckedSupplier
+    {
+        ResultMessage apply() throws Throwable;
```
frankgh: NIT: should we align with the JDKs `Supplier` API? In that case, this method would need to be `get` 
```suggestion
        ResultMessage get() throws Throwable;
```

src/java/org/apache/cassandra/cql3/statements/AlterRoleStatement.java
```diff
@@ -114,6 +124,19 @@ public void authorize(ClientState state) throws UnauthorizedException
 
     public ResultMessage execute(ClientState state) throws RequestValidationException, RequestExecutionException
     {
+        if (opts.isGeneratedPassword())
+        {
+            String generatedPassword = Guardrails.password.generate();
+            if (generatedPassword != null)
+                opts.setOption(IRoleManager.Option.PASSWORD, generatedPassword);
+            else
+                throw new InvalidRequestException("You have to enable password_validator and it's generator_class_name property " +
```
frankgh, line 133: should we defer validation to runtime? can we do it validation during startup as well?

src/java/org/apache/cassandra/cql3/statements/AuthenticationStatement.java
```diff
@@ -17,20 +17,37 @@
...
 import org.apache.cassandra.exceptions.RequestExecutionException;
 import org.apache.cassandra.exceptions.RequestValidationException;
 import org.apache.cassandra.exceptions.UnauthorizedException;
+import org.apache.cassandra.schema.SchemaConstants;
 import org.apache.cassandra.service.ClientState;
 import org.apache.cassandra.service.QueryState;
 import org.apache.cassandra.transport.Dispatcher;
 import org.apache.cassandra.transport.messages.ResultMessage;
 
+import static com.google.common.collect.ImmutableList.of;
```
frankgh: Use JDK's List.of instead. `java.util.List#of`

src/java/org/apache/cassandra/db/guardrails/CustomGuardrail.java
```diff
@@ -0,0 +1,176 @@
...
+     * are replaced by the values of the same key in the new configuration.
+     * <p>
+     *
+     * @param newConfig if null or the configuration is an empty map, no reconfiguration happens.
+     * @throws ConfigurationException when new validator can not replace the old one or when it is not possible
+     *                                to instantiate new validator or generator.
+     */
+    void reconfigure(@Nullable Map<String, Object> newConfig)
+    {
+        Map<String, Object> mergedMap = new HashMap<>(getValidator().getParameters());
```
frankgh, line 152: should we validate that newConfig is not null , we can even exit early if it's null or empty

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
```diff
@@ -1178,6 +1184,18 @@ public void setMaximumReplicationFactorThreshold (int warn, int fail)
...
+    @Override
+    public Map<String, Object> getPasswordValidatorConfig()
+    {
+        return password.getConfig();
+    }
+
+    @Override
+    public void reconfigurePasswordValidator(Map<String, Object> config)
+    {
+        password.reconfigure(config);
```
frankgh, line 1196: it feels like we need some flag to disable the ability to reconfigure the password validator via JMX?

test/unit/org/apache/cassandra/db/guardrails/CassandraPasswordValidatorTest.java
```diff
@@ -0,0 +1,245 @@
...
+        validationResult = validator.shouldFail("A1$asdfghjkl");
+        assertTrue(validationResult.isPresent());
+        assertThat(validationResult.get().message, containsString("Password contains the illegal QWERTY sequence 'asdfghjkl'."));
+    }
+
+    @Test
+    public void testWhitespace()
+    {
+        CassandraPasswordValidator validator = new CassandraPasswordValidator(new CustomGuardrailConfig());
+        Optional<ValidationViolation> validationResult = validator.shouldFail("A1$abcd efgh");
```
frankgh: should we allow spaces? I guess a more general question is whether we should allow configuring the special characters via cassandra.yaml

frankgh (commented):
The PR looks great! I have a few comments.

src/java/org/apache/cassandra/cql3/statements/AlterRoleStatement.java
smiklosovic: @frankgh No, we should't do it during startup. Because imagine that you have `cassandra.yaml` "as is", that will not enable password validation. It is enabled when you explicitly enable it in `cassandra.yaml`. But we can not "dynamically enable" `GENERATED PASSWORD` stuff, that will be always there. So we need to treat the situation when a user wants to generate passwords but it is not enabled. What would we detect during startup? We do not know in advance if a user wants to use `GENERATED PASSWORD` or not. If they never use it, then there is not any problem, no?

conf/cassandra.yaml
smiklosovic: `IllegalSequenceRule` of Passay has itself minimum value for this to be 3. I made a check that it can not be set to anything lower than 3 in our configuration too. By default this value will be set to 5 (also Passay's default)

In order to detect this early / pro-actively, we would need to trigger the validator construction in advance. For now it is instantiated on demand so this kind of error is deferred to runtime. I will look into that ...

src/java/org/apache/cassandra/db/guardrails/CustomGuardrail.java
smiklosovic: well ... the way how you effectively "reset" this feature is to reconfigure it by setting the config to an empty map. Imagine you have this guardrail working and for some reason you do not want that anymore (hard to believe why but anyway). So you reset that to use no-op validator by empty map.

pylib/cqlshlib/cql3handling.py
```diff
@@ -34,7 +34,7 @@ def __str__(self):
 
 
 SYSTEM_KEYSPACES = ('system', 'system_schema', 'system_traces', 'system_auth', 'system_distributed', 'system_views',
-                    'system_metrics', 'system_virtual_schema', 'system_cluster_metadata')
+                    'system_metrics', 'system_virtual_schema', 'system_cluster_metadata', 'system_guardrails')
```
smiklosovic: should be removed, probably some leftover.

test/unit/org/apache/cassandra/db/guardrails/CassandraPasswordValidatorTest.java
smiklosovic: I do not think we should allow spaces or whitespace characters. Let's just make it simple. Passay's whitespace rule rejects this:

    protected static final char[] CHARS = new char[]{'\t', '\n', '\u000b', '\f', '\r', ' '};

I don't know who "sane" would want to use these characters in a password. A user would have quite a hard time to figure out how to enter it upon login. Not talking about how they would actually save it to e.g. cqlshrc and how cqlsh would interpret that. That's a can of worms I am not super keen on opening.

The current generator implementation does not use these characters so while we are not generating them in password, they would be valid with these characters in it. I do not like this discrepancy.

For special characters, we might enable it to be configurable with the defaults as provided. One caveat thought. For simplicity, I would not support illegal character to be " ' " (single quote). Because if you go to do this:

    create role 'stefan' with password = 'abcd'efgh'

notice that we would support `'` as valid character, but CQLSH would think that your password is `abcd` and it would reject everything after `'` and it will fail the query. It might be probably escaped but again ... why the hassle?

EDIT:

Thinking about it more, I think that we should not support the configuration of special characters. I think there is a plenty of special chars to choose from already. I just do not see valid scenario when somebody wants so much to change it. And if they want, they can just provide their own validator implementation - it is pluggable for that very reason. It was also explained in the CEP that if somebody doesnt want what we provided they can just code up their own.

.build/parent-pom-template.xml
smiklosovic: As long as it is Apache licensed (which it is) it is fine. People probably saw your comment on the voting thread as well and nobody objected.

src/java/org/apache/cassandra/db/guardrails/Guardrails.java
smiklosovic: https://github.com/apache/cassandra/pull/3384/commits/ef034a41084200ba3dcdcc7d3682fd65db34abf8

src/java/org/apache/cassandra/db/guardrails/CustomGuardrail.java
smiklosovic: https://github.com/apache/cassandra/pull/3384/commits/ef034a41084200ba3dcdcc7d3682fd65db34abf8

frankgh: good point, passing an empty map would allow you to reset the feature to its defaults. However passing a null value would cause an NPE. One additional ask I'd have, I think it's worth adding a log entry at info level here that would allow an operator determine what is effectively configured and when a configuration happened

frankgh: ahh, I see the null check here. Thanks for adding that.

test/unit/org/apache/cassandra/db/guardrails/CassandraPasswordValidatorTest.java
frankgh: one of the passwords generated for me during testing had a backtick `` ` ``  and I had a hard time logging in with that password, so I see your point.

frankgh (approved):
+1 Thanks for addressing the comments

PR #3385, opened 2024-06-25
CASSANDRA-19704 UnsupportedOperationException when reducing scope for LCS compactions

maedhroz:
patch by Caleb Rackliffe; reviewed by ? for CASSANDRA-19704

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/LeveledCompactionTask.java
+++ b/src/java/org/apache/cassandra/db/compaction/LeveledCompactionTask.java
@@ -56,7 +56,7 @@ public class LeveledCompactionTask extends CompactionTask
     @Override
     protected boolean partialCompactionsAcceptable()
     {
-        throw new UnsupportedOperationException("This is now handled in reduceScopeForLimitedSpace");
+        return level == 0;
     }
 
     protected int getLevel()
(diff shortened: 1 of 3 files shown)


src/java/org/apache/cassandra/db/compaction/LeveledCompactionTask.java
```diff
@@ -56,7 +56,7 @@ public CompactionAwareWriter getCompactionAwareWriter(ColumnFamilyStore cfs,
     @Override
     protected boolean partialCompactionsAcceptable()
     {
-        throw new UnsupportedOperationException("This is now handled in reduceScopeForLimitedSpace");
+        return level == 0;
```
jasonstack, line 59: `reduceScopeForLimitedSpace` allows to remove sstable on level 0 for task with `level<=1`.

should we change it to `level <= 1`?

maedhroz: I think you're right. We could potentially not compact fully expired SSTables if we don't change to `level <= 1` from what I'm reading.

CC @krummas

maedhroz: Already fixed in the trunk patch, but I might also add some inline commentary: https://github.com/apache/cassandra/pull/3388

maedhroz: Will fix on commit for 4.0, 4.1, and 5.0

maedhroz:
Committed as https://github.com/apache/cassandra/commit/d54646e40957ffc18f651753941fac3068b79f8e

PR #3391, opened 2024-07-01
[CASSANDRA-19651][4.1] Fix latency reported by ideal consistency level monitoring

netudima:
idealCLWriteLatency metric reported the worst response time instead of the time when ideal CL is satisfied.

Patch by Dmitry Konstantinov; reviewed by TBD for CASSANDRA-19651

### DIFF
--- a/src/java/org/apache/cassandra/service/AbstractWriteResponseHandler.java
+++ b/src/java/org/apache/cassandra/service/AbstractWriteResponseHandler.java
@@ -189,7 +189,7 @@ public abstract class AbstractWriteResponseHandler<T> implements RequestCallback
         }
     }
 
-    public final void expired()
+    protected final void logFailureOrTimeoutToIdealCLDelegate()
     {
         //Tracking ideal CL was not configured
         if (idealCLDelegate == null)
@@ -207,6 +207,11 @@ public abstract class AbstractWriteResponseHandler<T> implements RequestCallback
             //Have the delegate track the expired response
             idealCLDelegate.decrementResponseOrExpired();
         }
+    }
+
+    public final void expired()
+    {
+        logFailureOrTimeoutToIdealCLDelegate();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


test/unit/org/apache/cassandra/service/WriteResponseHandlerTest.java
```diff
@@ -153,17 +154,23 @@ public void idealCLLatencyTracked() throws Throwable
         //dc1
         awr.onResponse(createDummyMessage(0));
         awr.onResponse(createDummyMessage(1));
+
+        // there are no enough responses for ideal EACH_QUORUM yet
```
aweisberg:
```suggestion
        // there are not enough responses for ideal EACH_QUORUM yet
```

src/java/org/apache/cassandra/service/AbstractWriteResponseHandler.java
```diff
@@ -280,6 +289,8 @@ public void onFailure(InetAddressAndPort from, RequestFailureReason failureReaso
 
         failureReasonByEndpoint.put(from, failureReason);
 
+        logFailureOrTimeoutToIdealCLDelegate();
```
aweisberg, line 292: Ouch looks like the failure path didn't decrement at all before?

I guess the only reason it gave some signal was because we would call expired explicitly if gossip thought the endpoint was down so it would be a non-zero metric.

aweisberg (changes requested):
This is great. Thank you for fixing this. One typo in a comment was all I found. Can you create branches for 4.0, 4.1, 5.0, and trunk and also remove the update to `CHANGES.txt`?

src/java/org/apache/cassandra/service/AbstractWriteResponseHandler.java
netudima: yes, I think so, for prolonged node loss cases detected by gossip we have another path via expired() method.

PR #3394, opened 2024-07-02
CASSANDRA-19743: add keyspace name to exception message in StorageService#getEffectiveOwnership for better debugging

maoling:
- more details in the [CASSANDRA-19743](https://issues.apache.org/jira/browse/CASSANDRA-19743)

### DIFF
--- a/src/java/org/apache/cassandra/service/StorageService.java
+++ b/src/java/org/apache/cassandra/service/StorageService.java
@@ -4184,14 +4184,14 @@ public class StorageService extends NotificationBroadcasterSupport implements IE
         if (keyspace != null)
         {
             if (isLocalSystemKeyspace(keyspace))
-                throw new IllegalStateException("Ownership values for keyspaces with LocalStrategy are meaningless");
+                throw new IllegalStateException("Ownership values for keyspaces " + keyspace + " with LocalStrategy are meaningless");
 
             KeyspaceMetadata keyspaceInstance = metadata.schema.getKeyspaces().getNullable(keyspace);
             if (keyspaceInstance == null)
                 throw new IllegalArgumentException("The keyspace " + keyspace + ", does not exist");
 
             if (keyspaceInstance.replicationStrategy instanceof LocalStrategy)
-                throw new IllegalStateException("Ownership values for keyspaces with LocalStrategy are meaningless");
+                throw new IllegalStateException("Ownership values for keyspaces " + keyspace + " with LocalStrategy are meaningless");
 
             strategy = keyspaceInstance.replicationStrategy;
             replicationParams = keyspaceInstance.params.replication;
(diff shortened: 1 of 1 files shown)


src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4184,14 +4184,14 @@ private LinkedHashMap<InetAddressAndPort, Float> getEffectiveOwnership(String ke
         if (keyspace != null)
         {
             if (isLocalSystemKeyspace(keyspace))
-                throw new IllegalStateException("Ownership values for keyspaces with LocalStrategy are meaningless");
+                throw new IllegalStateException("Ownership values for keyspaces " + keyspace + " with LocalStrategy are meaningless");
```
smiklosovic, line 4187: @maoling I do not think this is necessary. This code path is exercised when `keyspace != null` so a user knows which keyspace it is talking about, no? (Except of that, the constructed message is grammatically incorrect).

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4184,14 +4184,14 @@ private LinkedHashMap<InetAddressAndPort, Float> getEffectiveOwnership(String ke
...
 
             KeyspaceMetadata keyspaceInstance = metadata.schema.getKeyspaces().getNullable(keyspace);
             if (keyspaceInstance == null)
                 throw new IllegalArgumentException("The keyspace " + keyspace + ", does not exist");
 
             if (keyspaceInstance.replicationStrategy instanceof LocalStrategy)
-                throw new IllegalStateException("Ownership values for keyspaces with LocalStrategy are meaningless");
+                throw new IllegalStateException("Ownership values for keyspaces " + keyspace + " with LocalStrategy are meaningless");
```
smiklosovic, line 4194: @maoling same as above

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -4209,7 +4209,10 @@ private LinkedHashMap<InetAddressAndPort, Float> getEffectiveOwnership(String ke
                 for (String keyspaceName : userKeyspaces)
                 {
                     if (!Schema.instance.getKeyspaceInstance(keyspaceName).getReplicationStrategy().hasSameSettings(replicationStrategy))
-                        throw new IllegalStateException("Non-system keyspaces don't have the same replication settings, effective ownership information is meaningless");
+                        throw new IllegalStateException(String.format(
+                        "Non-system keyspaces: %s and %s don't have the same replication settings, effective ownership information is meaningless",
```
smiklosovic, line 4213: @maoling you might probably gather all keyspaces for which replication settings are different and enrich exception message with them?

However, there might be in theory dozens of keyspaces and it might pollute the output unnecessarily. Not sure how to control that.

maoling:
> However, there might be in theory dozens of keyspaces and it might pollute the output unnecessarily. Not sure how to control that.

You make a valid point. This PR was created last year, and it appears to have no value now. Let me close it. @smiklosovic Thanks for your review.

PR #3395, opened 2024-07-02
CASSANDRA-19744 Accord migration and interop correctness

aweisberg:
[CASSANDRA-19744](https://issues.apache.org/jira/browse/CASSANDRA-19744)

There are several different fixes/changes here.

* Batch log replay splits mutations based on whether they should be Accord/non-Accord routed
* Failed batchlog replay of Accord txns are sent to hints as a special host ID and hints replays them and if necessary converts them back to per host hints, per host hints are never converted to Accord hints so the amplification is bounded in the rare case this occurs.
* Hints split and retry Accord/non-Accord and handle Accord tens with no specific endpoint just for the batchlog
* Batch log is skipped if the entire batch is routed to Accord, if the part of the batch is non-Accord then he entire batch goes through the batchlog
* Barriers now return the precise Accord owned ranges that were barriered which may be a subset of what was requested 
* Repair now drives progress for migration based on what was actually barriered
* Read repair routes the repair mutation based on whether the token is Accord managed, but does not retry if it fails. We just fail the read since this should be rare.
* Regular mutations will loop splitting and retrying until timeout or until the routing is correct.
* Mutations now have a flag that indicates whether they bypass validation for a conflict with a transaction system, Accord always sets this flag since it doesn't conflict with itself. Everything else doesn't and a CFS will reject a mutation that should have been managed by Accord. Counter and virtual tables don't use this since they don't work with Accord right now.
* Timestamps of non-transactional writes are preserved in that the timestamp that is used by Accord to write data is the minimum of the Accord timestamp and the original coordinator generated timestamp to avoid data resurrection from batch and hints applied at a later time. CAS, Accord transactions, and Read Repair don't preserve timestamps. 
* Fixed handling of non-existent tables and tables that are system tables where we would NPE because not all schema sources contain non-distributed tables.
* Fixed registration race conditions in TestChangeListener pausing enactment of epochs
* Fixed `cep-15-accord` specific breakage of handling `CoordinatorBehind` responses
* `CoordinatorBehind` errors no longer fail operations, the coordinator will retry 

~~TODO CoordinatorBehind retry could retry operations that are not idempotent, but may have already succeeded~~
~~TODO Check that barriers of keys check the key was actually barriered~~
TODO The batchlog failed routing test doesn't actually experience failed routing it's also a timeout, but it should be possible to trigger a failed routing by doing a transaction on the key and then allowing the out of sync coordinator to catch up and be forced to execute the transaction in the new epoch.


src/java/org/apache/cassandra/service/accord/api/AccordAgent.java
```diff
@@ -122,7 +122,7 @@ public long preAcceptTimeout()
     @Override
     public Txn emptyTxn(Kind kind, Seekables<?, ?> seekables)
     {
-        return new Txn.InMemory(kind, seekables, TxnRead.EMPTY, TxnQuery.EMPTY, null);
+        return new Txn.InMemory(kind, seekables, TxnRead.EMPTY, TxnQuery.UNSAFE_EMPTY, null);
```
bdeggleston: could we rename this method to `emptySystemTxn` or `emptyUnsafeTxn` or something so people know that it doesn't offer all the normal guardrails

src/java/org/apache/cassandra/db/CassandraKeyspaceWriteHandler.java
```diff
@@ -89,7 +89,7 @@ private CommitLogPosition addToCommitLog(Mutation mutation)
                     if (update.metadata().params.memtable.factory().writesShouldSkipCommitLog())
                         ids.add(update.metadata().id);
                 }
-                mutation = mutation.without(ids);
+                mutation = mutation.filter(ids::contains);
```
bdeggleston: This doesn't seem right. If I'm reading this right then when we encounter mixed `writesShouldSkipCommitLog` mutations we'll _only_ write commit log entries for tables which aren't supposed to be written to the commit log?

src/java/org/apache/cassandra/db/IMutation.java
```diff
@@ -75,4 +78,22 @@ default boolean allowsOutOfRangeMutations()
...
+     * True if this mutation is being applied by a transaction system or doesn't need to be
+     * and conflicts between this mutation and transactions systems that are managing all or part of this table
+     * should be assumed to be handled already (by either Paxos or Accord) and the mutation should be applied.
+     *
+     * This causes mutations against tables to fail if they are from a non-transaction sub-system such as mutations,
+     * logged and unlogged batches, hints, and read repair against tables that are being managed by a transaction system
+     * like Accord that can't safely read data that is written non-transactionally.
+     *
+     */
+    default boolean allowPotentialTransactionConflicts()
```
bdeggleston: I had a hard time parsing this comment
TODO: propose something easier to understand

test/unit/org/apache/cassandra/cql3/PagingTest.java
```diff
@@ -140,7 +140,7 @@ public boolean isWorthMergingForRangeQuery(ReplicaCollection merged, ReplicaColl
         }
         Thread.sleep(1500);
 
-        Statement stmt = new SimpleStatement(String.format("SELECT DISTINCT token(id, id2), id, id2 FROM %s", table));
+        Statement stmt = new SimpleStatement(String.format("SELECT DISTINCT token(id, id2), id, id2 FROM %s WHERE token(id, id2) >= -9223372036854775808", table));
```
bdeggleston: what's up with the token bit you added here?

src/java/org/apache/cassandra/batchlog/BatchlogManager.java
```diff
@@ -340,66 +391,116 @@ private static class ReplayingBatch
...
+                logger.debug("Unexpected batchlog replay exception", e);
+                failure = Throwables.merge(failure, e);
+            }
+
+            if (failure != null)
+                throw Throwables.unchecked(failure);
         }
 
-        private int addMutations(int version, List<ByteBuffer> serializedMutations) throws IOException
+        private int addMutations(List<Mutation> unsplitMutations, int version, List<ByteBuffer> serializedMutations) throws IOException
```
bdeggleston: this (and `addMutation`) can be made static if you pass `writtenAt` in

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
```diff
@@ -0,0 +1,399 @@
...
+        List<T> normalMutations = null;
+        for (int i=0,mi=mutations.size(); i<mi; i++)
+        {
+            T mutation = mutations.get(i);
+            Pair<T, T> accordAndNormalMutation = splitMutationIntoAccordAndNormal(mutation, cm);
+            T accordMutation = accordAndNormalMutation.left;
+            if (accordMutation != null)
+            {
+                if (accordMutations == null)
+                    accordMutations = new ArrayList<>(Math.min(mutations.size(), 10));
```
bdeggleston: why are you allocating a minimum of 10 entries here? Was this meant to be max?

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
```diff
@@ -0,0 +1,399 @@
...
+                // commit consistency
+                ConsistencyLevel commitCL = mode.commitCLForStrategy(consistencyLevel);
+                if (commitCL != null)
+                    return commitCL;
+            }
+        }
+        return null;
+    }
+
+    public static <T extends IMutation> Pair<List<T>, List<T>> splitMutationsIntoAccordAndNormal(List<T> mutations)
```
bdeggleston: Instead of using Pair, can we just define a simple container class that contains accord and normal mutations? Pair's generic left and right labels make reading the code more difficult / error prone. Plus you can roll some of the repetitive list instantiation stuff into the container class.

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
```diff
@@ -0,0 +1,399 @@
...
+        }
+        // Potentially ignore commit consistency level if the strategy specifies accord and not migration
+        ConsistencyLevel clForCommit = consistencyLevelForCommit(mutations, consistencyLevel);
+        TxnUpdate update = new TxnUpdate(fragments, TxnCondition.none(), clForCommit, true);
+        Txn.InMemory txn = new Txn.InMemory(Keys.of(partitionKeys), TxnRead.EMPTY, TxnQuery.NONE, update);
+        IAccordService accordService = AccordService.instance();
+        return accordService.coordinateAsync(txn, consistencyLevel, queryStartNanoTime);
+    }
+
+    public static void dispatchMutationsWithRetryOnDifferentSystem(List<? extends IMutation> mutations, ConsistencyLevel consistencyLevel, long queryStartNanoTime)
```
bdeggleston: can we move this to StorageProxy?

src/java/org/apache/cassandra/service/BatchlogResponseHandler.java
```diff
@@ -31,6 +31,8 @@ public class BatchlogResponseHandler<T> extends AbstractWriteResponseHandler<T>
     AbstractWriteResponseHandler<T> wrapped;
     BatchlogCleanup cleanup;
     protected volatile int requiredBeforeFinish;
+    // Defer cleanup until `maybeAckMutationForCleanup` is called
+    private boolean deferCleanup = false;
```
bdeggleston: this could be made immutable if we split mutations in `StorageProxy#mutateAtomically` before calling `wrapBatchResponseHandler` and determine if we need to defer batchlog cleanup based on the existence of accord mutations.

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1344,66 +1266,151 @@ public static void mutateAtomically(Collection<Mutation> mutations,
...
+                                                                                       consistencyLevel,
+                                                                                       batchConsistencyLevel,
+                                                                                       WriteType.BATCH,
+                                                                                       cleanup,
+                                                                                       queryStartNanoTime);
+                        // exit early if we can't fulfill the CL at this time.
+                        wrappers.add(wrapper);
+                    }
+
+                    // TODO (review): A lot of the code duplication in splitting is from WriteResponseHandlerWrapper, but we could split before making the wrappers
```
bdeggleston: see my comment on [BatchlogResponseHandler.java](https://github.com/apache/cassandra/pull/3395/files/96a7efd671fbd3592676af7e5cbf5db9b7512cf0#diff-f7c3660cfeb41cef884d6c0a9d8d857ff39580193b650a9ae1ca338011bc56ef). I agree splitting up the mutations before wrapping them would make this a little neater

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -364,8 +392,16 @@ public IVerbHandler<? extends Request> verbHandler()
         return requestHandler;
     }
 
-    private <S extends Seekables<?, ?>> long barrier(@Nonnull S keysOrRanges, long epoch, long queryStartNanos, long timeoutNanos, BarrierType barrierType, boolean isForWrite, BiFunction<Node, S, AsyncResult<SyncPoint<S>>> syncPoint)
+    private <S extends Seekables<?, ?>> Seekables barrier(@Nonnull S keysOrRanges, long epoch, long queryStartNanos, long timeoutNanos, BarrierType barrierType, boolean isForWrite, BiFunction<Node, S, AsyncResult<SyncPoint<S>>> syncPoint)
     {
+        // TODO (review): So do we get what we really want picking a random cluster metadata and only doing a barrier on a subset of what was asked?
+        // We will definitely return a result indicating exactly what we ended up doing the barrier on so it's not misleading and won't falsely
+        // make progress when it shouldn't and now we don't have to be as precise as an operator specifying what to do the barrier on
+        keysOrRanges = (S)intersectionWithAccordManagedRanges(keysOrRanges);
```
bdeggleston: Well it's not random, it's the most recent metadata we're aware of :)

To turn the question around, why would we want to run barriers on ranges that we can't run barriers on? As long as we're not breaking the use case of intentionally migrating only a handful of keys/ranges, and we provide the user with clear feedback about what we are/are not doing, I think it's a good middle ground.

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -592,12 +715,9 @@ public TopologyManager topology()
         }
         catch (TimeoutException e)
         {
+            // TODO (review): This is j.u.c timeout so maybe this should mark?
```
bdeggleston: j.u.c?

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
```diff
@@ -0,0 +1,399 @@
...
+    {
+        while (true)
+        {
+            try
+            {
+                // Keep attempting to route the mutations until they all succeed on the correct system
+                // TODO (review): Metrics are going to double count when we split between the two systems
```
bdeggleston: that seems reasonable... a write is following both write paths

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
```diff
@@ -0,0 +1,399 @@
...
+                List<? extends IMutation> accordMutations = accordAndNormal.left;
+                // TODO this can race with migration to/from Accord and Accord doesn't know what range we are referring to or it just hasn't updated to the epoch we are aware of
+                // client side retry can still succeed but the error could be exposed
+                Pair<TxnId, Future<TxnResult>> accordResult = accordMutations != null ? mutateWithAccordAsync(accordMutations, consistencyLevel, queryStartNanoTime) : null;
+                List<? extends IMutation> normalMutations = accordAndNormal.right;
+
+                Throwable failure = null;
+                try
+                {
+                    if (normalMutations != null)
```
bdeggleston: I think this whole try block should be contained in this if block?

src/java/org/apache/cassandra/service/consensus/migration/ConsensusRequestRouter.java
```diff
@@ -85,11 +87,17 @@ ConsensusRoutingDecision decisionFor(TransactionalMode transactionalMode)
...
     private static TableMetadata metadata(ClusterMetadata cm, String keyspace, String table)
     {
-        KeyspaceMetadata ksm = cm.schema.getKeyspaceMetadata(keyspace);
-        TableMetadata tbm = ksm != null ? ksm.getTableOrViewNullable(table) : null;
-
+        Optional<KeyspaceMetadata> ksm = cm.schema.maybeGetKeyspaceMetadata(keyspace);
+        if (ksm.isEmpty())
+        {
+            KeyspaceMetadata ksm2 = Schema.instance.getKeyspaceMetadata(keyspace);
+            if (ksm2 == null)
```
bdeggleston: what's happening here? Is this meant to allow for system and virtual tables?

The call to `ClusterMetadata#current` in `Schema#getKeyspaceMetadata` creates a race that may cause us to erroneously pick paxos for accord tables. 

If the cluster metadata object doesn't contain the keyspace we're looking for, we'll check w/ Schema. If the schema changes and the keyspace we're looking for appears between calls we'll return null and we'll pick paxos when we should have either thrown an exception or picked accord

src/java/org/apache/cassandra/service/consensus/migration/ConsensusRequestRouter.java
```diff
@@ -98,18 +106,37 @@ public ConsensusRoutingDecision routeAndMaybeMigrate(@Nonnull DecoratedKey key,
...
     }
 
+    public static TableMetadata getTableMetadata(ClusterMetadata cm, TableId tableId)
+    {
+        TableMetadata tm = cm.schema.getTableMetadata(tableId);
+        if (tm == null)
+        {
+            // It's a non-distributed table which is fine, but we want to error if it doesn't exist
+            // We should never actually reach here unless there is a race with dropping the table
+            TableMetadata tm2 = Schema.instance.getTableMetadata(tableId);
```
bdeggleston: same issue as `getKeyspaceMetadata`

src/java/org/apache/cassandra/service/StorageProxy.java
```diff
@@ -1263,53 +1218,22 @@ public static void mutateWithTriggers(List<? extends IMutation> mutations,
...
+        List<Mutation> augmented = TriggerExecutor.instance.execute(mutations);
 
         String keyspaceName = mutations.iterator().next().getKeyspaceName();
-        boolean updatesView = Keyspace.open(mutations.iterator().next().getKeyspaceName())
+        boolean updatesView = Keyspace.open(keyspaceName)
                               .viewManager
                               .updatesAffectView(mutations, true);
 
-
+        // TODO (review): This is not the same as augmented? Doesn't seem right
```
bdeggleston: yeah seems like a metrics bug

src/java/org/apache/cassandra/hints/HintsDispatcher.java
```diff
@@ -138,8 +196,11 @@ private Action sendHintsAndAwait(HintsReader.Page page)
          *
          * If that is not the case, we'll need to perform conversion to a newer (or an older) format, and decoding the hint
          * is an unavoidable intermediate step.
+         *
+         * If Accord is enabled or these hints are from the batchlog and were originally attempted on Accord then
+         * we also need to decode so we can route the Hint contents appropriately.
          */
-        Action action = reader.descriptor().messagingVersion() == messagingVersion
+        Action action = reader.descriptor().messagingVersion() == messagingVersion && !DatabaseDescriptor.getAccordTransactionsEnabled() && !hostId.equals(RETRY_ON_DIFFERENT_SYSTEM_UUID)
```
bdeggleston: Always skipping the buffer sending path if accord is enabled seems excessive for what (I think) is a situation we'd only encounter while migrating. I think we could wait for the hint recipient to complain or make it possible to efficiently determine if a serialized hint belongs to a table that needs accord support

src/java/org/apache/cassandra/hints/HintsDispatcher.java
```diff
@@ -194,38 +307,106 @@ private <T> Action sendHints(Iterator<T> hints, Collection<Callback> callbacks,
         return Action.CONTINUE;
     }
 
+    // TODO (review): Could add the loop, but how often will it be needed?
```
bdeggleston: agreed looping here would be excessive

src/java/org/apache/cassandra/hints/HintsDispatcher.java
bdeggleston: Alternatively, what if the hint recipient took over accord coordination instead of sending it back to the hint dispatcher? It does something similar when it receives a hint for a key it no longer replicates

src/java/org/apache/cassandra/hints/HintsDispatcher.java
```diff
@@ -165,10 +226,62 @@ private Action sendHintsAndAwait(HintsReader.Page page)
...
+    {
+        ClusterMetadata cm = ClusterMetadata.current();
+        Hint hint;
+        while ((hint = hintsNeedingRehinting.poll()) != null)
+        {
+            HintsService.instance.writeForAllReplicas(hint);
+            Mutation mutation = hint.mutation;
+            // Also may need to apply locally because it's possible this is from the batchlog
+            // and we never applied it locally
+            // TODO (review): Additional error handling necessary? Hints are lossy
```
bdeggleston: agreed, they're best effort

src/java/org/apache/cassandra/hints/HintsDispatcher.java
```diff
@@ -194,38 +307,106 @@ private <T> Action sendHints(Iterator<T> hints, Collection<Callback> callbacks,
...
 
-        private Callback(long hintCreationTimeMillisSinceEpoch)
+        private Callback(@Nullable InetAddressAndPort to, long hintCreationTimeMillisSinceEpoch, long accordTxnStartNanos, @Nullable Pair<TxnId, Future<TxnResult>> accordTxnResult)
         {
+            this.to = to != null ? to : ACCORD_HINT_ENDPOINT;
             this.hintCreationNanoTime = approxTime.translate().fromMillisSinceEpoch(hintCreationTimeMillisSinceEpoch);
+            this.accordTxnStartNanos = accordTxnStartNanos;
+            this.accordTxnResult = accordTxnResult;
+            if (accordTxnResult != null)
+                accordTxnResult.right.addListener(this, ImmediateExecutor.INSTANCE);
```
bdeggleston: shouldn't we register as a result/throwable consumer so we are notified if the txn fails unexpectedly?

src/java/org/apache/cassandra/service/consensus/TransactionalMode.java
```diff
@@ -112,6 +112,10 @@ public ConsistencyLevel commitCLForStrategy(ConsistencyLevel consistencyLevel)
         return consistencyLevel;
     }
 
+    // TODO (required): This won't work for migration directly from none to full because there is no safe system to read from
```
bdeggleston: What a migration from `none` to `full` effectively ran accord in `mixed_reads` mode during migration?

src/java/org/apache/cassandra/service/AbstractWriteResponseHandler.java
```diff
@@ -294,8 +329,20 @@ public void onFailure(InetAddressAndPort from, RequestFailure failure)
         if (blockFor() + n > candidateReplicaCount())
             signal();
 
-        if (hintOnFailure != null && StorageProxy.shouldHint(replicaPlan.lookup(from)))
-            StorageProxy.submitHint(hintOnFailure.get(), replicaPlan.lookup(from), null);
+        // If the failure was RETRY_ON_DIFFERENT_TRANSACTION_SYSTEM then we only want to hint once
```
bdeggleston: wouldn't we want to retry with accord instead of writing a hint?

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
```diff
@@ -0,0 +1,399 @@
...
+                    cfs.metric.mutationsRejectedOnWrongSystem.mark();
+                logger.debug("Rejecting mutation on wrong system to table {}.{}", cfs.keyspace.getName(), cfs.name);
+                Tracing.trace("Rejecting mutation on wrong system to table {}.{} token {}", cfs.keyspace.getName(), cfs.name, dk.getToken());
+            }
+        }
+        if (throwRetryOnDifferentSystem)
+            throw new RetryOnDifferentSystemException();
+    }
+
+    public static boolean tokenShouldBeWrittenThroughAccord(@Nonnull ClusterMetadata cm, @Nonnull TableMetadata tm, @Nonnull Token token)
```
bdeggleston: We should be passing in the `TableId` and using it to retrieve the table metadata from the provided `ClusterMetadata`. Passing in table metadata directly like this makes it possible to read table params from a different epoch than the cm passed in, which I think is possible in invocation in `BlockingReadRepair`

bdeggleston (changes requested):

src/java/org/apache/cassandra/service/accord/api/AccordAgent.java
aweisberg: Will rename it to `emptySystemTxn` and add commments. Will also update the tests to stop using this.

src/java/org/apache/cassandra/db/CassandraKeyspaceWriteHandler.java
aweisberg: Oh yikes... you are right this should be inverted. Looks like there is no test for this and no implementation of `Memtable.Factory` that skip the commit log.

src/java/org/apache/cassandra/db/IMutation.java
aweisberg: I'll admit it's not a fun concept. Reminder to get back to this todo.

test/unit/org/apache/cassandra/cql3/PagingTest.java
aweisberg: Not supposed to be included, I think at the time I was checking to see how it was planned if you specified the starting point. I'll revert it.

src/java/org/apache/cassandra/batchlog/BatchlogManager.java
aweisberg: Will make them static

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
aweisberg: When splitting we don't know what portion will end up in accord and non-Accord so if `mutations.size()` is larger than 10 (default empty `ArrayList` size) only allocate 10 and then let `ArrayList` resizing take care of the rest. Specifying the allocation size is to reduce it below 10 (better than just using no-arg `ArrayList`).

I fussed with these loops a bunch of times trying to make them low/no allocation and then sort of gave up and just tried to make the common case (array list size 1) relatively minimal even if it isn't free. The simplest optimization is probably just to check for the single mutation routed to a single place case and return for that without allocating. I'll also add that.

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
aweisberg: We can, I was trying to not further explode the amount of code that is in `StorageProxy`. Is there any way we can accomplish that?

I actually want to reduce the scope of the change for `mutateAtomic` that is in `StorageProxy` as well.

src/java/org/apache/cassandra/service/accord/AccordService.java
aweisberg: It's double marking the metric, because `coordinateAsync` in the callback also marks the timeout. It seems like we need to be consistent about responsibility for firing the timeout. Should it be `getTxnResult` or `coordinateAsync`?

I am thinking the highest level is where we should do it because that is where we are blocking and Accord might fail to fire the timeout?

aweisberg: An now I remember why this was tricky, the issue is how I did CAS where it maps the future and never calls `getTxnResult`. I need to fix that so it goes through the `getTxnResult` code so they share this error handling and metrics.

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
aweisberg: I do it this way just because I don't normally ask the question "does this code throw", which in this case is `if (normalMutations != null)` just because putting something in a `try` block is free and then later changes that add/remove exceptions still get whatever the intended exception handling is.

So the example here would be if `normalMutations != null` were replaced with something more complex in the future that could NPE or `ClassCastException` it would still be handled.

The outer try block in this loop doesn't handle all exception types, and we can't rely on it because there are two sources of exceptions and we always want to get both (regular mutation path, accord path) before going through the loop again.

TL;DR I would argue that when using `try/catch/finally` the default approach should be to have everything possible inside the the `try` block since the cost of failing to handle exceptions is > than the cost of the extra indentation.

src/java/org/apache/cassandra/service/consensus/migration/ConsensusRequestRouter.java
aweisberg: CAS is supported on non-distributed keyspace (ya really) so this is all about preserving support for it. Since it's not distributed and Accord doesn't support non-distributed keyspaces this is letting Paxos continue to provide that support.

There was a bug where we could try to route to keyspaces that are not distributed and we would not return the right error here so this is differentiating between when it's an error and the keyspace doesn't exist and when it's just that it's not a distributed keyspace. I tried to standardize on this `metadata` and `getTableMetadata` method so we consistently perform this check.

aweisberg: Paxos racing is acceptable because Paxos (and even non-SERIAL) can misroute. The ClusterMetadata check itself is inherently racy and could miss the fact that other nodes have adopted this table as an Accord table.

I'll add a few more checks to confirm the keyspace we find is actually local, but they don't eliminate this race and we can still pick paxos on the coordinator when the `ClusterMetadata` has been updated to claim it's an Accord table.

src/java/org/apache/cassandra/hints/HintsDispatcher.java
aweisberg: If you are migrating to Accord you won't be using hints for very long so I didn't think it was worth optimizing. There is also the issue of having to deliver hints from the batchlog that aren't for any specific host so we would need implementation and both sender and receiver to do the routing.

I was happy to leave it as a wart, but yeah it's definitely possible to improve it further. I think responding back if hints can't be applied is probably the way to go so all the logic remains here and we can have the same logic for resending hints that need to be split and sending hints that are from the batchlog.

src/java/org/apache/cassandra/hints/HintsDispatcher.java
aweisberg: Except they are how the batchlog does retries and we go to surprising lengths to preserve hints. I recall seeing that we stream hints to replacement nodes.

src/java/org/apache/cassandra/hints/HintsDispatcher.java
aweisberg: We get the error either way because in `run` it calls `getTxnResult` which does the error handling. Not sure it's worth refactoring this in particular because it makes implementing `getTxnResult` a little more complicated since it would have to work as both a callback and as something invoked on a promise.

src/java/org/apache/cassandra/service/consensus/TransactionalMode.java
aweisberg: Yeah that is what it would need to do and it's what I implemented later when splitting into two phases. `TransactionalMode` takes in a `ClusterMetadata` and `Token` and makes a determination about `ConsistencyLevel` that is specific to whether the range is the `Token` is in has completed migration so it runs the appropriate interop `ConsistencyLevel` until it is no longer needed.

src/java/org/apache/cassandra/service/AbstractWriteResponseHandler.java
aweisberg: TBH I am having trouble remembering why I went to the trouble of doing this here. The purpose of this code was mostly to reduce the number of redundant Accord transactions created by the coordinator not to change the behavior of whether a hint is created or not.

This doesn't mean the entire request failed with RETRY_ON_DIFFERENT_SYSTEM it just means we got one response of such and it could still succeed or fail at other nodes for other reasons, succeed overall entirely, or write regular hints for some nodes that return other error responses.

Part of the reason hints are written here is that the callback may return before timeouts/responses come from all nodes so you can't rely on the callback completion to then figure out which hints to write. I believe this was the race condition I ran into.

The direction of the migration here isn't guaranteed. It could be that it was rejected during a migration away from Accord because the node that we asked to mutate thought it should have been on Accord. If we don't hint then something like ANY doesn't do the right thing.

With Accord ANY isn't really honored at the moment so you get an error instead of a hint written at the coordinator.

That's a very long way of saying I didn't think too hard here to try and prove it was safe to not hint since it should be a rare occurrence and I figured that hinting was better than not hinting.

It's possible to optimize it a little more by having the Accord hint written only if the overall outcome is retry on different system, but I didn't think this was something worth optimizing much since it is only an issue during a very brief window where nodes disagree on the correct system to use along with a concurrent write to that range.

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
aweisberg: That forces redundant map lookups if it is the approach we take everywhere. If we went to the trouble of looking up the `TableMetadata` shouldn't we keep using it?

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
aweisberg: Replaced a bunch of usage of Pair with dedicated container types.

src/java/org/apache/cassandra/service/BatchlogResponseHandler.java
aweisberg: I ended up refactoring this to remove `deferCleanup` entirely since `Cleanup` can be interacted with more directly to account for the actual number of wrappers that need to be created + 1 ack for the Accord txn.

src/java/org/apache/cassandra/service/StorageProxy.java
aweisberg: I was able to refactor this pretty cleanly to use a consumer for the split mutations to either aggregate into pair of mutation lists or in the batch case into a list and some the wrappers.

src/java/org/apache/cassandra/service/accord/AccordService.java
aweisberg: I refactored this so that everything goes through `getTxnResult` and the path in CAS/serial read that wants `ConsensusAttemptResult` uses a conversion method on the `TxnResult` returned by `getTxnResult`. I do like having that extra step just because `TxnResult` is a bit of a wonky class at this point that works in a very proprietary way depending on what kind of transaction we are running (transaction statement, range read, key read, CAS update).

The range/key reads are in a commit after this PR.

src/java/org/apache/cassandra/service/StorageProxy.java
aweisberg: Changed this to use `augmented` if `augemented` is not null.

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
bdeggleston: Not if it can break correctness like it can here, that part definitely needs to be fixed. Unless we can assert that the TableMetadata object is from the same epoch as the ClusterMetadata object, we shouldn't be doing this.

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
bdeggleston: it's a 3400 line file, what's 50 more lines? ;)

Seriously though, I think having StorageProxy be a bit longer is better than the vanilla mutation path taking a detour through this class. +1 to splitting up StorageProxy though

src/java/org/apache/cassandra/service/accord/AccordService.java
bdeggleston: I just didn't know what `j.u.c` meant, although now I realize it's `java.util.concurrent`

src/java/org/apache/cassandra/service/consensus/migration/ConsensusMigrationMutationHelper.java
aweisberg: OK, WFM. Will change it and keep an eye out for similar instances.

bdeggleston (approved):
+1

PR #3408, opened 2024-07-08
Bring back Journal simulator (w/o Accord at least for now); add semap…

ifesdjeen:
…hore interceptor.


test/simulator/main/org/apache/cassandra/simulator/systems/InterceptorOfGlobalMethods.java
```diff
@@ -383,12 +404,12 @@ public static CountDownLatch newCountDownLatch(int count)
...
         public static Semaphore newSemaphore(int count)
         {
-            return new Standard(count, false);
+            return methods.newSemaphore(count);
         }
 
         public static Semaphore newFairSemaphore(int count)
         {
-            return new Standard(count, true);
+            return methods.newSemaphore(count);
```
dcapwell:
```suggestion
            return methods.newFairSemaphore(count);
```

this is doing `false`

test/simulator/test/org/apache/cassandra/simulator/test/SimulationTestBase.java
```diff
@@ -229,10 +230,16 @@ public static void simulate(IIsolatedExecutor.SerializableRunnable[] runnables,
...
+                                                                                       NemesisFieldSelectors.get(),
+                                                                                       ClassLoader.getSystemClassLoader(),
+                                                                                       sharedClassPredicate.negate())::apply);
 
         ThreadGroup tg = new ThreadGroup("test");
-        InterceptorOfGlobalMethods interceptorOfGlobalMethods = new InterceptingGlobalMethods(NONE, null, failures, random);
+        InterceptedWait.CaptureSites.Capture capture = new InterceptedWait.CaptureSites.Capture(false, false, false);
+        InterceptorOfGlobalMethods interceptorOfGlobalMethods = IsolatedExecutor.transferAdhoc((IIsolatedExecutor.SerializableQuadFunction<InterceptedWait.CaptureSites.Capture, LongConsumer, Consumer<Throwable>, RandomSource, InterceptorOfGlobalMethods>) InterceptingGlobalMethods::new, classLoader)
+                                                                                .apply(capture, (ignore) -> {}, failures, random);
```
dcapwell, line 262: not familiar enough, can you explain this change?

dcapwell (commented):
overall makes sense, should be able to wrap up review tomorrow

test/simulator/test/org/apache/cassandra/simulator/test/SimulationTestBase.java
ifesdjeen: Essentially, without this, we are not properly intercepting some of the things, like a `WaitQueue` and `Semaphore` so they would throw something like below. I do not know how previous test worked without this, but by he time I uncommented it was already throwing.

```
java.lang.NullPointerException: null
	at org.apache.cassandra.simulator.systems.InterceptorOfGlobalMethods$Global.ifIntercepted(InterceptorOfGlobalMethods.java:437)
	at org.apache.cassandra.simulator.systems.InterceptingWaitQueue.register(InterceptingWaitQueue.java:52)
	at org.apache.cassandra.journal.Journal.awaitAvailableSegment(Journal.java:468)
	at org.apache.cassandra.journal.Journal.advanceSegment(Journal.java:454)
	at org.apache.cassandra.journal.Journal.start(Journal.java:157)
	at org.apache.cassandra.simulator.test.AccordJournalSimulationTest.check(AccordJournalSimulationTest.java:100)
	at org.apache.cassandra.simulator.test.AccordJournalSimulationTest.lambda$simpleRWTest$81c80a4a$2(AccordJournalSimulationTest.java:95)
	at org.apache.cassandra.simulator.systems.InterceptedExecution$InterceptedThreadStart.run(InterceptedExecution.java:216)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

```
java.lang.NullPointerException: null
	at org.apache.cassandra.simulator.systems.InterceptorOfGlobalMethods$Global.ifIntercepted(InterceptorOfGlobalMethods.java:437)
	at org.apache.cassandra.simulator.systems.InterceptingAwaitable$InterceptingCondition.maybeIntercept(InterceptingAwaitable.java:137)
	at org.apache.cassandra.simulator.systems.InterceptingAwaitable.awaitUninterruptibly(InterceptingAwaitable.java:97)
	at org.apache.cassandra.simulator.test.TrivialSimulationTest.lambda$semaphoreTest$81c80a4a$1(TrivialSimulationTest.java:119)
	at org.apache.cassandra.simulator.systems.InterceptedExecution$InterceptedThreadStart.run(InterceptedExecution.java:216)
```

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptorOfGlobalMethods.java
ifesdjeen: Well spotted, thanks!

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
```diff
@@ -0,0 +1,180 @@
...
+            return super.permits();
+
+        return permits.get();
+    }
+
+    @Override
+    public int drain()
+    {
+        if (ifIntercepted() == null)
+            return super.permits();
```
dcapwell:
```suggestion
            return super.drain();
```

?

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
```diff
@@ -0,0 +1,180 @@
...
+        assert res;
+        return current;
+    }
+
+    @Override
+    public void release(int release)
+    {
+        if (ifIntercepted() == null)
+        {
+            super.release();
```
dcapwell:
```suggestion
            super.release(release);
```

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
```diff
@@ -0,0 +1,180 @@
...
+            return;
+        }
+
+        int current = permits.get();
+        permits.compareAndSet(current, current + release);
+        int remaining = permits.get();
```
dcapwell:
```suggestion
        int remaining = permits.addAndGet(release);
```

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
```diff
@@ -0,0 +1,180 @@
...
+    @Override
+    public boolean tryAcquire(int acquire)
+    {
+        if (ifIntercepted() == null)
+            return super.tryAcquire(acquire);
+
+        int current = permits.get();
+        if (current >= acquire)
+        {
+            return permits.compareAndSet(current, current - acquire);
```
dcapwell: no loop needed?  We would enqueue even though we could have acquired?

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
```diff
@@ -0,0 +1,180 @@
...
+        {
+            return false;
+        }
+    }
+
+    @Override
+    public boolean tryAcquire(int acquire, long time, TimeUnit unit) throws InterruptedException
+    {
+        if (ifIntercepted() == null)
+            return super.tryAcquire(acquire);
```
dcapwell:
```suggestion
            return super.tryAcquire(acquire, time, unit);
```

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
```diff
@@ -0,0 +1,180 @@
...
+
+        int current = permits.get();
+        if (current >= acquire)
+            return permits.compareAndSet(current, current - acquire);
```
dcapwell: same as above, no loop needed?

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
```diff
@@ -0,0 +1,180 @@
...
+
+    @Override
+    public boolean tryAcquireUntil(int acquire, long nanoTimeDeadline) throws InterruptedException
+    {
+        if (ifIntercepted() == null)
+            return super.tryAcquireUntil(acquire, nanoTimeDeadline);
+
+        int current = permits.get();
+        if (current >= acquire)
+            return permits.compareAndSet(current, current - acquire);
```
dcapwell: shouldn't this be a loop?  if we fail to update because something else we *should* try again no?

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
```diff
@@ -0,0 +1,180 @@
...
+            signal.await();
+        }
+    }
+
+    @Override
+    public void acquireThrowUncheckedOnInterrupt(int acquire) throws UncheckedInterruptedException
+    {
+        try
+        {
+            acquire();
```
dcapwell:
```suggestion
            acquire(acquire);
```

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
```diff
@@ -18,250 +18,359 @@
...
-        simulate(arr(() -> run()),
-                 () -> check());
+        simulate(arr(() -> {
+                    ListenableFileSystem fs = new ListenableFileSystem(Jimfs.newFileSystem());
+                    File.unsafeSetFilesystem(fs);
+                    DatabaseDescriptor.daemonInitialization();
+                    DatabaseDescriptor.setCommitLogCompression(new ParameterizedClass("LZ4Compressor", ImmutableMap.of())); //
+                    DatabaseDescriptor.setCommitLogWriteDiskAccessMode(Config.DiskAccessMode.standard);
+                    DatabaseDescriptor.initializeCommitLogDiskAccessMode();
+                    System.out.println("DatabaseDescriptor.getCommitLogWriteDiskAccessMode() = " + DatabaseDescriptor.getCommitLogWriteDiskAccessMode());
```
dcapwell: Logger please

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
```diff
@@ -18,250 +18,359 @@
...
+
+                    DatabaseDescriptor.setPartitionerUnsafe(Murmur3Partitioner.instance);
+                    long seed = 1L;
```
dcapwell: dead code?

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
```diff
@@ -18,250 +18,359 @@
...
+                    new File("/journal").createDirectoriesIfNotExists();
+
+                    DatabaseDescriptor.setDumpHeapOnUncaughtException(false);
+
+                    Keyspace.setInitialized();
+
+                    CMSTestBase.CMSSut sut = new CMSTestBase.CMSSut(AtomicLongBackedProcessor::new,
+                            false,
+                            new FakeSchema(),
+                            new TokenPlacementModel.SimpleReplicationFactor(1));
```
dcapwell: dead code?

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
```diff
@@ -18,250 +18,359 @@
...
+                            new AccordSpec.JournalSpec(),
+                            new TestCallbacks(),
+                            new IdentityKeySerializer(),
+                            new IdentityValueSerializer());
+                }),
+                () -> check());
     }
 
-    private static void run()
+    public static void check()
```
dcapwell, line 87: having the check be the test and not the validation is much cleaner, good idea

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
```diff
@@ -18,250 +18,359 @@
...
+            {
+                int finalI = i;
+                State.executor.submit(() -> State.journal.asyncWrite("test" + finalI, "test" + finalI, Collections.singleton(1), null));
+            }
+
+            State.latch.await();
+
+            for (int i = 0; i < 100; i++)
+            {
+                System.out.println("Reading " + State.journal.readFirst("test" + i));
```
dcapwell: Logger please

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
```diff
@@ -18,250 +18,359 @@
...
+                Assert.assertEquals(State.journal.readFirst("test" + i), "test" + i);
+            }
+        }
+        catch (Throwable e)
         {
-            int finalI = i;
-            State.executor.execute(() -> State.append(finalI));
+            e.printStackTrace();
```
dcapwell: write to logger please

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
```diff
@@ -18,250 +18,359 @@
...
-            State.exceptions.forEach(error::addSuppressed);
-            throw error;
         }
 
-        State.journal.shutdown();
-        State.logger.info("Run complete");
+        @Override
+        public void onFlushFailed(Throwable cause)
+        {
+            new RuntimeException("Could not flush", cause).printStackTrace();
```
dcapwell: this is a big issue as the test could pass and this *will* be ignored.  You should save this to `State` and assert this in the test.  Any failures is unexpected and should lead to a test fault

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
```diff
@@ -18,250 +18,359 @@
...
-            State.logger.info("All events are durable done!");
+            State.latch.decrement();
         }
-        catch (InterruptedException e)
+
+        @Override
+        public void onWriteFailed(String key, String value, Object writeContext, Throwable cause)
         {
-            throw new AssertionError(e);
+            State.latch.decrement();
```
dcapwell, line 139: this is a big issue as the test could pass and this will be ignored. You should save this to `State` and assert this in the test. Any failures is unexpected and should lead to a test fault

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
```diff
@@ -18,250 +18,359 @@
...
+        {
+            int size = in.readInt();
+            byte[] key = new byte[size];
+            for (int i = 0; i < size; i++)
+                key[i] = in.readByte();
+
+            int maxSize = 16 - TypeSizes.INT_SIZE;
+            int remaining = maxSize - size;
+            for (int i = 0; i < remaining; i++)
+                in.readByte(); // todo assert 0
```
dcapwell: want to fix the todo?  Also `0` is common so could be good to use another value like `4` (randomly selected from XKCD)

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
```diff
@@ -18,250 +18,359 @@
...
+        {
+            int size = buffer.getInt();
+            byte[] key = new byte[size];
+            for (int i = 0; i < size; i++)
+                key[i] = buffer.get();
+
+            int maxSize = 16 - TypeSizes.INT_SIZE;
+            int remaining = maxSize - size;
+            for (int i = 0; i < remaining; i++)
+                buffer.get(); // todo assert 0
```
dcapwell: want to fix the todo?  Also `0` is common so could be good to use another value like `4` (randomly selected from XKCD)

dcapwell (changes requested):

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
ifesdjeen: Good point; removed.

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
ifesdjeen: removed; probably cleaner this way

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
ifesdjeen: good point!

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
ifesdjeen: switched to add loop here

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
ifesdjeen: switched to add a loop

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
ifesdjeen: On closer inspection, you are right, since this is not `AccordJournal` we do not need to set up schema here!

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
ifesdjeen: switched to just collect and throw

test/simulator/test/org/apache/cassandra/simulator/test/AccordJournalSimulationTest.java
dcapwell: also goes to the right log file, simulator has its own logger configs =)

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
```diff
@@ -0,0 +1,187 @@
...
+        throw new IllegalStateException("Too much contention");
+    }
+
+    @Override
+    public boolean tryAcquire(int acquire, long time, TimeUnit unit) throws InterruptedException
+    {
+        if (ifIntercepted() == null)
+            return super.tryAcquire(acquire, time, unit);
+
+        return tryAcquireUntil(acquire, Clock.Global.nanoTime() + unit.toNanos(time));
```
dcapwell: `Clock.Global.nanoTime()` is this right in the context of Simulator?  This feels wrong to me...

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
```diff
@@ -0,0 +1,187 @@
...
+            if (current >= acquire)
+            {
+                if (permits.compareAndSet(current, current - acquire))
+                    return true;
+            }
+            SemaphoreSignal signal = new SemaphoreSignal(acquire);
+            interceptible.add(signal);
+            signal.awaitUntil(deadline);
+        }
+        while (Clock.Global.nanoTime() < deadline);
```
dcapwell: `Clock.Global.nanoTime()` feels wrong as time is owned by Simulator, so the logical time could be multiple hours but the physical time is just a few ms, which would lead to this doing the wrong behavior (and being non-deterministic)?

dcapwell: if we waited until the deadline, then check again and not able to get the lock... shouldn't we stop right away?  If we loop again are we not doing it wrong?

test/simulator/test/org/apache/cassandra/simulator/test/TrivialSimulationTest.java
```diff
@@ -85,9 +97,97 @@ public void componentTest()
...
+    }
+
+    @Shared
+    public static class State
+    {
+        static final List<Tick> ticks = new ArrayList<>();
+        static boolean record = true;
+        static int i = 0;
+
+        public static void tick(int thread, int iteration)
```
dcapwell: is this thread safe?  I don't think so

ifesdjeen: It does not have to be thread safe. It is executed effectively in a single thread; this is also why we have same interleavings no matter how many times you run it.

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
ifesdjeen: BTW, loop is not needed. I am adding it only because it might be easier for people to reason about this. We will never have contention there, confirmed by @belliottsmith

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
ifesdjeen: My assumption was that`Clock.Global.nanoTime()` is intercepted and redirected to simulated time. I will double check if that's the case.

ifesdjeen: > if we waited until the deadline, then check again and not able to get the lock... shouldn't we stop right away? If we loop again are we not doing it wrong?

I don't think so. javadoc says "Try to take permits, waiting until the deadline", a opposed to "Take permits, waiting indefinitely until available". It does not specify you should wait for a single release. Default implementation does a `for (;;) {` loop.

test/simulator/main/org/apache/cassandra/simulator/systems/InterceptingSemaphore.java
ifesdjeen: I've double-checked with @belliottsmith, and using `nanoTIme` in the context of simulator is ok. That said, I did find a problem with the way we were simulating time for this specific test, and turned out that `System#nanoTime` was not intercepted. I've fixed this separately.

Also, for consistency of behaviors, I've switched to delegate time-related methods to awaitable that we already hold.

ifesdjeen (commented):
.

dcapwell (approved):
+1

PR #3409, opened 2024-07-08
Write side of the replay determinism: persist intermediate Command in…

ifesdjeen:
… the journal

      * Simplifies AccordJournal by removing Framing
      * Saves command outcomes to the log
      * Reconstructs latest command state from the log entries


src/java/org/apache/cassandra/journal/ActiveSegment.java
```diff
@@ -455,14 +459,18 @@ void write(K id, ByteBuffer record, Set<Integer> hosts)
...
             {
-                EntrySerializer.write(id, bytes, hosts, keySupport, out, descriptor.userVersion);
+                EntrySerializer.write(id, record, hosts, keySupport, out, descriptor.userVersion);
                 index.update(id, position);
                 metadata.update(hosts);
-                callbacks.onWrite(descriptor.timestamp, position, size, id, record, writeContext);
+            }
+            catch (IOException e)
+            {
+                throw new JournalWriteError(descriptor, file, e);
```
belliottsmith, line 473: Haven't thought about error handling here properly, but should rationalise with the callback handler that is notified of write failures. 

We also need to generally think about how we handle exceptions with respect to database state - do we try to rollback the state changes, or is it fine to just not send responses? _Probably_ the latter is sufficient, but we should add a `TODO (required)` to consider it more carefully and centrally document our assumptions alongside any protocol documentation.

PR #3411, opened 2024-07-09
CASSANDRA-19746: Update CQLSH documentation to reflect compatibility with Python 3.

Suhel0328:
This patch updates the CQLSH documentation to remove references to Python 2.7 and clarify compatibility with Python 3.
patch by Mohammad Suhel; for CASSANDRA-19746

### DIFF
--- a/doc/modules/cassandra/pages/managing/tools/cqlsh.adoc
+++ b/doc/modules/cassandra/pages/managing/tools/cqlsh.adoc
@@ -7,7 +7,7 @@ executable.
 
 == Compatibility
 
-`cqlsh` is compatible with Python 3.6 and later.
+`cqlsh` is compatible with Python 3.
 
 In general, a given version of `cqlsh` is only guaranteed to work with the
 version of Cassandra that it was released with.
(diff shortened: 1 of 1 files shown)


bschoening:
I think this should say simply Python 3 and not mention a specific Python release.  The current Python driver supports Python 3.8 - 3.12.

Suhel0328:
I have updated the documentation to state that `cqlsh` is compatible with Python 3, as requested. Please review the changes.

bschoening:
It looks like this was just a problem with the 4.1 docs, and that trunk had already been fixed.  I think the current (trunk) section on compatibility is fine as-is, having already removed 2.7, and it's not likely worth fixing it on the 4.1 branch of the docs.

I'd therefore suggest we close this PR and Jira as won't fix.

Suhel0328:
Thank you for the review and feedback.

I understand that the trunk branch has already addressed the Python 2.7 references, and updating the 4.1 branch documentation isn't necessary at this point.

I will close this PR and the associated Jira issue as "won't fix." 

Thanks again for the guidance.

PR #3412, opened 2024-07-09
CASSANDRA-19760 Expose the execution results of a cleanup command to a virtual table

Mmuzaf:
 
```

```

### DIFF
--- a/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
+++ b/src/java/org/apache/cassandra/db/compaction/CompactionManager.java
@@ -18,6 +18,7 @@
 package org.apache.cassandra.db.compaction;
 
 import java.io.IOException;
+import java.util.AbstractMap;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -27,14 +28,21 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.NavigableMap;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentNavigableMap;
+import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.BiFunction;
 import java.util.function.BooleanSupplier;
+import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
@@ -48,6 +56,7 @@ import com.google.common.collect.ArrayListMultimap;
 import com.google.common.collect.Collections2;
 import com.google.common.collect.ConcurrentHashMultiset;
 import com.google.common.collect.Iterables;
+import com.google.common.collect.Iterators;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Multimap;
@@ -57,6 +66,7 @@ import com.google.common.util.concurrent.RateLimiter;
 import com.google.common.util.concurrent.Uninterruptibles;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
 import com.codahale.metrics.Meter;
 import net.openhft.chronicle.core.util.ThrowingSupplier;
 import org.apache.cassandra.cache.AutoSavingCache;
@@ -108,6 +118,8 @@ import org.apache.cassandra.schema.TableMetadata;
 import org.apache.cassandra.service.ActiveRepairService;
 import org.apache.cassandra.service.StorageService;
 import org.apache.cassandra.streaming.PreviewKind;
+import org.apache.cassandra.tcm.ClusterMetadata;
+import org.apache.cassandra.tcm.ownership.DataPlacement;
 import org.apache.cassandra.utils.FBUtilities;
 import org.apache.cassandra.utils.JVMStabilityInspector;
 import org.apache.cassandra.utils.MBeanWrapper;
@@ -115,8 +127,6 @@ import org.apache.cassandra.utils.OutputHandler;
 import org.apache.cassandra.utils.Throwables;
 import org.apache.cassandra.utils.TimeUUID;
 import org.apache.cassandra.utils.WrappedRunnable;
-import org.apache.cassandra.tcm.ClusterMetadata;
-import org.apache.cassandra.tcm.ownership.DataPlacement;
 import org.apache.cassandra.utils.concurrent.Future;
 import org.apache.cassandra.utils.concurrent.ImmediateFuture;
 import org.apache.cassandra.utils.concurrent.Refs;
@@ -124,6 +134,7 @@ import org.apache.cassandra.utils.concurrent.Refs;
 import static java.util.Collections.singleton;
 import static org.apache.cassandra.concurrent.ExecutorFactory.Global.executorFactory;
 import static org.apache.cassandra.concurrent.FutureTask.callable;
+import static org.apache.cassandra.config.CassandraRelevantProperties.CASSANDRA_OPERATIONS_HISTORY_SIZE;
 import static org.apache.cassandra.config.DatabaseDescriptor.getConcurrentCompactors;
 import static org.apache.cassandra.db.compaction.CompactionManager.CompactionExecutor.compactionThreadGroup;
 import static org.apache.cassandra.service.ActiveRepairService.NO_PENDING_REPAIR;
@@ -174,6 +185,8 @@ public class CompactionManager implements CompactionManagerMBean, ICompactionMan
     final Multiset<ColumnFamilyStore> compactingCF = ConcurrentHashMultiset.create();
 
     public final ActiveCompactions active = new ActiveCompactions();
+    private final ConcurrentNavigableMap<TimeUUID, OperationInfo> operations =
+        new BoundedConcurrentSkipListMap(CASSANDRA_OPERATIONS_HISTORY_SIZE.getInt());
 
     // used to temporarily pause non-strategy managed compactions (like index summary redistribution)
     private final AtomicInteger globalCompactionPauseCount = new AtomicInteger(0);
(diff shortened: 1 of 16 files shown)


test/unit/org/apache/cassandra/db/virtual/StorageOperationsTableCleanupTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+import org.apache.cassandra.db.CleanupTest;
+import org.apache.cassandra.db.compaction.CompactionManager;
+import org.apache.cassandra.db.compaction.OperationType;
+import org.apache.cassandra.schema.SchemaConstants;
+import org.apache.cassandra.tools.ToolRunner;
+
+import static org.apache.cassandra.tools.ToolRunner.invokeNodetool;
+
+public class StorageOperationsTableCleanupTest extends CleanupTest
```
smiklosovic: This is an interesting way of doing it. Honestly I have never encountered something like this. I understand what you are trying to do here but I think this is not a good practice. Each test should be just standalone or extend some hierarchy of abstract class(es). Overriding the test methods is really strange to my taste. I do not know yet how to approach this differently, looking into `CleanupTest`, there is a lot of private methods which might be `protected` and each test method would be repeated here and you would add the bits you just added after you called test methods from `CleanupTest` into them.

src/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -2578,16 +2578,14 @@ public int forceKeyspaceCleanup(int jobs, String keyspaceName, String... tableNa
...
         logger.info("Starting {} on {}.{}", OperationType.CLEANUP, keyspaceName, Arrays.toString(tableNames));
-        for (ColumnFamilyStore cfStore : getValidColumnFamilies(false, false, keyspaceName, tableNames))
-        {
-            CompactionManager.AllSSTableOpStatus oneStatus = cfStore.forceCleanup(jobs);
-            if (oneStatus != CompactionManager.AllSSTableOpStatus.SUCCESSFUL)
-                status = oneStatus;
-        }
-        logger.info("Completed {} with status {}", OperationType.CLEANUP, status);
-        return status.statusCode;
+        CompactionManager.AllSSTableOpStatus result = CompactionManager.instance
```
smiklosovic: would you mind to fix the alignement?

src/java/org/apache/cassandra/db/virtual/model/StorageOperationsRow.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Column
+    public String processedByKeyspace()
+    {
+        return formatMap(entry.getValue().processedSStables());
+    }
+
+    private static String formatMap(Map<?, ?> map)
+    {
+        return format("[%s]", map.entrySet().stream()
```
smiklosovic: would you mind to rewrite this to for loops?

https://lists.apache.org/thread/65glsjzkmpktzmns6j9wvr4nczvskx36

src/java/org/apache/cassandra/db/virtual/SystemViewsKeyspace.java
```diff
@@ -45,6 +50,12 @@ private SystemViewsKeyspace()
                                                               new ThreadPoolRowWalker(),
                                                               Metrics.allThreadPoolMetrics(),
                                                               ThreadPoolRow::new))
+                    .add(CollectionVirtualTableAdapter.create(VIRTUAL_VIEWS,
+                                                              STORAGE_OPERATIONS,
+                                                              "Storage operations currently running or completed for the last 7 days",
```
smiklosovic: you hardcoded 7 days here not reflecting what a user might set it to

src/java/org/apache/cassandra/db/virtual/SystemViewsKeyspace.java
```diff
@@ -28,6 +31,8 @@
 
 public final class SystemViewsKeyspace extends VirtualKeyspace
 {
+    public static final String STORAGE_OPERATIONS = "storage_operations_history";
```
smiklosovic: what other operations do you plan to add? now it seems like it is just about "compaction". Should not we reflect that? There is also "thread_pools" not on constant if we play the constants game here.

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -2537,4 +2608,104 @@ public interface CompactionPauser extends AutoCloseable
...
+        {
+            return result;
+        }
+
+        public Map<CFInfo, AllSSTableOpStatus> processedCfs()
+        {
+            return processedCfs;
+        }
+
+        public Map<String, AtomicInteger> processedSStables()
```
smiklosovic: `processedSSTablesPerKeyspace()`?

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -2537,4 +2608,104 @@ public interface CompactionPauser extends AutoCloseable
...
+        /** The unique id for this operation */
+        public final TimeUUID operationId = TimeUUID.Generator.nextTimeUUID();
+        /** The type of operation being performed */
+        public final OperationType operationType;
+        /** The keyspace names for which this operation is being performed */
+        public final Set<String> keyspaceNames = new ConcurrentSkipListSet<>();
+        /** The keyspace and table names for which this operation is being performed */
+        public final Map<CFInfo, AllSSTableOpStatus> processedCfs = new ConcurrentHashMap<>();
+        /** The number of sstables that have been completed for each keyspace */
+        public final Map<String, AtomicInteger> processedSStables = new ConcurrentHashMap<>();
```
smiklosovic: `numberOfSSTablesPerKeyspace`, `sstableCountPerKeyspace`, `processedSSTablesPerKeyspace` ... you got the idea

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -1219,45 +1267,57 @@ public void forceUserDefinedCleanup(String dataFiles)
...
+        for (Map.Entry<ColumnFamilyStore, List<Descriptor>> entry : descriptors.entrySet())
         {
             ColumnFamilyStore cfs = entry.getKey();
             Keyspace keyspace = cfs.keyspace;
             final RangesAtEndpoint replicas = StorageService.instance.getLocalReplicas(keyspace.getName());
             final Set<Range<Token>> allRanges = replicas.ranges();
             final Set<Range<Token>> transientRanges = replicas.onlyTransient().ranges();
             boolean hasIndexes = cfs.indexManager.hasIndexes();
-            SSTableReader sstable = lookupSSTable(cfs, entry.getValue());
+            List<SSTableReader> sstables = entry.getValue().stream().
```
smiklosovic: would you mind to rewrite this to for loops?

https://lists.apache.org/thread/65glsjzkmpktzmns6j9wvr4nczvskx36

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
```diff
@@ -174,6 +183,12 @@ public class CompactionManager implements CompactionManagerMBean, ICompactionMan
     final Multiset<ColumnFamilyStore> compactingCF = ConcurrentHashMultiset.create();
 
     public final ActiveCompactions active = new ActiveCompactions();
+    private final Cache<TimeUUID, OperationInfo> operations = CacheBuilder.newBuilder()
```
smiklosovic: I wanted to say that his should be "space-bound" instead of "time-bound" but I noticed that you actually did both which is great. If somebody goes crazy and they cleanup like mad then without the limit on size this would grow and grow without any control.

I would just like to know based on what you did it like 7 days and 300 entries. Is this empirical based on your workload? I think revisiting the decision about the figures here would be nice.

Also, does the cache preserve any insertion order? If you do operations.toMap(), do I get entries without any particular order in the virtual table? I would like them to be ordered the same way every single time.

smiklosovic (changes requested):

test/unit/org/apache/cassandra/db/virtual/StorageOperationsTableCleanupTest.java
Mmuzaf: Fixed this one.

src/java/org/apache/cassandra/service/StorageService.java
Mmuzaf: Fixed.

src/java/org/apache/cassandra/db/virtual/model/StorageOperationsRow.java
Mmuzaf: Fixed.

src/java/org/apache/cassandra/db/virtual/SystemViewsKeyspace.java
Mmuzaf: Fixed.

src/java/org/apache/cassandra/db/virtual/SystemViewsKeyspace.java
Mmuzaf: Updated to the `compaction_operations_history`

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
Mmuzaf: Fixed.

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
Mmuzaf: Fixed.

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
Mmuzaf: Fixed. However, I don't think this is a hot path :-)

src/java/org/apache/cassandra/db/compaction/CompactionManager.java
Mmuzaf: I tend to use "space-bound" only in favor of both options altogether. Both of them are configurable via the system settings, so it should be no problem to customize them for any type of deployment. Based on the SLA we have and my previous experience it should be fine having 7d and 300 entries, but of course we can update these numbers.

The insert order is timestamp-UUID, which is a primary key for the virtual table we are trying to add, so with a default query it will be sorted naturally. I've checked the caching API and haven't found any sorting capabilities, but I don't think it should matter for such a small table.

src/java/org/apache/cassandra/db/virtual/SystemViewsKeyspace.java
```diff
@@ -45,6 +50,13 @@ private SystemViewsKeyspace()
                                                               new ThreadPoolRowWalker(),
                                                               Metrics.allThreadPoolMetrics(),
                                                               ThreadPoolRow::new))
+                    .add(CollectionVirtualTableAdapter.create(VIRTUAL_VIEWS,
+                                                              COMPACTION_OPERATIONS_HISTORY,
+                                                              "Compaction operations that are currently " +
+                                                              "running or have been completed since the node start",
```
smiklosovic: @Mmuzaf `since the node started` or `since node's start`

Mmuzaf: Fixed.

PR #3416, opened 2024-07-12
CASSANDRA-19769: CEP-15: (Accord) sequence EpochReady.coordinating to allow syncComplete to be learned from newer epochs

dcapwell:


src/java/org/apache/cassandra/gms/FailureDetector.java
```diff
@@ -312,8 +312,8 @@ public boolean isAlive(InetAddressAndPort ep)
             // registration via the metadata log, or a full gossip round). This is perfectly harmless, so no need to log
             // an error in that case.
             ClusterMetadata metadata = ClusterMetadata.current();
-            if (!metadata.directory.allJoinedEndpoints().contains(ep) && !metadata.fullCMSMembers().contains(ep))
-                logger.error("Unknown endpoint: " + ep, new IllegalArgumentException("Unknown endpoint: " + ep));
+//            if (!metadata.directory.allJoinedEndpoints().contains(ep) && !metadata.fullCMSMembers().contains(ep))
```
dcapwell: will add back, way too much spam in the test logs making it hard to find *real* issues...

src/java/org/apache/cassandra/schema/DistributedMetadataLogKeyspace.java
```diff
@@ -149,7 +149,8 @@ public static boolean tryCommit(Entry.Id entryId,
         }
         catch (Throwable t)
         {
-            logger.error("Caught an exception while trying to CAS", t);
+            //TODO (now): add back... it spams the logs making the real errors get truncated
```
dcapwell: will add back, way too much spam in the test logs making it hard to find *real* issues...

src/java/org/apache/cassandra/service/accord/AccordCallback.java
```diff
@@ -46,7 +46,7 @@ public AccordCallback(AgentExecutor executor, Callback<T> callback, AccordEndpoi
     @Override
     public void onResponse(Message<T> msg)
     {
-        logger.debug("Received response {} from {}", msg.payload, msg.from());
+//        logger.debug("Received response {} from {}", msg.payload, msg.from());
```
dcapwell: will add back, way too much spam in the test logs making it hard to find *real* issues...

src/java/org/apache/cassandra/service/accord/AccordSafeCommandStore.java
```diff
@@ -246,28 +247,27 @@ private <O> O mapReduceForRange(Routables<?> keysOrRanges, Ranges slice, BiFunct
     {
         if (commandsForRanges == null)
             return accumulate;
+        CommandsForRanges cfr = commandsForRanges.current().slice(slice);
```
dcapwell: this was found by `TopologyMixupTest`.  The `CommandsForRanges` shows every txn that could be found on *this* node, but we are processing a `Shard`, which is a smaller range.  So need to post process and `slice` to the shard's subset

src/java/org/apache/cassandra/service/accord/AccordSafeCommandStore.java
```diff
@@ -204,6 +204,7 @@ public RangesForEpoch ranges()
     @Override
     public void registerHistoricalTransactions(Deps deps)
     {
+        if (deps.isEmpty()) return;
```
dcapwell: not required, just a common case

src/java/org/apache/cassandra/service/accord/AccordTopology.java
```diff
@@ -134,22 +138,34 @@ private static KeyspaceShard forRange(KeyspaceMetadata keyspace, Range<Token> ra
...
+                                                      Function<KeyspaceMetadata, DataPlacement> placements,
+                                                      Function<KeyspaceMetadata, DataPlacement> settledPlacements,
+                                                      Directory directory, ShardLookup lookup)
         {
-            ReplicationParams replication = keyspace.params.replication;
-            DataPlacement placement = placements.get(replication);
+//            ReplicationParams replication = keyspace.params.replication;
+            DataPlacement placement = placements.apply(keyspace);
+            DataPlacement settled = settledPlacements.apply(keyspace);
+            Set<InetAddressAndPort> leaving = Sets.difference(placement.writes.byEndpoint().keySet(), settled.writes.byEndpoint().keySet());
```
dcapwell: I should cleanup this logic, we are already using the `KeyspaceMetadata` so we really don't need functions, we can just pass in the 2

src/java/org/apache/cassandra/service/accord/txn/TxnWrite.java
```diff
@@ -261,7 +261,7 @@ public Update complete(AccordUpdateParameters parameters)
             if (!staticRow.isEmpty())
                 updateBuilder.add(staticRow);
 
-            Row existing = !baseUpdate.isEmpty() ? Iterables.getOnlyElement(baseUpdate) : null;
+            Row existing = baseUpdate.hasRows() ? Iterables.getOnlyElement(baseUpdate) : null;
```
dcapwell: found by `TopologyMixupTest`. The test used `UPDATE` on a row that didn't exist, but the table had a static column... the query was

```
UPDATE ...
SET
  a += 42 -- row doesn't exist, null + 42 = null
  b = 7 -- this is a static column, so creates an update
```

the update had a static row updated, but no non-static row!

dcapwell: maybe it would be good to be explicit about this in one of our tests rather than rely on a fuzz test... regression tests are great to avoid the same bugs happening again...

I should add...

src/java/org/apache/cassandra/tcm/PaxosBackedProcessor.java
```diff
@@ -87,7 +88,8 @@ public ClusterMetadata fetchLogAndWait(Epoch waitFor, Retry.Deadline retryPolicy
             {
                 JVMStabilityInspector.inspectThrowable(t);
                 TCMMetrics.instance.fetchCMSLogConsistencyDowngrade.mark();
-                logger.warn("Could not perform consistent fetch, downgrading to fetching from CMS peers.", t);
+                //TODO (now): add back... it spams the logs making the real errors get truncated
```
dcapwell: will add back, way too much spam in the test logs making it hard to find *real* issues...

src/java/org/apache/cassandra/tcm/PaxosBackedProcessor.java
```diff
@@ -136,7 +138,17 @@ public ClusterMetadata fetchLogAndWait(Epoch waitFor, Retry.Deadline retryPolicy
             while (iter.hasNext())
             {
                 FetchLogRequest request = iter.next();
-                if (request.condition.awaitUninterruptibly(Math.max(0, nextTimeout - Clock.Global.nanoTime()), TimeUnit.NANOSECONDS) &&
+                boolean awaitResult;
```
dcapwell: filed CASSANDRA-19761, this is a TCM issue and Sam said he will take a look... my work is kinda a hack more than anything, I just wanted to unblock my tests!

src/java/org/apache/cassandra/tcm/RemoteProcessor.java
```diff
@@ -190,7 +190,7 @@ public static <REQ, RSP> RSP sendWithCallback(Verb verb, REQ request, CandidateI
         {
             Promise<RSP> promise = new AsyncPromise<>();
             sendWithCallbackAsync(promise, verb, request, candidates, retryPolicy);
-            return promise.awaitUninterruptibly().get();
+            return promise.await().get();
```
dcapwell: see CASSANDRA-19761

src/java/org/apache/cassandra/tcm/log/LocalLog.java
```diff
@@ -678,7 +678,7 @@ public void runOnce(DurationSpec duration) throws InterruptedException, TimeoutE
                 {
                     if (duration == null)
                     {
-                        current.awaitUninterruptibly();
+                        current.await();
```
dcapwell: see CASSANDRA-19761

src/java/org/apache/cassandra/tcm/log/LocalLog.java
```diff
@@ -707,7 +707,7 @@ else if (!current.await(duration.to(TimeUnit.MILLISECONDS), TimeUnit.MILLISECOND
                 if (runnable.subscriber.compareAndSet(null, ours))
                 {
                     runnable.logNotifier.signalAll();
-                    ours.awaitUninterruptibly();
+                    ours.await();
```
dcapwell: see CASSANDRA-19761

src/java/org/apache/cassandra/utils/HeapUtils.java
```diff
@@ -130,7 +130,7 @@ public static String maybeCreateHeapDump()
                 }
                 else
                 {
-                    logger.debug("Heap dump creation on uncaught exceptions is disabled.");
+//                    logger.debug("Heap dump creation on uncaught exceptions is disabled.");
```
dcapwell: will add back, way too much spam in the test logs making it hard to find *real* issues...

test/distributed/org/apache/cassandra/distributed/impl/AbstractCluster.java
```diff
@@ -280,7 +281,7 @@ protected class Wrapper extends DelegatingInvokableInstance implements IUpgradea
         protected IInvokableInstance delegate()
         {
             if (delegate == null)
-                throw new IllegalStateException("Can't use shutdown instances, delegate is null");
+                throw new IllegalStateException("Can't use shutdown node" + config.num() + ", delegate is null");
```
dcapwell: we know the node, why not tell users!

test/distributed/org/apache/cassandra/distributed/impl/AbstractCluster.java
```diff
@@ -1095,7 +1096,27 @@ public void close()
...
                            .map(IInstance::shutdown)
                            .collect(Collectors.toList());
-        FBUtilities.waitOnFutures(futures,1L, TimeUnit.MINUTES);
+        try
+        {
+            FBUtilities.waitOnFutures(futures,1L, TimeUnit.MINUTES);
+        }
+        catch (RuntimeException t)
+        {
+            if (t.getCause() instanceof TimeoutException)
```
dcapwell: a timeout is really not useful as you don't have enough information to know what happened... this will try to show what threads are left so you *could* figure out what is blocked.  This has been so helpful to find the threads that are blocking!

test/distributed/org/apache/cassandra/distributed/impl/INodeProvisionStrategy.java
```diff
@@ -26,140 +26,69 @@
 
 import static org.apache.cassandra.utils.Shared.Recursive.INTERFACES;
 
+//TODO (now, correctness): one network changes the port where as multiple network changes the address... yet AbstractNodeProvisionStrategy always changes the port too!
 @Shared(inner = INTERFACES)
 public interface INodeProvisionStrategy
```
dcapwell: I need to double check things, I do think I caused a regression here... I did enough to get my test to do what I needed... I will need to look closer here

test/distributed/org/apache/cassandra/distributed/impl/IsolatedExecutor.java
```diff
@@ -282,7 +282,7 @@ public static <T> T waitOn(Future<T> f)
         }
         catch (ExecutionException e)
         {
-            throw Throwables.throwAsUncheckedException(e.getCause());
+            throw new RuntimeException(e.getCause());
```
dcapwell: asked in the dev list, no one complained... this will break tests so shouldn't be done in this patch... will revert when testing is done and file a ticket against our other branches

test/distributed/org/apache/cassandra/distributed/test/accord/AccordHostReplacementTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.cassandra.service.consensus.TransactionalMode;
+import org.apache.cassandra.utils.CassandraGenerators;
+import org.apache.cassandra.utils.CassandraGenerators.TableMetadataBuilder;
+
+import static accord.utils.Property.qt;
+import static org.apache.cassandra.distributed.shared.ClusterUtils.stopUnchecked;
+import static org.apache.cassandra.utils.AccordGenerators.fromQT;
+import static org.apache.cassandra.utils.CassandraGenerators.insertCqlAllColumns;
+
+public class AccordHostReplacementTest extends TestBaseImpl
```
dcapwell, line 50: this test is redundant with `TopologyMixupTest`, but it is also super explicit... so still nice to have

test/distributed/org/apache/cassandra/distributed/test/accord/AccordHostReplacementTest.java
```diff
@@ -0,0 +1,104 @@
...
+                                                                  .set("accord.shard_count", "1"));
+        TokenSupplier tokenRing = TokenSupplier.evenlyDistributedTokens(3, clusterBuilder.getTokenCount());
+        int nodeToReplace = 2;
+        clusterBuilder = clusterBuilder.withTokenSupplier((TokenSupplier) node -> tokenRing.tokens(node == 4 ? nodeToReplace : node));
+        try (Cluster cluster = clusterBuilder.start())
+        {
+            fixDistributedSchemas(cluster);
+            init(cluster);
+
+            qt().withSeed(5215087322357346205L).withExamples(1).check(rs -> {
```
dcapwell: should remove the seed

test/distributed/org/apache/cassandra/distributed/test/accord/AccordHostReplacementTest.java
```diff
@@ -0,0 +1,104 @@
...
+                                                .withTableKinds(TableMetadata.Kind.REGULAR)
+                                                .withKnownMemtables()
+                                                //TODO (coverage): include "fast_path = 'keyspace'" override
+                                                .withTransactionalMode(TransactionalMode.full)
+                                                .build())
+                                         .next(rs);
+                maybeCreateUDTs(cluster, metadata);
+                cluster.schemaChange(metadata.toCqlString(false, false));
+
+                Gen<ByteBuffer[]> dataGen = fromQT(fromQT(new CassandraGenerators.DataGeneratorBuilder(metadata).build(ignore -> 10)).next(rs));
```
dcapwell: now that I have `ast.Statement` I could replace this and handle far more possible txn CQL

test/distributed/org/apache/cassandra/distributed/test/accord/AccordHostReplacementTest.java
```diff
@@ -0,0 +1,104 @@
...
+                    cluster.coordinator(1).execute(insertStmt, ConsistencyLevel.ANY, (Object[]) dataGen.next(rs));
+
+                cluster.forEach(i -> i.nodetoolResult("repair", "--full", "--accord-only", KEYSPACE).asserts().success());
+
+                stopUnchecked(cluster.get(nodeToReplace));
+                ClusterUtils.replaceHostAndStart(cluster, cluster.get(nodeToReplace));
+
+                cluster.forEach(i -> {
+                    if (i.isShutdown()) return;
+                    i.nodetoolResult("repair", "--full", "--accord-only", KEYSPACE).asserts().success();
```
dcapwell: I think I added the repairs because I found a bug with `--accord-only`, it isn't required for the test and could be removed... doesn't hurt though.

test/distributed/org/apache/cassandra/distributed/test/accord/AccordTestBase.java
```diff
@@ -373,21 +379,37 @@ private static SimpleQueryResult executeWithRetry0(int count, Cluster cluster, S
...
-            TransactionStatement stmt = AccordTestUtils.parse(cql);
-            return isIdempotent(stmt);
+        return inst.callOnInstance(() -> {
+            CQLStatement.Raw parsed = QueryProcessor.parseStatement(cql);
+            if (parsed instanceof TransactionStatement.Parsed)
+            {
+                TransactionStatement stmt = (TransactionStatement) parsed.prepare(ClientState.forInternalCalls());
+                return isIdempotent(stmt);
+            }
+            else if (parsed instanceof ModificationStatement.Parsed)
```
dcapwell, line 482: CAS queries gets converted to Accord, so we still want to support...

test/distributed/org/apache/cassandra/distributed/test/cms/SafeShutdownTest.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.cassandra.distributed.test.TestBaseImpl;
+import org.apache.cassandra.tcm.Epoch;
+import org.apache.cassandra.tcm.PaxosBackedProcessor;
+import org.apache.cassandra.tcm.Transformation;
+import org.apache.cassandra.tcm.log.Entry;
+import org.apache.cassandra.utils.Shared;
+
+import static net.bytebuddy.matcher.ElementMatchers.named;
+
+public class SafeShutdownTest extends TestBaseImpl
```
dcapwell: ill remove, I created this for Sam and forgot to remove.  Its for the TCM hanging bug

test/unit/org/apache/cassandra/cql3/RandomSchemaTest.java
```diff
@@ -80,38 +73,32 @@ public void test()
...
             TypeGenBuilder withoutUnsafeEquality = AbstractTypeGenerators.withoutUnsafeEquality()
-                                                                         .withUserTypeKeyspace(KEYSPACE)
                                                                          .withUDTNames(udtName);
             TableMetadata metadata = new TableMetadataBuilder()
                                      .withKeyspaceName(KEYSPACE)
                                      .withTableKinds(TableMetadata.Kind.REGULAR)
                                      .withKnownMemtables()
                                      .withDefaultTypeGen(AbstractTypeGenerators.builder()
                                                                                .withoutEmpty()
-                                                                               .withUserTypeKeyspace(KEYSPACE)
```
dcapwell, line 104: when creating UDTs we respect the keyspace that is used while creating the table... so no longer need this.  This was also not that safe... recursion is hard...

test/unit/org/apache/cassandra/cql3/ast/And.java
```diff
@@ -0,0 +1,46 @@
...
+
+package org.apache.cassandra.cql3.ast;
+
+import java.util.stream.Stream;
+
+public class And implements Conditional
+{
+    private final Conditional left, right;
+
+    public And(Conditional left, Conditional right)
```
dcapwell, line 27: could always be `List<Conditional>`, open to changing...

test/unit/org/apache/cassandra/cql3/ast/CollectionAccess.java
```diff
@@ -0,0 +1,58 @@
...
+        this.column = column;
+        this.element = element;
+        this.type = type;
+    }
+
+    @Override
+    public void toCQL(StringBuilder sb, int indent)
+    {
+        column.toCQL(sb, indent);
+        sb.append('[');
```
dcapwell, line 42: I feel like this isn't generic... when the AST was first made some of the collection stuff was not done yet, so may want to relook into this

test/unit/org/apache/cassandra/utils/CassandraGenerators.java
```diff
@@ -825,4 +876,145 @@ public static Gen<DecoratedKey> decoratedKeys(Gen<IPartitioner> partitionerGen,
     {
         return rs -> partitionerGen.generate(rs).decorateKey(keyGen.generate(rs));
     }
+
+    public static String insertCqlAllColumns(TableMetadata metadata)
```
dcapwell: I prob should migrate all usage to the `ast.Statement`, this was kinda a limited hack and switching to the ast made things far better covered

test/unit/org/apache/cassandra/utils/CassandraGenerators.java
```diff
@@ -825,4 +876,145 @@ public static Gen<DecoratedKey> decoratedKeys(Gen<IPartitioner> partitionerGen,
...
+                }
+                else
+                {
+                    pending.add(next);
+                }
+            }
+        }
+    }
+
+    public static class DataGeneratorBuilder
```
dcapwell, line 1240: I prob should migrate all usage to the `ast.Statement`, this was kinda a limited hack and switching to the ast made things far better covered

src/java/org/apache/cassandra/service/accord/AccordSafeCommandStore.java
ifesdjeen: Should we do the same in `mapReduceForKey`? There we also have 

```
for (Key key : commandsForKeys.keySet())
```

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -358,6 +464,79 @@ public synchronized void startup()
...
+        }
+        if (peers.isEmpty())
+            return Collections.emptyList();
+
+        Long minEpoch = findMinEpoch(MessagingService.instance(), peers, optMaxEpoch);
+        if (minEpoch == null)
+            return Collections.emptyList();
+        List<ClusterMetadata> history = new ArrayList<>(Math.toIntExact(current.epoch.getEpoch() - minEpoch));
+        for (long epoch = minEpoch; epoch < current.epoch.getEpoch(); epoch++)
+            history.add(cms.loadHistoricEpoch(Epoch.create(epoch)));
```
ifesdjeen: This approach will only be able to reconstruct epochs from the local log. If we want an arbitrary epoch - we will need to go to the CMS node and ask it to reconstruct it from the distributed log. 

There unfortunately be extremely rare cases where there's an epoch gap (such as during disaster recovery), too.

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -358,6 +464,79 @@ public synchronized void startup()
...
+        if (!localRanges.isEmpty()) // already joined, nothing to see here
+            return Collections.emptyList();
+
+        Map<InetAddressAndPort, Set<TokenRange>> peers = new HashMap<>();
+        for (KeyspaceMetadata keyspace : current.schema.getKeyspaces())
+        {
+            List<TableMetadata> tables = keyspace.tables.stream().filter(TableMetadata::requiresAccordSupport).collect(Collectors.toList());
+            if (tables.isEmpty())
+                continue;
+            DataPlacement placement = current.placements.get(keyspace.params.replication);
```
ifesdjeen, line 479: I _think_ if you are only looking at write placements, you do not need the `whenSettled`, since we first expand, then remove from read, then remove old.

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -688,6 +866,182 @@ public Id nodeId()
         return node.id();
     }
 
+    @Nullable
+    @Override
+    public Long minEpoch(Collection<TokenRange> ranges)
+    {
+        return node.topology().minEpoch();
+//        TxnId txnId = node.nextTxnId(Kind.ExclusiveSyncPoint, Routable.Domain.Range);
```
ifesdjeen: Nit: this probably should be removed (i.e. commented code), but I'm missing context to understand why it might have been needed.

src/java/org/apache/cassandra/service/accord/AccordTopology.java
ifesdjeen: FWIW, we could integrate more closely and have `AccordTopology` inside TCM. I somehow thought this also was a plan at some point.

src/java/org/apache/cassandra/service/accord/FetchMinEpoch.java
```diff
@@ -0,0 +1,150 @@
...
+            List<TokenRange> ranges = new ArrayList<>(size);
+            for (int i = 0; i < size; i++)
+                ranges.add(TokenRange.serializer.deserialize(in, version));
+            return new FetchMinEpoch(ranges);
+        }
+
+        @Override
+        public long serializedSize(FetchMinEpoch t, int version)
+        {
+            int size = TypeSizes.sizeofUnsignedVInt(t.ranges.size());
```
ifesdjeen: nit: should we use `long` for consistency with a return value?

src/java/org/apache/cassandra/service/accord/FetchMinEpoch.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+    };
+    private final Collection<TokenRange> ranges;
+
+    public FetchMinEpoch(Collection<TokenRange> ranges)
+    {
+        this.ranges = ranges;
+    }
+
+    public static Future<Long> fetch(MessageDelivery messaging, Map<InetAddressAndPort, Set<TokenRange>> peers)
```
ifesdjeen: Just to make sure: do we want to just ignore non-successes? Or do we want to retry on failure? I'd say the latter. There's a retry function in TCM, `sendWithCallbackAsync` which can be useful.

src/java/org/apache/cassandra/service/accord/TokenRange.java
```diff
@@ -48,13 +48,28 @@ public TokenRange(AccordRoutingKey start, AccordRoutingKey end)
 
     public TableId table()
     {
-        return ((AccordRoutingKey) start()).table();
+        return getStart().table();
+    }
+
+    public AccordRoutingKey getStart()
```
ifesdjeen: It feels like distinction between `start` and `getStart` is getting very blurry. I suggest to make `start` method non-final and downcast to `AccordRoutingKey` here in override, keeping `start`/`end` names

src/java/org/apache/cassandra/tcm/Startup.java
```diff
@@ -365,21 +367,29 @@ public static void startup(Supplier<Transformation> initialTransformation, boole
...
+        NodeState startingstate = metadata.directory.peerState(self);
+        switch (startingstate)
+        {
+            case REGISTERED:
+            case LEFT:
+                break;
+            default:
+                AccordService.startup(self);
+        }
+        switch (startingstate)
```
ifesdjeen, line 384: I'm a bit confused by this change. Above we're exiting on `REGISTERED` and `LEFT` and then we go back to the `switch` where we do startup on those. It feels like the intention was to only leave the below version in, could you check this out?

test/distributed/org/apache/cassandra/distributed/impl/AbstractCluster.java
ifesdjeen: Super useful! Great idea.

test/distributed/org/apache/cassandra/distributed/impl/INodeProvisionStrategy.java
```diff
@@ -26,140 +26,69 @@
...
-            return create(subnet, null);
-        }
-
-        abstract INodeProvisionStrategy create(int subnet, @Nullable Map<String, Integer> portMap);
     }
 
-    String seedIp();
-
-    int seedPort();
+//    String seedIp();
```
ifesdjeen: Nit: cleanup?

test/distributed/org/apache/cassandra/distributed/test/accord/AccordHostReplacementTest.java
```diff
@@ -0,0 +1,104 @@
...
+                                                .withTransactionalMode(TransactionalMode.full)
+                                                .build())
+                                         .next(rs);
+                maybeCreateUDTs(cluster, metadata);
+                cluster.schemaChange(metadata.toCqlString(false, false));
+
+                Gen<ByteBuffer[]> dataGen = fromQT(fromQT(new CassandraGenerators.DataGeneratorBuilder(metadata).build(ignore -> 10)).next(rs));
+                String insertStmt = wrapInTxn(insertCqlAllColumns(metadata));
+
+                for (int i = 0; i < 10; i++)
```
ifesdjeen: As long as the data is written sequentially you can also reuse Harry machinery from `ConsistentBootstrapTest`, it is super simple: 

```
            ReplayingHistoryBuilder harry = HarryHelper.dataGen(new InJvmSut(cluster),
                                                                new TokenPlacementModel.SimpleReplicationFactor(3),
                                                                SystemUnderTest.ConsistencyLevel.ALL);
            Runnable writeAndValidate = () -> {
                System.out.println("Starting write phase...");
                for (int i = 0; i < WRITES; i++)
                    harry.insert();

                System.out.println("Starting validate phase...");
                harry.validateAll(harry.quiescentLocalChecker());
            };
```

test/distributed/org/apache/cassandra/distributed/test/accord/AccordHostReplacementTest.java
ifesdjeen: I personally would remove it, and switch to writing with `ALL` above, at least if the intention is to test bootstrap streaming.

test/distributed/org/apache/cassandra/distributed/test/topology/TopologyMixupTest.java
```diff
@@ -0,0 +1,856 @@
...
+            {
+                int finalI = i;
+                Awaitility.await()
+                          .atMost(Duration.ofMinutes(1))
+                          .until(() -> HackSerialization.accordEpoch(cluster.get(finalI)) == waitForEpoch);
+            }
+
+            currentEpoch.set(waitForEpoch);
+
+            Gen<Statement> select = (Gen<Statement>) (Gen<?>)  fromQT(new Select.GenBuilder(metadata).withLimit1().build());
```
ifesdjeen: Since we are not testing anything Accord-specific here, and regular mutation statements would do, any reason not to use Harry here just like the rest of `Consistent*Test` do? It validates the placements after streaming by checking data presence even locally, so it has some big advantages.

test/distributed/org/apache/cassandra/distributed/test/topology/TopologyMixupTest.java
```diff
@@ -0,0 +1,856 @@
...
+import org.apache.cassandra.utils.CassandraGenerators;
+import org.apache.cassandra.utils.Isolated;
+import org.apache.cassandra.utils.Shared;
+import org.awaitility.Awaitility;
+
+import static accord.utils.Property.multistep;
+import static accord.utils.Property.stateful;
+import static org.apache.cassandra.utils.AccordGenerators.fromQT;
+
+public class TopologyMixupTest extends TestBaseImpl
```
ifesdjeen: I think this test brings a lot of value, but what I am not seeing is checking if streaming was successful / the data is present on the target nodes. We have a lot of machinery for testing these things in TCM, which can be reused.

Also, we either have to simplify this test a bit and reduce the number of things it does, or refactor it a bit and make it a bit simpler to get overview. Some middle-ground would be to open a Jira for creating a general framework for driving state changes and integrating it with the rest of our fuzz tooling, that would deduplicate efforts. We have done a lot of work to consolidate things in TCM, so it would be great to reuse some of that.

src/java/org/apache/cassandra/service/accord/AccordSafeCommandStore.java
dcapwell: we do already.  That section you showed is actually

```
Routables<?> sliced = keysOrRanges.slice(slice, Routables.Slice.Minimal);
for (Key key : commandsForKeys.keySet()) {
  if (!sliced.contains(key)) continue;
```

we *could* slice the keys but this is a set, so we would first need to sort them, wrap them, then sort them... the contains check is slightly less work?

src/java/org/apache/cassandra/service/accord/AccordTopology.java
dcapwell: I personally prefer the separation... we support the following yaml `accord.enabled: false`, so we won't create `Topology` when disabled, but if we link in TCM we would and would need to also handle that in the `ClusterMetadata`?

test/distributed/org/apache/cassandra/distributed/test/accord/AccordHostReplacementTest.java
dcapwell: > writing with ALL

this is Accord, we don't look at `ConsistencyLevel` =D

src/java/org/apache/cassandra/service/accord/AccordService.java
dcapwell: so, my thinking of putting it in `cms`  was that `ClusterMetadataService` figures out how to do it correctly... From Accord's point of view local log or remote peer doesn't matter (other than the fact its slower)...

I do know that this edge case exists but had no clue how to use CMS well enough to resolve... =D

src/java/org/apache/cassandra/service/accord/AccordService.java
dcapwell: yes, cleanup is desired here...

so, I tried to find the smallest epoch we still cared about, but the logic kept failing.  Talking with @belliottsmith he said to just use `minEpoch`...  This works just fine, but right now we don't *purge* older epochs... there is a ticket for that... right now epochs grow and grow and grow... we do know when an epoch is redundant (all txn in that epoch have executed) so we could always purge... we just need to do that =)

src/java/org/apache/cassandra/service/accord/FetchMinEpoch.java
dcapwell: let me add a TODO and get back to this feedback...

src/java/org/apache/cassandra/service/accord/TokenRange.java
dcapwell: yeah thats what I wanted to do but when I saw `final` I was risk adverse... @belliottsmith are you cool with removing `final`?

test/distributed/org/apache/cassandra/distributed/test/accord/AccordHostReplacementTest.java
dcapwell: we talked in slack about the new AST and how it plays with Harry, trying to make all these fancy things work together is a good idea...

Ill try to take a look... for this tests point of view harry would be fine as we have a simple pattern and no real requirements that would limit harry...

test/distributed/org/apache/cassandra/distributed/test/topology/TopologyMixupTest.java
dcapwell: We talked in Slack, it's mostly the lack of coverage.  Right now `AbstractTypeGenerator` supports every single type  (even super weird ones we hope users never touch, but are not blocked...); so we benefit from border coverage.

The other thing is that the new `ast` was built with Accord in mind (I wrote it to help validate Caleb's first CQL integration path), so it coverts more than harry does.  We spoke in slack today about migrating Harry to use the ASTs rather than `StringBuilder` which will make it much easier for harry to do those things as well (examples missing: type hints, `+=` (super common with CAS and txn), etc.).

Now, one thing I did have in mind when writing this test was to make the Accord part pluggable... I just never got around to doing that...  My thinking was that the schema + statements are accord specific but the rest wasn't, so we could make this test abstract and have one for non-accord and one for accord...  the non-accord one would be a good place to use Harry.  Might even be good to have 4 versions of the test: 2 that validate (using harry) and 2 just fuzzing to see what breaks...

Now, we both agree that long term we should converge, which makes these differences disappear...

test/distributed/org/apache/cassandra/distributed/test/topology/TopologyMixupTest.java
dcapwell: I don't think I fully understand this comment, but I am not at all against reusing existing things to improve validation.

> but what I am not seeing is checking if streaming was successful / the data is present on the target nodes. We have a lot of machinery for testing these things in TCM, which can be reused.

If you can point me to stuff, that would be great.  Accord bootstrap is different than normal, but the streaming part is the same... so reusing shouldn't be too hard to get working for accord tables

> and reduce the number of things it does
you mean topology changes, or number of "steps"?

> or refactor it a bit and make it a bit simpler to get overview.

I don't know what you just said.  "get overview"?  what do you mean by this?

> Some middle-ground would be to open a Jira for creating a general framework for driving state changes and integrating it with the rest of our fuzz tooling, that would deduplicate efforts. We have done a lot of work to consolidate things in TCM, so it would be great to reuse some of that.

If there is existing validation logic from TCM to validate that the change was *correct* I am more than happy to reuse.  Each step is single threaded and exists once everything is complete, so wouldn't be hard to add validation checks afterwards to make sure that the cluster is in the right state... I just don't know what logic exists.

src/java/org/apache/cassandra/service/accord/AccordService.java
ifesdjeen: Right; unfortunately the way log works in TCM is slightly different, so to make it work you would need to:

 * create a special verb for CMS nodes that would receive the epoch and construct CMS for that epoch; alternatively, you ask it for the bottom epoch, and ask it to give you a base state for that epoch, and all subsequent events, so you could reconstruct a range of epochs yourself on the non-cms node.
 *  in this handler, you perform a SERIAL read from distributed metadata table, and reconstruct metadata for given epoch

The problem with the current approach is that local log is only guaranteed to witness highest epoch during specific operations, but, for example, during bootstrap, a new node may receive a snapshot rather than an event log, so reconstructing metadata for the epoch preceeding it will not work.

test/distributed/org/apache/cassandra/distributed/test/topology/TopologyMixupTest.java
ifesdjeen: > If you can point me to stuff, that would be great. Accord bootstrap is different than normal, but the streaming part is the same... so reusing shouldn't be too hard to get working for accord tables

Sure, just take a look at https://github.com/apache/cassandra/blob/trunk/test/distributed/org/apache/cassandra/fuzz/ring/ConsistentBootstrapTest.java#L69-L71

In short, all you need to do is 

```
            ReplayingHistoryBuilder harry = HarryHelper.dataGen(new InJvmSut(cluster),
                                                                new TokenPlacementModel.SimpleReplicationFactor(3),
                                                                SystemUnderTest.ConsistencyLevel.ALL);

                for (int i = 0; i < WRITES; i++)
                    harry.insert();

                harry.validateAll(harry.quiescentLocalChecker());
```

Quiescent local checker (https://github.com/apache/cassandra/blob/42d18a83e4468827da234c68fad3e3058b368d7c/test/harry/main/org/apache/cassandra/harry/sut/injvm/QuiescentLocalStateChecker.java#L37) will infer the placements from the current ring, and will validate that every row is where it should be. In other words, if ownership of a range has changed from node A to node B, it will ensure that the data has moved.

> "get overview"? what do you mean by this?

I think the test itself in general is ok, and the only changes I am proposing are optical / structural, and not functional. Just to make it easier for people approaching it for the first time to navigate and expand it.

This class is currently at almost 900LOC. To get an overview is to be able to glance at the class, and understand what it is doing, how to extend it, or what to do when I am trying to debug an issue with the test failure. Maybe I was overly vague, so let me be more specific. I would do the following things: 
  * Most of the methods that create `SimpleCommand` can be made into static factory methods, such as `repairCommand`, `stopInstance`. They can be moved to the top level of the class, and bunched together. This way, anyone looking at the class can see what commands it offers, and navigate to a specific command that interests them quickly.
  * Some code that creates `SimpleCommand` is now inside a `topologyCommand` method. I think that the method itself does make sense, but I also think that individual commands (i.e. `addNode`, `decomission`, `removeNode`, `assasinate`) can benefit from being moved to their own static methods to be near other operations. `topologyCommand` can simply call those methods
  * Expand some of the abbreviations. For example, I'd use `rng` instead of `rs` since it is more commonly used, and expand `th` into `topologyHistory` or just `history`.
  * I'd move enums such as `TopologyChange` and `RemoveType` up, since they help the person understand what's going on in the class, but auxiliary methods such as `asSet`, `executeTxn` and `maybeCreateUDTs`, since they're incidental to the implementation.

With these changes, the person will be able to look at the signatures to better understand what exactly operations are already implemented, and will be able to pick them up and maybe use them in other contexts, for example for testing bootstrap, or decom, etc. Adding a new operation can then also be more straightforward: a person would simply add a new `static` method specifying the operation and then hook it up to the test quickly.

Please let me know if this was helpful!

src/java/org/apache/cassandra/service/accord/AccordService.java
dcapwell: > create a special verb for CMS nodes that would receive the epoch and construct CMS for that epoch

in my testing this list can be large (but should hopefully shrink once we start pouring), so something that says the min epoch would make the most sense to me (rather than 40 calls, do 1)..

One thing I wonder, couldn't we just have the log pull the rest?  Or is that too annoying for things in the past (as it assumes to more forward)?

src/java/org/apache/cassandra/service/accord/TokenRange.java
belliottsmith: could somebody summarise a bit what's going on here?

src/java/org/apache/cassandra/service/accord/txn/TxnWrite.java
dcapwell: wrote the test `org.apache.cassandra.distributed.test.accord.AccordCQLTestBase#testNonExistingKeyWithStaticUpdate`

test/distributed/org/apache/cassandra/distributed/test/accord/AccordHostReplacementTest.java
dcapwell: spoke with Alex in slack, this might be a better place for Harry, as this class isn't testing the full domain of CQL, just host replacement and needs a load gen

src/java/org/apache/cassandra/service/accord/AccordService.java
dcapwell: thought I replied to this... when the node starts up we need to know how placement changes, this currently does show the nodes leaving.  I think the only way to work around `whenSettled` is to diff write from read, but that feels a bit brittle for me as it makes assumptions on *why* we are in that state.

src/java/org/apache/cassandra/tcm/Startup.java
dcapwell: @ifesdjeen I tried to document this. 

```
// When Accord starts up it needs to check for any historic epochs that it needs to know about (in order
// to handle pending transactions), in order to know what nodes to check with it needs to know what the
// settled placement is (so it knows what peers to reach out to).
```

test/distributed/org/apache/cassandra/distributed/test/accord/AccordHostReplacementTest.java
dcapwell: I made this change in topology mixup test (we now have 2 versions of the test), so wouldn't be hard to do here as well

test/distributed/org/apache/cassandra/distributed/test/topology/TopologyMixupTest.java
dcapwell: I think I addressed all your comments.  I did the refactors requested and see that the code is easier to read (at least for me).

I also pulled the *accord* bits out and made the class `abstract`.  We now have an accord version and a harry version.

The thinking is that Accord wants to also validate we handle weird and random schemas, so the `ast` generators is better then harry there, but in the "how does topology changes work in Cassndra" version harry works better.

test/distributed/org/apache/cassandra/distributed/test/accord/AccordHostReplacementTest.java
dcapwell: harry doesn't support accord timestamps... so harry here fails

```
Caused by: org.apache.cassandra.harry.model.Model$ValidationException: Timestamps in the row state don't match ones predicted by the model:
```

test/distributed/org/apache/cassandra/distributed/test/accord/AccordHostReplacementTest.java
dcapwell: harry doesn't support accord timestamps... so harry here fails

```
Caused by: org.apache.cassandra.harry.model.Model$ValidationException: Timestamps in the row state don't match ones predicted by the model:
```

test/distributed/org/apache/cassandra/distributed/impl/AbstractCluster.java
dcapwell: I think I broke this when I improved the error msg... ill try to fix

test/unit/org/apache/cassandra/utils/CassandraGenerators.java
dcapwell: this is actually used in `RandomSchemaTest`.. the reason for the refactor is no longer needed, but I still prefer this...

test/distributed/org/apache/cassandra/distributed/test/topology/TopologyMixupTest.java
dcapwell: updating more for history than to inform Alex...

Alex and I spoke and worked to get Harry working with Accord... `TopologyMixupTest` got split into 2 tests: with accord / without accord...  the one without accord uses harry for the validation and schema creation.

The accord one still uses this logic as it helps validate the custom CQL that exists for Accord

test/distributed/org/apache/cassandra/distributed/test/topology/TopologyMixupTest.java
dcapwell: @ifesdjeen I think I have address your comment, can you resolve if you agree?  else what is missing?

src/java/org/apache/cassandra/service/accord/FetchMinEpoch.java
dcapwell: Looking at `org.apache.cassandra.tcm.RemoteProcessor#sendWithCallbackAsync` it is a bit CMS specific, but can be expanded.  There is also overlap with `org.apache.cassandra.repair.messages.RepairMessage#sendMessageWithRetries` which is very similar, but also very Repair specific...

Ill see if I can extract both into a reusable pattern

dcapwell: pushed this change.  I did a few refactors:

1) TCM and Repair's logic were super similar but subtly different... I tried to create a new `MessageDelivery.sendWithRetries` method that supported both cases and replaced both
2) epoch fetch now uses this new `MessageDelivery.sendWithRetries`
3) Testing extracted (but didn't modify... this patch is already giant) Repair's and EpochSyncTest simulated cluster logic and made it a reusable class... this let me test the new fetch logic with/without failures in simulation quickly.

src/java/org/apache/cassandra/service/accord/AccordConfigurationService.java
```diff
@@ -350,7 +371,6 @@ void maybeReportMetadata(ClusterMetadata metadata)
     @Override
     public void notifyPostCommit(ClusterMetadata prev, ClusterMetadata next, boolean fromSnapshot)
     {
-        maybeReportMetadata(prev);
```
dcapwell, line 353: this caused a bug where the first epoch (epoch=5) would add a older epoch (epoch=4) and we wouldn't have anything sync it... so rely on TCM only for `next`

src/java/org/apache/cassandra/tcm/RemoteProcessor.java
```diff
@@ -200,51 +203,27 @@ public static <REQ, RSP> RSP sendWithCallback(Verb verb, REQ request, CandidateI
 
     public static <REQ, RSP> void sendWithCallbackAsync(Promise<RSP> promise, Verb verb, REQ request, CandidateIterator candidates, Retry retryPolicy)
     {
-        class Request implements RequestCallbackWithFailure<RSP>
```
dcapwell: @ifesdjeen if you want I can revert, I tried to make sure the subtle semantics did match (other than allocation of callback), but figured could have all 3 usage use 1 thing that has tests rather than all 3 locations doing custom logic

test/distributed/org/apache/cassandra/distributed/shared/ClusterUtils.java
```diff
@@ -572,25 +574,44 @@ private static class ClusterMetadataVersion
     {
         public final int node;
         public final Epoch epoch;
-        public final Epoch replicated;
```
dcapwell: @ifesdjeen I couldn't figure out why this was present... its the same as `epoch` and only used in one code path... so removing made the error string easier to read... I can revert if you want

test/distributed/org/apache/cassandra/distributed/shared/ClusterUtils.java
```diff
@@ -611,8 +632,8 @@ public static void waitForCMSToQuiesce(ICluster<IInvokableInstance> cluster, Epo
                 if (cluster.get(j).isShutdown())
                     continue;
                 Epoch version = getClusterMetadataVersion(cluster.get(j));
-                if (!awaitedEpoch.equals(version))
-                    notMatching.add(new ClusterMetadataVersion(j, version, getClusterMetadataVersion(cluster.get(j))));
+                if (!awaitedEpoch.equals(version) && version.getEpoch() < awaitedEpoch.getEpoch())
```
dcapwell: not sure if this breaks anything on your end, but I found that I would see epoch=33 then fail because every node is on epoch=34... so I use this more as a min wait barrier than anything else...

test/distributed/org/apache/cassandra/distributed/test/accord/AccordCQLTestBase.java
```diff
@@ -79,6 +80,27 @@ public static void setupClass() throws IOException
         SHARED_CLUSTER.schemaChange("CREATE TYPE " + KEYSPACE + ".person (height int, age int)");
     }
 
+    @Override
+    protected void test(FailingConsumer<Cluster> fn) throws Exception
```
dcapwell: one test called this which used `full`, so was missing coverage

test/unit/org/apache/cassandra/io/IVersionedSerializers.java
```diff
@@ -0,0 +1,39 @@
...
+
+package org.apache.cassandra.io;
+
+import java.io.IOException;
+
+import org.apache.cassandra.io.util.DataInputBuffer;
+import org.apache.cassandra.io.util.DataOutputBuffer;
+import org.assertj.core.api.Assertions;
+
+public class IVersionedSerializers
```
dcapwell: this has to already exist... right?

src/java/org/apache/cassandra/index/sai/utils/IndexTermType.java
```diff
@@ -745,7 +745,7 @@ private AbstractType<?> collectionCellValueType(AbstractType<?> type, IndexTarge
                         return CompositeType.getInstance(collection.nameComparator(), collection.valueComparator());
                 }
             default:
-                throw new IllegalArgumentException("Unsupported collection type: " + collection.kind);
+                throw new IllegalArgumentException("Unsupported collection type: " + collection.kind + "; index type: " + indexType.name());
```
dcapwell, line 748: unrelated to the patch, more a product of the fuzz testing...

`RandomSchemaV2Test` now does far more randomness than before, includes accord, *and* tests indexes...

This test is failing due to accord blocking indexes (which is a bug)... Celeb and Ariel are looking into this

src/java/org/apache/cassandra/schema/KeyspaceMetadata.java
```diff
@@ -84,6 +85,12 @@ private KeyspaceMetadata(String keyspaceName, Kind kind, KeyspaceParams params,
         this.replicationStrategy = AbstractReplicationStrategy.createReplicationStrategy(keyspaceName, params.replication);
     }
 
+    @VisibleForTesting
+    public static KeyspaceMetadata createUnsafe(String keyspaceName, Kind kind, KeyspaceParams params, Tables tables, Views views, Types types, UserFunctions functions)
```
dcapwell, line 89: used by fuzz testing....

src/java/org/apache/cassandra/schema/ReplicationParams.java
```diff
@@ -63,6 +63,11 @@ private ReplicationParams(Class<? extends AbstractReplicationStrategy> klass, Ma
         this.options = ImmutableMap.copyOf(options);
     }
 
+    public static ReplicationParams copy(AbstractReplicationStrategy strategy)
```
dcapwell, line 66: used by fuzz testing

src/java/org/apache/cassandra/service/accord/AccordConfigurationService.java
```diff
@@ -293,27 +303,54 @@ synchronized void updateMapping(ClusterMetadata metadata)
...
-    private synchronized void onNodesRemoved(long epoch, Set<Node.Id> removed)
+    public synchronized void onNodesRemoved(long epoch, Set<Node.Id> removed)
     {
+        if (removed.isEmpty()) return;
         syncPropagator.onNodesRemoved(removed);
         for (long oldEpoch : nonCompletedEpochsBefore(epoch))
         {
             for (Node.Id node : removed)
-                receiveRemoteSyncComplete(node, oldEpoch);
+                receiveRemoteSyncCompletePreListenerNotify(node, oldEpoch);
```
dcapwell, line 353: doing remote sync complete would double count as `onRemoveNodes` would also ack for the node... we need this to update disk state

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -350,6 +374,7 @@ private AccordService(Id localId)
         Invariants.checkState(localId != null, "static localId must be set before instantiating AccordService");
         logger.info("Starting accord with nodeId {}", localId);
         AccordAgent agent = FBUtilities.construct(CassandraRelevantProperties.ACCORD_AGENT_CLASS.getString(AccordAgent.class.getName()), "AccordAgent");
+        agent.setNodeId(localId);
```
dcapwell, line 377: this was added for testing.  `AccordTopologyMixupTest` defines its own `AccordAgent` so it can provide debug details for timed out txns

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -437,12 +557,14 @@ public IVerbHandler<? extends Request> verbHandler()
             if (cause instanceof Timeout)
             {
                 TxnId txnId = ((Timeout) cause).txnId();
+                ((AccordAgent) node.agent()).onFailedBarrier(txnId, keysOrRanges, cause);
```
dcapwell, line 571: we kept having issues with barriers, and the root cause was the following:

1) Cassandra does not do durability scheduling (last I heard its still broken)
2) sync points don't use the same epoch discovery logic as normal txn, they rely on `ExclusiveSyncPoint`s, so without those running we kinda include every single epoch when a barrier runs... and when host replacements happen older epochs are no longer safe to touch (shards can't reach quorum)...  this has been improved in this patch, but these hooks are here for tests to provide useful error msg

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -973,6 +1096,30 @@ private static CommandStoreTxnBlockedGraph.TxnState populate(CommandStoreTxnBloc
         return cmdTxnState.build();
     }
 
+    @Nullable
+    @Override
+    public Long minEpoch(Collection<TokenRange> ranges)
+    {
+        return node.topology().minEpoch();
```
dcapwell, line 1114: Benedict said not to be fancy and just return the min known.... we do *not* evict epochs atm, so this is old in long running clusters, but the argument is that we will be able to purge once durability scheduling is working... we need the older epochs to be `redundant` (aka nothing is ever going to touch them) so we can purge...

TL;DR - atm this is returning very old epochs, but once other things come into play this will return more "recent" epochs.  That something is out of scope for this work

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -973,6 +1096,30 @@ private static CommandStoreTxnBlockedGraph.TxnState populate(CommandStoreTxnBloc
...
+    }
+
+    @Override
+    public void tryMarkRemoved(Topology topology, Id target)
+    {
+        if (node.commandStores().count() == 0) return; // when starting up stores can be empty, so ignore
+        Ranges ranges = topology.rangesForNode(target);
+        if (ranges.isEmpty()) return;
+        tryMarkRemoved(ranges, 0).begin(node().agent());
```
dcapwell, line 1123: we don't need to wait on these sync points... we just need to schedule them and hope for the best... its kinda "yolo" semantics until we have something working that is dealing with `ExclusiveSyncPoints`

src/java/org/apache/cassandra/tcm/Processor.java
```diff
@@ -74,4 +78,20 @@ default ClusterMetadata fetchLogAndWait(Epoch waitFor)
     ClusterMetadata fetchLogAndWait(Epoch waitFor, Retry.Deadline retryPolicy);
 
     LogState reconstruct(Epoch lowEpoch, Epoch highEpoch, Retry.Deadline retryPolicy);
+
+    default List<ClusterMetadata> reconstructFull(Epoch lowEpoch, Epoch highEpoch)
```
dcapwell, line 82: `LogState reconstruct`  was added by @ifesdjeen for this patch (though he added in a different ticket) but `LogState` is repetitive to work with in our context and we just want the `List<ClusterMetadata>`, so added this here to simplify Accord logic

test/distributed/org/apache/cassandra/fuzz/topology/AccordTopologyMixupTest.java
```diff
@@ -0,0 +1,397 @@
...
+
+    private static Spec createSchemaSpec(RandomSource rs, Cluster cluster)
+    {
+        TransactionalMode mode = rs.pick(TRANSACTIONAL_MODES);
+        boolean enableMigration = allowsMigration(mode) && rs.nextBoolean();
+        TableMetadata metadata = fromQT(new CassandraGenerators.TableMetadataBuilder()
+                                        .withKeyspaceName(KEYSPACE)
+                                        .withTableKinds(TableMetadata.Kind.REGULAR)
+                                        .withKnownMemtables()
+                                        //TODO (coverage): include "fast_path = 'keyspace'" override
```
dcapwell, line 104: this is a method now in `https://issues.apache.org/jira/browse/CASSANDRA-18675` and found bugs with our fast path serializers...

test/distributed/org/apache/cassandra/fuzz/topology/AccordTopologyMixupTest.java
```diff
@@ -0,0 +1,397 @@
...
+        operations.put(txn, 1);
+        Gen<Statement> statementGen = Gens.oneOf(operations);
+        return (rs, state) -> cqlOperation(rs, state, statementGen);
+    }
+
+    private static Command<State<Spec>, Void, ?> cqlOperation(RandomSource rs, State<Spec> state, Gen<Statement> statementGen)
+    {
+        Statement stmt = statementGen.next(rs);
+        String cql;
+        //TODO (usability): are there any transaction_modes that actually need simple mutations/select to be wrapped in a BEGIN TRANSACTION?  If not then this logica can be simplified
```
dcapwell, line 138: writes are now supposed to be transparent but reads are not... I do think this is also `transaction_mode` explicit... since we are single threaded it would be good to delegate but still need the read side

test/distributed/org/apache/cassandra/fuzz/topology/AccordTopologyMixupTest.java
```diff
@@ -0,0 +1,397 @@
...
+        {
+            super(rs, AccordTopologyMixupTest::createSchemaSpec, AccordTopologyMixupTest::cqlOperations);
+
+            SharedState.listeners.add(this);
+        }
+
+        @Override
+        protected void onConfigure(IInstanceConfig c)
+        {
+            c.set("accord.shard_count", 1)
```
dcapwell, line 218: one shard makes debugging easier, no other reason

test/distributed/org/apache/cassandra/fuzz/topology/AccordTopologyMixupTest.java
```diff
@@ -0,0 +1,397 @@
...
+            {
+                Exhausted e = (Exhausted) failure;
+                SharedState.debugTxn(self.id, "Bootstrap#" + phase, e.txnId().toString());
+            }
+            super.onFailedBootstrap(phase, ranges, retry, failure);
+        }
+    }
+
+    @Isolated
+    public static class Json
```
dcapwell: think this is dead code now... with the new vtable to inspect txn we don't need this...

test/distributed/org/apache/cassandra/fuzz/topology/TopologyMixupTestBase.java
```diff
@@ -131,9 +131,9 @@ private enum RemoveType
                                    state -> ClusterUtils.waitForCMSToQuiesce(state.cluster, state.cmsGroup));
     }
 
-    private Command<State<S>, Void, ?> stopInstance(int toRemove)
+    private Command<State<S>, Void, ?> stopInstance(int toRemove, String why)
```
dcapwell, line 134: this was a logging bug... `nodetool removenode` also calls this method, but we would log that it was for `assassinate`!

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -521,6 +535,27 @@ protected static void addVirtualKeyspace()
         VirtualKeyspaceRegistry.instance.register(SystemViewsKeyspace.instance);
     }
 
+    protected void clearSchema()
```
dcapwell, line 538: most of my tests start with this pattern

```
qt().check(rs -> {
  clearSchema();
  ...
});
```

The `check` block gets called 1k times by default.... so needed a simple way to reset all state

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -2121,7 +2156,10 @@ public static void assertRows(UntypedResultSet result, Object[]... rows)
                 if (!((cellValidator == null && actualValue == null) || (cellValidator != null && cellValidator.equals(actualValue))))
                 {
                     Object actualValueDecoded = actualValue == null ? null : column.type.getSerializer().deserialize(actualValue);
-                    if (!Objects.equal(expected != null ? expected[j] : null, actualValueDecoded))
+                    Object expectedValueDecoded = expected != null ? expected[j] : null;
```
dcapwell, line 2159: honestly its been way too long I don't remember why this was done... this patch is from march... my brain struggles with last week =D

test/unit/org/apache/cassandra/cql3/CQLTester.java
```diff
@@ -2551,6 +2589,116 @@ protected void assertUnauthorizedQuery(String errorMessage, String query, Object
                                   values);
     }
 
+    protected CassandraGenerators.KeyspaceMetadataBuilder createKeyspaceMetadataBuilder()
```
dcapwell, line 2592: these methods are used by `RandomSchemaV2Test`

test/unit/org/apache/cassandra/cql3/RandomSchemaV2Test.java
```diff
@@ -0,0 +1,326 @@
...
+import org.apache.cassandra.transport.ProtocolVersion;
+import org.apache.cassandra.utils.AbstractTypeGenerators;
+import org.apache.cassandra.utils.CassandraGenerators;
+import org.assertj.core.api.Assertions;
+import org.awaitility.Awaitility;
+
+import static accord.utils.Property.qt;
+import static java.lang.String.format;
+
+public class RandomSchemaV2Test extends CQLTester
```
dcapwell, line 67: this class isn't really related to this PR... this PR had things that this needed...

The history of this test is that I am giving a presentation at Community Over Code on property testing and I show how to convert the following test into a property test

```
createTable("CREATE TABLE %s (pk int, v int, PRIMARY KEY (pk))")
execute("INSERT INTO %s (pk, v) VALUES (?, ?)", 1, 1);
assertRows(execute("SELECT * FROM %s WHERE pk=?", 1),
                    row(1, 1))
```

So I wrote the test... which is this class... I then expanded it to include Accord and Indexes... and now its failing due to Accord and Index issues... but they are not *caused* by this patch... =D

Caleb and Ariel are trying to get Accord to work with Indexes... it was assumed to be simple then turned out not to be...

test/unit/org/apache/cassandra/cql3/RandomSchemaV2Test.java
```diff
@@ -0,0 +1,326 @@
...
+            if (mode == Mode.AccordEnabled)
+            {
+                // enable accord
+                schemaChange(String.format("ALTER TABLE %s WITH " + TransactionalMode.full.asCqlParam(), metadata));
+                metadata = Objects.requireNonNull(Schema.instance.getTableMetadata(ks.name, metadata.name));
+            }
+            Map<ColumnMetadata, CreateIndexDDL> indexedColumns = createIndex(rs, metadata);
+            Mutation mutation = nonTransactionMutation(rs, metadata);
+            if (mode == Mode.AccordEnabled)
+                mutation = mutation.withoutTimestamp().withoutTTL(); // Accord doesn't allow custom timestamps or TTL
```
dcapwell, line 107: we don't actually block `TTL` atm... I have a patch for that... I have too many things open atm...

test/unit/org/apache/cassandra/cql3/ast/CreateIndexDDL.java
```diff
@@ -0,0 +1,201 @@
...
+import java.util.stream.Stream;
+
+import org.apache.cassandra.cql3.ast.Symbol.UnquotedSymbol;
+import org.apache.cassandra.db.marshal.AbstractType;
+import org.apache.cassandra.db.marshal.UTF8Type;
+
+import static org.apache.cassandra.cql3.ast.Elements.newLine;
+
+//TODO (now): replace with IndexMetadata?  Rather than create a custom DDL type can just leverage the existing metadata like Table/Keyspace
+public class CreateIndexDDL implements Element
```
dcapwell, line 40: `IndexMetadata` can't use the different versions for the format (though we can fix) so not sure if its still best to have this... it doesn't hurt at least...

src/java/org/apache/cassandra/service/accord/AccordConfigurationService.java
```diff
@@ -293,27 +303,54 @@ synchronized void updateMapping(ClusterMetadata metadata)
...
-        });
+        }
+        Topology current = isEmpty() ? Topology.EMPTY : currentTopology();
+        reportTopology(topology);
+        // for all nodes removed, or pending removal, mark them as removed so we don't wait on their replies
+        Sets.SetView<Node.Id> removedNodes = Sets.difference(current.nodes(), topology.nodes());
+        if (!removedNodes.isEmpty())
+        {
+            onNodesRemoved(topology.epoch(), removedNodes);
+            for (Node.Id node : removedNodes)
```
dcapwell, line 328: add comment that this is a hack...

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -387,10 +412,48 @@ public synchronized void startup()
...
+        if (!ref.historic.isEmpty())
+        {
+            List<ClusterMetadata> historic = ref.historic;
+            long lastHistoric = ref.historic.get(historic.size() - 1).epoch.getEpoch();
+            // new epochs added while loading... load the deltas
+            for (ClusterMetadata metadata : ClusterMetadataService.instance().processor().reconstructFull(Epoch.create(lastHistoric + 1), Epoch.create(current.epoch.getEpoch() - 1)))
+            {
+                historic.add(metadata);
+                configService.reportMetadataInternal(metadata);
+            }
```
dcapwell: this might be a bug, we already are listening and new epochs would have us... so my parania to block bugs maybe added a bug.

src/java/org/apache/cassandra/service/accord/AccordService.java
```diff
@@ -387,10 +412,48 @@ public synchronized void startup()
...
+            // epochClosed implies no "new" txn can be proposed
+            for (ClusterMetadata m : historic)
+            {
+                Topology t = AccordTopology.createAccordTopology(m);
+                long epoch = t.epoch();
+                for (Id id : t.nodes())
+                    node.onRemoteSyncComplete(id, epoch);
+                //TODO (correctness): is this true?
+                node.onEpochClosed(t.ranges(), t.epoch());
+                node.onEpochRedundant(t.ranges(), t.epoch());
```
dcapwell, line 450: double check with @belliottsmith on this

